PR #20, opened 2026-07-06
Implement Niraj Architecture Feedback

LarsGradion:
Implements the accepted Codex/Niraj architecture-readability review end to end. The DAG (`shared → platform → modules → compose → cmd`) is unchanged — this is legibility work before the module surface grows further.

## What's in here

- **`decisions/0018-module-growth-policy.md`** — the growth rulebook: modules start flat (concept-prefixed files like `offer*.go` are the sanctioned interim shape); a subpackage requires ONE named trigger (external protocol/provider adapter · independently testable engine · policy/ruleset to hide · generated code · stable-API sub-capability); anti-triggers (no noun-splits, no wide exported internal APIs, no sibling-import workarounds); the compose corollary (orchestration only, never a durable business-entity owner).
- **`docs/explanation/authorization.md`** — why the RBAC/row-scope gate sits at store/service entry points below the handlers: REST, MCP, agents, workers, and compose flows share one gate; 403 (object denial) vs 404 (row-scope miss, existence-hiding); an FK reference to a row-scoped record is a read; RLS as the structural backstop.
- **`backend/internal/compose/integration/`** — the cross-module integration suites (61 files) move out of the flat 117-file compose root. ONE shared harness lives in a non-test `harness.go` that imports modules/platform only (never compose), so the white-box suites remaining in the compose root and in `briefs` import it without a cycle. **Test-function count parity proven: 468 before == 468 after** (`go test -tags integration ./... -list`). The genuinely white-box suites (closedate, coldstart, reconcile, report_forecast, scrape, comms, dedupe_budget, preference_agent) stay in the root by design.
- **`backend/internal/compose/briefs/`** — the pilot orchestration split (Morning-Brief rank/score/store/L2/transport); `compose.Server` embeds `briefs.Handlers`; the shared `unfence` model-output trim is rehomed as `modules/ai.Unfence` (one spelling, both callers). Remaining groups (reporting, exports, enrichment, public surface) follow the same recipe later, each only when touched.
- **Gates re-pointed, not weakened** — go-arch-lint's compose component gains the `/**` glob; `bench-perf` targets `./internal/compose/integration`; the write-shape/table-ownership waivers are re-keyed to `compose/briefs`; both fitness walkers now skip `//go:build integration` files (test scaffolding can never reach a shipped binary).
- **AGENTS.md / CLAUDE.md / docs/explanation/architecture.md / STATUS.md** updated (fourteen modules incl. signals, growth-policy pointers).

## Spec-side record (sibling repo, committed on its main as `da05f3e`)

ADR-0054 §3 amendment (**A99**), conventions 11 §2 growth rule, new leaves **B-EP01.17/18/19** marked POCV1 DONE, and `HANDOFF-to-code-session-2026-07-06.md` so no session re-executes them. Backlog validator: 0 cycles/dangling/dupes.

## Verification

- `make check` green (build, vet, lint 0 issues, arch-lint, unit tests incl. the re-keyed fitness gates, contract drift, craft-drift).
- `make test-integration` green (full real-Postgres lane); the lane still fails loudly without a database (verified — the DSN-less run aborts with the run-`make db-up` message).
- Integration test-function count parity 468 == 468; pre-push craft static gate passed on the diff.


LarsGradion:
Superseded: the branch is fully contained in #21, which merged — this PR's diff is empty.

PR #78, opened 2026-07-16
fix(dev): run ad-hoc SQL through the compose postgres container

lstrojny:
`scripts/dev.sh` shelled out to host `psql` for the slugged CREATE DATABASE, the `scripts/seed-dev.sql` apply, and the `--drop` DROP DATABASE — but psql is not a host requirement (hosts need Go + Docker only), so `make dev` failed at the SQL seed step on hosts without a Postgres client.

Route all three call sites through `docker compose exec` on the postgres container via a `psql_owner()` helper, the same pattern `make db-init` already uses. The DSN-based paths (cmd/migrate, the api) are unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved development environment database setup and cleanup reliability.
  * Database creation, seeding, and forced deletion now provide clearer failure handling.


dungbanguyen:
Superseded by #85. This PR was filed from a fork, so GitHub withholds `SONAR_TOKEN` from its CI run and the required `SonarCloud Code Analysis` check can never post. #85 carries the identical commits from an in-repo branch so the scan runs. Closing in favor of #85.

PR #81, opened 2026-07-16
feat(automation): the deterministic automation backend (closed 7×7 catalog, both permission gates, event+clock runtimes)

tinnhlt-gra:
## What

Builds the **deterministic automation backend** (WP1 core spine, AUTO chapter): a **closed 7×7 catalog** of triggers × actions a workspace turns on and parameterizes, plus the runtime that actually fires them — event-driven and clock-driven, both through one governed dispatch path. Automation is extracted into its own module (`internal/modules/automation`, the seventeenth), taking the catalog from 2 entries to the six seeded starter templates.

## Why

Automations must be **legible and bounded**, not a rule engine: ADR-0035 permanently rejects a visual/expression builder. Adding a trigger or action is a **code-and-test change, never data** — the closure test pins the code against the spec's RC-11 vocabulary in both directions, so neither a silent addition nor a silent drop can land. A firing acts **on behalf of its owner**, re-gated against that owner's live authority at fire time — a robot may never carry an action its owner could not perform by hand.

## Changes

- **Closed 7×7 registry** — `TriggerKind` × `ActionType` (the user-facing catalog) mapped onto `workflow.ActionKind` (the executor vocabulary, one layer down). Two deliberately distinct types; a both-direction closure test is the anti-builder guard. `assign_owner` carries the ADR-0026 §3 **dynamic** tier (🟢 owner-scoped, 🟡 at scale).
- **Two entries, one dispatch** — event triggers arrive via the outbox → relay → Redis `cg:workflows` → `HandleEvent`; clock triggers arrive via a River `time_scan` periodic job against an injected clock. Both converge on the shared `runOne`, so the occurrence key and the owner gate apply to both.
- **The occurrence key + the `recordSkip` fix** — clock triggers are continuously true, so their idempotency key is the **anchor timestamp** (last genuine activity / the date). A clock non-match no longer goes through `recordSkip` — which would have claimed the anchor key and made the real firing **silently never run** (the subtlest bug fixed here; the load-bearing test asserts the *first* firing happens).
- **Two permission gates** — an **author-time ceiling** (`ceiling.go`, fast-fail UX when a human creates an automation) and **the security boundary: an owner-bound match-time gate** (`gate.go`) that re-resolves the owner's live **seat + RBAC** via the `authz.Resolver` seam and **blocks** a firing whose owner lost the permission — or was downgraded to a read seat (ADR-0055's "live seat/RBAC", mirroring `platform/auth`'s admission path). Fail-closed on a gone/suspended owner; a transient resolver error retries, never allows. A NULL-owner (system-seeded) automation runs ungated. Run reasons are sanitized — a firing that trips a DB error lands `failed` with a client-safe reason, never a raw provider/SQL string (T2).
- **The 🟡 staging path** — `ApplyActions` stages a 🟡 action through the `Approvals` seam, creating a **real** approval row and parking the run behind its id; rejection lands the run `blocked` with a reason (fixes the AUTO-T05 dead end where a 🟡 firing parked forever with no approval).
- **A no-recipient firing skips, never poisons the event** — a matched trigger whose `Plan` finds nothing to act on (e.g. `stage_change_notify` on an ownerless deal — no recipient to name) now records a `skipped` run with an honest reason (`declinedFiring`), mirroring the no-transport notify skip. It previously returned a hard `Plan` error out of `HandleEvent`; with no dead-letter store yet (`subscriber.go`) that left the bus entry **unacked and re-delivering forever**, poisoning every sibling handler on the same event. A skip is the honest outcome: nothing went wrong and a redelivery meets the same record — and it stays visible in run history without fabricating a recipient.
- **`workflow_run.error` → structured `detail jsonb`** — one column carried two payloads (a failure reason and a staging pointer matched by string equality); now a queryable jsonb (`detail->>'approval_id'`).
- **Six seeded templates** + `assign_lead_owner` (authorable, not seeded). Every catalog `Key` == its handler's `Spec().Name` (the orphan-key invariant, gated by a fitness test). Per AUTO-NOTE-2, "Route new lead to a task" creates a task; the old owner-assignment behaviour moved to `assign_lead_owner` (migration `0077` re-keys live rows).
- **`gen-workflow` scaffold generator** (`backend/tools/gen-workflow`) — scaffolds a handler + test stub, **write-once** (never clobbers).
- **Dev stack** — `cmd/worker` (the only `cg:workflows` consumer + River runner) now always runs in `make dev`, so automations fire locally (reconciled with main's PR #75, which made the same change).
- Docs: `docs/explanation/automation.md`, `docs/how-to/add-an-automation-trigger-or-action.md`; CLAUDE.md/AGENTS.md → seventeen modules.

## Honest scope & known limitations (do not read this as chapter-DONE)

This PR is **backend-only**. Against the chapter UAT (`UAT.md`):
- **S2 (fires, trail explains), S3 (pause/resume/delete), S5 (no-DSL, REST half)** — yes.
- **S4 (agent composer) / AUTO-AC-7** — **unverifiable here**; needs the agent authoring path, out of scope.
- **`renewal_reminder`** ships registered but its candidate enumeration is **deferred** — there is no thin-adapter path to enumerate entities by a *custom* renewal-date field value (`fieldcatalog.Reader` is metadata-only, `datasource.SystemOfRecordProvider` is frozen V1). It is an honest no-op until that seam exists.
- **AUTO-AC-6** ("Sam gets the reminder") is proven through the real River job, but the reminder reaches the owner via **record ownership (row-scope)**, not an assignee-based personal task queue — the created task sets no `assignee_id`, and `GET /activities?assignee_id=` is contract-declared but unwired (a separate, still-planned personal-queue epic).
- **Clock-trigger transient-block latch (a product-intent call for review):** a human-owned *clock* automation whose owner *transiently* loses the permission mid-quiet-spell records a `blocked` run on the stable anchor key, so even after the owner regains access that one occurrence won't fire — it self-heals the next quiet spell. Unlike a silent miss this is **visible in run history** (`blocked` with a reason) and only affects human-owned clock automations (the seeded six are NULL-owner, ungated). Left as-is pending a product decision on whether a transient block should be terminal-per-occurrence.

## Upstream discrepancies (raised, not worked around — P3)

- **AUTO-T04 (author ceiling) and AUTO-T05 (🟡 staging) are marked `done` upstream but were not implemented** — this PR implements both.
- **ADR-0054's generated registration manifest / QG-8 `gen-manifests-check` is not implemented** — the whole repo hand-registers (compose is the compile-time glue); building it for automation alone would be a lone island. Deferred; reconcile upstream.
- **No `liveuat` lane exists** — the repo's live tag `livesmoke` is deliberately manual. AUTO-AC-6 is proven in the CI-gated **integration** lane instead (better than a manual lane).
- **Pre-existing bug:** `stage_change_create_task.Match` reads a `to_semantic` payload field that `deals` emits as `to_status` — its open-move filter never fires. Moved unchanged in the extraction; flagged for a follow-up.
- Migrations renumbered to `0076`/`0077` after main added `0074_system_log`/`0075_audit_log_entity_required`.

## Gates run locally (all green)

`make check` (backend + frontend) · `make test-integration` — `OK: integration passed with 0 skips (15 packages)` · `golangci-lint` baseline **and** `--config .golangci.strict.yml` (0 issues) · `arch-lint` · contract-breaking (green on main's crm.yaml) · craft static diff-scoped (**0 BLOCKER**; 8 advisory MAJOR: 3 long-func, 5 naked-any in the JSON-schema builders) · file-length · rls-store-path · **`make dev` §1**: dev log shows `worker dispatching workflows (cg:workflows)`.

## Manual test guide (for tomorrow's live-stack pass)

> This is deliberately small — only what no automated lane can observe. It is **not** a substitute for the chapter's own `UAT.md`.

1. **Dev stack fires** — `make dev`, then `grep 'worker dispatching workflows' .tmp/dev/*/dev.log` (present = the only `cg:workflows` consumer is running). ✅ pre-verified here.
2. **Dev side-effects survivable** — the always-on worker runs `privacy.RunRetention` + `close_date_sweep` (RunOnStart) at boot; confirm the seeded demo data is intact and close dates aren't rewritten (mitigated by a long `--retention-interval` in dev).
3. **A clock trigger fires against real time** — with a seeded quiet deal and a "no activity for N days" rule enabled, wait one time-scan tick: it fires once and the reminder lands in the owner's work queue; a second tick changes nothing.
4. **`notify` fails honestly** — enable Stage-change notify and trigger a stage change: expect a `workflow_run` with a **non-applied status and a readable reason** ("no notification transport configured"), never a silent nothing or a fake success (this repo wires no notification transport).
5. **The generator** — `cd backend && go run ./tools/gen-workflow my_test_handler && make check`; run it **again** → it refuses to overwrite (write-once). Then delete the scaffold.
6. **The screen still works, now shows the catalog** — open `#/automations` (via Settings): the seeded templates list, pause/resume/delete work, no console errors.


tinnhlt-gra:
## 🧪 Manual verification guide + live test run

Ran the guide below against an **isolated dev stack** (`make dev DEV_SLUG=automation`, clean DB built from this branch — no cross-worktree contamination) via three parallel agents. Results:

| Section | Result |
|---|---|
| §1 catalog closed + seeded | ✅ 8 entries; six seeded pinned by `seed_test.go` (seeded-ness is domain-internal, not on the wire) |
| §2 CRUD + param validation | ✅ create→enable→pause→delete; bad key/param/type → **422** (never 500) |
| §3 event fires **open-only** (the `to_status` fix) | ✅ open move → `fired`; won/lost close → `skipped` (never fires on close); no double-fire |
| §4 preview (dry-run) | ✅ `matches_now`/`would_have_fired` returned; writes nothing |
| §5 🟡 staging | ⏭️ no shipped starter emits a 🟡 action; proven in the integration lane instead |
| §6 permission gates | ⏭️ needs a demoted second user (not run live) |
| §7 `notify` no-transport | ✅ honest `skipped` — "no notification transport configured" |
| §8 clock trigger | ⚠️ inconclusive live (time-scan interval is 1h; no 500 on create/enable) — proven by CI integration test |
| §9 `gen-workflow` write-once | ✅ writes scaffold, refuses on re-run |

No 5xx and no behavioral regressions surfaced. (Aside: the shared `margince` dev DB on `:8080` currently 500s on any authed call because a sibling overlay-connections worktree seeded an `overlay_connection` RBAC object this branch doesn't define — environmental, not this PR; use `DEV_SLUG` isolation or reseed.)

---

# Manual verification — Deterministic Automation Backend (AUTO / WP1)

`curl`-first click-paths for the automation module (`internal/modules/automation`): the closed 7×7
catalog, CRUD + enable/pause, the event and clock runtimes, run history, preview, both permission
gates, and the 🟡 staging path — plus the `stage_change_create_task` open-only fix. This PR is
**backend-only**; a UI step is given where a screen exists (Settings → Automations lists + pause/
resume), curl otherwise. Format mirrors `.tmp/retrieval-intent-tools-ui/manual-guide.md`.

## 0. Bring-up (once per session)

```bash
# from the repo root
make dev              # infra (PG16 + Redis) + api (:8080) + worker + Vite web (:5173)
```

`make dev` now runs **`cmd/worker`** too (the only `cg:workflows` consumer + the River time-scan), so
automations actually fire locally. Confirm it came up:

```bash
grep 'worker dispatching workflows' .tmp/dev/*/dev.log    # present = the consumer is running
```

Seeded workspace + admin (from `scripts/seed-dev.sh`):

- Workspace slug: **`demo-workspace`** (sent as `X-Workspace-Slug` locally; prod uses the subdomain)
- Admin: **`admin@demo.test`** / **`demo-password-123`**

**Session cookie for curl** (every automation endpoint is `human-only`, so the slug header alone is not enough):

```bash
BASE=http://localhost:8080/v1
WS='X-Workspace-Slug: demo-workspace'
SESSION=$(curl -sS -D - -o /dev/null -X POST "$BASE/auth/login" \
  -H "$WS" -H 'Content-Type: application/json' \
  -d '{"email":"admin@demo.test","password":"demo-password-123"}' \
  | sed -n 's/^[Ss]et-[Cc]ookie: crm_session=\([^;]*\).*/\1/p' | tr -d '\r')
COOKIE="crm_session=$SESSION"
curl -s -H "$WS" --cookie "$COOKIE" "$BASE/me" | jq '{user:.user.email, roles}'   # sanity
```

Handy helper for the rest of the guide — `J` posts JSON:

```bash
J(){ curl -s -H "$WS" --cookie "$COOKIE" -H 'Content-Type: application/json' "$@"; }
```

### ⚠ Heads-up: the dev DB is shared across worktrees

`make dev` uses the shared `margince` database on the local Postgres container — and so does every
other worktree pointed at the same container. If another branch (e.g. an overlay-connections
worktree) has seeded RBAC objects THIS branch doesn't define, the policy loader rejects them and
**every authorized call 500s** with a line like:

```
unhandled error … err="crmauth: role \"admin\": policy: unknown object \"overlay_connection\" in permissions document" status=500
```

That is **not** an automation bug — it's a cross-worktree dev-DB collision (this branch's
`coreObjects` doesn't define that object). Two ways out:

```bash
# A) reset the shared demo workspace so its roles come from THIS branch's policy (keeps :8080)
make seed-reset && make dev
# B) or run THIS worktree on its own isolated DB (recommended when >1 worktree is active) —
#    note the API port shifts; read the printed `api http://localhost:PORT` line and set BASE to it.
make dev DEV_SLUG=automation
```

The integration lane (`make test-integration`) is immune — it builds a throwaway DB from this
branch's own migrations + seed, which is why CI stays green while a contaminated dev DB 500s.

---

## 1. The catalog is closed and seeded (7 triggers × 7 actions)

**curl:**

```bash
J "$BASE/automations/catalog" | jq '.data[] | {key, name, trigger, action, tier}'
# the closed set (8 entries today). Each carries key/name/description/trigger/action/tier/params_schema.
# NOTE: "seeded" (which templates auto-enrol on a fresh workspace) is a DOMAIN-INTERNAL flag pinned by
# seed_test.go to exactly six — it is deliberately NOT on the wire, so there is no `seeded` field here.

# the six seeded templates already exist as enabled/paused instances in the demo workspace:
J "$BASE/automations" | jq '.data[] | {id, key, name, status}'
```

**UI:** Settings → **Automations** lists the same catalog and the workspace's instances, each with a
status toggle. No console errors.

**Closure guard (code):** the 7×7 set is pinned in both directions by `catalog_closure_test.go` —
adding/removing a member without updating the spec-pinned list fails `make check`.

---

## 2. CRUD + instantiation & param validation

`createAutomation` always creates **paused** — enabling is a deliberate second step.

```bash
# create a stage_change_create_task instance (authorable, not seeded) — starts PAUSED
AID=$(J -X POST "$BASE/automations" \
  -d '{"key":"stage_change_create_task","name":"QA: next-step task","params":{"due_in_days":2}}' \
  | jq -r '.id')
J "$BASE/automations/$AID" | jq '{id, key, status}'            # status: "paused"

# enable it (PATCH flips status; there is no separate enable/pause verb)
J -X PATCH "$BASE/automations/$AID" -d '{"status":"enabled"}' | jq '{status}'   # "enabled"

# pause / re-enable round-trips the same way; delete removes it
# J -X PATCH "$BASE/automations/$AID" -d '{"status":"paused"}' | jq '{status}'
# J -X DELETE "$BASE/automations/$AID" -o /dev/null -w '%{http_code}\n'   # 204
```

> If that enable PATCH returns **500** with `unknown object … in permissions document`, that's the
> shared-DB collision from §0 (not the automation) — reset the DB or use `DEV_SLUG`.

Leave this instance **enabled** for §3. (UI: the same toggle lives on the Automations screen.)

**Automations are instantiated, not defined, over the API.** The catalog is code (§1); `POST
/automations` only stamps a per-workspace row against a **closed** `key` + `params` — you cannot
create a new automation *type* over the wire. A bad recipe is a **422**, never a 500:

```bash
# unknown catalog key → 422 "not a catalog automation type"
J -o /dev/null -w '%{http_code}\n' -X POST "$BASE/automations" -d '{"key":"make_coffee","name":"x","params":{}}'   # 422
# unknown param key → 422 "params.foo: not a parameter"
J -X POST "$BASE/automations" -d '{"key":"stage_change_create_task","name":"x","params":{"foo":1}}' | jq '.detail // .title'
# wrong param type → 422 "params.due_in_days: must be an integer"
J -X POST "$BASE/automations" -d '{"key":"stage_change_create_task","name":"x","params":{"due_in_days":"soon"}}' | jq '.detail // .title'
```

Params are validated against the catalog entry's own `Validate` (single-knob starters and
`assign_lead_owner`'s nested rule schema alike), so a mistyped recipe is caught at save — never stored
as a blob the engine chokes on at fire time. (Contrast the §0 500: that's an infra/RBAC-policy error,
a different layer entirely from param validation.)

---

## 3. ⭐ Event automation fires — and ONLY on open moves (the fix under review)

`stage_change_create_task` mints a follow-up task on an **open** stage move and must **not** on a
won/lost close. (This is the `to_semantic`→`to_status` correctness fix: before it, the filter was
always-true and closed deals got a spurious task.)

```bash
# pick a seeded deal and its pipeline's stages (note each stage's semantic)
DID=$(J "$BASE/deals?limit=1" | jq -r '.data[0].id')
PIPE=$(J "$BASE/deals/$DID" | jq -r '.pipeline_id')
J "$BASE/stages?pipeline_id=$PIPE" | jq '.data[] | {id, name, semantic}'
OPEN_STAGE=$(J "$BASE/stages?pipeline_id=$PIPE" | jq -r '[.data[]|select(.semantic=="open")][0].id')
WON_STAGE=$(J  "$BASE/stages?pipeline_id=$PIPE" | jq -r '[.data[]|select(.semantic=="won")][0].id')

# (A) OPEN move → should fire
J -X POST "$BASE/deals/$DID/advance" -d "{\"to_stage_id\":\"$OPEN_STAGE\"}" -o /dev/null -w '%{http_code}\n'
sleep 2   # let the bus deliver to cg:workflows

# (B) WON move → should NOT fire (a close ends the cadence)
J -X POST "$BASE/deals/$DID/advance" -d "{\"to_stage_id\":\"$WON_STAGE\"}" -o /dev/null -w '%{http_code}\n'
sleep 2

# run history: expect ONE fired (the open move) and ONE skipped (the won close)
J "$BASE/automations/$AID/runs" | jq '.data[] | {outcome, trigger_evidence, action_result, reason}'
```

**Expect:** the open move → `outcome:"fired"`, `action_result` naming a created task; the won move →
`outcome:"skipped"` (not `fired`). Cross-check the timeline: exactly **one** "Plan the next step…"
task was created for this deal, not two.

```bash
# The per-instance run outcomes above are the authoritative check (one fired, one skipped). There is
# no /deals/{id}/activities route — list activities and filter by subject (the list projection doesn't
# expose deal linkage, so this is a workspace-wide count):
J "$BASE/activities?limit=50" | jq '[.data[]|select(.subject|startswith("Plan the next step"))]|length'
# On a shared/re-used dev DB this may read >1 if leftover QA instances from other runs also fired — the
# per-instance run history (GET /automations/{id}/runs), not this global count, is the source of truth.
```

> Regression guard in code: `TestStageChangeCreateTaskMatchOnlyFiresOnOpenMoves`
> (`handlers_event_test.go`) + the `workflow_integration_test.go` end-to-end assertion, both now
> feeding the real `to_status` payload deals emits.

---

## 4. Preview — the designer's dry-run (🟢 read, no writes)

```bash
J -X POST "$BASE/automations/$AID/preview" -d '{"window_days":30}' \
  | jq '{matches_now, would_have_fired, window_days, sample, excluded_by_permission}'
# matches_now = open deals right now; would_have_fired = open stage moves in the last 30d.
```

Preview writes **nothing** — confirm the run count is unchanged after previewing:

```bash
J "$BASE/automations/$AID/runs" | jq '.data | length'   # same as before the preview call
```

A draft override previews an unsaved recipe (`{"key":"…","params":{…}}`); an unsupported type
(e.g. `no_activity_reminder`) returns an honest **422** naming why, never a wrong scope.

---

## 5. The 🟡 staging path — a held action creates a real approval

When a firing plans a 🟡 action it stages instead of executing: the run parks in
`queued_for_approval`, a **real** approval row lands in the Inbox, and **rejecting** it flips the run
to `blocked` (matched back by the approval id on the run's `detail`).

**Honest scope:** *no shipped starter emits a 🟡 action yet* — `assign_owner` resolves 🟢
single-entity (no bulk scope is wired), and nothing plans `request_approval`/`send_email`. So the
staging→reject→`blocked` loop is **proven end-to-end in the integration lane**
(`workflow_approval_staging_integration_test.go` over a real Postgres), not by a click-path here.

To watch it live, scaffold a one-off handler whose `Plan` returns a `request_approval`
(`ActionEmitFlowEvent`) action, register it, enable an instance, trigger it, then reject the staged
approval and re-read `GET /automations/{id}/runs` — `outcome` goes `queued_for_approval` → `blocked`.
This is a code step, so most testers should rely on the integration proof above.

---

## 6. Both permission gates

**Author-time ceiling** (fast-fail at creation): as a user who *cannot* perform the effect by hand,
creating that automation is refused up front.

```bash
# as a read-only / limited user (log in a second session), attempt to create an assign_owner-style
# automation → expect 403/422 naming the missing permission, never a silent save.
```

**Match-time owner gate** (the real boundary, on every firing): a human-authored automation acts on
behalf of its `owner_id`. Demote the owner (revoke the needed role / drop them to a read seat), then
trigger the automation → the firing is refused and lands as a durable **blocked** run:

```bash
J "$BASE/automations/$AID/runs" | jq '.data[] | select(.outcome=="blocked") | .reason'
# e.g. "the automation's owner no longer has a seat that can make changes"
#      "owner's authority no longer permits update on deal"
```

The six seeded templates are NULL-owner (system-seeded) and run **ungated** — only human-authored
instances carry an owner to re-check.

---

## 7. `notify` with no transport → honest skip (not a fake success)

This repo wires no notification channel, so a matched `notify` firing records a visible `skipped`
run with a readable reason rather than silently doing nothing:

```bash
SN=$(J -X POST "$BASE/automations" -d '{"key":"stage_change_notify","name":"QA: notify","params":{}}' | jq -r '.id')
J -X PATCH "$BASE/automations/$SN" -d '{"status":"enabled"}' -o /dev/null -w '%{http_code}\n'
# advance the deal (§3 has the stage-move curl), then read the run:
J "$BASE/automations/$SN/runs" | jq '.data[0] | {outcome, reason}'
# outcome:"skipped", reason:"no notification transport configured"
```

(stage_change_notify's Plan reads the deal's owner as the recipient, so use a deal that has an owner;
an owner-less deal is a `skipped` run with a different reason — "no assigned owner to notify".)

---

## 8. Clock trigger fires against real time (time-scan)

`no_activity_reminder` has no event — the River `time_scan` job sweeps stale records. With a seeded
deal that has had no genuine activity for N days and the automation enabled, one scan tick mints a
reminder task; a second tick changes nothing (the occurrence key is the last-activity **anchor**, not
a per-pass id).

```bash
# enable no_activity_reminder (params: {"no_activity_days": 7})
NID=$(J -X POST "$BASE/automations" -d '{"key":"no_activity_reminder","name":"QA: quiet","params":{"no_activity_days":7}}' | jq -r '.id')
J -X PATCH "$BASE/automations/$NID" -d '{"status":"enabled"}' -o /dev/null -w '%{http_code}\n'

# wait one time-scan interval (dev interval is short; watch the worker log), then:
grep 'time-scan' .tmp/dev/*/dev.log | tail
J "$BASE/automations/$NID/runs" | jq '.data[] | {outcome, trigger_evidence, action_result}'
# expect one fired run + a reminder task on the stale deal's owner's queue;
# a second tick adds NO new run for the same anchor.
```

> Proven in CI by the real-River integration test (`no_activity_reminder_workqueue_integration_test.go`),
> so this UI/dev pass is a live-stack confirmation, not the only evidence.

---

## 9. The scaffold generator is write-once

```bash
cd backend
go run ./tools/gen-workflow qa_probe_handler         # writes handlers_qa_probe_handler.go + _test.go
go run ./tools/gen-workflow qa_probe_handler          # re-run → refuses (write-once), non-zero exit
rm internal/modules/automation/handlers_qa_probe_handler*.go   # clean up the scaffold
```

The generator emits `handlers_<name>.go` (the module's handler-file convention) and prints the manual
next-steps (add a `Catalog()` entry, register in `StarterWorkflows()`); it deliberately does **not**
wire the handler for you — there is no generated manifest.

---

## Deferred / not-a-gap (recorded, not built this PR)

- **S4 / AUTO-AC-7 — agent-composed automations.** The NL "build me an automation" authoring path
  needs the agent-run surface, out of scope here; verify once that lands.
- **`renewal_reminder` candidate enumeration.** Handler is registered and correct, but the time-scan
  has no candidate source for an arbitrary `cf_*` renewal-date column yet — it is an honest no-op until
  that seam exists (see `handlers_clock.go`).
- **Personal task queue (AUTO-AC-6).** Reminders reach the owner via record row-scope, not an
  assignee-based "my tasks" queue (`GET /activities?assignee_id=` is contract-declared but unwired) —
  a separate planned epic.
- **No notification transport.** §7's skip is the intended state until a channel adapter is wired.

PR #90, opened 2026-07-17
feat: single-organization installation — config-driven bootstrap, honest login, password reset (ADR-0061/A107)

LarsGradion:
Implements the ratified single-organization installation concept (margince-foundation ADR-0061 / DECISIONS A107 / ADR-0043 Amendment 2) end to end. Contract-first: the spec-side ratification merged first (margince-foundation#1059); this PR is the build-repo re-sync + implementation.

## Backend

- **Contract re-sync** (`backend/api/crm.yaml`, deliberate breaking change — see CI note): `POST /workspaces` + `BootstrapWorkspaceRequest` removed; the `{workspace}` subdomain server template replaced by a single installation host; `LoginRequest.mfa_code` and the 403 `mfa_required`/`sso_enforced` states removed (unbuilt challenge flows are not advertised); anonymous **`GET /auth/capabilities`** added; the A74 account-recovery pair (`/auth/forgot-password`, `/auth/reset-password`) added and implemented.
- **`platform/deployconfig`**: strict `margince.yaml` loader — organization, bootstrap_admin (secrets as `*_file` references), **seeds** (pipeline stages, consent purposes, starter automations, booking page — every key optional, omitted = built-in default), auth, email/SMTP. Unknown keys are boot errors.
- **Boot state machine** (`identity.BootstrapInstallation` under a pg advisory lock): 0 active workspaces → create organization + admin + roles + configured seeds atomically (no session minted); 1 → bind; >1 → refuse for an operator-led migration. Enforced at boot, **not** as a schema constraint — the cross-tenant RLS suites keep inserting a second workspace directly.
- **Singleton binding per request**: the middleware resolves the cached singleton for every request; `X-Workspace-Slug` and subdomain tenant selection are gone; pre-bootstrap requests answer 503 (availability, never an auth failure). `cmd/mcp` loses `--workspace` and refuses to start pre-bootstrap (no human exists yet who could have granted a passport).
- **Password reset (A74)**: `auth_token` migration (hash-only, composite tenant-local FK, FORCE RLS), STARTTLS-required SMTP mailer, enumeration-resistant request (async send — response timing discloses nothing), single-use ~1h token whose redemption swaps the hash, clears the §27 lockout, and revokes every session. `cmd/migrate reset-password --email` is the operator-only recovery path (password on stdin). The capabilities probe reports `password_reset` from the same wiring.
- `make dev` writes a demo `margince.yaml` + password file and passes `--config`; `seed-dev.sh` only signs in.

## Frontend

- The default unauthenticated screen is a single centered **login** column — signup mode, split hero, and the workspace-slug field are gone; real form (Enter submits), password visibility toggle, Caps Lock hint, email preserved on failure.
- Forgot-password + reset screens render exactly when `GET /auth/capabilities` reports the flow operational; the emailed deep link's token is scrubbed from the address bar/history on read.
- `api/client.ts` drops the localStorage slug + header middleware; settings loses the slug card; offers drops the dev-only PDF fetch workaround.

## CI note

`ci.yml` sets `CONTRACT_STABILITY=pre-live` for this deliberate spec re-sync (the gate's advisory stance — it still prints every breaking change). **A follow-up PR reverts it to stable immediately after merge.** Also fixes a real bug in `check-contract-breaking.sh`: the pre-live branch passed `--fail-on NONE`, which oasdiff rejects.

## Verification

- `make check` (backend + frontend) green locally; frontend 498/498.
- `make test-integration`: **OK, 0 skips, 15 packages** — including the new reset suite, bootstrap atomicity/idempotency/singleton-refusal, and the composite-FK + RLS fitness gates over `auth_token`.
- Background security review findings addressed: STARTTLS-required mailer, async reset send (timing enumeration), throttle-before-parse, token scrubbed from browser history.

## Summary by CodeRabbit

* **New Features**
  * Added deployment-configured installation setup, including organization, administrator, and starter data initialization.
  * Added forgot-password and password-reset flows with secure, single-use links and session revocation.
  * Added authentication capability detection so supported sign-in options are shown appropriately.
  * Simplified MCP, API, and web app access by automatically resolving the installation’s organization.

* **Documentation**
  * Updated setup, configuration, and connection guidance for the single-organization model.

* **Bug Fixes**
  * Improved handling of unavailable or inconsistently configured installations with clearer service errors.


LarsGradion:
Addressed the CodeRabbit review in 9efb39d — disposition per finding:

**Fixed**
- `cmd/migrate`: hidden terminal input via x/term when stdin is a TTY (pipes/tests keep the plain reader); prompt + confirmation writes error-checked.
- `compose/installation.go`: the booking-page seed filters by the explicit workspace id (`storekit.MustWorkspace`).
- `identity/installation.go`: bootstrap now writes `system_log` as the **system** actor (`installation_bootstrap`) — no human logged in.
- `identity/reset.go`: the **whole** account-dependent path (lookup, token mint, mail) now runs off the request — the 202 and its timing are identical for known and unknown addresses; a zero-row user update refuses the reset instead of consuming the token around an unchanged password.
- `deployconfig`: SMTP port validated 1–65535; **`auth.password.enabled=false` fails closed at boot** — it would disable the only implemented login method, so the switch becomes meaningful when OIDC ships (A107 §14 "authentication configuration changes fail closed"). This also resolves the handlers.go finding: the capability can never honestly be false today.
- `mailer`: one connection deadline bounds the entire SMTP exchange.
- reset expiry test fixture: workspace-bound owner transaction + RowsAffected==1.
- `auth.tsx`: the reset-failure note carries `role="alert"`.

**Skipped, with reasons**
- *storekit Audit/Emit for reset mutations*: auth events are the ADR-0060 `system_log` class (non-entity operational events) — the same posture as login, failed-login, and lockout writes in this module. `app_user.password_hash`/`session` mutations on the auth path have never ridden `audit_log`+outbox; routing only the reset through storekit would split one invariant into two spellings. If the spec later reclassifies credential changes as record mutations, that's a module-wide sweep, tracked upstream.
- *frontend "honor password:false / render OIDC"*: the server can no longer produce `password:false` (fail-closed boot), and `oidc_providers` is structurally empty until the OIDC flow ships — branching the UI on states that cannot occur is speculative code. The capability-driven rendering lands with the OIDC work (ADR-0061 §6).

PR #145, opened 2026-07-20
feat(settings): per-user Voice DNA tab + personal/organization split

LarsGradion:
## What

Delivers the per-user **Voice DNA** management surface and splits Settings into a personal group and an admin-only Organization group.

### Voice DNA tab (`frontend/src/screens/voice-dna.tsx`)
The '…later in Settings' surface the onboarding Voice step promises. Over the owner-private `/voice-profiles` lifecycle (landed in #134):
- View the **derived voice** (`voice_profile_md`), status, and quality band.
- Edit **your preferences** (`personality_md`) — If-Match version-guarded; the model never overwrites it.
- Manage **writing-sample corpus sources** (add via paste, remove).
- **Rebuild** with honest queued/deferred/failed states.

### Personal / Organization split (`settings.tsx`)
- **Your settings** (every member): Account, Voice DNA, AI & autonomy (keeps the caller's own agent passports).
- **Organization** (admin/ops only): Company, Data model, Catalog, Privacy, Audit.
- The AI tab deliberately stays personal so a rep never loses passport minting; its admin-only cards remain gated per-card.
- This also puts the **company website-refresh behind the admin-only Organization group** (addresses the 'reread only for admins' ask at the UI; the server keeps `organization:update` as defense — a hard server-side admin-only pin is a spec follow-up).

### Rename
Finishes the user-facing **'voice profile' → 'Voice DNA'** rename in the crm.yaml operation summaries + i18n (identifiers/paths/tables stay frozen/drift-gated).

## Verification
- `make check-fe` green (biome + tsc + vitest + build; DS purity/spacing gates pass; settings.test.tsx updated for the grouped, role-gated nav).
- `make check-backend`: contract regenerated, drift + contract-breaking (description-only, non-breaking) clean.
- Manual (live stack, curl-verified lifecycle): create → ingest → build → derived voice; PATCH personality_md; delete source.

## Follow-up
- PR-D (next): admin **User management** (invite / change-role / deactivate) fills the Organization group's Users card.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added a **Voice DNA** section to Settings, including status, quality band, version, and derived voice details.
  * Manage personality preferences and “Schreibproben” (writing samples) with add/remove and paste-to-create source support.
  * Build/rebuild Voice DNA with progress updates and terminal build outcomes.
[...]

PR #147, opened 2026-07-20
feat(identity): admin user management — invite, roles, deactivate, reactivate

LarsGradion:
## What

Delivers the admin user-management surface the contract flagged as a fast-follow (crm.yaml: `/users … CRUD fast-follow`), and fills the **Users & roles** card in the admin-only Organization settings group (PR #145).

### Contract + backend (`internal/modules/identity`)
- **POST /users** — invite a member: create with no password + the target system role + a single-use set-password token (mailed when a sender is wired), one transaction, duplicate email → 409, emits `user.invited`.
- **PATCH /users/{id}/role** — set the member's system role (wires the existing `ChangeUserRole`).
- **POST /users/{id}/deactivate** — deactivate + revoke sessions/passports (wires the existing `DeactivateUser`).
- **POST /users/{id}/reactivate** — new `ReactivateUser`; restores `active`, emits `user.reactivated`.
- **GET /users?include_inactive** — admin-only widening so deactivated members are visible to reactivate; the default active-only roster still serves the share/assignee pickers.
- All writes admin-only (`hasRole(admin)`, re-checked server-side); each returns the resulting member. Registers `user.invited`/`user.reactivated` on the identity stream.

### Frontend (`screens/users-admin.tsx`)
The **Users & roles** card: include-inactive roster with status badges, an invite form (email + name + role), a per-row set-role control, and deactivate/reactivate — in the admin-only Organization group.

## Invite credential flow
An invite creates a passwordless member and mints a single-use set-password token via the existing A74 account-recovery path; the new member sets a password through the standard reset flow (mailed when a sender is configured). No password is ever chosen by the admin.

## Verification
- `make check-backend` green (build, vet, lint incl. goconst, arch-lint, auth-gate + license + drift fitness tests, contract-breaking non-breaking).
- `make test-it DIR=backend/internal/modules/identity` — invite/reactivate/deactivate/role over real Postgres (create + role + token + events; duplicate → 409, unknown role → 404, non-admin → 403; reactivate idempotent + admin-only).
- `make check-fe` green (biome + tsc + vitest incl. the Users card suite + build); `make frontend-e2e` 37/37 (axe WCAG AA).

## Follow-up
- The `User` contract carries no per-user role, so the set-role control sets a role without showing the current one — surfacing effective roles per member is a small contract follow-up.

## Summary by CodeRabbit

* **New Features**
  * Added a **Users & Roles** settings area for managing organization members.
  * Admins can **invite**, **assign roles**, **deactivate**, and **reactivate** members; roster status updates immediately.
  * Added an admin-only option to **include inactive** members in the user list.
  * Invite emails now include **set-password links valid for 7 days**.
* **Localization**
  * Added **English and German** translations for the new users-management UI.
* **Bug Fixes**
  * Improved conflict handling for duplicate invites and role/admin changes, including correct audit/event behavior.
  * Added protections to prevent deactivating/demoting the **last active admin**.
* **Tests**
  * Added end-to-end and integration coverage for the new admin user management flows.


LarsGradion:
Addressed in 7c7cc87:
- **Last-admin lockout (deactivate + role):** DeactivateUser and ChangeUserRole now refuse with 409 when the target is the only active admin (documented on both ops; the HTTP test proves the sole bootstrap admin can't deactivate or demote themselves).
- **inviteUser 404:** documented (unknown role → 404, same mechanism as changeUserRole).
- **Invite trim:** the form now trims email/name before sending.

LarsGradion:
## Review disposition (commits 7c7cc87, ab1f826, 2b8f2df)

**Fixed**
- **Reactivate only reverses `deactivated`** (409 for suspended/other) — no longer silently clears a lockout; the card offers Reactivate only for deactivated members.
- **Users card is admin-only** — an ops seat in the Org group no longer sees controls that only 403.
- **Invite validation** — email format + display-name length rejected with 422; deactivate reason length validated.
- **Accessibility** — invite is a real `<form>` (keyboard submit), fields have accessible labels, dynamic errors are `role=alert`.
- **Deactivation is confirm-first** (it revokes sessions + passports).
- **Role control** controlled-at-empty so re-selecting after a failure re-fires; **row wraps/shrinks** instead of overflowing at 390px.
- **Last-admin lockout guard + its TOCTOU race** (advisory lock, proven by a concurrency test) — earlier in the review.
- **inviteUser 404**, **invite trim**, **duplicate-email 409** documented/handled.

**Deferred (follow-ups)**
- **Roster pagination** beyond the first page — a single-org install's member list is small; a Load-more is a follow-up.
- **Set-password link when no mailer** — invited members are recoverable today via the operator `migrate reset-password` path; returning the link in the response is a contract-shape decision for a follow-up.
- **Advisory-lock 32-bit key** (P3) — worst case is occasional cross-workspace serialization of admin ops, never incorrectness.
- The `users-admin.test.tsx` "duplicate `calls`" flag is a false positive — each declaration is in its own `it()` scope; tsc + vitest are green.

Coverage gate now passes (new-code ≥ 80%).

LarsGradion:
All six follow-up review points fixed in 134a356: rune-safe display-name/reason limits, the invite mail now goes to the parsed/normalized address, a failed deactivation surfaces inside the confirm dialog, long display names wrap, and the German confirm copy is singular/gender-neutral.

LarsGradion:
Both test-quality points fixed in d552e06: the card now gates its body on the /me role probe (so the admin-only notice can't flash while the probe loads, and the test can't pass on the loading render), and the non-admin mock throws on any unexpected request instead of serving roster data.

PR #153, opened 2026-07-21
fix(ai,compose): capture embed payloads; recover panics in site-extraction fan-out

tinnhlt-gra:
## Summary
- **`Router.Embed` never captured a payload.** Every other AI lane (completions, structured calls) captures request/response into `ai_call_payload` when `ai.capture_payloads` is on; the embed lane built its trace with no `buildPayload`-equivalent call at all, so its trace UI always read "no payload captured" regardless of the config setting. Added `buildEmbedPayload`, which captures the embedded inputs and a `{vector_count, dims}` summary for the response (raw vectors are opaque floats, not reviewable content).
- **Unrecovered panics in the site-extraction fan-out could crash the whole worker process.** `crawlAndExtract`/`extractSite` run the profile lane and up to 40 concurrent per-page fact lanes, each in its own goroutine, none panic-recovered — a panic in any one of them kills the entire process instantly, including any other lane mid-flight on its own model call and trace write. Added `safeExtractPageFacts`/`safeExtractProfile`, thin wrappers that recover a panic into an ordinary error so one lane's panic degrades like any other lane failure (the file's own "collect-don't-cancel" contract), instead of taking the whole fan-out down.

Both fixes came out of investigating a user-reported gap where `site_extract` AI calls appeared in the usage-summary table but not in the call-trace table. That specific historical gap did not reproduce live during investigation (a fresh deep-read against the same organizations traced correctly), so it's not fixed by this PR — but the panic-recovery gap is a real, independently valid hardening fix found along the way, and the embed-payload gap is an unrelated, fully-diagnosed bug in the same UI area (screenshotted by the reporter).

## Test plan
- [x] `go build ./...`, `go vet ./...` clean
- [x] `go test ./internal/modules/ai/... ./internal/compose/...` green, including two new tests proving the panic wrappers degrade instead of crashing the test process, and one new test proving the embed payload captures inputs (with a secret stripped) and a vector-count summary
- [x] `craft static` (diff-scoped): 0 blocker, 2 pre-existing MAJOR (unrelated line-count debt in functions this diff doesn't restructure)
- [x] Two independent rounds of `craft-reviewer` + `security-redteam` subagent review, all findings applied (fixed a double-prefixed panic error message, clarified a load-bearing comment on the embed-payload stripping-order invariant, removed fix-narration from test comments)

---
## Summary by cubic
Capture payloads for embed calls and harden site extraction by recovering panics so a single page or profile failure doesn’t crash the process. Add a count cap to embed capture and log stack traces for recovered panics.

- **Bug Fixes**
  - Embedding: capture request inputs and a response summary (vector_count, dims) when payload capture is on; secrets are stripped and raw vectors are not stored; add a 200-input cap with `inputs_truncated`, and speed up truncation by scanning runes without large allocations.
  - Site extraction: recover panics in page-fact and profile goroutines, converting them to errors so other work continues; also record the stack in internal logs for diagnosis.

<sup>Written for commit d74e8b3b7e5427328483e8e8bd3dfdaf4a8728d5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/153?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved extraction reliability by preventing isolated processing failures from crashing the application.
  - Other page and profile results now continue processing when one extraction task fails.

- **New Features**
  - Added payload capture for embedding requests when call tracing is enabled.
  - Captured embedding details include sanitized inputs and summary information about generated vectors.

### DIFF
--- a/backend/internal/compose/siteextract.go
+++ b/backend/internal/compose/siteextract.go
@@ -13,6 +13,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log/slog"
+	"runtime/debug"
 	"slices"
 	"sort"
 	"sync"
@@ -154,6 +156,12 @@ func crawlAndExtract(ctx context.Context, crawler *siteCrawler, x evidenceExtrac
 func safeExtractPageFacts(ctx context.Context, x evidenceExtractor, page crawlPage) (res pageFactsResult, err error) {
 	defer func() {
 		if p := recover(); p != nil {
+			// recover() suppresses the runtime's own stack dump, so this is
+			// the only place that trace still exists — capture it into the
+			// internal log now, before it's gone; the returned error stays
+			// a short, stack-free line since it can surface on a dossier's
+			// warnings, not just internal diagnostics.
+			slog.ErrorContext(ctx, "extraction panic recovered", "lane", "page_facts", "url", page.URL, "panic", p, "stack", string(debug.Stack()))
 			err = fmt.Errorf("panic: %v", p)
 		}
 	}()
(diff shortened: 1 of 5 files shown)


tinnhlt-gra:
@coderabbitai review

PR #151, opened 2026-07-21
chore(dev): make dev boots cold and sweeps every stray stack first

LarsGradion:
## Why

Two things kept biting during local testing.

**Old stacks lingered.** Repeated `make dev` runs (and `DEV_SLUG` envs) left api/worker/vite processes running on slug-derived ports, each against its own `margince_dev_*` database. Nothing failed loudly — the browser kept talking to whichever api answered `:8080`, sometimes a binary from an earlier branch pointed at a different database. That breakage is indistinguishable from a real bug.

**`make dev` always seeded.** The demo workspace was created on every boot, so onboarding, empty states, and first-run flows were permanently skipped — the state a developer worked against was never the state a first customer sees.

## What changed

**`make dev` boots cold.** No demo records on boot: just the organization and admin seat the api bootstraps from `config/margince.yaml`. Demo data stays the explicit opt-in it already was — `make seed-dev` against a running stack — and the ready banner points at it. The banner's login line now reads the admin address back from the deployment config rather than the manifest in `seed-dev.sql`, so an edited config prints the truth.

**`make dev` sweeps before it starts.** A bare invocation:

1. kills every stack recorded in `.tmp/dev/*/env` (pids + ports);
2. kills orphans whose state file is gone, and margince servers from *other* checkouts (matched on binary name **and** a margince DSN, so an unrelated program called `api` is never touched) plus vite processes resolved out of this worktree;
3. evicts anything holding `:8080`/`:5173`, margince or not;
4. drops leftover `margince_dev_*` databases.

TERM, then KILL after 5s. The shared `margince` and the test lane's `margince_test*` / `margince_it_*` namespaces never match the pattern, so `make test-integration` is unaffected. One stack runs, always on `:8080` against `margince`.

`make dev-stop` is the mirror: bare, it stops **every** stack rather than only the one it recorded. `DROP=1` drops the per-slug databases; `margince` always survives.

`DEV_SLUG` still starts an isolated env and the sweep spares a running one — but the next bare `make dev` takes it down. Exclusivity is the point; an isolated stack is the deliberate exception.

## Verification

Manual, on a deliberately messy machine:

- base stack on `:8080` **plus** a `DEV_SLUG=ghost` stack on `:8932` → bare `make dev` swept 11 processes, dropped the stray databases, came up on `:8080`;
- three stray `margince_dev_*` databases at once → all three dropped, `margince` / `margince_test*` / `margince_it_*` intact;
[...]

PR #156, opened 2026-07-21
feat(onboarding): keep the legal identity a site states, and let the human pick

LarsGradion:
Stacked on #151 (base `chore/dev-cold-start`); rebase to `main` once that merges.

## The bug

The most important thing a CRM records about a company — registered name, address, VAT/register number — was the one thing onboarding never filled in. Not because the read missed it: **because the read threw it away.**

A group's legal notice states one block per entity. `gradion.com` names five. The profile prompt said to extract the legal trio *"ONLY when the site's legal pages name exactly one entity"*, and `applyLegalGate` dropped anything that slipped past. The abstention is correct — picking one of five at random writes another company's legal identity into this one's CRM — but discarding the blocks leaves a human retyping five lines the page already printed, which is the exact work the read exists to remove.

## What changed

The census now keeps what it reads. Each entity carries the address and register number printed **with its name**, gated like every other value: present on the page or absent, never guessed. Blocks sharing a register number are one company however the page labels them, so locale variants and market headings ("Gradion Singapur" above "Gradion Pte. Ltd.") fold into one entry. The dossier stores them (migration `0111`), the contract exposes them, and the confirm step offers them — one click fills the three fields, carrying the imprint's own quote as their evidence rather than marking them as the human's typing.

Two extraction fixes were needed to get there, both found by running the real read rather than by reasoning:

- **The detail keys had to become required in the reply schema.** As optional fields the model omitted them every single time, which is why the address never arrived and only the register number did.
- **An address survives a model round trip with its punctuation rearranged** — a page printing `Singapore (179433)` comes back as `Singapore 179433`. A detail now counts as printed when every content token is on the page, not only on exact containment. Rearranged punctuation passes; an invented street cannot, because its tokens are nowhere on the page.

## The step itself

Rebuilt around the work instead of the pitch:

- Titles were advertising — *"Let Margince understand your business."*, *"Make this unmistakably yours."* — and are now what the screen is for: **"Your company"**, **"Check your company details"**.
- The "Nothing was invented" explainer block is gone; the subheading carries it in one line.
- The ~100 evidence cards that sat **between the heading and the form** now sit after it, collapsed behind a count. A review step that opens with a hundred cards to scroll past collects a rubber stamp, not a confirmation.

## Verification

End-to-end on a fresh cold database against the live Anthropic model, `gradion.com`:

- all five entities reach the wire with complete blocks, matching the imprint exactly;
- picking *NFQ Solutions GmbH* fills legal name / address / VAT, labelled **"read from site"** with the imprint quote and URL as evidence;
- `GET /v1/company` after confirm returns `legal_name: "NFQ Solutions GmbH"`, `registered_address: "Deliusstraße 7, 24114 Kiel, Germany"`, `register_vat: "DE346175276"`.

## Note

`people/siteread.go` crossed the 500-line cap, so the claim-and-close half moved to `sitereadfinish.go` along the seam that was already there.

The deeper "rethink the flow from scratch" is **not** in this PR — this restructures and cuts, it does not redesign the five-step wizard.

---
## Summary by cubic
Onboarding now preserves all legal entities from a site’s legal notice, exposes them via the API, and lets the user pick one to auto-fill legal name, address, and VAT/register number. The entity name now uses the same whole‑token gate as details (names checked against the whole page; details against the cited block) to block partial, recombined, or cross‑block identifiers.

- **New Features**
  - Backend: keep entity blocks (name, address, register), dedupe by register number (fallback: normalized name), store as `site_read.legal_entities`, and expose as `CompanySiteRead.legal_entities` with `CompanySiteReadLegalEntity`; optional details are absent, not empty; debug report includes entity details; finish logic split to `sitereadfinish.go`.
  - Backend: whole‑token gate applied to both entity names (scoped to the whole page) and details (scoped to the cited block); rejects truncated, recombined, or cross‑block identifiers.
  - Frontend: legal‑entity picker fills the three fields and is labeled as your input; starting a new read resets the applied version, clears the picked entity and fact selection, and persists this state; facts list moved below the form and collapsible with a selected/total count; clearer copy.
  - DX: `make dev-fresh` for a rebuilt first‑run stack; app served on :8080 with the API proxied; safer `--fresh` refusal; ignore `.playwright-mcp/` test debris.

- **Migration**
  - Adds `legal_entities` jsonb column to `site_read` (migration 0112). Run migrations; no manual backfill.

<sup>Written for commit 385d7aecdcff74dd6fa23e62caa35219ccb4032c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/156?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Onboarding now identifies and lets you choose among multiple legal entities found on a company website.
  * Legal entity details such as name, registered address, registration or tax number, and source evidence are captured and shown in site-read results.
  * Fact selection is collapsed by default and displays the number of selected facts.
  * Added `make dev-fresh` to recreate a clean local database for first-run testing.

* **Documentation**
  * Updated onboarding copy and development setup guidance.


LarsGradion:
Reviewed every open thread. Fixed in this branch:

- **Invented identifiers via token recombination** — `groundedDetail` matched an unordered token set, so a notice printing "24114 Kiel" and "HRB 123456" vouched for the fabricated "HRB 24114". Tokens must now appear as a contiguous run, which still tolerates rearranged punctuation.
- **Truncated identifiers** — substring containment accepted `1234` against `HRB 123456`; whole-token matching only.
- **Cross-block attribution** — details are judged against the *cited block*, not the whole page. Honest limit documented in the code: the passage packer merges short lines, so a compact notice IS one block and this scoping cannot separate siblings inside it.
- **Provenance** — the picker no longer labels the legal trio "read from site", because confirmation records it as human input. Label and record agree; server-side binding is the follow-up.
- **Stale entity across reads** — starting a read now resets the version counter, clears the fact selection and the legal trio, and persists all of it with the new dossier id, so a reload restores the site you are actually reading.
- **DSN password in the `--fresh` refusal** — it names no connection at all now.
- **Committed Playwright captures** — removed from the index and `.gitignore`d; the squash merge keeps them out of main's history entirely.

Deferred deliberately, recorded in #162 so they outlive this branch: `FinishSiteRead`'s CAS versus a stale worker (pre-existing — this PR only moved the function), a VAT group folding two companies into one census entry, and a legal-only partial read being recorded as failed.

Also renumbered the migration to 0112 after main landed its own 0111; verified by migrating a clean database to head.

Resolving the threads on that basis.

PR #159, opened 2026-07-21
feat(overlay): continuous-sync deletion feed — purge incumbent-deleted records (branch 1b)

dungbanguyen:
## What & why
Branch 1 (#91) converged the overlay mirror on the incumbent's **modified** records and purged everything on disconnect, but a record deleted/archived in HubSpot during normal operation stayed readable in the mirror **forever** (PR #91 review defect: no deletion signal on the seam). This adds the opposite-direction convergence so an incumbent-side deletion stops being readable rather than lingering until disconnect.

Contract-first: the upstream spec pin merged first — margince-foundation #1123 (`mirror.deleted` in event-bus.md + EVT-SEM-13; `OVA-AC-5` + `OVA-EVT-4` continuous-sync deletion guarantee).

## How
- **Inner seam** grows `Incumbent.Deletions(objectClass, since, cursor)` + `Deletion`/`DeletionPage` (additive; the frozen `datasource` outer seam is untouched).
- **HubSpot adapter** implements it over the archived-object list feed (`?archived=true`, `archivedAt`→`DeletedAt`), keyed by canonical class. Documented: the list is not `archivedAt`-ordered, so the sweep pages it fully each pass — correct because purge is idempotent (a cheaper incremental feed is a later optimization).
- **`MirrorStore.PurgeRecord`** removes the mirror row + association edges (**both** endpoints) + visibility projection in one tx, reporting whether the row existed (so `mirror.deleted` fires only for records actually mirrored). **No tombstone** — a routine incumbent deletion is not a GDPR erasure hold; a restored record must be free to re-mirror.
- **`ReconcileDeletions`** drives the sweep on its own `deleted_at` watermark (new `overlay_deletion_watermark` table), emitting `mirror.deleted` per purged record. The poller runs it **after** the modified sweep each tick (HubSpot excludes archived from Search, so the two never fight).

## Tests / gates
- TDD across 5 cycles (fake feed → `PurgeRecord` → watermark → HubSpot adapter → reconcile sweep → poller wiring).
- End-to-end: the poller purges an incumbent-deleted record so a previously-mapped reader gets existence-hiding not-found.
- `make check-backend` green; overlay integration **89 pass**, compose **28**, dispatch **5**; `craft static` clean; table-ownership + event-catalog fitness tests updated.

---
## Summary by cubic
Adds continuous deletion sync so HubSpot-archived/deleted records are purged from the mirror and stop being readable. Runs after the modified sweep and emits `mirror.deleted` only when a mirrored record was actually removed.

- **New Features**
  - Added `Incumbent.Deletions(objectClass, since, cursor)` and types; HubSpot adapter uses `?archived=true`, maps `archivedAt` → `DeletedAt`, and keys by canonical class.
  - Added `MirrorStore.PurgeRecord` to remove the mirror row, both association endpoints, and visibility in one transaction, and emit `mirror.deleted` atomically; idempotent; no tombstone.
  - Added `ReconcileDeletions` to full-scan the archived feed each sweep (idempotent), run after the modified sweep, and meter per deletion; wired `mirrorDeletedTotal` and exposed it via `/metrics`.
  - Registered `mirror.deleted` in the events catalog; fake adapter supports the deletion feed; end-to-end poller test proves a deleted record becomes unreadable.

- **Migration**
  - New indexes to speed per-record purge and visibility clear: `idx_overlay_association_to`, `idx_mirror_visibility_record`.

<sup>Written for commit 0ee6842a973ddf7aa617b9ed51cce8351c6d3358. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/159?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added support for detecting deleted or archived records during synchronization.
  * Removed deleted records from mirrored data, associations, and visibility projections.
  * Added `mirror.deleted` events for records removed from the mirror.
  * Added progress tracking so deletion processing can resume reliably.
  * Added HubSpot archived-record support, including pagination.

* **Documentation**
  * Documented deletion synchronization, event behavior, and restore handling.

* **Bug Fixes**
  * Disconnect cleanup now also clears deletion synchronization progress.

### DIFF
--- a/STATUS.md
+++ b/STATUS.md
@@ -598,14 +598,20 @@ Open work, roughly in priority order:
 
 - **Overlay branch 1b — the review-deferred hardening** (from PR #91's
   three-lens review; the branch itself ships read + poller sync with the
-  human `/v1` surface seam-backed): budget-pace the initial backfill and
+  human `/v1` surface seam-backed). **In flight:** the **deletion/archive
+  feed** is implemented on branch `fix/overlay-deletion-feed` (local commit
+  `9b8fc80`, gates + integration green, not yet pushed) — `Incumbent.Deletions`
+  + HubSpot `?archived=true` + `MirrorStore.PurgeRecord` (row + assoc +
+  visibility, no tombstone) + `ReconcileDeletions` on its own
+  `overlay_deletion_watermark` + the `mirror.deleted` event, run after the
+  modified sweep. Contract-first blocker: it lands after the upstream spec pin
+  (foundation branch `spec/overlay-deletion-feed`: `mirror.deleted` +
+  EVT-SEM-13 + OVA-AC-5) merges. Still open in 1b: budget-pace the initial backfill and
   honor `Retry-After`/ADR-0063-style backoff on a failing connection
   (today a failed connection re-sweeps hot every tick); one budget meter
   per deployment, not per process/consumer (combined spend can exceed
   the HubSpot quota N×); composite keyset watermark (a >10k
   same-timestamp block wedges the sweep — disclosed in reconcile.go);
-  hard deletes never reach the mirror (no deletion signal, no ID-set
-  diff — an incumbent-deleted record stays visible until disconnect);
   the webhook-as-signal lane returns only WITH portal-id→workspace
   binding in the HMAC basis (the unmounted receiver was deleted, not
   fixed); a reconnect flow (Connect today refuses a workspace with any
--- a/backend/internal/compose/jobs_overlay.go
+++ b/backend/internal/compose/jobs_overlay.go
@@ -171,40 +171,67 @@ func reconcileConnection(ctx context.Context, vault keyvault.Vault, ms *overlay.
 	}
 
 	for _, objectClass := range overlayObjectClasses {
-		// Initial full load before the incremental sweep: Backfill lists
-		// the object class id-cursor style AND fetches its associations
-		// (design.md §4.4), checkpointing overlay_backfill_cursor so
-		// SyncStatus's backfillComplete answers truthfully. It is a cheap
-		// no-op once its cursor has converged, so every later sweep skips
-		// straight to the Modified pass below — the first sweep after a
-		// connect (via the poller, or on-demand through POST
-		// /overlay/reconcile) does the load, the rest ride the watermark.
-		// A backfill failure is logged and skips only this class's sweep
-		// this tick (the NEXT sweep resumes from the checkpoint), never
-		// aborting the other classes.
-		if err := overlay.Backfill(ctx, inc, ms, objectClass); err != nil {
-			log.WarnContext(ctx, "overlay reconcile: backfill pass failed, skipping this object class this tick",
-				"workspace", d.Workspace.String(), "object_class", objectClass, "err", err)
-			continue
-		}
-		since, err := ms.LoadReconcileWatermark(ctx, objectClass)
-		if err != nil {
-			log.WarnContext(ctx, "overlay reconcile: loading the persisted watermark failed, skipping this object class",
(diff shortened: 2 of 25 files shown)


dungbanguyen:
@coderabbitai review

dungbanguyen:
Addressed the review in a rework (force-pushed). Summary of resolutions:

**P1 — purge-then-emit non-atomic (CodeRabbit + cubic):** fixed. `PurgeRecord` now emits `mirror.deleted` **inside the same transaction** as the purge (`mirrordeletion.go`), so the purge and its promised event commit together or not at all — the separate-tx emit that could drop the event is gone.

**P1 — deletion could remain readable forever via the unordered archived feed (cubic):** fixed by **removing the watermark entirely**. HubSpot's `archived=true` list is not `archivedAt`-ordered, so a watermark could permanently skip a record archived behind the cursor mid-sweep. `ReconcileDeletions` now full-scans the archived feed from the epoch each pass and purges idempotently; the adapter pages the whole archived set regardless of `since`, so full-scan costs no more. This also removes the `overlay_deletion_watermark` table/migration/Save-Load and resolves the dependent threads (watermark regress, duplicate-of-reconcile-watermark, migration comment, teardown-comment).

**P1 — concurrent modified+deletion re-insert (cubic):** the poller is single leader-elected and runs the modified sweep then the deletion sweep **sequentially** per object class each tick, and HubSpot excludes archived records from the Modified/Search feed — so a stale Ingest from this poller cannot resurrect a just-purged row. The tombstone/fencing this would need for a concurrent *push* lane is tracked with the webhook-as-signal branch (which does not exist yet); no code change here, by design.

**P2/P3 taken:** `MirrorDeletedTotal` wired into `/metrics`; metrics doc corrected (three counters); deletion-event assertion tightened to match `object_class` too; `Deletion.ObjectClass` documented as canonical-not-incumbent (a future adapter returning the raw class would purge nothing); fake `SeedDeletion` call-order documented (archived leaves the live feed); fake deletion test rewritten to actually prove ascending ordering.

**P2 — per-record purge scans association/visibility without a supporting index:** noted for the branch-1b indexing pass (the association/visibility tables carry PKs but not an index matching the OR-endpoint delete); deferred as a performance item, not a correctness one, tracked in the overlay hardening backlog.

Re-verified: `make check-backend` + overlay/compose integration green; `craft static` clean.

dungbanguyen:
Second follow-up (force-pushed):

- **records.go — "a deletion behind the cursor can remain readable forever":** resolved by the watermark removal. `ReconcileDeletions` now full-scans the archived feed from the epoch every sweep, so a record missed on one pass (appearing behind that pass's cursor) is caught on the next full scan — bounded to one poll interval, never forever.
- **mirrordeletion.go — per-record purge scans the workspace's association/visibility rows:** fixed. Added migration `20260721130000_overlay_purge_indexes` — `idx_overlay_association_to (workspace_id, to_type, to_id)` (the to-side delete predicate the PK prefix didn't cover) and `idx_mirror_visibility_record (workspace_id, object_class, external_id)` (the visibility clear/recompute key). The from-side association delete already rides the PK prefix.

The remaining threads are obsolete (the `overlay_deletion_watermark` table/migration/Save-Load are gone) or already covered in the previous summary; resolving them.

PR #160, opened 2026-07-21
fix(overlay): close visibility remap races + revoke on late email ambiguity (branch 1b)

dungbanguyen:
Three concurrency/ambiguity defects in the overlay visibility projection, from the PR #91 three-lens review (P0/P1). Downstream-only (no spec change).

## Fixes
1. **Serialize remaps per `(workspace, app_user, incumbent)`** (P0, review #23) — `UpsertUserMap` read the prior mapping then upserted + recomputed without serialization, so two concurrent remaps of the same user could interleave their clear-then-grant recomputes and leave the user granted on **both** the old and new owners' records. A transaction-scoped `pg_advisory_xact_lock` keyed on the tuple (the same primitive `identity`'s admin guard uses) makes the read-decide-write-recompute sequence atomic.
2. **Row-lock the recompute** (P0, review #24) — `recomputeForOwnerTx` scanned an owner's mirror rows unlocked, so a concurrent ownership reassignment could finish mid-scan and re-apply the stale owner's grant to a record whose owner had changed. `SELECT … FOR UPDATE` serializes it; Postgres re-evaluates the `WHERE` against the latest committed row version after the lock, so a row reassigned away drops out rather than being mis-projected.
3. **Revoke on late ambiguity** (P1, review #28/#43) — `SeedUserMap` enforced "ambiguous email → no row" only at first seed. Once a second incumbent owner acquired an already-seeded email, the sweep skipped re-seeding but left the stale mapping granting access. It now revokes the email-sourced mappings (and their visibility grants) of every owner sharing a now-ambiguous email.

## Tests
- `TestConcurrentRemapsLeaveExactlyOneOwnerVisible` — two goroutines remap the same user; asserts exactly one owner's records stay visible (invariant, not timing — caught the race live before the fix). Same shape as identity's `TestConcurrentLastAdminDeactivationsKeepOneAdmin`.
- `TestSeedUserMapRevokesAMappingThatBecameAmbiguous` — a mapping is revoked once its email turns ambiguous.
- `make check-backend` green; overlay integration **87 pass**; `craft static` clean.

Part of the overlay branch-1b hardening (follows the deletion feed, #159).

---
## Summary by cubic
Serializes all overlay visibility changes with a per-workspace advisory lock and revokes email-based access when an email becomes ambiguous (counting distinct owners). This removes races that caused stale or double grants during remaps, owner reassignments, and revalidation.

- **Bug Fixes**
  - Use a single per-workspace, transaction-scoped advisory lock across all visibility mutators (UpsertUserMap after email resolution, Ingest owner projection, RecomputeForOwner, RevalidateEmailMappings, and the ambiguity revoke) to make read/decide/clear/grant atomic; recomputes now run under this lock instead of per-owner row locks to avoid deadlocks or missed transitions.
  - When a previously unique email becomes shared, delete email-sourced mappings for all affected owners and recompute in the same transaction under the lock so access is dropped immediately; manual overrides are preserved.
  - Treat duplicate owner directory entries as one distinct owner when checking email ambiguity to prevent accidental revokes or skipped seeds.

<sup>Written for commit 1182f5394e9dc8abd18c7c4e24da6e1d8785c4ca. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/160?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency when user ownership and visibility mappings are updated concurrently.
  * Prevented records from remaining accessible when an email match becomes ambiguous.
  * Ensured visibility updates remain synchronized during mapping, remapping, ingestion, and email revalidation.
* **Tests**
  * Added coverage for ambiguous email ownership and concurrent user remapping scenarios.

### DIFF
--- a/backend/internal/modules/overlay/mirrorstore.go
+++ b/backend/internal/modules/overlay/mirrorstore.go
@@ -136,6 +136,16 @@ func (s *MirrorStore) Ingest(ctx context.Context, rec Record) error {
 	}
 	var landed bool
 	err := database.WithWorkspaceTx(ctx, s.pool, func(tx pgx.Tx) error {
+		// Ingest's owner projection (ProjectOwnerVisibility, and the
+		// owner-change revalidation) mutates mirror_visibility, so it takes
+		// the same per-workspace visibility lock every other mutator takes —
+		// serializing an owner reassignment against a concurrent manual remap
+		// so a record transitioning between owners can never leave a stale
+		// grant. Overlay ingest is driven by the single leader-elected
+		// poller, so this lock is uncontended on the hot path.
+		if err := lockWorkspaceVisibility(ctx, tx); err != nil {
+			return err
+		}
 		var priorOwner string
 		if err := tx.QueryRow(
 			ctx,
--- a/backend/internal/modules/overlay/usermapseed.go
+++ b/backend/internal/modules/overlay/usermapseed.go
@@ -43,24 +43,55 @@ func (s *MirrorStore) SeedUserMap(ctx context.Context, incumbent string, owners 
 	// seeding a user to "whichever owner the directory listed last" would
 	// be a nondeterministic remap that revokes the prior owner's records
 	// every sweep. Only an email owned by exactly one owner is seedable.
+	// Track the DISTINCT owner ids per normalized email as a set, not a raw
+	// occurrence count: a paginated owners directory can list the SAME owner
+	// twice (overlapping pages), and counting that as two owners would
+	// misclassify one legitimate owner as ambiguous — revoking and
+	// withholding a user's visibility over duplicate input rather than a
+	// genuine two-owner collision. Ambiguity is "more than one DISTINCT
+	// owner claims this email."
 	byEmail := make(map[string]OwnerRef)
-	ambiguous := make(map[string]bool)
+	ownersByEmail := make(map[string]map[string]struct{})
 	for _, owner := range owners {
 		email := normalizeEmail(owner.Email)
 		if owner.ExternalID == "" || email == "" {
 			continue
 		}
-		if _, seen := byEmail[email]; seen {
-			ambiguous[email] = true
-			continue
+		if ownersByEmail[email] == nil {
+			ownersByEmail[email] = make(map[string]struct{})
 		}
-		byEmail[email] = owner
+		ownersByEmail[email][owner.ExternalID] = struct{}{}
+		if _, seen := byEmail[email]; !seen {
+			byEmail[email] = owner
+		}
 	}
 
 	var errs []error
+
+	// Revoke any pre-existing email-sourced mapping whose owner email has
+	// BECOME ambiguous since it was seeded (a second DISTINCT incumbent
+	// owner now carries the same email). design.md §4.6's "ambiguous → no
+	// row" must hold going FORWARD, not only at first seed: skipping the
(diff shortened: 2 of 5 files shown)


dungbanguyen:
Reworked (force-pushed): replaced the fine-grained per-owner/per-user locking with **one per-workspace visibility advisory lock** (`lockWorkspaceVisibility`) that every visibility-mutating transaction acquires first. This resolves the concurrency findings by construction:

- **P0 — record transitioning INTO an owner during a concurrent remap-away:** fixed. Ingest's owner projection and `UpsertUserMap` now take the same lock, so a reassignment and a remap serialize — no interleaving that leaves a stale grant. (The `FOR UPDATE` that only locked already-owned rows is gone.)
- **P1 — revoke racing a re-seed:** fixed. `revokeEmailMappingsForOwners` takes the same lock as `UpsertUserMap`, so a concurrent seed can't restore a mapping right after the ambiguity revoke drops it.
- **P2 — crossed remaps deadlock (opposite lock order) / ambiguity-sweep deadlock (nondeterministic map order):** fixed. A single lock key has no ordering, so no deadlock.
- **P2 — slow incumbent lookup held under the lock:** fixed. The email match (which may call HubSpot) now runs BEFORE the lock is acquired.
- **P2 — the concurrency test may pass without exercising the race:** it's an invariant test (two goroutines remap the same user; exactly one owner's records stay visible) — the same shape identity's `TestConcurrentLastAdminDeactivationsKeepOneAdmin` uses, and it did catch the race live before the fix. A stronger guarantee needs a fault-injection seam the store doesn't have; the invariant assertion never flakes and fails if the serialization regresses.
- **P1 — deactivated/recreated owner with the same email:** the ambiguity check is over HubSpot's **active** owners directory (`Owners`). A deactivated owner sharing an email is not a seedable candidate and owns no active assignment, so mapping the user to the single active owner is the correct call — detecting archived-owner email collisions would need an archived-owners fetch (a separate adapter/seam addition), out of scope here. Documented as the deliberate posture.

Serializing overlay visibility mutations per workspace is free in practice: they come from one leader-elected poller plus occasional manual remaps.

dungbanguyen:
Both addressed (force-pushed):

- **usermapseed.go — duplicate owner listing miscounted as ambiguous:** fixed. `ownersByEmail` now tracks the DISTINCT owner-id set per normalized email, so an overlapping/duplicate directory entry for the same owner is one owner, not an ambiguous pair. Added `TestSeedUserMapTreatsADuplicateOwnerListingAsOneOwner`.
- **visibility.go — `lockWorkspaceVisibility` no-op on unset workspace GUC:** fixed. Dropped `missing_ok` from `current_setting('app.workspace_id')`, so an unset GUC RAISES instead of resolving to NULL (which the STRICT `hashtext`/`pg_advisory_xact_lock` would have turned into a lock-free no-op, silently bypassing the serialization). This fails closed on an unset GUC, matching how the RLS policies behave.

PR #185, opened 2026-07-22
feat(extensions): walking skeleton of the stable extension tier (ADR-0069 slice 1)

lstrojny:
An extension directory under extensions/ now composes into the role binaries without a hand-edit to any upstream-owned file, and the boot observes it:

- backend/pkg/extension is the published registration surface (the first pkg/ inhabitant): Extension + Registry with exactly ONE capability, RegisterJurisdiction — the jurisdiction pack contract moved to pkg/extension/jurisdiction, ports/jurisdiction aliases it so core call sites and the coming de pilot register the same type. pkg is the outermost Tier-0 leaf, held stdlib-only three ways (depguard pkg-purity, TestPublishedSurfaceIsPure, go-arch-lint component).
- tools/gen-composition materializes build/composition/ (gitignored): the composed go.work(+sum), the composition module wiring the enabled set, the degenerate frontend/contract outputs, and composition.json binding input digests to reproducible output hashes. Deterministic: sorted sets, content digests, no timestamps; -verify is the regeneration-reproduces gate (make check-composition); every unbuilt capability layer (api/, frontend/, migrations/) is a loud refusal.
- The four role binaries wire the constant import path github.com/gradionhq/margince/composition; bare go builds resolve the committed vanilla stub via backend/go.mod's replace, make lanes run GOWORK=build/composition/go.work where the workspace `use` wins. The stub is byte-identical to the generator's vanilla output (held by the generator, -verify, and the tool's unit test), so both lanes provably wire the same thing. migrate gets the set when the extension migration namespace lands.
- compose.RegisterExtensions is collect-then-apply: Register turns capabilities into data, nothing lands in a core registry unless the whole composed set collected cleanly, and a stashed Registry used after boot hits an accident-gate panic (EXT-P13).
- compose.ObserveExtensionInventory writes one system_log row (extension.composition_observed) per composed-set change at api/worker boot — install, upgrade and removal happen in source, so boot is where they become attributable (ADR-0069 §5).
- Fitness: extensions (and fixtures) import only backend/pkg/**, never internal/cmd/composition/sibling extensions; the backend imports the composition module only from cmd/*/ and never an extension module.
- CI: extension-reference job copies fixtures/extensions/crm-hello into extensions/ and proves gen → composed build → unit lane → the extension's own test lane. make test now also runs ./tools/... (the tools module was outside ./... — gen-aitasks/gen-workflow tests join the gate).

Spec findings to reconcile upstream on margince-foundation#1128: go.mod is required for a Go-bearing extension (a go.mod-less directory cannot join a Go workspace — ADR-0069 §2 says optional); the committed vanilla stub module is how "build entrypoints consume composed paths on every build" coexists with bare-go/gopls usability; a bare go build in a tree WITH extensions still produces a vanilla binary (only make lanes compose — observable via the §5 boot log).

## How verified

- [x] `make check` is green
- [ ] `make craft-static` reports no new blockers

## Accountability

off (`git commit -s`, DCO). See [CONTRIBUTING.md](../CONTRIBUTING.md).

---
## Summary by cubic
Adds the stable extension tier walking skeleton. Extensions under `extensions/` compose into `api`/`worker`/`mcp` without touching core; boot registers them once and logs inventory changes.

- New Features
  - Published `backend/pkg/extension` with the `jurisdiction` contract on typed `Code`/`Period` plus `extension.Name`/`extension.Version`; re-exported from `internal/shared/ports/jurisdiction`.
  - Deterministic `tools/gen-composition` generates `build/composition/` (composed `go.work`, composition module, frontend/contract stubs, manifest) and is gated by `make check-composition`; rejects unbuilt layers (`api/`, `frontend/`, `migrations/`).
  - Role binaries import `github.com/gradionhq/margince/composition`; boot calls `compose.RegisterExtensions` and `compose.ObserveExtensionInventory` (one `system_log` row per set change).
  - CI composes and tests a reference extension by copying `fixtures/extensions/crm-hello` into `extensions/`; local dev (`scripts/dev.sh`) and make lanes build under the composed workspace.

- Bug Fixes
  - One `composition.Extensions()` snapshot per boot serves registration and inventory.
  - Stronger validation: unit name grammar, non-empty `Version`, and duplicate `jurisdiction.Code` preflighted against declared and live registries; `ports/jurisdiction.Register` enforces lower-case ISO 3166-1 alpha-2.
  - Retention floors now use ISO 8601 calendar intervals via `jurisdiction.Period` in queries (no 365-day shortening across leap years).
  - Inventory observation serialized with a per-workspace advisory lock; last-observation ordering fixed.
  - Generator hardening: rejects symlinked entries and non-regular files; strict `composition.json` schema/bytes; errors on unsupported root `go.work` directives; pins each unit to `extensions/<name>`; go binary for `go.work.sum` resolves from the running toolchain; vanilla output matches the committed `composition/` stub.
  - CI filters watch `extensions/**`, `fixtures/**`, `composition/**`; `.gitignore` anchors `/build/composition/`; make targets depend on `composition`; `cmd/api` config parsing moved to `config.go`.

<sup>Written for commit b4f2af3275037327ff1d22c962e444a56bdd1706. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/185?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added a composable extension framework for registering jurisdiction capabilities.
  * Applications now load enabled extensions during startup and track extension inventory changes.
  * Added a reference CRM extension demonstrating extension registration and jurisdiction support.

* **Bug Fixes**
  * Builds, tests, development runs, and end-to-end checks now consistently include enabled extensions.

* **Documentation**
  * Documented new composition and reproducibility checks.

* **Tests**
  * Added validation for extension compatibility, registration behavior, and reproducible composition.


lstrojny:
Review triage — all 37 inline findings, disposition per item. The branch is being force-pushed with two commits: the reworked slice (85fb25f — the registration idiom changed from `Register(Registry)` to a declaration value returned by `New()`, per review discussion) and the fixes commit (07ad762).

**Fixed in 07ad762**
- ci.yml extension-reference `working-directory` (cubic P0) + `extensions/**`/`fixtures/**`/`composition/**` in the backend change filter
- `e2e-ai` missing `composition` prerequisite (coderabbit + cubic ×2); root Makefile now delegates `composition`/`check-composition`
- one `composition.Extensions()` snapshot per boot for registration + inventory (cubic ×3)
- generated wiring pins each unit to its `extensions/<name>` directory — `mustBe` aborts a mismatched declaration (cubic ×2)
- scanner refuses symlinked units (was silently skipped) and non-regular files (FIFO hang) (cubic ×2, coderabbit ×1)
- `-verify` checks schema, re-encodes composition.json byte-for-byte, and rejects stale/injected extras under build/composition/ (cubic ×2)
- unsupported root go.work directives (replace/toolchain/godebug) refuse loudly instead of vanishing (cubic)
- RegisterExtensions preflights jurisdiction codes against the declared set AND the live registry, so the apply phase can no longer abort halfway (coderabbit + cubic — the collect-then-apply atomicity gap); requires non-empty Version (cubic)
- inventory: per-workspace advisory lock serializes concurrent api/worker observations (cubic + coderabbit); last observation orders by `occurred_at, id` (cubic)
- `ports/jurisdiction.Register` rejects non-ISO-3166-alpha-2 codes (cubic)
- `.gitignore` anchored to `/build/composition/` (cubic); `backend/go.mod` comment names the three wired binaries (cubic); exact name grammar + version contract documented on `pkg/extension` (cubic ×2)

**Moot after the rework (85fb25f)** — the Registry object no longer exists; capabilities are inert declaration values reconciled after whole-set validation:
- collector mutex/closed race, registry-retained-after-error, document-registry-lifetime (cubic ×3)

**Correction on second look** — the `strconv.Unquote` finding (cubic) was NOT moot: it targets the fitness tests themselves, and a raw-string (backtick) import path would have slipped past the quote-trimmed matching. Fixed in 07ad762: import paths decode via `strconv.Unquote`, an undecodable path fails the test.

**Deferred with reason**
- retention-class names free-form (cubic): class vocabulary is deliberately out of v1 — ADR-0069 §13 defers vocabulary registration to its own RFC; the only consumer today is the core-owned de pack
- `Years*365` leap-year shortening (cubic): pre-existing core retention-evaluator behavior, not this diff — tracked as a follow-up fix in the privacy module
- Tier-0 `internal/shared` importing `backend/pkg` (coderabbit): deliberate tier amendment, not an accident — `pkg` sits BELOW shared as the outermost stdlib-only leaf (held by depguard `pkg-purity` + `TestPublishedSurfaceIsPure`); ports alias published seam types so a core module and an extension register the same type (ADR-0069 §3)
- go.work.sum not independently regenerated in `-verify` (cubic): known bound, documented — the sum is a pure function of the members' go.mod/go.sum graph, which the input digests cover; a temp-workspace regeneration is preflight-slice work
- parallel `make -j` composition race (cubic): the gate lanes run serially by design; `-j` on `make check` isn't a supported invocation
- dev.sh worker-build failure orphaning the api (cubic): the `make dev` machine sweep reaps any orphan on the next run — self-healing by design

backend/pkg/extension/jurisdiction/jurisdiction.go
```diff
@@ -0,0 +1,40 @@
...
+// jurisdiction (A57) — there is no LocaleBundle on this seam.
+package jurisdiction
+
+// Pack is one jurisdiction's compiled-in behavior set. Retention is the
+// only obligation packs carry today; the further ADR-0042 contributions
+// (FiscalFormatter, ConformityRegime, …) return when a work package pays
+// for them.
+type Pack interface {
+	// Code is the ISO 3166-1 alpha-2 code, lower-case ("de").
+	Code() string
```
lstrojny: ISO codes should be a type

backend/pkg/extension/jurisdiction/jurisdiction.go
```diff
@@ -0,0 +1,40 @@
...
+type Retention interface {
+	Classes() []RetentionClass
+}
+
+// RetentionClass names one statutory retention floor: the core engine
+// treats Years as a minimum — a workspace policy may keep longer, never
+// destroy earlier.
+type RetentionClass struct {
+	Name  string
+	Years int
```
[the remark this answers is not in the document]
lstrojny: This should be an ISO 8601 date interval

backend/pkg/extension/extension.go
```diff
@@ -0,0 +1,46 @@
...
+
+// Extension is one installed unit's declaration.
+type Extension struct {
+	// Name is the canonical extension name and must equal the
+	// extensions/<name> directory name, stable across versions. Exact
+	// grammar: lower-case [a-z0-9] segments joined by single hyphens —
+	// `^[a-z0-9]+(-[a-z0-9]+)*$` — so no leading, trailing, or doubled
+	// hyphen. It keys the namespace at every layer (x_<name>_ tables,
+	// /x/<name>/ paths, the x_<name> database role).
+	Name string
```
lstrojny: Self-validating type?

backend/pkg/extension/jurisdiction/jurisdiction.go
lstrojny: Done in b4f2af3: `jurisdiction.Code` is a named type carrying the alpha-2 grammar (`Validate()`); the registry and the compose preflight delegate to it and `For` takes a `Code`.

backend/pkg/extension/jurisdiction/jurisdiction.go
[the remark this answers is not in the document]
lstrojny: Done in b4f2af3: `RetentionClass.Keep` is now a `jurisdiction.Period` (the date part of an ISO 8601 duration); `Period.String()` renders the interval literal ("P6Y") and the evaluator passes it into `$3::interval`, so Postgres does calendar arithmetic — this also fixes the leap-year shortening cubic flagged (Years*365 let destruction run early on statutory floors). The floor pin test passes on the new path.

backend/pkg/extension/extension.go
lstrojny: Done in b4f2af3, with the Go-shaped caveat: composite literals cannot force parse-don't-validate, so `Name` and `Version` are named types carrying their own `Validate()`, enforced at the one boot chokepoint (`RegisterExtensions`). The grammar now lives once on the published type instead of in compose.

backend/internal/modules/de/pack.go
```diff
@@ -17,20 +17,21 @@ func init() {
...
 type retention struct{}
 
 func (retention) Classes() []jurisdiction.RetentionClass {
 	return []jurisdiction.RetentionClass{
-		{Name: "commercial_correspondence", Years: 6},
-		{Name: "accounting_records", Years: 8},
-		{Name: "books_and_annual_accounts", Years: 10},
+		{Name: "commercial_correspondence", Keep: jurisdiction.Period{Years: 6}},
+		{Name: "accounting_records", Keep: jurisdiction.Period{Years: 8}},
+		{Name: "books_and_annual_accounts", Keep: jurisdiction.Period{Years: 10}},
```
lstrojny, line 35: Mhh, why not P6Y?

PR #234, opened 2026-07-23
feat(extensions): serve a governed agent tool — the yogi reference unit (ADR-0069 §7)

lstrojny:
**Stacked on #231** (the manifest slice — this branches off `feat/ext-manifest`, so review #231 first; the diff here is just the serving work). Rebases onto the base as it moves.

First *served* governed capability: an extension tool that is actually callable through the same MCP registry and admission gate the core tools ride.

## What
- **Published handler seam** — `extension.Tool` gains `Handle` (behavior) and `InputSchema`/`OutputSchema` (client-facing docs advertised via `tools/list`), all stdlib-typed so the published surface stays off `internal/`. The manifest reader **skips** them — behavior and I/O docs aren't part of the §5 governance descriptor.
- **compose adapter + registration** — each handler-bearing tool is adapted to the core `mcp.Tool` seam (mapping the published `Tier`/`Scope` → `RiskTier`/`principal.Scope`) and registered in `registryWithGate`, the one chokepoint every registry (HTTP MCP transport, stdio, runner) funnels through. So extension tools are governed **identically** to core tools.
- **Governance scope** — first-party units serve at their declared tier, exactly as `de`'s jurisdiction ships enabled. Third-party `approvals.lock` (§7 operator resolution) stays a later slice; a handler-less tool remains inert (declared in the manifest, not served).
- **`extensions/yogi`** — a read-only 🟢 tool returning a random Yogi Berra quote, enabled by default: the reference served extension.

## Verification
`make check` green. Tests: adapter mapping + inert-tool skip; **end-to-end `Invoke` through the real admission gate** (admitted with the scope, refused without); the yogi handler returns a known quote.

## Notes
- Serving without full §7 resolution is the deliberate first-party path (matches `de`), not an oversight.
- `InputSchema`/`OutputSchema` are hand-written here; struct-reflected schemas (invopop/jsonschema) compose cleanly since the manifest skips these fields — deferred to keep the published surface stdlib-only.

---
## Summary by cubic
Serve governed extension tools through the same MCP registry and admission gate as core tools, and add `extensions/yogi`, a read-only tool that returns a random Yogi Berra quote.

- **New Features**
  - `extension.Tool`: added `Handle`, `InputSchema`, `OutputSchema` (stdlib-typed); manifest/generator skip these fields.
  - Compose adapts handler-bearing tools to `mcp.Tool` and registers them via the single `agents.Registry` chokepoint; handler-less tools are inert; name collisions with core verbs fail at boot.
  - `extensions/yogi`: `yogi_quote` (auto-execute/read) with multiline schemas and a simple handler. Tests cover mapping and admission allow/deny.

- **Bug Fixes**
  - Reject serving confirmation-required tools until approval staging is supported.
  - Derive `Egress` from the send scope; default a missing input schema to `{"type":"object"}`; require declared schemas be JSON objects rooted at `type: "object"`. MCP `tools/list` now includes `outputSchema` and omits an empty “Maps to …” clause.
  - Boot now validate-then-apply: build tool adapters before applying jurisdiction packs to avoid half-applied state on failure. Removed remaining spec/ADR citations from extension comments.

<sup>Written for commit c980d79db63b6791af32952362a161982ea7d10e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/234?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ coverage.out
 !/extensions/.gitkeep
 !/extensions/approvals.lock
 !/extensions/de
+!/extensions/yogi
 
 # Compiled cmd binaries left in the module dir by `go build ./cmd/<role>`
 # (anchored so they never match the cmd/<role> source directories; the api
--- a/backend/internal/compose/extensions.go
+++ b/backend/internal/compose/extensions.go
@@ -29,6 +29,14 @@ func RegisterExtensions(exts []extension.Extension) error {
 			jurisdiction.Register(p)
 		}
 	}
+	// Adapt the handler-bearing tools to the core seam and stash them for
+	// every registry compose builds. Mapping failures are impossible after
+	// preflightTools, but stay fail-closed: a bad mapping aborts the boot.
+	tools, err := buildExtensionTools(exts)
+	if err != nil {
+		return err
+	}
+	setComposedTools(tools)
 	return nil
 }
 
--- a/backend/internal/compose/extensions.go
+++ b/backend/internal/compose/extensions.go
@@ -24,17 +24,19 @@ func RegisterExtensions(exts []extension.Extension) error {
 	if err := validateExtensionSet(exts); err != nil {
 		return err
 	}
+	// Do every fallible step before applying anything, so the whole
+	// reconciliation stays validate-then-apply: adapting the handler-bearing
+	// tools to the core seam can (in principle — preflightTools already
+	// precludes it, but this stays fail-closed) fail, and it must not fail
+	// with a jurisdiction pack already half-applied.
+	tools, err := buildExtensionTools(exts)
+	if err != nil {
+		return err
+	}
 	for _, e := range exts {
 		for _, p := range e.Jurisdictions {
 			jurisdiction.Register(p)
 		}
-	}
-	// Adapt the handler-bearing tools to the core seam and stash them for
-	// every registry compose builds. Mapping failures are impossible after
-	// preflightTools, but stay fail-closed: a bad mapping aborts the boot.
-	tools, err := buildExtensionTools(exts)
-	if err != nil {
-		return err
 	}
 	setComposedTools(tools)
 	return nil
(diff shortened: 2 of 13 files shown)


lstrojny:
## Review wave addressed (`c980d79d`)

- **P1 — served 🟡 tools** ✅ A handler-bearing `TierConfirmationRequired` tool is now rejected at build. You're right: the data-only adapter can't implement the registry's staging seam, so a served 🟡 tool would be refused on every call with no way to stage its approval. Fail closed until the staging seam is wired; a handler-less 🟡 tool (a manifest request) is still fine.
- **P2 — Egress** ✅ Derived from the send scope.
- **P2 — inputSchema null/non-object** ✅ Declared schemas must now be a JSON object rooted at `type:object`; a tool that omits an input schema defaults to `{"type":"object"}` so `tools/list` never advertises `null`.
- **P2 — outputSchema not advertised** ✅ `tools/list` now serializes `outputSchema` (it was dropped for **core** tools too — this fixes both).
- **P3 — "Maps to ."** ✅ The description omits the clause when a tool maps to no crm.yaml operation.
- **P3 — "inert" docs** ✅ The `Tool` doc now distinguishes a served handler-bearing tool from an inert handler-less one.
- **P3 — schema gen/boot parity** ✅ `Validate`'s doc notes the schema checks are boot-only (the manifest omits schemas by design).

**P0 — third-party tools served at their declared tier without per-capability resolution:** this is a *deliberate* pre-governance scope decision, not an oversight. Until an approvals record binds a decision to each tool's digest (the governance slice), the composed set **is** the trust boundary — the vanilla tree ships only first-party units and an installation adds a unit deliberately, the same trust a jurisdiction pack rides when it ships enabled. Now documented in `buildExtensionTools`. Per-capability operator resolution lands with the governance slice; a distributed, less-trusted unit is not the model until then.

Tests added for the served-🟡 rejection, egress + object-schema defaulting, and non-object/malformed schema rejection. Note the `outputSchema` + "Maps to" changes touch the shared `stdio.go` transport (they were pre-existing gaps for core tools); no tools/list test asserts that content, so core behaviour is unchanged apart from now advertising outputSchema.

PR #253, opened 2026-07-25
capture: T1 correspondence gate on a provider-attested outbound signal (ADR-0072 2b slice 2)

LarsGradion:
ADR-0072/A118 phase 2b, slice 2 — the T1 correspondence-positive gate, on evidence a forger cannot write.

## Why

The tiered creation gate ran T2 (transactional/ESP suppression) with no T1 spare in front of it, so a known contact whose newsletter footer carries a `List-Unsubscribe` header was suppressed as bulk infrastructure. ADR-0072 §1 puts correspondence-positive **first** for exactly that reason, and calls the order load-bearing.

The naive version of this gate was pulled from phase 2a on a security review, and that finding is what shapes this PR. `activity.direction` is derived by string-comparing a message's `From` header against the mailbox owner. A spoofed `From:owner` message delivered into the synced inbox therefore reads as the owner's own correspondence — and if T1 honored it, that message would whitelist any address it names past T2 suppression.

## What

**The evidence is the provider's, never the message's.** New `activity.counterparty_outbound_attested` (migration `0124`, plus a partial index serving only the `EXISTS`) is stamped from what the provider vouches for:

| connector | signal | notes |
|---|---|---|
| Gmail | the `SENT` system label | read off the same `messages.get` response the body already needs — no extra call |
| IMAP | an RFC 6154 `\Sent` **special-use** mailbox | a mailbox merely *named* "Sent" attests nothing: that name is operator config text |
| Microsoft Graph | `parentFolderId` vs. the SentItems well-known folder | backfill only — the incremental delta is inbox-only, so it can attest nothing |

`mailmap.Parse` cannot set the attestation at all; only `AttestSentByOwner`, called by a connector holding an authenticated provider handle, can. A provider that attests nothing — or a LIST / folder probe that fails — attests `false` and is logged. Under-attestation suppresses a legitimate contact (recoverable, visible); over-attestation would create records for a forger.

**A spare is as visible as a suppression.** When T1 overrides a matched suppression rule, that lands its own `capture_correspondence_spared` breadcrumb carrying the rule that would have fired — this is the one path that turns registry-declared infrastructure into a record, so a wrong registry entry stays diagnosable.

## Tests

- `mailmap` — a message whose `From` claims the owner parses as `direction=outbound` **and still attests nothing**.
- `gmail` — `SENT` attests, `INBOX` does not.
- `imap` — a mailbox *named* `Sent` with no special-use attribute attests nothing through the full sync path (against the in-memory server); the attribute check itself covered both ways.
- `graph` — SentItems-filed mail attests, inbox-filed mail whose `From` claims the owner does not; an unresolved sent folder keeps the backfill running, unattested.
- integration (real Postgres) — writing to an address spares its later `List-Unsubscribe` mail from suppression **and** logs the spare; a forged `From:owner` neither attests nor whitelists the ESP address it names.

`make check-backend` green; full zero-skip `make test-integration` green.

## Upstream spec raise (flagged, not worked around)

ADR-0072 §1's ladder reads `T1 → ensure person+org NOW`. Taken literally, a free-mail address the owner has corresponded with would mint a "Gmail" organization — precisely the junk the ADR exists to prevent. This build keeps T3's free-mail org suppression under a T1 spare (**T1 overrides T2 only**), gated by a test. The ladder wording needs reconciling upstream (P3 — raised, not patched from here).

## Still open in 2b

The pending-counterparty ledger + deferred creation, the `capture_counterparty_verdict` job, the review queue, and noise hide-then-redact. Then phase 3.

---
## Summary by cubic
Adds a T1 correspondence gate that uses provider‑attested “sent by owner” evidence to spare known contacts before T2 suppression. T1 now outranks T2 and records a `capture_correspondence_spared` breadcrumb; includes migration 0124 adding `activity.counterparty_outbound_attested` and keeps free‑mail org suppression.

- **New Features**
  - New `activity.counterparty_outbound_attested` with a partial index; T1 overrides T2 and logs `capture_correspondence_spared`.
  - Attestation requires BOTH provider filing and outbound authorship; enforced via `connector.Counterparty.WithOwnerAttestation` (unexported) and guarded by `backend/attestationproducer_test.go` to keep minting in `capture/mailmap`.
  - Provider signals:
    - `gmail`: `messages.get` returns RFC822 plus a `SENT` label flag; carried through `mailmap.AttestSentByOwner`.
    - `imap`: attests only when the selected mailbox has RFC 6154 `\Sent` special‑use; a name‑only “Sent” is ignored; LIST failures stop and retry.
    - `graph`: backfill compares `parentFolderId` to `SentFolderID`; refuses a 2xx with no Sent Items id and refuses listed messages missing `parentFolderId`; delta is inbox‑only and unattested. Pages stop and retry when evidence can’t be resolved.

- **Bug Fixes**
  - `graph` backfill now counts a per‑message `ErrSkip` (deleted since listing or oversized MIME) as skipped and continues the page instead of failing and retrying it.

<sup>Written for commit ac802ff7f7ea7c1721d93b640a15bf460ccff8ff. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/253?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added provider-attested “sent by owner” correspondence detection across Gmail (SENT label), IMAP (\Sent), and Microsoft (Sent Items).
  * Attested outbound correspondence can spare eligible emails from suppression, and recorded activity now carries the attestation signal.
* **Bug Fixes**
  * Prevented forged sender headers and non-owner-authored “sent-filed” messages from receiving “sent by owner” credit.
  * Improved backfill/sync safety: progress stops when sent-folder evidence can’t be verified.
  * Added clearer system diagnostics for correspondence spare decisions.
* **Tests**
  * Expanded coverage for attestation, suppression/spare behavior, and sent-folder/backfill edge cases.
* **Documentation**
  * Updated status documentation for the landed slice and remaining open work.

### DIFF
--- a/STATUS.md
+++ b/STATUS.md
@@ -900,10 +900,51 @@ Open work, roughly in priority order:
   activity row + raw_capture under their own retention (the "noise is not stored
   in the append-only spine" hardening; human-authored activities keep their full
   image).
-  **Still open (contract-first-gated on ADR-0072):** the rest of 2b — the T1 gate
-  with an authenticated outbound signal + the pending ledger + deferred creation
-  + the `capture_counterparty_verdict` job + review queue + noise
-  hide-then-redact — and 3 (corroborated signature org-name promotion).
+  Slice 2 (landed): **the T1 correspondence-positive gate on a provider-attested
+  outbound signal.** T1 now runs BEFORE T2 (order is load-bearing: a known
+  contact's `List-Unsubscribe` newsletter is no longer suppressed as bulk
+  infrastructure), and its evidence is a new `activity.counterparty_outbound_attested`
+  column (migration 0124 + a partial index serving the EXISTS) stamped only from
+  what the PROVIDER vouches for — Gmail's `SENT` label (read off the same
+  `messages.get` response the body needs), an IMAP `\Sent` **special-use**
+  mailbox (the folder NAME attests nothing — it is operator config text), and
+  Microsoft's SentItems `parentFolderId` (backfill only; the incremental delta is
+  inbox-only). `activity.direction` is never consulted: it string-compares the
+  forgeable `From` header against the owner, so a spoofed `From:owner` landing in
+  the synced inbox would otherwise whitelist any address it names past T2 — the
+  security review's finding, now gated by an integration test. A T1 override of a
+  matched suppression rule is its own `system_log` breadcrumb
+  (`capture_correspondence_spared`), so a spare is as diagnosable as a
+  suppression. A provider that attests nothing (or a LIST/folder probe that
+  fails) attests false and is logged — under-attestation suppresses, it never
+  creates. `make check` + the full zero-skip integration lane green.
+  Attestation requires BOTH halves — the provider filed the message as sent AND
+  the message names the owner as author — because the two are derived
+  independently: a server-side rule can file a third party's mail into a `\Sent`
+  mailbox or Sent Items, and that message's counterparty is its *sender*, so
+  attesting on placement alone would buy a stranger the same bypass the forged
+  header would have. Provider evidence accrues asymmetrically and this is
+  inherent, not a bug: continuously from Gmail, from Graph only during backfill
+  (the delta is inbox-only), and from IMAP only when the operator points the
+  connection at a `\Sent` mailbox — so an absent T1 spare on a default-INBOX
+  IMAP workspace is expected.
+  **Residual risk for the ADR (raised, no code change):** a rule-based Outlook
+  "reply with template" emits a genuine `From:owner` message with no
+  `Auto-Submitted` header, so an outsider can still induce a T1 spare by mailing
+  in and letting the rule answer. Inherent to the ADR's premise that writing to
+  someone is intent; the cheap mitigations if it ever matters are requiring two
+  attested outbound messages, or discounting an attested message that is an
+  immediate reply to an inbound one.
+  **Upstream spec raise (not worked around here):** ADR-0072 §1's ladder reads
+  T1 → "ensure person+org NOW", which taken literally would mint a "Gmail"
+  organization for a free-mail address the owner has corresponded with — exactly
+  the junk the ADR exists to prevent. The build keeps T3's free-mail org
+  suppression under a T1 spare (T1 overrides T2 only) and gates it with a test;
+  the ladder wording needs reconciling upstream.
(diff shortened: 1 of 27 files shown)


LarsGradion:
Both CodeRabbit threads addressed.

**1. `graph/backfill.go` — skippable message failed the page.** Confirmed and fixed in 08ff0f98. `GetMIME` returns an `ErrSkip`-wrapped error for a message it deliberately refuses (an oversized MIME blob); the incremental pull steps over it, the backfill returned it as a page fault, so the engine retried the page from its committed token onto the same message forever. One unreadable mail would wedge a whole backfill. It is now counted as skipped and stepped over, with a regression test. Pre-existing on `main`, fixed here because the file was already open.

**2. `mailmap_test.go` — forged `From:` plus Sent placement.** This one is a known, documented residual rather than a fixable defect, and the requested regression test would assert the opposite of the intended behaviour.

Attestation deliberately requires BOTH halves: the provider filed the message as sent AND the message names the owner as author. A message the owner genuinely sent has `From: owner` — so `From: owner` + Sent placement attesting is the *correct* path, not the bug. The test asserting `SentByOwner() == true` for that shape is asserting the feature.

The residual CodeRabbit is pointing at is real but narrower: an owner-side rule that files *inbound* mail into the sent container (sieve `fileinto "Sent"`, an Outlook move-to-SentItems rule) could combine with a self-domain spoof to defeat the conjunction on Graph and IMAP. It is recorded as residual (a) in migration `0124`, alongside three others. Two independent adversarial reviews concluded it is not reachable by an unaided outsider: it needs mailbox write access, or owner-side misconfiguration *plus* a self-domain spoof that DMARC and M365/Gmail internal-spoof protection exist to stop. Gmail is not reachable this way at all, since `SENT` is applied by the send path and filters cannot set it.

The case CodeRabbit says the second test misses — a third party writing from their own address into the sent container — **is** covered, by `TestAttestationRequiresAuthorshipNotJustPlacement`: placement alone does not attest, because the counterparty of such a message is its sender.

PR #257, opened 2026-07-25
capture: stop auto-enrich spending when it is switched off, cap automatic reads, and take authority from the dossier row

LarsGradion:
Two spend defects on a **default-ON** path, and the authority question underneath them. Found by an external review of the merged ADR-0072 phases; independent of #255/#256.

## Turning auto-enrich off did not stop the spending

The sweep read the flag before enqueuing and the worker never read it again, so a job queued while the feature was on **outlived the decision to turn it off**. An administrator who disabled auto-enrichment kept paying for crawls and model calls until the queue drained — while the sweep's own comment claimed toggling off *"stops new reads on the next sweep even for already-queued work"*, a promise the worker did nothing to honour.

The re-read now happens **before any page is fetched**, and that placement is the whole point: cancelling after the crawl records the same status and saves nothing. The test asserts on `pageCalls`, not on the outcome. Only the automatic lane is gated — a human who asked for a read is not governed by the automatic-enrichment setting.

Such a read is now `cancelled`, a new terminal status, because none of the existing ones were true: it did not fail, nothing went wrong, and a failure is something an operator investigates while this is not.

**That one value travelled seven layers**, and every gate caught its own: database CHECK → Go allowlist → OpenAPI enum → generated backend types → generated frontend types → the screen's exhaustive status→label `Record` (which refuses to compile) → both locale files. Labelled "Cancelled" / "Abgebrochen", because the entire reason it exists rather than reusing `failed` is that an operator should be able to tell a withdrawn read from a broken one.

⚠️ One mistake worth flagging for reviewers: rebuilding the status CHECK from the table's *original* definition silently dropped `deferred`, which migration **0104** added later. The budget-deferral tests caught it. The migration now states why restating every value matters.

## Automatic reads had no page ceiling

They crawled on the deployment-wide budget (40 pages) while the setting ships on and sweeps up to ten organizations per workspace per day. ADR-0072 §9 puts automatic reads at twelve: a read nobody asked for should cost a fraction of one somebody did. STATUS.md called this "not blocking" — I think that was the wrong call for something that spends money on a default-on path.

The ceiling is enforced at the worker **and** declared in the payload, and both only ever **narrow**. `withPageCeiling` ignores a value that is not lower than the configured cap, so a payload cannot raise what an operator set — the one direction a cap must never move, and what the test pins hardest. It copies the crawler rather than mutating it, so a per-run cap cannot outlive its run.

## The authority question underneath

`BeginSiteRead` already `RETURNING`s the claimed row's own `requested_by`, with a comment saying that is exactly what it is for. Three decisions were keyed on the **job payload** instead — and the sharpest one **predates this branch**: the auto-apply lane decides whether a read applies fields to an organization *directly* or stages a confirm-first proposal. Keyed on job metadata, a payload disagreeing with the row could skip the human.

All decisions governing spend or authority now read `claim.RequestedBy`. So does the **principal**: provenance is written once and never re-derived, so a payload naming the wrong human hangs their name on rows permanently, with no later gate to catch it.

Both are pinned by mutation rather than assumed — reverting either argument to the payload fails a named test:

```
0 proposals on behalf of the requester the row names, want 1
1 proposals attributed to the payload's requester — the row is the authority
```

The settings-read fault also now records itself instead of returning a bare error while the read sat claimed.

`make check-backend` exit 0, `make check-fe` exit 0, full zero-skip integration lane green.

---
## Summary by cubic
Stops spend on auto‑enrich jobs when turned off by re-checking at claim, caps automatic reads at 12 pages, and attributes spend/authority/provenance to the dossier row’s requester. Adds a new “cancelled” terminal status across API, DB (with migration), generated types, and UI.

- **Bug Fixes**
  - Re-check the auto-enrich setting on claim and cancel before any crawl; only automatic jobs are gated and settings-read faults are recorded.
  - Spend/authority/provenance now use the claimed row’s `RequestedBy` (not the payload), and the auto-apply lane keys off it; the worker principal is rebuilt from the claim.
  - Only `human:<uuid>` yields an `on_behalf_of` owner; system namespaces no longer attribute to a person.

- **New Features**
  - Automatic reads are capped at 12 pages; the sweep sets `MaxPages` in the payload and the worker clamps it with `withPageCeiling` (narrowing only, no mutation).
  - Contract/DB/UI: add “cancelled” to the OpenAPI enum and DB CHECK (with migration), regenerate backend/frontend types, and label the status in both locales.

<sup>Written for commit 05050141d054eb30a3c7f53872fa9ca6017a56d0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/257?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added a **Cancelled** terminal status for deep reads, with localized UI labels in English and German.
  * Automatic enrichment runs now apply a per-run page ceiling and can stop/cancel when auto-enrichment is disabled.
  * Deep-read approvals and results are attributed to the requester shown on the originating record.
* **Bug Fixes**
  * Fixed deep-read status handling end-to-end so cancelled/failed terminal outcomes are correctly recorded and enforced.


LarsGradion:
All three addressed in 05050141.

**`requestedByUserID` accepted any namespace** — valid and the one with teeth. A `system:<uuid>` would have populated `OnBehalfOf`, attributing a read to a person who never asked for it, which is precisely the provenance mistake this PR exists to fix. It now requires the `human` namespace, with a table pinning every shape that must yield no owner (`system:`, `agent:`, bare uuid, unparseable, empty).

**Both comment findings valid too.** The terminal-state invariant above `run` still listed three outcomes and `FinishSiteReadInput.Status` still documented three. Worth stating the general lesson: a new value in a CHECK constraint, a Go allowlist and an OpenAPI enum is not done until the prose describing them agrees — that value crossed seven layers in this PR and the comments were the two that no gate could catch.

PR #259, opened 2026-07-25
fix: two post-paint races carried forward from the overlay-UI PR

tinnhlt-gra:
The two fixable items the overlay-UI PR (#258) explicitly carried forward.
Both are the same species — work scheduled after the browser paints, on a
context or a state that has already moved on — and neither is
overlay-specific in its blast radius.

## 1. The record modals seeded after paint (`fix(forms)`)

`EditRecordModal` prefilled its fields in a `useEffect`. React runs passive
effects only after the browser has painted, so for that gap the form is on
screen with every field blank — focused and typeable. Anything typed into
that gap is written through empty form state and then thrown away by the
prefill landing a commit later: the field snaps back to the record's old
value and Save carries the edit the user never made.

It reproduced in ~22 of 25 e2e runs that typed into a freshly-opened modal
without first waiting for a known prefilled value, and it reaches **every
`EditAction` call site, native mode included** — this was never an overlay
bug.

`CreateRecordModal` had the identical shape with a milder symptom: a
reopened form showed the abandoned previous attempt for one frame before
the reset landed.

Both now seed **during render** on the closed→open transition (React's
documented "adjust state when a prop changes" shape), so the inputs' first
commit already carries their values and the gap does not exist. Keying off
the transition — not off `record`/`fields`, which a background refetch can
re-identify at any moment — is what still keeps a re-render from wiping
live input; that part is unchanged.

## 2. Disconnect's credential delete died with the request (`fix(overlay)`)

`Disconnect` commits its teardown and then deletes the sealed incumbent
credential. That delete ran on the caller's **cancellable** request context,
so a client that hung up the moment it had its 204 cancelled the cleanup
before it started — deterministically, every time — leaving an orphaned
(revoked, inert) blob in the vault. No retry can ever reach it: a second
`Disconnect` finds no active connection and answers `ErrNotFound`.

`reconnect.go` already guarded this with `context.WithoutCancel` plus its
own deadline, and `cleanupOrphanedRef` carries the same shape. Rather than
write it a third time, the post-commit delete both lifecycle paths need is
now one function — `deleteUnreferencedRef` — owning the detach, the
deadline, and the never-fail-the-caller posture in one place.

## What I dropped, and why

My earlier STATUS.md note also blamed `RecordFormBody`'s per-field writers
for spreading a `values` snapshot instead of using a functional update, and
I started by fixing that too. **It is not a reachable defect** and I removed
the change: React flushes discrete input events synchronously, so two field
writes never share a batch. I verified this three ways — `fireEvent` inside
`act`, native `dispatchEvent` inside `act`, and reverting the fix with the
test in place — and the test could not be made to fail. Shipping a
prop-contract change plus a test that can only pass would have been noise
(T11), so the diff is the seeding fix alone. STATUS.md now records the
correction.

## Verification

Every test here was falsified — reverted the fix, confirmed the test goes
red, restored it:

- `TestDeleteUnreferencedRefOutlivesTheCancelledRequestThatTriggeredIt` —
  red without the detach (`vault.Get after cleanup = <nil>, want ErrNotFound`).
  Its vault stub refuses work on a cancelled context the way any real
  backend would; the in-memory keyvault ignores `ctx` entirely, so **without
  that stub the test would have passed either way**.
- `edit.test.tsx` first-frame test — red with the effect restored.
- `create.test.tsx` reopen test — red with the effect restored
  (`expected 'Abandoned' to be ''`).

Both frontend tests read the input inside a **layout effect**, which runs in
the commit itself, before paint and before any passive effect — so they
observe exactly the first frame a user could act on, not the settled state.

## Gates (all run locally, exit codes read directly)

- [x] `make check-backend` — **exit 0**
- [x] `make check-fe` — **exit 0**
- [x] `make test-integration` — **exit 0**, `OK: integration passed with 0 skips (25 packages, parallel)`
- [x] `make fe-uat` — **exit 0**, 5 stories captured (`screens-create--open`, `--dedupe-error`, `--form-body-filled`, `screens-edit--prefilled`, `--version-skew-error`)
- [x] `make frontend-e2e` — **exit 0**, 45 passed
- [x] `craft static` over this diff — **PASS, 0 blocker / 0 major / 0 minor**
- [x] pre-push craft hook — passed

One trap worth recording for whoever hits it next: the first `make
check-backend` failed with 13 gosec findings in files that **do not exist** —
a stale `golangci-lint` cache still holding the deleted
`margince-next-overlay-ui` worktree. `golangci-lint cache clean` (the
GOPATH-pinned binary, not PATH) cleared it; nothing was wrong with the code.

## Still carried forward (not this PR)

- **Real-portal verification** of the HubSpot adapter against the live API —
  still the one unchecked box from #258; needs a HubSpot developer account.
- **Upstream (P3)**: per-record freshness on REST reads is contract-blocked;
  owner-on-create would unblock create in overlay; capture writes native
  `activity` rows that overlay reads never serve.

---
## Summary by cubic
Fixes two post-paint races: record modals now seed during render to avoid a blank first frame and dropped keystrokes, and post-commit credential cleanup now outlives the request to prevent orphaned vault blobs.

- **Bug Fixes**
  - Record modals (`EditRecordModal`, `CreateRecordModal`) seed form state during render on the closed→open transition. Removes the blank-first-frame gap and prevents lost input or stale values on reopen.
  - `Disconnect`/`Reconnect` delete unreferenced credentials on a cancellation-detached context with a 5s timeout. Cleanup is best-effort and never fails the caller.

- **Refactors**
  - Introduced `deleteUnreferencedRef` to centralize post-commit credential deletes and reuse the detach + deadline behavior across lifecycle paths.

<sup>Written for commit 9a977f8745fe056926f394c09ff0240514ef91d5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/259?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Create forms now reopen with a clean, default state after being canceled.
  - Edit forms display existing record details immediately when opened, avoiding a briefly blank form.
  - Improved cleanup of unused connection credentials during disconnects and reconnects.
  - Connection cleanup now completes reliably even if the triggering request is canceled.


tinnhlt-gra:
**Re: the Docstring Coverage pre-merge warning (25% vs. an 80% threshold) — pushing back, no change made.**

I checked every symbol this PR adds. The threshold is counting four things, and none of them should carry a doc comment:

| Symbol | Why no docstring |
|---|---|
| `deleteUnreferencedRef` | **Has one** — 17 lines, covering the detach, the deadline, and why it never fails its caller. |
| `cancelAwareVault` | **Has one** — explains why the stub exists and what would go untested without it. |
| `cancelAwareVault.Delete` | A three-line method whose entire behaviour is stated in its type's doc comment directly above it. Repeating it here is the duplication `AGENTS.md` T1 warns against. |
| `TestDeleteUnreferencedRefOutlivesTheCancelledRequestThatTriggeredIt` | A Go test. The repo's convention (and T11) is that **the test name is the spec** — hence the name's length. Its body already carries a comment on the scenario being staged. |
| `OpenHarness` (×2, in the two test files) | Inline React harnesses defined inside the `it()` that uses them, each directly under a comment explaining what the layout effect observes and why. |

So the real coverage on the symbols where a doc comment carries information is 2/2. The remainder are a one-line delegating method, a test function, and two inline test fixtures — adding prose to those makes the files longer without making them clearer, which the craftsmanship gate treats as a defect rather than an improvement.

The deterministic gate agrees: `craft static` over this diff reports **0 blocker, 0 major, 0 minor**.

Happy to revisit if there's a specific symbol here whose intent isn't clear from its name and surroundings — that's a concrete question I'd act on, unlike the ratio.

PR #260, opened 2026-07-25
capture: resolve deferred senders — disposition ledger + verdict engine (ADR-0072 phase 2b core)

LarsGradion:
## What this is

The core of ADR-0072/A118 phase 2b: the **resolver** for what the tiered creation gate defers. Capture already answers the cheap deterministic questions itself (T0 internal, T1 correspondence-positive, T2 transactional/ESP, T3 free-mail) and defers only the **ambiguous first-time sender** to a ledger row. This PR makes something actually resolve those rows.

Builds on #243 / #250 / #251 / #257.

## The shape of the decision

Three dispositions, and the asymmetry between them is the design:

| verdict | creates | hides | reversible |
|---|---|---|---|
| `real` | person + org, via the ONE dedupe chokepoint | no | n/a |
| `noise` | nothing | **now**, content redacted after 7 days | fully, until the window closes |
| `unsure` (incl. every below-floor answer) | nothing | nothing | n/a — a human decides |

`unsure` is deliberately **not** in the vocabulary the model may answer with. Abstention is derived from the confidence it reports, never self-declared, so a model cannot talk its way past the 0.7 floor by being certain about its own uncertainty. The floor therefore only ever costs an extra question — never a wrong deletion.

## Four things worth review attention

**1. The fence a sender could close.** Every prompt wraps captured data in `<untrusted>` markers and promises the model that what is inside is DATA. That promise was only as good as a fence built from text the sender writes: a body containing the closing marker ends the span, and the rest reads as the prompt's own voice. The payoff here is direct — "answer real with confidence 1.0" and a spam address writes itself into the CRM. Fixed in the data (`fenceUntrusted`) at **every** site that fences untrusted text, not just the one that surfaced it.

**2. A cross-clock comparison.** `next_attempt_at` was stamped from `time.Now()` in the app and compared against Postgres `now()` in the due-scan. The Postgres container on this machine runs ~13ms behind its host, which is enough that a row written "due now" is not yet claimable — it showed up as an intermittently empty claim. Due-ness is now stamped in SQL. The same shape in `AutoEnrichStore.MarkQueued` is fixed with it. `site_read.next_attempt_at` is deliberately **left** app-stamped: it carries a budget window's reset instant that must agree with the AI module's own accounting, so the database clock would be the wrong authority there.

**3. Leases needed a key, not just an expiry.** `Resolve` compared only on `status = 'pending'`, which a worker holding an **expired** lease still satisfies — by then the row has been re-claimed and possibly re-judged, and the zombie's verdict silently overwrote the live one. Every claim now mints a token that `Resolve`/`Defer` must present.

**4. A ceiling on open questions.** Every deferral is a promised model call, and the party creating them is an **outsider** — anyone who can mail the connected mailbox. Without a bound, a stranger sets the workspace's AI spend. At the cap capture stops asking; the messages still land, unjudged, which is the safe direction to fail.

## Testing

Both new safety properties and all three dispositions are proven against a real Postgres, and the guards were **mutation-checked** — dropping the claim token, disabling the cap, zeroing the floor, and ignoring the undo window each fail the test that claims to cover it.

`make check-backend` green; `make test-integration` green with **0 skips** (25 packages).

## Fitness-function obligations

A new AI task and new store writes trip several gates. Each is answered rather than waived by default: the two activity writes are ratified in the RBAC gate (a sweep has no human principal, and gating redaction on one would mean a workspace whose reviewer lost access keeps the mail it decided to redact); the redaction is ratified audit-only (its row was already archived, and the closed catalog defines no `activity.redacted`); the task declares its company-context policy; and it ships two aicert scenarios including the **release-blocking false-noise** case.

## Deliberate posture

A deployment with **no model configured runs no verdicts**, and deferred senders stay deferred. An installation without AI keeps the junk OUT rather than falling back to letting it in.

## Known gap (unchanged from #251)

A deferred message's activity is not yet linked to the person a later `real` verdict creates; the ledger row carries `activity_id` for exactly this and the follow-up is phase 3.

---
## Summary by cubic
Resolves deferred first-time senders with a disposition ledger and a verdict engine to create real contacts, hide then redact qualifying noise, and stage unsure for human review — with per-sender judging, atomic/resumable redaction, per-message undo windows, and strengthened untrusted fencing to block prompt injection.

- **Bug Fixes**
  - Redaction is atomic and resumable: activity text and `raw_capture` purge commit together; the sweep recovers half-done states and late reinserted originals; destruction runs only if the message is still hidden, and the undo window is measured from each message’s `archived_at`.
  - Per-sender judging and accounting: one sender per call blocks cross-sender injection; budget stops now refund the attempt before releasing the lease; created records carry a grammar-correct `captured_by` (`agent:capture-counterparty-verdict`) from the actor.
  - Scope and precedence are tightened: `noise` applies only to the evaluated inbound, non-attested, unlinked mail; later mail raises a new question; it stops once the workspace replies and when a person already exists; time decisions use the capture clock, not sender headers.
  - Input bounds and privacy safeguards: subject, display name, and body are truncated in SQL before prompting; the `<untrusted>` fence is hardened by stripping “<” from data and fencing the whole region (blocks zero‑width/in‑word and subject/body splice escapes) across verdict, `capture_classify`, enrich, and site prompts; bounds also apply on review proposals and SAR reads; the verdict task is pinned no‑payload and never persists to `ai_call_payload`; deferred senders are excluded from the classify backlog; interval bindings use SQL `make_interval`. Documented the fence’s fidelity tradeoff and the planned per‑call nonce boundary (no code change). 
  - Erasure and exports: erased-address correction now applies (CAS on the new status, not `pending`); embeddings are removed when mail is redacted or erased; SAR exports include capture dispositions.
  - Free‑mail invariant is enforced by tier order: removed the inert ledger flag, added an integration check that free‑mail never defers, fixed a dangling claim query reference, bounded display name separately, and removed dead code.

<sup>Written for commit 50b9aa847d2a7838fa7a073fe265eb4fe0ba0745. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/260?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added a deferred “counterparty verdict” pipeline for ambiguous first-time senders with `real`, `noise`, and `unsure` outcomes, including human review and automated ledger reconciliation.
  * Introduced a disposition ledger and background processing for counterparty verdicts.
  * Expanded noise handling: immediate hiding, with redaction after the undo window.
  * Added input hardening to prevent prompt injection/fence escapes, plus a workspace deferral cap (500).
  * Updated privacy exports to include capture dispositions.
* **Bug Fixes**
  * Improved claim/lease retry behavior to avoid stale-worker overwrites and fixed timing comparisons.
  * Tightened safety so batch verdicts can’t misclassify other senders’ mail without solo confirmation.
* **Documentation**
  * Expanded STATUS “Pick up here” for 2b core and remaining work.


LarsGradion:
Addressing the CodeRabbit threads before merge.

**Fixed in this branch:**

- **Go duration strings into `::interval`** — correct and important. Postgres cannot parse Go's duration format in general (the micro sign in `1µs`), and it only worked because every constant happened to be whole minutes; tuning one would have broken at runtime with no compile-time or test signal. Every site in `capture` now binds seconds through `make_interval(secs => $n)`. Eight sibling sites outside `capture` (identity, activities, privacy/retention — the last taking its duration from a jurisdiction extension pack) still have the old shape; that is a follow-up sweep, deliberately not widened into this PR.

- **Unicode bypasses of the `<untrusted>` prompt fence** — also correct. The recognition-based regex leaked repeatedly (ASCII-only `\s`, then a vertical tab, then U+2028/U+2029), and no character class could ever have caught an invisible rune placed *inside* the word, or a marker spliced across two separately-fenced fields (a subject ending `<` plus a body beginning `/untrusted>` — plain ASCII, no Unicode needed). The fence now removes the bracket instead of trying to recognise the marker, and prompts fence the whole untrusted region in one call rather than field by field.

**Known cost, recorded rather than hidden:** bracket-stripping is blunt. Three callers feed evidence gates that want verbatim quotes, so a pricing page reading `<10 users` is quoted back with a lookalike bracket. The gates still pass (both sides compare the same fenced text); what suffers is fidelity of stored evidence. The right fix is a per-call nonce boundary — unguessable by the sender, so the data needs no editing at all — applied to all twelve prompt builders at once. A first attempt at that was written and **reverted**: migrating one prompt while narrowing the shared strip left the other eleven weaker than before, and its hand-rolled case-insensitive scanner panicked on `Ⱥ` and let `İ</untrusted>` through, both reachable from an email body. It is one migration with `regexp` doing the case folding, or it is nothing. Written up in STATUS.md with the failure modes.

**Stale threads:** several comments reference `DueForRedaction` / `MarkRedacted` / `redacted_at`, which no longer exist — the redaction sweep became content-keyed (the absence of unredacted mail *is* the completed state), so a one-shot flag was removed along with the column.

PR #262, opened 2026-07-25
projects: the project object, its surface, and core-record parity (ADR-0073/A119, E21)

LarsGradion:
Implements the **project** — the body of work a client relationship is made of — from the merged spec (`margince-foundation` ADR-0073 / decision A119 / epic E21, chapter `specs/subsystems/projects-and-engagements.md`).

A deal is the wrong noun for what a client relationship is made of: too small, because a programme spawns several contracts over years, and too short, because it ends at closed-won and the relationship does not. So everything about a client piled onto the person and the company, undivided. This is the missing grouping key.

Per the chapter's own frontmatter the project lives in the **existing `deals` bounded context** (ADR-0054 §3 / A99 — a noun never earns its own module), and the stakeholder edge lives with `relationship` in `people`, the module that owns that table.

## What's here

**The vocabulary gate first.** The record vocabulary was restated by a dozen polymorphic CHECKs and again in Go, with nothing holding the spellings together — widening eleven of twelve passed every gate and failed at one insert nobody ran. Nine columns are now bound to two named vocabularies (`datasource.RecordType`, `datasource.EntityType`) through `TestEveryDomainEnumMatchesItsSchemaCheck`. It was landed green against the old 4/5-type sets *before* the sweep, so it demonstrably catches a half-done job — it named all nine sites when the migration ran ahead of the Go consts.

**Schema** (`0127`, `0128`, `0129`): `project` + `project_phase_history`; the `activity_link` project arm with `uq_activity_link_project` making a second project link on one activity structurally impossible; `deal.project_id` guarded by a `deal_project_same_org` constraint trigger; the `project_stakeholder` relationship arm; the RBAC default-grant backfill for existing workspaces; `advance_phase` in the audit verb vocabulary.

**Surface**: nine operations (`/projects`, `/projects/{id}`, `/advance`, `/stakeholders`), four events, MCP tiers, and the deal's project pointer end to end.

**Parity** (PROJ-AC-10): searchable in both lanes, a context-graph anchor and hop-2 neighbour, taggable/listable/segmentable/exportable, custom-fieldable, and reachable by every agent record verb.

## Two things worth a reviewer's eye

- **The activity link-walk gained its project branch.** Without it an activity linked *only* to a project matches no branch of the scope clause and quietly leaves every scoped read — a failure that looks like missing data rather than a permissions bug. Pinned in both directions by `TestAnActivityLinkedOnlyToAProjectFollowsItsRowScope`.
- **The phase move is its own verb**, not a settable column, so the row change, the history row and `project.phase_changed` are written from one transaction. `UpdateProjectRequest` has no `phase` field by design.

## Deliberately deferred

`project_link_candidate` (PROJ-DDL-3) is **not** in this PR. Its `evidence` column holds a snippet quoted from a message body, and erasure *scrubs* activities rather than deleting them — so an `ON DELETE CASCADE` would never fire to clean it. It lands with its writer, its erasure coverage and its SAR section together. The deterministic link ladder (PROJ-FORM-1..4) and the AI proposal path follow in their own PRs.

## Raised for upstream reconciliation (not worked around)

1. The chapter's frontmatter names `frontend/src/features/projects`; this repo has no `features/` tree (screens live in `frontend/src/screens/`).
2. The spec tiers `createProject`/`updateProject` 🟡 while the sibling `createDeal`/`updateDeal` are 🟢. Implemented as specced (`confirmation_required`) — the conservative direction — but the inconsistency with `deal` is worth settling upstream.
3. **Overlay is an open edge** (PROJ-EXT-4): the chapter says no ticket may assume overlay coverage, so `project` is deliberately not wired into `modules/overlay`.

## Verification

`make check-backend` green. `make test-integration` green for every package this change touches (`compose`, `compose/integration`, `deals`, `people`), including all seven new project cases.

Note for whoever runs the lane locally: `margince_test` is a machine-wide template shared across worktrees, so a parallel session on another branch can rebuild it out from under a run — that produced several misleading failures here before it was spotted.

---
## Summary by cubic
Introduces `project` as a first‑class record with full API, search/graph/list/segment/export parity, events, RBAC, and per‑record sharing. Closes remaining parity gaps by unifying the activity link vocabulary (filters now include `lead` and `project`) and aligning schemas/messages with the canonical record set.

- **New Features**
  - API: `/projects` list/get/create/update/archive and `/projects/{id}/advance`; project stakeholders via `people`; idempotent `POST /v1/projects` and `POST /v1/projects/{id}/advance`.
  - Record parity: searchable (text + embeddings), graph anchor/related, listable/taggable/segmentable/exportable; custom fields; per‑record sharing; activity relink/links now include `project` and the activity list filters by `project_id` and `lead_id` via a single link vocabulary.
  - Integrations: deals/leads accept and filter by `project_id`; MCP tools support search/read/update/archive/advance for `project`; confirm‑first routes are version‑pinnable and agent admission policy includes `project`; public events `project.created|updated|phase_changed|archived`.
  - Validations: mapped 422 responses for key format, date range, close‑without‑reason, and cross‑company deal→project; relink second‑project refusal returns 422 with guidance to use `replace_existing_of_type`; update/archive honor `If‑Match` with 409 `version_skew` on stale writes; list validation/error messages derive from the canonical record vocabulary.

- **Migration**
  - Run migrations: `0131_project`, `0132_project_rbac`, `0133_audit_advance_phase`, `0134_lead_project` (down sequence made data‑safe; deal↔project same‑org trigger is self‑contained).
  - Client: use `/projects` (+ `/projects/{id}/advance`); optional `project_id` on deal/lead create/update and filters; subscribe to `project.*` events.

<sup>Written for commit 90a453ac466dfc409c424219db6bdf04c56c374b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/262?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Introduced first-class **Projects** with full REST support (list/create/get/patch/archive), phase advancement, and project↔stakeholder management.
  * Added public webhook events: **project.created**, **project.updated**, **project.phase_changed**, **project.archived**.
* **Enhancements**
  * Extended **search**, **saved views/exports**, **attachments**, **activity relink**, and assistant/MCP tools to recognize **Projects**.
  * Added project-aware deal/relationship filters and improved visibility behavior, including key collision/refusal details.
* **Tests**
  * Expanded project lifecycle coverage across unit, HTTP, integration, and event behavior.


LarsGradion:
**Update — the lifecycle questions, and a defect they surfaced.**

Reviewing this against the spec turned up four lifecycle questions ADR-0073 left implicit. They are now founder-decided and pinned upstream as **A119 Amendment 1** (margince-foundation#1188 — `PROJ-LIFE-1..4`, `PROJ-DDL-4`), and implemented here.

**The one that was a bug, not a feature.** `MergeOrganization` relinks domains, employment, deals and relationships — but nothing moved `project.organization_id`, because the project did not exist when that code was written. A project's anchor is `NOT NULL … ON DELETE RESTRICT`, so it cannot stay on a dissolved company: merging A into B moved the deals to B and left the projects on A, and `deal_project_same_org` then refused the **next** edit of those deals. A deal that had been fine became un-editable, reported as a company mismatch nobody created. Merging duplicate company records is routine, so this would have been hit.

Merge now re-anchors projects — **before** the deals, because the trigger is `INITIALLY IMMEDIATE` and moving a deal while its project still points at the old company fires it mid-merge. Where **both** companies carry live projects the merge is refused and names them (`409 both_companies_have_projects`) rather than piling two sets of bodies of work onto one company on a guess: the same posture the link ladder takes, applied to merge.

**Also here.** A lead may now belong to a project (`lead.project_id`) — the object had conversations, people and deals, which assumed it is only populated once there is commercial intent, and that is wrong at the head of its own ladder where a project is born before any deal exists. The arm carries **no** same-company guard and cannot: a lead has no company. That limit is stated in the migration rather than left to be discovered. `phase` is confirmed advisory (a closed project still accepts work) and closing the last deal does nothing — both now pinned rather than implicit.

**A second defect, caught by the test written for the first.** The shared list prelude was returned by value while its `arg` closure captured the local, so every filter a caller registered afterwards appended to a dead copy and the query went out one placeholder short. It surfaced as an opaque driver error on the MCP search path, nowhere near the cause. Now a pointer, with a filtered-list test that fails loudly if it is ever copied again. That extraction itself started as a `dupl` finding and ended as the right shape — the query skeleton is what repeats, not the record.

`make check-backend` green. `make test-integration`: **OK: integration passed with 0 skips (25 packages, parallel)**.

LarsGradion:
Worked through all 14 open threads. Summary of what changed, what was already fixed, and what is deliberately deferred.

## Fixed

- **`ListActivities` link vocabulary** — worse than reported: the filter map held `person|organization|deal`, missing **both** `lead` and `project`, while writes accepted all five. An activity linked to a lead or project was unfindable by filtering on the link just written. It now calls `linkColumn`, the same resolver the write uses, so there is one vocabulary. Covered by `TestTheTimelineFilterKnowsEveryLinkTargetTheWriteAccepts`.
- **`ActivityLink` schema excluded `project`** — real, and caused by widening `relinkActivity`. The governing spec already carries `project` there; this copy was behind. Fixed.
- **`CreateList` refusal message** — derived from `memberEntityTables` instead of restating it, so it cannot drift again.
- **Rollback aborts while project rows exist** — correct, and the comment had the ordering backwards. `ADD CONSTRAINT … CHECK` validates existing rows, so narrowing before deleting aborts the rollback. On a fresh schema either order looks fine, which is why apply-reverse never caught it. Rows are deleted first now.
- **`assert_deal_project_same_org` scoping** — added the `workspace_id` predicate. The composite FK already forbids a cross-workspace `project_id`, so no outcome changes; the point is that the trigger can be judged correct without reading the RLS policy beside it.
- **Relink disclosure** — the ungated delete is fixed (scoped to links the caller can see) and the residual is now *gated by a test* rather than described in a comment. The one bit that remains — refuse-vs-succeed on an invisible project link — cannot be closed from that layer, because hiding a link's existence and enforcing one-per-activity are the same question. What is pinned: the refusal names neither the project nor its id, and the link survives.

## Already fixed since the review

- **`advance_phase` missing from `AuditLogEntry.action`** — it is present in `crm.yaml` now.

## Deferred, with reasons

- **`PATCH /v1/projects/{id}` idempotency allowlist** — I added it and a fitness test rejected it: the map must mirror the contract, and the governing spec does not declare `IdempotencyKey` there, while it does for the identical deal route. That is an upstream inconsistency to reconcile in `margince-foundation`, not to patch into this derived copy.
- **Typed `PublicEventProjectUpdated` payload / storekit call-site shape** — an established-pattern question for the author, not a correctness defect.
- **`search/graph.go`** — untouched by my work; leaving it to the author.
- **Derive the row-scoped FK target set** — legitimate and a genuinely better design, but a heavy-lift refactor of a shared fitness gate. Adding `project` closed this PR's gap and immediately surfaced five undeclared FKs, each now classified against the code that gates it.
- **`project.visibility` unused in Go** — noted; it is inert today and matches how other tables carry it.
- **RBAC down-migration removing grants the up did not write** — real but narrow (only on a database whose roles already carried `project`), and it is a down-migration edge on an unshipped migration.

PR #263, opened 2026-07-26
docs(status): hand off the prompt-fence nonce migration

LarsGradion:
Follow-up handoff for the work merged in #260.

The prompt fence that shipped is sound but blunt: it strips `<` from untrusted text, which is what makes the marker unforgeable, but it costs fidelity in the three callers whose evidence gates want verbatim quotes (a pricing page reading `<10 users` is quoted back with a lookalike bracket).

The intended fix is a per-call **nonce boundary** — unguessable by the sender, so the data needs no editing at all — applied to all twelve prompt builders in one change.

This records it as a *startable* piece of work rather than a wish: the site list, plus the three traps each already hit once during #260 and each reverted —

1. narrowing the shared `fenceUntrusted` while migrating only some callers makes the unmigrated ones **weaker** than before (they rely on strip-all as their whole defence);
2. hand-rolled case-insensitive matching that indexes the original string with offsets from a lowercased copy **panicked** on `Ⱥ</untrusted` and let `İ</untrusted>` through intact — both reachable from an email body;
3. appending a nonce sentence to a system prompt that still names the generic `<untrusted>` marker leaves the model a contradiction to resolve.

Docs only.

---
## Summary by cubic
Document the follow-up to the prompt fence: move from replacing ASCII `<` with a lookalike to per-call nonce boundaries across all twelve prompt builders. STATUS.md now enumerates all sites to update (`internal/compose` and `modules/agents/runner/window.go`), clarifies that per-caller migrations are safe but narrowing `fenceUntrusted` must be the final step, and lists the three traps (no hand-rolled case-insensitive matching; replace—don’t append—the boundary sentence).

<sup>Written for commit da544e0560ff537d14df5bc009cdffaf80d23b2e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/263?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Documentation**
  * Added a “Next Session Starts Here” work item documenting planned security hardening for handling untrusted prompt content.
  * Documented the migration scope, boundary requirements, and key implementation pitfalls.

### DIFF
--- a/STATUS.md
+++ b/STATUS.md
@@ -1002,6 +1002,42 @@ Open work, roughly in priority order:
   `<untrusted>` prompt fence was forgeable by sender-controlled text, now defused
   in the data at every fencing site (verdict, classify, signature-enrich, deep-read
   passages).
+  **NEXT SESSION STARTS HERE — the prompt-fence follow-up (#260 is merged).**
+  This is a scoped, self-contained piece of work; everything it needs is below.
+
+  *What to build:* replace the bracket-stripping prompt fence with a per-call
+  NONCE boundary, across ALL prompt builders in `internal/compose` in ONE change.
+  Each block gets a marker like `untrusted-<uuidv7>`; the block's own system
+  prompt names that marker as the only data boundary; the untrusted text is then
+  passed through byte for byte, because a sender who has never seen the nonce
+  cannot close a span bounded by it.
+
+  *The twelve sites* (a half-migration is worse than none — see below):
+  `captureverdictask.go`, `captureclassify.go`, `captureenrich.go`,
+  `enrichextract.go`, `sitesnippet.go`, `modelraterefresh.go`, `fxrefresh.go`,
+  `fxextract.go`, `offerdraft.go`, `sitepagefacts.go`, `siteprofile.go`, and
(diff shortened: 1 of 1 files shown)


LarsGradion:
Both corrections applied in `da544e05` — worth making precisely, since this block exists to be read cold by a fresh session.

- The fence **replaces** every ASCII `<` with the lookalike `‹`; it does not strip or delete anything. Reworded, and the mechanism is now stated explicitly next to the cost it causes (that is *why* quoted evidence loses fidelity).
- The hazard is narrowing the shared fence **before all callers are converted**, not migrating callers incrementally — converting them one at a time with the shared fence left alone is safe. Trap 1 now says exactly that, and the site list points at it rather than claiming a half-migration is inherently worse.

PR #264, opened 2026-07-26
prompts: bound untrusted data with a per-call nonce the sender cannot spell

LarsGradion:
Finishes the prompt-fence work #260 left open, in one change across every prompt builder.

## The problem

Every prompt that shows a model captured text wrapped it in a fixed `<untrusted>` marker. The sender writes that text, so the sender can write the marker: a body carrying the closing tag ends the span early and everything after it reads as the prompt's own voice. In the counterparty verdict that pays directly — escape the fence, instruct "real" at confidence 1.0, and a spam address writes itself into the CRM.

Recognising a forged marker turned out to be unwinnable (four regexes, four holes; and two attacks no character class catches at all — an invisible rune *inside* the word, and a marker spliced across two separately fenced fields). What shipped in #260 replaces every ASCII `<` in untrusted text with the lookalike `‹`. Unforgeable, but blunt: three callers feed evidence gates that quote a page verbatim, so a pricing page reading `<10 users` was stored as `‹10 users`.

## The change

The boundary is minted per call in `shared/kernel/promptfence` and named in that call's own system prompt. A sender who has never seen the nonce cannot close a span bounded by it — in any script, from any number of fields — so the data goes through byte for byte and the evidence gates get their fidelity back.

Notes on the parts that aren't mechanical:

- **Every builder converted in one change.** Not because half-migrations are inherently unsafe — converting callers one at a time is fine — but because the shared strip-the-bracket helper had to be deleted last, and it can only go once nothing needs it.
- **Boundary sentences are replaced, never appended to.** Wording that also names a generic marker re-teaches the model the exact thing an attacker forges, and leaves it to resolve the contradiction.
- **`agents/runner/window.go` gains a boundary it never had** — tool output and T2 grounding were interpolated bare. Its fence belongs to the *run*, not one call, because the transcript is cumulative; it travels in the suspended-run snapshot. A snapshot written before this change carries spans no marker can bound, so `Resume` refuses it (`ErrConflict`) rather than continuing under a boundary that is not one.
- **`promptfence.Fence` JSON-marshals** for exactly that snapshot, and `UnmarshalJSON` accepts only a marker this package could have minted.
- **The zero `Fence` panics** rather than emit `<untrusted->`, which every sender can spell. It is a programming error in a prompt builder, decided by code alone — no captured text can reach it.

## Gates and certification

`backend/promptfence_test.go` derives the obligation from the tree: no non-test file may build a boundary out of the fixed marker again.

The aicert corpus moves onto the nonce with one documented example marker, plus a new scenario (`capture_counterparty_verdict/forged_fence_01.yaml`) that forges the old one. The two pin tests now lift the marker out of the scenario, rebuild the shipped prompt around it, and compare byte for byte — so the boundary sentence's wording stays certified, not just the body.

**This changes shipped prompts, so the committed cert records are stale.** A re-certification run is an operational follow-up, not part of this PR.

## Verification

`make check-backend` green (build, vet, lint, arch-lint, unit + fitness tests, contract drift, script gates); `craft static` green on the diff.

---
## Summary by cubic
Replaced the fixed <untrusted> marker with a per-call nonce across all prompts and the agent runner to block forged spans while keeping verbatim data. Cache keys and certification ignore the nonce for stability; onboarding/reply/voice and agent transcripts are fenced by default.

- **New Features**
  - Added `shared/kernel/promptfence`: mints a unique fence per call, names it in the system prompt via `Fence.Rule`, JSON-marshals for suspend/resume, rejects markers it didn’t mint, and panics on zero.
  - All prompt builders adopt the nonce; the old bracket-stripper is removed. Reply drafting, voice build/eval, onboarding company messages/acts, and company context now fence their data (the injector adopts the caller’s boundary or mints one). Crawled page URLs move inside the fence; page headers carry only a stable ordinal outside it.
  - Agent runner wraps tool output and grounding in a run-scoped fence by default (only explicitly trusted tiers are left outside), bounds tool names, and uses a closed source-label vocabulary. The fence travels in the suspension snapshot; resume rejects pre-change snapshots (`ErrConflict`).
  - Router cache keys canonicalize the fence away so identical prompts with different nonces still hit; prompt-provenance fields remain out of the key.
  - Certification: `PromptVersion` is a SHA-256 over canonicalized prompts (nonce removed); records are re-certified. Added a forged-fence scenario; pin tests rebuild prompts around the scenario nonce and compare bytes.
  - Fitness/tests: forbid fixed markers and require `promptfence.New`/`FromMarker` plus `Fence.Rule`; the matcher parses code (comments ignored) and enforces `promptfence`-qualified construction (per-file scope). Capture verdict/classify helpers read the boundary from the SYSTEM prompt, scan only spans opened by it, and assert all captured text appears only within those spans; ids are read via `fencedIDs`.
  - Voice builder parsing now resolves sample spans using the declared nonce from the system prompt to avoid treating user-shaped markers as boundaries.

- **Migration**
  - Suspended runs saved before this change cannot resume; restart those runs.

<sup>Written for commit 2416c4adb4c959e8e412d2e931e3992aa0f341d3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/264?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Security**
  * Enforced per-call prompt data boundaries across all AI flows and tightened handling of boundary-forgery attempts.
  * Preserved user/evidence text verbatim, including bracket-like content, and hardened agent resume to refuse snapshots missing the required boundary context.
  * Added stricter tool-name validation during tool execution.
* **Reliability**
  * Updated AI certification tracking to use scenario-scoped prompt digests and refreshed certification records.
  * Improved completion cache-key stability by canonicalizing boundary markers while still differentiating distinct fenced inputs.
* **Testing**
  * Expanded certification and integration coverage across extraction, drafting, onboarding, voice, and agent workflows, including new forged-fence scenarios.
* **Documentation**
  * Documented prompt-fence/certification expectations in the certification corpus README and status notes.


LarsGradion:
Addressed the four review threads.

**`window.go` — fencing on `TrustTier == "T2"` fails open.** Correct and the most important of the four. `TrustTier` is a free-form string on an exported `Job`, so `"t2"`, `"T2 "` or any tier a later provider invents printed raw. Inverted to default-deny against a closed `trustedTiers` set — the same default `groundingRef` below it already takes — with a test covering all three bypasses.

**`promptfence_test.go` — verify fence construction, not an import-path substring.** Fair; the check was a proxy. It now matches `promptfence.New(`/`FromMarker(` together with `.Rule(`, i.e. a fence exists *and* the model is told which marker it is. Still textual rather than AST, but it checks use rather than presence.

**`FromMarker` — reject markers not minted in this process.** Not taking this one, and the reasoning is now in the code with a test pinning the behaviour. The check is shape, not authenticity, deliberately: the only writer of a stored marker is the runner's own store, inside the workspace's own database. An attacker who can choose values in `agent_run.pending` can already rewrite the transcript that marker delimits and every prompt built from it — a token proving provenance would be a lock on an open door. What a caller must not do is take a marker from anywhere a request can reach, and that is stated.

**Corpus scenarios — keep crawled URLs inside the fence.** The underlying concern is real and is **fixed in production** by this PR: `sitesnippet.go`, `sitepagefacts.go` and `enrichextract.go` now put crawl URLs inside the span, leaving only a page ordinal in the frame. The corpus inputs you flagged don't match production's shape either way — `fxrefresh`/`modelraterefresh` put no URL in the prompt at all, so adding one would make the scenario *less* accurate. That mismatch is the known 'corpus is not pinned to the shipped prompt' gap, recorded in STATUS.md as owed work: the fix repins the scenarios, which changes their `prompt_version` and invalidates the certification records this PR just generated, so it needs its own run.

Verification: `make check-backend` green, `make test-integration` green (25 packages, 0 skips). One local run of `TestPerfBudgetsHoldOnSeededVolumeTier` timed out under load from the certification batches; it passes in 3.4s in isolation.

LarsGradion:
Fixed in 2416c4ad — good catch, and it was the third copy of this defect.

`sampleIDPattern` matched any marker-*shaped* string and `sampleSpanOpen` was the bare ` id="` attribute, so neither was bound to the boundary the call declared. Writing samples are the user's own prose and reach the prompt byte for byte, so a sample containing `<untrusted-<36 chars> id="…">` would have been read as a span the builder never opened — letting corpus text name the sample a quote is attributed to.

Both now resolve through `promptfence.MarkerIn(req.System)`, reusing the `fencedIDs` helper the capture verdict and classify lanes already share.

PR #271, opened 2026-07-27
ai: fence the retry echo, and bound every validator's model echo

LarsGradion:
Follow-up to #264. Two holes on the structured-output **repair** path, both reached by the same sender who steered the first attempt. #264 fenced the first attempt; neither of these was in its diff.

## 1. The §5.2 retry re-injected unfenced text

`withValidatorFeedback` appended the model's rejected output as a bare `assistant` turn, and the validator's message — which quotes tokens out of that output — as a bare `user` turn. Meanwhile the system prompt still says the nonce markers are the **only** boundary.

So a sender who engineers a first-attempt validation failure gets their text into the instruction region on attempt 2 — and `completeEscalated` then sends attempt 3 to a **stronger** model carrying the same turns.

Both echoes now go inside the boundary the prompt already declared, adopting its marker via `promptfence.MarkerIn` → `FromMarker` rather than minting a second one beside it (the same adoption rule as `companycontextprompt.contextFence`). A malformed marker is refused rather than borrowed.

A prompt that declares **no** boundary was shown no untrusted data, so its failed output is not attacker-steered and the echo stays plain — the retry keeps the detail it needs to correct itself.

## 2. Two of three validators echoed model text unbounded

`clampToken` existed only in the verdict lane, with a comment explaining exactly why model-chosen text must not reach a log unbounded. Its siblings never called it:

- `validateClassifyPayload` echoed the model's `id` and `label`
- `signatureShapeValid` echoed its `field` name

Both are written to the operator's log **and**, now that the retry is a real prompt input, fed back into the model. `id` is `schema.String()` — unbounded. Same invariant, three lanes, one spelling.

## Verification

Mutation-tested with the fixes committed: unclamping either validator, or dropping the fence from the feedback turns, fails the new tests; all pass at baseline and after revert.

- `make check-backend` green
- `make test-integration` green, 0 skips, 25 packages

New tests cover the fenced echo, a forged `</untrusted>` inside the rejected output failing to end the span, the unfenced-prompt passthrough, and the malformed-marker refusal.

---
## Summary by cubic
Fenced the retry echo with the prompt’s own boundary and neutralized model-shown markers; also clamped all validator-echoed tokens to keep attacker-chosen text out of instructions and logs.

- **Bug Fixes**
  - Wrapped the failed output and validator message inside the declared fence using `feedbackFence` and `promptfence.WrapAuthored`; adopt the existing marker, refuse malformed ones, keep echoes plain when no boundary is declared. Tests now assert the retry instruction sits after the closing marker and that near‑miss markers pass through as data while exact markers are neutralized.
  - Applied `clampToken` across validators (verdict, classify, enrich) to bound model-chosen `id`, `label`, and `field` before logging or feeding them back into a retry prompt.

<sup>Written for commit 44c4c2e41e403d2b277917a311034dedc7637f12. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/271?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Security & Reliability**
  * Improved protection against oversized or malicious model-generated content in validation errors and retry feedback.
  * Prevented untrusted output from prematurely escaping feedback boundaries during structured-output retries.
  * Added safeguards for malformed or forged boundary markers.

* **Bug Fixes**
  * Validation messages are now limited in size while retaining useful error details.

* **Tests**
  * Added coverage for oversized model responses, boundary handling, and adversarial feedback content.

### DIFF
--- a/backend/internal/compose/captureclassify.go
+++ b/backend/internal/compose/captureclassify.go
@@ -252,15 +252,18 @@ func validateClassifyPayload(payload classifyPayload, batch []unlabeledMessage) 
 		want[m.ID.String()] = true
 	}
 	for _, r := range payload.Results {
+		// Every echoed token is MODEL output, and a sender who got the model to
+		// obey can choose it — so it is bounded before it reaches an error string
+		// that ends up in the operator's log and, on a retry, back in the prompt.
 		if !want[r.ID] {
-			return fmt.Sprintf("result id %q was not requested", r.ID)
+			return fmt.Sprintf("result id %q was not requested", clampToken(r.ID))
 		}
 		if seen[r.ID] {
-			return fmt.Sprintf("result id %q appears twice", r.ID)
+			return fmt.Sprintf("result id %q appears twice", clampToken(r.ID))
 		}
 		seen[r.ID] = true
 		if !classifyLabels[r.Label] {
-			return fmt.Sprintf("label %q is not commitment|meeting|noise", r.Label)
+			return fmt.Sprintf("label %q is not commitment|meeting|noise", clampToken(r.Label))
 		}
 		if r.Confidence < 0 || r.Confidence > 1 {
 			return fmt.Sprintf("confidence %v is outside [0,1]", r.Confidence)
--- a/backend/internal/compose/captureclassify_test.go
+++ b/backend/internal/compose/captureclassify_test.go
@@ -9,6 +9,7 @@ package compose
 // surface, not the model.
 
 import (
+	"strings"
 	"testing"
 
 	"github.com/gradionhq/margince/backend/internal/shared/kernel/ids"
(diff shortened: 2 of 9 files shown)


LarsGradion:
Both fixed in 44c4c2e4 — both were assertions that could only ever pass.

**Instruction placement:** you're right, `fence.Wrap(cause.Error())` is built fresh in the test and cannot contain `retryInstruction`, so it proved nothing. It now reads `out.Messages[2].Content` and requires the instruction to appear *after* the closing marker — the property that actually matters, since the one sentence this codebase wrote must not sit inside the span it tells the model to distrust. Mutation-checked: moving the instruction inside `echo(...)` fails it.

**Forged delimiter:** also right that `"</untrusted>"` is not `fence.Close()`, so the count could only be 1. Rather than swap in `fence.Close()` there, I split the two halves, because a commit landed between your review and now that changes the answer:

- `TestTheModelCannotCloseTheSpanWithTheMarkerItWasShown` injects the real `fence.Close()`. That case needed more than a test — the model is *shown* the marker in the prompt that produced the output being echoed, so it can close the span exactly, without forging anything. `promptfence.WrapAuthored` now neutralises the exact marker before wrapping.
- `TestAForgedMarkerInRejectedOutputSurvivesAsData` keeps the near-miss and asserts it passes through byte for byte.

Both halves together are what separate 'remove one known string' (complete, because exactly one byte sequence closes the span) from the blocklist promptfence exists to avoid.

PR #294, opened 2026-07-28
capture: the auto-enrich lane stages its write instead of applying it

LarsGradion:
## What this closes

The captured-organization auto-enrich lane applied a deep read's organization fields and facts **directly** to the record — no proposal, no human. With foundation #1198 ratifying `capture_auto_enrich` **default ON** as the shipping posture, that path is live for every company the CRM captures mail from.

The input to that path is the company's own website: text an outsider controls, fetched and fed to a model. Staging the write was the last unbuilt mitigation on it.

## The change

`compose/deepreadautoapply.go` → `deepreadautostage.go`. The lane now calls the same `stageProposals` the human lane calls:

- one confirm-first `deepread` proposal for the org fields + facts,
- one `site_lead` per stranger (unchanged),
- `ApplyDeepRead` has **no auto-lane caller left**.

The person-matching step the lane used to inline is now `fillKnownSitePeople`, which fills the people the workspace can already identify and returns the strangers. Staging itself is the one shared spelling in `stageProposals` — the two lanes no longer keep separate copies of it.

### One direct write stays, by design

`ApplySitePersonFields` fills only the **empty** columns of a person the workspace **already records at that company**, on an unmistakable match (exact live email, or exactly one employee matching at ≥0.92). It overwrites nothing and touches no record a human doesn't already have. Left as-is deliberately; called out here so the exception is visible rather than assumed.

### Cursor outcome

Terminal outcome is now `staged`. Migration **0136** widens the `capture_auto_enrich_state.last_outcome` CHECK; `applied` stays in the set because rows written before this migration carry it, and nothing writes it any more.

The set of outcomes that *retire* an org was spelled twice — the Go constants in compose, and an inline `$2 IN ('applied','empty')` in `MarkResolved`'s SQL. Adding `staged` to only one of them is exactly the bug the first test run hit. It is now one `autoEnrichTerminalOutcomes` list in the `capture` module that owns the table.

## Upstream raise required

**This contradicts ADR-0072/A118 as written**, which specifies the auto lane applies directly. Contract-first (P3) says the spec wins, so A118 must either ratify staging or this reverts. Recorded in STATUS.md; it rides with the ADR-0075 `<untrusted>`-boundary raise.

## Gates

- `make check-backend` green.
- Full integration lane green: `OK: integration passed with 0 skips (26 packages, parallel)`.
- `TestAutoEnrichLaneAppliesDirectlyInsteadOfStaging` is inverted into `TestAutoEnrichLaneStagesAndWritesNothingToTheOrganization` — it now asserts exactly 1 staged `deepread` proposal and **zero** rows in `organization_profile_field` / `organization_fact`.

### Unrelated observation

The lane's default `INTEGRATION_TIMEOUT=300s` is marginal for `internal/compose/integration` on this machine (it timed out at 301.9s on one run, passes at 1200s). Pre-existing, not touched here.

---
## Summary by cubic
Auto-enrich now stages deep-read organization fields and facts instead of applying them. Adds a new `staged` cursor outcome, unifies terminal outcome handling, and expands tests for empty and fault paths.

- **Refactors**
  - Replaced the auto-apply lane with staging: calls `stageProposals`; no auto-lane caller of `people.ApplyDeepRead`; `deepreadautostage.go` replaces `deepreadautoapply.go`.
  - Auto lane routes via `autoStage` in `deepread.go`.
  - Known site people are filled via `people.ApplySitePersonFields`; strangers stage as `site_lead`. Introduced `fillKnownSitePeople`; staging path shared with the human lane.
  - Consolidated terminal outcomes in `capture.autoEnrichTerminalOutcomes`; `AutoEnrichStore.MarkResolved` now uses `ANY($3)`.
  - Integration tests cover: stage-only path (1 `deepread` proposal, 0 org writes, cursor `staged`), honest empty reads (`empty`), staging faults (`failed` keeps cursor armed), and cursor-write faults (do not fail the read).
  - Note: This contradicts ADR-0072/A118 (which specified direct apply) and needs an upstream update.

- **Migration**
  - Migration 0136 widens `capture_auto_enrich_state.last_outcome` to include `staged`; keeps `applied` for legacy rows.
  - Down migration maps `staged` back to `applied` before restoring the old CHECK.

<sup>Written for commit 5b34e8baefc2ad824359ee299d9d92e9c3e2ebf8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/294?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Changes**
  * Captured-organization auto-enrichment now stages deep-read findings for confirmation instead of immediately applying organization fields and facts.
  * Auto-enrichment records a new **“staged”** terminal outcome when proposals are generated.
  * For already-existing people, missing site details can still be auto-filled; others follow the normal review proposal flow.
* **Bug Fixes**
  * Improved auto-enrichment cursor/outcome handling so staged work completes correctly and does not trigger unnecessary retries.


LarsGradion:
Pushed `7d24c42` — the coverage gate is addressed and the two code findings are answered below. The ADR question I've escalated to the repo owner; see the last section.

### CodeRabbit: "do not suppress terminal-cursor write failures" (`deepreadautostage.go:64`)

Gated with a test rather than left as a comment: `TestAutoEnrichLaneFinishesTheReadWhenTheCursorWriteFaults` runs the lane with `autoEnrich` over a dead pool and asserts the read still finishes `done` with its proposal staged.

Returning the error here would be worse, not better. At that point the proposals are already staged and waiting for a human. Failing the read would mark the dossier `failed` and hand River a job to retry — re-crawling the site and re-calling the model to recover a bookkeeping row whose only consumer is the sweep's due-scan. Losing it costs one reconsidered org on the next pass, which the dossier-exists gate then filters out.

The genuine failure mode — a fault that strands the org — is the `failed` outcome, and that one is now pinned too: `TestAutoEnrichLaneLeavesTheCursorArmedWhenStagingFaults` asserts the cursor stays armed with a retry time so the sweep picks the org back up.

### cubic: cursor resolved before `finish` runs (`deepreadautostage.go:59`)

Checked, and the stated consequence doesn't hold. Two reasons:

1. **A `finish` failure blocks the org through the dossier, not the cursor.** `ListDueOrgs` excludes any org that already has a dossier row, and the row exists from `StartSiteReadQueued` onward. Whether the cursor says `staged` or stays armed changes nothing about whether the sweep reconsiders that org. (The stuck-`running` dossier itself is the known `FinishSiteRead` CAS gap tracked in STATUS.md under the site-read legal-census entry — pre-existing, not introduced here.)
2. **The retry is not duplicative.** Both `stage` and `stageSiteLead` set `JoinPending: true`, so a re-run with identical findings takes `stageOrJoinPendingInTx` and joins the existing pending approval on `(kind, target_entity_id, diff_hash)` instead of creating a second one. A prior *rejected* offer short-circuits the stage entirely.

Reordering would therefore add a second code path across two files to change no observable behaviour, so I've left it.

### The ADR conflict — both of you flagged it, and you're right

This diverges from ADR-0072/A118, which specifies the direct apply. Contract-first (P3) says the spec wins, so this should not merge on a promise of a later raise. I've escalated the ordering decision rather than deciding it here: either A118 is amended upstream in `margince-foundation` first and this merges behind it, or this branch waits. STATUS.md records the raise. **Please don't merge until that's resolved.**

LarsGradion:
Closing — founder decision (2026-07-28): **auto-write stays.**

The call is that a captured company's web findings should land on the record immediately, marked as AI-written, and be correctable afterwards if they turn out to be wrong. Not staged behind an approval.

That is a legitimate product call and it reverses this PR's premise. The safety argument moves from "a human approves before the write" to "the write is attributable and reversible":

- every AI-written field already carries `captured_by = agent:<task>`, its evidence snippet, its source URL and its confidence;
- a human edit flips that field's provenance to `human:*` and later AI passes never overwrite it;
- what was missing is the ability to FIND them, which is [#300](https://github.com/gradionhq/margince-poc-v1/pull/300) plus a follow-on for fields (see below).

`ApplyDeepRead` therefore keeps its auto-enrich caller and `deepreadautoapply.go` stays as it is. ADR-0075's A118 amendment is withdrawn in [margince-foundation#1201](https://github.com/gradionhq/margince-foundation/pull/1201); §1–§2 (the per-call prompt boundary) are unaffected and stand.

One gap this leaves, being addressed separately: #300 finds records an AI **created**. The auto-enrich case creates nothing — it fills fields on an organization the Gmail connector created — so a record-level provenance filter does not surface it. Finding "organizations with AI-written fields" is the follow-on.

PR #302, opened 2026-07-28
capture: drop the activity index 0137 shipped with

LarsGradion:
**Follow-up to a defect that reached `main`.** cubic raised it on [#295](https://github.com/gradionhq/margince-poc-v1/pull/295); I pushed the fix to that branch and it lost a race with auto-merge, so the index shipped.

## The problem

`0137` created a plain index on `activity`:

```sql
CREATE INDEX idx_activity_bulk_mail_attested
  ON activity (workspace_id, counterparty_email)
  WHERE bulk_mail_attested AND archived_at IS NOT NULL;
```

A plain `CREATE INDEX` holds a **write-blocking lock for the whole build**. On a real installation `activity` is the largest and hottest table in the schema, so applying `0137` pauses mail capture until the build finishes.

`CREATE INDEX CONCURRENTLY` cannot run inside a transaction, and `dbmigrate.Up` wraps every migration in one — so there was no concurrent path to use instead. No migration in this repo has ever used `CONCURRENTLY`.

## The fix

`DROP INDEX IF EXISTS`. Cheap: the lock a drop takes is held for the drop, not for a build over every row. `IF EXISTS` so an installation that hasn't applied `0137` yet (it applies both in one run) and one that has both end in the same place.

**What it costs:** the redaction sweep's index. That sweep is a daily background job with a `LIMIT`, reached through the noise senders' own ledger rows — it was never the query that needed the help. The column `bulk_mail_attested` and all of #295's behaviour are untouched.

**The down is deliberately empty.** Recreating the index is the hazard this removes; a down-migration must not reintroduce it. `0137`'s own down still drops the column.

## If the sweep ever does need an index

It belongs behind a non-transactional migration path built for concurrent builds. That path doesn't exist, and inventing it inside a migration is how this happened. Same reasoning removed the equivalent index from [#300](https://github.com/gradionhq/margince-poc-v1/pull/300) before it merged — where it was doubly wrong, since a partial index on a bind-parameter prefix can't be used by the planner anyway.

## Gates

`make check-backend` green.

---
## Summary by cubic
Drops the `activity` index from 0137 to remove write-blocking that paused mail capture. Adds a 3s lock timeout for the drop; rollback now restores the index; app behavior is unchanged.

- **Bug Fixes**
  - 0139 up: `SET LOCAL lock_timeout = '3s'`; `DROP INDEX IF EXISTS idx_activity_bulk_mail_attested`.
  - 0139 down: recreates the index with `SET LOCAL lock_timeout = '3s'` so reverting past 0139 matches 0137; 0137 is unchanged.

<sup>Written for commit 427eb740c2cda7839af46d3c5e4a2be272d0979f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/302?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


LarsGradion:
Pushed a fix for a fair catch: **the drop was replacing one lock hazard with another.**

The first version argued the drop is cheap because there's no build over every row. That's only half of it. `DROP INDEX` needs `ACCESS EXCLUSIVE` on `activity`, and on a busy table that *acquisition* queues behind whatever transaction is already running — while every request arriving after it queues behind the acquisition. The stall has the same shape as the one this migration removes; it just has a different cause.

`SET LOCAL lock_timeout = '3s'` makes it a fast, loud failure instead: the lock is taken promptly or the migration errors and rolls back. `DROP INDEX IF EXISTS` is idempotent, so the operator retries in a quiet window at no cost.

The asymmetry is what makes this the right trade: **failing to drop the index is a performance footnote; stalling mail capture is an outage.**

`0116` already anticipated this pattern in a comment — *"revisit with a bounded lock_timeout"* — this is the first migration to actually use it.

PR #304, opened 2026-07-29
fix(overlay): stop the agent surface answering for a system of record it cannot reach

dungbanguyen:
Closes the two Critical findings from a pre-open-source review of the overlay
feature, plus nine defects found while fixing them. Reviewed over six
craft/security rounds; the last round was the first to find no blocker.

## C1 — five agent tools answered for a system of record they cannot reach

`run_report`, `catch_me_up_on`, `prep_for_meeting`, `whats_slipping_this_week`
and `draft_follow_ups_for` are wired to engines that read the native domain
tables directly. Reports have a seam verb but it carries only ad-hoc plans, and
the mirror holds no context-graph or pipeline projection at all, so none of them
could ride the Dispatcher. Handed an overlay workspace they queried tables
holding none of its records and returned a well-formed empty result as a
**successful** answer — "nothing is slipping" from a portal with a live
pipeline. `overlay.Provider.RunReport` already declared the correct refusal; no
production caller reached it.

The composition layer now guards each dependency (`nativeonlytools.go`), so the
tools stay mode-unaware. Both REST report operations are shadowed — `RunReport`
and the `ExplainReport` drill-through that returns source rows — and both answer
the declared `unsupported_by_sor` sentinel, the same machine code the tool half
gives. The SPA hiding Reports in overlay is exactly why this must exist
server-side: the callers it does not cover are an agent and a direct API client.

## C2 — agent write-back reached the customer's CRM ungoverned

`update_record` auto-executed into HubSpot. Its only brake was per-field
human-edit precedence, which reads the audit trail — and a mirrored record has
no human audit history, so the conflict set was always empty and every patch
went outward. A closed loop with T2 mirror content, which is
attacker-influenceable by design.

The refusal is at the **seam**, not in the tool, because the tool was never the
only route: the REST twin under an agent passport runs the same toothless split,
and `qualify_lead` writes through the provider with no gate of its own.
Dispatcher's update/archive dispatch is what every write caller funnels through
— including the REST write shadow, which is why the gate sits on
`updateInMode`/`archiveInMode` rather than `Update`/`Archive`. It gates only
verbs the adapter actually serves, so "every write" holds by construction as
`SupportsWrite` changes.

### ⚠️ Scope decision to confirm before merge

It answers **`unsupported_by_sor`, not "needs approval"** — i.e. agent
write-back to the incumbent is a *declared refusal*, which is **stricter than
AC-OV-5's confirm-first intent**.

The reason: no approval an agent could get exists here. Staging one needs an
authority object a human can both see and release, and for a mirrored target
neither holds — `approvals/authority.go`'s decidability probe and
`approvals/redeem.go`'s version pin both read *our own* tables, which the record
has no row in. Naming a path that dead-ends is what the declared sentinel exists
to avoid. Confirm-first agent write-back is therefore blocked on the approvals
layer being able to describe a non-authoritative target; the released-approval
check is the seam that change plugs into. **Human seats are unaffected** — that
stays object RBAC's question, per ADR-0055's split.

If reviewers prefer confirm-first now, that is a governance-module change and
should be its own PR rather than smuggled into a security fix.

## Nine further defects, each fixed at the invariant

- The egress probe first read the **cached** mode on a mutation boundary; a
  stale `native` answer would open the gate while the dispatch, resolving fresh,
  still routed the patch to the incumbent.
- The MCP confirm-first redemption branch redeemed an approval without marking
  the call released — asymmetric with the auto-execute branch three lines above.
- Six staging sites minted approvals no human could release (the version pin
  re-reads a row a mirrored record does not have). All six now refuse.
- REST could **release** an agent overlay write with nothing re-verified: a 🟡
  twin stages from `If-Match` alone, which a mirror row has no version for, so
  the pin was nil and redemption checked nothing.
- `ExplainReport` — the unguarded sibling of the endpoint C1 guarded.
- REST reports answered `validation_error`, telling callers their input was
  wrong when the request was fine.
- A declared refusal reached agents as a retryable internal fault, in stdio's
  explanation **and** in the Surface-B runner's observation — the loop with a
  step budget to spend on a permanent no.
- T2 taint dropped on `update_record`'s read-back (finding J12).
- `frontend`: the Reports comment claimed these endpoints 404 in overlay; they
  now answer 422.

## Tests

Unit specs for every guard, the backstop's principal and capability handling,
and both directions of the human/agent discriminator. Two derived fitness
functions rather than hand lists: every verb `SupportsWrite` can answer true for
must be egress-gated, and every stageable tool must refuse an externally-held
target (a new one with no args entry fails). An integration suite drives
`compose.NewRegistry` — the constructor `cmd/mcp` and the api role use — against
a real overlay workspace, including the seam itself so the REST and
`qualify_lead` routes are covered rather than assumed.

**Every regression test here was confirmed to fail against the unfixed code**,
not merely to pass against the fixed one. Three tests passed for the wrong
reason during development and were caught that way — one asserted only an HTTP
status where the route returns the same status for a malformed query, so it
stayed green with the guard removed. Re-verified after the rebase: both Critical
controls still fail on mutation.

## Gates

`make check` ✅ · `make test-integration` ✅ (0 skips, 25 packages) ·
`craft static` ✅ 0 blocker. The 6 remaining craft advisories are all
pre-existing and in scope only because this touches their files (`newServer` is
one line *shorter* than on main; the three `naked-any` in `provider_writes.go`
predate this). Refactoring code this PR didn't write inside a security fix would
widen the blast radius.

## Three items for upstream, not this PR

1. The `/me` contract description tells clients unservable reads answer
   `unsupported_in_overlay_mode` — a code that only ever appears nested under
   `validation_error` in this build. Needs spec reconciliation (P3), not a local
   edit.
2. `connectOverlay` / `disconnectOverlay` / `executeOverlayFlip` are annotated
   `Access: "tool"` rather than `human-only`, so an agent can reach
   system-of-record posture changes. Reads like ADR-0055's human-only
   governance class.
3. The remaining 45 review findings are unaddressed here, incl. the mirror's
   fail-closed 2×SLO visibility floor, GDPR erasure not reaching the mirror
   while the docs claim it does, custom-pipeline deals reading `status: "open"`,
   and the explanation doc's nine false claims.

---
## Summary by cubic
Stops agents from answering or writing to a system of record they can’t reach in overlay mode. Adds a seam egress backstop and uncached mode-aware guards so tools return `unsupported_by_sor`, and agent runs don’t retry.

- **Bug Fixes**
  - Guard native-only tools (`run_report`, `catch_me_up_on`, `prep_for_meeting`, `whats_slipping_this_week`, `draft_follow_ups_for`); both tool and REST report ops (`RunReport`/`ExplainReport`) return 422 `unsupported_by_sor` in overlay; guards use the uncached SoR probe.
  - Add seam egress backstop `refuseUngovernedAgentEgress` in `Dispatcher.updateInMode`/`archiveInMode`; gate only verbs `overlay.SupportsWrite` serves; refuse agent writes unless a released approval is present; human seats unaffected.
  - REST write guard refuses agents in overlay even for seam‑served verbs; fast path for human seam‑served writes; resolve SoR mode fresh for writes and ignore stale caches.
  - Mark released calls via `agents.RedeemAndMark`; cannot mark if redeem fails; reads use `agents.ApprovalRedeemed`; setter kept unexported to prevent forging.
  - Refuse staging for mirrored targets with `refuseStagingElsewhere` (applies to `archive_record`, `promote_lead`, `merge_records`, `advance_deal`, `progress_deal`, and `update_record` staging); validate both merge survivor and source.
  - Normalize errors and agent UX: HTTP 422 `unsupported_by_sor`; stdio says “do not retry”; the runner adds a terminal directive outside the data fence and never exceeds the cap (space reserved for the directive).
  - Integration and fitness tests cover native-only guards, stale‑mode cases, seam backstop, human/agent discrimination, no‑marking‑on‑failed‑redeem, and that every served write verb is backstopped.

- **Refactors**
  - Wrap native‑only dependencies with uncached, mode‑aware adapters in `nativeonlytools.go` and wire via `registry.go` using a named `nativeOnlyModeProbe`; rename write resolver to `isOverlayUncached` and clarify docs.
  - Expose `overlay.AllWriteVerbs()` and add tests so every served write verb is covered by the backstop.
  - Record overlay review follow‑ups and two upstream contract items in `STATUS.md` (remaining findings, `/me` code spelling, overlay lifecycle ops access).

<sup>Written for commit 2028bf7be96450829c96a112beaac82dab19519d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/304?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added an egress backstop that blocks agent writes to external system-of-record boundaries in overlay mode until approval is redeemed.
  * Added native-only guards for selected tools/endpoints in overlay workspaces (returns `422 unsupported_by_sor`).
* **Bug Fixes**
  * Overlay agent seam-served writes are now refused.
  * REST report endpoints now return `422 unsupported_by_sor` in overlay mode.
* **Tests**
  * Expanded unit and end-to-end coverage for approval redemption, overlay/tool refusals, and staging “elsewhere” protections.

---

## Relationship to ADR-0075 §3 (write posture)

ADR-0075 §3 landed after this branch was cut and points the other way on write
posture: staging every model-derived write **was proposed and rejected**
(founder, 2026-07-28), and what is pinned instead is *direct apply, attributed,
reversible, findable*. This PR refuses an agent write outright, which is further
than the staging that was rejected — so the distinction is recorded here rather
than left implied.

**This PR does not touch the lane §3 governs.** The auto-enrich and capture
writes run as `PrincipalSystem` (`compose/captureautoenrich.go`), and this gate
fires only on `PrincipalAgent`. The self-filling CRM is unaffected.

**Why egress is a different risk class.** §3's safety argument rests on three
legs and states that "weakening any one of them puts staging back on the table."
For a write into the customer's HubSpot, two are weakened by construction:

- **Reversible** — a human edit cannot flip a HubSpot field's provenance to
  `human:*`, and our fill-only-empty / human-precedence rule does not govern
  another vendor's store. The correction that "settles it for good" natively has
  no equivalent there.
- **Attributed** — HubSpot carries no `captured_by`. The attribution exists only
  in our audit log, not where the data now lives and not where the customer's
  other integrations read it.

§3 itself distinguishes by risk class (citing ADR-0072 §3, where the
irreversible half does require corroboration), and AC-OV-5 independently
classifies incumbent write-back as *egress*, confirm-first "like any external
send." That is the doctrine this PR implements.

**What would change it.** Confirm-first — the AC-OV-5 intent — needs the
approvals layer to describe a non-authoritative target: today its decidability
probe and its redemption version pin both read our own tables, which a mirrored
record has no row in. When that lands, the released-approval check already in
`egressbackstop.go` is the seam it plugs into.

### DIFF
--- a/STATUS.md
+++ b/STATUS.md
@@ -30,6 +30,35 @@ Open work, roughly in priority order.
 
 ### Correctness and security
 
+- **Overlay: 45 of 49 pre-open-source review findings are still open.** The two
+  Critical ones are fixed (the agent surface answering from native tables for an
+  overlay workspace, and ungoverned agent write-back into the incumbent). What
+  remains, in the order worth taking: `docs/explanation/overlay-augmentation.md`
+  carries nine verifiably false claims and is the first thing an OSS reader meets
+  (cheapest, highest exposure); a single unmappable incumbent record freezes its
+  object class forever, because a mapping failure aborts the whole page and the
+  cursor is never saved; `Reconcile` discards the partial watermark it returns on
+  error, so a portal past HubSpot's 10k search window livelocks; backfill is
+  entirely unmetered and nothing paces the 4 req/s bound `meter.go`'s own doc
+  claims it enforces; every closed deal in a custom pipeline reads
+  `status: "open"` because only the default pipeline's stage keys are recognised;
+  ADR-0044's 2×SLO fail-closed visibility floor is unimplemented (`snapshot_at`
+  is written and never read); and Art. 17 erasure never reaches the mirror while
+  the explanation doc says it does. The last two are compliance-shaped.
+
+- **Overlay: agent write-back is a declared refusal, not confirm-first.** A
+  mirrored target has no authority object a human can see and release — the
+  approvals decidability probe and the redemption version pin both read our own
+  tables, which the record has no row in — so staging one would name a release
+  path that dead-ends. `egressbackstop.go` therefore answers
+  `unsupported_by_sor`, which is stricter than AC-OV-5's confirm-first intent.
+  Reconciled against ADR-0075 §3 in PR #304: §3's posture (direct apply,
+  attributed, reversible, findable) governs writes to OUR records, and two of its
+  three legs are weakened by construction across a boundary we do not own. The
+  released-approval check in that file is the seam a real confirm-first
+  implementation plugs into once approvals can describe a non-authoritative
+  target.
+
 - **Recorded idempotency bodies survive Art. 17 erasure.**
   `idempotency_key.response_body` (migration 0033) holds full 2xx `Person`/
   `Lead`/`Activity` bodies for 24h, and `privacy/erasure.go` does not touch that
@@ -375,6 +404,22 @@ this build repo.
   0141), and `interfaces.md` §1 gains an optional `BackfillProgress` seam
   beside `Backfiller`/`Watcher`/`Sender`. Both are additive; neither changes
   what a committed run reports.
+
+- **`/me`'s `system_of_record` description promises a code this build never
+  emits at top level.** It tells clients that unservable reads answer 422
+  `unsupported_in_overlay_mode`, but that spelling only ever appears nested in
+  `details.errors[].code` under a top-level `validation_error`; the overlay read
+  shadows and the report shadows answer top-level `unsupported_by_sor`. Either
+  the description or the split needs to move, and the choice is the contract's.
+
+- **Overlay lifecycle ops are agent-reachable.** `connectOverlay`,
+  `disconnectOverlay` and `executeOverlayFlip` carry `x-agent-access: tool`
+  rather than `human-only`, so an agent acting for an admin can command a
+  system-of-record posture change (connect, or revoke-and-purge). That reads like
+  ADR-0055's human-only governance class, alongside approval decisions and
+  pipeline config. Raised from the overlay review; the annotation source is the
+  contract.
+
 - **ADR-0072 §1 ladder wording** — the ladder reads T1 → "ensure person+org
   NOW", which taken literally would mint a "Gmail" organization for a free-mail
   address the owner has corresponded with, exactly the junk the ADR exists to
--- a/STATUS.md
+++ b/STATUS.md
@@ -46,6 +46,25 @@ Open work, roughly in priority order.
   is written and never read); and Art. 17 erasure never reaches the mirror while
   the explanation doc says it does. The last two are compliance-shaped.
 
+- **The released-approval marker is context-wide, and one transport forwards
+  its pin while the other does not.** `agents.RedeemAndMark` returns a context
+  marked as released, and `egressbackstop.go` acts on that marker — but it
+  authorizes every external write inside that request or `Handle`, not the one
+  `(tool, diff_hash)` the human released, and a workspace flipped to overlay
+  inside the redemption TTL turns a change approved as local into third-party
+  egress. Separately, the REST gate forwards the redeemed version pin as its own
+  `If-Match` (closing the redeem-tx→write-tx window) while the MCP registry
+  discards it, so that window is shut on one transport and open on the other.
+  Both are pre-existing and bounded; binding the marker to the redeemed call
+  rather than to the context would remove the class.
+
+- **The REST merge twin stages a mirrored target without the authority guard.**
+  `POST /v1/{people,organizations}/{id}/merge` reaches `stageRefusal`, which
+  never calls `refuseStagingElsewhere` for either record — the MCP twin does. It
+  fails closed today only as a side effect: resolving the version pin reads the
+  native row, which a mirrored record does not have. That is the pin's doing, not
+  a guard, so a target type with no version column would open it. Worth the pin.
+
 - **Overlay: agent write-back is a declared refusal, not confirm-first.** A
   mirrored target has no authority object a human can see and release — the
   approvals decidability probe and the redemption version pin both read our own
--- a/backend/internal/compose/agentgate.go
+++ b/backend/internal/compose/agentgate.go
@@ -231,7 +231,12 @@ func redeemIfPresented(w http.ResponseWriter, r *http.Request, next http.Handler
 	if pinned {
 		r.Header.Set("If-Match", strconv.FormatInt(pin, 10))
 	}
-	next.ServeHTTP(w, r)
+	// Mark the released call, exactly as the MCP registry does after its own
+	// Redeem. Without this the seam's external-egress backstop would refuse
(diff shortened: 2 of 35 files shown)


dungbanguyen:
All four review findings are fixed in `ee98a6ab`. Each was valid, and each had been raised by the pre-push review rounds and deferred there — which is the more useful takeaway than the fixes themselves.

**`tools_confirm.go` — merge source unvalidated (CodeRabbit, Major).** Correct: `mergeRecords.StageInfo` pinned and validated only the survivor, but a merge archives and relinks the source too, so an externally-held source under a locally-authoritative survivor still minted an approval no human could release. Both records now go through `refuseStagingElsewhere`, and the mixed-authority case you asked for is pinned by `TestMergeRefusesAnExternallyHeldSourceUnderALocalSurvivor`.

**`approvals.go` — exported marker was not enforcement (cubic, P1).** Correct, and this one had already been raised internally with the same proposed shape; I recorded it as advisory and moved on, which was the wrong call. `withApprovalRedeemed` is unexported again and `RedeemAndMark` is now the only way to obtain a released context — marking is a *consequence* of redeeming rather than a restriction stated in a comment. Both dispatch layers (the MCP registry, the REST agent gate) go through it, and the compose test that proves the gate opens on a released call now obtains its context by actually redeeming.

**`nativeonlytools.go` — capability guards read the cached mode (cubic, P1).** Correct. I had documented this as an accepted trade, naming the stale-`native` cost honestly but still accepting it. That was wrong on the asymmetry: a stale `overlay` costs a retry, whereas a stale `native` serves a well-formed empty report *as an answer* — the exact defect these guards exist to remove. They take the uncached read now, including `ExplainReport` and the registry-wired agent guards. Each guards a report or a graph walk, so one indexed workspace-row read is noise against what it protects.

**`runner.go` — trace entries exceeded the cap (cubic, P3).** Correct, and a genuine improvement on the fix that preceded it: an earlier round had me truncate the payload *before* joining so attacker-influenceable length could not evict the terminal directive, which then overran the bound. Reserving the directive's room satisfies both constraints at once — the finding always survives and the entry stays within the cap, marker included. Pinned by an assertion on both properties.

Gates: `make check` ✅ · integration lane ✅ (0 skips) · `craft static` ✅ 0 blockers. Every one of these fixes was mutation-checked — the guard removed, the test watched to fail, then restored.

PR #314, opened 2026-07-29
comms: key a sent message on the identity the provider stamped, not the one we asked for

tinnhlt-gra:
PR #310's live run through a real Gmail mailbox settled a question STATUS.md had carried open since
#303, and settled it against the design: **Gmail rewrites a client-supplied `Message-ID`.** Three
sends produced six `activity` rows — ours under the minted `019fad38-…@localhost`, and the
separately captured Sent-folder echo under a Gmail-minted `CAFAR1txEuKW…@mail.gmail.com`.

#303 reasoned that the sent row and its provider echo share one natural key, so the echo's upsert
would find the row and write nothing. The reasoning is sound; its premise is false. A `Message-ID` a
sender writes is a *request*, and Gmail declines it.

## It was four bugs, not one

| # | What broke | Why |
|---|---|---|
| 1 | **Every sent email appeared twice on the timeline.** | The echo's natural key is Gmail's identity, ours was the minted one, so `ON CONFLICT (workspace_id, source_system, source_id) DO NOTHING` never fired. |
| 2 | **Replies attributed to the wrong row.** | The counterparty's reply roots its `thread_key` at Gmail's identity, and the reply matcher joins outbound activities on `thread_key`. It found the *echo* — a row with none of the send's links, no consent purpose, no draft outcome. |
| 3 | **Our own follow-up sends quoted a chain nobody had seen.** | `anchorThreading` rebuilds `In-Reply-To`/`References` from the stored identity. |
| 4 | **The at-least-once retransmission guard was silently dead.** | `gmail.Send` tells "already transmitted" from "never sent" by searching `rfc822msgid:<minted id>`. Against a rewritten identity that search can never match. |

1–3 share one fix. **4 is deliberately out of scope** — and it is what makes the rest of this PR
shaped the way it is.

## The invariant, and the one it is subordinate to

> An email activity's natural key is the identity the transmitted message carries on the wire.

Minting stays; the system stops *assuming* it was honoured and asks. `gmail.Send` reads the identity
back off the message Gmail just created — via the existing `GetRaw` and **`mailmap.Parse`, the same
function capture parses the echo with**, so the identity the send records and the identity capture
derives are one function of one set of bytes and cannot drift apart.

That invariant is subordinate to an older one:

> **The receipt commits whenever the provider accepted the message. Nothing about bookkeeping may
> roll that back.**

Getting this backwards would have shipped a worse bug than the one being fixed. A reconcile coupled
to the receipt means any bookkeeping error rolls `RecordSent` back, the delivery stays `pending`,
River redelivers, and the retry guard searches for an identity Gmail already discarded (defect #4) —
**mailing the recipient a second time.** On exactly the provider this PR exists for.

So the receipt commits in its own transaction, and the re-key runs afterwards, best-effort, under a
non-cancelable bounded context with a recover boundary. Once the receipt is committed, nothing from
the reconcile phase can reach the dispatcher — not an error, not a panic, not a commit failure.

## The absorb

Capture cannot learn of a send until the send is recorded, so an echo captured before the reconcile
commits already owns the stamped key. Our row absorbs it: links **copied** (not moved), the queued
counterparty review moved, and the echo's natural key **released** — `uq_activity_source` is a
partial index, so `source_system = NULL, source_id = NULL, archived_at = now()` frees the identity
without destroying anything.

An earlier revision hard-deleted the echo. Two reviews killed that: it would have been the only hard
`DELETE FROM activity` in the codebase, it orphaned the echo's attachments (and their S3 objects),
embeddings and provenance where Art. 17 erasure can never reach them — breaking `eraseAttachments`'
stated invariant, "no bytes are ever orphaned with their only key gone" — and it destroyed a
`CommercialCorrespondence` row that carries a 6-year GoBD floor under `extensions/de`.

The echo lookup is also **constrained**, because the stamped identity is parsed from a remote
provider's response bytes: matching on `(source_system, source_id)` alone would let a hostile or
corrupted response nominate any captured activity in the workspace. The candidate must also be
`kind='email'`, `direction='outbound'`, `captured_by LIKE 'connector:%'`, `created_at >= survivor`,
and name the same counterparty. A collision that does not match is a fault, not an absorb.

## Review

Four adversarial reviews — two on the design, two on the code — all dispositioned in the design
directory. They found things that changed the architecture, not just the code:

- **Fable (design)** — the receipt-coupling blocker above. The design as approved would have turned
  every bookkeeping fault into a double-send.
- **Security redteam** — the hard delete, and the provider-string-selects-the-victim lookup.
- **Fable (code)** — the error-reporting path itself could poison the receipt transaction and cause
  the double-send it existed to report.
- **Codex (code)** — two more paths to the same double-send: job-context cancellation after Gmail
  accepted, and the fact that a savepoint cannot protect an *uncommitted* receipt from a failed
  `RELEASE`, a lost connection, or a panic outside the guard. This is what forced the two-transaction
  split. Codex also caught two contract dishonesties: auditing an archive as a `merge`, and emitting
  `activity.updated` with an empty required `changed_fields` delta.

Both blocker fixes were **mutation-tested** — remove the detached context and the cancellation test
fails with `pg: begin: context canceled`; remove the savepoint and the collision test fails with a
real `25P02` on a rolled-back receipt.

## Testing

`make check-backend` green; `make test-integration` green with **0 skips** (28 packages);
`craft static` over the branch diff — 0 blocker, 0 major, 0 minor.

The tests that matter are the failure paths, not the happy one:

- the receipt stands when the reconcile returns an error, hits a real Postgres `23505`, panics, has
  no reconciler wired, loses its transaction's connection, or has its job context cancelled;
- `Send` returns **no error** when the read-back fails — the re-mail guard;
- the absorb refuses a row it cannot show is this message's echo (inbound / human-authored /
  not-an-email / another recipient / no recipient), and never reaches another workspace's row on the
  same identity;
- the absorbed echo **keeps** the links Art. 17 erasure reaches it by;
- the end-to-end proof: a send through a Gmail stub that rewrites the identity, then the echo
  captured from the bytes the provider stored — exactly one activity survives, and
  `engagement.reply` names the send's own row.

**Live testing is not included and is not claimed.** The branch was booted against the real
connected mailbox (`tin.nguyen@gradion.com`, grant holds both `gmail.send` and `gmail.readonly`) and
confirmed syncing, but no mail was sent. The guide for the live run is in the design directory. One
trap it records: **do not test by mailing yourself** — see R4 below.

## Residuals, stated not discovered

- **D1 — the retransmission guard stays broken.** Gmail exposes no idempotency key and once the
  identity is rewritten there is nothing left to search for. Deferring it is only defensible
  *because* the reconcile can no longer cost a second email; had it stayed coupled, this PR would
  have widened the very window it declines to close.
- **D2/R2 — no backfill.** Pre-existing duplicate pairs keep mis-attributing until those threads die.
- **R1** — a follow-up staged before its anchor's reconcile commits forks the thread visually; reply
  detection survives.
- **R3** — the archived echo is never swept (the retention selector requires `archived_at IS NULL`).
  The price of respecting the correspondence floor.
- **R4** — a self-addressed send does not collapse: the send stamps the first `To` as counterparty,
  capture stamps the first *non-owner*. Fixing it means one spelling of "who was this message with",
  which is an ADR-0072 semantics change.
- **R5** — an E10 subscriber is never told the identity moved, because the survivor's re-key is
  audit-only (below).

## Upstream (P3)

1. `data-model` §7/§8 fixes the mail natural key as the RFC822 `Message-ID` without saying **whose**
   — the sender's request or the transmitting provider's assignment. That ambiguity is what licensed
   #303's assumption, and this build now has live evidence the two differ.
2. `activity.updated`'s `changed_fields` is a closed, bounded delta that cannot represent a
   transport-identity change. The survivor's re-key is therefore **audit-only** and ratified as such,
   rather than publishing an empty required delta that would misrepresent what changed.
3. `SendReceipt`'s doc argued a provider's own identifiers "join nothing here". True of Gmail's
   `threadId`, false of the RFC822 identity.
4. Capture does not persist the provider's internal message id, which is why the absorb's echo
   lookup is a strong heuristic rather than a proof.

---
## Summary by cubic
Record sent emails under the provider-stamped RFC822 `Message-ID`, collapsing Sent-folder echoes to remove duplicates and fix reply attribution. If the provider doesn’t return an identity on read-back, the send still commits and the reconcile degrades without re-keying (no resend).

- **Bug Fixes**
  - Re-key sent messages to the provider’s `Message-ID`; echoes collapse onto the send; replies/threading now attach to the send.
  - Reject invalid or overlong provider-reported identities; if read-back yields no identity, reconcile logs and skips while the send stands.
  - Echo absorb copies links, moves pending counterparty review, releases the echo’s key, archives it (emits `activity.archived`), and matches strictly (same workspace, outbound email, connector provenance, same counterparty, recent).

- **Migration**
  - `connector.SendReceipt` now includes a validated, length-bounded `RFC822MessageID`.
  - `Dispatcher` passes the full receipt to `RecordSent(ctx, id, receipt)`; `Store` requires a `MessageIdentityReconciler`.
  - `gmail` reads back the stored message and parses the identity via `mailmap.Parse`.
  - Send worker binds a system actor and correlation id for audit/outbox during reconcile.

<sup>Written for commit 755db64a17ba1c671c22cd72854e0530488bb163. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/314?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Gmail sends now recognize provider-assigned message identities and consolidate duplicate sent-message echoes into one activity.
  * Replies are more reliably attributed to the correct sent message, including when Gmail rewrites message identifiers.
  * Sent receipts are preserved even when post-send reconciliation encounters an error.

* **Bug Fixes**
  * Improved handling of duplicate echoes, identity collisions, and pre-transmission capture races.
  * Invalid or unavailable provider identities no longer disrupt successful sends.
  * Added safeguards for excessively long message identifiers.


tinnhlt-gra:
## Testing — what was run, and what was deliberately not

Posting this after the merge so the live-run instructions live with the change rather than in a
gitignored working directory.

### Automated — run and verified locally, not just reported

`make check-backend` and `make test-integration` both green, the latter ending
`OK: integration passed with 0 skips (28 packages, parallel)`. `craft static` over the branch diff:
0 blocker, 0 major, 0 minor. Both were re-run after merging `origin/main` (which brought #309/#313
and migration `0142` — no number race, this change adds no migrations).

**The receipt survives everything.** These are the tests that matter, because the failure they
guard is a second email to a real recipient, not a wrong row:

```
TestRecordSentKeepsTheReceiptWhenTheReconcileFails
TestRecordSentKeepsTheReceiptWhenTheReconcileHitsAUniqueViolation
TestRecordSentKeepsTheReceiptWhenTheReconcilePanics
TestRecordSentKeepsTheReceiptWhenTheReconcilesTransactionIsLost
TestRecordSentKeepsTheReceiptWhenTheFaultReportItselfPanics
TestRecordSentKeepsTheReceiptWhenTheBreadcrumbItselfCannotBeWritten
TestRecordSentKeepsTheReceiptWhenTheStoreHasNoReconciler
TestDispatchRecordsTheReceiptWhenTheJobContextIsCancelledDuringTheSend
```

Two of these were **mutation-tested** rather than trusted:

- remove the savepoint → `TestRecordSentKeepsTheReceiptWhenTheReconcileHitsAUniqueViolation` fails
  with the real Postgres evidence — `current transaction is aborted (SQLSTATE 25P02)` caused by
  `duplicate key value violates unique constraint "uq_activity_source" (SQLSTATE 23505)`, and
  `commit unexpectedly resulted in rollback`. That is the double-send, reproduced.
- remove `context.WithoutCancel` → `TestDispatchRecordsTheReceiptWhenTheJobContextIsCancelledDuringTheSend`
  fails with `pg: begin: context canceled`, i.e. `OutcomeRetry` on an already-sent message.

**The read-back, including its degradations:**

```
TestSendReportsTheIdentityGmailStampedWhenItRewroteIt
TestSendReportsTheMintedIdentityUnchangedWhenGmailHonouredIt
TestSendSucceedsWithNoIdentityWhenTheReadBackFails      ← the re-mail guard
TestSendReportsNoIdentityWhenTheReadBackIsNotAUsableOne
TestSendOnRetryFindsThePriorSendAndReadsNothingBack
TestRecordSentRefusesAnIdentityNoMessageCouldCarry
```

**The absorb, mostly as refusals** — an absorb that fires when it should not is worse than one that
never fires:

```
TestAbsorbRefusesARowThatIsNotThisMessagesEcho          (inbound / human-authored / not-an-email /
                                                         another recipient / no recipient)
TestAbsorbNeverReachesAnotherWorkspacesRowOnTheSameIdentity
TestAbsorbLeavesTheEchoTheLinksErasureReachesItBy
TestAbsorbDoesNotDuplicateALinkTheSurvivorAlreadyHolds
TestAbsorbLeavesTheEchosProjectLinkBehindWhenTheSurvivorHasOne
TestAbsorbGivesTheSurvivorTheEchosProjectLinkWhenItHasNone
TestAbsorbRePointsAQueuedCounterpartyReview
TestAnEchoCapturedBeforeTransmitIsAbsorbedByTheReceiptItself   ← production shape, real reconciler
```

**The end-to-end proof and the wiring gate:**

```
TestAGmailRewrittenIdentityStillYieldsOneActivityAndOneReplyTarget
TestEveryComposedDeliveryStoreCarriesAMessageIdentityReconciler
```

The first drives a real send through a Gmail stub that rewrites the `Message-ID`, then feeds capture
the bytes the provider *stored* — exactly one activity survives, nothing remains under the minted
identity, and `engagement.reply` names the send's own row. The second is an AST scan, verified to go
red when a reconciler is nil'd.

### Manual — what was actually done

The branch was booted against the **real connected mailbox** (`tin.nguyen@gradion.com`) and confirmed:

- `/healthz` and `/readyz` 200; boot line `applied 0 core+custom + 0 river migration(s); schema is at head`;
- the running binary built from the branch, process bound to the dev database `margince`;
- the Gmail connector enabled and the grant holding **both** `gmail.send` and
  `gmail.readonly` — the second matters, because `Send` verifies only the send scope and a
  send-but-not-read grant would 403 the read-back and make this fix structurally dead;
- the mailbox actively syncing (`last_synced_at` advancing against a live `history_id`).

Every SQL query in the run-book below was executed against that live database and corrected where
the schema disagreed with the obvious guess (`audit_log.occurred_at`, not `created_at`;
`last_synced_at` is a projection of `capture_sync_state`, not a `capture_connection` column).

### Not done, and not claimed

**No mail was sent.** Sending is outward-facing and irreversible, and the only safe recipient was the
owner's to choose. The live confirmation of the headline claim — one timeline row per send — is
still outstanding.

### Live run-book

Send through the UI under a **fresh subject token**, wait for `comms_outbound.status = 'sent'`, then:

```sql
-- 1. THE PASS/FAIL. One row is a pass; two rows with differing source_id is the defect.
SELECT subject, source_id, thread_key, direction, archived_at
  FROM activity WHERE subject LIKE '<token>%' ORDER BY created_at;

-- 2. The delivery and the timeline agree, and the provider receipt is recorded.
SELECT o.status, o.message_id, o.provider_message_id, a.source_id
  FROM comms_outbound o JOIN activity a ON a.id = o.activity_id
 WHERE a.subject LIKE '<token>%';

-- 3. The evidence that this mailbox rewrites: a system:comms-send 'update' audit row
--    whose before/after name the two identities.
SELECT actor_id, action, before->>'source_id' AS minted, after->>'source_id' AS stamped
  FROM audit_log WHERE entity_type = 'activity' AND action = 'update'
 ORDER BY occurred_at DESC LIMIT 5;
```

Then reply from the counterparty mailbox and assert the reply's `thread_key` matches the send's
re-keyed `source_id`, and that `engagement.reply`'s `matched_outbound_activity_id` is the **send's**
activity id rather than an echo's.

Connect with:
`docker compose -f infra/docker-compose.dev.yml exec -T postgres psql -U margince_owner -d margince -c "<query>"`

### Three traps that each read as a failure of this change

1. **Do not test by mailing yourself.** The send stamps `counterparty_email` as the first `To`;
   capture stamps the first **non-owner** address. A self-addressed send gives the two rows
   different counterparties, the absorb's counterparty predicate correctly declines, and you get two
   rows. That is residual R4, not a broken re-key. Use a real second recipient.
2. **Use a fresh subject token.** The dev database still holds #310's duplicate pairs — this change
   deliberately does not backfill them (D2/R2), so a `subject LIKE` that catches an old pair reads
   as a failure of the new code.
3. **If you ever test from a git worktree**, copy `.env.local` in first. A fresh worktree gets the
   all-commented template, and its `MARGINCE_KEYVAULT_ROOT_KEY` differs from the `scripts/dev.sh`
   fallback — the connected mailbox's refresh token will not decrypt, and the send fails in the
   vault looking exactly like a bug in the code under test.

### Follow-up this change does not do

A read-back that cannot be answered leaves the message under the minted key, and its echo lands as a
duplicate — the degradation is accepted, not repaired. `comms_outbound.provider_message_id` is
durable, so a later pass can re-ask the provider for the identity of a message it already accepted.
That pass is the repair and does not exist yet.

PR #318, opened 2026-07-29
security: patch the four accepted preference-token findings

ancorn-gradion:
## What

The four preference-token findings the owner accepted in `TRIAGE.json`, one commit each. They are four halves of one credential-handling defect: the `pref_` token minted for a marketing send is a bearer credential over one person's consent record — on `/v1/public/preferences/{token}` it reads their per-purpose state, withdraws, and **grants**, under a system principal that short-circuits `auth.Require` and `auth.EnsureVisible`.

| Finding | Commit | The gap |
|---|---|---|
| WEB-f001 | `security(consent): row-scope the preference-token mint` | The mint held the `person:read` object grant but not the row-scope probe its sibling read applies, so a `row_scope=own` rep obtained standing authority over a person who 404s to them. |
| WEB-f007 | `security(activities): keep the preference token off the durable record` | One body was used twice — wire *and* `activity.body` — so the token landed in a durable row `GET /v1/activities` serves back, and in the 202. The seeded `read_only` seat (activity:read, `RowScope: all`, zero write grants) harvested every recipient's token workspace-wide and replayed it to drive a governed GDPR write attributed to the data subject. |
| WEB-f002 | `security(privacy): Art. 17 erasure retires the preference token` | Erasure anonymizes the person in place, so 0048's `ON DELETE CASCADE` never fired and the erased subject's link kept resolving forever. The same erasure already NULLs `list_unsubscribe` out of `comms_outbound` — it destroyed the record of the link while leaving the capability live. |
| WEB-f010 | `security(consent): bound the preference token's lifetime` | No expiry column, one permanently-reused token per person, and a `revoked_at` no production code wrote — so the resolver's only filter passed vacuously and one leaked newsletter was un-rotatable authority for the life of the installation. |

Two follow-ups ride along: a pure extraction to keep `anonymizeSubjectRows` under the 80-line gate, and a renumber of the migration to `0143` after `main` took `0141`/`0142` mid-flight.

## Why

Each fix restores an invariant the surrounding code already claimed:

- `handlers_email.go` says the footer "belongs on the mail, not on this response to the API caller" — the durable copy was the seam neither that comment nor the access-log redaction covered.
- `erasure.go` says erasure "must reach every store that holds the data subject"; `preference_token` appeared in neither `piiTables` nor the cross-store-writer allowlist.
- `rbacgate_test.go` waived the RBAC gate on `ResolvePreferenceToken` because "the signed single-use token is the authority" — it is neither signed nor single-use. That waiver now states the properties this credential actually has.

Both fitness maps are updated rather than just the call sites, so the obligations outlive this PR: `TestErasureAndSARReachEveryPIITable` now fails if the erasure sweep is ever dropped.

Reuse is deliberate and stays — the preference centre is revisitable, and one message's link must keep working after the next goes out. What ends is reuse without a bound.

## How verified

Local, against a real migrated Postgres (`infra/docker-compose.dev.yml`). `make` is not available on this host, so each target's underlying commands were run directly:

- `go build ./...`, `go vet ./...` — clean.
- Unit lane `go test ./...` — failure set **identical to the pre-change baseline** (56 both sides; all are this Windows checkout's CRLF artifact, e.g. 1458 untouched files "missing" a licence header they have).
- Integration lane, packages serialized: `compose/integration` (190s), `modules/{activities,consent,comms,privacy}`, `migrations` — all `ok`.
- `golangci-lint` baseline + `--config .golangci.strict.yml` (the new-code pass, integration build tags on) — **0 issues**.
- `craft static` — 0 blockers; `erasure.go`'s only MAJOR is the pre-existing `ErasePerson` one, unchanged from `origin/main`. `craft residue` clean.
- `check-image-pins`, `check-test-lanes`, `check-go-file-length`, `check-rls-store-path`, `check-no-jurisdiction` — all OK.
- Composition reproducible; no generated-file drift; `crm.yaml` untouched, so no contract change and no frontend lane.

Each new test was checked against a vacuous pass. Reverting the `EnsureVisible` guard makes `TestPreferenceTokenMintRefusesAnInvisibleRecipient` fail with the real credential in the message (`pref_UqaPsw…` handed to a rep who 404s on that person), which is the finding reproduced.

- [x] `make check` is green — run as its constituent commands (see above); the only red is the CRLF checkout artifact, present identically on `origin/main`
- [x] `make test-integration` is green — run directly with the `integration` build tag against the compose Postgres, packages serialized

## Behaviour changes

Reviewers should look here first — these are the changes most likely to surprise someone downstream.

1. **A send to a recipient outside the caller's row scope now 404s** instead of silently succeeding. It deliberately does *not* fall through to "no unsubscribe surface": that would transmit marketing mail with no working `List-Unsubscribe` URL, trading a credential leak for an RFC 8058 violation. An address no person carries still answers `found=false`, unchanged, so the send path does not become an in-CRM/not-in-CRM oracle.
2. **`activity.body` and the 202 render the footer with the token segment as `token-redacted`.** `comms_outbound.body` (the transmitted copy) is unchanged and remains the only place the credential is stored. The integration harness now lifts the token from the staged delivery — where a recipient reads it — instead of the 202.
3. **A token minted more than 30 days before this migration stops resolving on deploy**; the recipient's next message mints a live replacement. The backfill dates existing rows from their own `created_at`, not the deploy, so an old token is not handed a fresh window.
4. `preference_token.expires_at` is `NOT NULL` with no default — the TTL has one spelling, in Go, and a SQL default would be a second.

One existing test asserted the vulnerable behaviour (`logged activity body does not match the transmitted body`). Its assertion and its comment now state the new invariant: the record and the transmission differ by exactly the capability segment, and the test also asserts the live token is absent.

Also fixed in passing, in a function this PR rewrites: `ensurePreferenceTokenTx` returned the token variable *before* the scan that filled it, so a send that lost the INSERT race handed back an empty token and a nil error, rendering a link with no token segment.

## Residue, stated rather than hidden

The age ceiling bounds a leaked token for an actively-mailed recipient, but a token whose recipient stops receiving mail just before the ceiling simply ages out — so the worst case is the ceiling plus one TTL. That is in the constant's doc comment, not left implicit.

## AI involvement

Authored by Claude Opus 5 (Claude Code) under human direction and review, from findings the owner accepted in `TRIAGE.json`. All production code, migrations, tests and commit messages here were AI-generated; every gate listed above was run locally and its real output read, so no result claimed here is asserted from expectation. The triage and threat-model markdown was deliberately kept out of the branch — this PR is Go and SQL only.

## Accountability

off (`git commit -s`, DCO). See [CONTRIBUTING.md](../CONTRIBUTING.md).

---
## Summary by cubic
Hardened preference-token handling to stop leaks and limit authority. Tokens are row‑scoped with a strict live check, kept off durable records and 202s via redaction, expire and rotate with an age ceiling, are deleted on erasure, excluded from SAR, and unsafe sends are refused.

- **Bug Fixes**
  - Gate `PreferenceTokenForEmail` with `auth.EnsureVisibleLive`; out‑of‑scope recipients 404, closing the erasure race. Unknown addresses remain silent (no token, no error).
  - Keep the token off durable records: deliverability returns transmitted and recorded bodies; `activity.body` and the 202 use `token-redacted`, only `comms_outbound.body` stores the live token.
  - Art. 17 erasure deletes `preference_token` rows; added sweep in `privacy` and updated fitness maps.
  - Bound token lifetime: added `expires_at` with sliding refresh and an age ceiling. Resolver requires unrevoked and unexpired tokens; send rotates and revokes when expired or past the ceiling. Fixed a concurrent mint path that could return an empty token.
  - Refuse unsafe sends: fail when the public base URL is unset, and when the unsubscribe linker errors; added tests for both.
  - Exclude tokens from SAR: marked `preference_token` as `sarForbidden` and added a gate to fail if SAR ever reads it.

- **Migration**
  - Apply `0144_preference_token_expiry` (adds `expires_at`, backfilled from `created_at`).
  - On deploy, tokens older than 30 days stop resolving; the next send mints a new token. TTL is defined in Go (no SQL default).

<sup>Written for commit 9540a2becd66a0db81b223e8fdc1b0ba8394abcb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/318?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Security & Privacy**
  * Preference-center tokens now expire, rotate, and become unavailable after erasure.
  * Token creation respects recipient visibility and remains silent for unknown addresses.
  * Live preference tokens are redacted from recorded email activity and timeline views.
  * Marketing emails are refused when required unsubscribe configuration is unavailable or fails.

* **Data Protection**
  * Preference-center tokens now have a bounded lifetime, including existing tokens.


ancorn-gradion:
Thanks both — three of these were real defects in the branch. Addressed in `69cb998`.

**1. `EnsureVisible` → `EnsureVisibleLive` on the mint (cubic P1) — accepted.**

This one is the best catch of the set, and it undercut my own WEB-f002 patch. Statements in a read-committed transaction each take a fresh snapshot, so an Art. 17 erasure committing between the email lookup and the probe left the plain check answering "yes, still yours" for the tombstone — `EnsureVisibleLive`'s own doc comment names exactly that case ("erasure … stamps `archived_at` while LEAVING `owner_id` alone"). The mint would then have issued a **new** public credential for the subject whose old one the erasure two commits earlier had just deleted. Switched to the strict twin. No live recipient is affected: the email lookup already joins `p.archived_at IS NULL`, so only the race changes behaviour.

**2. `sarRead: false` doesn't forbid, it only skips (cubic P1 / CodeRabbit Major) — accepted.**

Correct, and it's precisely the distinction this repo's own rule draws: a known gap in a comment is a rationalization, in a drift-failing table it's a control. I'd written the invariant in prose and enforced nothing. There is now an explicit `sarForbidden` that fails when `AssembleSAR` reads the table. I verified it isn't vacuous by injecting a `SELECT t.token FROM preference_token …` into `sar.go` and confirming the gate fails:

```
--- FAIL: TestErasureAndSARReachEveryPIITable
    piicoverage_test.go:247: SAR reads PII table preference_token — it is
    registered sarForbidden because its rows must never leave in an Art. 15
    package; drop the section in AssembleSAR
```

It also fails a table registered both `sarRead` and `sarForbidden`, since the export cannot both require and refuse it.

**3. Hostile-origin send status discarded (cubic P2 / CodeRabbit Minor) — accepted.**

Right, and it was a weakness I introduced: when I moved the token lookup off the 202 body onto the staged delivery, `transmittedBody` began reading the *newest* row, so a hostile send refused before staging would have left the assertion re-reading the previous benign delivery and passing without ever proving the forged `Host`/`X-Forwarded-Proto` were ignored. The 202 is now asserted before the link is read.

**4. Section header still said "single-use" (cubic P3 / CodeRabbit Minor) — accepted.**

The header claimed one bound for every entry under it, which the updated entry contradicted. It no longer claims any; each entry names its own bound (single use, a signature, an expiry, or nothing but entropy).

---

**Not taken: splitting the 0143 backfill across deploys (cubic P1).**

I don't think this one applies to this class of change here, and the siblings agree:

- **The shape is the house convention.** `0100_ai_call_attempts` and `0107_voice_profile_lifecycle` both do ADD COLUMN + `UPDATE` backfill + `SET NOT NULL` in a single migration. Splitting only mine would make it the inconsistent one.
- **The table is small by construction.** This is a single-organization installation (A107/ADR-0061), and `preference_token` holds at most one live row per contact who has been sent consent-gated bulk mail in that one org — not a general-purpose event table.
- **There's no mechanism to split into.** `cmd/migrate up` applies the embedded sequence in one pass; a resumable batched backfill would need a staged-deploy facility this repo doesn't have, which is a much larger change than the finding.
- The `ADD COLUMN` itself is nullable with no default, so on PG11+ it's a metadata-only change — the cost is the `SET NOT NULL` scan, and I'd rather keep that invariant than trade it for a nullable column a future writer can forget to set.

Happy to revisit if the deployment model changes.

**Verification after the fixes:** `compose/integration` green (196s) plus activities/consent/privacy; unit lane failure set unchanged from baseline; `golangci-lint` strict pass 0 issues; `craft static` 0 blockers.

PR #320, opened 2026-07-30
impl(overlay): the overlay→native cutover lifecycle (ADR-0071 / OVA-AC-6)

dungbanguyen:
Implements gradionhq/margince-foundation#1179 — the overlay→native flip and its ADR-0071 cutover lifecycle, with the OVA-AC-6 (a)–(d) integration lanes and the UC-E18-05 / UC-J-05 e2e derivations green.

## What lands

**The contract (OVA-GAP-1).** `/overlay/flip:preflight` and `/overlay/flip` existed as schema-less stubs; they are now real operations — the preflight's `{ready, blocking[], unresolved_conflicts[]}` verdict with the sealed snapshot and parity preview, the execute request with its typed confirmation phrase and the emergency mode, the 202's disclosed-lossy body, and the 409/404/422 failure surfaces. Also adds `GET /overlay/export`, the pre-flip export's producer. The backwards "read-mode→overlay" direction strings are corrected: the flip is overlay→native.

**The migration engine** (`internal/modules/migration`) — the shared importer the flip reuses (IEM-FORM-1): the RLS-scoped `import_run` store with audited transitions, the checkpointed resumable run loop, the zero-write dry-run whose disposition table discloses every skip with a reason, and `GuardIncumbentSource` — the OVA-AC-6(a) refusal the flip preflight and the direct importer share through one constant. Sources and native writers are injected seams, so the module never imports the record modules it feeds; the UC-E11-03 connectors plug into the same engine later.

**The flip.** Preflight gates conjunctively (incumbent reachable, force-fresh converged, `pending_sync` drained, pre-flip export current), seals the frozen snapshot when green — fenced mirror writes then refuse, sweeps skip the workspace — and unseals on *every* other exit, so a failed preflight is a no-op return to a healthy overlay. Execute is confirm-first behind `FLIP TO SOR`, refuses with 409 `overlay_flip_blocked` naming the reasons, imports the estate with counts, relationships and provenance preserved, then flips `x_sor_mode` to native and clears `x_incumbent` in one audited transaction.

**Emergency cutover** (OVA-AC-6 b): available *only* while the incumbent is unreachable — refused in both directions, never a silent substitute — and its 202 returns the snapshot's staleness plus an explicit unverifiable-parity notice.

**Retirement and reconstruction** (OVA-AC-6 c/d): disconnect-after-flip purges the mirror and derivatives while native data, audit, and the pre-flip bundle survive; the fake incumbent's record state deep-compares byte-identical across the whole lifecycle; and `ReconstructFromBundle` rebuilds a clean native instance from the bundle's mirror snapshot with zero incumbent calls — reversibility as rebuild, with no native→overlay claim.

## Spec-fills disclosed for upstream reconciliation

The spec leaves these open; each is implemented deliberately and flagged rather than silently invented:

1. The `blocking[]` reason literals, including `incumbent_unreachable` — OVA-AC-6(a) names it only in prose.
2. The emergency variant's API shape — no OVA-WIRE row exists. Implemented as a `mode` field with a reachable-refusal rule.
3. `import_run.connector` gains `'mirror'` — IEM-DDL-1's enum never named the flip's source.
4. `x_incumbent` is cleared at flip time (the `x_overlay_iff_incumbent` CHECK forces it); the connection row survives active for retirement.
5. The export bundle's retention window has no pinned value (A92 says "short auto-expiry", no number).
6. Deal pipeline/stage materialization falls back to the default pipeline's first open stage with a per-row disclosure — OVA-MAP-6 leaves it open.
7. `GET /overlay/export` is the lifecycle's bundle producer until IEM-WIRE-1/2's export-run lifecycle is minted.

UC-E18-05 **F2** (disconnecting an un-flipped overlay workspace) and **F3** (teardown partial-failure recovery) are named spec gaps — "do not invent behavior in the test" — so no case asserts them.

## Review

Both review agents ran against the diff and every confirmed finding is fixed in `768f0a0b`, notably: the preflight seals state so it is human-only rather than agent-reachable at `auto_execute`; the freeze is no longer a one-way latch that could permanently block credential revocation and PII purge; concurrent executes are serialized by a workspace advisory lock; seal/unseal are audited; and the pre-flip export is whole-estate for an unbounded actor, without which the OVA-AC-6(d) rebuild promise was nominal.

## Verification

- `make check` — green (backend + frontend)
- `make test-integration` — `OK: integration passed with 0 skips (29 packages, parallel)`
- `craft static` on the diff — 0 blockers

---
## Summary by cubic
Implements the overlay→native cutover lifecycle (ADR-0071 / OVA-AC-6): preflight with snapshot sealing and parity preview, execute (fresh-sync and emergency), retirement, and reconstruction from the pre‑flip bundle — powered by a resumable `migration` engine and a secure external→native identity map. Adds `GET /overlay/export` to produce pre‑flip bundles with human‑only access and export auditing.

- **Bug Fixes**
  - Ensures the flip’s advisory lock is fully released on unlock failure by closing the connection, preventing stuck claims from returning to the pool.
  - `import_record_map` cross‑workspace run references now return NotFound instead of leaking composite FK constraint details.
  - The pre‑flip export gate no longer clears on a zero cutoff; an actual sweep or mirror watermark is required.
  - Malformed bundle errors no longer echo untrusted identifiers.

- **Performance**
  - Memoizes incumbent owner resolution per flip to avoid repeated lookups, reducing import queries during migration.

<sup>Written for commit b08d670c28d719ea6ed437fa150e99cb2972d370. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/320?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added an operator-only endpoint to download pre-cutover overlay export bundles.
  * Implemented overlay-to-native cutover with readiness checks, confirmation safeguards, snapshot sealing, parity previews, emergency options, and resumable execution.
  * Added export-bundle reconstruction and migration run tracking.
  * Extended sync status to show when data is frozen for cutover.
* **Bug Fixes**
  * Prevented disconnects and mirror writes from racing with cutover.
  * Rejected reserved mirror source identifiers in client writes.
* **Documentation**
  * Updated cutover, emergency, retirement, and recovery guidance.


dungbanguyen:
Worked through the 13 unresolved review threads. Seven were real; six are fixed here, and the rest are tracked in #336 rather than resolved silently.

**Fixed in `3430799f` and `c9881f7d`:**

- **Payload-less owned rows were imported as blank records.** `_owner_external_id` was smuggled into `Row.Fields`, so `len(Fields) == 0` never matched for a row that had an owner and the `empty_payload` skip never fired — the writer landed a nameless native record instead of disclosing the system entry. The owner is now a first-class `Row` field, which keeps `Fields` the canonical payload the emptiness check reads. Pinned by a test at the engine and a no-leak assertion on the bundle source.
- **A cancelled preflight left the mirror frozen.** Both unseal sites ran on the request context, so the abandoned preflight that most needed the undo was the one that could not perform it. They now use `context.WithoutCancel`, matching the lock release beside them.
- **The export-recency gate was vacuous on an empty estate.** With no row watermark the cutoff was the zero time, so any export ever written cleared it. The cutoff is now the later of that watermark and the last successful sweep.
- **A resumed run reported only its final leg.** Failures persisted no report at all and completion replaced rather than merged, so someone resuming a crashed cutover was told a fraction of what actually imported. Failures now record their partial dispositions and completion folds by class; proven end-to-end through a real JSON round-trip in the Postgres lane.
- **A malformed bundle row printed the record it could not read.** `%v` of the raw row put every name, email and address into the error. It now names only the missing identifiers.
- **Reconstruction could strand a run at `running`.** Three fallible steps ran after `runs.Create`, and their early returns left a row nothing resumes and nothing clears. Everything fallible now happens before the row exists.

Also swapped the chained type assertions in the flip integration lane for the comma-ok form, and split the readiness verdict out of `flip.go`, which the additions pushed past the 500-line cap.

**Tracked in #336, not fixed here:** the create/`RecordIdentity` atomicity window (needs one transaction across a module boundary — ADR-0054 §9), raw driver text in `import_run.error` (no transport reads it today), bundle whole-estate validation (mitigated by RBAC, since export is admin/ops-gated and those roles are row-scope unbounded), the seal/fenced-write ordering window (affects preview counts, not what lands), and `pgtype.Numeric` in CSV export (pre-existing, relocated not introduced).

**Declining, with the reasoning already in-source:** the missing `event_outbox` rows for import-run state changes are ratified in `auditOnlyWrites` — IEM-GAP-3 declines run-lifecycle events, and the fitness test records that rather than letting it pass unnoticed.

`make check` and `INTEGRATION_JOBS=4 make test-integration` are both green (0 skips, 29 packages).

dungbanguyen:
Second round — all four new threads were real, all four fixed in `4f37d966`.

- **A failed unlock returned a still-locked session to the pool.** The advisory claim is session-level and `pgxpool.Release` hands the connection back with the session intact, so a failed `pg_advisory_unlock` left the claim held by an idle pooled connection and wedged every later flip in the workspace. Worse, the comment beside it asserted the opposite — that the lock dies with the connection. The failure path now hijacks and closes the connection so the claim cannot outlive it.
- **`RecordIdentity` leaked the composite FK's constraint name** when the run belonged to another workspace, which both exposed database shape and confirmed that a run the caller cannot see exists. It is a row-scope miss and now answers as one, through `ErrNotFound`, using the existing `storekit.IsForeignKeyViolation` idiom. The integration assertion checks the sentinel and that neither the table nor the constraint appears in the message.
- **The export gate still cleared with no watermark at all.** My earlier fix fell back to the sweep timestamp, but a workspace that has neither mirrored a row nor completed a sweep has both at zero, so every export in history post-dated the cutoff. A zero cutoff now proves nothing rather than proving everything.
- **Renamed the export-transport subtest** — its principal holds read on `overlay_connection` and is refused for want of update, so describing it as lacking the grant misstated what was under test.

`make check` and `INTEGRATION_JOBS=4 make test-integration` green again (0 skips, 29 packages).

dungbanguyen:
Third round. Four fixed in `b08d670c`, two deferred to #336 with reasoning.

**Fixed:**
- **One owner lookup per imported record.** The import asked `mirror_user_map` who owned every single row it landed — an estate has dozens of owners and hundreds of thousands of records, so the run re-answered the same question per record. Live lookups are memoized for the run, safe because the mirror is frozen while it executes. Owner resolution moved to its own file as it went: it decides whether a flipped record is visible to one seat or the whole workspace, which is a security decision rather than a mapping one.
- **The malformed-bundle refusal still echoed `object_class`,** which is untrusted bundle content like the rest of the row. It now names only which identifier is absent.
- **A lifecycle assertion's message still said "want 2"** after the expectation moved to 3.
- The zero-watermark export case was already closed in `4f37d966` — the re-report predates that commit.

**Deferred to #336 (items 6 and 7):** `x_sor_mode`/`x_incumbent` being read in their own transaction rather than the readiness one — a disconnect/reconnect in that window could pair the old incumbent name with new mirror state, and that name is the import's provenance stamp. And the freeze audit recording `flip_snapshot_id`/`mirror_frozen_at` against `incumbent_connection` when those columns belong to `overlay_sync_state`. Both are transaction- and ownership-boundary changes in the seal path rather than local fixes, and neither is reachable without the operator already holding the flip claim.

Gates green (`make check`, `INTEGRATION_JOBS=4 make test-integration`, 0 skips, 29 packages).

PR #322, opened 2026-07-30
feat(company): the connections card — one hop of the account's graph

LarsGradion:
PR 6 of the company-view rebuild. `GET /organizations/{id}/graph` serves the account's one-hop neighbourhood as nodes and edges, and the company rail draws it as a connections card between the people and deals cards.

## What the endpoint returns

One hop from the account, and only one:

- **contacts** — current employment edges, each weighted by its §4 relationship strength;
- **deals** — the account's open deals, plus the stakeholder seats on the ones it draws;
- **related companies** — the parent, the children, and the A41 partner counterparties in either direction;
- **intro path** — which contact the active signal's warm-intro path routes through.

A contact's other employers, a deal's other accounts and a partner's own partners are not walked. The second hop is a different read with a different cost, and a card that sometimes went two hops would have no honest cap.

## Three rules worth reviewing

**Authorization is per group.** Reading the organization is mandatory; contacts need the person grant, deals the deal grant, the intro path the signal grant. A withheld group is absent from `nodes`/`edges` and named in `groups_omitted`, so the card says "hidden from you" instead of drawing a company nobody works at. Within a group every row carries its own object's row scope — and a stakeholder edge needs **both** the deal and the person to be visible, because an edge names two records.

There is deliberately **no** `related_organizations` group. Parent, children and partner companies need no grant beyond the organization read the endpoint already demands, so they are row-scope pruned like every other node and can never be withheld wholesale. A value nothing can emit would be vocabulary a client had to handle and would never see.

**The intro path reuses the warm room's ranking rather than re-deriving it.** `signals.RankRouteIn` and `signals.RouteInEdges` are extracted from `Warmth`, so both callers rank through one spelling. The card reports a path only when that exact contact is already one of its nodes; when the route-in contact's only seat is on a deal the card did not draw, it says nothing rather than promoting the next contact it happens to have. It never names a different person than `GET /signals/{id}/intro-path`.

**The caps report themselves.** Node order is fixed — contacts by strength then id, deals by amount then id, organizations by name then id — so two reads of an unchanged account draw the same picture. `dropped_count` says how many nodes the caps left out. Stakeholder contacts are the one exception: they have no cap of their own because the deal cap already bounds them, and the contract says so.

## Frontend

`frontend/src/screens/connections.tsx` renders one payload two ways, and the **list is the primary one**. The ego diagram is a hand-rolled SVG on a fixed radial layout, `aria-hidden` with nothing focusable inside it, because everything it draws is in the node list beside it — each row routing through `EntityRef`. "See it larger" reopens the same graph and the same list in a `Modal size="wide"`.

The card reads its own endpoint rather than riding the 360, so it owns its states: a failed read says unavailable, and only a successful one may say the account has nothing attached.

## One thing to expect in the diff

Regenerating `internal/contracts/api_gen.go` reshuffles which enums win the bare constant name — `VoiceBuildStatusCode`, `ListRelationshipsParamsKind` and three inline params enums gain or lose their type prefix. That is oapi-codegen's conflict resolution responding to the new schemas, not a hand edit. No hand-written code referenced any of the renamed constants; the drift gate and the build both pass.

## Gates

- `make check` green (backend + frontend halves).
- `craft static` over every changed backend file: 0 blocker, 0 major, 0 minor.
- The graph's nine integration tests pass in `make test-integration`. The overlay failures in that lane (`TestOverlayRefetchWorker*`, `TestReconcile*`, `TestAcceptance_AC_OV_*`) reproduce identically on a clean `origin/main` — they are pre-existing, not from this branch.

---
## Summary by cubic
Adds the company connections card and `GET /organizations/{id}/graph`, a one‑hop graph of an account’s contacts, deals, related companies, and warm‑intro path. The UI is list‑first with a decorative, aria‑hidden SVG; the API uses per‑group auth with deterministic caps.

- **New Features**
  - Endpoint `GET /organizations/{id}/graph` returns nodes/edges one hop from the org: employees (weighted by relationship strength), open deals with stakeholder seats, parent/child/partner orgs, and the intro path when that contact is present. Ordering is fixed; `dropped_count` reports cap overflow.
  - Per‑group authorization with row‑scope pruning; withheld groups are omitted and named in `groups_omitted`. Edges render only when both ends are visible.
  - New `ConnectionsCard` on the company rail (`frontend/src/screens/connections.tsx`) fetches its own data, renders an accessible node list that states how each item is connected, and an `aria-hidden` SVG ego diagram. “See it larger” opens the same data in a wide modal.

- **Bug Fixes**
  - Enforced the person grant at each read: contacts, stakeholder seats, and `signals.RouteInEdges` now require person access directly (closing a grant‑by‑order gap and aligning Warmth).
  - Fixed caps/counts and bounded reads: totals come from the same statements that fetch rows (no negative `dropped_count`), and SQL now orders/limits deals, contacts, and related orgs while still reporting exact totals.
  - Related‑org cap now counts companies, not rows: SQL picks the first distinct companies by name and collapses duplicate edges; `graphScanCap` now only bounds the contact read.
  - Unified “signals of this org” with `signals.OfOrganizationWhere` so the graph’s active‑signal logic matches `/signals` listing.
  - Named nodes from the payload instead of fetching each via `EntityRef`; an empty `name` still falls back to a lookup, and tests assert one request per card and ID fallback rendering. Authz test now also requires the `referred_by` edge to exist, not just be correctly oriented.
  - Test cleanup and docs: removed an unused map in a cap test and clarified the comment on what `graphScanCap` actually bounds.

<sup>Written for commit 691315505d17480063dd6c5276e664699b80e113. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/322?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added a Connections card to organization pages, showing one-hop contacts, deals, related companies, and relationship details.
  * Added an expanded Connections view with deterministic diagram, accessible node list, intro-path indicators, and “show more” counts.
  * Added a connections graph endpoint supporting capped results and warm-introduction routing.
* **Bug Fixes**
  * Improved handling of unavailable or restricted connection data with clear messaging while preserving permitted information.
  * Added responsive layouts for smaller screens.


LarsGradion:
Both review findings are fixed in 23294186.

**cubic — N+1 fetch per node.** Valid. The list named every record through `EntityRef`, which reads the record to resolve a display name, so a card showing ten nodes fired ten extra requests and each row showed its raw uuid until one landed — for names `GET /organizations/{id}/graph` had already returned. `EntityRef` now takes an optional `name`: supplied, it skips the lookup and keeps the link and the id fallback unchanged. A *blank* name still falls through to the lookup, because a blank is the caller saying it has nothing rather than saying the record is nameless; otherwise the row would render a button with nothing in it. Two tests come with it — one proves the card issues exactly one request, the other that an unlabelled node reads as its id.

**CodeRabbit — vacuous orientation check.** Valid. The `referred_by` loop judged the edges it found and asserted nothing about finding any, so it could only fail on a wrong edge, never on a missing one. It now counts them and requires exactly one.

Auto-merge is armed.

LarsGradion:
Both addressed in 69131550.

**P3 — dead `plain` map.** Correct, it was written and never read. Removed.

**P2 — the employment count is not bounded by the scan cap.** Accurate, and deliberate. Taking it in two parts:

What the cap *does* bound is the rows returned and the work done per row — `people.StrengthForPeople` folds §4 per contact, joining `activity` and `activity_link` for each one. That is the cost that used to grow with the account, and it is now capped at 500 regardless of headcount.

What it does not bound is the count, because the headcount has to ride the same statement as the rows: two statements each take their own Read Committed snapshot, and a concurrent hire between them makes `dropped_count` negative against the schema's own `minimum: 0` (fixed in 57403444). So the count reads one account's employment rows through `idx_rel_org_people` — a single index range scan, scoped to one organization, not to the workspace.

I am keeping that trade rather than relaxing the response contract. An exact `dropped_count` is the promise the whole cap design rests on: a truncated graph reporting no count reads as the whole neighbourhood. One index range scan per account is a fair price, and the alternative — a has_more flag — would lose the number the card shows the rep.

The comment on `graphScanCap` overstated this ("a graph's query size follows the cap rather than the account") and now states both halves plainly, so the next reader is not misled into thinking the count is bounded too.

LarsGradion:
One correction on top of 69131550, in d9dd9bed.

That commit fixed the `graphScanCap` comment and left the **contract** claiming the opposite — "the reads themselves are bounded, so one request's cost follows the caps and not the account". That sentence is false, and putting the honest account in a Go comment while the published description kept the promise is worse than either alone: the people who need it are the ones who cannot see it.

What is actually true, and now what the description says:

- an exact `dropped_count` needs a count over each group's whole membership, so this read **is** proportional to the account it describes;
- what the caps bound is the rows returned and the per-contact §4 fold done on them, which is the part that grows fast;
- if that trade ever has to change, the response shape changes with it — a floor plus a flag — rather than the count quietly becoming approximate.

The paragraph also lost its reassurance about request cost. A client cannot observe our index scans, and prose that exists only to reassure is the worst kind of line in a contract.

PR #325, opened 2026-07-30
feat(deploy): generic container deployment materials

dungbanguyen:
## What

Adds **deployment-target-agnostic** container materials so anyone can self-host margince, on any platform. This repo is open-source and clients choose where to deploy, so it carries only the **generic** pieces — no D13/NFQ/Gradion-instance specifics (those live in a separate private deployment repo).

| File | Purpose |
|---|---|
| `Dockerfile.api` | `cmd/api` (HTTP) + bundled `cmd/migrate`; applies migrations at boot |
| `Dockerfile.worker` | `cmd/worker` — outbox relay, retention, Surface-B AI |
| `Dockerfile.web` | Vite SPA behind nginx-unprivileged |
| `scripts/deploy/api-entrypoint.sh`, `worker-entrypoint.sh` | migrate-as-owner → serve-as-app |
| `scripts/deploy/db-bootstrap.sql` | one-time DB role + database + extension setup |
| `frontend/nginx.conf` | SPA static serving |
| `docs/deployment.md` | platform-neutral self-hosting guide |

## Key design points

- **No baked instance config.** The images read every setting from the runtime environment (`MARGINCE_*` env fallbacks) and, for first boot, a `margince.yaml` mounted at `MARGINCE_CONFIG`. No baked `.env`. Standard 12-factor; the same web image works for any domain (the SPA calls the API same-origin at `location.origin + /v1`).
- **Composed-workspace build.** The Go images run `gen-composition` then build under `GOWORK=build/composition/go.work` (folds in `extensions/de` + `extensions/yogi`).
- **Two-role DB model, bootstrapped out of band.** margince enforces tenant isolation with `FORCE ROW LEVEL SECURITY`, which only a superuser bypasses — so the app must never connect as one. `scripts/deploy/db-bootstrap.sql` creates the non-superuser `margince_owner` / `margince_app` roles + database + extensions (`vector` is untrusted) once, as a superuser. The api entrypoint then migrates as owner and serves as app. This is a hard requirement of the app on **any** Postgres, not a D13 thing.
- **Non-root containers.** api/worker run as uid 10001; web on nginx-unprivileged (uid 101, port 8080). `pnpm install --ignore-scripts`.

## Verification

- ✅ All three images build clean from the repo-root context, with **no `.env` present**.
- ✅ Images run as non-root; entrypoints enforce their required env vars; `/app/config` is empty (mount-only), nothing baked.
- ✅ `make check` (merge gate) passes; SonarCloud security findings resolved at source.

## Follow-up (separate private repo)

The D13-specific deployment (`.d13.*.yaml`, `Jenkinsfile`, the Gradion instance config, VPN whitelist, Vault wiring, runbook) moves to a private `margince-d13-deploy` repo that consumes this one as a git submodule. Not part of this PR.

## Summary by CodeRabbit

* **New Features**
  * Added containerized deployment options via new Dockerfiles for the API, web UI, and background worker, including build-time OpenAPI generation and static frontend bundling.
  * Added container entrypoint scripts for the API (runs migrations) and worker (runs without migrations).
  * Added one-time PostgreSQL bootstrap SQL with separate non-superuser roles and required extensions.
  * Added SPA hosting (gzip, immutable asset caching, and client-side routing) plus deployment documentation.
  * Added a `.dockerignore` to streamline Docker build context.
* **Security**
  * Runtime containers run as non-root users.
  * Added secure admin password handling via protected secret files during API startup.


dungbanguyen:
Addressed all review findings in the latest commit (`fix(deploy): address review findings on the deployment materials`):

- **db-bootstrap.sql (cubic P1 / CodeRabbit Critical — real bug):** `:'app_pw'` isn't interpolated inside a dollar-quoted `DO $$…$$` body, so the bootstrap aborted. Rebuilt with `SELECT format('CREATE ROLE … %L', :'app_pw') WHERE NOT EXISTS(…) \gexec`. **Verified against a live Postgres:** creates a role with a special-char password (`%L` escaping), non-superuser + non-`BYPASSRLS`, idempotent. Also corrected the docs to pass passwords **raw**, not pre-quoted.
- **api/worker entrypoints (CodeRabbit Major — real bug):** sourcing `/.env` set shell vars without exporting them, so the `exec`'d binary never saw them. Wrapped the source in `set -a`/`set +a` (proven a child process now inherits the var).
- **docs/deployment.md:** split the api vs worker entrypoint rows (only the api needs the owner DSN); stated RLS precisely (superusers **and** `BYPASSRLS` roles bypass, so both runtime roles must be neither); quoted `location.origin + "/v1"`.
- **nginx.conf:** dropped the D13-specific ingress wording (target-agnostic), added `gzip_vary on`, and gave `index.html` an explicit `Cache-Control: no-cache`.

Thanks for the catches — the two psql/dotenv bugs would have broken a real deploy. Resolving the threads.

dungbanguyen:
Second review pass addressed:

- **db-bootstrap.sql — existing role could keep SUPERUSER/BYPASSRLS (P1):** the `NOT EXISTS` guard skipped a pre-existing role. Added unconditional `ALTER ROLE … NOSUPERUSER NOBYPASSRLS` for both roles (verified against Postgres). This makes the RLS-binding posture explicit even when a role already exists.
- **api-entrypoint.sh — password_file mismatch (P1):** the write path is now `MARGINCE_ADMIN_PASSWORD_FILE` (default `/app/secrets/admin-password`) and mkdir's its directory; docs now state the mounted `margince.yaml`'s `password_file` must match, so image/docs/config agree.
- **docs bootstrap example (P1):** pass passwords **raw**, not pre-quoted (matches the SQL's own guidance).
- **.dockerignore (P2):** also exclude `.pnpm-store/` and `scratchpad/`.
- **Pin base images to `@sha256:` (P2):** intentionally **not** doing this here. These app Dockerfiles use tag refs by house convention (matching the sibling `gradion-workspace` deployment Dockerfiles), and the repo's supply-chain pin gate (`check-image-pins`, `make check-image-pins`) deliberately scopes to CI workflow `uses:` + compose `image:` refs, not these build Dockerfiles. Pinning here would diverge from the established pattern and add digest-bump maintenance without a gate requiring it. Resolving as won't-fix with that rationale.

PR #328, opened 2026-07-30
docs(status): close the company-view arc, move its narrative to the archive

LarsGradion:
The company-view rebuild ended at six PRs — #309, #313, #315, #317, #319, #322 (plus #326 correcting its contract). A seventh was expected and does not exist, so STATUS.md's "PR 7 is the last one" pickup pointed at work nobody had specified. A dead pointer is worse than no pointer.

STATUS.md's own header promises it carries open work only, so the arc's narrative moves to STATUS-ARCHIVE.md: what the six PRs shipped, the three decisions the original scope line left open, and the four rules the connections card's review rounds produced. Each rule is a trap the first version fell into, which is why they read as rules rather than as a change log:

- **gate a group inside its own read, never from the omitted set** — the graph's person reads were gated by the order the group list ran in, so reordering a slice literal would have turned a gated read into an ungated one;
- **a cap must count what the cap MEANS** — `graphOrgCap` is ten companies, but the read bounded itself by rows, and one company attached many ways filled the budget and starved the others;
- **every group total rides the same statement as its rows** — two Read Committed snapshots can disagree and drive `dropped_count` negative against the schema's own `minimum: 0`;
- **a gap documented where the reader cannot see it is not documented** — one commit corrected the Go comment and left the published contract promising the opposite.

STATUS.md keeps a four-line pointer plus the open questions the arc deliberately left, which are genuinely still open: which deal edits count as working a stalled deal, the O(N) suggestion read, the uncapped `/ask` model call, and the `org_ask` corpus gap.

Docs only. `make check` green.

---
## Summary by cubic
Moves the company-view docs arc from `STATUS.md` to `STATUS-ARCHIVE.md` and removes the dead pointer to a non-existent “PR 7”. `STATUS.md` now only lists open work and links to the archive.

- **Refactors**
  - `STATUS.md`: trimmed to a short pointer; keeps the open questions (stalled-deal edits, O(N) suggestion read, uncapped `/ask` call, `org_ask` corpus gap).
  - `STATUS-ARCHIVE.md`: documents the six feature PRs (#309, #313, #315, #317, #319, #322) and clearly separates them from the follow-up #326; captures the three decisions and four review rules from the connections card reviews; notes that graph level 2 is deferred and does not replace `RecordContextPanel`.

<sup>Written for commit a3ce1d6fe1c8ecc64f31fe6f0523230b273e78d7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/328?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Documentation**
  - Updated project status documentation to mark the company-view rebuild as complete.
  - Added an archive entry summarizing the rebuild’s delivered behavior and key correctness rules.
  - Clarified that advanced graph behavior remains deferred and does not replace other company-view panels.
  - Documented remaining open questions and known scope gaps for future work.


LarsGradion:
First finding applied in a3ce1d6f; second one I am declining, with the check.

**Six feature PRs vs #326 — valid.** The heading listed seven identifiers and then said a seventh did not exist, which contradicts itself on its face. It now names the six features and #326 separately, and says what #326 is: the contract correction that missed #322's squash by one push, not a seventh feature.

**The `stalledDeal.episode()` pointer — not valid.** The rejected shapes are in that function's own doc comment, which is what STATUS.md says. `backend/internal/compose/org360/suggestionreads.go` lines 124-159 carry both rejections and the reasoning for each: `deal.version` ("would catch all of it and is monotone, but it also bumps on writes no person made: CloseDateCorrector patches expected_close_date from a sweep") and `wait_until` ("a deferral can be set, expire, and be cleared, returning the deal to a shape the rep already dismissed").

`suggestions_test.go` contains no discussion of rejected shapes at all — `grep -n 'rejected\|Rejected'` returns nothing in that file, and the two matches in the package are both in the `episode()` comment. Repointing the reader at the test would send them somewhere the rationale is not.

PR #338, opened 2026-07-31
fix(overlay): stop a resumed flip creating records the crash already landed

### LINKED ISSUE #336: overlay flip: hardening deferred from PR #320 review
Findings from the bot review of #320 that were assessed as real but deliberately not fixed in that PR. Each is recorded here rather than rationalized in a source comment.

### 1. Native create and identity recording are not atomic (highest value)
`flipWriters.ensureX` creates the native record through the module store (its own transaction), then calls `remember` → `RecordIdentity` in a second one. `Exists` resolves only through the engine-owned identity map — deliberately, so a client-planted `source_system`/`source_id` cannot suppress a real record. The consequence is that a crash between the two transactions leaves a created record the resume cannot see, and the row is created a second time.

Window is narrow, but the outcome is a duplicated record in a one-way cutover. A real fix needs both writes in one transaction, which crosses the module boundary — ADR-0054 §9 single-tx exception territory.

### 2. `import_run.error` stores raw driver text
`failRun` persists `cause.Error()`, whose tail can carry pgx text naming tables and columns. No transport reads `Run.Error` today (the backfill surface transports a classified `ErrorClass` from a different table), so there is no client exposure now — but the column is a leak waiting for its first endpoint. The `ErrorClass` precedent is the shape to follow.

### 3. Reconstruction does not verify the bundle is a whole-estate export
`parseBundle` accepts any bundle carrying `canonical_data_resides_in`. A bundle produced under a narrower actor would rebuild a silently incomplete estate. Mitigated today because `/overlay/export` is admin/ops-gated and those roles are row-scope unbounded, so in practice every bundle is whole-estate — mitigated by RBAC, not by validation. The manifest should record scope and the parser should require it.

### 4. Seal/fenced-write ordering
A fenced mirror write can pass its freeze check before `SealFlipSnapshot` commits and land after it, so the sealed snapshot differs from the one readiness was measured against. The import reads the mirror after sealing, so the late row is still imported — the inaccuracy is in the parity preview's counts, not in what lands. Fixing properly means serializing fenced writers and the validate/seal transition on the sync-state row.

### 5. `pgtype.Numeric` renders as a struct in CSV export
`csvCell`'s fallback serializes pgx's internal representation rather than a decimal, so numeric columns such as `deal.fx_rate_to_base` are corrupt in a CSV member. Pre-existing behaviour, relocated but not introduced by #320.

dungbanguyen:
Closes the first and highest-value item of #336, found in the review of #320.

## The defect

A record lands in two steps that are **not** one transaction: the owning module store writes the native row under its own write shape (domain row + audit + outbox), then the engine's identity map records what that external id became. `Exists` reads the identity map, so a process that died between the two steps left a record nothing could recognize — and the resumed run created it again. In a one-way cutover that is a duplicate nobody asked for and nothing removes.

## Why not just use one transaction

That was the obvious fix and it is the wrong shape here: it means a module store accepting an externally-owned transaction, across three modules, which is exactly the boundary ADR-0054 draws. The alternative makes the crash *repairable* instead, using an invariant the cutover already relies on.

## The approach

The flip previously stamped `source` with a value a client could also write — which is precisely why the identity map had to be the sole authority on "already imported" (a caller could otherwise pre-plant a row under a guessed incumbent id and have the flip treat the real record as already handled).

It now stamps `source` inside the **reserved import namespace**, and every create wire that accepts provenance refuses that namespace. `person`, `organization` and `deal` join the `lead` and `activity` paths that already guarded it — the last three matter most here, because they have no `(source_system, source_id)` replay key and `source` is the only provenance they carry.

With that closed, a native row bearing the prefix can only be this importer's own work. So a resumed run opens by adopting rows that carry its prefix and are not yet mapped. It runs **only** on resume — a first attempt has nothing to adopt and should not pay five table scans to discover that.

## Testing

`TestAResumeAdoptsRecordsTheCrashLandedButNeverMapped` reproduces the exact process death (record landed, identity write lost) and asserts the resumed run writes it once. Verified to fail against the unfixed code — without the repair it reports `person/p-1 was written 2 times`.

The engine's fake writer now models the two steps as separate state. With a single map it could not express "landed but never recorded", so the defect would have passed the suite untouched — which is how it survived eleven review rounds on #320.

`TestEveryProvenanceWireRefusesTheImporterNamespace` covers all five wires as one rule rather than per-call-site, and the deals module gets its own guard and test.

Gates green: `make check`, and `make test-integration` at 0 skips across 29 packages (at `INTEGRATION_JOBS=2`; the machine hits `out of shared memory` in unrelated fixture setup at 4).

---
## Summary by cubic
Stops a resumed flip from creating duplicates by adopting previously landed, unmapped rows marked with a reserved provenance prefix; the repair runs at the start of every attempt (even with a zero checkpoint), only adopts live rows, and finishes closing adopted deals when needed. Closes #336 items 1 and 8.

- **Bug Fixes**
  - Added early `ReconcileIdentities` that runs unconditionally, adopts rows stamped with `mirror:` and missing from the identity map, skips archived/merged rows, rejects ambiguous external IDs, uses `starts_with`, extracts the external ID once, and enforces a class allowlist.
  - Adopted deals are closed only if the incumbent is terminal and the native status is still open; the check now uses `deals.DealStatus` (type-safe) and remains idempotent on replay.
  - Flip stamps `source` with the reserved `mirror:` prefix; all create wires refuse that namespace on `source` and `source_system` with 422 field errors, while ordinary provenance values remain allowed; integration queries updated to `mirror:`.

- **Refactors**
  - Centralized reserved-namespace enforcement in `provenance` (`Refuse`, `ReservedError`); added bulk `RecordIdentities`; updated the `Writers` interface and engine to invoke reconciliation once per attempt.
  - Tests cover zero-checkpoint runs, the live-row rule, prefix strictness, the class allowlist, the deal-close rule, and adoption/finishing of half-landed deals.

<sup>Written for commit 7cfe20150636f2d22d5661fd1c32909454273a31. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/338?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added native deal importing with names, amounts, close dates, owners, currencies, and deal stages.
  - Imported closed deals now retain their closed or lost status and relevant details.
  - Interrupted imports can resume without duplicating records or losing identity mappings.
  - Existing deals are safely adopted and finalized after interrupted imports.

- **Bug Fixes**
  - Prevented client-provided values from using the reserved importer provenance namespace.
  - Improved recovery of deals whose closing process was interrupted.

### DIFF
--- a/backend/internal/compose/flip_test.go
+++ b/backend/internal/compose/flip_test.go
@@ -215,3 +215,28 @@ func TestStageCatalogPlacement(t *testing.T) {
 		}
 	})
 }
+
+func TestAdoptedDealClosesOnlyWhenItIsStillOpen(t *testing.T) {
+	closed := flipPlacement{closedStage: &[]ids.StageID{ids.New[ids.StageKind]()}[0], closedSemantic: stageSemanticWon}
+	open := flipPlacement{}
+
+	// The case the repair exists for: the crash created the deal on its
+	// open birth stage and died before the close. Left alone it would be
+	// reported converged while sitting open, and the estate's won revenue
+	// would simply be missing.
+	if !adoptedDealNeedsClosing(closed, "open") {
+		t.Error("a deal the incumbent calls closed, still open natively, must be closed on the resumed attempt")
+	}
+	// The ordinary replay: the attempt that created it also closed it.
+	// Advancing again would refight a settled close and its FX freeze.
+	for _, status := range []string{"won", "lost"} {
+		if adoptedDealNeedsClosing(closed, status) {
+			t.Errorf("a deal already %s was advanced again", status)
+		}
+	}
+	// The incumbent says open, so open is correct and there is nothing
+	// to assert — closing here would invent a terminal state.
+	if adoptedDealNeedsClosing(open, "open") {
+		t.Error("an open incumbent deal must not be closed")
+	}
+}
--- a/backend/internal/compose/flip_test.go
+++ b/backend/internal/compose/flip_test.go
@@ -16,6 +16,7 @@ import (
 	"time"
 
 	crmcontracts "github.com/gradionhq/margince/backend/internal/contracts"
+	"github.com/gradionhq/margince/backend/internal/modules/deals"
 	"github.com/gradionhq/margince/backend/internal/shared/apperrors"
 	"github.com/gradionhq/margince/backend/internal/shared/kernel/ids"
 )
@@ -224,19 +225,19 @@ func TestAdoptedDealClosesOnlyWhenItIsStillOpen(t *testing.T) {
 	// open birth stage and died before the close. Left alone it would be
 	// reported converged while sitting open, and the estate's won revenue
 	// would simply be missing.
-	if !adoptedDealNeedsClosing(closed, "open") {
+	if !adoptedDealNeedsClosing(closed, deals.DealOpen) {
 		t.Error("a deal the incumbent calls closed, still open natively, must be closed on the resumed attempt")
 	}
 	// The ordinary replay: the attempt that created it also closed it.
 	// Advancing again would refight a settled close and its FX freeze.
-	for _, status := range []string{"won", "lost"} {
+	for _, status := range []deals.DealStatus{deals.DealWon, deals.DealLost} {
 		if adoptedDealNeedsClosing(closed, status) {
 			t.Errorf("a deal already %s was advanced again", status)
 		}
 	}
 	// The incumbent says open, so open is correct and there is nothing
 	// to assert — closing here would invent a terminal state.
-	if adoptedDealNeedsClosing(open, "open") {
+	if adoptedDealNeedsClosing(open, deals.DealOpen) {
 		t.Error("an open incumbent deal must not be closed")
 	}
 }
--- a/backend/internal/compose/flipdeals.go
+++ b/backend/internal/compose/flipdeals.go
@@ -1,0 +1,134 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+package compose
+
+// Landing an incumbent deal natively. It is the only estate class whose
+// arrival takes two steps — born on an open stage, because the store
+// forbids any other birth, then advanced to its terminal one — and that
+// second step is what makes it the only class the crash repair has to
+// finish rather than merely recognize.
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/gradionhq/margince/backend/internal/modules/deals"
+	"github.com/gradionhq/margince/backend/internal/modules/migration"
+	"github.com/gradionhq/margince/backend/internal/platform/database/storekit"
+	"github.com/gradionhq/margince/backend/internal/shared/kernel/ids"
+)
+
+func (w *flipWriters) ensureDeal(ctx context.Context, row migration.Row) (migration.EnsureResult, error) {
+	owner, disclosure, err := w.resolveOwner(ctx, row, flipObjectDeal)
+	if err != nil {
+		return migration.EnsureResult{}, err
+	}
+	stages, err := w.stageCatalog(ctx)
+	if err != nil {
+		return migration.EnsureResult{}, err
+	}
+	rawStage := fieldString(row.Fields, "stage_id")
+	placement := stages.place(rawStage)
+
+	name := strings.TrimSpace(fieldString(row.Fields, "name"))
+	if name == "" {
+		name = overlayUnnamed
+	}
+	in := deals.CreateDealInput{
+		Name:       name,
+		Currency:   fieldStringPtr(row.Fields, "currency"),
+		PipelineID: placement.pipeline,
+		StageID:    placement.birthStage,
+		OwnerID:    owner,
+		Source:     w.provenance(flipObjectDeal, row.ExternalID),
+	}
+	if minor, ok := fieldInt64(row.Fields, "amount_minor"); ok {
+		in.AmountMinor = &minor
+	}
+	if closeAt, ok := overlayTime(row.Fields, "expected_close_date"); ok {
+		in.ExpectedClose = &closeAt
+	}
+	deal, err := w.deals.CreateDeal(ctx, in)
+	if err != nil {
+		return migration.EnsureResult{}, fmt.Errorf("flip import: creating deal %s: %w", row.ExternalID, err)
+	}
+	dealID := ids.From[ids.DealKind](ids.UUID(deal.Id))
+	if err := w.remember(ctx, flipObjectDeal, row.ExternalID, ids.UUID(deal.Id)); err != nil {
+		return migration.EnsureResult{}, err
+	}
+
+	// A closed estate deal is born open (the store's open-birth-stage
+	// rule), then advanced to the terminal stage — the same won/lost path
+	// a native close takes, FX freeze included.
+	if placement.closedStage != nil {
+		var lostReason *string
+		if placement.closedSemantic == "lost" {
+			reason := "imported closed-lost from the incumbent estate"
+			lostReason = &reason
+		}
+		if _, err := w.deals.AdvanceDeal(ctx, dealID, deals.AdvanceDealInput{ToStageID: *placement.closedStage, LostReason: lostReason}); err != nil {
+			return migration.EnsureResult{}, fmt.Errorf("flip import: closing imported deal %s: %w", row.ExternalID, err)
+		}
+	}
+	notes := stageDisclosure(placement, rawStage, row.ExternalID)
+	if disclosure != "" {
+		if notes != "" {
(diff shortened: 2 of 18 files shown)


dungbanguyen:
Status: everything green except the SonarCloud gate, which now fails on one condition only.

- **Duplication is fixed** (was 4.7%, threshold 3%). Adding the same refusal to three module handlers extended an error-mapping ladder Sonar already flagged. The rule is one invariant with one wire meaning, so it now lives once in the kernel and maps once in `httperr` — the three module ladders lost the entry entirely.
- **The integration shard failure was flaky** — `TestSubscriberDeliversAcksAndFiltersWorkspaces` in `platform/events`, unrelated to this change, green on re-run.
- **`new_coverage` is 37.6% against an 80% threshold**, and this one is a genuine gap rather than a metric artifact. The repair's SQL scan (`flipreconcile.go`) runs *only* on resume, and nothing exercises it against real Postgres. The engine's fake writer proves the engine calls the repair; it cannot prove the scan finds anything.

I tried to close it by flipping, forgetting the person identities, and re-calling `/v1/overlay/flip` — that does not work, and the reason is worth recording: a completed flip clears `x_sor_mode`, so the endpoint 404s. The crash this repairs is one where the flip never completed, which the HTTP path gives no way to induce.

Closing it properly needs a test hook into the writers, in the shape this package already uses for `ClaimFlipForTest` and `ReconstructForTest`. That is a small addition but it is a design choice about test surface, so I have left it rather than improvising it at the end of a long session. **This PR should not merge until that coverage exists** — a resume path with no real-database test is exactly the gap that produced the defect it fixes.

dungbanguyen:
End-of-work review round. Both reviewers independently found the same two critical defects **in the repair itself**, and one reproduced the first against the real engine. Fixed in `40e6c3df`.

### The repair skipped the case it exists for
The checkpoint advances only *after* a row lands, so a crash on the **first** row leaves it at zero — and a re-created run, which is every bundle re-upload and every re-sealed snapshot, also starts at zero with the previous attempt's orphans still on disk. `run.Checkpoint > 0` skipped the repair in both. The bug this PR fixes was still fully live for the first record of every attempt.

It now runs unconditionally. The conditional was the added complexity rather than the saving, and my comment justifying it ("a first attempt has nothing to adopt") was simply false.

### The scan adopted dead rows — exploitable
Merge preserves the loser's provenance and sets only `archived_at`/`merged_into_id`. So given the exact state this PR repairs — a crash orphan carrying `mirror:hubspot:person:p-1` with no map row — anyone who could see that row could merge it into their own record, or just delete it. The next attempt would bind the incumbent's external id to the tombstone; the genuine contact would **never be created**, reported as `Unchanged` rather than skipped, and every activity resolving through that identity would follow the tombstone to its survivor. That is the escalation the whole design set out to prevent, reintroduced by the fix.

Adoption is now restricted to live rows, so a dead one is left alone and the import recreates the record properly.

### Also fixed
`starts_with` rather than a LIKE pattern (the prefix carries the incumbent's name, and `_`/`%` in it would widen the match across incumbents while the length-based strip took the wrong characters); the external id extracted once in SQL instead of the same strip spelled twice in two languages that agreed only by accident; an external id claimed by two live rows now binds to neither, rather than letting the insert and the writer's cache disagree about which won; the import grant checked before the scan reads five module tables rather than after, with the deliberate absence of a row-scope clause stated; and `lookup`'s comment, which claimed provenance is never read back — false sixty lines from the code that reads it back.

`TestAFirstAttemptDoesNotReconcile` was pinning the defect as correct behaviour and is replaced by one that fails at the `failAt: 1` boundary the old suite stepped around.

### Not fixed, recorded instead
- **An adopted deal never reaches `AdvanceDeal`.** `ensureDeal` is create → remember → advance, and adoption short-circuits at `Unchanged`, so a closed-won estate deal can land parked in its open birth stage with no disclosure. Real, and adoption now happens more often, so it matters more. Needs the close folded into the create or terminal placement re-asserted for adopted deals — a design change, not a patch.
- **Adopted leads/activities lose their deliberate non-adoption disclosure**, whose comment is now the stale sibling of this change.
- **The namespace guard is seven hand-placed calls**, not a derived obligation; repo rule 2 wants a fitness function walking the contract for request types with a `source` field whose table is in `flipImportOrder`. Both reviewers confirmed the invariant holds today by exhaustive trace — including that `promote.go` copies `lead.Source` onto a person but survives only because the class segment differs.

I'll add these to #336.

**Coverage:** the reconcile now runs on every flip rather than only on resume, so its SQL is exercised by the existing integration lanes — which also addresses the Sonar new-code coverage gap I flagged earlier. Both gates green (`make check`; integration 0 skips, 29 packages).

dungbanguyen:
Rebased onto current main and closed **#336 item 8** as well. Force-pushed as `105fa33c`.

**Item 8 — an adopted deal never reached its close.** A deal is the only estate class whose arrival takes two steps: born on an open stage, because the store forbids any other birth, then advanced to its terminal one. An orphan recovered by the repair completed only the first, so `Ensure` answered `Unchanged` and a closed-won estate deal stayed parked open, counted as converged — wrong revenue after a one-way cutover, with no disclosure. `Ensure` now re-asserts the close for an adopted deal. It is idempotent by construction: `adoptedDealNeedsClosing` returns true only when the incumbent calls the deal terminal *and* it is still open natively, so the ordinary replay finds it already closed and does nothing rather than refighting a settled close and its FX freeze.

**The rebase changed the approach in one place, for the better.** Main had independently solved the same error-mapping duplication I did, with `apperrors.FieldFault` — a module's error states its own field/code/message and `httperr` routes it generically. That is better than my central special-case, because it also answers on the MCP surface, which never runs a module's HTTP mapper. So the shared `provenance.ReservedError` now implements `FieldFault()` and my `httperr` change is gone. Net effect against main: one kernel type replaces the two per-module `ReservedSourceSystemError` copies, and it composes with main's mechanism rather than competing with it.

Deal handling also moved to its own file — it was pushing `flipwriters.go` past the length cap, and the two-step landing is exactly the concept that earns one.

Gates: `make check` green, integration 0 skips across 28 packages.

Still watching the Sonar coverage condition — the reconcile now runs on every attempt rather than only on resume, so its SQL is exercised by the existing lanes, and `adoptedDealNeedsClosing` is unit-covered. I'll report where the number lands rather than predict it.

dungbanguyen:
Six threads, triaged in `7cfe2015`.

**Fixed:**
- **`deal.Status` compared as a bare string** (Major, correct). Converting to `string` and testing against `"open"` meant a status rename would silently start treating closed deals as open and refighting settled closes with their FX freeze. Now `deals.DealStatus` against `deals.DealOpen`, so that stops compiling instead. The unit test moved to the typed constants too.
- **The activity guard's ordinary-`source` case.** The rule is a prefix, not a ban on the field, and nothing pinned that a connector's own `source` still passes the wire.

**Stale — resolved by the rebase, not by an edit:**
- The `httperr_test.go` substring assertion. Main had independently solved this duplication with `apperrors.FieldFault`, which is a better mechanism than my central special-case because it answers on the MCP surface too, so I adopted main's and dropped mine. That test no longer exists in this diff; the equivalent assertion now lives with the type, in `provenance`, checking the `FieldFault()` triple directly rather than substring-matching a rendered body.
- The reserved-`source` coverage for `activityLogInput` — already added before this round; only the ordinary-value case was genuinely missing.

**Deferred, with reasoning:**
- **Prefix indexes for the reconcile scan.** Correct that `starts_with` over five `source` columns is a sequential scan. Deliberately not adding them here: a flip happens once per workspace, ever, and it already reads the whole mirror estate — five scans are not what makes it slow. Five index migrations on shipped tables to speed up a once-in-a-lifetime operation is the wrong trade, and I would rather that decision be made with a measurement than with my guess. Recorded on #336.

**One I want to answer directly** — cubic's point that the new integration test seeds the row then runs a *fresh* flip rather than a resumed one, contradicting the PR's design note that a first attempt has nothing to adopt.

That note was true of the design as first written, and the review of the previous round is exactly what killed it. Gating on `run.Checkpoint > 0` skipped the repair for the first record of any attempt (the cursor advances only *after* a row lands) and for every re-created run, which is every bundle re-upload. So the repair now runs unconditionally, and a first attempt adopting orphans is the intended behaviour rather than a gap in the test. The stale claim was in the PR text, not the code — the comment asserting it was deleted in `40e6c3df`. The resume path itself is covered by the engine's unit lane, which drives a crash at `failAt: 1` and asserts the record is written once.

All checks green including the Sonar coverage condition, which the new lane closed.

PR #343, opened 2026-07-31
feat(company): make the facts read off the site readable

LarsGradion:
Stacked on #342 — review that first; GitHub retargets this to `main` when it merges.

## Why

One real account (ScaleCommerce) returns **99 facts**. The card drew every one, unranked, at full weight:

- "Shop Devs" and "Shop-Devs" as two separate entries
- the same platform listed under `product`, `service` **and** `capability`
- a card taller than the page beside it

At that length nobody reads any of it, and the visible duplicates make the reader doubt the rest.

## What changed

**Grouping by canonical value** (`factview.ts`) collapses the spellings: case, punctuation and spacing carry no meaning in a scraped value. Diacritics are deliberately **kept** — in German they distinguish real words, and folding them would merge facts that differ.

Facts collapse **within a category, not across the set**. A company can genuinely be both a `partner` and a `named_customer`, and the category is what keeps those two true statements apart. Inside `offering` the collapse spans fields, because product/service/capability are three ways of saying one thing; the most concrete spelling survives, and a human-held value outranks any machine one whatever the model scored it.

**Only the first five rows per category are drawn**, with the total on the button — a truncated list with no number reads as "that is everything".

Result on the real account: **20 rows instead of 99**, the rest one click away.

| Category | Facts | Shown |
|---|---|---|
| Company | 5 | 5 |
| Offering | 50 | 5 + "Show all 50" |
| Market | 13 | 5 + "Show all 13" |
| Signals | 31 | 5 + "Show all 31" |

## The untranslated-enum bug

Every fact field now has a name in both catalogs. The label map they were reaching for (`COLD_FIELD_LABELS`) covers the **profile-field** vocabulary, not this one, so all seventeen fact fields fell through its fallback and reached German readers as English snake_case: *"served industry"*, *"company size"*. Verified in the browser: the rail now reads Kontakt-E-Mail / Standort / Fähigkeit / Größe / Zertifizierung.

A test derives the field list from the schema, so a new fact field fails there rather than on the page.

## Relationship to the backend work

The server is being taught the same canonical form (planned). Once it is, this stays correct and simply has nothing left to collapse — it is not a competing source of truth, it is how the card decides what the reader sees one of and in what order.

## Tests

10 new unit tests over `canonical` and `groupFacts`, including the two cases that must NOT collapse (partner vs named_customer; diacritics). `make check-fe` green.

Note: the UAT lane covers the companies **list**, not the detail page — no e2e touches this screen yet. Closing that gap is a later slice.

---
## Summary by cubic
Make company facts readable by collapsing duplicates and showing a short preview per category. On a real account, this reduces 99 rows to 20.

- **New Features**
  - Group facts by canonical value within each category; keep diacritics.
  - In offering, merge product/service/capability; prefer the most concrete field and human-sourced values.
  - Show only the first 5 rows per category with a "Show all {count}" toggle; keep a consistent category order and omit empty categories.
  - Tighten spacing in the facts card for faster scanning.

- **Bug Fixes**
  - Localize all fact fields in `en` and `de`; replaced the wrong label map that caused snake_case in German.
  - Add unit tests for canonicalization, grouping, and label keys, including cases that must not collapse.

<sup>Written for commit 0354427a66d7e11630ac83d210a94af823d835bf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/343?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


LarsGradion:
Folded into one integration branch so the whole company-page rework can be tested as a single thing before any of it merges. Superseded by the branch on #342.

PR #358, opened 2026-08-01
feat(linkedin+network): the review queue, the reach view, and the three risk rules that were named but never fired

LarsGradion:
## What this does

Three things that were each half-built, finished and reconciled.

**The LinkedIn suggestions can be decided.** The import card counted "awaiting your confirmation" for a queue that existed nowhere — no list endpoint, no confirm, no reject, no screen. The only route to a confirmed match was the automatic exact-email one, which on a real 5,064-row export fired exactly once, because LinkedIn exports an address only for connections who allowed it. The matcher's middle tier, where all the volume is, was inert.

Four owner-scoped endpoints and a review queue in Settings → Integrations. Confirming writes the connection's own LinkedIn URL onto the contact — the one thing a ghost ever contributes to a real record. Migration 0164 adds the column the importer needed: `Connections.csv` has carried a `URL` column in every format LinkedIn has shipped and this importer read every other one.

**The reach view** answers what the import was for: which accounts your network already reaches, and how many of those people are not contacts yet. It is also the half that works on a one-person workspace, where asking which *colleague* knows an account has no content.

**Three risk rules that were `Kind` constants with no detector.** `going_cold`, `champion_left` and `stakeholder_left`. A surface listing the kinds it can show tells a rep those checks are running, so declaring them without implementing them is worse than their being absent. Plus `who_knows` and `network_risks` in `AssembleContext`, and the `intro_path_to` / `at_risk_relationships` tools — before this the projection was visible on the person page and invisible to the assistant answering "who should introduce me".

## Review

Codex reviewed all 17 commits with security weighted heavily. 17 findings: **13 fixed here, 4 pushed back with reasons in STATUS.md.** The ones worth naming:

- **Capture privacy.** The re-match sweep runs under a system principal, which is exempt from owner-private by design, so `ScopeClauseFor` returned an empty clause and the matcher linked one member's ghost to a contact another member had captured privately. The review list then hid the name and returned the uuid — an id alone proves the record exists. The boundary is now carried on the match itself, keyed to the ghost's owner.
- **A reversed human decision.** The duplicate collapse ranked the matcher's own `suggested` equal to a person's `confirmed`, so the older row won on age and then inherited `confirmed` from the copy it had overwritten.
- **The write shape.** A rejection emitted nothing; a confirmation's `person.updated` cited an audit row for the LinkedIn connection.
- **GDPR.** `profile_url` reached neither the Art. 15 package nor the Art. 17 ghost sweep.

## Verification

`make check`, `make test-integration` (0 skips), `make frontend-e2e` (61 passed) and the craft gate all exit 0. Regression tests cover the two most serious findings against a real database.

## Note for whoever merges next

`origin/main` moved during this work (#345) and is merged in. Migration **0160 is claimed by two unmerged branches** with different content — this one and `capture-domain-triage`. Whichever merges second must renumber; details in STATUS.md.

---
## Summary by cubic
Adds LinkedIn review and reach with event-driven matching and an hourly per-owner re-match; also ships deal coverage risk rules and UI so reps and the assistant can see who knows whom and what’s at risk. Closes the remaining privacy and webhook gaps and hardens confirm, contracts, and tool behavior.

- **Bug Fixes**
  - Per-person match now runs under each ghost owner’s authority; organization placement uses the same owner scope to prevent reach leaks.
  - Webhooks: classify `linkedin_match.decided`, `linkedin_account.changed`, and `linkedin_network.imported` as self-only to avoid workspace-wide fan-out.
  - Confirm path: keep `person_id` on chunked requests so corrections apply to the intended contact.
  - Handle removal: audited and emits `person.updated`, symmetric with adds.
  - Tools: `intro_path_to` and `at_risk_relationships` are native-only; overlay mode refuses instead of returning misleading “nobody” answers.
  - Contracts: add `user` to `ContextEntityRefType` so who-knows context responses are consumable.
  - GDPR: include LinkedIn `profile_url` in SAR and erasure; read handles before deletion during ghost sweeps.
  - Tests: grant real read-people authority to ghost owners so per-owner matching and privacy checks are exercised.

<sup>Written for commit 3cd4a158a8dbbfe3f920427a2417d3200f31a343. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/358?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added LinkedIn onboarding, account management, connection import, match review, confirmation/rejection, and network reach reporting.
  * Added colleague relationship visibility and introduction-path suggestions.
  * Added deal coverage risks for departed stakeholders and deals going cold, including days since last touch.
  * Added LinkedIn review and reach panels, plus network and coverage cards across relevant views.

* **Bug Fixes**
  * Improved privacy controls, matching accuracy, duplicate handling, data erasure, and onboarding completion for empty results.


LarsGradion:
Reconciled the cubic review. Fixed in `bd385ad4`:

- **Four CSS custom properties I invented do not exist** (`--radius-2`, `--surfaceRaised`, `--surfaceSunken`, `--focusRing`). Every declaration using them was invalid at computed-value time, so three new cards rendered transparent and square with no focus ring. The design gate checks that a colour *reads* a token, not that the token is *defined* — worth knowing.
- **The upload focus ring could never match** — adjacent-sibling combinator pointing at a label that precedes the input. Now `:has()` on the wrapper.
- **A connective is not a legal form.** `"&"/"und"/"and"` in `legalSuffixes` collapsed `"Research and"` → `research`. That key drives *all* organization dedupe. Now consumed only between two legal forms, with a test.
- **`person.read` missing** on the review list and `intro_path_to`.
- The `none` band no longer renders `0 interactions`; the suggestion queue discloses when more are waiting; German onboarding copy uses `du` like every sibling step.

Not acting on these now, with reasons:

- **Read Committed snapshots** (`coveragefacts.go:52`, `linkedinreach.go:86`) — real, and the same property every multi-statement read in this codebase has. Changing isolation for these two is a repo-wide decision, not a PR-local one.
- **Reach pagination** (`crm.yaml:8260`) — the response reports `accounts_total` and `unresolved_connections`, so a truncated list already says what it is not showing. A cursor is warranted when someone needs page 2; nobody does yet.
- **Person-merge ghosts, URL canonicalization, org-placement scoping, the partial-index suggestion, `orgMatchKeys` stopword floor** — all real and all follow-ups; recorded in STATUS as carry-forward rather than widened into this PR.
- **`writeLinkedInHandle` lock ordering** — `touchPerson` takes `storekit.LockRow` on both paths, which is one order; the deadlock described needs a second order I do not see.
- **Two weak tests** (owner-scoping asserted only from the caller's side) — correct criticism, and the boundary is covered by `TestTheSweepNeverMatchesOutsideTheGhostOwnersRowScope` and `TestAConnectionNeverMatchesAContactItsOwnerMayNotSee` against a real database.

LarsGradion:
Merged `origin/main` (#360). All local gates green: `make check`, `make test-integration` (0 skips), `make frontend-e2e`, craft static — all exit 0.

Closing the review loop here. The P0 (the per-person match ran under the unbounded system actor, same oracle as the workspace sweep one path over) and every P1 from the last round are fixed in `308d10c9`. The remaining findings are recorded as carry-forward in STATUS rather than widened into this branch: Read Committed snapshot consistency on two multi-statement reads, LinkedIn URL canonicalization for erasure suppression, person-merge ghost re-evaluation, org-placement cost scoping, reach pagination, and an onboarding restore path that can skip the LinkedIn act after a reload.

This is the fifth review round on one branch. None of the open items is a security hole, and each is the kind that earns its own PR with its own tests — holding this open through further rounds costs more than it buys, and two other sessions are queued behind it.

PR #371, opened 2026-08-03
feat(company): the company page tells the truth about an account (P0)

LarsGradion:
Phase 0 of the company-page overhaul: make the page trustworthy before rebuilding it. Spec side is merged — [ADR-0079](https://github.com/gradionhq/margince-foundation/pull/1226) and the [score amendment](https://github.com/gradionhq/margince-foundation/pull/1228).

## The audit that started it

The ScaleCommerce record holds an inbound email ending the contract on 31 July 2026 and a live opportunity with two of its people. The page said **Prospect**, showed **Relationship 2/100**, and claimed **Activity (30d): 0** above a timeline with three messages in it. Every one of those was a different bug.

## What changed

**`classification` is split in two.** It held one value and answered two questions — two of its nine values are stages of a sales motion, six are kinds of company. So `partner` erased whether they buy from us, and there was no churn state at all: an account that ended its contract could only be relabelled by overwriting `customer`. Now `organization.lifecycle` says where the account stands and `organization_relationship_type` says what the company is to us, multi-valued.

The sharpest finding: `classification` was `NOT NULL DEFAULT 'prospect'` and **never had a writer**. ADR-0032 said enrichment would set it; that was never built. So "Prospect" was the column's default rendered as a finding — which is also why this migration carries almost no stored meaning across.

**The partner invariant is enforced instead of described.** ADR-0032 said an org is a partner iff `classification='partner'` and a partner row exists; nothing checked either direction. Both are checked now, in one locked transaction.

**"Activity (30d)" counts what the timeline shows.** It matched only direct organization links while the timeline walks three; since capture files mail against the *person*, a busy account's correspondence never reached the count.

**The header says who wrote last.** `last_inbound_at` and `last_outbound_at`, kept apart — an account we mailed a fortnight ago with no reply and one that wrote this morning share a date and mean opposite things. The 0–100 score is gone: it was PO-F-3's MAX over contacts, so one talkative contact spoke for the whole account. The factors stay computed and shown.

**Enrichment writes show as the columns they changed.** Cold-start and deep-read audited with a metadata bag, so field history rendered pseudo-fields called "source" and "facts" — changes that never happened — while the `legal_name` just written appeared nowhere.

**Smaller ones:** the visit baseline is recorded (nothing ever called `view-ack`, so "first time opening" showed on every visit); Partner is no longer a permanent tab; Connections is gone from the rail; facts whose value contradicts their field say so.

## Review

Codex reviewed the phase before this went up and found two blockers, both since fixed: the invariant only bound one way, and the migration's own backfill created violating rows for old `agency`/`reseller` orgs. The full integration lane caught three more that per-package runs had missed.

## Open decision

`agency`/`reseller`/`tech_vendor`/`platform` map to no relationship type; a human retags them. Mapping them automatically was the original plan and is what created the violating rows, since none of them have a partner programme.

## Gates

`make check` green. `make test-integration` green — 28 packages, 0 skips.

---
## Summary by cubic
Make the company page honest. Split “classification” into clear fields, fix activity and header readings, and gate Partner and visit baselines so what you see matches reality.

- New Features
  - New fields: `organization.lifecycle` (single) and `organization_relationship_type` (multi), with migration 0169. Editable on the company form and shown as badges. API/list filters support `lifecycle` and `relationship_type`; `classification` stays one release for compatibility.
  - Header shows last_inbound_at and last_outbound_at instead of a 0–100 score; omitted when the caller lacks activity access.
  - Records a visit baseline after a 5s dwell to power “since your last visit”.
  - Facts that contradict their field show a small suspect badge.
  - UI: Partner tab only when the account is a partner; “Relationships” → “People & companies”; Connections card removed.

- Bug Fixes
  - “Activity (30d)” matches the timeline by walking activity, deal’s org, and contact’s employer; query budget 25 → 27 for the new last-touch section.
  - Field history for enrichment/deep reads shows real column diffs, not pseudo-fields.
  - Partner invariant enforced both ways: a partner must have the programme row and the `partner` relationship type; migration stops auto-tagging `agency`/`reseller` as partners.
  - Merge carries the union of relationship types to the survivor; archive retires them with the org.

<sup>Written for commit 4424424b9787ab1e1c52cf8f27cb998fce92a943. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/371?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Filter organizations by lifecycle and relationship type.
  * Manage lifecycle values and multiple relationship types, including partner status.
  * View inbound and outbound contact timestamps in company summaries.
  * See warnings for potentially contradictory organization facts.
  * Partner tools and tabs now appear only for applicable organizations.
  * Company views are acknowledged automatically after remaining open briefly.

* **Improvements**
  * Activity rollups now include relevant organization, deal, and contact connections without double-counting.
  * Organization merges and updates preserve relationship information more reliably.
  * Company pulse summaries now emphasize strongest contributors and contact counts instead of relationship scores.

### DIFF
--- a/backend/api/crm.yaml
+++ b/backend/api/crm.yaml
@@ -13086,6 +13086,17 @@ components:
         evidence_snippet: { type: [string, 'null'] }
         source_url: { type: [string, 'null'], format: uri }
         confidence: { type: [number, 'null'], minimum: 0, maximum: 1 }
+        suspect_reason:
+          type: [string, 'null']
+          enum: [null, phone_shaped_location, not_a_phone, not_a_year, not_an_email, not_a_size]
+          readOnly: true
+          description: >-
+            Why this fact's VALUE contradicts its FIELD, or null when it does not. The extractor
+            picks the field from a closed per-page menu and the category follows the field, so a
+            phone number on a contact page can land as `location` and a register number as
+            `employee_range` — the row is well-formed and still wrong. Computed at read from the
+            value's shape; it never gates the fact, because a heuristic that hid data would be
+            worse than one that flags it.
         updated_at: { type: string, format: date-time }
 
     OrganizationFactListResponse:
--- a/backend/api/crm.yaml
+++ b/backend/api/crm.yaml
@@ -10733,12 +10733,28 @@ components:
           format: date-time
           description: The instant the assembling transaction read. Sections are consistent to this moment under Read Committed.
         organization: { $ref: '#/components/schemas/Organization' }
+        last_inbound_at:
+          type: [string, 'null']
+          format: date-time
+          description: >-
+            When they last wrote to us, over the same three-link walk the timeline uses (the
+            activity's own link, its deal's organization, the employer of the contact it is filed
+            against). Null means nothing inbound was ever captured — which is a fact about the
+            account, not a missing field. Absent entirely when the caller has no activity grant,
+            named in `sections_omitted` as `last_touch`.
+        last_outbound_at:
+          type: [string, 'null']
+          format: date-time
+          description: >-
+            When we last wrote to them, same walk. Shown BESIDE last_inbound_at rather than folded
+            into one "last touch": which direction went last is the whole question — an account we
+            mailed a fortnight ago with no reply is not the same as one that just wrote to us.
         sections_omitted:
           type: array
           description: The sections withheld for lack of a grant — so a client can say "you can't see this" instead of "there is none".
           items:
             type: string
-            enum: [people, deals, strength, activities, tags, list_memberships, pending_approvals, next_steps, since_last_visit, suggestions]
+            enum: [people, deals, strength, activities, tags, list_memberships, pending_approvals, next_steps, since_last_visit, suggestions, last_touch]
         people:
           type: object
           required: [data, page]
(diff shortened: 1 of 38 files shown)


LarsGradion:
Closing unmerged: #392 strictly contains this work and is queued to merge.

This was the P0 half — the classification split (ADR-0079, now migration 0175), the partner invariant enforced in both directions, and the header. All of it ships in #392, which carries it plus the rest of the account page and everything three review rounds turned up on top. Landing both would apply the same migration twice.

Kept open until now as a fallback in case #392 hit trouble in review. It did — 72 findings across three cubic passes, two review subagents, three stop-gate rounds — and came out the other side rather than needing the smaller PR, so the fallback has done its job.

PR #391, opened 2026-08-03
docs(status): phase 1 A and B landed, and C is what is left

tinnhlt-gra:
`STATUS.md`'s pickup section still listed PRs 2 through 6 as open work and still
named the `embed_reindex` caveat as live. #390 shipped all five and deleted that
caveat from both files that carried it, so the entry was describing a tree that
no longer exists.

Rewrites it to what is actually left — C, the fleet metrics and the per-workspace
`/v1` read endpoint — and records three things a reader of the job layer needs
that would otherwise only exist in #390's body:

- **What #390 left honest rather than hidden.** `populated_identity` means "last
  *released* under", not "last completed under", so a run whose children all
  failed still stamps and `/readyz` reports `active`; and a forced reindex steal
  does not cancel the run it dispossesses.
- **Why a blocker survived five task reviews.** No fitness test asserts that a
  workspace worker declares a `Timeout`. The GDPR retention worker shipped
  without one and was caught only by the whole-branch pass — under River's
  1-minute default it would have been cancelled mid-pass nightly and left a
  permanently failing row for the one obligation whose point is auditability.
- **That migration renumbering must be the last action before merge.** #390's
  migration moved twice, and the second collision was invisible to local gates:
  `TestEmbeddedMigrationNamespacesLoad` passes on each side independently and the
  duplicate exists only once the trees are combined.

Docs only — no code, no contract, no migration.

---
## Summary by cubic
Update `STATUS.md` to show Phase 1 A/B merged in #390 and set Phase 1 C as the remaining work: fleet metrics and the per‑workspace /v1 read endpoint. Also document two behavior clarifications from #390 (`populated_identity` stamps on release; reindex steals don’t cancel prior runs), flag the missing worker `Timeout` fitness test, and note migration renumbering must be done last to avoid CI‑only collisions.

<sup>Written for commit 3a873a6a1c482aa45a325db4e26504b4227e13f5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/391?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/STATUS.md
+++ b/STATUS.md
@@ -285,7 +285,7 @@ Vite/React web UI. What is deliberately still stubbed (answering explicit
 The merge gate (`make check`), the real-Postgres integration lane
 (`make test-integration`), and the live-boot job are all green.
 
-## Session pickup — 2026-08-03 (job observability, Phase 0 and Phase 1 PR 1, merged)
+## Session pickup — 2026-08-03 (job observability, Phase 0 + Phase 1 A/B, merged)
 
 **Every unit of tenant work now names one workspace, and spells it one way.**
 PR #367 bound each job to a single workspace; PR #374 made the wire agree with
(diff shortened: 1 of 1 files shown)


tinnhlt-gra:
Parking this rather than chasing the merge block. The content is recorded in `.tmp/job-observability/STATUS-MD-OWED.md` and will land soon; nothing depends on it, since #390 is already merged.

For the record, in case it recurs: this PR sat `BLOCKED` by base-branch policy with all 10 applicable checks green and 10 skipped by the change classifier — identical in every measurable respect to #376, a `STATUS.md`-only PR that merged the same morning (same passing set, same review state, same merger, 0 commits behind `main`). Branch protection reads 404 for this token, so the unmet requirement could not be identified, and `--admin` would bypass the rule rather than satisfy it.

PR #462, opened 2026-08-04
fix(sbom): normalize the three SBOMs to one tree, guard parity (#459)

### LINKED ISSUE #459: make sbom: CycloneDX absolute paths + SPDX directory entries make the three SBOMs disagree (blocks dist release validation)
## Summary

`make sbom` produces three SBOMs (CycloneDX, SPDX 2.2.1, SPDX 3.0) that do **not** describe the same tree, so the release-pipeline `dist` service (margince-constellation) rejects every drafted release with `release artifacts do not agree`.

Two syft-format quirks cause it. Neither is a missing file — all 2992 real files agree once normalized. Fix and verification below are **verified against the current SBOM output** (all three sets reconcile to an identical 2992 files, zero diff).

---

# Spec: make the three Margince SBOMs agree

Apply this in the **Margince CRM repo** (the repo whose `make sbom` produces
`margince.cdx.json`, `margince.spdx221.json`, `margince.spdx300.json`). The
constellation `dist` service rejects a release when the three SBOMs do not
describe one and the same tree. They currently do not.

## 1. Contract the dist gate enforces

`constellation/pkg/sbom` parses all three formats into one model: a map of
**repo-relative file path → digests**. The `dist` gate then requires the three
maps to hold the **same key set** and agreeing digests
(`constellation/services/dist-api/internal/domain/release.go`).

Per format, the gate reads:

| Format | File list | Path field | Digest field | Algorithm spellings |
|---|---|---|---|---|
| CycloneDX JSON | `components[]` where `type == "file"` | `name` | `hashes[].{alg,content}` | `SHA-1`, `SHA-256`, `SHA-512`, `SHA3-256` |
| SPDX 2.2/2.3 JSON | `files[]` | `fileName` | `checksums[].{algorithm,checksumValue}` | `SHA1`, `SHA256`, `SHA512`, `SHA3-256` |
| SPDX 3.0 JSON-LD | `@graph[]` where `type == "software_File"` | `name` | `verifiedUsing[].{algorithm,hashValue}` | `SHA1`, `SHA256`, `SHA512`, `SHA3-256` |

Paths pass through `CanonicalPath`, which only cleans and strips a leading `/`.
It does **not** strip a directory prefix. So the SBOMs must already carry
repo-relative paths.

## 2. Defects measured in the current output

Root cause: syft scans the clean export at `/src/.tmp/sbom-src` inside the
build container (`SBOM_SRC := .tmp/sbom-src`, container workdir `/src`). The
three writers then diverge:

1. **CycloneDX carries the absolute scan path.** Every one of 2992 file names is
   `"/src/.tmp/sbom-src/<relpath>"`. SPDX names are relative (`<relpath>`). So
   **no** CycloneDX path matches any SPDX path — every file reports as
   "absent from cyclonedx-json".

2. **SPDX lists directories; CycloneDX does not.** SPDX 2.2 and 3.0 emit 3173
   entries; CycloneDX emits 2992. The 181-entry difference is **not missing
   files** — it is 180 directory pseudo-entries (`backend`, `backend/api`, …),
   each with a placeholder `SHA1 = 0000…0`, plus one scan-root entry with an
   **empty** `fileName` (SPDX 2.2) / **null** `name` (SPDX 3.0). Every real file
   present in CycloneDX is present in SPDX.

So the real file sets already agree. Only the CycloneDX prefix and the SPDX
directory/root pseudo-entries make them look different.

## 3. Required outcome

After `make sbom`, all three SBOMs must:

- name every file with the **same repo-relative path** (no `/src/...` prefix, no
  leading `/`), and
- enumerate the **same set of regular files** — no directory entries, no
  empty/null root entry.

## 4. Fix

Two independent problems. Fix both. A deterministic post-generation
normalization pass is the most reliable approach, because it does not depend on
syft flag behaviour staying constant across versions.

### 4a. Strip the CycloneDX scan-root prefix

Preferred — remove the prefix at the source by scanning **relatively**: run syft
with its working directory set to the export root and scan `dir:.`, instead of
passing the absolute path `/src/.tmp/sbom-src`. A relative source root makes
CycloneDX names relative too. Concretely, in the `sbom` recipe, `cd` into
`$(SBOM_SRC)` and scan `dir:.`:

```make
# inside the container, before invoking syft
cd $(SBOM_SRC) && syft scan dir:. -o cyclonedx-json=... -o spdx-json=... -o spdx-json@2.2=...
```

If the recipe cannot scan relatively, strip the prefix in a normalization pass
(see 4c).

### 4b. Drop directory and root pseudo-entries from SPDX

syft's SPDX writers record a `SpdxFile` for directories with a zero SHA1 and a
root element with no name. CycloneDX omits both. Remove them from the two SPDX
files so all three enumerate the same regular files. Identify a pseudo-entry by
an empty/null name **or** a lone `SHA1 = 0000…0`.

### 4c. Deterministic normalization pass (recommended, covers 4a + 4b)

Add a `jq` normalization step to the `sbom` recipe, run on each SBOM after syft
writes it and **before** cosign signs it, so the signature covers the
normalized bytes.

```bash
# CycloneDX: strip the scan-root prefix from every file component name.
jq --arg p "/src/$(SBOM_SRC)/" '
  .components |= map(
    if .type == "file" and (.name | startswith($p))
    then .name |= sub("^" + $p; "")
    else . end)
' margince.cdx.json > tmp && mv tmp margince.cdx.json

# SPDX 2.2: drop directory/root pseudo-entries (empty name, or lone zero SHA1).
jq '
  .files |= map(select(
    (.fileName | length) > 0 and
    (.checksums | any(.checksumValue != "0000000000000000000000000000000000000000"))
  ))
' margince.spdx221.json > tmp && mv tmp margince.spdx221.json

# SPDX 3.0: drop software_File elements with null/empty name or lone zero SHA1.
jq '
  ."@graph" |= map(select(
    (.type != "software_File") or (
      ((.name // "") | length) > 0 and
      ((.verifiedUsing // []) | any(.hashValue != "0000000000000000000000000000000000000000"))
    )
  ))
' margince.spdx300.json > tmp && mv tmp margince.spdx300.json
```

Note: `$(SBOM_SRC)` is `.tmp/sbom-src` and the container workdir is `/src`, so
the CycloneDX prefix is `/src/.tmp/sbom-src/`. Derive it from the variable — do
not hardcode — so it tracks any change to `SBOM_SRC`.

If you adopt 4a (relative scan), keep only the two SPDX filters from 4c.

## 5. Verification — must pass before publishing

Cross-check the three normalized SBOMs describe the same file set. This is the
same comparison the `dist` gate makes. Run it in the `sbom` recipe (or CI) and
fail the build on any diff:

```bash
jq -r '.components[]|select(.type=="file")|.name'                 margince.cdx.json     | sort -u > /tmp/cdx.set
jq -r '.files[].fileName|select(length>0)'                        margince.spdx221.json | sort -u > /tmp/s22.set
jq -r '[.["@graph"][]|select(.type=="software_File")|.name]|.[]' margince.spdx300.json  | sort -u > /tmp/s30.set

# every pairwise diff must be empty
diff /tmp/cdx.set /tmp/s22.set && diff /tmp/s22.set /tmp/s30.set \
  && echo "OK: three SBOMs list the same file set" \
  || { echo "FAIL: SBOM file sets differ"; exit 1; }
```

Expected after the fix: identical sets (~2992 regular files), no `/src/...`
prefix, no directory entries.

## 6. Regression guard

Add the section-5 cross-check to the Margince CI `sbom` job, so a future syft
upgrade that reintroduces the prefix or directory entries fails the build rather
than the release. The constellation side already guards the shared parser and
the dist agreement rules with unit tests; the enumeration parity is the Margince
build's responsibility.

## Out of scope (constellation side, already handled)

- Patch parse of the generator's blank-line `x-checksum` block — constellation
  PR #146.
- The unreadable rejection message — constellation PR #148 (now one summary line
  per out-of-step SBOM).

lstrojny:
Closes #459.

## Problem

`make sbom` produces three SBOMs (CycloneDX, SPDX 2.2.1, SPDX 3.0) that do not describe the same tree, so the constellation `dist` gate rejects every drafted release with `release artifacts do not agree`. Two syft-writer divergences cause it — the real file sets already agree:

1. **CycloneDX carries the absolute scan-root prefix.** Every file component is named `/src/.tmp/sbom-src/<relpath>`; SPDX names are repo-relative — so no CycloneDX path matches any SPDX path.
2. **SPDX lists directories; CycloneDX does not.** The SPDX 2.2/3.0 writers emit a pseudo-entry per directory (lone `SHA1 = 0000…0`) plus one empty-name scan-root entry that CycloneDX omits.

## Fix

A deterministic post-generation normalization pass (jq), so the result does not depend on syft flag behaviour staying constant across versions.

- **`sbom-normalize`** — strips the CycloneDX scan-root prefix (derived from `SBOM_SRC`, not hardcoded) and drops the SPDX directory/root pseudo-entries. Runs on the syft output **before any signature**, so the signed bytes are the normalized bytes. Both filters are idempotent — a future syft that already emits relative names / omits directories is a no-op.
- **`sbom-parity`** — the `dist` gate's own comparison run locally: the three SBOMs must enumerate the identical file set, and it **fails the build on any diff**. `make sbom` runs it after normalization, so a syft upgrade that reintroduces the prefix or directory entries breaks the SBOM CI job rather than release validation (§6 regression guard).

## Verification

- Reproduced the two defects on live output (3026 → after fix: three sets reconcile to an identical file set, zero diff).
- `sbom-normalize` is idempotent (re-run leaves bytes unchanged); all three SBOMs remain valid JSON.
- `sbom-parity` exits 0 on agreement and fails loudly (`FAIL: SBOM file sets differ`, make Error 1) on an injected prefix regression.
- Full `make sbom` runs generate → normalize → parity end-to-end and passes on the current tree.

---
## Summary by cubic
Normalize the three SBOM outputs to a single repo‑relative file tree, add a local parity check, and gate signing on parity. Fixes #459 where the dist gate rejected releases due to path prefixes and directory pseudo‑entries.

- **Bug Fixes**
  - Added `sbom-normalize` to strip the CycloneDX scan-root prefix (from `SBOM_SRC`) and remove SPDX directory/root pseudo-entries; runs before signing and is idempotent.
  - Added `sbom-parity` to compare the three SBOM file sets and fail on diffs; `make sbom` runs normalize + parity, and `sbom-sign` now depends on parity so only normalized, matching bytes are signed.

<sup>Written for commit e5ee80be3955035f173dd732133fe76dc07131d3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/462?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved software bill of materials (SBOM) consistency by normalizing generated files and verifying they contain matching repository file listings.
  * Added validation to detect missing prerequisites and report failures during SBOM generation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@
 # one target here that invokes the compiler directly instead of delegating.
 GO ?= go
 
-.PHONY: help install ai-routing-local dev-fresh check check-backend check-q check-go check-gates check-fe build test test-v test-cover test-integration e2e-ai e2e-ai-report ai-probe test-db-up test-it test-integration-serial bench-perf lint arch-lint vet gen gen-types gen-types-check drift composition check-composition test-extensions db-up db-init db-wait migrate migrate-up migrate-down run psql redis-cli tidy dev dev-stop dev-logs clean tools tools-go infra-up infra-down infra-logs infra-reset seed-dev seed-dev-db seed-reset verify-boot frontend-check frontend-e2e fe-install fe-typecheck fe-lint fe-build fe-preview fe-format fe-test ds-purity font-lock icon-lint fitness-jurisdiction storybook fe-uat craft-static craft-residue check-craft-doc check-image-pins contract-breaking-check test-lanes go-file-length rls-store-path no-jurisdiction pkg-freeze hooks sbom sbom-sign sbom-check
+.PHONY: help install ai-routing-local dev-fresh check check-backend check-q check-go check-gates check-fe build test test-v test-cover test-integration e2e-ai e2e-ai-report ai-probe test-db-up test-it test-integration-serial bench-perf lint arch-lint vet gen gen-types gen-types-check drift composition check-composition test-extensions db-up db-init db-wait migrate migrate-up migrate-down run psql redis-cli tidy dev dev-stop dev-logs clean tools tools-go infra-up infra-down infra-logs infra-reset seed-dev seed-dev-db seed-reset verify-boot frontend-check frontend-e2e fe-install fe-typecheck fe-lint fe-build fe-preview fe-format fe-test ds-purity font-lock icon-lint fitness-jurisdiction storybook fe-uat craft-static craft-residue check-craft-doc check-image-pins contract-breaking-check test-lanes go-file-length rls-store-path no-jurisdiction pkg-freeze hooks sbom sbom-normalize sbom-parity sbom-sign sbom-check
 
 # Bare `make` lists every command instead of running the first target.
 .DEFAULT_GOAL := help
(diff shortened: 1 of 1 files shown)


lstrojny:
Review findings at `dd99e7a` — changes requested:

1. **BLOCKER — SPDX 3 normalization leaves a broken graph.**

   The filter removes only `software_File` nodes. In the SBOM artifact uploaded by this PR, that leaves **368 `Relationship.from` references** whose target no longer exists and **184 stale IDs in `software_Sbom.element`**. SPDX 3 requires relationship endpoints and collection members to be Elements:
   - https://spdx.github.io/spdx-spec/v3.0.1/model/Core/Classes/Relationship/
   - https://spdx.github.io/spdx-spec/v3.0.1/model/Core/Classes/ElementCollection/

   Compute the pseudo-file ID set, remove every relationship whose `from` or `to` touches it, remove both the file IDs and the removed relationship IDs from `software_Sbom.element`, and then assert that every local `SPDXRef:` reference resolves.

2. **HIGH — `sbom-parity` does not mirror the dist contract.**

   It compares only names with `sort -u`. A digest mismatch or duplicate path would pass. The current artifact itself is good: all 3,118 canonical `path → digests` records agree and none are duplicated. The guard nevertheless needs to compare canonicalized path-and-digest records and explicitly reject duplicate paths.

3. **MEDIUM — the pseudo-entry predicate is broader than documented.**

   The filters drop entries whenever all hashes are zero. They do not require the exact pinned-Syft marker: one SHA-1 hash, the correct algorithm spelling, and the zero value. SPDX does not reserve an all-zero SHA-1 as a directory sentinel. Match the exact generator-specific shape and fail on unexpected zero-hash shapes rather than silently deleting them.

4. **MEDIUM — the documented host-binary override is broken.**

   The normalizer assumes `/src/$(SBOM_SRC)/`, which is specific to the Docker invocation. With `SYFT=syft`, Syft 1.50.0 emits the actual absolute host scan path, so the prefix remains and parity fails. Scanning `dir:.` also emits an absolute path in Syft 1.50.0, so the prefix must be derived from the actual scan environment or the host override must be handled explicitly.

The CycloneDX transformation and SPDX 2.2 cleanup otherwise look sound. Valid JSON and the green SBOM job do not establish SPDX model validity; the uploaded SPDX 3 graph is currently internally inconsistent.

lstrojny:
Thanks @coderabbitai — went through all three:

**Sign non-normalized bytes (Major) — addressed, adjusted form.** Added a prerequisite so signing can only cover normalized, mutually-agreeing bytes, but `sbom-sign: sbom-parity`, not `sbom-sign: sbom`. In CI, signing is a deliberately isolated job that holds `id-token: write` and consumes the generation job's artifact; making `sbom-sign` depend on `sbom` would run the full syft scan under that token, which is exactly the generate/sign isolation the SBOM workflow forbids. `sbom-parity` re-checks the existing files cheaply (no scan, no network) and refuses to sign a stale or un-normalized set.

**SPDX 3.0 filter by `software_fileKind` (Trivial) — declined, false positive for the pinned syft.** The suggestion assumes syft tags directories `fileKind == "directory"`. Verified against the actual output of the pinned syft (v1.50): every SPDX 3.0 element, directories included, carries `software_fileKind: "file"` — e.g. the `backend` directory pseudo-entry is `{"name":"backend","software_fileKind":"file","verifiedUsing":[{"algorithm":"sha1","hashValue":"0000…0"}]}`. So the kind field cannot single a directory out here; switching to it silently kept all 180 directory entries in SPDX 3.0 and broke parity. The lone zero-SHA1 placeholder (never present alongside a real file's SHA-256/512) is the only signal syft actually gives. Kept the hash heuristic and documented why in the recipe comment.

**Extract to helper scripts (Trivial) — declined.** `checkmake` is not a gate in this repo (no CI or Makefile reference), so `maxbodylength` isn't binding. The inline recipes match the existing `sbom` / `sbom-check` / `sbom-sign` targets and the SBOM toolchain's design where the Makefile owns the scan/normalization policy directly; a separate script tier plus tests would be more surface for a build target, not less.

PR #464, opened 2026-08-04
fix(auth): the reset token stops leaving the browser, and one spelling for the emailed link

joshbruh-lang:
Lifts two of @luitpold's commits out of #339 so the onboarding PR carries no backend changes. Both are cherry-picked whole — frontend and backend together — because they are one fix each and splitting them would break the flow.

Commits are unchanged and authorship is preserved. The only edit is a conflict resolution in `frontend/src/main.tsx`, where main's 403 handler and this branch's early `applyTheme` both wanted the same spot; both are kept.

## What these fix

**The password-reset token was travelling in the query string**, and three paths leaked it before `auth.tsx` could scrub it:
- the service worker cached it to disk (a Cache Storage key includes the query, and it was then served cache-first),
- the browser sent the full URL in the `Referer` on same-origin API calls, since no `Referrer-Policy` was set,
- nginx logged the query by default.

The emailed link now carries the token in the fragment, which never reaches a server, never appears in a `Referer`, and is not part of a cache key. `Referrer-Policy: no-referrer` and a service worker that refuses to cache any query-bearing URL are added as depth. No compatibility shim: `password_reset` is false on every installation without a mailer, so no link of the old shape exists in the wild, and a test asserts the query form is ignored.

**The set-password link had two spellings**, one per call site. Now built in one place, gated.

## Verification

- `make check-fe`: 125 files, 1324 tests, lint, tsc and build green.
- `go build ./...` clean; `go test ./internal/modules/identity/...` green across identity, password and policy.

## Sequencing

Once this merges, merging `main` into `feat/onboarding-surface-redesign` drops these files from #339's diff — no history rewriting on the shared branch.

---
## Summary by cubic
Keep password reset and invite tokens in the URL fragment so they never leave the browser, and unify deep-link building. Also fixes unauthenticated screens with correct theming, clearer error messages, and better accessibility.

- **Bug Fixes**
  - Security: tokens now use `#/reset-password?token=…`; all emails go through `passwordLink`; query-string tokens are ignored; `Referrer-Policy: no-referrer` added; the service worker stops caching any URL with a query; reset mail goroutine is panic-safe; tests enforce fragment-only.
  - Reset UX: classify errors (token/password/rate-limited/server); show “Request a new link” only for token errors; error region gets `role="alert"`; improved EN/DE copy.

- **Refactors**
  - Theming and a11y: theme is resolved/applied at boot (`app/theme.ts`) to avoid flashes and cover unauthenticated pages; stronger non-text contrast via `--borderControl` and dark `--textOnAccent`, with themed shadows; remove nested `<main>`; set `document.documentElement.lang` from i18n; add stable input `name`s and mobile/email hints; tests updated.

<sup>Written for commit 258e3be6d61453400927d333cc0b84a6640f9e07. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/464?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/internal/modules/identity/handlers_users.go
+++ b/backend/internal/modules/identity/handlers_users.go
@@ -138,7 +138,7 @@ func (h Handlers) sendInvite(r *http.Request, email, rawToken string) {
 	if h.resetMailer == nil || rawToken == "" {
 		return
 	}
-	link := h.resetBaseURL + "/reset-password?token=" + rawToken
+	link := passwordLink(h.resetBaseURL, rawToken)
 	body := "You've been invited to Margince.\n\n" +
 		"Set your password within seven days to sign in:\n\n  " + link + "\n\n" +
 		"If you weren't expecting this, you can ignore this email."
--- a/backend/internal/modules/identity/passwordlink.go
+++ b/backend/internal/modules/identity/passwordlink.go
@@ -1,0 +1,33 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+package identity
+
+// passwordLink is the ONE spelling of an emailed set-password deep link, and the
+// reason it is a function rather than a string per caller is that the token's
+// PLACEMENT in the URL is a security property — not a decision each caller gets
+// to make. Both mails that use it carry a live single-use credential: the reset
+// link for one hour, the invite link for seven days.
+//
+// The token rides in the FRAGMENT, which a BROWSER never puts on the wire. That
+// closes the three legs a query string loses on: the credential cannot reach an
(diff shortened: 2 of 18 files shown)


joshbruh-lang:
Closing — the reset-token and set-password-link fixes stay on feat/onboarding-surface-redesign and ship with #339.

PR #465, opened 2026-08-04
feat(person): the Relationship Room — the person page opens on a reason

LarsGradion:
## The person page opens on a reason, not a record

"Warm, 73" describes a contact. "They replied after 41 quiet days and nobody has answered" is something to do. This branch builds the second, and the machinery that makes it honest.

Built against the shared dev database — 247 people, ~2,950 captured activities — rather than fixtures. Three of the defects below only exist to be found that way.

## What landed

**Multi-party participants.** Mail capture never parsed `Cc` at all; the calendar folded attendees into body text. Both now emit `NormalizedRecord.Participants`, **resolved to a colleague or a known contact at stamping time** rather than left for a promotion that never comes — the interaction graph joins `user_id` to `person_id`, so an address-only row is invisible to it. The counterparty can wait because the ensure path promotes its row moments later; nothing promotes a CC.

A replay pass recovers history from the stored originals. Migration 0185 adds its marker: the two-end backfill needs no state because its predicate shrinks as it runs, and this one cannot borrow that trick, because most messages have no CCs and the result therefore cannot say whether an original was ever re-read.

**Relationship change, derived and never stored.** `relstrength.Changes` folds the same §4 curve over a window ending in the past, so the system can say *when* a relationship moved without storing yesterday's number. Only a **band crossing** is reported — the score decays continuously, so "73 became 71" is arithmetic and reporting it would fire on every read. No table, so an erased activity takes its derived change with it.

**The correction ledger (`ai_feedback`).** Specified upstream since the ai-runtime chapter (AIRT-SCHEMA-1) and never built. `claim_key` hashes the claim's **path**, never its value, so a human's verdict survives the evidence shifting underneath it — keyed on the value it would evaporate exactly when their answer matters most. Migration 0186. `POST /ai/feedback` is human-only and gated on the **subject's** update grant: correcting what the system says about a contact is editing that contact, so no new RBAC object is invented that could drift from the one it mirrors.

**Moments.** Five deterministic rules over what the 360 already read, ranked in a fixed editorial order. Deterministic buys three things: each carries evidence a reader can open, the page opens on a reason at first paint with no model in the path, and a moment cannot assert something that did not happen. Dismissal writes an `ai_feedback` verdict keyed on the moment's path.

**The local graph.** `GET /people/{id}/graph`. A `direct` arm carrying up to three real messages, each passing the caller's own activity row scope **inside the query** — so a message they may not read is absent from the proof while still counted in the total. An `account` arm carrying pooled counts and dates only: interaction metadata is disclosable where the correspondence is not. Row scope is applied **per arm**, never once at the root, or the graph would disclose by adjacency what the record list withholds.

**Frontend.** Moment card with dismissal, change lines on the pulse, correction UI on enriched fields, the Connections panel, timeline filters.

## GDPR, declared up front rather than discovered by the gate

`ai_feedback` holds a human-typed value about the subject, so it is registered in `piiTables`, Art. 17 deletes it in the single erasure transaction beside `field_provenance`, and Art. 15 exports it as `corrections`.

## What running it against real data caught

- Three wrong column names.
- An employment predicate keyed on `is_current_primary`, which answers *which of several employers is the main one*, not *are they still there* — it would have dropped any colleague holding a second post.
- A comma inside an unquoted YAML flow-mapping description in `crm.yaml`, which split into a sibling key and made the **whole contract fail OpenAPI validation** while codegen stayed happy about it.
- The graph panel crashed the entire record page on a response missing its `nodes` array. The first regression test for it asserted an empty container — which is also what a crashed React tree produces, so it passed with the guard removed. It now asserts a **sibling survives**, and waits for the read to resolve first. Mutation-checked both ways.

## Gates

`make check` green. `craft static` **PASS — 0 blocker, 0 major, 0 minor** under the now-strict bar. `make test-integration`: 28/28 packages, zero failures, zero skips.

The integration lane took five attempts to produce a trustworthy answer, and the reason belongs in the record: a parallel session was running the same lane against the shared `margince_test` template, producing 1,235 deadlocks and a half-migrated schema underneath them. It found exactly one real defect — `activity_participant_replay.activity_id` needed a row-scope classification — now fixed.

## Deliberately not here

**C2 (person brief and ask) and D (commitments) are blocked upstream, not deferred.** Both need AI task declarations generated from `ai-tasks.yaml`, which still calls the relevant task `planned`. Building D's lifecycle alone would ship a queue nothing fills — the extraction that would stage a proposal is itself one of the blocked tasks. Contract-first: they wait for a spec change rather than a workaround.

**Search-based enrichment is built and dormant.** The `websearch` seam (ADR-0081) and LinkedIn-URL discovery are in; they stay off until `BRAVE_SEARCH_API_KEY` is bound, which is the sovereign posture rather than a degraded one.

---
## Summary by cubic
Redesigned the person page to open on a clear, actionable reason and load the whole view in one composite read. Adds trustworthy evidence, sane disclosure rules, governed discovery/enrichment, and a local graph with a deterministic route.

- **New Features**
  - `GET /people/{id}/360` single payload: identity, roles, strength, split last inbound/outbound, timeline, consent, enrichment evidence, since‑last‑visit; omitted‑and‑named sections when grants lack. `POST /people/{id}/view-ack` records the baseline.
  - Moments: deterministic reasons with evidence; dismissal writes a ledger verdict keyed to the moment path via `POST /ai/feedback`.
  - UI: RelationshipPulse (words first; math on disclose), Enriched field corrections, Connections panel (`GET /people/{id}/graph`), thin state for correspondence‑free records, All/Messages/Meetings/Tasks filters.
  - Capture: records CCs and calendar attendees as `connector.NormalizedRecord.Participants`; replay pass re‑reads stored originals to recover historic CCs/attendees.
  - Local graph: `GET /people/{id}/graph` with direct edges (up to 3 receipts, each row‑scoped) and account edges (counts/dates only); per‑arm RLS and a deterministic recommended route.
  - Auto‑enrichment: consumer `cg:person-auto-enrich` fills empty person fields from employer site reads; governed `websearch` seam discovers public profile URLs from result metadata (optional; off unless `BRAVE_SEARCH_API_KEY` is set).
  - Migrations: `0184` widens `user_record_view` to include `person` (rollback fixed for RLS); `0185` adds `activity_participant_replay`; `0186` adds `ai_feedback`.

- **Bug Fixes**
  - Security: prevented forged relationships — further‑party stamping trusts recipient lists only when the provider attests the owner sent the message.
  - GDPR: profile‑fields read is live‑only; `person_profile_field` covered by erasure, retention, SAR; lead‑subject corrections are deleted on erasure; merge relinks enriched fields onto the survivor.
  - Graph: true `dropped_count` for account; direct arm’s `dropped_count` stated as bounded; fixed `has_more` and evidence ordering; per‑arm refusals (404 on out‑of‑scope/archived subjects, receipts withheld without losing routes, no leakage via counts, no route from empty histories).
  - Search guardrails: whole‑word name matches with consistent tokenization (handles hyphens/apostrophes); derive profile field from host; drop `xing`; rune‑safe snippet clipping; refuse credentialed/plaintext URLs; strict URL/snippet canonicalization; skip paid search when a LinkedIn URL exists; provider/decoder errors do not leak query terms; refuse non‑person results; degrade cleanly when no provider is bound; search‑discovered fills now write field history correctly with evidence captured in the right column.
  - UI: timeline filter moved to header; thin state respects `sections_omitted`; graph panel guarded so it cannot take the page down; view‑ack monotonic; corrections applied on both the profile‑fields sidecar and the 360.
  - Tests: stronger integration coverage for participants replay, person 360 wire and moments dismissal, local graph refusals, auto‑enrich, and search discovery; stubs assert request shape to catch silent regressions.

<sup>Written for commit b18ab67fc670e0a276f8040bc6b2c0043cdbb0a1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/465?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added a comprehensive Person 360 view with identity, career, relationship, activity, consent, and network details.
  * Added relationship graphs with recommended routes, interaction evidence, and visibility-aware results.
  * Added recommended moments, relationship changes, and persistent dismissals.
  * Added evidence-backed profile enrichment with confirmation and correction controls.
  * Added automatic person enrichment from approved professional profile discoveries.
* **Improvements**
  * Improved participant capture for email and calendar activities, including historical backfill.
  * Added directional interaction metrics, German translations, and expanded privacy export and erasure coverage.


LarsGradion:
Review triage, closing out the threads on this PR.

Every **P1** raised by cubic and CodeRabbit is fixed on this branch. The **P2/P3** findings are carried in #488 — presentation/locale issues in the new screens, correction-UI affordances, graph panel state, the suppressed-verdict contract gap, the not-yet-wired parts of the ADR-0081 search seam, and a set of smaller correctness items. Filing them there is what the repo rule asks for: anything found but not fixed in the change becomes an issue rather than disappearing with a resolved thread.

Resolving the threads now on that basis.

LarsGradion:
Second review round, all eleven threads addressed in b18ab67f.

**Two real defects, both cubic's:**

- `mentionsName` cut the result text at every non-alphanumeric but split the name on whitespace alone, so `Jean-Luc` and `O'Connor` stayed whole and could never match the tokens the page was cut into. Every contact with punctuation in their name was undiscoverable. Both sides now use one tokenizer, with a test that fails if the name reverts to a whitespace split.
- The search-fill audit put bookkeeping keys in the after-image, which `storekit` reserves for the record's own per-field images — so a discovered field never reached field history. The source moves to the evidence column and the images carry each field with the value written. The sibling site-fill writer has the same shape but predates this branch (#283), so it goes to #488 rather than widening this PR.

**Six tests that asserted less than they claimed**, all correct catches:

- The Brave stub discarded the request, so the site-narrowing test never inspected the query it narrows and the no-call test never proved no call. The stub now records queries and both assert on them.
- The graph test counted nodes and rejected duplicate ids, which a graph that drew the colleague twice would also satisfy. It now names the shared end and requires both edges to meet there.
- The no-employer case passed on an empty database; a foreign organization now stages a page naming the same person, so the employer gate is what the assertion rests on.
- A survivor read ignored its iterator error, a replay helper folded a failed query into "no row", and a withdrawal assertion read the database wall clock — now compared against the row's own creation time instead.

One skipped: the redundant file-header line was removed as asked.

PR #479, opened 2026-08-05
feat(agents): the passport cap an operation spends comes from the contract

tinnhlt-gra:
## The defect

`compose/agentgate.go` admitted any tool verb with no registered MCP tool under a hardcoded cap:

```go
return mcp.ToolSpec{Name: pol.Tool, RequiredScope: principal.ScopeWrite, Tier: tier}, true
```

Eleven contract-declared verbs have no registered tool, so all eleven ran on `write`. Three of them egress:

| verb | what it does | ran on | should be |
|---|---|---|---|
| `enrich` | fetches the target's website (`enrichextract.go:498`, `deepread.go:134`) | `write` | `enrich` |
| `reconcile_overlay` | queues a sweep that calls the incumbent's API (`reconcile.go:117`) | `write` | `enrich` |
| `send_offer` | the contract promises delivery leaves the workspace (`crm.yaml` `sendOffer`) | `write` | `send` |

`ScopeSet.Has` is exact membership with no hierarchy, so a passport whose granting human deliberately withheld `enrich` or `send` spent `write` instead and the cap was never enforced.

`agentpolicysynthesis_test.go` recorded this in three prose maps. Documenting is not enforcing: nothing in product code read them, and the runtime behaved identically whether a verb was pinned or not.

## The fix — the cap is contract-declared, like the tier already was

The tier was declared once in the contract and enforced below the transport. The cap now is too. They answer different questions and neither substitutes for the other: **a tier says whether a human confirms the act; a scope says whether the act was ever delegable.**

- **`crm.yaml`** — `AgentAdmissionPolicy` gains a `scope` vocabulary (`read|draft|write|send|enrich`, mirroring `principal.Scope`). All **104** `x-mcp-tool` annotations declare one.
- **`gen-agentpolicy`** — reads and validates it through the existing vocabulary machinery, so a typo fails generation naming the `operationId`. Two new build-time defects:
  - `x-mcp-tool` with no `scope`. Absent is **not** a meaningful state here, unlike `tier` and `record_type` — a default is precisely what made every verb look internal.
  - `scopeCoherence`: one verb, one cap. Tier may legitimately differ per route (A34 tighten-only), but scope is a property of the *act*, not the route reaching it — a verb that is `enrich` on one path and `write` on another lets a denied agent pick the other door.
- **`agentgate.go`** — uses `principal.Scope(pol.Scope)`. Dynamic-tier-without-a-registered-tool still fails closed, unchanged.

### Which cap an act spends

Decided by the act's **purpose**: `send` delivers to a counterparty, `enrich` pulls from a third party, and an act whose purpose is a durable state change is `write` even where it makes network calls to do it.

That puts **`connect_incumbent` at `write`, not `enrich`**, even though it does call the incumbent's API (`overlay/connection.go:314-323,341`). It seals a credential and flips `x_sor_mode`; the fetches are incidental. Because scopes are flat, `enrich` would have let an enrich-only passport reach both. This is the one non-obvious call in the table and it is stated in the pin.

## Fitness functions replacing the prose

Per the repo rule "prefer fitness functions over point fixes":

- **Contract ↔ registry parity** — for every policy row whose verb IS registered, the contract's scope must equal the tool's `RequiredScope`. A verb can no longer mean `enrich` on MCP and `write` on REST.
- **Every tool row declares a grantable cap** — a contract scope the kernel doesn't know would make the gate demand a cap no passport can hold: fail-closed but silently broken.
- **`Egress` agrees with scope** — derived via the new `principal.Scope.Egresses()` rather than hand-set per spec.
- **`outboundHoles` is deleted.** Its entire premise was "egresses but admitted at `write`", now true of nothing. Its four verbs move to `synthesizedVerbs` with reasons rewritten to what is actually true. `deadEndVerbs` (`share_record`) stays — unrelated, still true.

## ⚠️ Behaviour change

A tightening with teeth. A passport holding `write` but **not** `enrich`/`send` is now refused with `scope_exceeds_grantor` (403) where it was previously admitted and staged a 🟡 approval:

- `POST /v1/organizations/{id}/enrich`, `/deep-read`, `POST /v1/coldstart`, `/coldstart/preview` → need `enrich`
- `POST /v1/overlay/reconcile` → needs `enrich`
- `POST /v1/offers/{id}/send` → needs `send`

Existing passports minted without those caps lose access to those routes. That is the point of the change, but it is visible.

## Two adjacent fixes this change forced

1. **`extensiontools.go`** derived `Egress: scope == principal.ScopeSend`. Once `enrich` is a real egressing cap, an extension declaring it would fetch the web while reporting itself workspace-local in the operator inventory. Now `scope.Egresses()`.
2. **`modules/agents/scope_fitness_test.go`** asserted *every* egressing tool must require `send`. The first `enrich`-scoped registered tool would have failed a test that was itself wrong. Narrowed to "an egressing tool must spend a cap that egresses."

## Verification

- `make check` — **EXIT 0** (build, vet, lint, arch-lint, unit, contract drift; frontend 1635 tests + build)
- `make test-integration` — **`OK: integration passed with 0 skips (29 packages)`**
- `make craft-static` — `PASS — 0 blocker, 0 major, 0 minor`
- Frontend `schema.d.ts` regenerated; `make gen` reproducible

Every new assertion was **mutation-verified** — flipping `search_records` to `enrich`, blanking a scope, setting `book_meeting`'s `Egress` false, and reverting `operationSpec` to the hardcoded cap each produced a failure naming the specific defect.

The integration tests are the ones that matter most here, because the defect lived in the gate at runtime. Note the subtlety they had to handle: **a missing-cap refusal and a staged 🟡 approval are both 403.** The observable difference is whether an approval row exists — a staged approval is a durable authority object (ADR-0036) — so an assertion on status alone would have passed against the old code. Reverting `operationSpec` makes both new tests fail with a staged approval, which is exactly the old shape.

## Out of scope

Registering an actual `enrich` MCP tool. The verb stays absent from `tools/list`; only its REST admission is corrected. Follow-up issue to be filed.

## Upstream (P3)

Adding `scope` to the `x-mcp-tool` vocabulary is an upstream-visible contract change. Made here and raised as an issue for `margince-foundation` reconciliation against `interfaces.md §2`, rather than edited into the spec repo from this side.

---
## Summary by cubic
Make the gate spend the contract-declared `x-mcp-tool.scope` (`read|draft|write|send|enrich`) instead of defaulting to `write`, so withheld `enrich`/`send` can’t be bypassed via REST or missing MCP tools. Also keep scope vocabularies equal to grantable passport scopes and make the parity checks strict.

- **New Features**
  - Contract + generator: add `x-mcp-tool.scope` to all tools; validate vocabulary and one-verb→one-cap (`scopeCoherence`); hold the scope vocabulary equal to grantable passport scopes; fail if tier/scope enums are missing or empty (no silent pass).
  - Gate/runtime: `operationSpec` uses the contract scope when no MCP tool is registered; `Egress` is derived via `principal.Scope.Egresses()`; removed `outboundHoles`; outbound synthesized verbs are pinned with reasons.
  - Parity + tests: contract scope must match tool `RequiredScope`; an egressing tool must spend an outbound cap; integration tests verify missing-cap refusals (no staged approval) vs staged 🟡 approvals over HTTP.
  - Docs/types + inventory: updated authorization/how-to guides and `schema.d.ts`; README/STATUS note scopes; operator inventory uses `scope.Egresses()`; identity drops method→scope mapping and REST send routes spend `send`.

- **Migration**
  - Passports missing `enrich` or `send` lose access to: `/v1/organizations/{id}/enrich`, `/deep-read`, `/coldstart`, `/coldstart/preview`, `/overlay/reconcile`, `/offers/{id}/send`.
  - Mint agent passports with needed `enrich`/`send`.
  - All `x-mcp-tool` annotations must include `scope` (generation fails otherwise); re-run generation.

<sup>Written for commit d0d1d61d4a6be968f1975eada081bf69bcc93675. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/479?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added explicit passport-scope requirements to MCP tools and API operations.
  - Introduced scope validation for `read`, `draft`, `write`, `send`, and `enrich`.
  - Added stricter protection for outbound actions, including enrichment and sending.

- **Bug Fixes**
  - Requests missing the required scope are now consistently refused before authorization or approval processing.
  - Prevented scope caps from being bypassed across transports or unregistered actions.

- **Documentation**
  - Updated authorization, governance, and endpoint guidance with the new scope requirements.


tinnhlt-gra:
Addressed all three cubic findings in 2a02db28.

**1. Egress coverage for synthesized verbs was dropped — valid, and the important one.** Confirmed by mutation: weakening `enrich` to `scope: write` in the contract passed every gate on the previous commit. Deleting `outboundHoles` did take a real guarantee with it.

One correction to the suggested fix, though. *"Assert an egressing synthesized verb spends an egress cap"* cannot be implemented as stated, for two reasons:

- It would **fail on `connect_incumbent`**, which does call the incumbent's API yet is deliberately `write`. The cap follows the act's *purpose*, and what connect durably does is seal a credential and flip `x_sor_mode`. Scopes are a flat set, so `enrich` would admit an enrich-only passport to both.
- It would be **circular**. A synthesized spec's `Egress` is computed *from* its scope via `principal.Scope.Egresses()`, so a scope weakened to `write` makes `Egresses()` report `false` and agree with itself. Nothing fails. The test cannot ask the code whether the act egresses — the code's answer is derived from the very field under test.

Egress is a fact about the world, not a derivable one, so `TestEgressingSynthesizedVerbsSpendTheCapTheyArePinnedTo` states it once per verb with its reason and holds the contract to it. The pin also fails if a verb *gains* a registered tool — at which point `agentscopeparity_test.go` owns it and the pin must be deleted, so the two sweeps hand off cleanly instead of overlapping.

**2. `len(delivers) == 0` guards the wrong direction — valid.** It caught "no spec declares egress" but not "no contract route resolved to one," which is the case that would actually go quiet. Now counts what it checked and fatals at zero.

**3. README wrapping — valid.** Re-wrapped before "Governance actions".

`make check-backend` EXIT 0, `make craft-static` PASS 0/0/0.

---

Also worth recording here: UAT ran against a live stack — 27 cases across the regression, no-regression, error-taxonomy and build-gate sections, **zero failures**. `A1` vs `A2` is the change in two rows: same route, same tier, one cap apart — `403 scope_exceeds_grantor` without `enrich`, `403 approval_required` with it. Both are 403, which is precisely why the tests assert on the code and on whether an approval row exists rather than on status.

UAT also surfaced a **pre-existing** defect unrelated to this branch, filed as #484: `connect_incumbent` is declared 🟡 but has no approval-kind mapping, so an agent-initiated overlay connect always ends in `permission_denied` and can never be staged. Its admission is bit-identical to before this PR (the contract declares `write`, which is what the gate hardcoded), so it is not a regression here.

tinnhlt-gra:
Fixed in d0d1d61d. Valid, and it applied to more than the new line.

Both cross-vocabulary checks skipped themselves when the schema they compare against was missing or declared no enum — the `ok` guard and the `len(...) > 0` guard each turned the check off silently. A check that quietly compares against nothing is worse than no check, because it reads on the page as an enforced invariant and the next person stops looking. For the scope pairing the cost is concrete: a cap an operation demands with no grantable counterpart refuses **every** caller, fail-closed and silent, which is indistinguishable from the endpoint simply being broken.

`closedEnum` now refuses each way the declaration can go missing — absent schema, absent property, empty enum — and both the tier and the scope pairings read through it.

I extended it to the `tier` pairing deliberately. That one carried the same weakness before this branch, and making its twin strict while leaving it lax would have left the reason unstated three lines apart.

Mutation-verified: removing the `enum` from `IssuePassportRequest.scopes.items` now fails generation with `components.schemas.IssuePassportRequest.scopes declares no enum — the vocabulary must be closed to be checkable against AgentAdmissionPolicy`. Before the fix, that same edit generated 248 policies and said nothing.

`make check-backend` EXIT 0, `make craft-static` PASS 0/0/0.

Issue #512, opened 2026-08-06
The data reset's five stores are not proven to compose (the HTTP door now is)


dungbanguyen:
`POST /v1/admin/reset-data` now spans five stores: Postgres rows, `river_job`, Redis streams and keys, object storage, and every running process's in-memory caches. Each surface has its own tests and the orchestration has unit tests over fake purgers, but nothing exercises a real reset through HTTP against real infrastructure, so nothing proves the surfaces compose.

An end-to-end test existed and was reverted before merge at the requester's direction. What it had covered, for whoever restores it:

- a queued `river_job` row, a bus entry XADDed to a catalog stream, and a stored object all seeded, then all gone after the reset
- every `kevents.Groups()` consumer group still present afterwards (a purge that left them destroyed would wedge every subscriber — the `NOGROUP` recovery on that same change exists for exactly this)
- a sibling tenant's object untouched, proving the blob sweep is prefix-scoped
- the `audit_log` evidence read back from Postgres and compared against the response counts

It was mutation-tested against six broken purges, three of them production-code mutations.

Two gaps it never closed either, worth folding in when it returns: nothing drives `drain_timed_out` true end to end, and nothing asserts `QueuePause` actually paused (no running job, no `river_queue` read). Both behaviours are covered by lower-level tests today.

Context is in STATUS.md under "the data reset has no end-to-end proof, by request".

dungbanguyen:
Retitled: an end-to-end test now exists, so the old title read as done when the thing it was filed about is not.

**What landed** — `internal/compose/integration/datareset_http_e2e_integration_test.go` (`TestResetDataOverHTTP`, 59 lines) proves the transport half: the `WithDataReset` / `WithNonProduction` wiring, the `non_production` posture `/me` carries for the client gate, a wrong confirmation refused with 422 before anything is deleted, the org-name confirmation answering 200 with `tables_cleared > 0`, and — per its own comment — that the live session path populates the admin `RoleKeys` `RequireAdmin` gates on.

**What this issue asked for, still absent** — every bullet of the reverted test:

| asked for | on `main` (2ffbdb1f) |
|---|---|
| a queued `river_job` row, a bus entry XADDed to a catalog stream and a stored object seeded, then all gone after the reset | absent — none of the three is seeded |
| every `kevents.Groups()` consumer group still present afterwards | absent |
| a sibling tenant's object untouched, proving the blob sweep is prefix-scoped | absent |
| `audit_log` evidence read back and compared against the response counts | absent |
[...]

Issue #511, opened 2026-08-06
Parallel dev stacks share one Redis database, so their event buses cross-talk


dungbanguyen:
`--redis` takes only `host:port` — no database index, no key prefix — so every `DEV_SLUG` stack connects to db 0 and shares the `gw:events:crm:*` streams. Their relays XADD into the same streams and their subscribers consume each other's events, which is an isolation break that predates the data-reset work.

It became visible while adding the reset's bus purge: that purge works from a declared key inventory (the stream catalog, `gw:dedupe:*`, `ovb:<workspace>:*`) and never `FLUSHDB`, so its blast radius is the shared namespace rather than the whole instance — but a reset in one stack still clears another stack's bus.

Options: a `--redis-db` flag, or an installation-bound key prefix. The second also makes the reset's installation-wide cut precise rather than exact-by-coincidence.

Found while implementing the reset's runtime purge; not fixed there because the isolation break is older and wider than that change.

LarsGradion:
Fixed — closing. PR #2456 (`688ec4543`) took the first option this issue names: `--redis` now accepts a logical database index as a `host:port/N` suffix (`events.ClientOptions`), and `scripts/dev.sh` claims one per `DEV_SLUG` stack under a lock, recorded as `REDIS_DB` so a restart reclaims the same index.

Redis serves 80 databases in three blocks that cannot overlap: `0` for bare `make dev`, `1–63` for the integration lane (one per test package, and its suites `FLUSHDB`), `64–79` for slugged stacks. The index is claimed rather than hashed, because hashing collided for two slugs while their ports still differed — reachable with two stacks, not sixteen.

A unix-socket address is still parsed as an address, not a suffixed host, so a deployment on `/var/run/redis.sock` is unaffected.

Verified with two simultaneous slugged stacks on dbs 64 and 65: a €10,000 deal and a €20,000 deal, both at 25%, accrued €2,500 and €5,000 in about three seconds with neither taking the other's event. That is the scenario that lost an event and cost an hour of misdiagnosis (#2428).

The related reset concern in #512 stays open on its own terms.

PR #513, opened 2026-08-06
Reset the runtime, not just the rows

dungbanguyen:
## What

`POST /admin/reset-data` swept Postgres rows and left the runtime around them
intact: queued `river_job` rows, XADDed bus entries and their consumer PELs,
dedupe marks and overlay-budget counters, every running process's in-memory
caches, the object bytes under the workspace prefix, and the sealed
credentials those rows referenced. A reset installation was therefore not a
first-boot installation — queued work went on executing against records that
no longer existed.

The reset now quiets the fleet, purges the queue and the bus, sweeps and
reseeds Postgres, returns an overlay installation to native, clears the budget
counters, the stored objects and the vault secrets, resumes, and announces a
cache flush to every process.

```
gates (env 404 → human → admin) → decode → verify confirmation
QueuePause("*")        River's all-queues literal — DB-mediated, so the api
                       quiets the queues the WORKER process owns
drain, bounded 10s     a timeout never fails the reset; it is reported
clear event_outbox     first, so the relay has nothing left to ship
purge river_job        this workspace's rows + NULL-args fleet dispatchers
purge Redis            streams (+ groups re-created), dedupe marks, ovb:<ws>:*
DB tx                  collect vault refs → sweep → revert to native mode →
                       reseed → audit, with every count
drop cf_* columns, purge the blob prefix <ws>/, redeem the vault refs
ResumeQueues           deferred, detached context, fires after everything
PUBLISH gw:control:reset
```

## Notes for review

- **Purges run before the DB transaction**, so a mid-purge failure leaves a
  safe partial state — queue and bus clear, data intact — that a re-run
  recovers. The object and credential purges are the exception: neither can
  join the transaction, so they run post-commit.
- **The queue resume has exactly one owner**, registered before anything can
  pause and firing after every post-commit purge, on `context.WithoutCancel`
  with its own bound — an operator whose client disconnects mid-reset must not
  leave the fleet paused.
- **The control channel is unauthenticated by nature** (Redis pub/sub, no
  signature), so what it can reach is deliberately limited to cache drops.
  Clearing the auth lockout buckets — brute-force brakes — belongs only to the
  gated path. `backend/resetflushscope_test.go` pins that split, because the
  two flushes read almost identically and merging them back would restore the
  hole silently.
- **Overlay mode is reverted through `overlay.RevertToNative`**, the module
  that owns those fork-owned columns, which `Disconnect` now shares. It is not
  that teardown: the connection and mirror rows are already gone with the
  sweep, and `incumbent.disconnected` would be staged into an outbox this
  reset just drained.
- **`vault_secret` has no `workspace_id` by design** (the tenant is inside the
  ref and the AEAD's AAD), so the handles are collected inside the sweep's
  transaction, before the rows naming them go, and redeemed after it commits.
  Which tables hold one is derived from the catalog on `credential_ref`.
- **Contract drift, deliberate:** the 200 body gains five required fields.
  `specs/contract/crm.yaml` upstream is the source of truth (P3) and needs
  reconciliation; the spec repo was not edited from here. The response schema
  is inline, so no Go type is generated for it —
  `backend/resetwireshape_test.go` compares the contract's `required` list
  against the response struct's json tags to keep the two from drifting.

## Testing

Unit and integration coverage per surface (queue, bus, budget, blobstore,
limiters, vault, overlay-mode revert) plus the orchestration's ordering and
failure semantics.

**There is no end-to-end test.** One existed on this branch and was reverted
before merge at the requester's direction; #512 records what it covered. #511
records that parallel `DEV_SLUG` stacks share one Redis database.

## Summary by CodeRabbit

- **New Features**
  - Expanded non-production data reset to clear jobs, event streams, Redis counters, workspace objects, credentials, caches, and staged events.
  - Reset now restores overlay workspaces to native mode and reports detailed cleanup totals.
  - Added warnings when background jobs remain active after the drain window.
  - Added reset notifications and cache invalidation across running services.

- **Bug Fixes**
  - Ensured queues resume after reset failures, timeouts, cancellations, or unexpected errors.
  - Improved protection against deleting unrelated workspace data.

- **Documentation**
  - Documented reset behavior, limits, reporting, and partial-failure handling.


dungbanguyen:
## Review triage

Thanks — several of these were real, including one the branch's own security fix had falsified. Fixed in `9c72dffa`:

**Correctness**
- **Unset workspace GUC silently no-ops the mode revert.** `current_setting(…, true)` resolved to NULL, matched no row, and returned the same `(false, nil)` an already-native workspace does. Reads it strictly now, so a reset outside a bound transaction errors rather than reporting a revert it never made.
- **Post-commit work rode the request context.** A client disconnecting after the sweep committed could cancel the object and credential purges and the announcement, leaving bytes and sealed secrets behind for a reset the database had recorded as done. They now run detached under their own bound, as the deferred resume already did.
- **`PurgeStreams` used `DEL`.** Switched to `UNLINK`, for the reason `PurgeDedupe` already gives.

**Documentation**
- The reference doc still said the announcement drops auth lockout buckets in other processes. It does not, deliberately — that channel is unauthenticated, so it clears caches only. True before the split, false after; corrected.
- STATUS: dropped the reviewer narration, and made the contract sentence say what it means — `backend/api/crm.yaml` is this repo's authoritative contract, the spec repo is the normative upstream that needs reconciling to it.

**Tests that could not fail**
- The cross-tenant credential test sealed a foreign ref but attached it to no row, so any collector would skip it. It now hangs off a foreign workspace's connection row — the only arrangement that distinguishes correct scoping from none.
- Two bus tests reported from unjoined goroutines (Go turns that into an opaque panic, not a failure): the result comes back on a channel and cleanup waits for it, with the `<-ready` receive bounded.
- Frontend summary assertion covers all five counters; limiter no-op test covers all four buckets; an inverted failure message now says what it means.

### Deferred, with reasons

**[#515](https://github.com/gradionhq/margince-poc-v1/issues/515) — the reset does not serialize against concurrent writers.** Three findings are one gap: credential refs committed after the collection snapshot, unconditional `ResumeQueues` under overlapping resets, and a drain timeout letting the sweep race a live worker's locks. The fix is an installation-wide reset lock with an owner token — a design change, not a correction. The endpoint's exposure is already narrow (non-production, human-only, admin-only, typed confirmation). The same issue records the **versioned-S3-bucket** gap in `DeletePrefix` and the **one-shot `SubscribeReset`** needing reconnect/backoff.

**Accepted limitations, stated rather than fixed:**
- *Wire-shape gate compares names, not types.* True. A type change with unchanged names would pass. Names are the drift that actually happens when a field is added or renamed; a full schema comparison is worth more than it costs only if types start moving.
- *`resetflushscope` AST gate matches a direct call.* True — reaching `ResetRateLimits` through a helper or method expression would evade it. The realistic regression is someone folding the two flushes back together, which is exactly what it catches.
- *`/admin/reset-data`'s bespoke 404 rather than the shared `NotFound`.* Deliberate: the endpoint must not reveal that it exists outside non-production, and the wording carries that. Pre-existing, not introduced here.
- *Overlay-budget purge test seeds one key shape.* The prefix scan is shape-independent, and the batching path is covered by `PurgeDedupe`'s own mid-scan test.
- *Real clock in the subscriber-recovery test.* It bounds a wait against real Redis; injecting a clock there would fake the thing under test.
- *`quiesceTestPool` duplicates `migratedAppPool`.* Fair — worth folding together next time either is touched.

dungbanguyen:
## Second round

**Fixed in `d3f7b7ae`:**

- **The bounded ready-wait introduced a deadlock.** Correct catch, and worse than what it replaced: the select's `case err := <-done` consumed the goroutine's only value, so cleanup's `<-done` blocked forever on an empty channel. Cleanup is now the sole reader and is itself bounded; a subscription that fails outright never closes `ready`, so it lands in the ready timeout and cleanup reports the real error.
- **`resetFinishTimeout` moved to the file that uses it.** Not dead code — it is wired into the post-commit path via `context.WithTimeout` in `datareset.go:152` — but the constant sat in `datareset_runtime.go` while its comment described work in the other file. Moved beside its use.

**Not a defect — `datareset.go:145`, the deferred resume:**

`resumeResetQueues` takes the request context but does not run on it. It detaches internally, `datareset_runtime.go:148`:

```go
resumeCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), resetResumeTimeout)
```

A client disconnect therefore cannot leave the fleet paused — that was fixed in an earlier round, and the detachment lives inside the function rather than at the call site. The finding reads the call site only. No change.

PR #520, opened 2026-08-06
feat(settings): a setting is a row, and adding one costs no migration (ADR-0090/A135)

dungbanguyen:
First implementation step of **ADR-0090 / A135** (ratified upstream in margince-foundation#1253). Installation settings move off columns on `workspace` into one `setting` table, with the catalog in typed Go.

## Why

Adding a setting costs DDL today. `0121_capture_auto_enrich` is an `ALTER TABLE workspace ADD COLUMN` **plus** a hand-written backfill of the `capture_settings` RBAC object into every seeded workspace's role documents — for one boolean. Every future toggle pays that.

## What this keeps

Nothing the column form governed is given up. An `Entry` declares its own **RBAC object, audit verb, default, validator** and optional **freeze probe**, so per-setting audit verbs stay per-setting and `capture_settings` still gates the value it always gated.

What the schema stops enforcing, three fitness gates take over:

1. every key unique, well-formed and governed (an entry with no object or no audit verb fails the build);
2. every key's module prefix names the module that declares it — the table-ownership obligation, in a second place;
3. **no key settable in both the registry and `margince.yaml`'s runtime surface** — ADR-0061 §2's *"no key exists in both surfaces"* promise, enforced for the first time.

Gate 3 is derived from the config **struct** by reflection, not from `margince.example.yaml`. I wrote it against the example file first and it passed **vacuously** — the template has no `capture:` section at all, so it was comparing against nothing. Against the struct it has teeth (`capture.freemail_extra` and `organization.base_currency` both resolve).

## Structure

The catalog is assembled in `compose`, the way every cross-module edge is; each module declares the settings it owns and platform never learns what a currency is. Cross-module reads ride `shared/ports/settings`, shaped like `ports/fieldcatalog`, so a reader never imports a sibling.

`setting` is **non-tenant** — no `workspace_id`, no RLS — on the footing `embed_store_binding` already has. One installation serves one organization (ADR-0061 §3) and ADR-0091 retires the tenant boundary outright, so the table is built without it rather than acquiring a column now to drop it then.

That has a consequence worth reviewing: **with no RLS beneath it, the object gate at the writer is the only control.** So the settings store is brought *under* `rbacgate_test.go` (roots widened to include `internal/platform/settings`) rather than waived out of it — the gate's own doc says bringing a tier under it needs its own evidence, and the evidence is that both entry points take `auth.Require`.

## The migration

`capture.auto_enrich` is the first setting moved, end to end. 0189 creates the table and carries the value across — **written only when it differs from the registered default**, so an installation that turned auto-enrich off does not get it back, and a row that merely restates the default never appears (which keeps "has anyone ever changed this?" answerable).

The column is deliberately **not** dropped. Readers switch first; the drop follows once nothing reads it, which keeps this migration reversible without a data-restoring `down`.

## Two waivers moved, not invented

- `capture:workspace` is **gone** — capture no longer writes another module's table at all.
- `capture:Update` becomes `platform/settings:Set`, reason generalized: a settings write is audit-only under EVT-NOEVT-3, and now every setting inherits that posture instead of arguing it again.

## Verification — please read

- `make check-backend` **green**
- `make craft-static` **green** (0 blocker / 0 major / 0 minor)
- New unit tests for the mechanism: validation refusals, the registry, and the canonical no-op comparison
- **The integration lane was NOT run locally — Docker is not running on this machine.** Migration 0189 and the capture-settings integration suite are exercised for the first time **in CI**. That is the part of this PR I have the least evidence for; if CI's integration shards go red, the migration is where I'd look first.

## Two bugs I introduced and caught in review

Recording them because both were the kind that pass every gate:

1. **The empty-PATCH gate regression.** My first `Update` took only the *read* gate when the patch was empty, so a read-only role would have got a 200 where the column form gave 403 — and my comment *rationalized* it. The update gate is now taken unconditionally, before the branch.
2. **Byte-comparing JSON for the no-op check.** `next` is encoded by Go, `before` comes back from Postgres which normalizes `jsonb` on its own terms. Fine for a scalar; for the composite settings coming in the next step, every write would have looked like a change and written an audit row saying nothing happened. Now compared canonically through the entry's own type, with a test.

## Not in this PR

The other six columns (`name`, `slug`, `base_currency`, `timezone`, `x_sor_mode`, `x_incumbent`) and the column drops. `base_currency` needs the freeze probe wired to ADR-0085 §7's predicate; the overlay pair needs the composite-value handling. Those are the next steps, and the mechanism here is built for them.

One known follow-up for that work: the disjointness gate keys `bootstrapSections` on the *setting's* prefix, so an `installation.*` setting seeded from the `organization:` config section will pass for the right outcome by the wrong reasoning. Worth making that correspondence explicit when `installation.*` lands.

---
## Summary by cubic
Moved installation and capture settings into row-backed entries in the `setting` table with a typed catalog in `internal/platform/settings`. Added `/installation/settings` API and an Installation tab; base currency locks after conversions with a counted reason, and writes mirror to `workspace` columns until readers move.

- **New Features**
  - Centralized, typed settings with per-entry validation, RBAC, audit-only writes, optional freeze probes, and idempotent updates; adding a setting no longer needs DDL.
  - New `/installation/settings` API and UI; all roles can read, admin/ops (human-only) can patch; includes base-currency lock flag and reason; governed by `installation_settings`.
  - Bootstrap seeds `installation.*` rows from the organization created at install so first boot reads configured name/timezone/base currency; capture and workers now read via `NewSettingsStore`.

- **Bug Fixes**
  - PATCH updates the setting row and mirrors to `workspace` columns in one transaction; per-key advisory locks guard concurrent writes and stale audits.
  - Timezone validation rejects empty and `Local`; GET open to agents; Installation tab follows the `installation_settings` grant.
  - Data reset deletes configuration setting rows to restore defaults but keeps installation identity; integration tests cover defaults, lock, mirroring, and RBAC.
  - Migrations renumbered to `0190_setting` and `0191_installation_settings`; backfill/seeding run only when exactly one live workspace exists; down migrations carry current values back to `workspace` columns.

<sup>Written for commit 8e7d8f36d6ee944983fc6a01f2ec2eb18f8d0bcc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/520?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added an Installation settings tab for organization name, reporting timezone, and base currency.
  * Added centralized, validated, audited settings management.
  * Added `capture.auto_enrich` configuration with access controls and migration of existing preferences.
  * Base currency is automatically locked after conversion activity, with an explanation shown to users.
* **Bug Fixes**
  * Improved consistency when reading and updating settings.
* **Tests**
  * Added coverage for permissions, validation, migration, auditing, and locked settings.


dungbanguyen:
Applied the review round. Six confirmed findings — the two reviewers converged independently on the first, and it was about to make CI red.

**The integration tests toggled a column production no longer reads.** `capturedomaintriage_integration_test.go` and `deepread_integration_test.go` still wrote `workspace.capture_auto_enrich`. So `TestDeepReadCancelsAnAutoEnrichJobWhenTheSettingWentOff` would have **failed**, and `TestTriageOnCaptureRespectsTheSettingBeforeSpendingAnything` would have **passed vacuously** — the auto-enrich kill switch reporting green with its "don't crawl" branch never taken. Both now write through the settings store. Straight rule-1 miss: I moved the readers and left their sibling writers behind.

**`rbacgate` could not see the write path, and my comment said it could.** `isStoreEntryPoint` matches methods with a `*Store` receiver; `Set`/`Get` were package-level generic functions (Go forbids generic methods), so the gate collected only `Raw` while the comment claimed it covered *"the one store whose gate has no backstop"*. The entry points are now real methods — `Raw`, `SetRaw` — with the generic helpers as thin wrappers. Verified by the reviewer's own repro: delete the `auth.Require` from `SetRaw` and the gate now fails, where `check-backend` previously stayed green.

**`SetRaw` now resolves through the registry** instead of trusting the caller's entry. An entry a module declared but compose never registered was writable while invisible to every catalog gate — and unreadable through `Raw`, which did check.

**The catalog gate now asserts the RBAC object is in the closed object set**, not merely non-empty. An entry declaring `"person"` would have handed every rep holding `person:update` an installation-wide posture, with every gate green.

**Removed four abstractions with no caller** (T3/T8): the whole `shared/ports/settings` package and its three unbound `Installation*` keys, the Key-bound `Define` variant, and the freeze probe with its unreachable branch. They come back with the module that needs them — `base_currency`'s freeze belongs with ADR-0085 §7's predicate, which is the next step.

**The 0189 backfill is scoped to the live workspace.** It scanned every row, so a decommissioned workspace with auto-enrich off could hand its posture to an installation whose live workspace had it on — the mirror of the case it exists to prevent.

Smaller: `yamlPaths` recurses through slice elements so nested config sections enter the collision set; the audit field key moving `auto_enrich` → `capture.auto_enrich` is recorded where a reader meets it; the column drop is now #521 rather than a bare deferred obligation.

**Recorded, not fixed:** two concurrent settings writes can each record a stale `before` in their audit rows. Same shape as the column form this replaces, so not a regression, and locking a rare admin write isn't worth the complexity today. Also noted for later: nothing stops a future secret-bearing setting landing in `audit_log` — worth a `Redacted` marker on `Entry` before the first such setting exists, but building it now would be the same speculative surface I just removed.

`make check-backend` and `make craft-static` green on a clean run.

dungbanguyen:
All findings from cubic and CodeRabbit are addressed. Summary of what changed, since several were substantive:

**The blocking one.** Roll-ups, FX conversion, quota attainment and the report builder read `workspace.base_currency` and `workspace.timezone` directly — eight files. The PATCH wrote only the setting row, so an operator could change the base currency, see the settings screen confirm it, and have every roll-up keep computing in the old one. `UpdateInstallation` now writes the setting **and** mirrors it onto the column in one transaction, so the two cannot diverge; the mirror retires with the columns (#521). That also collapsed the multi-field patch into a single transaction, which fixes a caveat I had documented rather than solved.

**Concurrency** (raised by CodeRabbit, cubic, and the earlier security pass — I had twice recorded it as not-a-regression, wrongly). `SetRawTx` now takes a transaction-scoped advisory lock on the key before reading `before`. A row lock cannot serve: the row may not exist, and the first write to a setting is exactly when it does not.

**Timezone validation.** `LoadLocation` accepts `""` (→ UTC) and `"Local"` (→ whatever zone the *server* runs in), either of which would silently change what a reporting period means and make one installation compute different boundaries on different hosts. Both refused before the lookup.

**GET security.** Dropped the `cookieAuth`-only restriction — the endpoint's own description says every role may read these values and only the PATCH is human-only, so the narrower scheme contradicted it and excluded agents from a read the RBAC gate already governs.

**Both down migrations** carry the current values back onto the columns before dropping or deleting. After the move the row is the copy a writer updates, so a rollback without this would return whatever the column held before the move.

**Multi-workspace backfill.** First fixed by refusing — which broke `TestRBACBackfillsWriteTheIntendedGrants` and deserved to, because ADR-0061 §3 deliberately keeps the schema able to hold several workspace rows so cross-tenant isolation tests can seed a second. The seed is now conditioned on exactly one live workspace and does nothing otherwise; the setting stays unset and reads as its default, and nothing consumes it before the ambiguity is resolved because the API will not start until it is. Refusing and skipping look similar and are not: one blocks a legal database, the other declines to invent an answer for it.

**Frontend.** The Installation tab now follows the same live `installation_settings:update` grant the card asks for, rather than the admin/ops role name — the two disagreed in both directions. A refused save invalidates the query so the form shows what the server holds. Since the tab is object-gated now it is no longer one of the "objectless" tabs, and the settings-screen tests say so again. Em dash removed from the timezone hint (VOICE-RULE-5), and the eleven keys added to the Vietnamese locale that landed in #526.

Also: the table-ownership header now names the `internal/platform/settings` tier it walks.

**Not fixed, and I would rather name it than bury it:** the base-currency freeze is atomic with its own write but not with the deal transaction that stamps `fx_rate_to_base`, so a conversion committing concurrently with a re-base can still interleave. Closing it needs a lock shared with every FX-freeze path — a larger change than this branch, and worth its own issue rather than a rushed fix here.

Issue #546, opened 2026-08-07
Split internal/compose/integration: one package is the integration lane's long pole


tinnhlt-gra:
## Context

`internal/compose/integration` ran at ~98% of the lane's old 300s per-package timeout and tipped over it — under the concurrency the lane itself creates, and on the unsharded local path (`make test-integration`, `make test-it`) on a loaded machine. That symptom and the timeout budget it argues about are tracked at **#524** (#538 was the same report and is closed as a duplicate of it).

The per-package budget is now 600s and the cost report prints each package's share of it, both landed in #537. That bought headroom; it did not address why one package needs it.

This issue is the other half: why one package is that large in the first place.

## The problem a timeout bump would not solve

`internal/compose/integration` is the largest package in the lane by wall time, and its growth is monotonic — every new cross-module integration suite lands there because that is where the shared harness lives. A single package is also a single scheduling slot: the lane's wall clock bottoms out at this one package no matter how many cores are free, and CI's 12-way shard split works only because it slices *by test*, not by package.

So the cost is paid twice: locally the lane is as slow as this package, and any future growth walks the budget back up to whatever the ceiling is raised to.

## What to do

Split it along the seams the suites already group by, so two or more slots run concurrently instead of one long pole. The shared harness stays where it is and the split packages import it.

Worth checking as part of the split, since they compound:
- #539 — setup forks ~5 Postgres backends per test in this package
- #535 — table-driven cases re-seed Postgres per case
- #536 — some suites here assert pure logic through a booted app

## Done when

No single integration package exceeds roughly half the per-package budget, and the `% of budget` column the lane prints (added in #537) shows it.

dungbanguyen:
Edited the body: it was filed on the premise that #538 had already raised the per-package timeout to 600s and added a `% of budget` column. Neither landed — `main` still has `IT_TIMEOUT="${INTEGRATION_TIMEOUT:-300s}"` with the 900s bump gated on `COVERDIR`, and `scripts/test-integration-one.sh:52` still hardcodes `-timeout=300s`. The timeout and reporting work lives on #524; #538 is closed as a duplicate of it. The split argued for here stands on its own and is not blocked on either.

dungbanguyen:
[...]

Issue #538, opened 2026-08-06
compose/integration runs at ~98% of the integration lane's 300s package timeout


tinnhlt-gra:
## Symptom

`make test-integration` intermittently fails with the lane's own teeth rather than a test assertion:

```
panic: test timed out after 5m0s
	running tests:
		TestAnArchivedOutboundIsNotAConversationToReplyInto (0s)
...
FAIL: the lane ran a different test set than discovery assigned:
  assigned but not run: backend|./internal/compose/integration|TestWorkflow... (22 tests)
```

The named test is innocent — it had just started (`0s`). The package hit `IT_TIMEOUT` mid-run, so everything after it never ran.

## Cause

`./internal/compose/integration` legitimately takes almost exactly the lane's default budget. Measured standalone on one machine, same env the lane gives a package:

| Tree | Package wall time | Margin against the 300s default |
|---|---|---|
| `main` | 296.5s | 3.5s |
| a feature branch | 292.4s | 7.6s |
| inside the parallel lane (25 packages, up to `$JOBS` concurrent) | 302.3s | **timeout** |

So the package passes when run alone and fails under the concurrency the lane itself creates. `INTEGRATION_TIMEOUT=600s make test-integration` passes cleanly (1829 tests, 0 skips).

`IT_TIMEOUT` defaults to `300s` (`scripts/test-integration-parallel.sh:96`) and is only raised to `900s` for single-shard coverage runs.

## Why it matters beyond the flake

The failure is indistinguishable from a real regression at a glance, and it silently removes 22 tests from the run. The lane's contract is that a thinner run must never read as a passing one — that part works — but the cost is that any branch can go red for a reason unrelated to its diff, which trains people to re-run rather than read.

## Options

1. Raise the default `IT_TIMEOUT` (cheapest; hides the growth rather than addressing it).
2. Split `compose/integration` — it is the largest package in the lane and the growth is monotonic; a split also lets two slots run concurrently instead of one long pole.
3. Have the lane report per-package time against the budget, so the margin shrinking is visible before it crosses.

2 and 3 together seem right: the package is a god-suite by wall time, and nothing currently reports how close any package is to the limit.

Found while verifying #537, whose diff was ruled out as the cause by the measurements above.

dungbanguyen:
[...]

PR #552, opened 2026-08-07
fix(test-lane): one budget rule, reachable from both integration lanes

### LINKED ISSUE #524: The unsharded integration lane gives internal/compose/integration a 300s budget it needs ~300s of
## Summary

`internal/compose/integration` runs the full package in 258–302s against a **300s** per-package `go test -timeout`. On a loaded machine it crosses the line, and the lane reports a red that says nothing about the code under test.

Observed on one branch, same commit, three consecutive runs:

| Lane | Wall | Result |
|---|---|---|
| `make test-integration` | 286.82s | pass |
| `make test-integration` | 301.72s | **timeout** |
| `make test-it DIR=backend/internal/compose/integration` | 300.59s | **timeout** |
| `make test-it DIR=backend/internal/compose/integration` | 258.77s | pass |

The failure surfaces as a `go test` panic with a River goroutine dump on top, so the first read is "the job runtime hung", not "the package ran 2s long".

## Why CI does not see it

CI shards the lane 12 ways (`integration shard (n/12)`, ~3m20s–3m46s each), so no shard is near the bound. The exposure is the **unsharded** path: `make test-integration` locally (`SHARD_TOTAL == 1`, whole packages) and `make test-it` on this one package. That is the inner loop, so the flake lands on whoever is iterating rather than on CI.

## Root cause

The 300s default is documented against the sharded assumption, in `scripts/test-integration-parallel.sh`:

```sh
# Per-package go-test timeout. Shard slices stay far under a full package even
# with coverage instrumentation, so one tight default serves both modes.
IT_TIMEOUT="${INTEGRATION_TIMEOUT:-300s}"
if [[ -n "${COVERDIR:-}" && -z "${INTEGRATION_TIMEOUT:-}" ]] && (( SHARD_TOTAL == 1 )); then
  IT_TIMEOUT=900s
fi
```

"Shard slices stay far under a full package" is true, and it is exactly why the default does **not** serve both modes: a slice is not what the unsharded run executes. The script already recognises this — the 900s branch below it bumps precisely the `SHARD_TOTAL == 1` whole-package case — but gates the bump on `COVERDIR`, when the thing that makes the run long is running whole packages, which coverage only adds to.

`scripts/test-integration-one.sh:52` hardcodes `-timeout=300s` with no env override at all, so `make test-it` cannot be nursed past it even deliberately.

## Suggested fix

Make the existing bump depend on the condition it is actually about — drop `COVERDIR` from the `SHARD_TOTAL == 1` branch, so an unsharded run gets the whole-package budget with or without coverage — and give `test-integration-one.sh` the same `INTEGRATION_TIMEOUT` default rather than a literal.

Worth a look while there: 960 tests in one package is the largest by a wide margin (the next is `internal/compose` at 292), which is what puts it alone against the bound. Whether that package should split is a separate question from the timeout, and the timeout should not be the thing that forces it.

## Not this

- #516 — one named test timing out under **shard** load. Different bound, different lane.
- #482 — SQLSTATE 53200 under 29-way parallelism.

Neither covers the unsharded whole-package budget.

dungbanguyen:
## What

`resolve_it_timeout` in `scripts/lib-testdb.sh` — the per-package `go test -timeout` and its validation, in one place, called by both integration lanes.

## Why

#537 moved the parallel lane's budget to 600s and added the `% of budget` column. `scripts/test-integration-one.sh` did not come along:

```sh
go test -p 1 -tags=integration -v -count=1 -timeout=300s …
```

Inline, no variable, no env override. So `make test-it DIR=backend/internal/compose/integration` — the inner-loop command for the one package this whole thread is about — still failed at a bound the lane it belongs to had already moved, and could not be nursed past it even deliberately. That is precisely when someone iterating on that package needs to.

Two entry points ran whole packages against the same suites while disagreeing about what a package is allowed to cost. Rather than copy the new default across, the budget rule moves to the lib both scripts already source. The `SHARD_TOTAL == 1` coverage bump stays in the parallel lane, where the condition it reads lives.

This is the remainder of #524. The budget half landed in #537; this is the sibling call site that got missed — the shape CLAUDE.md's rule 1 is about ("fixed the case under review, missed the sibling copy").

## Verification

13 assertions against the real `resolve_it_timeout`, sourced the way each lane sources it:

```
  ok   default                            0|600s
  ok   explicit 900s                      0|900s
  ok   10m rejected                       1|FAIL: INTEGRATION_TIMEOUT must be <seconds>s
  ok   1h30s rejected                     1|FAIL: INTEGRATION_TIMEOUT must be <seconds>s
  ok   bare number reject                 1|FAIL: INTEGRATION_TIMEOUT must be <seconds>s
  ok   0s rejected                        1|FAIL: INTEGRATION_TIMEOUT must be greater tha
  ok   test-integration-parallel.sh       calls resolve_it_timeout / uses $IT_TIMEOUT
  ok   test-integration-one.sh            calls resolve_it_timeout / uses $IT_TIMEOUT
  ok   no hardcoded -timeout= left in scripts/
```

`bash -n` clean on all three scripts; `make check-backend` green. Behaviour for the parallel lane is unchanged by construction — the same default, the same two guards, the same messages, relocated.

**Not run:** the integration lane itself. It rebuilds the shared migrated template, and a parallel session on this machine is working on that template's staleness logic. No Go changed.

## Note on the comment move

The relocated comment drops its `(issue #538)` reference — history belongs to git, and the invariant reads the same without it.

Closes #524. The package-size half is #546.

### DIFF
--- a/STATUS.md
+++ b/STATUS.md
@@ -47,12 +47,15 @@ that has no meaning without Postgres.
 
 Where the cost actually is, and what is worth doing about it:
 
-- **#524** — `internal/compose/integration` is half the lane (960 tests) and
-  runs 258–302s against a 300s per-package timeout, so it flakes on a loaded
-  machine. CI shards 12 ways and never sees it; the unsharded local lane and
-  `make test-it` do. The 300s default is documented against the sharded
-  assumption, and the script already bumps to 900s for the whole-package case
-  but gates that on `COVERDIR` rather than on `SHARD_TOTAL == 1`.
+- **#524** — **closed.** `internal/compose/integration` is half the lane (960
+  tests) and ran 258–302s against a 300s per-package timeout, so it flaked on a
+  loaded machine. CI shards 12 ways and never saw it; the unsharded local lane
+  and `make test-it` did. #537 raised the budget to 600s and made the cost report
+  print each package's share of it, so a margin shrinking to nothing is visible
+  before it crosses. The one-package lane kept a hardcoded `-timeout=300s` that
+  no variable could reach, which left `make test-it` on this package failing at
+  a bound the lane itself had already moved; both lanes now resolve the budget
+  through one `resolve_it_timeout` in `scripts/lib-testdb.sh`.
 - **#539** — that package's setup forks about five Postgres backends per test.
   Sharing them measured ~18% (170.9s vs a 209.6s baseline) and is **blocked**:
   the harness drops `cf_*` columns between tests, so a pooled connection
--- a/scripts/lib-testdb.sh
+++ b/scripts/lib-testdb.sh
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 # Shared helpers for the parallel integration lanes (test-integration-parallel.sh
 # and test-integration-one.sh): parse this repo's owner + app DSNs, clone/drop a
-# throwaway per-package database, and derive a per-slot MinIO bucket. Source
-# this; don't execute it.
+# throwaway per-package database, derive a per-slot MinIO bucket, and resolve the
+# per-package test timeout both lanes answer to. Source this; don't execute it.
 #
 # This repo's clone-per-package test-DB shape:
 #   - TWO roles, not one — MARGINCE_TEST_DSN (owner: migrates + seeds) and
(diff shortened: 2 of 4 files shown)


dungbanguyen:
Rewritten. This PR originally implemented the 600s bump and the `% of budget` column itself; #537 landed both at 10:56 today while this was in flight, so that work is now redundant and has been dropped. What remains is the piece #537 did not touch: `test-integration-one.sh`, which still hardcoded `-timeout=300s`. Rather than copy the new default into it, the budget rule moves to the lib both lanes already source.

dungbanguyen:
Resolving the cubic thread on `share()`: that function no longer exists. It was part of this PR when the branch implemented the budget column itself; #537 landed an equivalent column while this was in flight, so the implementation here was dropped in the force-push and the report now comes from main unchanged by this PR.

The underlying point is also already met by what main prints: elapsed is its own column (`%8.2fs`), the limit is named in the header (`budget 600s`), and the percentage carries a decimal rather than being rounded to an integer —

```
test-integration-parallel: per-package cost (advisory, budget 600s)
  ./internal/compose/integration                 296.53s    960 tests    308.9 ms/test   49.4% of budget
```

so the margin is visible without a second duration pair on the same row.

PR #553, opened 2026-08-07
feat(people): one authoritative value per company fact, with its evidence beside it (ADR-0085/A130)

LarsGradion:
Gate 0 / slice B1 of the company record page V2 plan: settle the canonical company fields and field-level evidence before any page work depends on them.

Three stores held a company's industry and nothing said which was true — the `organization` column, the profile-field sidecar written by enrichment with a snippet and a confidence, and an overlapping fact value. The same for legal name and address. This slice decides: **for any fact that is a column on `organization`, the column IS the value, and the sidecar holds a proposal plus its evidence, joined at read.**

## What lands

- **Migration 0191.** `organization.linkedin_url` as a validated, live-unique column (identity semantics — matching, dedupe, enrichment — that a governed custom field cannot express). `retrieved_at` / `verified_at` / `verified_by` on both evidence sidecars, with a CHECK pairing the verification to its actor. Deliberately no website column: the primary `organization_domain` row stays canonical and the readable website is derived from it.
- **Five contract operations.** `setBaseCurrency`, `updateOrganizationProfileField`, `confirmOrganizationProfileField`, `updateOrganizationFact`, `confirmOrganizationFact`.
- **A correction changes the record, not just its receipt.** Where a profile field has a canonical `organization` column, both move in one transaction. The machine's proposal is not overwritten — the snippet, source URL and confidence stay on the row, and the audit before-image carries the whole original claim.
- **The workspace base currency becomes a real setting.** Freely correctable until a closed deal has frozen `fx_rate_to_base` against it, locked afterwards, with the frozen count served so a surface can say so before the refusal rather than only on the 422.

## Three defects the new tests found

None of these was reachable by an existing gate; all three are fixed in `9665d557`.

1. **Every correction and confirmation failed with a raw SQL error.** `storekit`'s guarded patch always asked for `archived_at IS NULL`, and neither sidecar has that column — a receipt is deleted with the organization it describes, never retired on its own. `ApplyGuardedIn` / `ApplyWithVersionIn` now carry the archive filter through the lock into the update. That also closes an older gap: a `RowLock` taken with `IncludeArchived` was applied through a liveness clause the lock itself never used.
2. **A fact correction reached the wrong row.** The store located a fact by `value_key` alone, but the contract addresses one as `<field>:<value_key>` and every company fact carries an empty `value_key` by `org_fact_value_key_cardinality`. Correcting the phone number reached whichever of `phone` / `founded_year` / `contact_email` the scan returned first. Both columns are matched now, which `uq_org_fact` and `org_fact_field_vocab` together make exact.
3. **A LinkedIn URL the store accepted, the database refused.** `normalizeOrgLinkedInURL` passed `/company/voltaq/about` through to a `organization_linkedin_url_shape` violation. It now cuts to the company slug, so every tab of a company page normalizes to the one spelling the unique index compares.

## Tests

- `organization_evidence_write_integration_test.go` — the correction moves the column; a field with no column touches none; a confirmation keeps the value and the extraction's evidence; a stale `If-Match` is refused without landing; one single-value fact correction leaves its siblings alone; a multi-value fact is addressed by its key; a malformed key is a 422 and an unknown key a 404; the audit before-image keeps the machine's claim.
- `basecurrency_integration_test.go` — correctable while nothing is frozen; locked once a won deal froze a rate, with the count; a malformed code is refused before anything is written; both verbs need the rate-sheet grant.
- `organizationlinkedin_test.go` — normalization is checked against the column CHECK's own regex, so a value the store accepts can never be one the database then rejects.

Defect 2's test was mutation-checked: reverting the fix fails `TestCorrectingOneSingleValueFactLeavesItsSiblingsAlone`.

## Gates

`make check`, `make craft-static` and the full integration lane are green (`OK: integration passed with 0 skips`, 25 packages).

## Still open in this slice

The plan's remaining Gate 0 items — finance ingestion vs the no-accounting decision, the dossier and growth-fit contracts, account-started email and attachment transport, finance/document RBAC and retention — are **not** in this PR and stay upstream.

Draft until the rest of Gate 0's contract work is decided.

---
## Summary by cubic
Make company facts single-source: `organization` columns are the canonical values, and evidence lives beside them with provenance. Adds LinkedIn as a first-class column, derives website from the primary domain, introduces correction/confirmation endpoints, and makes base currency a real, lockable setting once deals freeze rates.

- **New Features**
  - Canonical fields: for any fact with an `organization` column, the column is the value; sidecars store proposal + evidence with `retrieved_at`, `verified_at`, `verified_by`.
  - New column: `organization.linkedin_url` (normalized, shape-checked, unique among live rows). Website is derived from the primary `organization_domain` (no new column).
  - Corrections: update/confirm profile fields and facts; when a field has a column, the column and receipt update atomically; original extraction evidence is preserved and audited.
  - Currency: `GET /fx-rates` now returns `base_currency`; `PUT /fx-rates/base-currency` sets it until any deal has frozen `fx_rate_to_base`, after which edits return 409 `base_currency_locked` with a blocking count. Same grant as the rate sheet.
  - Migration 0191: adds `organization.linkedin_url` and evidence timestamps/actor to both sidecars.

- **Bug Fixes**
  - `storekit` guarded updates no longer assume `archived_at`; archive filtering is correctly applied through row locks to prevent SQL errors.
  - Fact corrections now match `<field>:<value_key>` on both columns, avoiding writes to the wrong fact row.
  - LinkedIn URL normalization trims to the company slug so accepted values always satisfy the DB shape check.

<sup>Written for commit 9665d55716b838b5174c07c3ca8e4954a858b713. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/553?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


LarsGradion:
Superseded locally: main's PR #520 moved installation.base_currency into the new setting table, so this branch's base-currency commit is being reworked into the freeze probe that identity's entry is waiting for. The branch stays intact and will come back as one PR once the full company-record-page V2 slice is built and tested locally.

Issue #551, opened 2026-08-07
The base-currency freeze is not atomic with the deal FX freeze it guards (ADR-0085 §7)


dungbanguyen:
`installation.base_currency` freezes once a deal has stamped `fx_rate_to_base` (ADR-0085 §7, shipped in #520). The guard is correct in isolation and **is not atomic with the transactions it guards**.

## The window

`identity.InstallationSettingsStore.UpdateInstallation` opens a transaction, takes a per-key advisory lock, runs `deals.BaseCurrencyFrozen` (which counts deals with a non-null `fx_rate_to_base`), and writes. Nothing in the deal FX-freeze paths participates in that lock, so:

1. A re-base transaction reads `count(*) = 0` and proceeds.
2. Concurrently, a deal transaction stamps `fx_rate_to_base` against the **old** base and commits.
3. The re-base commits.

The result is a deal whose frozen rate converts to a base the installation no longer uses — exactly the state ADR-0085 §7 exists to prevent, and it converts silently. Every roll-up built on that deal is then computed on a basis nothing records.

The reverse order is harmless: the freeze probe sees the row and refuses.

## Where the other side would have to participate

The rate is stamped in at least four places in `internal/modules/deals`:

- `deal.go` (`p.Set("fx_rate_to_base", nil, rate)`)
- `deal_advance.go` (same, on the advance path)
- `offer_lifecycle.go` — both the patch at the offer's own freeze and the direct `UPDATE deal SET … fx_rate_to_base = $4 …`

## Why it was not fixed in #520

Closing it means a lock shared between the settings write and every FX-freeze path — a cross-module coordination primitive that does not exist today, and one that has to be taken in a consistent order to avoid trading this race for a deadlock. That is a larger and riskier change than the surface it would protect, and it did not belong in the PR that introduced the surface.

## Shape of a fix, for whoever picks it up

The cheap version is a single advisory lock key (the installation's base-currency setting) taken by both sides: `UpdateInstallation` already takes one per setting key, so the deals side would take the same key before stamping a rate. That serializes a rare admin write against a common deal write, so the cost lands on the wrong side — worth measuring before committing to it.

The alternative is to make the freeze a database constraint rather than a probe: if the base currency lived somewhere a deal's FX write could reference, the check becomes ordinary transactional read-write conflict handling rather than coordination. That composes better with ADR-0091's direction and is probably the right shape, but it depends on where the value lands once the `workspace` columns go (#521).

## Practical severity

[...]

Issue #521, opened 2026-08-06
Drop the remaining workspace columns: slug, x_sor_mode, x_incumbent (ADR-0090/ADR-0091 follow-up)


dungbanguyen:
`0189_setting.up.sql` moved the capture auto-enrich posture into the `setting` table (ADR-0090/A135) and deliberately left `workspace.capture_auto_enrich` in place: readers switch first, the column drop follows, which keeps 0189 reversible without a data-restoring `down`.

Until it is dropped the column is **live-looking but governs nothing**, which has already misled code once — two integration suites kept toggling it after the readers moved, so one test would have failed and one passed vacuously (caught in review on #520).

**To do**
- Confirm no reader remains (`rg capture_auto_enrich` should hit only the migration and this issue).
- Add an additive migration dropping the column, with a `down` that re-adds it `NOT NULL DEFAULT true` — the original 0121 shape.
- Remove the reference to this issue from 0189's comment.

Sensible to batch with the other column drops as the remaining ADR-0090 settings land (`name`, `slug`, `base_currency`, `timezone`, `x_sor_mode`, `x_incumbent`), rather than one migration per column.

dungbanguyen:
Done on \`main\` (2ffbdb1f). The column is dropped by core migration \`0206_drop_capture_auto_enrich_column.up.sql\`, after the readers moved to the \`setting\` row (#572, and the installation-column sweep in #817/#857/#862). What still carries the name is the feature itself — \`modules/capture/autoenrich.go\`'s \`capture_auto_enrich_state\`/\`capture_auto_enrich_budget\` tables — not the workspace column this issue was about.

Closing as delivered.

dungbanguyen:
Reopening — I closed this too early, and the tree says so in four places.

The headline ask **is** delivered: `0206_drop_capture_auto_enrich_column.up.sql` drops the column (carrying an OFF posture forward into `setting` first, for the multi-workspace population 0190 skipped), and `TestRollingBackTheAutoEnrichColumnRestoresTheCurrentPosture` covers the data-carrying reverse. The two remaining `capture_auto_enrich*` names in the tree are the `_state` and `_budget` tables, which are a different thing.

But this issue's last line scoped it wider — *"Sensible to batch with the other column drops … (`name`, `slug`, `base_currency`, `timezone`, `x_sor_mode`, `x_incumbent`)"* — and four shipped migrations now name **#521** as the tracker for exactly that:

[...]

PR #555, opened 2026-08-07
docs(status): record where ADR-0090 landed and where ADR-0091 picks up

dungbanguyen:
Docs only. STATUS.md was never updated for the settings work, and it is the file the next session reads first.

Two open items and one pickup point:

**The settings mirror is a dual-write on ADR-0091's critical path.** `UpdateInstallation` writes the setting *and* mirrors it onto the `workspace` column, because roll-ups, FX conversion, quota attainment and the report builder still read the column — eight files. That is not a settings leftover: ADR-0091 phase 4 drops the row those readers use, so moving them is that phase's first step. Tracked as #521.

**#551** — the base-currency freeze is atomic with its own write but not with the deal transaction that stamps `fx_rate_to_base`.

**Pickup point for ADR-0091/A136**, including the reminder that §9's sequencing is binding — the Go plumbing collapses while RLS is still armed, because the tenant-isolation suite is the only mechanical proof an 887-call-site edit stayed faithful, and the schema phase deletes it.

Also carried forward: #520 lived long enough for main to move under it five times — two migration-number collisions, a semantic conflict with #523 that it would have silently regressed, a new locale, and a dependency bump. **Four of the five were invisible to the branch's own gates.** Phase 2 is larger again, so the note says to slice it into PRs that land in hours.

---
## Summary by cubic
Updates STATUS.md to mark ADR‑0090 shipped and define ADR‑0091’s pickup with concrete next steps and risks. Clarifies phase‑2 magnitude without stale counts.

- `UpdateInstallation` now dual‑writes (settings row + `workspace` `base_currency`/`timezone`) as ADR‑0091 phase‑4 prep; move remaining readers, then drop columns — tracked in #521.
- Base‑currency freeze isn’t atomic with deal writes that stamp `fx_rate_to_base` — tracked in #551.
- Not migrated yet: `slug` (likely drop) and overlay `x_sor_mode`/`x_incumbent` (needs composite value).
- Withdraws process‑wide installation fallback (#557): no consumer (`identity/middleware.go` binds; `cmd/worker` never binds); two bugs (ctx/GUC divergence via `WithWorkspaceTx` + `storekit.MustWorkspace`; global leak across `compose/integration`).
- Guidance: fleet loops should resolve `identity.InstallationWorkspace` once under a singleton and stop enumerating; no global fallback.
- ADR‑0091 §9 sequencing is binding; phase 2 is roughly nine hundred `WithWorkspaceTx` sites across a couple hundred files — avoid precise counts and land in same‑day slices to prevent staleness and hidden conflicts.

<sup>Written for commit 1f6bfa0b0f032031d3b6fd5014636fe35be43997. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/555?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/STATUS.md
+++ b/STATUS.md
@@ -55,7 +55,7 @@ a boot-resolved singleton when the context carries none. It has no consumer:
 paths included, before the `isPublicRequest` branch — so in `cmd/api` nothing
 reaches the database unbound. And `cmd/worker` never calls `EnsureInstallation`,
 so there the pointer stays nil. The fallback fires nowhere while removing the
-loud `ErrNoWorkspace` guard from all 910 call sites.
+loud `ErrNoWorkspace` guard from every one of them.
 
 Two things it also got wrong, worth knowing before anyone tries again.
 `WithWorkspaceTx` hands `fn` the ORIGINAL context, so a fallback-bound
@@ -74,10 +74,14 @@ enumerating. No fallback, no guard removed, one loop per PR.
 
 The sequencing in ADR-0091 §9 is **binding, not advisory**: the Go plumbing
 collapses while RLS is still armed, because the tenant-isolation suite staying
-green is the only mechanical proof that an 887-call-site edit stayed faithful,
+green is the only mechanical proof that an edit of that size stayed faithful,
 and the schema phase deletes that suite. Do not reorder it.
 
-Phase 2 spans 910 `WithWorkspaceTx` call sites across 378 files. Land it in
+Phase 2 spans roughly nine hundred `WithWorkspaceTx` occurrences, a bit over
+four hundred of them outside tests, across a couple of hundred non-test files.
+Deliberately not a precise count: it moved by eleven while this note was being
+written, so an exact figure here would be wrong by the time anyone read it and
+would invite arguing with the number instead of the shape. Land the work in
 slices small enough to merge the day they are written: a branch open longer than
 that accumulates conflicts with `main` that its own gates cannot see — migration
 numbers, locale key sets, and semantic overlap with whatever else is in flight.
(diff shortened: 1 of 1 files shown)


dungbanguyen:
Fixed — you were right, and it cites the file's own header: STATUS.md carries open work, and session narrative belongs in STATUS-ARCHIVE or git history.

Recounting how often `main` moved under #520 was narrative. What survives is the durable instruction: phase 2 spans 910 call sites across 378 files, so land it in slices small enough to merge the day they are written, because a longer-lived branch accumulates conflicts its own gates cannot see. That stands on its own without the history that taught it.

dungbanguyen:
Fixed, though not quite as diagnosed — and the discrepancy is the more useful finding.

My `910 / 378` pair came from **two different scans**: 910 counted `WithWorkspaceTx` occurrences; 378 counted files containing *either* transaction helper, tests included. Measured cleanly just now: **921 occurrences across 375 files, 439 of them outside tests across 211 non-test files.** Your 374 is close to the total-file figure rather than the non-test one, so neither of us had a consistent pair.

The number also moved by eleven between when I measured it and when you asked. That is the real point: any precise figure in the file the next session reads first will be wrong when they read it, and invites arguing with the decimal instead of the shape. The note now states magnitude — roughly nine hundred, a bit over four hundred outside tests, a couple of hundred non-test files — and says why it is deliberately imprecise. The instruction it exists to carry, land slices in a day, is unaffected.

Also dropped the adjacent `887` you flagged: that was ADR-0091's measurement at an older commit, and restating a spec's snapshot in a status file is the same staleness by another route.

PR #556, opened 2026-08-07
refactor(test): the account-360 suites become their own integration package

dungbanguyen:
First cut of #546. Ten files, 61 tests, out of the long pole and into their own package.

## Why a split and not a bigger budget

`internal/compose/integration` is 960 of the lane's 1828 tests and ~296s of its ~493s. The lane runs `JOBS = min(nproc, 8)` **whole packages** concurrently, so its wall clock is `max(longest package, summed/8)` = `max(296, 62)` = **296s**. The lane is bounded entirely by this one package while seven cores idle. #537's 600s budget bought headroom; it did not move that number, and nothing about a timeout can.

CI is unaffected either way — it slices by test across 12 shards, which is why CI never saw the problem and does not see this fix. This is a local inner-loop change.

## Why org360 first

Of 258 test files, this group was the only large one that is a genuinely closed seam:

| | org360 | capture | overlay | approval |
|---|---|---|---|---|
| files / tests | 10 / **61** | 20 / 57 | 12 / 35 | 9 / 21 |
| rides the booted-app fixture | **0** | 1 | 4 | 2 |
| external symbols needed | **0**\* | 7 | 9 | 8 |

\* after the four helpers below move. Every file rides the already-exported `Env` harness, so this cut needs no app-harness promotion at all — that larger refactor (462 field accesses + 158 call sites) is still ahead for the groups that do ride it, and this PR deliberately does not start it.

## What moved, and why there

Four helpers were declared inside the group but used from outside it, so they could not simply travel with it. They go to the harness, beside the seeding vocabulary they belong to: `LinkToOrg`, `Employ`, `AccountMailDirectedAt`, `AgentWithOrgRead`.

The rep fixture behind the last one lands as **`AccountRepPerms`**, not merged into the existing `RepPerms`. They are different grants — `RepPerms` is person/deal/pipeline; this one adds the organization, its activities, and the tag/list chips. Folding them together would have silently widened every suite that reads `RepPerms` as *a bounded rep*, which is exactly what several of them are asserting.

## Two gate-forced consequences, fixed not waived

- **`harness.go` crossed the 500-line cap.** The row-seeding helpers become `seed.go`. One concept per file, and the seeding vocabulary can now grow without growing the harness. Adding a length waiver for a file I had just grown would have been the wrong direction.
- **`TestEveryIdentityInsertGoesThroughTheChokepoint` enrolls sanctioned mint sites by path**, so the waiver follows the seeding code to its new file. Worth noting that the fitness test named the move within seconds — including that the old waiver now matched nothing. Working as designed.

Also constified `objPerson` / `objActivity`: strict lint applies the full ruleset to moved lines, and `goconst` fired once they were re-seen as new. Same shape as `platform/auth/linkscope.go`'s `tablePerson`.

## Verification

- `make check-backend` green; `make craft-static` 0 findings across backend/extensions/fixtures.
- `go vet -tags=integration ./internal/compose/integration/...` clean for both packages.
- `go list -tags=integration` confirms the lane now sees two packages where it saw one: `integration` (248 test files) and `integration/org360` (10) — a second scheduling slot, picked up by discovery with no script change.

**Not run:** the integration lane itself. It rebuilds the shared migrated template and a parallel session on this machine is working in the same tree; CI's shards are the check that matters for a move like this, since the transformation is name-qualification and file relocation with no logic change.

Next cuts, in the order their seams look cleanest: customfields (5 external symbols), oauth (8), then the app-harness promotion before overlay/telegram/approval can follow.

---
## Summary by cubic
Split the Account 360 integration tests into their own package (`backend/internal/compose/integration/org360`) to add a second scheduling slot and speed up local runs. Moved shared seeders/helpers and tightened fixtures; also fixed a seeding helper for determinism and safety.

- **Refactors**
  - Created `backend/internal/compose/integration/org360` and moved 10 files (61 tests); suites now import `github.com/gradionhq/margince/backend/internal/compose/integration` and alias production `compose/org360` as `org360svc`.
  - Added `seed.go` for raw-SQL helpers (`DealFixture`, `SeedStakeholder`, `LinkActivity`, `LinkToOrg`, `AccountMailDirectedAt`, `SeedRow`); harness keeps store-mediated helpers.
  - Promoted `AgentWithOrgRead` and added `AccountRepPerms`; confined `objPerson`/`objActivity` to permission fixtures; `AgentWithOrgRead` now deep-copies its grant map.
  - Kept `employ` in the `org360` package; clarified when a suite group earns its own package.

- **Bug Fixes**
  - One timeout rule for whole-package runs via `resolve_it_timeout` (600s) used by both integration test lanes.
  - `SeedStakeholder` now pins `created_at` and binds `direction` to avoid clock flakiness and SQL injection; identity-mint waiver path updated to `internal/compose/integration/seed.go`.
  - Reverted an accidental `Employ` export; `orgask` tests now use `AgentWithOrgRead`.

<sup>Written for commit d94d71c0c9e594b9810069f47f9f5f5860c8b645. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/556?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Test Improvements**
  - Expanded integration-test coverage for account timelines, graphs, permissions, suggestions, visits, signals, and related scenarios.
  - Added shared test fixtures for deals, stakeholders, activities, employment, and directed account mail.

- **Reliability**
  - Increased the default integration-test timeout from 300 to 600 seconds.
  - Added configurable timeout validation through `INTEGRATION_TIMEOUT`.
  - Improved integration-test cost reporting.

### DIFF
--- a/backend/dedupespine_test.go
+++ b/backend/dedupespine_test.go
@@ -60,7 +60,7 @@ var sanctionedMintSites = gatekit.Waive(map[string]string{
 	"internal/modules/people/lead.go": "the direct lead write shape",
 
 	// Test seeding, not a product path.
-	"internal/compose/integration/harness.go": "integration-harness row seeding",
+	"internal/compose/integration/seed.go": "integration-harness row seeding",
 })
 
 func TestEveryIdentityInsertGoesThroughTheChokepoint(t *testing.T) {
--- a/backend/internal/compose/integration/harness.go
+++ b/backend/internal/compose/integration/harness.go
@@ -153,22 +153,47 @@ func OwnerConn(t *testing.T) *pgx.Conn {
 	return conn
 }
 
+// Object names the permission fixtures and the link helpers both spell. Named
+// once so a fixture's grant vocabulary cannot drift from the entity type the
+// seeding helpers write.
+const (
+	objPerson   = "person"
+	objActivity = "activity"
+)
+
 // permissions fixtures mirror the RBAC matrix rows the suites
 // exercise; the seeded JSONB↔these shapes is identity's policy tests.
 var (
 	RepPerms = principal.Permissions{
 		RoleKeys: []string{"rep"},
 		Objects: map[string]principal.ObjectGrant{
-			"person":   {Create: true, Read: true, Update: true},
+			objPerson:  {Create: true, Read: true, Update: true},
 			"deal":     {Create: true, Read: true, Update: true},
 			"pipeline": {Read: true},
+		},
+		RowScope: principal.RowScopeTeam,
+	}
+	// AccountRepPerms is RepPerms widened to what the account sections read —
+	// the organization itself, its activities, and the tag/list chips. Row scope
+	// stays team on purpose: the interesting failures here are row-scope ones,
+	// and an unbounded admin short-circuits every scope clause.
+	AccountRepPerms = principal.Permissions{
+		RoleKeys: []string{"rep"},
+		Objects: map[string]principal.ObjectGrant{
+			"organization": {Read: true},
+			objPerson:      {Create: true, Read: true, Update: true},
+			"deal":         {Create: true, Read: true, Update: true},
+			objActivity:    {Create: true, Read: true, Update: true},
+			"pipeline":     {Read: true},
+			"tag":          {Read: true},
+			"list":         {Read: true},
 		},
 		RowScope: principal.RowScopeTeam,
 	}
 	ReadOnlyPerms = principal.Permissions{
 		RoleKeys: []string{"read_only"},
 		Objects: map[string]principal.ObjectGrant{
-			"person": {Read: true}, "deal": {Read: true}, "pipeline": {Read: true},
+			objPerson: {Read: true}, "deal": {Read: true}, "pipeline": {Read: true},
 		},
 		RowScope: principal.RowScopeAll,
 	}
@@ -181,11 +206,11 @@ var (
 	AdminPerms       = principal.Permissions{
 		RoleKeys: []string{"admin"},
 		Objects: map[string]principal.ObjectGrant{
-			"person":       {Create: true, Read: true, Update: true, Delete: true},
+			objPerson:      {Create: true, Read: true, Update: true, Delete: true},
 			"organization": {Create: true, Read: true, Update: true, Delete: true},
 			"deal":         {Create: true, Read: true, Update: true, Delete: true},
 			"lead":         {Create: true, Read: true, Update: true, Delete: true},
-			"activity":     {Create: true, Read: true, Update: true, Delete: true},
+			objActivity:    {Create: true, Read: true, Update: true, Delete: true},
 			"pipeline":     {Create: true, Read: true, Update: true, Delete: true},
 			// computed_field is read-only for every system role, admin
 			// included (RD-AC-7: no runtime formula-authoring surface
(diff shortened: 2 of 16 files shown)


dungbanguyen:
Review round applied. One real defect, plus a set of comments that claimed invariants the code did not hold.

**The defect: `Employ` was promoted on a false premise.** The analysis that decided which helpers had callers outside the org360 group matched `employ(` textually, and `network_http_integration_test.go` declares an unrelated `(e *env) employ` — a five-argument method that POSTs a relationship over HTTP. The blanket rename exported that method for nothing, and made the package-level helper look like it had a second client. It had none: org360 was its only caller, and org360 already owns a sibling employment seeder. The method is lowercase again, the helper moved back into the group, and `network_http_integration_test.go` is now byte-identical to `main`. `LinkToOrg`, `AccountMailDirectedAt` and `AgentWithOrgRead` were re-checked the same way and are genuine — each has a caller on both sides of the split.

**`objPerson` fused two vocabularies that merely coincide** — an RBAC object key from identity's policy, and the `activity_link.entity_type` value the seeders write. A rename of one would have silently dragged the other. Now confined to the permission fixtures; the seeding side keeps its literal (which drops it under goconst's threshold anyway).

**Three comments overclaimed.** `AccountRepPerms` said it was "RepPerms widened" while being a fixture in its own right. The neighbouring `SchedulerPerms` has long claimed to be "RepPerms plus the activity grant" while actually dropping `deal` and `pipeline` — pre-existing, but this diff touches those lines, so it is corrected. And `seed.go` claimed one-concept-per-file, which the store-mediated seeders still in `harness.go` disprove immediately; it now states the real line (raw SQL vs store-mediated) and that the file is the identity-mint site the dedupe gate sanctions **by path**, which is why an INSERT belongs there and nowhere else in the package.

**`AgentWithOrgRead` now deep-copies the grant map.** A plain struct copy shares it, and the fixture is cross-package now, so a grant added later would widen it for every suite at once — the failure mode where a negative test keeps passing while proving nothing. Same idiom the two sibling helpers in the file already use.

Also: the package doc now says when a suite group earns its own package (that judgement is the point of the arrangement and was nowhere); four comments pointing at `compose/org360` read as self-references from inside a directory now named `org360`; and the production package is imported under an alias so referring to it from a package of the same name is unambiguous.

Both reviewers independently re-derived the relocation as byte-for-byte semantics-preserving by normalising and diffing all ten files, and one ran the moved package against a real database: 61 tests, all pass, zero skips.

Deliberately **not** taken, recorded rather than silently dropped:
- `accountMailAt` is a near-duplicate of `AccountMailDirectedAt` with `"inbound"` hardcoded — a real cleanup, but it predates this PR and is not part of the move.
- 34 copies of `e.As(e.Rep1, []ids.UUID{e.Team1}, integration.AccountRepPerms)` would collapse into a three-line local helper. Worth doing, but it is a rewrite of assertions rather than a relocation, and I want this PR to stay provably a move.
- Renaming the package to `account360` and dropping the `org360_` file prefixes. Coherent with the `Account*` names, but it churns every file again for naming alone; the import alias removes the actual ambiguity.

`make check-backend`, `make craft-static` (0 findings across backend/extensions/fixtures) and `go vet -tags=integration` all green after the fixes.

dungbanguyen:
Addressed all five review threads. Two were mine, three belonged to `main`.

**Why three of them pointed at files this PR should not have touched:** the branch still carried the #524 commit as an ancestor, and `main` took that work as a *squash* — so the merge base stayed behind both and `scripts/lib-testdb.sh`, `scripts/test-integration-*.sh` and `STATUS.md` all showed as new here. The branch is rebased; git recognised the duplicate by patch-id and skipped it, and the diff is now purely the split.

**Fixed here (`seed.go`, genuinely this PR):**

- **`created_at` left to the column default** — the doc promised emails "at the fixed instant 2026-06-01T12:00Z" but the insert pinned only `occurred_at`, so `created_at` carried wall-clock time. That is part of the activity shape a reader sees, and enough to make an assertion about this fixture depend on when the suite ran. Both columns are pinned now, the way the sibling `AccountMailDirectedAt` already pins them. Good catch — it is exactly the real-clock flakiness the house rules forbid, and it survived the move unnoticed because the move was faithful.
- **`direction` interpolated into the SQL** — now bound, which also lets the failure name which direction it was.

The instant moved into a named var so the comment points at the value rather than restating it; they can no longer disagree.

**Fixed on `main` instead, in #561:**

- **The leading-zero timeout.** Reproduced before fixing: `08s` clears the `^[0-9]+s$` check, then `(( 08 == 0 ))` reads it as octal and fails with *value too great for base*. The arithmetic failing makes the `if` false, so the value is **accepted anyway** — the lane runs on the requested budget behind an error about nothing the caller did wrong. `08s` and `09s` trip it; `007s` is octal-legal and passes silently. Now matched as a string rather than evaluated. I took cubic's `^0+$` form over the `10#` variant: it answers a string question as one and avoids arithmetic entirely. The resolver suite gained the four cases that would have caught it (17 assertions, all green).
- **The STATUS.md placement.** Correct and my error — I marked #524 closed and then *grew* its entry, which is the opposite of what that file's header asks. The narrative is in STATUS-ARCHIVE.md now (append-at-top) and the open lane keeps a three-line pointer.

`make check-backend`, `make craft-static` (0 findings) and `go vet -tags=integration` green after the change.

PR #557, opened 2026-08-07
feat(tenancy): every transaction can bind the installation, not just a request that carries it (ADR-0091 §9 step 3)

dungbanguyen:
First slice of **ADR-0091 phase 2** — the tenancy retirement ratified upstream as A136. Deliberately small: the *decision* about which workspace a transaction binds moves, and **not one of the 910 call sites changes**.

## What changes

Today a workspace reaches the database only by riding a request context. A worker loop without one has to bind the GUC itself — which is why the fleet scans exist and why 25 sites carry an `rls-exempt` comment.

`resolveWorkspace` now takes the context's workspace when there is one, and the installation this process serves otherwise. `BindInstallation` names that installation once, at the boot path that already resolved it.

## What it costs

Stated in `database.go` where a reader meets it: a context with no workspace now reads the installation's data instead of failing with `ErrNoWorkspace`. With one tenant there is nothing to leak *across* — the fallback can only bind the id every request already binds — but the loud failure that used to catch a path forgetting its workspace is gone.

ADR-0091 §4 accepts that trade for the schema phase; it arrives here because the fallback is what lets the fleet loops collapse. **RBAC is untouched**: `auth.Require` gates every store entry point regardless of which workspace is bound.

**Pre-bootstrap still fails loudly.** There is no installation to guess at before one exists, and the worker polls exactly that state until the API creates one — so the binding is nil until boot sets it, and a query in that window errors rather than inventing an answer.

## Why `resolveWorkspace` is its own function

Because it is the thing this step changes, and because it is provable without a database: the decision happens before any SQL, which is the whole reason it can be tested at all. Three cases, all of which matter — nothing bound, nothing on the context, and a context that names its own workspace, which **must still win** or every cross-tenant isolation test would stop meaning anything.

## Sequencing

RLS stays armed and the schema is untouched, which is what ADR-0091 §9 requires of this phase: the tenant-isolation suite staying green is the proof the change was faithful, and that suite only exists until the schema phase deletes it.

Sliced this way on purpose. #520 lived long enough for `main` to move under it five times, and four of those collisions were invisible to its own gates. This one is three files.

`make check-backend` and `make craft-static` green. Integration runs first in CI (no Docker locally).

Next slices, each its own PR: collapse the fleet loops and their `rls-exempt` hatches onto the binding; then `WithInfraTx` folds into one `Tx`. #521 — moving the eight readers off `workspace.base_currency`/`timezone` — is phase 4's first step and independent of these.

---
## Summary by cubic
Let transactions bind the boot-resolved installation when no workspace is on the context (ADR-0091 §9 step 3). Requests still bind their own workspace; pre-bootstrap still fails with `ErrNoWorkspace`.

- **Refactors**
  - Added `database.BindInstallation`, called once during boot from `compose.EnsureInstallation`.
  - Introduced `resolveWorkspace` and updated `WithWorkspaceTx` to use it.
  - Added tests for pre-bootstrap failure, installation fallback, and context override.

<sup>Written for commit 700ea1db3ec8178499de2f14436d337dafc2d605. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/557?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/internal/compose/installation.go
+++ b/backend/internal/compose/installation.go
@@ -27,6 +27,7 @@ import (
 	"github.com/gradionhq/margince/backend/internal/modules/consent"
 	"github.com/gradionhq/margince/backend/internal/modules/deals"
 	"github.com/gradionhq/margince/backend/internal/modules/identity"
+	"github.com/gradionhq/margince/backend/internal/platform/database"
 	"github.com/gradionhq/margince/backend/internal/platform/database/storekit"
 	"github.com/gradionhq/margince/backend/internal/platform/deployconfig"
 	"github.com/gradionhq/margince/backend/internal/platform/settings"
@@ -66,6 +67,13 @@ func EnsureInstallation(ctx context.Context, pool *pgxpool.Pool, log *slog.Logge
 	if err != nil {
 		return err
 	}
+	// Name the singleton every transaction in this process binds (ADR-0091 §9
+	// step 3). Until now a workspace reached the database only by riding a
+	// request context; a worker loop without one had to bind the GUC itself.
+	// Set here, at the one point that has resolved the installation and before
+	// any request or job is served.
+	database.BindInstallation(wsID.UUID)
+
 	if created {
 		log.Info("installation bootstrapped", "workspace_id", wsID.String(), "organization", cfg.Organization.Name)
 	} else {
--- a/backend/internal/platform/database/database.go
+++ b/backend/internal/platform/database/database.go
@@ -12,11 +12,13 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	"github.com/jackc/pgx/v5"
 	"github.com/jackc/pgx/v5/pgxpool"
 
+	"github.com/gradionhq/margince/backend/internal/shared/kernel/ids"
 	"github.com/gradionhq/margince/backend/internal/shared/kernel/principal"
 )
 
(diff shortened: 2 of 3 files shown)


dungbanguyen:
Closing this rather than fixing it. The review round found the premise wrong, not the details, and patching six findings onto it would leave the thing that makes it wrong.

**The fallback has zero consumers.** `identity/middleware.go:107` binds the singleton into *every* request context — before the `isPublicRequest` branch — so in `cmd/api` no request ever reaches `WithWorkspaceTx` unbound. And `cmd/worker` never calls `EnsureInstallation`, so there the pointer stays nil for the process's life. The fallback fires on no real path in either role, while removing the loud `ErrNoWorkspace` guard for all 910 call sites. That is exactly the T3/T8 trade the craftsmanship rules forbid: an abstraction with no caller today, bought with a detector that was working.

**And if it ever did fire, it was wrong.** `WithWorkspaceTx` binds the GUC to the resolved id but hands `fn` the *original* context, so `storekit.MustWorkspace(ctx)` would return the zero UUID while the GUC held the installation. The domain row and its `audit_log`/`event_outbox` rows would name different tenants, and `storekit.LockWriteIdentity` would key its advisory lock on the zero workspace — silently degrading the lock whose comment says it exists to stop a concurrent write being overwritten. Two comments in storekit state the invariant this diff removed, verbatim, and I did not grep for them. That is rule 1.

**It also leaked in the test lane.** 155 call sites in `compose/integration` bootstrap through `EnsureInstallation` in one binary, nothing unbinds, and `testdb.Reset` truncates between tests — so from the first bootstrap the global names a workspace row that no longer exists. In the lane that owns the isolation proofs, an unbound context would open a transaction against a dead tenant and return zero rows, which is what an isolation test reads as PASS.

**My tests could not have caught any of it.** All three exercise `resolveWorkspace` only; revert the one line in `WithWorkspaceTx` that the slice exists to change and they stay green.

## What the right slice looks like

The value in ADR-0091 §9 step 3 is in the fleet loops, and they do not need a process-global. They already bind explicitly — by *enumerating* workspaces. Under a singleton the fix is for each to resolve `identity.InstallationWorkspace` once and stop enumerating, which needs no fallback and removes no guard. That is the next attempt, one loop at a time.

Recorded in STATUS.md (#555) so the next session does not rediscover this by rebuilding it.

PR #563, opened 2026-08-07
refactor(test): the search fixture becomes importable, like Env before it

dungbanguyen:
Prerequisite for the next cut of #546. No file moves here — this makes the move possible, and keeps the two kinds of change in separate PRs.

## Why

`searchEnv` was declared in `search_integration_test.go`. Forty suites ride it, but a test file's type is unreachable from another package, so no group that depends on it can be split out. The capture suites are 16 of those forty.

It moves to a non-test file (`searchenv.go`, beside `harness.go`) and exports its name, constructor, and the two lowercase fields. Body unchanged, seeding unchanged, suites that keep riding it keep riding it.

## What the fixture turned out to be

Worth recording, because the name undersells it. Of the forty riders, most want the migrated database and the seeded workspace, not the search store — **the capture suites reference `.Store` exactly zero times**:

| what capture touches on the fixture | refs |
|---|---|
| `.Pool` | 72 |
| `.Rep1` | 61 |
| `.WS` | 38 |
| `.owner` | 9 |
| **`.store`** | **0** |

So this is the package's *second general-purpose fixture*, not a search-specific one — which is why it now sits beside `Env` rather than with the suite that named it. That also rules out the cheaper-looking alternative of re-pointing capture at `Env`: `SetupSearch` seeds a different workspace slug and a different user set, so swapping fixtures would be a behaviour change wearing a refactor's clothes. Renaming the fixture to match what it is can wait for someone with a better name; the export is what unblocks the split.

## How the field renames were made

`owner` and `store` are field names on several structs in this package (`e2e`'s `env`, `customfields`' `store`, `runner`'s `store`), so a blanket rename would have hit the wrong ones — the same class of mistake as the `Employ` collateral in #556.

Instead every site came from the type-checker: compile, take errors of the form `x.store undefined (type *SearchEnv … but does have field Store)`, fix exactly those lines, recompile. **87 sites over twelve rounds.** The loop keys on the exported field *existing*, which only the fixture and structs embedding it satisfy — `reembedFleetEnv` embeds `*SearchEnv` and was picked up correctly by that rule.

## Three lint findings that came with it

Strict lint re-reads a moved line as new, so `unparam` and `revive` got a fresh look at old code. One was cosmetic (the newly-exported constructor wanted a doc comment). The other two were real and are now gone:

- `seedDealFixtures` returned an `orgID` that its one interested caller assigned and then discarded three dozen lines later with `_ = orgID`.
- `enableLeadRouting` returned an automation id that none of its four callers has ever read.

## Verification

`make check-backend` green, `make craft-static` 0 findings across backend/extensions/fixtures, `go vet -tags=integration` clean. The change is a rename and a file move with no logic edits, and the type-checker drove every field site.

**Not run:** the integration lane. It rebuilds the shared migrated template and a parallel session is working in this tree; CI's shards are the check that matters for a rename.

Next: the capture suites move to `internal/compose/integration/capture` — 16 files, ~45 tests, a third lane slot.

---
## Summary by cubic
Made the search test fixture importable so sibling packages can use it, and tightened the migrate-once gate to judge all build-tagged integration files. This unblocks moving capture suites into their own package without changing behavior.

- **Refactors**
  - Moved the fixture from `search_integration_test.go` to `searchenv.go` beside `harness.go`.
  - Exported `SearchEnv`, `SetupSearch`, fields `Owner` and `Store`, and methods `Admin()`, `AsTeamRep(...)`, and `Seed(...)` (plus the shared grant helper), moving these methods into `searchenv.go` so imports actually work.
  - Widened the migrate-once gate to scan build-tagged integration files (not only `_test.go`), updated its wording, and excluded `internal/platform/testdb` by rule.
  - Updated package docs to note both shared harnesses and the broader gate.

- **Migration**
  - Use `integration.SetupSearch(t)` and `*integration.SearchEnv` in tests.
  - Replace: `e.owner`→`e.Owner`, `e.store`→`e.Store`, `e.seed(...)`→`e.Seed(...)`, `e.asTeamRep(...)`→`e.AsTeamRep(...)`, `e.admin()`→`e.Admin()`.

<sup>Written for commit d4ceeb6b1b30693a8b6d6f323de7294cdacbebe2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/563?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/integrationmigrateonce_test.go
+++ b/backend/integrationmigrateonce_test.go
@@ -50,6 +50,13 @@ import (
 // than this gate owns today.
 const migrationsPackage = "migrations"
 
+// testdbPackage implements the migrate-once mechanism every suite is required to
+// ride, so it is where dbmigrate.Up is SUPPOSED to be called. Excluded by rule
+// rather than waived, for the same reason migrationsPackage is: a waiver would
+// read as "this one is allowed to misbehave" when in fact it is the definition of
+// behaving. Only reachable now that this gate walks non-test files too.
+const testdbPackage = "internal/platform/testdb"
+
 // inlineMigrators are the suites outside migrationsPackage ratified to migrate
 // on their own, each bound to what the exception costs.
 var inlineMigrators = gatekit.Waive(map[string]string{
@@ -63,8 +70,17 @@ func TestIntegrationSuitesMigrateOncePerProcess(t *testing.T) {
 	var offenders, inMigrations []string
 	fset := token.NewFileSet()
 	err := filepath.WalkDir(".", func(path string, d fs.DirEntry, err error) error {
-		if err != nil || d.IsDir() || !strings.HasSuffix(path, "_test.go") {
+		if err != nil || d.IsDir() || !strings.HasSuffix(path, ".go") {
 			return err
+		}
+		// Every file the lane compiles, not only the _test.go ones. A shared
+		// fixture that moves into a non-test file so sibling packages can import
+		// it would otherwise walk straight out of this gate's reach while still
+		// being the thing that migrates — and the gate would keep passing,
+		// covering less, saying nothing. Membership is the build tag, which is
+		// what actually decides whether the lane runs the file.
+		if !strings.HasSuffix(path, "_test.go") && !isIntegrationTagged(path) {
+			return nil
 		}
 		path = filepath.ToSlash(path)
 		file, err := parser.ParseFile(fset, path, nil, 0)
@@ -76,6 +92,9 @@ func TestIntegrationSuitesMigrateOncePerProcess(t *testing.T) {
 		}
 		if strings.HasPrefix(path, migrationsPackage+"/") {
 			inMigrations = append(inMigrations, path)
+			return nil
+		}
+		if strings.HasPrefix(path, testdbPackage+"/") {
 			return nil
 		}
 		if !inlineMigrators.Waived(t, path) {
--- a/backend/integrationmigrateonce_test.go
+++ b/backend/integrationmigrateonce_test.go
@@ -3,7 +3,7 @@
 
 package backendarch
 
-// Migrate-once discipline for every integration suite in the module, as a
+// Migrate-once discipline for everything the integration lane compiles, as a
 // fitness function. A suite migrates the schema once per test process
 // (internal/platform/testdb.EnsureSchema) and resets between tests with a fast
 // data-only reset (testdb.Reset); one that instead runs its own DROP SCHEMA +
@@ -13,7 +13,14 @@ package backendarch
 // subtree is where the obligated code lives, and that second claim is the one
 // nothing checks.
 //
-// Any test file that REFERENCES the migrate entry point is caught, not only one
+// Every file the lane COMPILES is judged, not only the _test.go ones: a shared
+// fixture that moves into a build-tagged non-test file so sibling packages can
+// import it is still the thing that migrates, and keying on the filename suffix
+// would let it walk out of reach here while the gate kept passing. Membership is
+// therefore the build tag, which is what actually decides whether the lane
+// compiles a file.
+//
+// Any such file that REFERENCES the migrate entry point is caught, not only one
 // that applies it: `up := dm.Up; up(...)` migrates just as much as `dm.Up(...)`,
 // and a file has no reason to hold the migrator except to run it. That closes the
 // function-value route without type-aware analysis, and the qualifier is resolved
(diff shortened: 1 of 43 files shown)


dungbanguyen:
Review round found that **this PR did not do what it claimed**, and both reviewers caught it independently. Fixed.

## The export was inert

Moving the type and exporting it is not enough. `Seed`, `Admin` and `AsTeamRep` were still declared in `search_integration_test.go`, and **a method declared in a `_test.go` file is not part of the package a sibling imports.** So a split-out group could hold a `*SearchEnv` and still be unable to build a principal context or seed a row. The capture suites this whole thing is for call `Admin` **51 times** — they would have been exactly as stuck after this PR as before it.

The three methods and the grant helper they share now live in `searchenv.go`. Same discipline as the field pass: `seed` is also a method on `overlayWriteEnv`, so all 109 call sites came from compiler errors naming the exported replacement, never a blanket rename.

**The claim is now checked instead of asserted.** I compiled a throwaway sibling package calling `SetupSearch`/`Admin`/`AsTeamRep`/`Seed`, confirmed it builds, and deleted it. It would not have compiled before this change. That check is what I should have run before opening the PR — "importable, like Env before it" was an assertion I had not tested.

## My rename loop had the same bug as the code it was fixing

It reported `round 1: clean` when it found no error it recognised — which is not the same as the package compiling. It said that while `searchenv.go` was missing an import. It now distinguishes "compiles" from "stuck on something else" and prints what it is stuck on. Same silent-success shape as the monitor that would have gone quiet on a repeated CI failure earlier in this session.

## A fitness gate silently lost coverage

`TestIntegrationSuitesMigrateOncePerProcess` walks files by the `_test.go` suffix. Moving the fixture into a non-test file took it **out of that gate's reach while it is still the thing that migrates** — the gate would have kept passing while covering less, which is precisely the failure its own doc comment warns about. It now keys on the build tag, which is what actually decides whether the lane compiles a file; that also re-covers `harness.go` and `seed.go`, which had been invisible to it all along.

`internal/platform/testdb` is excluded **by rule, not waived** — it implements the migrate-once mechanism, so calling `dbmigrate.Up` there is the definition of behaving, not an exception. Mutation-tested: injecting an inline migrate into a non-test tagged file makes the gate fail and name the file; restoring makes it pass.

## Two doc claims were false

`SearchEnv` is not "the second migrated-database fixture" — several suites still build their own; it is the second **exported** one. And the rider count is now dropped rather than stated, since this work exists to change it.

## Deliberately deferred, recorded not dropped

- **Renaming the fixture.** A reviewer argued (fairly, under rule 5) that the comment rationalizes a known misnomer and that renaming gets harder after the split. I kept the name: this change is already a rename of everything else, and one rename per pass is easier to verify than two. The comment now states the misnomer as a fact owed a fix rather than explaining it away.
- **`SetupSearch` duplicates `Setup`'s DSN/connect/EnsureSchema preamble** verbatim. Real, and now two non-test callers, so extraction is warranted — but four more copies exist elsewhere in the package, so it is its own change.
- **`report_graph_integration_test.go:237` logs where it should fail** (`t.Logf` in a branch that reads like an assertion). Pre-existing, adjacent to my edit, not mine to fold in silently.

`make check-backend`, `make lint` (0 issues), `make craft-static` (0 findings) and `go vet -tags=integration` all green.

dungbanguyen:
Applied. The gate's wording now matches what it judges.

Correct finding: I widened the walk to every file the lane compiles, but left the contract note saying *"any **test file** that REFERENCES the migrate entry point"* and the failure calling offenders *"integration **suite(s)**"*. For a fitness function the wording is the only place the obligation is written down, so a doc narrower than the enforcement teaches the next reader the wrong rule — and the next reader is who the gate exists for.

Three places updated: the opening line (now "everything the integration lane compiles"), the contract note (which now states *why* membership is the build tag rather than the filename suffix — a shared fixture moving into a non-test file to become importable is still the thing that migrates), and the failure message (now "integration-lane file(s) … a suite or a shared fixture in a build-tagged non-test file counts alike").

Re-ran the mutation test after the wording change, since a doc edit near a gate is exactly when you stop checking it still bites:

```
baseline:                    PASS (names searchenv.go: False)
with an inline dbmigrate.Up: FAIL (names searchenv.go: True)
restored:                    PASS (names searchenv.go: False)
```

Also rebased onto `main` — the branch had gone `BEHIND` (main took #559), and branch protection here requires the head branch up to date. Worth noting for anyone watching this queue: **auto-merge does not update the branch for you**, so an armed PR can sit green-and-unmergeable indefinitely with nothing saying why. That is what stalled #561 for an hour.

`make check-backend`, `go vet -tags=integration` and the migrate-once gate all green after the rebase.

PR #568, opened 2026-08-07
refactor(test): the capture suites become their own integration package

dungbanguyen:
Second cut of #546, and the payoff for #563: **24 files and 63 tests** out of the long-pole package, which the lane schedules as a third slot.

```
integration          test-files=224
integration/capture  test-files=24   <- new
integration/org360   test-files=10
```

## Why the group is wider than "capture"

Capture alone would have been the *more* expensive cut. Six staying files use `newTestCaptureRegistry`, thirteen use `newTestKeyvault` — so moving capture by itself meant pushing capture internals into the shared harness purely to keep the boundary working. Moving their users too makes both couplings vanish:

| | capture alone | capture + connectors + keyvault |
|---|---|---|
| tests moved | 53 | **63** |
| helpers needing promotion | 4 (incl. one used by 13 files) | **3**, all single-file |
| helpers owed outward | 3 (incl. one used by 6 files) | **2**, both single-file |

Gmail and Calendar *are* capture sources, so the wider group is also the more honest boundary.

## Two structural things org360 didn't hit

**A method cannot be declared on a type from another package.** `humanWithScopes` (45 call sites) and `captureSettingsCtx` were methods on `*SearchEnv` declared in capture files, so they could not travel as methods. They are plain functions taking the fixture now — converted while still in-package, so that step compiled and was verified on its own before the move.

**`asFullUser` was the mirror case, and my analysis missed it**: a method on the shared fixture declared *outside* the group and called *inside* it. My scan only looked for methods declared within the group. The compiler caught it, and I then enumerated the whole class rather than fixing the one — of 17 fixture methods in the parent, only `asFullUser` was needed by the move. It's promoted as `AsFullUser` beside `Admin`, and the object vocabulary both grant sets iterate is now named once: if the read and write sets drifted, a suite comparing a reader against a writer would be comparing two vocabularies and attributing the difference to permissions.

## What stayed behind, and why

`capturedbykind_http_integration_test.go` drives the booted-app fixture in `e2e_integration_test.go`. That fixture is declared in a test file, so it isn't importable — the same inertness #563 fixed for `SearchEnv`. It stays until that one is promoted too. The package doc says so, so the next person doesn't have to rediscover it.

## Promotion judgment

`AwaitKindCompleted` and `WriteJSON` are promoted — suites on both sides need them, and a per-package copy of a no-sleep wait loop is exactly the thing that decays into one that sleeps. `boolPtr` is **not**: it's `return &v`, and the moved package declares its own rather than coupling two packages over an address.

Both split packages gained the package doc they should have had; org360 has been missing one since it landed.

## Verification

- **Test census, checked not assumed:** 968 top-level tests on `main`, 968 across the three packages now. A relocation that quietly drops one looks exactly like a passing run.
- `make check-backend`, `make lint` (0 issues), `make craft-static` (0 findings), `check-test-lanes.sh`, `go vet -tags=integration` — all green.
- Three findings arrived with the moved lines, since strict lint and the craft gate re-read a moved line as new. Two were real and are fixed: `emailWithListUnsub` took a recipient every caller filled with `captureOwner`, implying a degree of freedom the scenarios don't have; and the new package needed a doc comment. `WriteJSON`'s naked `any` is waived with a reason — the value goes straight to `json.Encoder.Encode`, whose own signature is `any`, and no narrower type admits every shape a stub returns.

**Not run:** the integration lane itself — it rebuilds the shared migrated template and a parallel session works in this tree. CI's shards are the check that matters, and for a relocation the compiler plus the census carry most of the weight.

Next, if this lands: the remaining clusters all ride the booted-app fixture, so the app-harness promotion (~660 edits) is the gate on everything after this.

---
## Summary by cubic
Split the capture, connector, and keyvault integration test suites into `backend/internal/compose/integration/capture` to shorten the lane and clarify boundaries. Promoted `AwaitKindCompleted` and `WriteJSON`; no behavior changes and test count unchanged.

- **Refactors**
  - Moved 24 files (63 tests) to `backend/internal/compose/integration/capture`; kept `capturedbykind_http_integration_test.go` in the parent.
  - Promoted shared helpers: `AwaitKindCompleted` (`integration/jobwait.go`) and `WriteJSON` (`integration/httpstub.go`); updated call sites in `embedreindex`, `no_activity_reminder`, `telegram_fixture`, comms send provider, and Gmail/Calendar stubs.
  - API shape fixes: `AsFullUser` is now a `SearchEnv` method; `humanWithScopes` and `captureSettingsCtx` are package-level functions; unified object vocabulary as `searchObjects`.
  - Kept `boolPtr` local to the one test; simplified `emailWithListUnsub` by removing its unused recipient parameter; capture tests now use package `capture` and import `integration` helpers.

- **Documentation**
  - Added package docs for `capture` and `org360`; corrected boundaries by naming shared helpers they import.
  - Updated harness docs to point to subdirectories; clarified `SearchEnv` naming note; moved the `craft:ignore` waiver with `WriteJSON` and removed the stale copy.

<sup>Written for commit f77059ddffc5e340c01fa1313fa5993c7a45e14e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/568?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Tests**
  * Reorganized capture and account-360 integration test suites for clearer separation and more consistent setup.
  * Expanded shared test utilities for JSON responses, job completion waits, full-access test contexts, and optional settings.
  * Preserved existing test coverage and assertions across capture, messaging, lead scoring, search, and connector scenarios.

* **Documentation**
  * Added package documentation describing integration test coverage and conventions.


dungbanguyen:
Review round applied. One shape defect, and a set of docs claiming more than the code does.

**The reviewers ran what I could not.** One of them executed the moved package against a real database — `make test-it DIR=backend/internal/compose/integration/capture` → **63 tests, 63 PASS, 0 SKIP**, plus the affected parent tests. That closes the gap I had flagged in the PR body: I had skipped the integration lane because it rebuilds the shared migrated template a parallel session is working on. Both reviewers also independently re-derived the 24 moved files as pure relocation, and independently confirmed the 968-test census.

**The defect:** `AsFullUser` was a free function taking `*SearchEnv` while its two siblings `Admin` and `AsTeamRep` are methods on it. The function shape was forced only for helpers that had to live in the `capture` package, where a method on an imported type is illegal — `AsFullUser` landed in the fixture's own package, where nothing forced it. Callers read `AsFullUser(e)` next to `e.Admin()` in the same test. It is a method now; three call sites updated.

**An orphaned doc comment** stayed behind when `asFullUser` moved out, and ended up directly above `TestLeadScoreRecomputesFromLinkedActivities` — where godoc and every reader would take it as that test's documentation. Deleted.

**Both package docs claimed each package "owns everything else it needs", and both were false.** `capture` also rides `Env`, `ApplyRiverSchema`, `AwaitKindCompleted`, `WriteJSON`, `AsFullUser`; `org360` rides ten of the parent's seeding and principal helpers. A doc that overstates a boundary is worse than no doc — the next reader trusts it and never looks for the parent import. Both now name what they ride.

**`capture` called itself the second scheduling slot; org360 took that one.** I removed the ordinal rather than correcting it, and removed the matching `(org360 is the first)` enumeration from the package doc, pointing at the subdirectories instead. A list every future split must remember to extend is a list that will be wrong.

**"the capture split"** appeared in two comments — an event, not an invariant, with no referent once it is history. They name the packages now. The `searchenv.go` comment explaining *why* the fixture had not been renamed went the same way: it stated a misnomer and then argued about scheduling. It now just states the misnomer. That one was mine from #563, and leaving it a second time would have compounded it.

**`boolPtr` moved into the one test file that calls it** and `capture/ptr.go` is deleted — a non-test file was putting a test-only one-liner on the importable surface for nothing, and the parent keeps its copy in a test file, so that is the consistent home. The non-promotion decision itself stands.

Two sed artifacts had left the Go qualifier inside English prose ("in the `integration.SearchEnv` workspace"); both read as sentences again.

Re-verified after the fixes: census still 968 CONSERVED, all three packages compile, `make check-backend` and `make craft-static` (0 findings) green.

Recorded, not fixed: the `SearchEnv` name itself. Now that its riders are enumerated, most take it for the database rather than the store, so it is due a rename — but that crosses three packages now and is its own change.

dungbanguyen:
Both applied — and both are the same defect in different clothes: a comment describing code that is no longer under it.

**The orphaned `craft:ignore`.** The `naked-any` waiver for `writeJSON`'s `v any` parameter stayed behind when the function moved to the parent package, and came to rest above `TestGmailConnectorSyncsAnActivity`, which has no such parameter. A suppression pointed at the wrong code is worse than none — it silences a check for a function nobody meant to exempt. The waiver travelled with the function; only the stale copy is deleted. Good catch: this is the third orphaned comment this thread has produced, all from the same mechanical cause.

**`AsFullUser`'s missing `Delete`.** Correct that the doc and the grants disagreed. I fixed the **doc**, not the grants, and the reasoning matters:

- The grants are the original fixture's. Adding `Delete` would make this a behaviour change rather than a relocation, in a PR whose whole claim is that it relocates.
- Nothing riding the fixture deletes. A grant no caller needs is exactly how a future erasure test becomes a pass instead of a proof — the failure mode this repo names explicitly, and one I would rather not plant.

So the comment now says create/read/update, says `Delete` is deliberately absent, and says why — including that `AdminPerms` does grant it, so a reader comparing the two does not have to wonder which is the mistake. A suite that genuinely needs delete should ask for it rather than inherit it from a fixture called "full".

`make check-backend`, `make craft-static` (0 findings) and `go vet -tags=integration` green after both.

PR #572, opened 2026-08-07
feat(settings): drop workspace.capture_auto_enrich, now that nothing reads it (#521)

dungbanguyen:
First slice of #521, and the only part of it that is finished today.

0190 moved the auto-enrich posture into `setting` and switched its one reader — capture's own settings store — to it. The column stayed behind so 0190's reverse had somewhere to put the value back, and so a rollback landed on a schema whose readers still worked. That obligation ends here.

**Nothing in production reads the column.** The remaining matches were #523's reset test, two comments, and 0122's prose.

Worth stating why that matters rather than just being tidy: a column that governs nothing while looking live **already misled code once on this branch**. Two integration suites kept toggling it after the readers moved — one test failed, and one passed *vacuously*, reporting a spend control as proven while its "don't crawl" branch was never taken. Leaving it is an invitation to repeat that.

## The reverse

`0194.down` restores both the column and its value. The default matches 0121's, so a workspace that never had a setting row lands on the posture it shipped with; the `UPDATE` then carries across whatever the setting holds, because after 0190 the row is where every change since has been written. Restoring the column without it would resurrect a posture an operator turned off.

Ordering works out: a rollback past 0190 runs 0194.down first, so the column is back before 0190.down writes to it.

## #523's tests

Moved onto `x_sor_mode`/`x_incumbent`, which prove the same two properties — configuration returns to its default, a co-tenant's does not — with columns that still exist.

`ResetWorkspaceConfig` itself needs **no change**: it derives its column list from the catalog, so the drop removes the column from its sweep automatically. That is exactly the property that made deriving-from-the-catalog worth doing in #523.

## Not in this change

`name`, `timezone` and `base_currency` still have readers — roll-ups, FX conversion, quota attainment and the report builder read them directly, which is why `UpdateInstallation` mirrors onto them. Dropping a column out from under eight readers is a different change with a different risk, so **#521 stays open** for it. That is also ADR-0091 phase 4's first real step.

`make check-backend` and `make craft-static` green. Integration runs first in CI (no Docker locally).

---
## Summary by cubic
Dropped the unused `workspace.capture_auto_enrich` column; the auto‑enrich posture now lives only in `setting`. Migrations preserve an OFF posture across the drop, add guarded rollbacks, and tests/docs now target the overlay columns; aligns with Linear #521 with no runtime behavior change.

- **Migration**
  - Added 0206 up: if any live workspace had `capture_auto_enrich = false` and the setting row is missing, insert `setting('capture.auto_enrich' = false)`, then drop the column.
  - 0206 down: re‑add the column as boolean NOT NULL DEFAULT true and backfill from `setting`, guarded by `jsonb_typeof(value) = 'boolean'`.
  - Added the same `jsonb_typeof` guard to 0190 down.

- **Refactors**
  - Moved `datareset` and identity tests off the dropped column to `x_sor_mode`/`x_incumbent`; docs and comments updated to reflect the split.
  - New migration integration tests cover the rollback restoring the current posture and the undecodable‑value case (falls back to default).
  - Covered the empty‑column path in `ResetWorkspaceConfig` on core‑only trees by probing inside a transaction.

<sup>Written for commit 6cc379e5191aa1903e3f9ef0d148f9698d264de3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/572?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workspace configuration resets to restore supported settings correctly, including overlay configuration and core-only schemas.
  * Preserved auto-enrichment settings when migrating configuration storage or rolling back migrations.
  * Prevented invalid setting values from interrupting migration rollback.

* **Documentation**
  * Clarified workspace reset behavior, including migration-defined defaults, preserved identity fields, and separately stored settings.

### DIFF
--- a/backend/internal/compose/datareset.go
+++ b/backend/internal/compose/datareset.go
@@ -247,9 +247,9 @@ func (h dataResetHandlers) sweepAndReseed(ctx context.Context, wsID ids.UUID, co
 		// bootstrap leaves. The overlay flip above is not the general case, it
 		// is the reported one: the sweep is derived from the tables carrying a
 		// workspace_id column and workspace keys on id, so NO workspace-level
-		// setting is in its target set — capture_auto_enrich (CAP-PARAM-7)
-		// outlived every reset before this call, and so would the next setting
-		// added to the row.
+		// setting is in its target set. The auto-enrich posture outlived every
+		// reset before this call, which is what exposed it, and so would the
+		// next setting added to the row.
 		//
 		// identity owns that row and derives its own column list the same way,
 		// so nothing here has to be kept in step with the schema. It runs AFTER
--- a/backend/internal/compose/datareset_integration_test.go
+++ b/backend/internal/compose/datareset_integration_test.go
@@ -382,14 +382,16 @@ func TestResetReturnsAnOverlayWorkspaceToNativeMode(t *testing.T) {
 // is derived from the tables carrying a workspace_id column; workspace keys on
 // id, so it is not excluded from that list, it is not a candidate for it.
 //
-// capture_auto_enrich (CAP-PARAM-7) is the second setting on the row today and
-// stands in for every one added after it: an operator who wiped the
-// installation would otherwise find yesterday's settings still applied to a
-// database with nothing in it.
+// x_sor_mode stands in for every configuration column the row carries: an
+// operator who wiped the installation would otherwise find yesterday's
+// settings still applied to a database with nothing in it. The settings that
+// live in the `setting` table are the other half of the same obligation, and
+// the sibling test below covers them.
 func TestResetRestoresWorkspaceLevelSettings(t *testing.T) {
 	e := integration.Setup(t)
 	ctx := e.Admin()
-	e.WsExec(t, `UPDATE workspace SET capture_auto_enrich = false WHERE id = $1`, e.WS)
+	e.WsExec(t, `
+		UPDATE workspace SET x_sor_mode = 'overlay', x_incumbent = 'hubspot' WHERE id = $1`, e.WS)
 
 	h := dataResetHandlers{
 		pool:  e.Pool,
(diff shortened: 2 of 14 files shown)


dungbanguyen:
Both P1s check out as false positives against the current tree, and the evidence is in the run that just went green rather than in my reading of the code.

**`(s.value)::boolean` "fails whenever the row exists".** It does not — PostgreSQL has cast jsonb to boolean since 11, and this exact path is covered: `TestRollingBackTheAutoEnrichColumnRestoresTheCurrentPosture` writes `'false'::jsonb` into `capture.auto_enrich`, runs `dbmigrate.Down` to this migration, and asserts the re-added column reads `false`. A failing cast would abort the migration and fail the test at `dbmigrate.Down`. It **PASSED** in the last run, alongside `TestRollingBackTheInflightMarkerParksUnlearnedSends`.

That test exists precisely because `TestMigrations_applyReverseReapply` runs on a fresh schema where the `UPDATE` matches nothing — so a wrong cast, key or join would pass unnoticed. It was added in this PR for that reason.

**`datareset_integration_test.go` still uses the dropped column.** Not any more. `rg capture_auto_enrich backend/internal/compose/datareset_integration_test.go` returns nothing on the current head; `TestResetRestoresWorkspaceLevelSettings` was re-pointed at `x_sor_mode`/`x_incumbent` on 7 Aug, in the commit that addressed the first review round — where this same finding was raised and fixed. The lines quoted (~392, ~406) are from before that commit.

Both threads resolved. If either was raised against the current head rather than an older commit, say so and I will look again — I would rather re-check than dismiss, and the first review round on this PR found two real defects I had missed.

Issue #539, opened 2026-08-06
compose/integration setup forks ~5 Postgres backends per test; sharing them needs plan-cache invalidation first


dungbanguyen:
## Summary

`internal/compose/integration` is half the lane (960 of 1828 tests). Each test's setup forks about five Postgres backends — a `pgx.Connect` for the owner, a `pgxpool` for the app role, and another `pgxpool` opened and closed just to run River's one-row schema probe. Reusing them across a package run measured **170.9s against a 209.6s baseline, ~18%**, on the same commit and machine.

The reuse cannot be taken naively, and this issue exists to record why, so the next attempt starts from the failure rather than rediscovering it.

## What blocks it: pgx's per-connection plan cache vs the harness's own DDL

The harness changes the schema between tests. `testdb.Reset` drops the runtime `cf_*` columns; `testdb.EnsureSchema` drops the schema outright. pgx caches a query plan per connection. A pooled connection that outlives such a DDL then serves a plan whose result type no longer matches:

```
ERROR unhandled error method=POST path=/v1/activities/…/send-email
  err="jobs: enqueue comms_send_email transactionally:
       ERROR: cached plan must not change result type (SQLSTATE 0A000)"
```

That surfaced as `TestPreferenceCenterOneClickUnsubscribe` failing with `marketing send → 500` — a suite with nothing to do with custom fields, one that passes in isolation and fails only in a full-package run. Per-test pools were hiding this by accident: a pool that dies with its test can never hold a stale plan.

## A second constraint, independent of the first

Sharing also raises the steady-state connection count. Pools previously closed per test; a pool that lives for a package run holds its connections for the whole run, and the lane runs ~6 package workers against one Postgres with `max_connections=100`. `database.NewPool`'s defaults are an api process's (`MaxConns=16`, `MinConns=2`), so 6 workers × (owner conn + two pools) can exceed the server ceiling. It does not fail cleanly — a caller waits for a connection until its context expires, so the symptom was `TestPrivacyRetentionFansOutOneJobPerWorkspaceAndFailsOnlyTheFailedTenant` timing out at 60s with `0 of 3 outcomes`, nowhere near the real cause.

Any shared-pool design must size itself explicitly (`pool_max_conns`, `pool_min_conns=0`) rather than inherit those defaults.

## Suggested design

[...]

PR #574, opened 2026-08-07
refactor(test): the booted-app fixture becomes importable, in its own package

dungbanguyen:
The third and last integration fixture to become importable, and the prerequisite every remaining suite group in #546 was waiting on. `AppEnv`, its three constructors, the three methods every suite calls and the `AnyMap` alias now live in `internal/compose/integration/apptest`.

**Stacked on #568** — review the top commit; the rest is that PR.

## Why a separate package, not a non-test file next to harness.go

Because the build refused the obvious thing. `package compose`'s own white-box tests import `package integration`, so nothing there may import `compose` — and this fixture boots a compose handler stack:

```
package .../internal/compose/integration
	imports .../internal/compose/integration/apptest from activity_lifecycle_integration_test.go
	imports .../internal/compose/integration from appenv.go: import cycle not allowed in test
```

`harness.go`'s doc has stated that constraint since it was written. I identified it earlier in this work, then put the fixture there anyway; the compiler corrected me. One level down, importing compose is free.

The direction is load-bearing in **both** senses: `apptest` must not import `integration` either, or the cycle closes from that side, since `integration`'s suites import `apptest`. `applyRiverSchema` is duplicated across that boundary for exactly that reason and says so — and the original cannot move here, because `package compose`'s tests use it too and would then reach compose through it.

The type is `AppEnv`, not `Env`: `integration.Env` is the database fixture.

## What moved, and what deliberately did not

Moving the type to another package makes **every method on it illegal outside that package**, so the sixteen a suite kept for itself are now plain functions taking the fixture. Only three were promoted with it — `Call`, `BootstrapWorkspace`, `SetWorkspaceSeat` — plus the two package-level helpers those call, which a non-test file cannot reach in a `_test.go` one.

`inWorkspace` looked shared at twelve files and is not: all twelve are outbound messaging. `listTools`/`getJSON`/`raw` are the MCP suites. Those travel with their own group, which is the rule the package doc states.

## Scope note

The plan called for a smaller first step — type promotion with fields left unexported. I did the whole promotion instead, because stopping there ships an **inert** PR: a fixture whose fields no sibling can read. That is exactly the defect review caught in #563, and both halves touch the same 76 files, so splitting doubles the review surface rather than halving it.

## Four gates had something to say once the code left a test file

Same discovery as the migrate-once gate two changes ago: a gate keyed on the `_test.go` suffix silently judges less when a fixture is promoted.

- **noctx** wanted `NewRequestWithContext`. Right — test files were exempt, a non-test file is not.
- **bodyclose** lost sight of every close, because the closer moved out with the fixture. Waived per site, and the sweep **refused to waive one where it could not first see the `CloseBody`** — that one turned out to close eight lines down rather than deferred, so its waiver says that instead of the boilerplate reason.
- **revive** wanted the docs to name the exported symbols. They described unexported ones.
- **the fleet-scan gate** flagged the fixture's workspace lookup. `slug` is a unique key (`0002_identity.up.sql`, `workspace_slug_unique`), so a lookup by it reads one row for the same reason a lookup by id does. The gate is taught that once rather than granted a waiver per site. Mutation-tested: a bare `SELECT id FROM workspace` and a `WHERE slug = ANY(...)` set-read both still fail it, so widening the predicate did not blind it.

## Verification

- **Importability is checked, not claimed.** A throwaway sibling package calling `SetupApp`, `BootstrapWorkspace`, `SetWorkspaceSeat` and `Call` compiles; it would not have before. #563 asserted this and was wrong, so it is a test now.
- **968 top-level tests before, 968 after.**
- `make check-backend`, `make lint` (0 issues), `make craft-static` (0 findings), `go vet -tags=integration` — green.
- **The full integration lane was run** on this tree: `OK: integration passed with 0 skips (27 packages, parallel)`, 397s wall.

## What it buys, measured

Now that the lane has actually been run back-to-back on one machine, with untouched packages as a control: the split so far cuts the lane's floor about **20%**, not the 3× I projected earlier from a model. The parent is still 349.6s and 88% of the wall clock. This PR adds no speed of its own — it is the prerequisite that makes the remaining cuts possible, and by measured seconds only overlay (41.3s) is still worth cutting before the better lever, #539.

---
## Summary by cubic
Promoted the booted-app integration fixture into `internal/compose/integration/apptest` as `AppEnv`, making it importable and removing the compose/integration import cycle. All integration suites now use this package; behavior is unchanged, and comments/docs were updated (including removing a duplicate `CloseBody` doc).

- **Refactors**
  - Added `apptest` with `AppEnv` (TLS test server + cookie-jar client) and `AnyMap`.
  - Public API: `SetupApp`, `SetupAppWithOptions`, `SetupAppWithOriginOptions`, `Call`, `BootstrapWorkspace`, `SetWorkspaceSeat`; helper `BootstrapWorkspaceSession`; `CloseBody`.
  - Exported `AppEnv` fields: `Slug`, `Owner`, `Pool`, `TS`, `Client`; suites updated.
  - Kept suite-specific helpers local (e.g., MCP); duplicated `ApplyRiverSchema` to avoid cycles.
  - Tightened workspace lookup by slug; addressed lints (`NewRequestWithContext`, response body closes); added package docs on import direction.
  - Swept and fixed stale comments to match promoted symbols; removed the duplicate `CloseBody` doc block.

- **Migration**
  - Import `github.com/gradionhq/margince/backend/internal/compose/integration/apptest`.
  - Replace setup: `setup(...)` → `apptest.SetupApp(...)` (or `SetupAppWithOptions`, `SetupAppWithOriginOptions`).
  - Replace env members and calls: `e.slug/owner/pool/ts/client` → `e.Slug/Owner/Pool/TS/Client`; `e.call(...)` → `e.Call(...)`; `e.bootstrapWorkspace(...)` → `e.BootstrapWorkspace(...)`.
  - Use `apptest.BootstrapWorkspaceSession`, `apptest.AnyMap`, and `apptest.CloseBody` where needed.

<sup>Written for commit d981a774c785a6165bfb5fd4db0c68b65823c8ba. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/574?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/internal/compose/integration/activity_lifecycle_integration_test.go
+++ b/backend/internal/compose/integration/activity_lifecycle_integration_test.go
@@ -11,61 +11,26 @@ package integration
 // association whose target passes the visibility probe.
 
 import (
-	"context"
-	"io"
-	"log/slog"
 	"net/http"
-	"os"
-	"path/filepath"
 	"testing"
 
-	"github.com/gradionhq/margince/backend/internal/compose"
-	"github.com/gradionhq/margince/backend/internal/platform/deployconfig"
+	"github.com/gradionhq/margince/backend/internal/compose/integration/apptest"
 )
-
-// bootstrapWorkspaceSession provisions the installation's organization
-// through the A107 boot path (configuration-driven, exactly what cmd/api
-// runs at startup) and signs its admin in over HTTP — the arrange step
-// every e2e scenario shares. The login also primes the server's
-// singleton-organization resolution before a test seeds any
-// cross-tenant rows directly.
-func bootstrapWorkspaceSession(t *testing.T, e *env, organizationName, adminEmail, adminName string) {
-	t.Helper()
-	pwFile := filepath.Join(t.TempDir(), "admin-password")
-	if err := os.WriteFile(pwFile, []byte("correct-horse-battery"), 0o600); err != nil {
-		t.Fatal(err)
-	}
-	cfg := deployconfig.Config{
-		Version:      1,
-		Organization: deployconfig.Organization{Name: organizationName},
-		BootstrapAdmin: &deployconfig.BootstrapAdmin{
-			Email: adminEmail, DisplayName: adminName, PasswordFile: pwFile,
-		},
-	}
-	if err := compose.EnsureInstallation(context.Background(), e.pool, slog.New(slog.NewTextHandler(io.Discard, nil)), cfg); err != nil {
-		t.Fatalf("bootstrap: %v", err)
-	}
-	if status := e.call(t, "POST", "/v1/auth/login", anyMap{
-		"email": adminEmail, "password": "correct-horse-battery",
-	}, nil, nil); status != http.StatusOK {
-		t.Fatalf("login → %d", status)
-	}
-}
 
 // seedTaskAndTarget logs one task activity plus a person for it to be
 // relinked onto, returning both ids.
-func seedTaskAndTarget(t *testing.T, e *env) (personID, taskID string) {
+func seedTaskAndTarget(t *testing.T, e *apptest.AppEnv) (personID, taskID string) {
 	t.Helper()
 	var person struct {
 		ID string `json:"id"`
 	}
-	if status := e.call(t, "POST", "/v1/people", anyMap{"full_name": "Task Target"}, nil, &person); status != http.StatusCreated {
+	if status := e.Call(t, "POST", "/v1/people", apptest.AnyMap{"full_name": "Task Target"}, nil, &person); status != http.StatusCreated {
 		t.Fatalf("create person → %d", status)
 	}
 	var task struct {
 		ID string `json:"id"`
 	}
-	if status := e.call(t, "POST", "/v1/activities", anyMap{
+	if status := e.Call(t, "POST", "/v1/activities", apptest.AnyMap{
 		"kind": "task", "subject": "Send offer",
 	}, nil, &task); status != http.StatusCreated {
 		t.Fatalf("log task → %d", status)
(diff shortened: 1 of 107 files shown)


dungbanguyen:
All three applied, plus thirteen more of the same shape the sweep found.

**`CloseBody` had two stacked doc blocks.** Correct — the second was written to replace the first and both shipped, and they disagreed about what a failed close means. Merged into one that says both true things: a broken close can hide a truncated read, and a leaked body should be a red test rather than a slow drip.

**Two comments named identifiers the promotion renamed** (`setupWithOptions`, `setup()`). Also correct, and the interesting part is what happened when I looked for siblings rather than fixing the two reported: **thirteen more** — `setWorkspaceSeat`, `anyMap`, `e.call`, more `setup()` — across ten files. All prose pointing at symbols that no longer exist.

That is the failure mode this whole refactor keeps producing, and it is worth naming: a compiler-driven rename is exact about *code* and completely blind to *comments*. Every mechanical pass in this programme has left some behind, and each time a reviewer has caught one or two of them. The sweep is the fix for the class — comment lines only, matched against the list of promoted names, re-run until it reports zero.

A comment naming a symbol that no longer exists is worse than no comment, because a reader trusts it enough to go looking.

`make check-backend`, `make craft-static` (0 findings) and `go vet -tags=integration` green.

Issue #482, opened 2026-08-05
Integration lane intermittently fails with SQLSTATE 53200 (out of shared memory) under 29-way parallelism


tinnhlt-gra:
The parallel integration lane (`make test-integration`, `scripts/test-integration-parallel.sh`) intermittently fails with:

```
ERROR: out of shared memory (SQLSTATE 53200)
```

Observed on a local run while working on #479, in `internal/compose` (`TestForecastDerivationHonorsRowScope`) and `migrations` (`TestAttachmentScanStatusBackfill`). Neither test involves a passport or the change under test; an immediate clean rerun of the same commit passed with `OK: integration passed with 0 skips`.

**Why this is worth fixing rather than tolerating**

The failure surfaces in whichever packages happen to hold the most locks concurrently, so it looks like a random product-test failure. That is the same shape as a real regression, which makes it expensive: the honest response to a red lane is to investigate the diff, and here the diff is innocent. A lane that fails for environmental reasons trains people to rerun red instead of reading it.

**Likely cause**

Postgres lock-table exhaustion, not memory pressure per se: 29 packages each run against their own throwaway clone database in parallel, and `max_locks_per_transaction × (max_connections + max_prepared_transactions)` bounds the shared lock table. Schema-heavy transactions (the RLS-sweep and migration tests hold locks on many relations at once) are the ones that tip it over.

**Suggested fix**

Raise `max_locks_per_transaction` (and review `shared_buffers`) on the dev Postgres container in `infra/docker-compose.dev.yml`, so the lane's worst-case concurrent lock count fits. Alternatively cap the lane's parallelism, though that trades wall-clock for reliability and the sizing fix is cheaper.

Not yet observed in CI — filing so that if it appears there, the diagnosis is already written down rather than rediscovered.

tinnhlt-gra:
Data point, not a resolution: **did not recur** across two full lane runs today on `feat/agent-tools-match-the-contract`, both ending `OK: integration passed with 0 skips`.

Worth noting the lane is now **25 packages, not 29** — #472 landed the tagged-tests-only change in between, which also prints a per-package cost table. Fewer concurrently-open clone databases means fewer concurrently-held locks, so the pressure that produced `SQLSTATE 53200` is lower than when this was filed. That is a plausible reason it has gone quiet rather than evidence it is fixed: the sizing is unchanged, so a package that grows its lock footprint can still tip it over.

[...]

PR #584, opened 2026-08-07
refactor(test): the overlay suites become their own integration package

dungbanguyen:
The last split of #546 worth making, and the first one whose payoff was **measured rather than projected**.

**Stacked on #574** — review the top commit; the rest is that PR.

## Chosen on time, not on file count

41s of the parent package's 349s from only 35 tests — the most expensive remaining cluster. Ranking by test count would have picked customfields (55 tests) or oauth (45), which are 18.0s and 14.3s. That mis-ranking is exactly what made my earlier estimate of the split programme's payoff wrong by a factor of three, so this one was picked off the per-test timings in a real lane run.

Four of the eleven files ride the booted-app fixture, which is the whole point of #574: until that fixture left a test file, no group depending on it could go anywhere.

## What it measured

One valid lane run (`OK: integration passed with 0 skips (28 packages)`, 1853 tests), so no cross-run normalisation:

```
./internal/compose/integration/overlay     75.64s    35 tests   12.6% of budget
./internal/compose/integration/org360      42.16s    61 tests    7.0% of budget
./internal/compose/integration/capture     33.12s    63 tests    5.5% of budget
./internal/compose/integration            260.22s   808 tests   43.4% of budget
wall: 296s
```

A lane's wall clock cannot go below its longest package. Unsplit, those four would be one package of **411s**, so the wall could not have been under that. It measured **296s**.

Two honest caveats on that figure. Each split package re-pays its own template clone and migrate-once, so the true unsplit total is somewhat below 411s — by summed test time the overlay suites are ~41s of work, not 75.6s, which puts the unsplit parent nearer 300–336s and this single cut's gain at **14–23%**. And I attempted a same-machine A/B against the unsplit tree; the control run died on `SQLSTATE 53200 — out of shared memory` (#482, back-to-back lane runs exhaust Postgres), so **I am not reporting a number from it**. The same-run counterfactual above needs no control.

## Two structural notes

**`overlay_acceptance_seam_test.go` stayed behind.** It carries no `//go:build integration`, so it belongs to the unit lane, and it both declares and consumes `backendModuleRoot`, which a webhooks suite there also reads. Leaving it removed the group's only outward debt.

**Two gates in `package compose` read this suite's source by hardcoded path** and failed on the move — on a path, not on anything about the code. That is the third path-keyed gate this programme has tripped, so it is fixed by *derivation* rather than repointing: they now resolve the pin by name anywhere under `integration/`. Mutation-tested, because a resolver that survives a move must not also survive the pin vanishing:

```
pin in place:                     PASS
pin removed:                      FAIL (names the file)
pin moved to integration/capture: PASS
```

Neither trap the capture move hit applies here — nothing declares a method on a foreign fixture in either direction, checked both ways.

## Verification

968 top-level tests before, 968 after. `make check-backend`, `make craft-static` (0 findings), `go vet -tags=integration` green, and the full lane run above.

## Where this leaves #546

**Stop splitting here.** The parent is still 260s and **88% of the wall clock**, but what remains is 116 clusters averaging under 1.5s each — 46% of the package in pieces too small to be worth a package. The better lever is #539: sharing the Postgres connections measured ~18% off this package's *work*, which reaches the long tail that splitting structurally cannot.

---
## Summary by cubic
Split the overlay integration suites into `internal/compose/integration/overlay`, added shared ZIP/CSV helpers, and replaced path-based gates with a name-based pin to keep assertions stable across moves; integration lane wall time down by ~14–23% (296s measured).

- **Refactors**
  - Moved overlay suites (acceptance, e2e, dispatch, flip/claim, write shadow/audit, tool surface, user map) to `internal/compose/integration/overlay`; kept `overlay_acceptance_seam_test.go` in the unit lane.
  - Added `integration.BundleEntries` and `integration.CSVColumn` in `integration/bundle.go`; export and filtered-export tests now reuse them.
  - Switched the suite to `package overlay` and imported the production module as `overlaymod`.
  - Replaced path-keyed gates with name-based resolution via an `overlayPin` that finds `overlay_toolsurface_integration_test.go` anywhere under `integration/` and fails if the pin is missing or duplicated.

- **Bug Fixes**
  - `BundleEntries` now fails on duplicate ZIP members to avoid last-one-wins assertions.
  - Scoped the fleet-enumeration gate to exclude integration-tagged non-test files and restored the original single-row predicates, preventing hidden fleet scans via `slug`.

<sup>Written for commit 07cba2888e70c6165a28a3a6b5981b70acc18128. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/584?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Tests**
  - Improved integration test organization and shared test utilities.
  - Strengthened validation for exported ZIP and CSV data.
  - Made overlay test configuration discovery more resilient.
  - Refined fleet-scan coverage to better distinguish production code from integration tests.
  - Preserved existing test behavior while improving setup consistency and diagnostics.

- **Documentation**
  - Updated testing guidance and comments to reflect current integration harnesses and response-handling practices.

### DIFF
--- a/backend/internal/compose/dispatcher_test.go
+++ b/backend/internal/compose/dispatcher_test.go
@@ -14,7 +14,7 @@ package compose
 // touching a mirror store. d.native is left nil: the overlay=true seed
 // guarantees it is never dereferenced. The native-mode branch (and the
 // real cache-miss DB query) is proven end to end by
-// compose/integration/overlay_dispatch_integration_test.go, which needs
+// compose/integration/overlay/overlay_dispatch_integration_test.go, which needs
 // a real, migrated Postgres.
 
 import (
--- a/backend/internal/compose/integration/bundle.go
+++ b/backend/internal/compose/integration/bundle.go
@@ -1,0 +1,72 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+//go:build integration
+
+package integration
+
+import (
+	"archive/zip"
+	"bytes"
+	"encoding/csv"
+	"io"
+	"testing"
+)
+
+// Readers for the export bundle: a zip of CSVs. Exported because the suites that
+// assert on a bundle's contents sit on both sides of the overlay split — the
+// export suites here, the mirror-flip suite in the overlay package — and both
+// want the same parse rather than two that could disagree about a header row.
+
+// BundleEntries reads the produced ZIP into name→bytes.
+func BundleEntries(t *testing.T, raw []byte) map[string][]byte {
+	t.Helper()
+	zr, err := zip.NewReader(bytes.NewReader(raw), int64(len(raw)))
+	if err != nil {
+		t.Fatalf("opening bundle zip: %v", err)
+	}
+	entries := map[string][]byte{}
+	for _, f := range zr.File {
+		rc, err := f.Open()
+		if err != nil {
+			t.Fatalf("opening %s: %v", f.Name, err)
+		}
+		data, err := io.ReadAll(rc)
+		if err != nil {
+			t.Fatalf("reading %s: %v", f.Name, err)
+		}
+		if err := rc.Close(); err != nil {
+			t.Fatalf("closing %s: %v", f.Name, err)
+		}
+		entries[f.Name] = data
+	}
+	return entries
+}
+
+// CSVColumn parses a CSV entry and returns the values under one column —
+// the format-validity check (csv.Reader fails loudly on a malformed file)
+// doubling as the content probe.
+func CSVColumn(t *testing.T, raw []byte, column string) []string {
+	t.Helper()
+	records, err := csv.NewReader(bytes.NewReader(raw)).ReadAll()
+	if err != nil {
+		t.Fatalf("parsing csv: %v", err)
+	}
+	if len(records) == 0 {
+		t.Fatal("csv has no header row")
+	}
+	idx := -1
+	for i, h := range records[0] {
+		if h == column {
+			idx = i
+		}
+	}
+	if idx == -1 {
+		t.Fatalf("csv has no %q column; header=%v", column, records[0])
+	}
+	var out []string
+	for _, row := range records[1:] {
+		out = append(out, row[idx])
+	}
+	return out
+}
--- a/backend/internal/compose/integration/bundle.go
+++ b/backend/internal/compose/integration/bundle.go
@@ -19,6 +19,11 @@ import (
 // want the same parse rather than two that could disagree about a header row.
 
 // BundleEntries reads the produced ZIP into name→bytes.
+//
+// A repeated member name is a failure, not a last-one-wins overwrite: zip
+// permits duplicates, and a map would silently keep only the final copy — so a
+// bundle that shipped two conflicting overlay_mirror.csv entries would satisfy
+// every assertion against whichever came last.
 func BundleEntries(t *testing.T, raw []byte) map[string][]byte {
 	t.Helper()
 	zr, err := zip.NewReader(bytes.NewReader(raw), int64(len(raw)))
@@ -37,6 +42,9 @@ func BundleEntries(t *testing.T, raw []byte) map[string][]byte {
 		}
 		if err := rc.Close(); err != nil {
 			t.Fatalf("closing %s: %v", f.Name, err)
+		}
+		if _, dup := entries[f.Name]; dup {
+			t.Fatalf("the bundle carries %s twice — assertions would have read only the second copy", f.Name)
 		}
 		entries[f.Name] = data
 	}
--- a/backend/internal/compose/integration/capture/doc.go
+++ b/backend/internal/compose/integration/capture/doc.go
@@ -14,11 +14,8 @@
 // fixtures (SearchEnv mostly, Env in one suite) and its shared seed, job-wait and
 // HTTP-stub helpers; it owns its own connector stubs, seeding and assertions.
 //
-// One capture suite did NOT come along: capturedbykind_http_integration_test.go
-// drives the booted-app fixture, which was still trapped in a test file when
-// this package was carved out. It has since been promoted to
-// integration/apptest, so nothing holds that suite here any more — it should
-// join this package the next time someone is in the area.
+// One capture suite is not here: capturedbykind_http_integration_test.go rides
+// the booted-app fixture and lives in the parent package.
 //
 // The production capture module is imported as capturemod, because inside a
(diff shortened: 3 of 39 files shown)


dungbanguyen:
Review round applied. One finding was a real weakening **I introduced**, and the security reviewer proved it by mutation rather than argument.

## The fleet-scan gate: scope was wrong, not the predicate

Promoting the app fixture into a non-test file dropped its `SELECT id FROM workspace WHERE slug = $1` into `TestFleetEnumerationOnlyAtRatifiedSites` for the first time (`goFilesUnder` skips `_test.go`). I "taught the gate" that a slug lookup reads one row. The fact is true; the fix was still wrong — `singleRowPredicates` is consulted for **every file under `internal/`**, so a production module could then hide a fleet loop behind a bare `WHERE slug =` landing in the marker's two-line window. A tree-wide relaxation of a production gate, to accommodate a test fixture. That is the rule-5 shape: rationalised in a comment instead of restructured away.

The reviewer reverted just those two strings and confirmed the gate then names **only** `apptest/appenv.go` — so the widening existed solely for the fixture.

Fixed by scope: the gate judges production passes, an integration-tagged file is harness that never ships and has no job to take a workspace from, so it is excluded by rule and the predicate list is back to what it was. **Mutation-tested both directions**, because a rescoped gate that merely passes has proved nothing:

```
production `WHERE slug = ` scan   FAIL  (the widening would have ADMITTED this)
production bare fleet scan        FAIL
same scan in the harness          PASS  (correctly ignored)
```

Worth noting the two gates now scope **oppositely**, and that is correct: migrate-once *includes* integration-tagged non-test files because its obligation is the lane; fleet-scan *excludes* them because its obligation is production. Both now say so at the point where they do it.

## `overlayPin` took the last of several matches

Both reviewers caught this; the security one supplied the decoy repro. A split in progress leaves two copies of a suite for as long as it takes to delete the first, and both gates would have derived their verb lists from whichever the walk reached last — green against a stale pin, the exact shape they exist to catch. Now fails unless there is exactly one, naming both paths. Mutation-tested: one pin passes, a decoy copy fails and names both.

## Comments, again

- **`harness.go`'s package doc still stated the superseded split rule** — "rides Env or SearchEnv rather than the booted-app fixture" — which this branch falsified. That is the file the next person opens to plan a cut, and it told them not to do what this branch just did. Rewritten to name all three fixtures, the cycle direction, and the two things a split reliably runs into.
- **Seven files still pointed at `e2e_integration_test.go`** for the fixture. My earlier sweep repointed identifier *names* and missed their *homes*.
- **`capture/doc.go` stated a blocker that no longer exists** — it said a suite stays "until that fixture is promoted", and it now is.
- **Two comments the sweep missed**, so my "zero stale references" claim was wrong. Both fixed; the sweep now reports zero for real.
- **Sixteen `bodyclose` waivers** said "the deferred CloseBody" where the call is straight-line, or "on the next line" where it is four down. One spelling now, and a checkable one — a waiver's reason is the only thing between it and a rubber stamp.
- **`overlay/doc.go` carried a post-mortem** of a planning estimate. Deleted; it states what the package holds.
- **`raw` renamed to `mcpRaw`** — as a package-level function it collided with a local `raw` in 12 files, including inside its own body.

## Recorded, not fixed

- **Parameter order**: the 16 converted helpers take `(env, t, …)` while `apptest.BootstrapWorkspaceSession` takes `(t, env, …)`. Real inconsistency, but re-churning 16 signatures and ~100 call sites to fix ordering is its own change — and the reviewer's alternative (a one-line embedding struct so they stay methods) deserves consideration first.
- **`AnyMap` placement**: 14 files import `apptest` solely to spell `map[string]any`. It belongs in `integration`. Placement, not correctness.

968 tests conserved. `make check-backend`, `make craft-static` (0 findings), `go vet -tags=integration` green.

dungbanguyen:
Recorded the two deferrals as #588 rather than leaving them in a PR comment, per the repo rule that anything found and not fixed becomes an issue — the parameter-order split across the sixteen converted helpers, and `AnyMap` sitting in `apptest` when fourteen files import that package solely for it. The issue also notes the alternative a reviewer raised: a one-line embedding type would let those helpers stay methods and make the ordering question disappear, which may beat reordering sixteen signatures.

Also added today's reproduction to #482. The trigger is cheaper than the 29-way parallelism it was filed under: **two consecutive `make test-integration` runs at the default `JOBS`** were enough. Same code both runs, `briefs` untouched, first passed and second died on `SQLSTATE 53200` eleven tests in and aborted 75 tests short. Worth flagging there because it does not just fail — it invalidates measurement quietly, which is exactly how it wasted my A/B control for this PR.

dungbanguyen:
Four applied, two declined with reasons.

**Applied — duplicate zip members.** `BundleEntries` read the archive into a name→bytes map, so two members with the same name collapsed to the last and every assertion about that entry read a file the bundle also contradicted. I checked the premise rather than assuming it — `archive/zip` will happily write and read two members named `overlay_mirror.csv`, so the guard is reachable, not defensive theatre. It fails now, naming the member.

**Applied — three comments saying more than they should.** The fleet-scan scope comment retold how the harness stopped being excluded and which fix I rejected; that is the commit's job and the durable half stands alone. `capture/doc.go` told a future contributor to move a suite — that belongs in an issue, not in source. The overlay doc's build-lane narration went in the previous review round.

**Declined — `WHERE w.slug = ` symmetry.** This reviews a superseded commit. The slug predicates were reverted entirely in the previous round: the widening was the wrong fix, because `singleRowPredicates` is consulted for every file under `internal/`, so it loosened a production gate tree-wide to accommodate a test fixture. There are now zero `slug` mentions in that gate. Adding an aliased form would re-introduce what was removed.

**Declined — `filepath.SkipAll` on first match.** This directly contradicts the other reviewer, and the other reviewer is right. Stopping at the first match is *exactly* the last-match-wins ambiguity that was flagged and fixed: the walk deliberately collects **all** matches so it can fail when there are two, because a split in progress leaves two copies of a suite and both gates would otherwise derive their verb lists from whichever the walk reached first. Mutation-tested — a decoy copy fails and names both paths. The traversal cost is one `WalkDir` over `integration/` per gate run, which is not worth trading the ambiguity check for.

`make check-backend`, `make craft-static` (0 findings) and `go vet -tags=integration` green.

PR #602, opened 2026-08-07
feat(company): the company record page V2 — canonical fields, Today, evidence verdicts, coverage, documents, attachment carriage

LarsGradion:
Seven slices of the company-record-page V2 plan, built against the upstream ADRs that PR #1251 wrote for it. Each is independently useful and each carries the state distinctions the plan asks for.

## What lands

**Slice B1 — canonical company fields (`ADR-0085`/`A130`, migration 0194).** Three stores held a company's industry and nothing said which was true. Now: for any fact that is a column on `organization`, the column IS the value and the sidecar holds a proposal plus its evidence. `linkedin_url` becomes a validated, live-unique column; the readable website stays derived from the primary domain rather than gaining a second store; both evidence sidecars gain `retrieved_at` / `verified_at` / `verified_by`.

**Evidence verdicts.** B1 shipped confirm and correct and nothing in the product could call them, so a rep who could see a wrong extraction had no way to say so and the next enrichment run rewrote it. Every profile field and fact now carries both. Correcting `display_name` moves the column, stamps `name_source = human` and runs the duplicate recheck.

**Today on this account.** New 360 section `next_meeting` — the soonest meeting not yet started, with the attendees this caller may read. Facts, assessments and recommendations are labelled apart. It deliberately does **not** repeat the suggestions or the open commitments, which already render with their evidence and dismissal elsewhere.

**Relationship coverage.** Each contact names the strongest few colleagues who have actually exchanged messages, plus `+N`. A coverage explorer compares up to eight colleagues the reader picks — never a contact × every-colleague matrix, which is unreadable at forty reps.

**Documents (`documents-and-files`, migration 0195).** A file gains a category, a display title, a lifecycle state and a pin. The account roll-up scopes every candidate through its **own primary parent**, so a contract on a deal you cannot see contributes neither a row nor a count.

**Attachment carriage (`ADR-0086`/`A131`, migration 0196).** Carriage is a declared capability with no default; a channel that cannot carry files **parks** the message rather than sending the text alone. What was sent is snapshotted immutably.

**The base-currency freeze probe** counted only closed deals. A sent offer freezes the same rate and its figure has already reached a customer in a PDF — this was live in merged main, not just here.

## The distinctions this PR is really about

Several sections exist to keep two states apart that a naive implementation renders identically:

- **Untried vs cold** — nobody has written to this contact, versus somebody wrote repeatedly and got nothing. Opposite next moves.
- **No meeting booked vs you cannot see the calendar** — a fact about the account, versus a fact about the reader.
- **Hidden from you vs there is none** — a list assembled from three of five sources is not the same list.
- **Stripped vs parked** — a message that goes out without its files fails silently and permanently.

Each is pinned by test.

## Mutation-checked

Reverting each of these fails its test: the fact-key split, the participant row scope, the document parent scope, the carriage gate, the offer arm of the freeze probe, and the LinkedIn host shape.

Two bugs were found by real data rather than by fixtures — a meeting attendee listed once per role, and a fact correction that reached the wrong row because every company fact shares an empty `value_key`.

## Gates

`make check`, `make check-fe` (1,913 tests), `make craft-static`, and the integration lane all green locally. Two integration tests failed once under the parallel lane and pass in isolation — `TestBackfillCompletionBuildsTheDigest` and `TestSweepBackoffGatesDueOverlayConnections`, neither in code this PR touches.

## Not in this PR

PR 4 (dossier and growth fit) is a module on the scale of `compose/orgbrief`, and growth fit's deterministic floor is abstention — see the plan's §15 and STATUS.md. PR 6's other halves (the composer learning carriage before send, the transmit-time recheck), PR 8 and PR 9/10 are likewise outstanding.

---
## Summary by cubic
Delivers the core of the company record page V2: canonical company fields with verifiable evidence, a “Today” view, per‑contact coverage and explorer, an account documents library, and safe outbound attachments. Tightens LinkedIn URL validation and document idempotency, and fixes paging, visibility, download handling, and document anchoring on company merge.

- **New Features**
  - Canonical fields and evidence: `linkedin_url` is a validated, live‑unique column; website derives from the primary domain; evidence sidecars gain `retrieved_at` / `verified_at` / `verified_by`; confirm/correct update canonical columns atomically and preserve the machine’s proposal in the audit.
  - “Today” and coverage: new `next_meeting` section with row‑scoped, de‑duplicated attendees; withheld vs none is carried via `sections_omitted`; per‑contact routes (untried vs cold) and a coverage explorer comparing selected colleagues; one‑query reads where added.
  - Documents: metadata (category, title, `doc_state`, pin, supersedes); account library scopes each file through its own primary parent so hidden parents don’t leak rows or counts; uploads stamp `organization_id`; new `GET /v1/organizations/{id}/documents`.
  - Messaging safety: outbound attachments are snapshotted on delivery; channels declare `connector.AttachmentCarrier`; transmit‑time integrity rechecks scan state, visibility, and sender grants; a channel that can’t carry files, or a file that no longer passes, parks the message and names the missing/blocked files.

- **Bug Fixes**
  - Base currency: lock counts every frozen rate (deals and offers), returns 409 `base_currency_locked` with a reason; FX list reads the workspace base.
  - Evidence and PATCH safety: archived companies can’t be corrected; `display_name` corrections stamp `name_source=human` and recheck duplicates; LinkedIn normalization validates host and rejects trailing slashes; strict If‑Match parsing; handlers distinguish null vs absent on sparse PATCH.
  - Next meeting: attendees listed once per person and only those the viewer may see; canceled meetings excluded; absence conveyed via `sections_omitted`.
  - Coverage routes: withheld without the activity grant (section omitted, not empty); route‑band mapping fixed to declared enum.
  - Documents: keyset paging no longer strands items across the pinned boundary; download links use `GET /v1/attachments/{id}`; external identity must be a (source,id) pair; supersedes cycles refused and targets visibility‑gated; activity‑borne files now appear in the account library; filtered zero keeps controls so the filter can be cleared; organization merge re‑anchors `attachment.organization_id` to the survivor so files don’t vanish from the account library.
  - Tests and reliability: fixed clock in the activity fixture to avoid flakes; the Today card now surfaces a failed read even when a view is present, so errors aren’t silently treated as “quiet account.”

<sup>Written for commit 4ba63d4a0f99c023926abf1d596a99225afd6149. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/602?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added a redesigned company record experience with lifecycle, owner, LinkedIn, address, activity, task, history, and sharing actions.
  * Added company documents with filtering, metadata, pinning, supersession, download availability, and status indicators.
  * Added inline evidence confirmation and correction for company fields and facts.
  * Added “Today on this account,” contact route recommendations, and relationship coverage exploration.
  * Added FX base-currency visibility and safeguards against changes after rates are frozen.
* **Bug Fixes**
  * Improved permission handling, archived-record restrictions, document visibility, and attachment delivery validation.
* **Documentation**
  * Updated implementation status and handoff materials.


LarsGradion:
## Review disposition

A Codex review (`codex exec review --base origin/main`) was run over this branch in addition to CodeRabbit and cubic. It found seven defects, two of which no bot reported. Everything below is fixed with a test that fails when the fix is reverted, except where marked rejected.

### Fixed — user-visible breakage

| Defect | Why it mattered |
|---|---|
| Document downloads 404'd | The card linked `/v1/attachments/{id}/download`; the contract serves `GET /v1/attachments/{id}`. |
| The account library was always empty | `UploadAttachment` never wrote `organization_id`, which `ListOrganizationDocuments` filters on. The library's tests seeded that column themselves, so they proved the read filters on it and nothing more. |
| Pages after the pinned group were unreachable | The list ordered on `(pinned, created_at, id)` and continued the page on two of three. A boundary inside the pinned group excluded every newer unpinned document from every later page — permanently, not late. |
| Canceled meetings were offered as "next meeting" | Sent a rep to prepare for something that will not happen, and hid the real next meeting behind it. |
| Route bands were unrenderable | `relstrength` scores `none/weak/moderate/strong`; the contract declares `cold/developing/strong`. The cast shipped `weak` on a wire that never admitted it. |
| Successful edits looked like no-ops | The invalidations used `orgProfileFields`/`orgFacts`; the consumers register `org-profile-fields`/`org-facts`, and `["organization", id]` was never invalidated at all. |

### Fixed — security and data integrity

- **Contact routes leaked activity-derived data.** `contactRoutes` reads `graph_interaction_edge`, and an edge is derived from an activity — which is why `graphourside.go` demands `activity:read` before touching the same table. Routes now go **absent** for a caller without that grant, not empty: an empty set is an answer ("nobody can reach them"), and giving that answer to someone who was not allowed to ask is the same disclosure inverted.
- **Human corrections did not survive enrichment.** Both enrichment upserts decline to overwrite a row whose `captured_by` is a human, and they test that column rather than `source`. The evidence writer moved only `source`, so the next ordinary deep read put the machine's value back with nothing to show it had happened. The new test drives the real `ApplyDeepRead`.
- **A supersedes target was not visibility-gated.** Pointing at a document is a read of it; accepting an unreadable id both confirmed the target existed and built a relationship across a visibility boundary. Refused as `ErrNotFound` so existence stays hidden.
- **`0195` now backfills the account roll-up**, so files already attached to a company or its deals do not vanish from the library on upgrade.

### Fixed by withdrawal

**`If-Match` on `updateAttachmentMetadata` could never work.** `attachment` has no `version` column and the wire `Attachment` exposes none, so a client had nothing to send and a well-formed header would have failed as a raw SQL error against `version`. The precondition is withdrawn from the contract rather than advertised and unfulfillable. Concurrent metadata edits are last-write-wins, and each is audited.

### Rejected, with reason

**The `UpdateAttachmentMetadata` outbox waiver stays.** The suggestion was to define an attachment event and emit it. I implemented that, then reverted it: the spec's event catalog is closed (`events.md` §5 defines no `attachment.*` type) and the closed-verb law forbids inventing one build-side. Three sibling waivers say exactly this for `UploadAttachment`, `ArchiveAttachment`, and `MarkScanResult`. Contract-first (P3) means the spec wins here; changing it is an upstream spec question, not a build-side one.

### Known and not addressed in this PR

- The roll-up is not maintained on organization merge or activity relink, so a file can stay filed under an account its parent left.
- A person-parented file rolls up to nothing: `person`→`organization` runs through `relationship` kind `employment`, which is many-valued, so a contact at two companies has no single account.
- Activity-parented files are absent from the library's parent kinds.
- The documents card does not consume `next_cursor`, so an account with more than 50 documents shows only the first page.
- Concurrent metadata edits can still race a supersedes cycle past the check.

These are follow-ups, not regressions in this branch.

PR #625, opened 2026-08-08
fix(test): a Redis logical db per package, and a run that refuses to wrap

dungbanguyen:
The lane gives every package private state on three axes — its own clone
database, its own MinIO bucket, its own Redis logical db. The Redis one had
quietly stopped holding.

The mapping was `1 + (idx-1) % 15`, so it wrapped. With 28 packages, package 16
lands on the same db as package 1, and the two can run at the same time.

## Why a shared db is not just contention

`platform/events` and `overlaybudget/budgettest` both `FLUSHDB` between tests. A
collision wipes the other package's keys mid-test, and the failure lands wherever
those keys were about to be read — with nothing pointing back at the mapping.

It cost a run: `internal/compose` and `internal/platform/events` both drew db 9,
and `TestResetDataAuditEvidenceCarriesTheSameCacheKeyTallyAsTheResponse` read
`cache_keys_deleted = 0` for a counter it had just written.

## What changes

- Redis serves 64 databases instead of the default 16.
- The lane **refuses to start** when there are more packages than dbs, rather
  than wrapping the modulo. Wrapping is the part worth removing: a lane whose
  isolation silently degrades as packages are added reads exactly like a lane
  that is still isolated.
- The 1..15 bound was spelled twice in Go — once in `events`, once in
  `budgettest` — and both copies rejected the wider assignment, which is how the
  widening announced itself. They become one `testdb.RedisDB`, so the number of
  dbs the lane may use is stated in three places that have to agree rather than
  five that merely happened to.

## Verification

`make test-integration` green — 28 packages, 1906 tests, 0 skips — with the
assigned dbs now 1..28, all distinct.

---
## Summary by cubic
Give each integration test package its own Redis logical DB and refuse to run if packages exceed available DBs, preventing `FLUSHDB` collisions in parallel runs. Centralizes DB selection via `backend/internal/platform/testdb.RedisDB` and expands Redis to 64 DBs.

- **Bug Fixes**
  - Redis now serves 64 DBs (`redis-server --databases 64`); db 0 stays reserved for `make dev`.
  - Runner sets and exports `REDIS_DBS=63`, assigns dbs 1..`REDIS_DBS`, and fails fast when `NPKGS > REDIS_DBS` instead of wrapping.
  - New `backend/internal/platform/testdb/redis.go` with `RedisDBs` and `RedisDB(t)`; used by `platform/events` and `overlaybudget/budgettest`.
  - Guard test (`backend/internal/platform/testdb/redis_integration_test.go`) keeps the runner, compose, and the running server counts in sync.
  - Prevents `FLUSHDB` collisions from `platform/events` and `overlaybudget/budgettest`.

- **Migration**
  - Recreate the dev Redis container: run `make db-up` (or `docker compose up -d --force-recreate redis`).

<sup>Written for commit dbae9cb16e662fba9e8075c1e36291749ada95b2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/625?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved isolation for parallel integration tests using Redis, preventing database reuse between test packages.
  * Added validation to detect when more Redis databases are needed than are available.
  * Added consistency checks to ensure Redis database settings match across the test runner, development environment, and running server.

* **Test Improvements**
  * Standardized Redis test database selection and cleanup across integration tests.
  * Expanded support for isolated Redis databases during development and parallel testing.
  * Updated documentation for Redis database allocation and parallel execution.

### DIFF
--- a/backend/internal/platform/events/bus_integration_test.go
+++ b/backend/internal/platform/events/bus_integration_test.go
@@ -18,7 +18,6 @@ import (
 	"fmt"
 	"log/slog"
 	"os"
-	"strconv"
 	"sync/atomic"
 	"testing"
 	"time"
@@ -32,27 +31,6 @@ import (
 	kevents "github.com/gradionhq/margince/backend/internal/shared/kernel/events"
 	"github.com/gradionhq/margince/backend/internal/shared/kernel/ids"
 )
-
-// testRedisDB is the Redis logical database this lane isolates its streams in.
-// setup FlushDB's the whole db between tests, so it must never be db 0 — that
-// one is reserved for a running `make dev`, whose relay/subscriber use the
-// client default. The parallel runner hands each package a distinct db in
-// 1..15 via MARGINCE_TEST_REDIS_DB so concurrent packages never share a stream;
-// absent the env (a bare `go test`), db 15 is the isolated default. An
-// out-of-range or non-numeric value fails loudly rather than silently eating
-// the wrong db.
-func testRedisDB(t *testing.T) int {
-	t.Helper()
-	raw := os.Getenv("MARGINCE_TEST_REDIS_DB")
-	if raw == "" {
-		return 15
-	}
-	db, err := strconv.Atoi(raw)
-	if err != nil || db < 1 || db > 15 {
-		t.Fatalf("MARGINCE_TEST_REDIS_DB=%q is not a Redis db index in 1..15", raw)
-	}
-	return db
-}
 
 type busEnv struct {
 	pool *pgxpool.Pool
@@ -104,7 +82,7 @@ func setup(t *testing.T) *busEnv {
 	}
 	t.Cleanup(pool.Close)
 
-	rdb := redis.NewClient(&redis.Options{Addr: redisAddr, DB: testRedisDB(t)})
+	rdb := redis.NewClient(&redis.Options{Addr: redisAddr, DB: testdb.RedisDB(t)})
 	if err := rdb.Ping(ctx).Err(); err != nil {
 		t.Fatalf("redis at %s unreachable — run `make db-up`: %v", redisAddr, err)
 	}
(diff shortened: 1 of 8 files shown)


dungbanguyen:
All three review comments were right and are fixed in the latest push.

**cubic, the capacity guard's message** — it named two of the three places and would have sent someone to raise a pair that cannot work alone. It now names all three, and says which way each moves, because they are not the same number:

```
FAIL: N integration packages but only 63 Redis logical dbs — raise all three together:
REDIS_DBS here, testdb.RedisDBs in backend/internal/platform/testdb/redis.go (same value),
and --databases in infra/docker-compose.dev.yml (one MORE, it counts the reserved db 0)
```

**cubic, the `RedisDBs` comment** — correct, and my wording was the bug. It said "the same number lives in ... `--databases`", which is false: `--databases` is 64 because Redis counts the reserved db 0. Setting all three equal would silently take the lane's top db away. The comment now states that `RedisDBs` and `REDIS_DBS` are the highest assignable index and `--databases` is one greater, and the gate asserts the offset rather than equality.

**CodeRabbit, connecting on db 0** — the sharpest of the three, because it defeated the test's whole purpose. go-redis `SELECT`s on the first command, so against an under-provisioned server the `SELECT` failed and the test reported "reading the server's database count" — naming Redis instead of the drift. Verified by running the gate against a stock 16-db `redis:7`:

Before: a Redis read error. After:

```
the server at localhost:56379 serves 16 databases but the lane assigns up to db 63 —
recreate the container (`make db-down && make db-up`); a container started before the
count was raised keeps the old one
```

Db 0 is the one a running `make dev` owns and no test may flush; `CONFIG GET` reads a server setting and touches no key in it, which the code now says.

Both file-comparison arms were mutation-tested too — each fails and names the right file when its number is changed.

PR #626, opened 2026-08-08
perf(test): one Postgres pool per test process, not per test

### LINKED ISSUE #539: compose/integration setup forks ~5 Postgres backends per test; sharing them needs plan-cache invalidation first

dungbanguyen:
Closes #539.

**Stacked on #625** — that one fixes a Redis-db collision this change's timing
exposes, and must land first. Review this PR's second commit.

Every integration harness opened its own `pgxpool` and closed it again for each
test, sized by `database.NewPool`'s api-process defaults — MinConns 2, MaxConns
16. A package's setup therefore forked several Postgres backends per test and
threw them away, and the lane runs packages concurrently against one server with
`max_connections=100`.

`testdb.Pool` hands out one pool per DSN per test binary instead, sized for a
test process: MinConns 0, MaxConns 4. The three exported fixtures and the
River-schema helper take it, so the app-role and owner-role connections are each
dialled once per package rather than once per test.

## What sharing costs, and what pays it

pgx's default prepares each statement on the server and keeps it for the life of
the connection. That is safe only while a connection cannot outlive the schema it
was planned against — which per-test pools guaranteed by accident.

This schema changes constantly: the customfields engine `ALTER`s record tables
mid-test, the reset drops those columns again, and `ApplyRiverSchema` creates
River's tables the first time a suite asks for a real worker. A plan that
survives any of those draws `SQLSTATE 0A000 "cached plan must not change result
type"` in whichever suite runs next, with nothing in the failure pointing back at
the DDL that caused it.

The shared pools therefore run in `CacheDescribe` mode: same extended protocol
and the same type safety, but the cached artifact is the statement *description*
rather than a server-side plan, and the driver re-describes when the server
reports it stale.

Retiring the connections at each reset also works and was tried first. The driver
holding the invariant is both cheaper and honest — it needs no list of which
operations invalidate what, and nothing for a future schema-changing test to
remember to declare.

## Measured

Same machine, Postgres restarted before each run (#482 otherwise contaminates a
second consecutive lane run):

| | baseline (`main`) | this stack |
|---|---|---|
| lane wall clock | 220.5s | **190.3s** (−13.7%) |
| sum of packages | 512.3s | **443.6s** (−13.4%) |
| `compose/integration` | 188.3s | 160.6s |
| `compose` | 74.5s | 59.2s |

Both runs: 28 packages, 1906 tests, `OK: integration passed with 0 skips`.

The heaviest package measured 152.5s with the retire-per-reset approach and
135.9s with `CacheDescribe`, which is the difference the last commit buys.

---
## Summary by cubic
Share a single Postgres `pgxpool` per test process (per DSN) instead of per test. Pools open only after schema migration and we assert they’re idle at test end, cutting lane time by ~15%.

- **Refactors**
  - Introduced `testdb.Pool`: per-process, per-DSN pool via `database.NewPool` with DSN params `pool_min_conns=0`, `pool_max_conn_idle_time=30s`, `default_query_exec_mode=cache_describe` (keeps `MaxConns` at 16); gated on `EnsureSchema` (returns `ErrSchemaNotReady` before migration).
  - Updated integration fixtures/helpers to use `testdb.Pool`; removed per-test `Close()`. Registered `testdb.AssertPoolsQuiesced(t)` first so it runs last; DSNs are redacted in errors.
  - Added integration tests to pin parity with `database.NewPool` (`jit=off`, typed-id ANY binds), verify one pool object per DSN per process, and ensure refusal before migration.

<sup>Written for commit 8d97dbcbcc837daaeb82a7d25d03d97d933d6eaf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/626?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Reliability**
  * Improved integration test database pooling and cleanup.
  * Shared database connections are reused safely and verified to be inactive after tests complete.
  * Database access now waits for schema setup to finish successfully before use.
  * Improved connection diagnostics while protecting sensitive credentials.

* **Tests**
  * Added coverage for shared pool reuse, connection configuration, schema readiness, and cleanup behavior.

### DIFF
--- a/backend/internal/compose/integration/apptest/appenv.go
+++ b/backend/internal/compose/integration/apptest/appenv.go
@@ -22,7 +22,6 @@ import (
 	"github.com/jackc/pgx/v5/pgxpool"
 
 	"github.com/gradionhq/margince/backend/internal/compose"
-	"github.com/gradionhq/margince/backend/internal/platform/database"
 	"github.com/gradionhq/margince/backend/internal/platform/deployconfig"
 	"github.com/gradionhq/margince/backend/internal/platform/jobs"
 	"github.com/gradionhq/margince/backend/internal/platform/readmeter"
@@ -97,11 +96,15 @@ func SetupAppWithOriginOptions(t *testing.T, opts func(origin string) []compose.
 		t.Fatalf("resetting database: %v", err)
 	}
 
-	pool, err := database.NewPool(ctx, appDSN)
+	// Shared across the package's tests, and deliberately not closed here — see
+	// testdb.Pool for why the connections, not the pool object, are the cost.
+	pool, err := testdb.Pool(ctx, appDSN)
 	if err != nil {
 		t.Fatalf("opening app pool: %v", err)
 	}
-	t.Cleanup(pool.Close)
+	// Registered here, before the test adds any cleanup of its own, so it runs
+	// last and sees a package that has genuinely stopped.
+	t.Cleanup(func() { testdb.AssertPoolsQuiesced(t) })
 
 	// The delivery machinery every send transport is composed with in the api
 	// role. Without it a send refuses rather than log an activity claiming a
@@ -283,11 +286,10 @@ func applyRiverSchema(t *testing.T) {
 		t.Fatal("MARGINCE_TEST_DSN not set — run `make db-up` (integration tests fail loudly, they never skip)")
 	}
 	ctx := context.Background()
-	ownerPool, err := database.NewPool(ctx, ownerDSN)
+	ownerPool, err := testdb.Pool(ctx, ownerDSN)
 	if err != nil {
 		t.Fatalf("opening owner pool: %v", err)
 	}
-	defer ownerPool.Close()
 	if err := testdb.EnsureRiverSchema(ctx, ownerPool, jobs.Migrate); err != nil {
 		t.Fatal(err)
 	}
--- a/backend/internal/compose/integration/harness.go
+++ b/backend/internal/compose/integration/harness.go
@@ -133,11 +133,15 @@ func Setup(t *testing.T) *Env {
 		}
 	}
 
-	pool, err := database.NewPool(ctx, appDSN)
+	// Shared across the package's tests, and deliberately not closed here — see
+	// testdb.Pool for why the connections, not the pool object, are the cost.
+	pool, err := testdb.Pool(ctx, appDSN)
 	if err != nil {
 		t.Fatal(err)
 	}
-	t.Cleanup(pool.Close)
+	// Registered here, before the test adds any cleanup of its own, so it runs
+	// last and sees a package that has genuinely stopped.
+	t.Cleanup(func() { testdb.AssertPoolsQuiesced(t) })
 	e.Pool = pool
 	e.People = people.NewStore(pool)
 	e.Deals = deals.NewStore(pool)
@@ -464,11 +468,15 @@ func ApplyRiverSchema(t *testing.T) {
 		t.Fatal("MARGINCE_TEST_DSN not set — run `make db-up` (integration tests fail loudly, they never skip)")
 	}
 	ctx := context.Background()
-	ownerPool, err := database.NewPool(ctx, ownerDSN)
+	// The shared owner pool, not a fresh one: every suite that needs a real
+	// worker calls this, and each call is one existence probe. testdb.Pool
+	// refuses to open before EnsureSchema has run, so a caller that reached
+	// here without a harness Setup is told so rather than served a connection
+	// older than the schema.
+	ownerPool, err := testdb.Pool(ctx, ownerDSN)
 	if err != nil {
 		t.Fatalf("opening owner pool: %v", err)
 	}
-	defer ownerPool.Close()
 	if err := testdb.EnsureRiverSchema(ctx, ownerPool, jobs.Migrate); err != nil {
 		t.Fatal(err)
 	}
--- a/backend/internal/compose/integration/searchenv.go
+++ b/backend/internal/compose/integration/searchenv.go
@@ -15,7 +15,6 @@ import (
 	"github.com/jackc/pgx/v5/pgxpool"
 
 	"github.com/gradionhq/margince/backend/internal/modules/search"
-	"github.com/gradionhq/margince/backend/internal/platform/database"
 	"github.com/gradionhq/margince/backend/internal/platform/testdb"
 	"github.com/gradionhq/margince/backend/internal/shared/kernel/ids"
 	"github.com/gradionhq/margince/backend/internal/shared/kernel/principal"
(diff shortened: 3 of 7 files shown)


dungbanguyen:
@cubic-dev-ai on `RegisterIDTypes` — the diagnosis was right about an earlier revision of this branch and is already fixed; the test you asked for is now added, and writing it turned up something worth recording.

**The omission was real, and is gone.** An earlier push did assemble its own `pgxpool` config, and dropped `RegisterIDTypes` along with `jit=off` and the opening `Ping`. `testdb.Pool` now builds *through* `database.NewPool` (`pool.go:102`) and declares only its deltas as DSN parameters that constructor already honours, so all three come back.

**On the test.** Added as `TestSharedPoolKeepsTheProductionConnectionSetup`, and mutation-tested by rebuilding the pool from a bare `pgxpool.ParseConfig` — which is exactly the regression in question. The result is worth stating plainly, because it is not what either of us assumed:

- the `jit=off` assertion goes **red**
- the `[]ids.PersonID` / `ANY($n)` bind stays **green**

The bind cannot fail here, and that is the exec mode rather than a weak assertion. Under `cache_describe` the parameter OID always arrives from the server, so pgx never consults the registered type to encode the slice — `RegisterIDTypes` only feeds `TypeForValue`, which `planEncode` reaches only when `TypeForOID` misses. So the missing registration was inert on this branch, not a live break; the live loss was `jit=off`, which `database.go` documents as 475ms of JIT behind 12ms of work on the `/search` union, and the lane had been paying it.

Both assertions are kept, with the asymmetry written down: `jit` is the one that detects the construction regressing, and the slice bind pins the idiom itself, since a future exec mode that guesses OIDs would break it and `ANY($n)` over typed ids is used throughout the record stores.

Full lane green afterwards — 29 packages, 0 skips — with `make check-backend` and `make craft-static` clean.

dungbanguyen:
@coderabbitai you're right, and you caught me doing the exact thing the rule you quoted forbids. The comment explaining why the gate had no test *was* the rationalization — "restructure it away or gate it with a test" does not have a third option, and I took one.

Fixed with `TestPoolRefusesBeforeTheSchemaIsMigrated` in a new internal test file, since as you say the state is only reachable from inside the package: save `schemaReady`, clear it, call `Pool`, assert `errors.Is(err, ErrSchemaNotReady)`, restore in `t.Cleanup`. Safe to mutate for one call — tests in a package run serially under the lane's `-p 1`, nothing here calls `t.Parallel`, and `Pool` returns before it touches the pool map, so the process keeps whatever pools it already had.

Mutation-tested by deleting the guard from `Pool`, which is the regression it exists to catch:

```
poolready_integration_test.go:38: Pool handed out a pool for an unmigrated database
poolready_integration_test.go:41: Pool on an unmigrated database returned <nil>, want ErrSchemaNotReady
--- FAIL: TestPoolRefusesBeforeTheSchemaIsMigrated
```

The rationalizing comment is gone; what remains is a pointer to where the gate lives.

Full lane green afterwards — 29 packages, 1938 tests, 0 skips — with `make check-backend` and `make craft-static` clean.

Issue #646, opened 2026-08-08
The REST read door consults MCP-SESS-READS but charges nothing


tinnhlt-gra:
`Gate.AdmitRead` refuses a REST read once the window is spent, and nothing on that door ever charges a record back. A credential that reads only through `/v1` therefore never accrues toward the bound at all: the MCP→REST direction is closed (a window spent through `/mcp` refuses on `/v1`), the REST→anything direction is not.

This is the read half of the gap #623 already names. C2 closed the **mutating** half — `compose/agentgate.go` now charges `agents.Registry.ChargeAdmittedCall` before the handler and `ChargeEffect` after it, so writes, egress and calls bind on both doors — and deliberately did not extend that to reads, because a read's charge is **per record** and the REST handlers answer their own pages; there is no single point on that door where the record count is known.

**Fix shape.** The record count has to come from where records leave the surface, as it does on the MCP door (`newWireRecord` → `envelopeFacts.servedCount`). Either the REST handlers report what they served through a shared seam, or the response is counted at the edge — the second is cheaper and wrong in a way that matters (a page of 1 is not a page of 200 to a caller, but the counter would see bytes rather than records).

**The derivable form of the obligation:** a fitness test asserting that every counter the gate READS on a door has a charge point on that same door. That is what would have caught this and the mutating half together, rather than one at a time.

Issue #645, opened 2026-08-08
An approval's workspace_id is answered as the zero uuid on every read


tinnhlt-gra:
`GET /v1/approvals` and `POST /v1/approvals/{id}/approve` answer `"workspace_id":"00000000-0000-0000-0000-000000000000"` on every approval, while the stored rows carry the real one.

**Cause.** `modules/approvals/inbox.go`'s `columns` constant never selects `workspace_id`, so the scanned `row` has no such field and the contract's UUID serializes as its zero value.

**Why it matters beyond cosmetics.** The field is published in the contract, so a client that branches on it — or an operator reading an export — is reading a value that names no workspace. It is also the kind of wrong that a reader assumes is right, because a uuid-shaped zero looks like a real answer rather than an absent one.

**Provenance.** Found by a manual UAT of the C2 quota ladder (#PR pending) while reading approval rows back over HTTP. `inbox.go` is not in that branch's diff — this is pre-existing on `main` and unrelated to the session-removal work, so it is filed rather than folded in.

**Fix shape.** Either select the column and carry it on `row`, or stop publishing a field the read model cannot answer. A fitness test that every contract-declared field of `Approval` is populated by the read model would keep the next one honest.

PR #647, opened 2026-08-08
feat(agents): the session's implicit bound becomes four explicit counters, a client publishes its own identity, and the registry goes

tinnhlt-gra:
Implements **ADR-0092 / A141** §4 and §6, and carries **C-6** from the MCP roadmap.

The MCP session was silently acting as a volume boundary. ADR-0092 §6 removes the
registry so a call can land on any api replica — and makes the ordering normative
rather than advisory: the per-Passport counters must be live **before** the
registry goes, because removing a bound and adding its replacement in the wrong
order leaves the surface briefly unbounded in production.

Three changes, in that order.

## 1 — the bound becomes explicit

`platform/readmeter` becomes `platform/agentquota`: one meter, five counters
(`reads`/`writes`/`egress`/`calls`/`cost`), per Passport, per window.

**Which counter a call spends is derived, never listed** — `egress`-flagged →
`egress`, read-only → `reads` per **record**, otherwise → `writes`, and every
admitted call also spends one of `calls`. BYO-LIM-9 asks for exactly this ("the
row is defined by the egress flag, not by an enumerated list of names"), and the
ordering matters: every egress tool here also mutates, so asking "does it write?"
first would leave the 20-call send ceiling — the tightest one on the surface —
permanently unspent while the 200-call write quota absorbed every send.

**The gate reads and refuses; the surface charges.** A registry that could do
both could admit itself. One rule decides what a failed charge costs: *refuse
only while nothing has happened yet.*

**The §2.4 ladder's two halves are told apart by a type.** `reads`/`writes` are
step-ups: the call is refused **and** the question is put to the human who lent
the passport, whose approval widens that window by one more allowance. `egress`
and `calls` are hard stops that reach no inbox at all, because no approval lifts
them and a question nobody's yes can answer is worse than a plain refusal.
`cost` is soft — it refuses nothing and says so on the answer.

**The release is decidable by the lender and nobody else.** Not an admin, not
the workspace owner: an agent's ceiling is its granting human's authority, so
the only person who can widen what it may be handed is the person whose reading
it is doing. That is a `selfOnlyKinds` entry plus an empty `decisionGrants` set —
and the two are one decision, which is why the fitness test now derives the
empty-set exemption from the self-only clause rather than granting it by name.

`MCP-SESS-RATE` is not here and is not forgotten: `compose/mcpedge.go` already
bounds tool-call volume at 240/min per passport, before authentication costs
anything.

## 2 — CIMD, with DCR retained for the window

A `client_id` that is an HTTPS URL **with a path** resolves by fetching the
client's own metadata document. The fetch is the dangerous part — an
unauthenticated URL, dereferenced by this server — so every guard is named with
the attack it closes: `netguard` refuses a non-public address on the **resolved**
IP inside the socket's Control hook; redirects are **not followed** (a followed
hop is a second URL the caller chose, which the `client_id` equality check no
longer covers); 64 KiB; 5s; `application/json`; and the document's own
`client_id` must equal the URL **byte for byte** — no normalizer, because a
normalizer is a second reading of one value.

A validated document becomes an ordinary `oauth_client` row with
`created_via = 'cimd'`, because `oauth_grant`'s foreign key onto that table is
what makes a connection revocable. DCR keeps working and stays advertised
alongside `client_id_metadata_document_supported`.

The consent screen names the **host** the authorization goes back to, and says so
again when that host is on this machine — a metadata document can prove what a
client calls itself and cannot prove which program holds a loopback port.

## 3 — the registry goes

`agents/session.go` deleted. **Both eras** lose the session, not only the modern
one: the id was never authority (every call re-authenticates on its Bearer
passport), `Mcp-Session-Id` is optional in `2025-11-25`, and keeping it for one
era would keep the replica pinning ADR-0092 exists to remove — for the era with
the most clients. `DELETE /mcp` answers 405 in both framings with one sentence.

---

## What three review passes and two live UATs found

Six defects, none of which a green suite showed. Two reviewers found the same
one independently, and the live UAT reproduced it on the stack.

1. **`cmd/api` panicked at boot with no AI configured.** `resolveModelPath`
   answers nil for a supported deployment shape that every other consumer
   guards. `make dev` uses `--ai-fake`, so neither the gate nor a local UAT
   could see it.
2. **A target-less staging deduplicated nothing, and a human's "no" was never
   remembered.** `insertProposalInTx` writes NULL for a kind with no target;
   all three probes compared that column with `=` against the zero uuid, which
   never matches. An agent looping on a refusal staged one approval row *per
   call* — the UAT produced five identical rows — and a rejection was invisible
   to the next call. Fixed with `IS NOT DISTINCT FROM`, bound to the same NULL
   the insert writes.
3. **The step-up identity had to name the passport.** Without it two agents lent
   by two different humans ask a byte-identical question; the second joins the
   first's row, and whichever lender answers releases only the other's window.
4. **The call ceiling could burn a human's approval.** It charged inside
   `handle`, after `RedeemAndMark` had already committed its own transaction
   consuming the approval — so a Redis blip refused a call that never ran and
   could never be redeemed again. It charges before any redemption now.
5. **The REST door read the new counters and never paid them**, so a credential
   that never touched `/mcp` was unbounded on writes, egress and calls — while
   the comment claimed the opposite. It charges where the MCP door charges. The
   **read** half stays open for a real reason (the charge is per record and no
   single point on that door knows how many a handler served) and is now named
   in the source with #646 rather than implied away.
6. **A CIMD `http.Transport` per fetch** leaked a socket and two goroutines per
   request with no `IdleConnTimeout` to reap them; one client naming a fresh path
   each time exhausts the process.

Plus: the step-up quoted the *effective* limit as the delta an approval grants
(after one release it said "another 4000" where approving adds 2000); a refusal
promised a human had been asked on the door that stages nothing; the withholding
branch tested `ReadOnly()` where it reasoned about effects, which would have
withheld a read-scoped send that had already left; `asCount` accepted trailing
garbage against its own comment; the cost ceiling pro-rated against the default
window rather than the meter's; and a CIMD refusal told the *operator* nothing
at all (the caller learns nothing on purpose — this server must not be a network
prober reporting back to the prober — which left an operator with no signal).

## The defect only a live database could find

The staged step-up's window has to be a **string**. The approvals engine's
proposal identity is a map of string values matched against the stored payload by
jsonb containment, so a numeric window was refused at staging: every step-up
silently failed to stage, the refusal stood alone, and no human was ever asked.
No unit test against a stub could see it — the integration lane did, first call.

## Proof

- **`make check` + `craft static --strict` + the full integration lane** green
  (`OK: integration passed with 0 skips`, 1,965 tests).
- **The ladder, per rung, in the lane**: refused → staged against the passport
  that was refused → approved through the real HTTP surface → the same passport
  reads again; one release is one allowance; a hard stop reaches no inbox.
- **Manual UAT on a live stack**, transcripts in the comments: no
  `Mcp-Session-Id` in either era and `DELETE /mcp` = 405 × 3 versions; the
  step-up loop end to end; the egress hard stop staging nothing; and CIMD
  authorize → consent → token → `/mcp` against a genuinely public document, with
  each guard proved by the refusal it produces — including a listener showing
  **zero** outbound connections for a loopback URL.

## Not an aicert change

No tool registered or retired, no prompt or corpus text edited, so the catalog
the band sees is byte-identical and no record's pin moves. Verified against
`origin/main`, not assumed.

## Left filed rather than folded in

- **#645** — an approval's `workspace_id` answers as the zero uuid on every read
  (pre-existing on `main`; `inbox.go` is not in this diff).
- **#646** — the REST read door consults the bound and charges nothing.

---
## Summary by cubic
Replaced the session’s implicit bound with explicit per‑passport quotas (reads, writes, egress, calls, cost), added step‑up approvals, introduced Client ID Metadata Documents, and removed the session registry so requests can hit any replica and quotas apply consistently across REST and MCP. Charging now happens only when work actually runs.

- New Features
  - `internal/platform/agentquota`: five per‑passport counters — `reads`, `writes`, `egress`, `calls`, `cost`. Gate refuses via `auth.WithQuota`; both REST and MCP charge; `egress` outranks writes; every admitted call spends `calls`; `cost` is soft and only annotates responses. Model token spend flows from the AI router into `cost` (embed lane included).
  - Step‑up approvals: on `reads`/`writes` overflow, a `quota_release` is staged to the passport’s lender; approval widens that window. `egress` and `calls` are hard stops. Approvals include the passport in their identity and release the exact window.
  - MCP session registry removed and `Mcp-Session-Id` dropped in both eras; `DELETE /mcp` returns 405. All calls authenticate per request and can land on any API replica.
  - Client ID Metadata Documents: a `client_id` that is an HTTPS URL with a path is fetched under strict guards and persisted as `oauth_client` (`created_via = 'cimd'`). Discovery advertises `client_id_metadata_document_supported`. The consent screen shows the redirect host and warns on loopback.

- Bug Fixes
  - Quota charging correctness: writes charge only on 2xx; `calls` charge only when a handler runs (before the handler if nothing is committed, after redemption on approved retries); invalid/replayed approval tokens don’t spend; split updates charge the effect; REST and MCP both charge the same meter.
  - Stability and correctness: API boot works with no model runtime; negative Redis counters are treated as corruption; `RebindFrom` no longer carries a stale cost ceiling; strict numeric parsing and proper cost pro‑ration; window‑rolled releases judged in the right branch; clearer operator errors.
  - Step‑up robustness: correct dedup/decline with NULL targets (`IS NOT DISTINCT FROM`); include passport in identity; a call with no passport is not staged and the refusal stands; “nothing pending” copy; step‑ups cannot be edited.
  - CIMD robustness: reuse a bounded `http.Transport`; down migration deletes in dependency order and revokes CIMD connections; tests now exercise fetch guards correctly with a positive control.
  - Tool registration safety: registered specs now own their bytes; input schemas are copied on register to prevent later mutation affecting `tools/list` and validation.

<sup>Written for commit 125c638a2eccfc3226ab8e300b3edc5d4c3e4b36. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/647?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added per-passport quotas for agent calls, reads, writes, egress, and model usage.
  - Quota limits for eligible actions can request a one-time human-approved allowance.
  - Added support for validated HTTPS OAuth Client ID Metadata Documents.
  - OAuth consent screens now show redirect destinations and warn about loopback addresses.
- **Changes**
  - MCP connections no longer create or require session IDs; session deletion returns `405`.
  - Added clearer guidance when quota limits block an agent action.
- **Documentation**
  - Updated agent-tool and quota behavior documentation.


tinnhlt-gra:
## Manual UAT — CIMD, on a live stack

Run against `DEV_SLUG=c2 make dev` (api hit **directly** at `:18274`, never the Vite proxy — it coalesces duplicate headers and would make a header-shaped check lie). Tokens and codes redacted; every URL is throwaway fixture data.

### The happy path, against a genuinely public document

A CIMD document published on jsonblob.com (real HTTPS, real path, `application/json`). Worth recording: **a gist raw URL does not work** — `gist.githubusercontent.com` serves `text/plain`, which the media-type check refuses, and the caller cannot tell that apart from an unregistered client. The UAT script now says so.

```
GET /.well-known/oauth-authorization-server
  → client_id_metadata_document_supported = true
  → registration_endpoint = http://localhost:18274/oauth/register     # CIMD forward, DCR retained

GET /oauth/authorize?client_id=https%3A%2F%2Fjsonblob.com%2Fapi%2FjsonBlob%2F019fe217-…
  → 302 /#/oauth-consent?…   Set-Cookie: crm_oauth_consent=…; Path=/oauth/authorize; Max-Age=300
GET /v1/oauth/consent-request
  → {"client_name":"C2 UAT Client",…}          # the name shown to the human comes from the fetched document
POST /oauth/authorize  (nonce cookie + form nonce, Sec-Fetch-Site: same-origin)
  → 302 http://127.0.0.1:9876/callback?code=<REDACTED>&state=uat-c2
POST /oauth/token  client_id=<the document URL>
  → 200 {"access_token":"<REDACTED>","scope":"read","token_type":"Bearer"}
POST /mcp  tools/list  → 200, 18 tools — the read-scope surface only, matching the lent passport
POST /mcp  tools/call list_pipelines → 200, real data
```

```
client_id           | https://jsonblob.com/api/jsonBlob/019fe217-af67-7944-9bef-1af3bf6c03c9
client_name         | C2 UAT Client
redirect_uris       | {http://127.0.0.1:9876/callback}
created_via         | cimd
metadata_expires_at | 2026-08-08 16:04:12+00   (expiry_in_future = t)
```

The 5-minute TTL is the `cimdMinCache` floor, correct for that document's `Cache-Control: public, no-cache`. **Caching observed:** first authorize `duration_ms=750` (fetched), second `duration_ms=10` (not refetched).

### Each refusal, proving one guard

| case | result |
|---|---|
| document claims a **different** `client_id` | `400 invalid_client`, 1.34s — fetched, then rejected on the byte-for-byte equality. No row created |
| served via a **302 to another host** — built so a *following* server would have succeeded (the redirect target names the redirecting URL as its own `client_id`) | `400 invalid_client`, 0.53s. The second host was never dereferenced |
| `https://jsonblob.com` with **no path** | `400 invalid_client`, **0.0086s — no network at all**. Not CIMD → falls through to the table |
| `https://localtest.me/doc.json` (resolves to 127.0.0.1), `:9999`, and `https://169.254.169.254/…` | `400 invalid_client` in 4–70ms each |
| `redirect_uri` not in the document | `400 invalid_request` — "redirect_uri is not registered for this client", answered to the caller, never redirected |

**The loopback case has a positive control.** A listener on `127.0.0.1:9999`, verified to log real connections, recorded exactly one — the control's own curl:

```
listening on 127.0.0.1:9999
CONNECTION ACCEPTED from ('127.0.0.1', 49242)   ← the control. Nothing from the api, ever.
```

The refusal happens inside the dialer's Control hook, on the resolved IP, **before `connect(2)`**.

### Nothing leaks back to the prober

All five refusal shapes produce the identical body — `{"error":"invalid_client","error_description":"unknown client_id"}` — with no "connection refused", no "no such host", no "dial tcp", no status, no content-type, no hostname. Over the whole run:

```
grep -icE "connection refused|no such host|dial tcp|i/o timeout|jsonblob|localtest|169.254" .tmp/dev/c2/dev.log
0
```

The only distinguishable channel is latency, which is inherent to doing or not doing the fetch.

**One thing this run changed in the code:** it showed the *operator* had no signal either — every CIMD failure was one `status=400` line and nothing more. A refused document now logs its reason server-side (`WarnContext`, with the `client_id` and the error), which is the one place that reason may exist.

tinnhlt-gra:
## Manual UAT — the quota ladder and the session removal, on a live stack

Two runs. The first found a **FAIL** and a copy defect; both are fixed in this branch and the second run re-proved them against a rebuilt binary. Both transcripts matter, so both are here.

### Run 1 — what it caught

Against a binary built before the fixes. Sections 1 and 3 passed; **section 2.4 FAILED**:

```
                  id                  | status  |     dh     | obs  | lim
--------------------------------------+---------+------------+------+------
 019fe21a-4ebf-73a3-b3a3-cb58fed4dab7 | pending | a171242b7a | 2001 | 2000
 019fe21a-723c-7006-8806-77504e1013da | pending | a171242b7a | 2001 | 2000
 019fe21a-726d-7057-9857-3cec4d959f36 | pending | a171242b7a | 2001 | 2000
 019fe21a-728e-75cc-b5cc-b4553e732a76 | pending | a171242b7a | 2001 | 2000
 019fe21a-72b1-7158-a158-c7a5e9f93607 | pending | a171242b7a | 2001 | 2000
```

Five refused calls, five identical questions — the UAT script's own named "wrong if" (*a looping agent will bury its human*). It also caught the second question offering **"another 4000"** where approving adds 2000. Both are the defects the review passes had just found; this is the same bugs, seen from the wire.

### Run 2 — the fixed build, **12/12 PASS**

Provenance checked first: the api process started at `23:08:39`, after `approvals/staging.go` was last modified at `22:58:11`.

**Property 1 — one question per crossed window.** Six refused `search_records` calls, all quoting the same approval id:

```
 pending_rows_for_our_passport
-------------------------------
                             1
```

**And under concurrency** (added; not in the script, because sequential calls never exercise the identity advisory lock) — ten simultaneous crossings on a third passport:

```
for i in $(seq 700 709); do ( curl -sS -X POST "$API/mcp" … ) & done; wait
grep -oh 'approval [0-9a-f-]*' burst_*.out | sort | uniq -c
  10 approval 019fe229-f97a-775a-a75a-abce2d0dd504     ← one question
```

**Property 2 — a human's "no" is durable.** Rejected the step-up, then three more refused calls: **no new pending row**, and the agent gets a plain quota refusal rather than "a human has been asked". Final table for that passport: one `approved`, one `rejected`, nothing else.

**Property 3 — the sentence quotes what a yes buys.** The second question:

> This agent has been handed 4000 records against a limit of 4000 for this window (most recently through search_records). Approve to let it continue for **another 2000**.

`payload: {"observed":4000,"limit":4000,"allowance":2000,"bucket":"20673"}` — the effective limit where it means the ceiling, the configured allowance where it means the grant. The pre-fix row is still in the same table saying *"another 4000"*, which is the defect it replaces.

**The rest of the ladder.** Approve → the byte-identical call answers (3 records, counter `2000 → 2003`, `released=1` on a separate key so the observed count is untouched). Cross again → refused with a **new** id. Egress at 20/20 → refused with "no approval lifts it", approval count `8 → 8`.

**A control for that last one**, because "stages nothing" only means something if the tool stages something when it can — dropped egress to 19 and repeated the identical call: it staged (`8 → 9`, one `send_email`/pending row). The hard stop, not the tool, is what asks nobody.

### Session removal

```
POST /mcp initialize (2025-11-25)                → 200, grep -ci 'mcp-session-id' = 0
POST /mcp tools/list (2026-07-28, no handshake)  → 200, full catalog, grep -ci 'mcp-session-id' = 0
DELETE /mcp × {none, 2025-11-25, 2026-07-28}     → 405 405 405
    "This server establishes no session, so there is none to close."
DELETE /mcp with a fabricated Mcp-Session-Id     → 405, same body   (never 404)
```

### Old binary and new, in one table

| passport | status | count | written by |
|---|---|---:|---|
| `019fe217-24cd-…` | **pending** | **5** | pre-fix |
| `019fe224-15e0-…` | approved / rejected | 1 / 1 | fixed |
| `019fe229-be65-…` | pending | **1** | fixed — from **10 concurrent** crossings |

`grep -ciE 'level=ERROR' dev.log` → `0`, and zero "putting a step-up in front … failed" lines: every refusal above is the intended path, not the silent fallback.

**One thing this run changed in the code.** Reading the two message layers together as an agent would, a *declined* releasable counter was told both "no approval lifts it" (outer) and "it may continue once the person who connected it releases the window" (inner) — each true of its own layer, contradictory composed. Fixed in `fb730e67`: what is true of both is that nothing is pending, so that is what it says.

tinnhlt-gra:
@coderabbitai review

tinnhlt-gra:
@coderabbitai review

tinnhlt-gra:
## cubic review — disposition

Thirty findings across two rounds. **Eleven were real defects and are fixed**; the rest are answered below, including the four I disagree with and why.

### Fixed — the charging was wrong in three places, all in the same direction

I introduced these. A quota must count what an agent **did**, and each of these let a caller spend a bound on something that never happened:

| # | finding | what it cost |
|---|---|---|
| routes.go:43 | the REST-admission registry had **no charger at all** | both charge points were no-ops on that door; a credential that never touched `/mcp` was unbounded on writes/egress/calls |
| agentgate.go:78, agentgatestaging.go:90 | approved 🟡 retries bypassed charging | presenting an approval was the way to spend nothing |
| agentgate.go:222 | `update_record`'s field split returned past the charge | the most common REST mutation cost nothing |
| routes.go:60, agentsplit.go:62, agentgate.go:81 | a **rejected** mutation and a **staged** one both spent a write; an **invalid approval token** spent a call | a caller could exhaust its own ceiling on requests that executed nothing |

The effect is now charged on a **2xx**, read off what the handler actually wrote. The call is charged where it is known to **run**: before the handler when nothing has committed (refusable there), after redemption otherwise (**absorbed** there — refusing would burn the human's approval on a call that never happened). Those two cases are a named type rather than a bool, so a caller cannot pass the wrong one.

### Fixed — four small ones, each wrong in the direction that matters

- **meter.go:110** — `RebindFrom` kept a stale cost ceiling when the source had none, so a rebound meter judged cost against the composition it used to be in.
- **meter.go:364** — a negative Redis counter read as a count, putting an agent **under** a bound it had already passed. It is corruption; this meter only ever `INCRBY`s.
- **tracing.go:167** — the embed lane spends the workspace budget and was never charged to the agent's share, so a retrieval-heavy agent was the one shape the soft cost counter never saw.
- **oauth_cimd.go:318** — a near-`int`-ceiling `max-age` overflowed the `Duration` and came back negative, so the **longest** cache anyone asked for became the **shortest** offered.

### Fixed — three tests that could not fail

- **release_test.go:55** and **quotarelease_test.go:80** — both named a rolled-window release and never reached that branch: one because the nil-Redis guard sat above it (the window is an argument fact and is judged with the others now), one because the stub could only answer `applied=true`.
- **admitquota_test.go:251** — counted the counters consulted without checking **which**, so a regression that asked `Reads` twice and dropped the call ceiling would have passed.

### Fixed — copy and comments

**explain.go:52** — the step-up no longer quotes an approval id beside the words "do not send an approval_id"; it is the identifier a caller told that would reach for, and it opens nothing here. **explain.go:66** was already fixed before this round (a declined releasable counter no longer borrows the hard stop's words). **explain.go:62** — em dash out of user-facing copy (VOICE-RULE-5). **authority.go:60** — `quota_release` has a label in every shipped locale. The stale "read bound" comments on **server.go:231**, **serveroptions.go:218** and **counter.go:106** now describe all five counters, and the window is documented as **fixed**, which is what an operator needs to answer "when does this agent read again".

### Not taken, with reasons

- **staging.go:200** — `IS NOT DISTINCT FROM` does **not** match a NULL row against a non-null `$3`; it is equality that treats NULL=NULL as true. Target-scoped kinds are unaffected. (The change was needed for the opposite case: a target-**less** kind whose NULL never matched the zero uuid.)
- **meter.go:270** — `Read` and `Consume` are not atomic, so a concurrent burst can overshoot a hard ceiling. That is inherent to the split this design turns on: the gate reads and refuses, the surface charges, and a registry that could do both could admit itself. The spec's own model is that crossing a threshold refuses the **next** call.
- **quotarelease.go:88** — a transient meter failure leaving an approved step-up unreleased **self-heals**: only a *rejection* is remembered, so the agent's next call stages a fresh question and the lender approves again. The error is now framed like the sibling effect branch ("approved, but widening the window failed") so a human is not told to re-decide a committed row.
- **registration.go:93** — refusing a bare `{"type":"object"}` output schema would change what extension tools may register, which is ADR-0069's contract and not this PR's to move.

**registration.go:41** is taken — the registered spec now owns its own bytes, so a tool holding a reference to the RawMessage it registered cannot rewrite what `tools/list` advertises for every later request. And the ladder tests' clock exposure: they use the default window now, so a run crossing an hour boundary no longer resets the counter mid-test.

Issue #658, opened 2026-08-08
A composed binary is READY against a database whose extension migrations were never applied


tinnhlt-gra:
## The problem

A composed binary started against a database whose **extension** migrations have not been applied
becomes **ready** and publishes routes and jobs that then fail with undefined-table errors.

`AssertRuntimeRole` already established that a deployment-shape mistake belongs in a readiness check
(`backend/internal/compose/dbrole.go`, wired at `backend/cmd/api/main.go:85`,
`backend/cmd/worker/main.go:75` and `/readyz` via `backend/internal/compose/routes.go`). Extension
migration state has no equivalent: `/readyz` checks the pool and the runtime role and nothing else, and
API boot records the extension inventory without verifying that the schema those extensions need exists.

## Reproduction

1. Apply **core** migrations only against a fresh database — i.e. any state where
   `backend/migrations/core/**` is at head but the composed unit's namespace has not been applied. The
   easy version: drop `schema_migrations_ext_crm_demo` and `ext.ext_crm_demo_note`.
2. Boot the composed api (`make dev` with `extensions/crm-demo` present).
3. `GET /readyz` → **200 ready**. The route table is mounted; the boot log reports the extension
   inventory and the mounted-route count.
4. `POST /v1/ext/crm-demo/notes/list` → 500. The handler runs
   `SELECT … FROM ext.ext_crm_demo_note` (`extensions/crm-demo/notes.go:47`) against a table that does
   not exist.
5. The heartbeat dispatcher's child does the same on its cadence
   (`extensions/crm-demo/heartbeat.go:64`), so the failure is also a recurring job fault, not only a
   request-time one.

In a rolling deploy this is the ordinary window: new binary, old schema.

## Why this was filed rather than fixed in the tier PR

The check is not the two-line addition it looks like, and the reason is a **grant**, not effort.

The runtime role cannot see migration state. The tracking tables carry **no grants at all** —
verified against the dev database:

```
=> \dp public.schema_migrations*
 Schema |              Name              | Type  | Access privileges
--------+--------------------------------+-------+-------------------
 public | schema_migrations_core         | table |
 public | schema_migrations_custom       | table |
 public | schema_migrations_ext_crm_demo | table |
```

So `margince_app` — the role `/readyz` holds and the only role a runtime process may hold
(`AssertRuntimeRole` refuses anything wider) — cannot read `schema_migrations_ext_<name>` to compare
the applied versions against the composed units' embedded `migrations/` FS.

[...]

Issue #657, opened 2026-08-08
An extension route answers 500 on malformed arguments AND on legitimate runtime states; the surface needs a small set of published refusal classes


tinnhlt-gra:
## The problem

An extension's mounted REST route answers **500** when a caller sends malformed arguments. It should answer a 4xx: bad input from a caller is not a server fault, and a 500 tells an operator to go looking for a broken server when nothing is broken.

Found during the extension tier's click-through UAT (`extensions/crm-demo`), where posting an argument the unit's handler rejects produced a 500 plus an `"unhandled error"` log line per call.

## Why it was not fixed in the tier PR

The fix is not local. A unit's handler lives behind a **stdlib-only published surface** (`backend/pkg/extension/**`, enforced by the `pkg-purity` depguard rule and `TestPublishedSurfaceIsPure`), so it cannot reach `apperrors` to return a classified sentinel. Closing this needs two pieces:

1. a new **published** error sentinel on `backend/pkg/extension` that a unit returns for "the caller's arguments are wrong", and
2. a mapping in `backend/internal/compose` from that sentinel to the appropriate `httperr` classification, so the route answers 4xx and the agent path answers its equivalent refusal.

That is an addition to a frozen published surface. The tier PR already added two such fields (`SecretsRequest`, then `RbacAction`), each of which changed what a unit author must write, and a third wants its own reviewed change rather than being folded into a fix round late in a large branch.

## What closing this looks like

- Decide the sentinel's shape and name on the published surface.
- Map it in the compose adapter for both the REST route and the tool-invocation path — a caller-argument fault should read the same on both.
- Cover it with a test that posts malformed arguments to a mounted extension route and asserts a 4xx with no `"unhandled error"` log line.

## Related

- #627 — `Runtime.Tx` bypasses overlay-mode datasource routing
- #628 — no per-unit runtime database role
- #651 — customfields' `--schema-dsn` posture
- #656 — a cold-start install has no agent seat, so no extension scheduled job can run

---

## Confirmed still live at `feat/extension-tier-capabilities` @ d35f45d2 (independent whole-branch review, 2026-08-09)

Codex re-derived this independently and named the concrete paths, which the description above did not:

- `extensions/crm-demo/notes.go:90` — `errors.New("crm-demo: a note needs a body")`, and `:92`'s
  `fmt.Errorf` for the length bound. These are **caller mistakes** returned as unclassified errors.
- `extensions/crm-demo/notes.go:159` — the strict decoder in `decode[T]` wraps
[...]

Issue #656, opened 2026-08-08
No product path creates an agent seat, so no extension scheduled job can run on a fresh installation


tinnhlt-gra:
Found by Task 14's UAT of the extension tier (`feat/extension-tier-capabilities`). Recorded here rather than fixed in that branch because the fix is a product decision about seat bootstrapping, not something the extension tier owns.

## What happens

On a freshly bootstrapped installation, **no extension scheduled job can ever run**. Every tick fails, three times, once per cadence interval, forever.

```
worker | ERROR jobs: a worker failed with an unclassified cause
         err="compose: the principal this extension job was enqueued under is no longer live in this
              workspace: the dispatcher recorded no actor for this workspace"
worker | level=INFO msg="jobexecutor.JobExecutor: Job errored; retrying" job_kind=ext_crm_demo_heartbeat_ws
… ×3
```

## Why

`compose.extensionJobActor` (`backend/internal/compose/extjobsrun.go:151-168`) resolves the tick's initiator as the workspace's agent seat:

```sql
SELECT id FROM app_user
 WHERE is_agent AND status = 'active' AND archived_at IS NULL
 ORDER BY created_at LIMIT 1
```

A workspace with none is enqueued carrying a zero principal, and `deriveAuthority` then fails the attempt deliberately — that part is correct and well argued in the code: a tick that quietly did nothing because its actor was deactivated is indistinguishable, in River and in every gauge, from one that ran and found nothing to do.

The problem is upstream of that. **Nothing in the product ever sets `is_agent = true`.** The two write paths that create an `app_user` — `identity/users.go:93` (invite) and `identity/installation.go:183` (bootstrap admin) — both leave it at its column default of `false`. The only writers in the tree are test fixtures (`modules/overlay/testsupport_integration.go:231`). The contract exposes `is_agent` as a read-only field on `User` (`backend/api/crm.yaml:14021`) and offers no way to set it.

So the seat the dispatcher requires cannot be created through any endpoint, screen, seed or migration.

## Reproduce

```bash
make infra-up && make migrate && make composition
# bootstrap an installation (organization + bootstrap_admin in margince.yaml), then:
psql -c "select id, email, is_agent from app_user;"
#  → every row is_agent = f

# with any unit declaring a scheduled job enabled under extensions/, start the worker
# and watch the tick fail once per cadence interval.
```

Workaround, which is what the UAT used as an unplanned prerequisite:

```sql
[...]

Issue #652, opened 2026-08-08
Flake: company-context.test.tsx cannot find the "Review what changed" heading under CI load


tinnhlt-gra:
`CompanyContextCard refresh review > names the field each change checkbox selects` failed once in CI with:

```
TestingLibraryElementError: Unable to find role="heading" and name "Review what changed"
```

It passes locally, repeatedly, and the whole frontend suite (166 files, 1967 tests) is green on the same commit. The failing run was the full CI job under load, which points at a render that has not settled rather than at a wrong assertion — the heading appears after an async step and the query is not waiting for it.

**Provenance.** Seen on PR #647, whose diff does not touch `company-context.tsx`, `CompanyContextCard`, or any of its i18n keys. Re-run green. Filed rather than absorbed, because a re-run is not a fix and the next person to meet this will spend the same time deciding whether it is theirs.

**Fix shape.** Whichever query fetches that heading should be a `findBy*` (or wrapped in `waitFor`) rather than a `getBy*`, so it waits for the render it depends on instead of racing it. Worth a sweep of the same file for the other three assertions in that block.

The same run also logged two unrelated `TypeError: Cannot read properties of undefined (reading 'x')` lines from the legal-entity fill and voice corpus paths — those did not fail their tests, but they are noise that would hide a real one.

tinnhlt-gra:
Recurred twice consecutively on PR #793 (a change that touches nothing in this
subsystem), which is worth recording because the shape argues against "slow CI".

Both failures are the same step:

```
frontend/src/screens/company-context.test.tsx:182
TestingLibraryElementError: Unable to find role="heading" and name "Review what changed"
```

That is the `findByRole` after `fireEvent.click(refresh)`, bounded by
`SETTLE_MS = 10_000`. **A ten-second waiter expired.** Machine load can make a
300 ms render take 3 s; it does not plausibly make it take more than ten. The
more likely reading is that the work the click depends on never settled at all
under CI conditions — a stubbed `fetch` promise or a timer that resolves on a
turn the environment never reaches — and that raising the timeout would only
make the failure slower to arrive.

What I could establish from outside the subsystem:

- The file passes 3/3 in isolation locally, and passes inside the full 2115-test
  suite locally.
- The rendered DOM in the failure output shows `.company-context-shell` and
  `.company-context-hero` present, so the component mounted and the *initial*
  render completed — it is specifically the post-click review state that never
  arrived.
- Nothing on that PR touches this screen, its providers, or its fixtures. The one
[...]

PR #659, opened 2026-08-09
feat(extension): the extension tier gains migrations, routes, RBAC, secrets, jobs and a frontend — proven by crm-demo

tinnhlt-gra:
Extensions become a tier rather than a fork. A unit declares its surface in the same contract shapes core uses — an OpenAPI fragment merged into the published contracts — and registers through explicit seams: routes, RBAC vocabulary, job kinds, secrets, migrations. It never joins core's closed generated sets. Runtime handles reach unit code as a parameter at handler invocation, so a declaration stays inert data and validate-then-apply holds.

`extensions/crm-demo` is the reference unit and exercises five of the six surfaces end to end.

---

## On what this tier's isolation is, and is not

**Read this before reviewing the security surface.** This tier is built for reviewed, first-party or otherwise trusted units, and that is a load-bearing assumption rather than a caveat.

Units are ordinary Go compiled into the same process, so a unit that wanted to could read `MARGINCE_KEYVAULT_ROOT_KEY` from the environment exactly as `keyvault.FromEnv` does and decrypt any other unit's secrets, open its own database connection, or reach the network — none of which any code in this PR prevents, and none of which a published-surface design *can* prevent while the code runs in-process.

The database walls are similarly bounded. `Runtime.Tx` pins the transaction to the calling workspace and the RLS policies read that pin, but the pin is a `SET LOCAL` GUC and a unit can rebind it with `SELECT set_config('app.workspace_id', …)` through the very verbs the seam publishes. This was verified against the shipped schema as the real runtime role, and so was the obvious mitigation: re-binding before every statement is defeated by a *single* statement (a CTE that rebinds, with a sibling scan reading under the new value). No SQL-parsing defence was attempted, because a wall made of statement inspection is a wall made of guesses. Every handler also runs as the shared `margince_app` role, which holds DML on core tables, on every other unit's `ext_<name>_*` tables, and on `extension_secret`.

**So: every wall this PR builds is defence in depth against mistakes — it turns the accidental cross-tenant query, the forgotten scope and the retained handle into loud, early failures — and none of it is a sandbox against a hostile unit.** The composed set is the trust boundary. #628 (a per-unit database role) is the single change that would move any of this from convention to enforcement, and even then the in-process root-key reach remains.

Earlier drafts of `pkg/extension`'s docs and `DESIGN.md` §2 asserted stronger properties ("No statement a unit writes reaches another tenant's rows"; "Runtime is the ONLY way"). This PR corrects them at the seam, in the design, and in `docs/explanation/extensibility.md`, because a reader must not conclude this is a security boundary it is not.

---

## What the design claims, and what is delivered

| Claim | State | Mechanism |
|---|---|---|
| A declaration is inert data | **delivered** | AST gates reject a non-identifier `Handle`, package-level `init()`, call-bearing package vars, and in-unit Go subpackages |
| An empty tree reproduces the vanilla build byte-for-byte | **delivered** | zero-fragment composition is a byte copy, not a re-serialisation; pinned at two levels and re-verified in the UAT |
| A unit cannot reach another tenant's rows | **delivered against mistakes** | tenant GUC + `FORCE` RLS; see the isolation section — not a wall against malice |
| A unit cannot reach another unit's namespace | **partial** | the secrets store closes over the unit name and every statement is scoped by it; the in-process root key is outside that wall (#628) |
| Removing a unit leaves no trace | **partial** | routes, inventory and login all degrade cleanly; no `DROP OWNED BY` purge primitive exists (#628) |
| Nothing is granted silently | **not delivered** | `approvals.lock` is digested but never parsed; now stated in `DESIGN.md` §2 rather than asserted |

## Two divergences from the plan

**The reference unit's screen lives in the core frontend tree.** `extensions/<name>/frontend/` is still an unbuilt capability layer; lifting it means bundling unit-authored TSX into the SPA, which needs its own reviewed slice with a supply-chain decision, a per-unit isolation story and a CSP answer. So the `frontend/` capability is demonstrated *from* a composed descriptor registry rather than *by* a unit — five of six surfaces come from inside the unit. A consequence worth knowing: **removing a unit is a removal in two places** until unit-owned frontends exist, and the how-to says so.

**`GET`/`DELETE` are not declarable.** A served operation is a governed tool invocation and its arguments are the body, so the method validator admits only POST/PUT/PATCH — and the sibling rule forbids path parameters. crm-demo's three record operations are three POSTs on three paths. The consequences are real and worth stating once rather than rediscovering four times: an extension operation cannot be HTTP-cached, linked, bookmarked, or made a browser navigation.

## Proof of work

Two acceptance runs, both with raw output retained.

The first found five defects, including a demo screen where **every read returned `undefined`** (the routes returned the governed-tool envelope while the contract declared the bare payload) and an uninstall path that **500'd `/auth/login`** for every user holding the departed unit's grant. Both fixed.

The second is an independent re-run: **9 PASS / 1 FAIL**. The remaining failure is the removal leg's `check-q`, documented rather than papered over. It also found that any authenticated seat — including read-only, including via the agent — could **replace** the signing key, because the secrets operations declared no RBAC object; fixed, and generalised into a class rule that refuses any `write`/`draft` verb declaring no object.

Verified across both runs: the stored key is unrecoverable through the DOM, the API, the agent, the logs and a full `pg_dump`; signatures match an independently computed HMAC-SHA256; the empty-tree leg is byte-identical on every artifact.

## Review

Fifteen per-task reviews with fix rounds, plus two independent whole-branch reviews.

**Codex** found two things the per-task reviews could not: the GUC rebinding above, and a route-ownership bug where dispatch keyed on the global tool verb rather than `(unit, tool)` — so a second unit declaring an existing verb would dispatch the first unit's handler under its own published operation. Fixed and mutation-verified.

**Fable** found that both canonical docs still described the pre-branch tier, and a combination no per-task review could see: no product path creates an agent seat (#656) × crm-demo shipping enabled at a 60-second cadence = every fresh install erroring three times a minute, forever. Fixed by skip-and-gauge.

## Deferred, with reasoning

- **#628** — no per-unit runtime database role. The single change that would make the isolation story enforcement rather than convention.
- **#627** — `Runtime.Tx` bypasses overlay-mode datasource routing.
- **#657** — a unit cannot express refusal classes from a stdlib-only surface, so bad arguments and some legitimate runtime states answer 500.
- **#658** — a composed binary reports ready against a database whose extension migrations were never applied. Not built here because the check needs the runtime role to read migration-tracking tables, which widens its read surface — a reviewed decision, not a fix-round line.
- **#656** — no product path creates an agent seat.
- **#651** — customfields' `--schema-dsn` posture under the role-separation invariant.

## Notes for the reviewer

- Migrations were renumbered to `0199`/`0200` after the final `main` merge: `main` landed its own `0197`/`0198` while this branch was building. Git merged it cleanly — the collision is semantic, and it is **fatal**, since the migrator refuses duplicate versions.
- `pkg-freeze` raises an advisory on the narrowed `Tool` type and the three added published fields (`Migrations`, `SecretsRequest`, `RbacAction`). Intended: pre-v1.0.0 the surface is design-fluid, and it hard-fails from the first v1 tag.
- One known flake, unrelated and untouched by this branch: `company-context.test.tsx` under parallel load (#652). 264ms in isolation, 10s under load.

---
## Summary by cubic
Ships the extension tier end to end with routes, jobs, secrets, RBAC, and an admin UI — plus stronger safety: the API/worker refuse a superuser or BYPASSRLS runtime role, readiness probes it, and extension REST routes now return the declared payload shape.

- New Features
  - Frontend: generated `@composition/extensions` registry and `#/ext/<unit>` operations card; Settings → Extensions lists each unit’s routes/jobs/objects and links to `#/ext/<unit>`; links come from the generated registry so stale bundles say “no page in this build.”
  - Identity/Admin: `GET /v1/extensions`, `GET /v1/roles`, `PATCH /v1/roles/{key}/objects/{object}` with If-Match; roles are versioned; RBAC objects from extensions are registered and enforced on REST and MCP.
  - Runtime/Readiness: hard fail when the runtime DSN is superuser or `BYPASSRLS`; readiness includes the runtime-role probe; per-call Runtime pins the workspace for `Tx` and Secrets and is bound early in API/worker boot.
  - Secrets/Migrations/Routes/Jobs: audited `extension_secret` store; contract-derived verbs mounted at `/v1/ext/<unit>/…` with 501 for contract-only and RBAC enforcement; REST unwraps the MCP envelope and returns the declared body; composed jobs (dispatcher + workspace child) with seatless-workspace skipping and a `margince_extension_job_seatless_workspaces` gauge on the worker’s `/metrics`.

- Migration
  - Build composed binaries (`gen-composition`) for API/worker/web; CI runs both SPA lanes.
  - Run `migrate up` to apply core `0212_extension_secret`, `0213_ext_schema`, and `0214_role_version`, plus per‑unit migrations under `ext_<unit>`.
  - Use a non‑superuser runtime DSN without `BYPASSRLS`; readiness fails otherwise.
  - If no agent seats exist, extension jobs skip; watch the seatless-workspace gauge.
  - Removing a unit is one step: delete `extensions/<name>/`.

<sup>Written for commit 92b05313e9b201f02954796ee1a784eec93a266a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/659?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tinnhlt-gra:
## UAT — independent re-run (proof of work)

Executed by a subagent that did not write the code, at `cd7cf35a`. Method and raw output below; the first run's evidence follows in a second comment.


---

## Summary

| # | Step | Verdict |
|---|---|---|
| 0 | **Prerequisite** — RBAC grants via raw SQL | **PASS** |
| 0b | *(unplanned)* an agent seat must exist or the heartbeat cannot run | **PASS, F2 reproduced** |
| 1 | boot; inventory lists `crm-demo` | **PASS** |
| 2 | `#/ext/crm-demo` renders, "not connected" | **PASS** — and on the **documented** `make dev` path (F3 fixed) |
| 3 | Store a key → "connected"; sign → signature; key never displayed | **PASS** — "Connected" on screen, real signature, key unrecoverable through four channels |
| 4 | Add a note → appears; reload → persists | **PASS** |
| 5 | One tick → heartbeat row appears unprompted, naming its workspace | **PASS** (after 0b) |
| 6 | Ask the agent to list notes → it returns them | **PASS** |
| 7 | Read-only seat → list visible, `Add` gone; API refuses the write | **PASS** — the list is now observable, which it was not last run |
| 8a | Removal leg — 404s cleanly, inventory omits it, `check-q` green | **FAIL** — 404 ✓, inventory ✓, F4 ✓, but `make check-q` is **still not green** on the documented complete leg (**R2**) |
| 8b | Empty-tree leg — byte-for-byte vanilla, `check-composition` green | **PASS** — byte-identical on every artifact |

**Counts across the prerequisite and the eight steps (8a and 8b counted separately): 9 PASS, 1 FAIL,
0 BLOCKED.** Compare the first run's 5 PASS / 2 FAIL / 3 PARTIAL. The demo screen, which was
non-functional, now works end to end.

### Status of the five findings this run was commissioned to check

| ID | First run | This run |
|---|---|---|
| **F1** | Every screen read returned `undefined` | **GENUINELY FIXED.** Bare payloads at REST, `Trace-Id` header present, envelope still on the agent path. Verified against the merged contract's schemas, not just by eye. |
| **F2** | Cold start has no agent seat | **REPRODUCED AND CORRECTLY DEFERRED.** Issue [#656](https://github.com/gradionhq/margince-poc-v1/issues/656) exists, is OPEN, and the limitation is real — not silently worked around. |
| **F3** | `make dev` served the vanilla SPA | **GENUINELY FIXED.** The documented path `http://localhost:8581/#/ext/crm-demo` reaches the screen. |
| **F4** | Removing a unit 500'd `/auth/login` | **GENUINELY FIXED.** Login 200 for both seats with the orphaned grant still on the role; core grants retained; `row_scope` still fatal. |
| **F5** | Removal leg's `check-q` failed on `drift` | **NOT FULLY FIXED.** `drift` and `fe-typecheck-composed` are both fixed, and the documentation is honest about the two-place cost — but the documented **complete** leg still fails `check-q` on a third site the docs do not mention (**R2**). |

### Findings

| ID | Severity | Finding |
|---|---|---|
| **R1** | **High (new)** | **Any authenticated seat can replace the installation's signing key** — including a read-only seat, and including through the agent. The three secrets operations declare no `x-rbac-object`/`x-rbac-action`, so FR1's enforcement does not reach them. Audited, but not prevented. |
| **R2** | **Medium-High (new)** | **The documented complete removal leg is not green.** Removing the unit is a removal in **three** places, not the two the docs state: a core test hard-codes `extensions/crm-demo/crmdemo.go`. |
| **R3** | Medium (new) | **#657 understates its own scope.** Four of six routes 500 on malformed arguments, and a *legitimate runtime state* (signing with no key stored) also 500s — which the proposed `ErrInvalidArgument` sentinel would not cover. |
| **R4** | Low (new, docs) | `docs/how-to/add-an-extension.md` @ `d35f45d2` still says extension-owned tables and `/v1/ext/<name>/` surfaces "haven't landed yet" — this PR ships both. |

tinnhlt-gra:
## UAT — first run (the one that found the defects)

Kept as part of the record rather than superseded: this is the run that found the screen non-functional and the uninstall path breaking login. Both fixed before the re-run above.

# UAT evidence — the extension tier (`feat/extension-tier-capabilities` @ `738c1c1f`)

Task 14's click-through, run end to end on 2026-08-09. This file is the proof of work attached to
the PR: every step below carries the exact command, the real output, and a verdict. Nothing was
repaired to make a step pass — where the tier does not work, the failure is recorded as it happened
and the run continued.

**How the system was driven.** Two surfaces, both real:

- **The SPA** was driven headlessly with **Playwright + Chromium** (`@playwright/test` 1.62.1, the
  copy already in `frontend/node_modules`). Every screenshot in `uat/` is a real page render of the
  running app; nothing is mocked and no browser interaction is described that was not performed.
- **The API and the agent surface** were driven with `curl` — the mounted REST routes with a session
  cookie, and the agent path over the real **MCP transport** (`POST /mcp`, `tools/call`) with an
  Agent Seat Passport bearer.

**Environment.** Postgres/Redis/MinIO from `infra/docker-compose.dev.yml` (already up). The
installation ran on an isolated database `margince_dev_uat` on the shared cluster — see
[Environment note E1](#e1-the-shared-margince-database-was-unusable) for why the shared `margince`
database could not be used, and [E2](#e2-make-run-is-not-a-usable-boot-path-on-its-own) for why
`make run` alone does not boot.

---

## Summary

| # | Step | Verdict |
|---|---|---|
| 0 | **Prerequisite** — RBAC grants via raw SQL | **PASS** |
| 0b | *(unplanned prerequisite)* an agent seat must exist or the heartbeat cannot run | **PASS, with finding F2** |
| 1 | `make infra-up && make migrate && make composition && make run` → boot inventory lists `crm-demo` | **PASS** (inventory correct; `make run` itself needs flags — E2) |
| 2 | `#/ext/crm-demo` renders, "not connected" | **PARTIAL** — renders and says "No key stored", but on the *composed* lane only, and the notes card is already broken (F1, F3) |
| 3 | Store a key → "connected"; sign → signature shown; key never displayed | **FAIL in the UI / PASS at the API.** The key is genuinely never disclosed (strongest result in this run), but the screen never says "connected" and renders `undefined undefined` instead of a signature (F1) |
| 4 | Add a note → appears; reload → persists | **FAIL in the UI / PASS at the API.** Both notes were written and both survive a reload — the screen simply cannot read them back (F1) |
| 5 | One tick → heartbeat row appears unprompted, naming its workspace | **PASS** (after prerequisite 0b; F2) |
| 6 | Ask the agent to list notes → it returns them | **PASS** |
| 7 | Read-only seat → list visible, `Add` gone; API refuses the write | **PASS on enforcement, PARTIAL on the screen** — `Add`/`Remove`/`New note` are correctly absent and both caller paths return 403; "list visible" is unobservable because of F1 |
| 8a | Removal leg — 404s cleanly, inventory omits it, `check-q` green | **PARTIAL** — 404 and inventory are perfect; `make check-q` **fails** on `drift` (F5), and login **500s** for every user until the grant is dropped (F4) |
| 8b | Empty-tree leg — byte-for-byte vanilla, `check-composition` green | **PASS** — byte-identical on every artifact |

**Counts across the eight steps + the prerequisite: 5 PASS, 2 FAIL, 3 PARTIAL** (counting 8a and 8b
separately; step 3 and step 4 are counted FAIL because the step as written is a UI observation).

`git status` is **clean** at the end of this run — verified in [§8 restore](#restore-and-final-tree-state).

tinnhlt-gra:
## Demo — the tier working, end to end

One continuous Playwright take at `11e05fee`, 1:23, nothing cut or re-ordered. The dark strip is a caption overlay injected by the recording script; where it quotes a number, that number was computed live in the page at that moment. Everything above the strip is the real application.

| Time | What is on screen |
|---|---|
| 0:00 | Sign-in page; signing in as `admin@demo.test`. |
| 0:14 | **1/6** `#/ext/crm-demo` — the **Demo Notepad** screen, reached from the composed extension registry. Breadcrumb reads `ext · crm-demo`. |
| 0:20 | **2/6** A signing key is typed into the `Signing key` field and `Store key` clicked. |
| 0:29 | The status chip flips to **Connected**. The caption reports a check run in the page at that instant: the key string appears **nowhere** in the DOM (`false`) and **nowhere** in local/sessionStorage (`false`). The `Signing key` input is empty again. |
| 0:33 | **3/6** `hello-from-the-video` is signed. A real signature renders under the button: `hmac-sha256 b25ef76473352036e75e752c75f97dd6b63349bdc1a04272c0eb4b344eda4102`. |
| 0:38 | **4/6** A note is typed and `Add` clicked; the row appears at the top of the Notes list. |
| 0:50 | A **full page reload**. The note is still there — it lives in the unit's own `ext.ext_crm_demo_note` table. |
| 0:58 | **5/6** Nothing is clicked from here. Caption records 3 heartbeat rows visible and waits. |
| 1:03 | A **fourth heartbeat row arrives unprompted** (3 → 4) — `⟳ heartbeat — workspace 019fe43b-a518-7b95-be24-642be40ef8f1`. The unit's scheduled job ticks every 60 s; the screen polls every 15 s. Each row names the workspace the tick was for. |
| 1:08 | **6/6** The session cookie is cleared and the read-only seat `readonly@demo.test` signs in. |
| 1:17 | The same screen on the **read-only seat**: the Notes list renders all five rows, and `Add`, per-row `Remove`, the `New note` field and `Store key` are all **gone** (counts in the caption: 0 / 0 / 0). The status still reads `Connected` and `Sign` is still offered — that seat holds `read` on `ext_crm_demo_signing_key` but not `update`. |


**Signature verified out of band.** The `hmac-sha256 b25ef764…4102` shown at 0:33 matches `hmac.new(b'VIDEO-DEMO-KEY-0123456789abcdef', b'hello-from-the-video')` byte for byte — the screen is signing with the stored key, not echoing something.

**Not in the video, and not implied by it:** the MCP/agent path, the API's 403 on a read-only write, and step 8's removal and empty-tree legs. None are browser surfaces. The video shows the read-only *control* absent; the API refusal is proven in the UAT evidence above, not here.

**One thing the video updates.** The re-run recorded `storeKey: 1` for the read-only seat — the control was present. It is now absent (count 0), because the signing operations gained their own RBAC object. `Sign` remains, correctly: that seat holds `read` on the key object but not `update`.

tinnhlt-gra:
The recording is committed at [`docs/evidence/extension-tier/`](https://github.com/gradionhq/margince-poc-v1/blob/feat/extension-tier-capabilities/docs/evidence/extension-tier) so it outlives this comment — **[crm-demo.mp4](https://github.com/gradionhq/margince-poc-v1/blob/feat/extension-tier-capabilities/docs/evidence/extension-tier/crm-demo.mp4)** (1.0 MB, 1:23) renders in GitHub's file view. Stills below.

**Connected — the key is stored and never shown**

connected

**A real signature**

signature

**The heartbeat arrives unprompted, naming its workspace**

heartbeat

**Read-only seat — list visible, every write control gone**

read-only

tinnhlt-gra:
## CI status, and the one red check that stays red

Everything this branch can affect is green. One check is not, and it is not ours:

**`sbom` — pre-existing on `main`, filed as #670.** `jackc/pgerrcode`'s PostgreSQL licence is not on the compliance allowlist. It entered in `6af4f5e5` ("Update go dependencies (non-major)", #605, 2026-08-07) and `sbom` has failed on every `main` run since. No commit on this branch touches it. Whether to admit the licence is a policy decision, so #670 records it rather than this PR patching it.

## Two things CI found that were real, and worth recording

**Lock exhaustion under the parallel lane.** Three integration tests failed with `out of shared memory (SQLSTATE 53200)`. My first suspicion — that this branch's widening of the test reset to cover the `ext` schema was responsible — was **refuted by measurement**: that widening adds two locks to a ~630-lock transaction. The actual cause is pre-existing and much larger: one `DROP SCHEMA public CASCADE` takes **4,999** heavyweight locks against a 6,400-slot cluster table, so any two of the 16 concurrent lane packages collide and the loser fails outright. It never showed locally because a single shard has the whole lock table to itself.

Fixed by leaf-first ordering and per-extension drops — 4,999 → 653 — plus headroom in the committed compose file, labelled as headroom rather than as the fix. The lane also got ~27% faster.

**Migration numbers collided with `main`. Twice.** This branch took `0197`/`0198`; `main` landed its own and we moved to `0199`/`0200`; `main` then landed `0199_company_dossier` and `0200_approval_bundle` and we moved to `0201`/`0202`. Both merges were textually clean and semantically fatal — `dbmigrate.Up` refuses a duplicate version outright, which is why the second collision turned 18 jobs red at once rather than one.

That is not a defect in anyone's work. Nothing reserves a migration number at branch time, so any branch that sits will collide, and it fails hard rather than quietly. The cheap mitigation is merging promptly. The structural one — allocating at merge time, or keying the ledger on name rather than number — is a real change that deserves its own issue rather than being smuggled in here.

extensions/notes/note.go
```diff
@@ -0,0 +1,300 @@
...
+// maxNotesPerRead bounds one list. The screen renders the whole result, and a
+// notepad nobody pruned should still answer in one frame.
+const maxNotesPerRead = 200
+
+// note is one row, and mirrors the shape the contract's 200 response declares.
+type note struct {
+	ID        string `json:"id"`
+	Body      string `json:"body"`
+	CreatedAt string `json:"created_at"`
+}
```
lstrojny, line 120: I would demo here as well:
 * Map to a nested struct
 * Map to an enum (e.g. a status)

extensions/crm-demo/crmdemo.go
```diff
@@ -0,0 +1,99 @@
...
+// New returns the unit's declaration: inert data, holding no handle into the
+// core. Every capability arrives at the handlers below through the Runtime the
+// core mints for one invocation and releases when the handler returns.
+//
+// Every field is a literal, including the tool and job names, because the
+// operator manifest is derived from this function's AST without compiling it —
+// a named constant here would be a value the manifest reader cannot resolve.
+func New() extension.Extension {
+	return extension.Extension{
+		Name:    "crm-demo",
```
lstrojny: how about we call this extension "notes"?

extensions/crm-demo/migrations/0001_note.up.sql
```diff
@@ -0,0 +1,64 @@
...
+-- deliberately absent: it empties every tenant's rows without consulting the
+-- policy's USING clause.
+--
+-- Note what this grant is NOT, since every unit issues one: margince_app is the
+-- SHARED runtime role, so this line gives it to every unit's handlers, not only
+-- to crm-demo's. Within one workspace, another installed unit's SQL can read and
+-- write this table. That is inside the tier's trusted-unit threat model (see
+-- backend/pkg/extension/runtime.go) and it is what #628's per-unit role would
+-- close; it is stated here because this file is the template.
+GRANT SELECT, INSERT, UPDATE, DELETE ON ext.ext_crm_demo_note TO margince_app;
```
lstrojny: I think these two guarentees should be managed by the extension system, not left to individual extension authors

lstrojny (dismissed):
I like the change overall, it’s a huge step. Few findings, see below.

## Correctness and governance

### 1. Authorization must be operation-aware

Extension REST operations are RPC-shaped and may use POST even when they only
read. HTTP method therefore no longer determines whether an extension request
mutates.

Derive the classification from `Verb.RequestedScope`: `read` is read-only and
every other scope is mutating. Inject that operation lookup into the session
middleware, and recheck the human seat ceiling from the same declared scope in
the registry as defense in depth.

A blanket `/v1/ext/*` exemption would be unsafe, while forcing reads into GET
would rule out read operations with structured arguments. The same
classification should make the mutation capability unavailable to a
read-scoped operation; see finding 2.

### 2. Extension writes need a governed mutation capability

`Runtime.Tx` currently exposes arbitrary SQL while bypassing the repository's
mandatory mutation shape. Adding optional `Audit` and `Emit` methods would still
let an extension forget either half.

The published surface should instead separate:

- `ReadTx`, which opens a database read-only transaction.
- `Mutate`, whose callback performs domain SQL and must return at least one
  structured change descriptor.
- Core writes `storekit.Audit` and `storekit.Emit` for every descriptor before
  committing that same transaction.

`Mutate` must reject an empty descriptor set and expose no untracked write
primitive; otherwise a callback can still commit SQL while forgetting the
descriptor, recreating the defect behind a different interface. A descriptor
needs the namespaced entity and id, action, before/after images, and a declared
event payload. Tool calls and scheduled jobs must use the same mutation seam.

This also requires a decision about extension event types. The branch composes
`public-events.yaml`, but the runtime event catalog remains core-owned and
closed. Under P3, reconcile those contracts rather than bypassing the runtime
catalog with arbitrary event strings.

### 3. Retry support must be capability-based

The registry adds `idempotency_key` to every non-read tool, but replay assumes
evidence collected through the core datasource seam. Extension-owned records
never enter that seam, so a mutating extension tool advertises that a retry
returns the first result while its recorded result cannot pass the replay gate.

The governed mutation capability should attach evidence for every changed
extension record. Replay then needs a tool-specific visibility authorizer that:

- re-authenticates the caller and applies the current read-seat ceiling;
- requires current `read` permission on the extension RBAC object;
- runs the extension's visibility probe when the entity has finer row scope;
- supports a deletion receipt without replaying pre-deletion data the caller
  can no longer read.

This is read authority over a stored receipt, not a second request for the
original mutation scope or tier. Until it exists, mutating extension tools must
not advertise `idempotency_key`; the surface must not promise recovery it cannot
provide.

### 4. Extensions need proper error responses

Extension handlers cannot return the product's validation, not-found, conflict,
or legitimate-unavailable responses (#657), so caller mistakes and expected
runtime states become opaque 500s. The error vocabulary must be part of the
common handler surface, not only the mutation interface: read handlers need the
same vocabulary.

### 5. Readiness does not prove extension migrations are applied

A composed binary reports ready even when its extension migrations were never
applied (#658). `/readyz` must compare the composed migration set with the
database state before publishing extension routes and jobs.

The generator already refuses a unit that ships `migrations/` without declaring
and embedding that layer through `Extension.Migrations`; that case is covered
and is not an open finding.

## Frontend boundary

### 6. Keep foreign types out of the stable surface where practical

The exported `api` value exposes `openapi-fetch` through its inferred client
type, and exported components necessarily expose React types. TanStack Query is
not re-exported by the surface, but unit frontends depend on it directly as a
hosted peer, so it is a separate host-compatibility constraint.

Replace avoidable inferred foreign types with small core-owned interfaces. React
is likely part of the unavoidable component ABI. This does not make every
dependency upgrade impossible, but a breaking foreign-type change must not
silently become a breaking extension-surface change.

### 7. Close the secondary frontend boundary gaps

- Resolve `extensions/<unit>/frontend/package.json#main` canonically and require
  the result to remain under that unit's `frontend/` directory. Reject lexical
  prefix tricks and symlink escapes.
- Parse imports, re-exports, and dynamic imports with a TypeScript parser rather
  than double-quote-sensitive grep. If commented-out imports are intentionally
  prohibited, enforce that as a separate policy instead of confusing comments
  with syntax.
- Generate screen aliases with an injective encoding of the full unit name.
  The current title-casing maps valid distinct names such as `foo-1` and `foo1`
  to the same `Foo1Screen` identifier; a prefixed hexadecimal encoding avoids
  that class of collision.

### 8. The unit-owned frontend slice has not had whole-branch review

The unit-owned frontend code was committed on 2026-08-10, after both
whole-branch reviews and both acceptance runs. Those reviews therefore do not
cover it. It also supersedes the reviewed §4.5 ruling not to bundle unit-authored
frontends.

Re-review the complete frontend slice, including its generated artifacts and
supply-chain boundary, or move it into its own PR. Repeat that review after the
final merge from `main`, because the target branch is still moving.

### 9. Frontend units edit upstream files and bring dependencies into the host

A unit with npm dependencies changes the root `pnpm-lock.yaml`, so the rule that
a unit never edits an upstream-owned file no longer holds for frontend-bearing
units. `DESIGN.md` §4.6 records the exception, but the published package comment,
`docs/how-to/add-an-extension.md`, and `docs/explanation/extensibility.md` still
promise no upstream edits. The explanation document also still says units
cannot own frontends. Reconcile all canonical documentation before merge.

A unit's transitive packages enter the SPA bundle with access to the same origin
and session. The import gate checks that a dependency was declared; it does not
allowlist dependencies or require a specific lockfile review. That may be an
accepted trusted-unit posture, but it needs an explicit shipping rule. Finding 6
is about what core exports; this finding is about what a unit brings in.

### 10. The frontend RBAC object type is intentionally open-ended

The frontend RBAC object type accepts any `ext_`-prefixed string. A misspelled
extension object therefore compiles and is denied at runtime, which looks the
same as a missing grant. This is a known cost of a vocabulary that depends on the
composed set; retain it as an explicit trade-off rather than claiming extension
object typos are compile-time errors.

## Sequencing and shipping

### 11. RLS retirement conflicts with the extension tenancy design

Extension isolation currently depends on the workspace GUC, `FORCE RLS` on
`ext_` tables, and the catalog gate requiring `workspace_id NOT NULL` with the
standard policy. ADR-0091 (ratified 2026-08-06) removes `workspace_id` and the RLS
policies, and no replacement catalog obligation has been decided.

Decide the order explicitly: either merge this while the present tenant boundary
still exists and make the extension-gate redesign a named part of ADR-0091's
schema phase, or redesign this branch first. Do not leave “redesign afterwards”
as an unowned follow-up on ADR-0091's critical path.

The private `callerWorkspace` SQL constant in crm-demo is a symptom of the same
transition. While `workspace_id` exists, enforce the exact stamping shape through
a gate or core-owned mutation capability. Do not publish a long-lived helper for
an identifier ADR-0091 is about to retire.

### 12. crm-demo should not be enabled by default

crm-demo is enabled in the default composition, so every production build serves
`/v1/ext/crm-demo/*`: a reference notepad with a signing-key feature and a
one-minute heartbeat. The acceptance runs found two defects because that demo
was running as product code—a read seat could replace its signing key, and a
fresh installation produced a recurring heartbeat failure. Both defects are
fixed on this branch; they remain evidence that the reference unit has production
blast radius only because it ships by default.

Move reference/demo units to `fixtures/` or compose them only in a dedicated CI
lane. Keep default `extensions/` for capabilities intentionally shipped as part
of the product.

### 13. A per-unit database role is defense in depth, not a third-party sandbox

All extension SQL currently runs on the shared application role, so
`Runtime.Tx` can read `extension_secret` directly. #628's per-unit role should
land before the product claims database-level namespace isolation between
units.

That role is not sufficient to admit untrusted third-party code. An in-process
unit can still read the keyvault root key and application DSN from the
environment, open its own connection, or bypass the published surface entirely.
Reviewed first-party or otherwise explicitly trusted units remain the supported
model. Supporting untrusted third-party units requires an out-of-process or WASM
boundary, not only #628.

## Recorded follow-ups

- Runtime expiry has a race window. The aliveness check and the work after it
  are not atomic: a goroutine leaked by a handler can pass the check just before
  release and complete its transaction after the call returns. Closing the
  window would require request completion to wait on leaked work. The actual
  guarantee is “a runtime reused by a later call is refused,” not “no statement
  runs after release.”
- There is no published test double for `Runtime`, `Tx`, or `Secrets`; crm-demo
  hand-writes all three. The conformance kit from ADR §12 is the appropriate
  place for a supported fake or harness, and it is not part of this PR.

extensions/crm-demo/migrations/0001_note.up.sql
tinnhlt-gra: Yes ideally it should managed by the system. But we can not create role at runtime. Check `scripts/deploy/api-entrypoint.sh` 
- Migration -> using owner account 
- Runtime -> app account 

So we need somewhere to store the migration and run it at `margince-migrate up` step before api run

tinnhlt-gra:
@lstrojny — thanks, useful review. Status of all 13 findings below.

## Fixed on the branch

| # | What | Commit |
|---|------|--------|
| 3 | Mutating extension tools no longer advertise `idempotency_key`. Took the retreat you described. A caller who sends the key anyway now gets an explicit refusal instead of a call that quietly runs unprotected. | `cad5f51b` |
| 7c | Screen aliases are injective. `foo-1` and `foo1` no longer collide. Kept names readable, marked only digit-initial segments (`foo-1` → `Foo_1`); the result decodes back to the original name, so it cannot be many-to-one. `localeKeyPrefix` had the same bug and now shares the encoder — that one was worse, since copy keys are strings and would never fail to compile. | `cad5f51b` |
| 9 | Docs reconciled. The `pnpm-lock.yaml` exception is recorded in `extensibility.md` and `add-an-extension.md`; the "units cannot own frontends" paragraph is gone; the shipping rule is written down. | `cad5f51b` |
| 10 | Now stated as an explicit trade-off at `ExtensionRbacObject`: only the core half of the union catches typos at compile time. | `cad5f51b` |
| 13 | No doc claims sandboxing or third-party safety. | `cad5f51b` |
| 2, 6 | Cannot land the governed mutation capability today. Instead the package doc now says the surface is not stable, and names `Runtime.Tx` as expected to be **removed**, not joined by a sibling. | `cad5f51b` |
| — | Your inline comments: unit renamed `crm-demo` → `notes` (routes, tool ids, RBAC objects, DB names), and it now demonstrates enum + nested-struct mapping. | `cad5f51b` |

Also: the final merge from `main` you asked us to re-review against has landed
(`e4c3f568`). It surfaced a collision git could not see — both sides added a
`0205` and a `0206` under `backend/migrations/core`, so `pgmigrate` refused the
set. Ours moved to `0209`/`0210`.

## Split into their own PRs — decided, no answer needed

**#8 — we are splitting the frontend slice out.** You offered re-review here or a
separate PR; we are taking the separate PR. It also removes #6, #7a, #7b, #9's
lockfile half and #10 from this one, since all of those exist only because the
slice is here. What remains here is the backend tier, which has had a full review.

One thing that belongs with that split: **the unit's frontend tests are never
executed by any lane.** vitest's root is `frontend/`, so
`extensions/*/frontend/*.test.tsx` is not collected — 2230 tests run, 0 from
`extensions/`. They are typechecked and nothing more. The slice's own PR is where
that gets fixed.

**#1 — operation-aware authz: its own PR, next.** You are right that it is a real
hole. It touches session middleware, and we would rather it got a slow review
than a merge-day one.

**#12 — its own PR.** Agreed and easier than expected: the `extension-reference`
job already copies `fixtures/extensions/crm-hello` into `extensions/`, so it is a
move plus one line. It waits on #8 only because `notes` is the only unit with a
frontend.

**#11 — its own PR, in `margince-foundation`.** ADR-0091 lives there, so naming
the extension-gate redesign in its schema phase cannot be done from this repo.

**Extension access UI — its own PR.** Not one of your findings, but it came out of
testing this branch: nothing grants extension RBAC objects and there is no API to.
The reference unit ships enabled and every user can only ever see "you do not hold
access". We have built `GET /roles`, `PATCH /roles/{key}/objects/{object}`,
`GET /extensions` and a settings screen for it, separately.

davewilliamson:
#658 readiness/migration-state nudge

I rechecked the PR metadata before posting this: #659 is currently open, not draft, and the branch head I see via GitHub is `53e7d9afb7643259a486c557c43f0168025e0783` (`feat/extension-tier-capabilities`). If the intended stale head was `b161fd0e3a961c5c67cce69be82fcada0a64a30b`, please treat this as a request to refresh the evidence against the current exact head.

Required next artifact for #658 is still a terminal-validated, substantive source/test commit covering all of:

- retained runtime-before-bind behavior;
- exact catalog drift rejection for composed extension migrations;
- decoded digest-bound trust/provenance readback;
- real Postgres + sidecar tamper/restart evidence showing the unready/ready transitions;
- exact-head CI evidence for the commit under review;
- fresh independent review after that commit lands.

Please report the concrete commit SHA, changed files, and commands/tests run with real output, or state the executable-environment blocker explicitly. This does not touch #657 and does not route through Fred.

lstrojny:
One quick point, haven't checked all changes yet: the pnpm lock file must remain patchable, so it should not change. Can we use a gitignored alternative lock that is always regenerated?

tinnhlt-gra:
> One quick point, haven't checked all changes yet: the pnpm lock file must remain patchable, so it should not change. Can we use a gitignored alternative lock that is always regenerated?

Should not. since the pnpm lock required in docker steps - [here](https://github.com/gradionhq/margince-poc-v1/blob/eb9e57ab3cd80deb411534debf63e118113d14e9/Dockerfile.web#L24)

tinnhlt-gra:
@lstrojny  the Extensions Frontend topic moved to https://github.com/gradionhq/margince-poc-v1/pull/878

I will try to merge this PR first at the end of the days 

----
Add new Setting for extensions RBAC config 

image

PR #672, opened 2026-08-09
refactor(test): the job fan-out suites become their own package

dungbanguyen:
The lane's wall clock cannot fall below its longest package, and `internal/compose/integration` is that package by a wide margin. This moves the periodic job-dispatch suites out into `integration/jobfanout`.

## Measured, same sitting, Postgres restarted before each run

| | control (`main`) | this branch |
|---|---|---|
| `compose/integration` | 159.1s | **139.2s** (−12.5%) |
| lane wall clock | 190.7s | 185.3s (−2.8%) |

Green at 30 packages, 0 skips.

**Be honest about that second row.** The package shed 20s and the lane shed 5. Wall clock is no longer purely bound by the longest package — there is ~46s between the longest package and the lane's wall time that further splitting cannot reach. I'd stop splitting after this one unless that gap is attacked first; #671 documents what I found when I went looking for it.

## Three helpers had to be promoted, and the import rule decided where each could go

- `SeedExtraWorkspace` (seed.go) and `RetentionPassCtx` (new scope.go) are exported from this package: both sides need them, neither needs `compose`.
- The River job-runner ceremony **could not follow them**. It builds a runner, so it needs `compose` — and package `integration` may never import `compose`, because `compose`'s own white-box tests import `integration`. The compiler reports this as `import cycle not allowed in test`, which names neither the helper nor the file that moved. It becomes the `jobtest` package, for the same reason `apptest` exists.

That rule is now written into the package doc, since it is the first thing the next split will hit.

`harness.go` was one line under the 500-line ceiling, so recording the rule pushed it over. The package doc moves to `doc.go`, which is where it belongs anyway.

The suites that stayed — embed-reindex and webhook-retry fan-out, reembed — were already sharing that ceremony, so they take it from `jobtest` too.

---
## Summary by cubic
Split periodic job fan-out integration suites into a dedicated `integration/jobfanout` package to improve parallelism and reduce CI time. Also fixed flaky test failures by switching the shared test DB pool to `describe_exec` so it’s safe under schema changes. `compose/integration` runtime drops ~12.5% (159s→139s); lane wall clock drops ~3%.

- **Refactors**
  - Moved `agentscheduler`, `privacyretention`, `runner`, `runnerreap` suites to `integration/jobfanout`.
  - Promoted helpers to `integration`: `SeedExtraWorkspace` (seed.go) and `RetentionPassCtx` (scope.go).
  - Added a guard test (`backend/retentionscope_test.go`) to ensure `RetentionPassCtx` is only passed to `EvaluateWorkspace`.
  - Extracted River runner test harness to `integration/jobtest` to avoid `compose` import cycles; exposes `StartTestJobRunner`, event awaiters, and repeat-interval constants (`DispatchInterval`, `DispatchGapBound`).
  - Updated remaining suites (`embedreindex`, `webhookretry` fan-out, `reembed`) to use `integration/jobtest`.
  - Moved package docs from `harness.go` to `doc.go`. Tests-only; no production code.

- **Bug Fixes**
  - Shared test DB pool now uses `default_query_exec_mode=describe_exec` (was `cache_describe`) to avoid stale caches across schema changes, fixing intermittent XX000/0A000 errors (e.g., “cache lookup failed for type …”).

<sup>Written for commit 6fb663fbe615673cac234681b20734376dcb6855. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/672?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/internal/compose/integration/ai_call_integration_test.go
+++ b/backend/internal/compose/integration/ai_call_integration_test.go
@@ -255,7 +255,7 @@ func TestEmbedCallRetentionAgesOutOverAgeEmbeddingRows(t *testing.T) {
 	underAge := seedEmbedCall(t, e, 10)
 
 	svc := privacy.NewRetentionService(e.Pool, nil, slog.New(slog.NewTextHandler(os.Stderr, nil)))
-	if err := svc.EvaluateWorkspace(retentionPassCtx(e.WS)); err != nil {
+	if err := svc.EvaluateWorkspace(RetentionPassCtx(e.WS)); err != nil {
 		t.Fatal(err)
 	}
 
@@ -295,7 +295,7 @@ func TestAICallPayloadRetentionAgesOutContentKeepingMetadata(t *testing.T) {
 		VALUES (NULLIF(current_setting('app.workspace_id', true), '')::uuid, 'ai_call_payload', 'content', 365, 'erase')`)
 
 	svc := privacy.NewRetentionService(e.Pool, nil, slog.New(slog.NewTextHandler(os.Stderr, nil)))
-	if err := svc.EvaluateWorkspace(retentionPassCtx(e.WS)); err != nil {
+	if err := svc.EvaluateWorkspace(RetentionPassCtx(e.WS)); err != nil {
 		t.Fatal(err)
 	}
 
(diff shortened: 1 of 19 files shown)


dungbanguyen:
Rebased onto #769, which is the actual cause of the shard-7 failure this PR was showing — not the split.

The failure was `cache lookup failed for type N (SQLSTATE XX000)` in reembed's baseline seed. It is a soundness bug in #626's shared pool (`cache_describe` caches the parameter OID that `$5::vector` makes the server infer, and a migration recreating the extension changes it). This PR only exposed it, by changing which tests share shard 7.

**Verified on the rebase:** shard 7 green (172 tests, 0 skips), `make check-backend` and `make craft-static` clean.

Base is temporarily #769's branch so the diff here stays just the split — I'll retarget to `main` once #769 lands.

PR #673, opened 2026-08-09
feat(agents): a long wait becomes a Task, and the passport that polls it is the one that acts

tinnhlt-gra:
Implements the MCP Tasks extension (`io.modelcontextprotocol/tasks`) over the one operation on this surface that genuinely waits: a human decision. **Roadmap track F2**, unblocked by C1.

## The dead end this closes

A confirm-first (🟡) call is refused and told to come back with an `approval_id`. The agent is never told when — or whether — the person decided, so it either abandons an effect the human released or re-issues the call on a guess. Now the staged call answers a durable handle and the client polls `tasks/get` until the decision lands.

## A Task is never authority — which is why it executes on the poll

`tasks/get` is an authenticated MCP request like any other, so when a poll finds the approval released **the effect runs inside that poll**, under the polling passport, through the same `Registry.Invoke` a manual retry takes. Admission, object RBAC, row scope, the version pin, the volume counters and the retry claim all apply because it is the same path — not because this track re-checks them.

Nothing runs between requests, which is what makes the plan's two exit criteria structural rather than defended:

- **a cancelled Task never reports cancelled while writes continue** — there is no writer outside a request, and cancel is guarded on `claimed_at IS NULL` so it cannot land while a poll is executing;
- **a revoked passport stops its own running Task** — its next poll is refused with a 401 at the transport, before dispatch is entered.

No narrowed durable capability is minted, and none is needed. The task id carries no permission at all: every method requires the presenting passport to match, and a mismatch answers exactly what an unknown id answers.

## What is executed is the human's payload, not the agent's

A person may edit a staged proposal before releasing it (ADR-0036 §4), and doing so rewrites both `proposed_change` and `diff_hash` — *"the original hash no longer opens anything"*, in the approvals module's own words. A task replaying its own arguments would perform what the agent asked for rather than what the person allowed, and would be refused for the mismatch. The payload is read live through the seam.

## Scope, and what is deliberately excluded

Exactly one refusal becomes a task: a 🟡 call that **staged**. Everything else is unchanged.

| Not a task | Why |
|---|---|
| The C2 quota step-up | Releasing it widens a counter; it does not perform the call. A task completing on release would report an effect that never happened. Excluded **by construction** — `StepUpStagedError` does not unwrap to the confirm-first sentinel. |
| A 🟡 tool that cannot stage | There is no durable decision to hand a handle to. |
| Any client that did not declare the extension on the request | The specification's MUST NOT. Byte-identical refusal to today's. |
| Every handshake-era client | It has no per-request `_meta` to declare in, so it is a non-declaring client by construction. |

**No aicert re-run is owed.** No tool registered or retired, no description, prompt or corpus text edited; the catalog and every schema in it are byte-identical, and the cert band drives `ai.Task` through the AI runtime rather than the protocol. Verified, not assumed — the prompt-version currency gate in `make check` is what would fail otherwise.

## A spec conflict resolved rather than inherited

`ext-tasks` says `-32003` for a missing client capability. The **core 2026-07-28 specification's error table says `-32021`**, and it partitions the range explicitly: `-32000..-32019` is legacy, new implementations must not allocate there and receivers must assume no meaning for it. The same condition already has a specified code, so the shared table wins. This server emits `-32021` with `data.requiredCapabilities`.

That also made `modern.go`'s "-32021 is deliberately never emitted" false, so it is rewritten rather than left teaching the reverse of production.

## What three review passes changed

Craft, adversarial security, and Codex found eight defects in the first cut. Read these before the next track touches this code.

**A completed task was a free, un-rechecked read.** Its stored result is a whole record read-back for most of the 🟡 family, and every later poll handed those bytes straight back — no row-scope re-check, no read charge, for the life of the handle. That is exactly the hole A4b closed on the idempotency door, whose own comment says why: *"a recorded result is a receipt that outlives the authority it was produced under."* A caller whose grant, seat or row scope had since been pulled kept being paid the record, and an Art. 17 erasure that anonymized the row in place was still served from the snapshot. **Both doors share one gate now** — `Registry.ServeRecorded` re-reads every record the envelope names and re-charges what the call cost, and `replay()` is a call to it.

One consequence worth knowing: `disqualify_lead` archives the very lead its answer names, so its receipt cannot be re-proven and the second poll withholds the document while staying `completed`. That is the same trade `archive_record` already makes — the effect happened once, which is the promise; the receipt is what the caller gives up.

- **Cancel raced the executor.** It read only `status`, so a cancel landing mid-execution marked the task cancelled, the executor's settlement was discarded by the terminal guard, and the client was told "nothing was changed" for a mutation that committed.
- **Cancel claimed a withdrawal that did not happen.** `WithdrawInTx` only touches a pending row; that bool was dropped. A failed withdrawal was acked as success.
- **An approval spent outside the task read as unexecuted** — the poll invoked, was refused as already-redeemed, and recorded a *refusal* for an effect that had succeeded. Execution history was read from the task's claim while authority was spent on the approval: two components reading one value by different means.
- **The handle expired before its decision did.** It took the staging deadline, but a yes granted in that window's last second stays redeemable for fifteen minutes after it, so an approved effect was stranded behind "no such task".
- **The claim lease (2m) was shorter than the longest exchange (150s)**, so a poll still inside its handler could be declared dead and its task settled `failed` while the original execution went on to commit. It is derived from `mcpCallDeadline` now.
- **`Dispatcher.clock` was never assigned** while its comment claimed injection. Deleted.

One craft finding did **not** hold on checking: `approvalsAdapter.Stage` leaves `JoinPending` false, so a repeat call inserts its own proposal and collides with no index. `TestRepeatingAStagedCallAnswersItsOwnHandle` records that.

## Shape

`agent_task` (core **0201**) is transport-owned operational state like the idempotency claim — no audit row, no outbox row; the effect it performs carries the full write shape through the tool path. RLS + FORCE, composite `(workspace_id, id)` FKs, and an hourly retention sweep, because a completed task stores a verbatim record read-back and Art. 5(1)(e) does not let it outlive its purpose.

`status` holds **only wire values**; work in flight is `claimed_at`, a timestamp — a second vocabulary is how the two would eventually disagree.

## Testing

Unit (`modules/agents`, `modules/approvals`): the negotiation MUST NOT from both directions, `-32021` with its `requiredCapabilities`, `-32601` in the legacy era, the status machine, the existence-hiding refusal, the withheld-receipt path, an approval spent elsewhere, and the lease-vs-exchange bound.

Integration (real Postgres, real origin) — including the three the reviews asked for:

- a 🟡 call becomes a handle, a person approves, the effect lands **once**, and the second poll runs nothing;
- **two simultaneous polls redeem exactly once** (the claim SQL, not a fake that reimplements it);
- **a claim left unsettled past its lease fails without running**, redeeming nothing;
- a foreign passport gets the byte-identical answer an unknown id gets;
- cancel takes the proposal off the inbox and the effect never lands;
- a non-declaring client still gets the plain refusal, and no task row is created.

---
## Summary by cubic
Implements the `io.modelcontextprotocol/tasks` extension so a confirm-first (🟡) call that staged returns a durable `taskId`, and the approved effect runs on the next `tasks/get` poll under the polling passport. Legacy and non-declaring clients behave exactly as before.

- New Features
  - Added `tasks/get`, `tasks/update`, and `tasks/cancel`; execution happens on poll via `Registry.Invoke` using the human-approved payload.
  - Recorded task results are served through `Registry.ServeRecorded`, which re-reads rows and re-charges volume before returning stored bytes.
  - New `agent_task` table (core `0201`) with hourly retention sweep; HTTP wiring via `WithTaskStore`. The sweep spares a task with an active claim lease.
  - Only staged 🟡 calls become tasks; others are unchanged. Non-declaring and handshake-era clients get the existing refusal. Emits `-32021` for missing client capability; task methods require mirrored `Mcp-Name`.

- Bug Fixes
  - Extension negotiation is exact and case-sensitive; a mis-cased declaration no longer counts.
  - Cancellation no longer races execution, reports failed withdrawals, errors when a cancel cannot be recorded, and uses the same lease predicate as claim/sweep.
  - Completed answers are not free snapshots; replays re-prove row scope, re-charge reads, and withhold documents that can’t be re-proven. Poll replays now honor Calls ceilings via replay admission.
  - A settlement that loses now routes through `Registry.ServeRecorded`, avoiding unproven, uncharged reads.
  - Approval-consumed checks run after the claim to avoid misreporting under concurrent polls; approvals redeemed elsewhere are reflected as completed without re-running.
  - Task expiry matches the decision’s redemption window; claim lease aligned with `mcpCallDeadline` to avoid false failures.
  - Served record count persists across replays; schema/indexing avoids workspace-wide scans on passport deletes.
  - Zero-record successes keep their receipt on later polls; nothing is withheld when no records were served.

<sup>Written for commit ea0e9dc9659f165edebcbede2368ef82c5870977. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/673?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added durable MCP task handling for approval-based tool calls.
  * Modern clients can poll, update, cancel, and retrieve task results.
  * Added capability negotiation, task status reporting, expiration handling, and safe single execution.
  * Added hourly workspace-scoped cleanup of expired tasks.

* **Bug Fixes**
  * Improved approval expiration, ownership checks, replay handling, and error reporting.
  * Preserved completed results and prevented duplicate execution during retries or concurrent polling.


tinnhlt-gra:
## Live UAT — the real api on `:8080`, real database, real human approval

Driven with raw HTTP against `cmd/api` (not the test harness), so the composition root in `mcpedge.go` is the wiring under test. The human half goes through the ordinary REST session; the agent half through `/mcp` with a real passport.

```
=== 1. sign in as the admin human ===
login -> HTTP 200

=== 2. mint an agent passport (read+write) ===
passport minted

=== 3. server/discover advertises the extension ===
{
  "extensions": {
    "io.modelcontextprotocol/tasks": {}
  },
  "resources": {
    "listChanged": false,
    "subscribe": false
  },
  "tools": {
    "listChanged": false
  }
}

=== 4. create a lead for the confirm-first tool to act on ===
lead 019fe530-4e92-7661-9e61-772c6f26de7b

=== 5. a NON-declaring client calls the yellow tool -> the plain refusal, unchanged ===
resultType: complete | isError: True
text: This is a confirm-first (🟡) action: it needs human approval before it runs. Ask the user to perform it in the CRM, or wait for the approval flow. Noth

=== 6. a DECLARING client calls it -> a durable task handle ===
{
  "resultType": "task",
  "taskId": "019fe530-4f02-77b4-84b9-c11ca0c5427b",
  "status": "working",
  "statusMessage": "A person must approve this before it takes effect. Nothing has changed yet. Poll tasks/get with this taskId; it completes when they decide.",
  "ttlMs": 87299992,
  "pollIntervalMs": 5000
}

=== 7. poll before the decision -> working, and the lead is untouched ===
status: working | A person must approve this before it takes effect. Nothing has changed yet. Poll tasks/get
lead status: new

=== 8. tasks/get WITHOUT the capability -> -32021, naming what to declare ===
code: -32021
data: {"requiredCapabilities": {"extensions": {"io.modelcontextprotocol/tasks": {}}}}

=== 9. a human approves it in Margince ===
approval 019fe530-4efd-74c4-acc4-7e761c5e22c5
approve -> HTTP 200

=== 10. the next poll performs the effect under its own passport and completes ===
status: completed | A person approved this and it has now been carried out.
result isError: None
lead status: disqualified

=== 11. poll again -> nothing re-runs; the receipt is re-proven and withheld (the lead is archived) ===
status: completed
result: {"content": [{"text": "This was approved and carried out, but its result can no longer be shown to this agent \u2014 the records it named are no longer readable under the current access. Do not repeat the call; ask the u

=== 12. Mcp-Name that disagrees with the body -> -32020 ===
code: -32020 | header mismatch: Mcp-Name is missing or disagrees with the name its own body invokes, and 

=== 13. cancel: a fresh staged call, withdrawn before anyone decides ===
cancel resultType: complete
status: cancelled | This task was cancelled and its pending approval withdrawn. Nothing was changed.
live pending proposals: 1
lead status: new

=== 14. a task id from another passport reads exactly like an unknown one ===
another passport's task      -> -32602 | no such task: it was never created by this passport, or it has expired and been discarded. Do not poll it again.
an id nobody minted          -> -32602 | no such task: it was never created by this passport, or it has expired and been discarded. Do not poll it again.

=== 15. tasks/* in the handshake era do not exist ===
code: -32601 | method not found: tasks/get

=== UAT complete ===
```

### Reading two lines that deserve it

**Step 11 — `completed`, but the result is withheld.** Nothing re-ran. The recorded answer was re-proven against a live read and refused, because `disqualify_lead` archives the very lead its answer names — the same trade `archive_record` already makes: the effect happened once, which is the promise; the receipt is what the caller gives up. Before this PR's review round those bytes came back uncharged and un-rechecked for the life of the handle.

**Step 13 — "live pending proposals: 1".** That one is step 5's proposal, staged by the non-declaring client and never cancelled. The cancelled proposal is gone — confirmed in the table:

```
      kind       |  status  | lapsed
 disqualify_lead | pending  | f      <- step 5, still open
 disqualify_lead | approved | f      <- the task's, redeemed on the poll
 disqualify_lead | pending  | t      <- cancelled: withdrawal is forced expiry
```

Withdrawal deliberately mints no new status — the CHECK and the public enum stay closed — so a withdrawn row reads expired everywhere instead.

### What the transcript establishes

| | |
|---|---|
| The extension is advertised, and only in the era that can declare it | step 3 |
| A non-declaring client's 🟡 refusal is byte-identical to today's | step 5 |
| A declaring client gets `resultType: "task"` with a poll interval and a real TTL | step 6 |
| Polling before the decision changes nothing | step 7 |
| `-32021` names the capability to declare — not the extension's stale `-32003` | step 8 |
| The effect runs on the poll, under the polling passport, once | steps 9–10 |
| A second poll re-proves rather than re-serving | step 11 |
| `Mcp-Name` is mirrored on the task methods | step 12 |
| Cancel withdraws the proposal and the effect never lands | step 13 |
| A foreign task id and an unknown one are one answer | step 14 |
| `tasks/*` does not exist in the handshake era | step 15 |

tinnhlt-gra:
## Bot round — eight of nine taken

**`//nolint:dupl` → `//craft:ignore dupl` was not taken.** `craft` implements no `dupl` check (it is a golangci linter, and craft's checks are its own: `naked-any`, `swallowed-errors`, the size ceilings). A `//craft:ignore dupl` would suppress nothing and leave `lint` red. Verified against `cli/craft` before declining.

The other eight are fixed in `7cbc1213`. Two are worth reading:

**A mis-cased `Extensions` read as a declaration.** `encoding/json` matches struct members case-insensitively, so a client sending `"Extensions"` was handed a task handle it never claimed to understand — and the approved effect behind it would strand with nobody polling. It decodes through a map and matches exactly now, which is the rule `metaOf` already keeps for the reserved `_meta` keys and for the same reason. It has a table test of its own.

**The consumed-approval fix regressed concurrency, and the integration lane caught it.** Checking `consumed` *before* the claim cannot tell an approval spent elsewhere from one being spent by the poll executing this very task — so a losing concurrent poll declared the task interrupted while the winner was still committing. `TestTwoSimultaneousPollsRunTheReleasedCallOnce` went red on both polls reporting `failed`. The check belongs after the claim, where "consumed" means what it says.

Also this round: the retention sweep now spares a row an executor may still be inside (reading the executor's own lease, not a second copy of the number); a token refusal is re-asked before it is believed; a cancellation that cannot be recorded is an error rather than an ack; a replayed idempotency key's record count reaches the task, so its later polls charge what the answer cost; and the passport FK gets its own index.

New-code coverage was 78.5% against the ≥80% gate. The gap was the dependency-failure branches, which now have a suite on their own merit: an unreadable decision leaves the task `working`, an unwritable settlement still tells its caller the truth, and a handle that cannot be created falls back to the plain refusal.

tinnhlt-gra:
## The zero-record case was real, and it was on the flagship path

CodeRabbit's `idempotency.go:336` finding is fixed in `ea0e9dc9`. It deserves its own note because the reachability argument is what makes it a defect rather than a hypothetical.

`ensureReplayVisible` refuses an evidence-free envelope, and the task path reaches it for **any** success — including successes that legitimately name nothing. `tools_comms.go` and `tools_lifecycle.go` call `newWireRecord` nowhere: `send_email` answers an activity id and a status, reading no record through the seam. So a task-driven `send_email` completed, and its **second poll withheld the receipt** with *"the records it named are no longer readable"* — a claim about lost access that never existed, on a task whose effect had committed.

Zero is served as it stands now. The divergence from the idempotency door is deliberate: that door cannot tell *"this answer carried no records"* from *"this answer's evidence went missing"*, and refusing on that doubt is exactly what its gate is for. Here the count is a **positive fact** recorded at execution time by the same pass that collects the evidence — nothing left through the seam, so there is nothing to re-prove and nothing to re-charge. Above zero the gate is unchanged.

`TestAnAnswerThatNamedNoRecordsIsStillServedOnASecondPoll` pins it.

---

**Still not taking the `//craft:ignore dupl` waiver format.** `craft` implements no `dupl` check — it is a golangci linter, and craft's own checks are `naked-any`, `swallowed-errors` and the size ceilings. `//craft:ignore dupl` would suppress nothing and leave `make lint` red. Verified against `cli/craft` rather than assumed.

Issue #743, opened 2026-08-09
LinkedIn: a ghost row has no terminal state, so a refused or orphaned suggestion is swept forever


tinnhlt-gra:
The `linkedin_connection.match_status` column records the MATCHERs outcome and one
human outcome — `confirmed`, written by the accept effect. Two states that end a
suggestions life are never written back to the row at all, so nothing that reads the
row can tell a finished suggestion from a live one.

**1. A refusal is not written to the ghost row.** Rejecting a `linkedin_match` proposal
runs no effect (`approvals/decide.go` dispatches the effect table on `approve` only), so
the connection stays `suggested`, non-tombstoned, forever. The `rejected` value the
migration defines (`migrations/core/0159_linkedin_connection.up.sql`) has no writer
anywhere in the tree. This is the row half of #678.

**2. A suggestion whose contact leaves the live set is stranded.** The pending read joins
`p.archived_at IS NULL` (`people/linkedinmatchapply.go`), so an archived contact — which
includes the SOURCE of a person merge, since nothing re-points `matched_person_id` — can
never be proposed. The matcher cannot revisit it either, because both arms filter
`match_status = unmatched`. And no approval about it could be decided anyway:
`auth.EnsureVisibleLive` refuses the archived person. The row is permanently inert with
no state saying so.

Correctness is fine in both cases — a refusal survives (`StageUnlessDeclined` reads the
approval), and nothing wrong is proposed. **The cost is what changed.** Since the sweeps
enumeration widened to `match_status IN (unmatched,suggested)` (PR for #677), an
owner in either state above is enumerated on every hourly sweep and every organization
event, with a guaranteed-empty result: an RBAC resolve, a whole-network match, a pending
read, and one idempotent staging transaction per outstanding row. Refusals are supposed
to be the cheap terminal state; today they are a permanent per-pass cost that only grows
with the size of the imported network.

Related: the 24h staging TTL re-mints an approval (plus its audit row and
`approval.requested` event) once a day for every suggestion nobody decides, indefinitely
— defensible as re-proposing against fresh state, but it has no terminal state either.

**What would fix it, as one change rather than two:** give the ghost row the transitions
its lifecycle already has.

- a refused proposal writes `match_status = rejected` (needs the rejected-effect seam
  #678 asks for, and the audit + outbox tail #676 extracts)
- a suggestion whose matched person is archived or merged away is released back to
  `unmatched` with `matched_person_id` cleared, so the matcher can re-evaluate it against
  the survivor

[...]

Issue #678, opened 2026-08-09
Rejecting a linkedin_match approval has no effect on the ghost row


LarsGradion:
Rejecting a `linkedin_match` approval changes nothing on the ghost row. The approvals service dispatches effects only on approve: `approvals/decide.go:111` (`effect, ok := s.effects[a.Kind]; ok && approve && serverProposed(a)`), and the bundle path is explicitly inert on reject (`approvals/bundle.go:216-217`, "a rejection releases nothing"). There is no reject-side effect mechanism at all, and no non-test code ever writes `match_status = 'rejected'`.

Consequences:

- `matchRankOrder`'s `"rejected"` slot (`people/linkedinrenormalize.go:60`) and the `AND g.match_status <> 'rejected'` predicate (`people/linkedinmatch.go:354`) are dead branches — no row can ever be `rejected`.
- The rejected pair stays `suggested` in the people module forever; `PendingLinkedInMatches` (`people/linkedinmatchapply.go:84`) keeps re-reading it, and only the approvals-side `StageUnlessDeclined` guard (`compose/linkedinmatchproposal.go:139`) suppresses re-staging. The domain record and the approvals record disagree about what happened.

Fix:

1. Add `WithRejectEffect(kind, fn)` beside `WithEffect`, invoked where reject decisions land in `runDecisionEffect` (and the bundle path, which PR #660 added — a rejected bundle member needs the same dispatch).
2. `people.RejectLinkedInMatch`, single transaction: set `match_status='rejected'`, `matched_person_id=NULL`, guarded by `owner_user_id` and current state (`match_status='suggested' AND matched_person_id = <proposed person>`), plus audit row and outbox event. Requires `person:update` and target visibility, same as apply.

Tests owed: reject over HTTP leaves the ghost `rejected` with `matched_person_id` NULL; re-import plus sweep does not re-propose it; the previously dead `<> 'rejected'` predicate now has a row it excludes.

dungbanguyen:
picking this up

PR #744, opened 2026-08-09
fix(people): a LinkedIn match suggestion reaches the inbox, and an unrelated edit no longer cancels it (#677, #679)

### LINKED ISSUE #677: The event-driven LinkedIn matcher's suggestions never reach the approval inbox
Suggestions produced by the event-driven matcher never reach the approval inbox. `compose/linkedinmatchgen.go` `matchPerson` (lines 113-127) and `matchWorkspace` (lines 133-147) run the matcher but never call `StageLinkedInMatches`. A suggestion created there only becomes an inbox proposal when the next hourly sweep or the next CSV import happens to stage it.

And the hourly sweep cannot always repair that, because of a second gap: `compose/linkedinowner.go:63-65` enumerates owners with

```sql
SELECT DISTINCT owner_user_id FROM linkedin_connection
 WHERE match_status = 'unmatched' AND tombstoned_at IS NULL
```

Only `'unmatched'`. An owner whose ghosts are all already `'suggested'` but unstaged is never enumerated, so `StageLinkedInMatches` in `compose/linkedinrematch.go:148` never runs for them. Their suggestions sit invisible forever.

Fix:

1. One helper that runs the matcher then `StageLinkedInMatches`, used at all three call sites: the import handler seam (`people/handlers_linkedin.go:94-99` via the stager), the hourly sweep (`compose/linkedinrematch.go`), and `linkedinmatchgen.go` `matchPerson`/`matchWorkspace`.
2. `ghostOwners` enumerates `match_status IN ('unmatched','suggested')`.
3. Build the approvals service once per consumer/sweep pass, not per owner per event — `linkedinrematch.go:148` currently constructs `approvalsServiceWithEffects(w.pool)` inside the per-owner loop.

Test owed: a person event produces a suggestion AND a pending approval for the ghost's owner, without waiting for the sweep; an owner whose only ghosts are `suggested` still gets staged by the sweep.

### LINKED ISSUE #679: A linkedin_match approval pins the person version, so any contact edit cancels the waiting approval
A `linkedin_match` proposal pins the target person's version at staging, so ANY edit to the contact between staging and decision makes the redemption fail its version re-check. This contradicts the founder decision that a contact edit must not cancel a waiting LinkedIn-match approval.

Where: staging uses `TargetType: person` (`compose/linkedinmatchproposal.go:143-148`), `person` is in `versionTables` (`approvals/redeem.go:149-153`), and `resolveTargetVersion` (`approvals/staging.go`) pins every checkable, non-context-only kind. `linkedin_match` is not in the `contextTargetKinds` waiver (`approvals/redeem.go:170-177`, which today holds only `site_lead`).

The fix is NOT to add it to `contextTargetKinds` — that map means "the target is context the proposal is merely about", and this effect genuinely writes the target's `person_social` and bumps the person row. Misfiling it there would trade one lying label for another.

Instead: add a second, honestly-named waiver beside it — `unpinnedKinds` (kind → rationale, same fitness-test constraints as `contextTargetKinds`: rationale required, must be a registered effect kind). Rationale for `linkedin_match`: the proposal's claim is "this ghost is this contact", which no field edit on the contact can invalidate, and the write it authorizes is an additive idempotent `person_social` insert rather than a field patch. There is nothing for a version pin to protect — and pinning also makes the second of two matches onto one contact fail after the first bumps the version.

Test owed: stage a linkedin_match, edit the person, approve — the approval still applies.

tinnhlt-gra:
Closes #677, closes #679.

A member uploads their LinkedIn export during onboarding, and the contacts it could
match arrive over the following weeks. The event consumer that catches those arrivals
matched them, wrote `suggested` to the ghost row, logged a line, and stopped. The
pending question lives in the **approval**, not in `match_status` — so nobody was ever
asked, and nothing in the product could tell them a question existed.

The hourly sweep could not repair it either: it enumerated owners with `unmatched`
ghosts only, so a member whose ghosts had all been matched to `suggested` dropped out of
the enumeration entirely and their suggestions sat invisible for the row's lifetime.

And the ones that did reach the inbox were cancelled by an unrelated contact edit: the
staging pinned the person's version, so a title correction between staging and decision
made the redemption fail — against the founder decision that a contact edit must not
cancel a waiting match.

## What changed

**A pass proposes over the same scope it matched.** The per-person arm stages that
contact's pending matches; the workspace arm, the sweep and the import handler stage the
whole outstanding set. The narrow arm is what keeps the event path bounded — matching one
arrival can only raise questions about that arrival, so it is the complete answer as well
as the cheap one, and the whole-set alternative would cost one transaction per open
question on every contact write in the workspace.

**`ghostOwners` counts both undecided statuses**, so the sweep can reach residue at all.

**The approvals service is built once** per consumer, worker and stager rather than once
per member per pass — a dozen effect registrations that answer the same service every time.

**Writing the test found a second defect the issue does not name.** The sweep bound no
correlation id, and the write shape refuses one without it: its staging failed on the
first proposal it tried and failed the whole workspace's re-match with it — permanently,
since the retry rebuilds the same context. **The sweep's staging has never run in
production.** `systemContext` now binds a trace the way every other workspace job does.

**#679 gets a second, honestly-named pin waiver.** `linkedin_match` targets the person its
effect writes, so filing it under `contextTargetKinds` — "the target is context the
proposal is merely about" — would trade one lying label for another. `unpinnedKinds` says
the other thing: the target is the operand and the pin binds nothing a human judged. Both
maps are held to the same discipline, and a new gate keeps a kind from claiming both.

**Then both reviews found the same two sibling copies**, and they were right. `deal_follow_up`
and `capture_counterparty` were pinned to rows their effects never write. The follow-up
one is the sharper case: its stager has said "the proposal carries no target_version" in
prose the whole time, and that stopped being true when the pin moved server-side — an
overnight proposal waits for somebody's morning inbox, which is exactly the window a rep
edits the deal in. Neither failure is visible where it happens, because the decision
commits before the effect runs: the human sees an approved proposal whose work was never
done.

## Proof

**Every new test fails against the defect it describes** — verified by reverting each fix
and re-running:

| test | failure against the unfixed code |
|---|---|
| `TestAPersonEventStagesTheLinkedInSuggestionItProduced` | `0 pending linkedin_match proposals, want exactly 1` |
| `TestTheSweepStagesALinkedInSuggestionNobodyWasEverAskedAbout` | `0 pending linkedin_match proposals, want exactly 1` |
| `TestAContactEditDoesNotCancelAWaitingLinkedInMatch` | `target changed since approval (v1 → v2): version skew` |
| `TestADealEditDoesNotCancelAWaitingFollowUp` | same, on the deal |
| `TestEditingTheCapturedMessageDoesNotCancelItsWaitingReview` | same, on the activity |

Plus `TestTheSweepNeverReasksALinkedInMatchThatWasRefused` (the widened enumeration makes
durable refusal load-bearing in a new way), the `unpinnedKinds` discipline gates, and a
guard test for the narrow read's zero-contact refusal.

### UAT — a live stack, end to end through the real bus

`DEV_SLUG=lim`, worker on, events flowing through the outbox relay.

```
POST /v1/organizations                          201   Simio GmbH
POST /v1/me/linkedin-connections                200   {"imported":1,"suggested":0}   ← ghost waits, unmatched
GET  /v1/approvals?status=pending                     pending approvals: 0            ← nothing to ask yet
POST /v1/people            "Andreas Muller"     201   version 1
POST /v1/relationships     employment            201
                                                      t+3s  pending=1                 ← #677: the bus staged it
```

```json
{ "kind": "linkedin_match",
  "summary": "Andreas Müller at Simio GmbH looks like Andreas Muller",
  "target_entity_type": "person",
  "target_version": null }          ← #679: no pin on the wire
```

```
PATCH /v1/people/{id}  {"title":"Head of Procurement"}   200   version 1 → 2
POST  /v1/approvals/{id}/approve                         200   status approved, effect_error null
GET   /v1/people/{id}   → social {"linkedin": "https://www.linkedin.com/in/amueller-uat"}
psql  → linkedin_connection: Andreas Müller | confirmed | linked t
```

The edit bumped the contact to v2 while the question waited, and the approval still
applied.

**Negative case** — a colleague may not see or decide someone else's match question:

```
GET  /v1/approvals?status=pending   (rep@demo.test)   → 0 pending, kinds: []
GET  /v1/approvals/{admins-proposal}                  → 404 not_found      ← existence-hiding
POST /v1/approvals/{admins-proposal}/approve          → 404 not_found
```

### Gates, locally, before pushing

- `craft static --strict` — PASS, 0 blocker / 0 major / 0 minor
- `make test-integration` — `OK: integration passed with 0 skips (29 packages)`
- new-code coverage — every new function 75–100% (`StageLinkedInMatchesForPerson` 75%,
  `stagePendingLinkedInMatches` 82%, `PendingLinkedInMatchesForPerson` 100%,
  `suggestedMatches` 85%, `TargetVersionUnpinned` / `UnpinnedKinds` 100%,
  `systemContext` 100%)

No prompt, tool description, contract or aicert corpus file is touched, so no paid lane is owed.

## Filed, not fixed

**#743** — a ghost row has no terminal state. A refusal is recorded on the approval and
never on the row, and a suggestion whose contact is archived or merged away can never be
proposed, matched again, or decided. Correctness holds in both cases; the cost does not —
the widened enumeration now sweeps those owners forever for a guaranteed-empty result.
Both reviews reached it independently, from opposite ends, and the honest fix is the
rejected-effect seam #678 asks for plus a release transition, which is a change of its own.

---
## Summary by cubic
Fixes two onboarding issues: LinkedIn match suggestions now reach the inbox when they’re created, and unrelated edits no longer cancel pending approvals for `linkedin_match`, `deal_follow_up`, or `capture_counterparty` (addresses #677 and #679).

- **Bug Fixes**
  - Stage `linkedin_match` proposals in the same pass that matched them: per-person events stage that contact only; workspace and sweep paths stage all outstanding. Idempotent staging prevents duplicates.
  - Sweep now enumerates both `unmatched` and `suggested` owners and binds a correlation id, so staging actually runs and can rescue missed suggestions.
  - Approvals decline the version pin where it never protected content:
    - `linkedin_match` is declared in `UnpinnedKinds` (claim is “this connection is this contact”; no contact edit can falsify it).
    - `deal_follow_up` and `capture_counterparty` are declared as context-only targets, so edits to their filing rows don’t cancel approvals.
  - Contact/message/deal edits between staging and decision no longer break redemption; regression tests cover these paths and durable refusals are not re-asked.

- **Refactors**
  - Added person-scoped helpers: `PendingLinkedInMatchesForPerson` and `StageLinkedInMatchesForPerson`; zero-contact reads are rejected; the pending-match query uses `optionalPerson()` to express “all contacts” and keep the row-scope join single-sourced.
  - Built the approvals service once per consumer/worker/stager; event and sweep logs include `staged` counts.
  - Added `UnpinnedKinds` alongside `contextTargetKinds`, plus guard tests to ensure every waiver is explained, references a real kind, and no kind claims both.

<sup>Written for commit 7e6f837ada868c9bb7df9969ee028e9d022eae1f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/744?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/internal/compose/approval_contexttarget_test.go
+++ b/backend/internal/compose/approval_contexttarget_test.go
@@ -11,9 +11,10 @@ import (
 )
 
 // A staging's version pin is what stops a confirmed action executing against a
-// row that changed underneath it. Declining the pin is a waiver, so it is held
-// to the same discipline as the confirm-first one: every entry says why, and
-// an entry naming a kind nothing stages is removed rather than left to rot.
+// row that changed underneath it. Declining the pin is a waiver, so both maps
+// that decline it are held to the same discipline as the confirm-first one:
+// every entry says why, an entry naming a kind nothing stages is removed rather
+// than left to rot, and no kind may claim both reasons at once.
 
 func TestEveryContextTargetKindIsExplained(t *testing.T) {
 	declared := approvals.ContextTargetKinds()
@@ -29,13 +30,54 @@ func TestEveryContextTargetKindIsExplained(t *testing.T) {
 }
 
 func TestEveryContextTargetKindIsAKindWeStage(t *testing.T) {
-	registered := map[string]bool{}
-	for _, kind := range approvalsServiceWithEffects(nil).EffectKinds() {
-		registered[kind] = true
-	}
+	registered := registeredEffectKinds()
 	for kind := range approvals.ContextTargetKinds() {
 		if !registered[kind] {
 			t.Errorf("contextTargetKinds[%s] names no registered effect kind — stale waiver, remove it", kind)
 		}
 	}
+}
+
+func TestEveryUnpinnedKindIsExplained(t *testing.T) {
+	declared := approvals.UnpinnedKinds()
+	if len(declared) == 0 {
+		t.Fatal("no unpinned kinds declared — if the waiver is unused, delete the mechanism")
+	}
+	for kind, why := range declared {
+		if len(strings.TrimSpace(why)) < 40 {
+			t.Errorf("unpinnedKinds[%s] has no real rationale — a staging that declines its version "+
+				"pin over the very row its effect writes must say what the pin would have protected", kind)
+		}
+	}
+}
+
+func TestEveryUnpinnedKindIsAKindWeStage(t *testing.T) {
+	registered := registeredEffectKinds()
+	for kind := range approvals.UnpinnedKinds() {
+		if !registered[kind] {
+			t.Errorf("unpinnedKinds[%s] names no registered effect kind — stale waiver, remove it", kind)
+		}
+	}
+}
+
+// The two waivers make opposite claims about the same target — one says it is
+// context the proposal is merely ABOUT, the other that it is the operand the
+// effect writes. A kind declared in both asserts both, and whichever a reader
+// believes, the other is a rationale that reads true and is not.
+func TestNoKindIsBothContextOnlyAndUnpinned(t *testing.T) {
+	for kind := range approvals.UnpinnedKinds() {
+		if approvals.TargetIsContextOnly(kind) {
+			t.Errorf("%s is declared in BOTH contextTargetKinds and unpinnedKinds — its target is either "+
+				"context the proposal is about or the row its effect writes, and the two rationales say "+
+				"opposite things about what a pin would have bound", kind)
+		}
+	}
+}
+
+func registeredEffectKinds() map[string]bool {
+	registered := map[string]bool{}
+	for _, kind := range approvalsServiceWithEffects(nil).EffectKinds() {
+		registered[kind] = true
+	}
+	return registered
 }
--- a/backend/internal/compose/captureverdictreview_integration_test.go
+++ b/backend/internal/compose/captureverdictreview_integration_test.go
@@ -66,6 +66,76 @@ func TestCounterpartyAcceptCreatesTheRecordsAndClosesTheDisposition(t *testing.T
 	}
 }
 
+// The proposal is FILED under the message that carried the unrecognized sender,
+// because that message is the evidence a human judges it on — but the effect
+// creates a person and an organization and closes the disposition, and never
+// writes the activity. So the ordinary inbox work done to the message while the
+// question waits (a relink, a participant correction, a subject fix) must not be
+// able to cancel the answer.
+//
+// It could, and silently: the decision commits before the effect runs, so the
+// human saw an approved proposal whose records were never created.
+func TestEditingTheCapturedMessageDoesNotCancelItsWaitingReview(t *testing.T) {
+	e := integration.Setup(t)
+	activityID := seedCapturedMail(t, e, "dana@editco.example", "about your services")
+	dispositionID := seedPendingDisposition(t, e, "dana@editco.example", "editco.example", activityID)
+	retireToUnsure(t, e, dispositionID)
+
+	svc := approvalsServiceWithEffects(e.Pool)
+	engine := NewCounterpartyVerdictEngine(e.Pool, &scriptedVerdictBrain{}, slog.Default())
+	if err := engine.StageReviewsWorkspace(principal.WithWorkspaceID(context.Background(), e.WS), 0); err != nil {
+		t.Fatalf("staging reviews: %v", err)
+	}
+	approvalID := stagedProposalID(t, e, dispositionID)
+
+	before := activityVersion(t, e, activityID)
+	editCapturedSubject(t, e, activityID, "about your services (corrected)")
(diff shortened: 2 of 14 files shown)


tinnhlt-gra:
@coderabbitai review

PR #775, opened 2026-08-10
refactor(test): the job fan-out suites become their own package

dungbanguyen:
Replaces #672, which I broke: I left auto-merge armed on it and then retargeted its base onto a feature branch, which is unprotected and therefore had no required checks to hold it. It merged into my own branch with a red shard. Nothing unreviewed reached `main`, but the PR it landed in then described one file while carrying nineteen, so I reset that branch and am reopening the split here on its own.

Same content, reviewed and re-verified against current `main` (which now has #769).

The lane's wall clock cannot fall below its longest package, and `internal/compose/integration` is that package by a wide margin. This moves the periodic job-dispatch suites out into `integration/jobfanout`.

## Measured, same sitting, Postgres restarted before each run

| | control (`main`) | this branch |
|---|---|---|
| `compose/integration` | 182.5s | **157.5s** (−13.7%) |
| lane wall clock | 212.1s | **192.2s** (−9.4%) |

Both green, 0 skips. (These are larger than the numbers on the old #672 because `main` now runs `describe_exec` per #769, which costs a round trip per query. The comparison above is the only valid one — I measured both sides today.)

**Worth saying plainly:** the package sheds 25s and the lane sheds 20. Wall clock is no longer purely bound by the longest package, so further splitting has less left in it than this ratio suggests. I would stop after this one and look at the ~35s that sits between the longest package and the lane's wall time instead.

## Three helpers had to be promoted, and the import rule decided where each could go

- `SeedExtraWorkspace` (`seed.go`) and `RetentionPassCtx` (new `scope.go`) are exported from this package: both sides need them, neither needs `compose`.
- The River job-runner ceremony **could not follow them**. It builds a runner, so it needs `compose` — and package `integration` may never import `compose`, because `compose`'s own white-box tests import `integration`. The compiler reports this as `import cycle not allowed in test`, naming neither the helper nor the file that moved. It becomes the `jobtest` package, for the same reason `apptest` exists.

That rule is now in the package doc, along with the trap the review round found: **one package is one binary**, so process-global state does not follow a split. `retention_jurisdiction_integration_test.go` arms a statutory retention floor in `init()`; a retention suite moved to a sibling package would run with no floor armed and go green while asserting the opposite of its sibling. Harmless today (the floor covers `activity` retention, the moved suite seeds `lead/unconverted`) but silent, so both the `init()` and the split rule now say so.

`harness.go` was one line under its 500-line ceiling, so recording that rule pushed it over; the package doc moves to `doc.go`, which is where it belongs.

## Also from the review round

`RetentionPassCtx` binds a SYSTEM principal, which no row-scope clause narrows. No current gate asserts visibility through it — all 11 call sites pass it only to `EvaluateWorkspace`, and isolation there comes from the GUC — but exporting it puts it in reach of every suite in the tree, so `TestRetentionPassCtxOnlyDrivesTheRetentionPass` now bounds it to that one sink. Mutation-tested: a bare use fails it, and it refuses to pass if it ever matches nothing.

---
## Summary by cubic
Split the periodic job fan‑out integration tests into their own package to reduce the lane bottleneck. `compose/integration` runtime drops ~13.7% (182.5s → 157.5s) and lane wall clock drops ~9.4% (212.1s → 192.2s).

- **Refactors**
  - Moved fan‑out suites to `backend/internal/compose/integration/jobfanout` (agentscheduler, privacyretention, runner, runnerreap) to gain an extra parallel slot.
  - Promoted shared helpers to `backend/internal/compose/integration`: `SeedExtraWorkspace` (in `seed.go`) and `RetentionPassCtx` (new `scope.go`).
  - Introduced `backend/internal/compose/integration/jobtest` for the River runner setup and event awaits; exports `StartTestJobRunner`, await helpers, and `DispatchInterval`/`DispatchGapBound`.
  - Added package docs: `backend/internal/compose/integration/doc.go` and `backend/internal/compose/integration/jobfanout/doc.go`; moved long comments out of `harness.go`.
  - Added `backend/retentionscope_test.go` to keep `RetentionPassCtx` bound to `EvaluateWorkspace` and verify the sink is unique.
  - Updated existing suites to use `RetentionPassCtx`, `SeedExtraWorkspace`, and `jobtest` helpers; no behavior changes.

<sup>Written for commit 13231395d4bf435b2823c40cadd089ddbcdace51. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/775?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for integration test organization, scheduling, fixtures, and package boundaries.
* **Tests**
  * Improved integration test coverage and architectural safeguards for retention processing.
  * Preserved existing assertions for job fan-out, retention, retries, reaping, and scheduling behavior.
* **Refactor**
  * Consolidated shared test setup, workspace seeding, job runners, and outcome-waiting utilities for more consistent integration testing.

### DIFF
--- a/backend/internal/compose/integration/ai_call_integration_test.go
+++ b/backend/internal/compose/integration/ai_call_integration_test.go
@@ -255,7 +255,7 @@ func TestEmbedCallRetentionAgesOutOverAgeEmbeddingRows(t *testing.T) {
 	underAge := seedEmbedCall(t, e, 10)
 
 	svc := privacy.NewRetentionService(e.Pool, nil, slog.New(slog.NewTextHandler(os.Stderr, nil)))
-	if err := svc.EvaluateWorkspace(retentionPassCtx(e.WS)); err != nil {
+	if err := svc.EvaluateWorkspace(RetentionPassCtx(e.WS)); err != nil {
 		t.Fatal(err)
 	}
 
@@ -295,7 +295,7 @@ func TestAICallPayloadRetentionAgesOutContentKeepingMetadata(t *testing.T) {
 		VALUES (NULLIF(current_setting('app.workspace_id', true), '')::uuid, 'ai_call_payload', 'content', 365, 'erase')`)
 
 	svc := privacy.NewRetentionService(e.Pool, nil, slog.New(slog.NewTextHandler(os.Stderr, nil)))
-	if err := svc.EvaluateWorkspace(retentionPassCtx(e.WS)); err != nil {
+	if err := svc.EvaluateWorkspace(RetentionPassCtx(e.WS)); err != nil {
 		t.Fatal(err)
 	}
 
--- a/backend/internal/compose/integration/channelidentity_lifecycle_integration_test.go
+++ b/backend/internal/compose/integration/channelidentity_lifecycle_integration_test.go
@@ -151,7 +151,7 @@ func TestRetentionAnonymizeDropsTheChannelIdentityWithoutSuppressingIt(t *testin
 	e.WsExec(t, `UPDATE person SET created_at = now() - interval '800 days' WHERE id = $1`, person)
 
 	svc := privacy.NewRetentionService(e.Pool, nil, slog.New(slog.NewTextHandler(io.Discard, nil)))
-	if err := svc.EvaluateWorkspace(retentionPassCtx(e.WS)); err != nil {
+	if err := svc.EvaluateWorkspace(RetentionPassCtx(e.WS)); err != nil {
 		t.Fatalf("retention pass: %v", err)
 	}
 
--- a/backend/internal/compose/integration/doc.go
+++ b/backend/internal/compose/integration/doc.go
@@ -1,0 +1,61 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+//go:build integration
+
+// Package integration holds the cross-module integration suites — the
+// compose charter exercised end to end over a real migrated Postgres —
+// and the shared fixtures they ride. Three of those carry a database or a booted
+// app, and every suite starts from one of them — all exported, all in non-test
+// files:
+//
+//   - Env, here — a migrated database plus the core stores.
+//   - SearchEnv, in searchenv.go — lighter, and despite the name mostly taken
+//     for the database rather than the search store.
+//   - apptest.AppEnv, in the apptest subpackage — the booted application behind
+//     a TLS test server.
+//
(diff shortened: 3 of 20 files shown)


dungbanguyen:
@coderabbitai three findings, three different dispositions — one fixed, one declined with reasoning, one filed.

## 1. CI runner mechanics in `doc.go` — fixed

Fair, and worth saying why the paragraph got that way: a reviewer in the round before yours told me the opposite half of the same problem, that the doc *overstated* the CI case by implying package splitting helps there when the matrix slices by test. I corrected it by adding the mechanics, which traded one defect for yours.

The decision-relevant fact — a package is the unit the lane schedules, and this one is its long pole — stays. Everything about `-p 1`, sharding and worker pools is gone, replaced by a pointer to the runner script, which is where that lives and which already documents it.

## 2. "less than" → "more than" — declined, the comment is correct

The two clauses describe different things and the finding conflates them. The assertion rejects a gap *greater* than the bound:

```go
if gap := second.Sub(first); gap > jobtest.DispatchGapBound {
    t.Fatalf("the two %s dispatches were %s apart, over the %s bound …")
}
```

so a **pass** means the observed gap was *less than* the bound. The comment describes what that passing observation implies — "two dispatches less than `DispatchGapBound` apart can only happen if a cadence far shorter than that flag's own default is what River is scheduling on" — which is the inference the test rests on, not the predicate it rejects. Changing it to "more than" would make it describe the failure while claiming it proves the invariant.

Unchanged in both files.

## 3. River's `Test.Time` for the periodic suites — filed as #779

Good catch and correct about the API: v0.42.0 does expose a time generator. Not taken here because this PR is a pure move, and because it is a bigger change than it looks — three reasons in the issue, the load-bearing one being that these suites deliberately assert what *arrived* on the observer's own clock rather than what River *scheduled*, so a stub weakens the claim unless it drives the same path the operator's flag feeds.

Worth doing though: I measured the family at **27.8s across 12 tests**, the largest concentration of real waiting in the lane. #779 has the per-test numbers and a suggested shape (put the seam in `jobtest`, keep one suite on the real clock as a control — if all of them move to a stub, nothing proves River honours the interval at all).

Issue #776, opened 2026-08-10
agent_loop's three certification records went stale when the App tool surface merged


tinnhlt-gra:
All three `agent_loop` records now name a prompt version the build no longer produces:

```
agent_loop/gemini_gemini-3.1-flash-lite                  certified p8fcc673…, build is p6fd299e…
agent_loop/openai_compatible_deepseek/deepseek-v4-flash  certified p8fcc673…, build is p6fd299e…
agent_loop/openai_compatible_mistralai/ministral-14b…    certified p8fcc673…, build is p6fd299e…
```

`prompt_version` digests the REQUESTS this build constructs, and the tool listing is rendered
into the runner frame — so #739 (a tool may name a view) moved it the moment it merged. This is
exactly the consequence `.tmp/mcp-roadmap/PROGRESS-ISSUES.md` predicted for the first change to
the catalog, written down before it happened:

> the first merged change to the catalog — #688, D3a, or wave 3s implementation lever — moves
> `prompt_version` and makes those records stale

It is worth filing rather than absorbing quietly for three reasons:

1. **`TestEveryCommittedRecordNamesTheCurrentPromptVersion` WARNS, it does not fail.** A stale
   record is reported inside a passing test, so nothing goes red and nothing blocks. The records
   simply stop describing what ships, silently, until somebody reads the log line.
2. **The four paid passes behind #738 are the ones spent.** That branch shipped as a hardening
   whose records were current *at the time*, on the explicit argument that parking would forfeit
   them on the next catalog change. The catalog changed.
3. **Re-running is a decision, not a formality.** Per #740, one `RUNS=3` pass on the noisy
   bindings cannot support or refute a claim — the same prompt twice on deepseek gave 0.492 and
   0.619 with eleven of twenty-one scenarios moving. Whoever re-runs should decide the run count
   and which bindings are worth paying for first; gemini is the readable one.

Noticed while confirming that an unrelated change (#591/#592, the account brief) owed no paid
lane. It did not: the gate answers identically with and without it, and these three were already
stale on `origin/main`.

dungbanguyen:
## Decision, 2026-08-31: turn the gate red, then pay for gemini first

Re-triaged against `origin/main` at `8ecb691f1`. **Worse than filed.**

When this was written all three records named the same stale version. They now
name three *different* ones:

| record | prompt_version |
|---|---|
[...]

PR #777, opened 2026-08-10
fix(brief): the account brief reads amounts in the right scale, and knows an open task is open (#591, #592)

### LINKED ISSUE #591: Account brief reports deal amounts 100x too large (minor units sent to the model raw)
On an account with a 180,000.00 EUR deal, the brief writes:

> With a deal value of 18 million EUR, this opportunity represents a significant enterprise-level engagement.

The state strip on the same screen says EUR 180,000.00, so the page contradicts itself in AI-authored, customer-facing text.

Cause: `orgbrief.DealIn` (internal/compose/orgbrief/input.go) serialises `amount_minor` straight into the prompt JSON, and no prompt text says those are minor units. The model reads 18000000 as the major unit.

`orgbrief/ask.go:307` does divide by 100 for the same figure, so the Ask path is right and the Brief path is wrong — the two disagree on the same number.

Note for the fix: `/100` is not correct either. Zero-decimal currencies (VND, JPY, KRW) have no minor unit, so dividing understates them 100x. `internal/compose/offerdraft_price.go` already carries the ISO-4217 exception table as `minorUnitDigits`, but it lives in package `compose`, which `compose/orgbrief` cannot import (compose imports orgbrief). `internal/shared/kernel/values` already owns `Money` and is importable from both — that is the natural home for one shared table.

### LINKED ISSUE #592: Account brief claims open tasks were completed
On an account whose two tasks are both OPEN (one overdue), the brief writes:

> You recently completed tasks related to the contract walkthrough and the submission of the master services agreement.

The Next steps card immediately above shows the same two tasks as open, one flagged Overdue, so the page contradicts itself.

Cause: the same task reaches the model twice in `orgbrief.Input` (internal/compose/orgbrief/input.go):

1. as `tasks` — `TaskIn`, documented as "one open task", carrying id/name/due
2. as `recent` — `ActIn`, a timeline item with `kind: "task"`, carrying id/kind/subject/at and **no completion state**

Nothing in the second shape says the task is still open, and nothing links the two shapes to each other, so the model infers completion from a past-dated timeline entry.

`Activity` on the wire does carry `is_done` and `done_at` (contracts api_gen.go), so the state is available to pass. Either give `ActIn` that state, or drop from `recent` the task rows already listed under `tasks`.

tinnhlt-gra:
Closes #591, closes #592.

Two defects on one AI-authored, customer-facing screen. Both made the page contradict its
own cards, two inches above the prose.

## The money (#591)

The prompt carried minor units under a key that said `amount_minor` to this file and
nothing at all to the model. A 180,000.00 EUR deal reached it as `18000000`, and the brief
wrote:

> With a deal value of 18 million EUR, this opportunity represents a significant
> enterprise-level engagement.

while the state strip on the same screen said EUR 180,000.00.

**`/100` is not the fix either.** A zero-decimal currency has no minor unit — ¥18,000,000
*is* eighteen million yen — so dividing understates VND, JPY and KRW a hundredfold. The
ISO-4217 table that already existed for the offer-draft price check moves to
`shared/kernel/values`, beside `Money`, because the two callers sit on opposite sides of a
dependency edge: the price check is in `compose`, the brief is in `compose/orgbrief`, which
`compose` imports. Neither can hold the table the other needs; Tier-0 can. The duplicate is
retired.

**Derived at marshal time, not stored beside the integer.** The package still sums minor
units — the deterministic fallback totals open deals, and summing rendered decimals would
reintroduce the rounding an integer exists to prevent — but a rendered copy kept as a
second field is a second number a caller can set on its own. Taking it as the JSON is
written means there is only ever one, and the same rendering serves the Ask path, which had
its own `/100`.

**The won-to-date total had the identical defect and the issue names only the open deals.**
Both are fixed; all three `/100`s in the package are gone.

## The tasks (#592)

The same task reached the model twice — once under `open_tasks`, once on the timeline as a
past-dated row with **no state** — and nothing linked the two shapes or said the second was
still outstanding. The model did the reasonable thing with a dated entry:

> You recently completed tasks related to the contract walkthrough…

directly above a card showing one of them **Overdue**.

A timeline item that *can* be finished now says whether it has been. One that cannot says
nothing: a call is neither done nor outstanding, and a plain `false` would invent a state
the record does not have.

## Proof

The new tests assert **the JSON the model is shown**, which is where both bugs lived —
every existing test asserted the prose, which is why neither was caught:

| case | asserted |
|---|---|
| two-decimal | `"amount":"180000.00"`, and **not** `18000000` |
| zero-decimal (JPY) | `"amount":"18000000"`, and **not** `180000.00` — the `/100` failure |
| three-decimal (KWD) | `"amount":"18000.000"` |
| no currency | no `amount` at all — a figure without its code is a number whose scale the reader must guess |
| won lifetime | `"won_lifetime":"12000.00"`, and **not** its minor integer |
| open / done task | `"done":false` / `"done":true` |
| a call | no `done` key |

`make check-backend` green, `craft static --strict` PASS (0/0/0), `make test-integration`
`OK: integration passed with 0 skips (29 packages)`.

## No paid aicert lane is owed — established, not assumed

`prompt_version` digests the requests this build constructs, so a change to how a site
serialises its input *can* move it. This one does not:

- `TestEveryCommittedRecordNamesTheCurrentPromptVersion` answers **byte-identically** with
  and without this change (verified by stashing it and re-running).
- There are **no `summarize` records at all** — `org_brief` has corpus scenarios but was
  never certified, so there is nothing to go stale.
- The three records the gate *does* report stale are `agent_loop`, and they were already
  stale on `origin/main` from #739. Filed as **#776**, since that gate only WARNS inside a
  passing test.

---
## Summary by cubic
Fixes money scale across the org brief and account draft, and correctly marks open tasks and meeting outcomes so prose matches the cards. Addresses #591 (amounts) and #592 (tasks).

- **Bug Fixes**
  - Render deal amounts and won-to-date totals as major-unit strings in model JSON using ISO‑4217 minor-digit rules; remove all /100 assumptions.
  - Apply the same major-unit rendering to the `accountdraft` prompt; zero-priced deals show "0.00", amounts without a currency are omitted.
  - Add `done` to timeline tasks and include meeting `status` (booked/held/canceled/no_show).

- **Refactors**
  - Centralized ISO‑4217 digits and rendering in `values.MinorUnitDigits` and `values.MajorUnits`; used in `compose/orgbrief`, `compose/accountdraft`, `compose/offerdraft_price`, and `overlay`. Table adds four‑digit codes (CLF, UYW), UYI, and codes with no minor unit (XAU, XAG, XPT, XPD, XDR, XXX, XTS).
  - `ask` and deterministic text use the shared renderer; prompt version bumped to v6.
  - Added tests asserting model-input JSON for amounts and task/meeting state, including negative and int64 edge cases; timeline tests use a fixed instant (no wall clock).

<sup>Written for commit 3f8905e3521929e453236bf2aa8647a829fc4e9a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/777?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Monetary amounts are now displayed in currency-appropriate major units, including currencies with zero, three, or four decimal places.
  * Deal and lifetime totals are shown only when valid currency information is available.
  * Task completion states and meeting statuses are now included in organization summaries.

* **Bug Fixes**
  * Corrected formatting for negative, zero, and boundary monetary values.
  * Prevented unsupported or misleading activity outcome fields from appearing in summaries.


tinnhlt-gra:
## Both reviews landed — eight findings, all taken

Fable and Codex reviewed the finished diff independently. Three were the same shape as the
bug being fixed: **a claim the code did not hold.**

**The draft prompt had the identical defect.** `compose/accountdraft` serialises the same
`amount_minor` into the same kind of account summary — and it is the *more* consequential
half, because that prompt writes an outbound message and explicitly licenses stating figures
the summary supplies. A model told 18000000 can put eighteen million in front of the person
paying it. Fixed the same way, with its own tests.

**This PR claimed to retire a duplicate table while a THIRD one stood.**
`overlay.ISO4217MinorUnitExponent` carries the same ISO fact under a comment promising *"one
spelling of the ISO fact, never a second copy that could drift"* — and the two had already
drifted: overlay knew `UYI`, `CLF` and `UYW`; the new table knew none of them. Overlay now
delegates, the table holds the union, and that comment is true for the first time. The claim
in the description above was wrong when I wrote it.

**A meeting can be finished, and my comment said only tasks could.** Its states are not a
boolean — `canceled` and still-to-come are both not-held — and a booked meeting is dated at
its **slot**, so one that never happened arrives on the timeline looking exactly like one
that did. That is #592's mechanism on the kind whose dates run forward. A meeting now
carries its own outcome in its own vocabulary.

The rest, each verified before being taken:

| finding | disposition |
|---|---|
| a zero-priced deal lost its amount and read as unpriced | fixed — the DDL ties nullness, not positivity, so zero is a price somebody chose |
| the table claimed to list every non-two-digit code, listed no four-digit ones | fixed — says what it is, and carries CLF, UYW, UYI |
| `MajorUnits(math.MinInt64, …)` printed a sign in front of a negative quotient | fixed — the magnitude is taken unsigned, the one arithmetic that works for it |
| `promptVersion` unmoved though the input shape changed | bumped to `org-brief-v6`, per the constant's own rule |
| **the #592 mapping had no test** | fixed — every case built `ActIn` by hand, so deleting the fold that fixes the bug left the suite green. It now runs through `FromView` from a real `Organization360` |

Verified by mutation, not by inspection: deleting the fold fails
`TestATaskOnTheTimelineCarriesWhetherItIsDone`; restoring the zero-suppression fails
`TestAZeroPricedDealStillCarriesItsAmount`.

**Still no paid lane owed** — re-checked with the accountdraft change included. The
prompt-version gate is byte-identical with and without this branch.

Gates re-run after all of it: `make check-backend` green, `craft static --strict` PASS
(0/0/0), `make test-integration` `OK: integration passed with 0 skips (29 packages)`.

tinnhlt-gra:
Both threads addressed — one taken whole, one taken in part with the evidence for the part declined.

**Fixed timestamp: taken.** The timeline fixture asked what time it was four times and needed to know none of them. One fixed instant now, per the repo's own T11 rule.

**"Do not render an omitted ISO scale as two decimal places": half taken.** The comment asked for two things that point opposite ways.

*Naming the codes ISO gives no minor unit is right, and they are named* — `XAU`, `XAG`, `XPT`, `XPD`, `XDR`, `XXX`, `XTS`. For those the integer IS the amount, and two digits invents a scale the code does not have. Covered by a new case.

*Refusing an unnamed code is not, and would be a worse version of the bug this PR fixes.* The map holds the **exceptions**, not the supported set:

| | count |
|---|---|
| exception codes in the table | 26 |
| ordinary two-digit codes deliberately absent (GBP, CHF, SEK, AUD, CAD, INR, BRL, …) | most of the world |

Every ordinary currency reaches the default *because* two digits is what it carries — which is also ISO-4217's own default. Returning "unavailable" for anything unnamed would suppress the amount for nearly all of the world's money while admitting the two dozen unusual ones, so the brief would go back to being unable to state a deal's value — the exact defect #591 is about, reached from the other side.

The residue is real and now stated rather than implied: a code that IS an exception and is missing renders at two digits and is wrong for that code. The repair is one line in the table, not a scale withheld from everyone else. A new test pins both directions — `XAU` → `18000000`, `GBP` → `180000.00` — so neither can drift.

Gates re-run: `make check-backend` green, `craft static --strict` PASS (0/0/0).

PR #793, opened 2026-08-10
refactor(mcp-apps): the views move to frontend/, and the api reads them

### LINKED ISSUE #742: The MCP App views belong in frontend/, so they share the design system
Follow-up from #739 (the `ui://` App runtime). That PR deliberately did **not**
solve this, and it should not stay unsolved.

## Where it stands

#739 embeds each view as hand-written HTML + CSS + vanilla JS under
`backend/internal/modules/agents/apps/`, `go:embed`-ed and assembled in Go. That
was chosen for one narrow reason: a document with its stylesheet and scripts
**inline** can declare an empty origin allowlist, so "this view reaches no
network" is kept by having nothing to name rather than by an allowlist somebody
maintains — and the whole security argument stays finite.

It buys that at three real costs.

**1. The design system is not shared.** `frontend/src/design-system/` is where
every interactive control in this product comes from, and CLAUDE.md says so.
These views re-hand-roll type, spacing, colour tokens and the light/dark contract
in `apps/app.css`. Two definitions of one visual language diverge, and the one
nobody looks at is the one served to a third-party host.

**2. There is no test lane for them.** The Sonar coverage exclusion #739 added
states this openly: the FE lcov report comes from vitest over `frontend/`, which
never sees `backend/**`, so the views would report a false 0%. Their behaviour is
gated instead by a Go sweep that reads the assembled document as a **string** —
which is why that sweep needed a comment stripper, and why the stripper then
needed its own assumptions gate after two review passes found holes in it. All of
that machinery exists because the assets live where no JS tooling can reach them.

**3. Every check is a substring check.** The sweeps are explicitly a ratchet, not
a proof — they cannot see a computed property (`node['inner'+'HTML']`). A view
built in `frontend/` would be linted by biome, typed by tsc, and unit-tested in
jsdom: real analysis instead of grep.

## Probable shape

The views become a small Vite build target in `frontend/` with their own entry,
sharing the design-system tokens, emitting **one self-contained HTML file per
view** with everything inlined — which is what the empty-allowlist posture needs.
The Go side keeps only the resource provider and its descriptors, `go:embed`-ing
the built artifact rather than hand-written source.

## What has to be decided first, because it is not free

- **The api serves no SPA today** (`cmd/api` serves `/v1` only; the frontend is a
  separate static build), so a built asset has to reach the Go binary somehow —
  either committed build output, which this tree avoids elsewhere, or a build step
  that `make check` and CI both run before the Go build. That new cross-language
  ordering in the build graph is the real cost.
- **The empty-origin guarantee must survive bundling.** A bundler that emits a
  `<link>`, a font URL or a source-map comment reintroduces the origin the sandbox
  declaration says is not there. Keep the existing sweep and point it at the
  BUILT file, where it becomes a check on the bundler instead of on hand-written
  source.
- **Whether the design system can be used here at all.** It is React; a view is a
  document a third-party host sandboxes. Sharing the **tokens** (colour, type,
  spacing) is clearly right. Sharing components pulls React into every view and
  makes each document much larger, which a host fetches per render.

Two review comments on #739 were answered with "revisit when this moves" rather
than churned line by line — the off-origin sweep's raw-read scope, and copy voice
in the view text. Both belong in that pass.

tinnhlt-gra:
Closes #742.

The two MCP App views move out of `go:embed`-ed backend assets into
`frontend/src/mcp-apps/`, build into self-contained inlined HTML documents, and
the api fetches them over HTTP, validates them, and holds an immutable per-URI
snapshot that `resources/list`, `resources/read` **and** each tool's `_meta.ui`
all read from.

#742 named three costs of embedding, and this is what each buys:

| the cost | now |
|---|---|
| the design system is not shared | `view.css` reads `tokens.css` — the accent, the text roles, the border, the 4/8/12/16/24 spacing scale. The colours and rhythm visibly change (table below) |
| there is no test lane for them | 37 jsdom specs over the bridge and both renderers, 31 over the inliner, its two validators and the dev middleware, 51 over the Go fetch/admit/hold half, and a Playwright zero-request run — and the Sonar coverage exclusion is **deleted**, not moved |
| every check is a substring check | the build now runs a **parsed** pass over the document's nodes on top of the shared token list, and a browser proves what neither can see |

## The shape

`frontend/src/mcp-apps/{bridge,types,view.css,forbidden.json}` plus a directory
per view. One single-input vite build per view (`--mode <view>`) folds the entry
chunk and the stylesheet in, strips CSS comments, stamps the SPDX header after
the doctype, and asserts **exactly one file** — the cardinality check that
catches a worker, a wasm module, an `import.meta.url` sibling and a copied public
file at once. nginx serves them at `/mcp-apps/`; `try_files $uri =404` is
load-bearing, or the SPA fallback answers a missing view with HTTP 200 and the
api fetches the app shell believing it is a view.

Go keeps the catalog, the descriptors and `sandbox()`. `document()` and the
`go:embed` are gone; `fetch.go`, `validate.go` and `hold.go` are new.

## Availability — the blocker both review rounds found

`appsServed()` was derived from the registry and said so in its own comment, and
`ToolSpec.UI.ResourceURI` is a constant baked at registration — so a view whose
fetch failed was **still named by its tool**, and a host that prefetches it gets
a 404 and a panel that silently never appears. Availability is now per-URI and
read from one snapshot by all three surfaces. One view missing suppresses **one**
tool's `_meta.ui`; that tool keeps answering in text.

The advertised set freezes after a bounded startup fetch, because this server
declares `resources.listChanged: false` — the notification travels on a stream
this transport does not open, so a view that appeared after a background retry
would be invisible to any host that listed at connect time. A refresh replaces
only what is already held, and a failure keeps the last known-good copy.

## What the live stack found that no gate did

Both reviewers found it independently and the first `make dev` run confirmed it:
**`dev.sh` started vite strictly after the api**, so priming always met a cold
origin and both views were permanently unadvertised in *every* dev stack. Fixed
on both sides — the startup fetch re-asks a tier that is still coming up (a
rolling deploy hits the same thing), and `make dev` starts the FE first.

## Two adversarial review passes

Fable and Codex, independently, against the diff. Beyond the above:

- **One value, two readings — a configured origin.** Flag parsing admitted any
  bare origin while the fetch refuses cleartext to anything but a literal local
  address, so `http://web.internal:8080` — the private-network case the setting
  exists for — started cleanly and served nothing. Both sides ask
  `apps.ValidateOrigin` now; it is a boot error naming the setting, and the rule
  is documented.
- **One value, two readings — the shared vocabulary.** Go decoded it into a
  struct with a member per class, so a fifth class added upstream would be
  enforced by the frontend and silently ignored by the api *while the
  byte-equality test kept passing*, because the bytes really are identical.
- **A gate that would not have run.** `ci.yml` classified the authored
  `forbidden.json` as frontend-only, so a PR editing only it never ran the lane
  carrying the parity test.
- **Refusals that could be walked past.** An uppercased `<LINK`/`SRC=` matched
  neither validator; `xlink:href` matched neither the token list nor a plain
  `hasAttribute`. Both sides case-fold HTML-shaped tokens by the same derived
  rule, and the parsed pass matches an attribute's local name.
- **A silent rewrite.** The CSS comment strip was a regex, so
  `content:"/* label */"` became `content:""`.
- **Retrying what retrying cannot change**, a log line that reported an outage
  that was not happening, a test that asked a second provider whether it served
  what the first had advertised, and real clocks in unit tests (124s → 2s under
  `-race`).

## Visible change (UAT proof 5)

Measured through `getComputedStyle` on the real documents, driven through the
real handshake, before and after. Screenshots taken for all eight combinations.

| view | theme | era | body bg | body fg | font | score | row padding | head gap | band |
|---|---|---|---|---|---|---|---|---|---|
| account-brief | light | before | `rgb(255,255,255)` | `rgb(17,24,39)` | -apple-system | `rgb(29,78,216)` | 10px 12px | 6px | — |
| account-brief | light | **after** | `rgb(255,255,255)` | `rgb(54,67,61)` | system-ui | `rgb(11,122,83)` | 12px | 8px | — |
| account-brief | dark | before | `rgb(11,17,32)` | `rgb(229,231,235)` | -apple-system | `rgb(147,180,253)` | 10px 12px | 6px | — |
| account-brief | dark | **after** | `rgb(20,36,32)` | `rgb(195,206,200)` | system-ui | `rgb(22,163,74)` | 12px | 8px | — |
| relationship-map | light | before | `rgb(255,255,255)` | `rgb(17,24,39)` | -apple-system | — | 10px 12px | 6px | `rgb(4,120,87)` |
| relationship-map | light | **after** | `rgb(255,255,255)` | `rgb(54,67,61)` | system-ui | — | 12px | 8px | `rgb(21,128,61)` |
| relationship-map | dark | before | `rgb(11,17,32)` | `rgb(229,231,235)` | -apple-system | — | 10px 12px | 6px | `rgb(52,211,153)` |
| relationship-map | dark | **after** | `rgb(20,36,32)` | `rgb(195,206,200)` | system-ui | — | 12px | 8px | `rgb(74,222,128)` |

The blue accent becomes the product's emerald, the dark ground becomes the
product's dark green rather than a slate-blue of its own, spacing lands on the
scale, and the font stack is `system-ui` — **not** `var(--f-body)`, which
resolves to a web font a document with an empty origin allowlist can never load.
Every warmth band was re-measured against 4.5:1 on `--bgElevated` in both themes
(recorded by token name in `view.css`, because `check-ds-purity.sh` greps
comments too).

## Deviations from the approved design, with reasons

1. **jsdom via a per-file `// @vitest-environment` pragma**, not
   `test.environmentMatchGlobs` — deprecated in vitest 3.2 and it warns on every
   run of the whole suite. The design names the pragma as the alternative.
2. **The three-state theme contract is resolved in `bridge.ts`**, not by a
   `@media (prefers-color-scheme: dark)` arm in `view.css`. The design's arm is
   *unimplementable* under the design's own other constraint:
   `check-ds-purity.sh` fails any hex outside `tokens.css`, and the dark palette
   is literal values. Resolved at the root cause the design itself identified,
   with the pattern `src/app/theme.ts` already uses for the identical problem on
   the unauthenticated surface. There is now **no** second partial dark palette
   anywhere — strictly better than what it replaces.
3. `list()` lives once, in `types.ts` as `asList`, rather than duplicated on
   `bridge.ts`.
4. The per-view build is selected by vite `--mode`; a `.mjs` importing the TS
   config factory cannot resolve that config's own extensionless import under
   Node's type stripping.

## Gates

`make check` and `make check-fe` green; `craft static --strict` clean over the
whole tree; the Playwright zero-request lane green — and **verified to fail**
when a computed sink (`window["fet"+"ch"](url)`) is injected into a built
document, which is the only way to know a zero-request assertion is doing
anything.

**No aicert re-run is owed**, verified rather than assumed: no tool registered or
retired, no tool description, prompt or corpus text touched.

MCP Inspector UAT to follow as a comment.

## Summary by CodeRabbit

* **New Features**
  * Added MCP App views for account briefs and relationship maps.
  * MCP App documents are securely fetched, validated, refreshed, and served when available.
  * Tool responses now show UI metadata only for views currently being served.
  * Added support for themed, sandboxed view rendering and build-revision diagnostics.

* **Documentation**
  * Documented MCP App base URL configuration, defaults, reachability requirements, and validation rules.

* **Bug Fixes**
  * Improved startup handling and preserved previously available views during refresh failures.


tinnhlt-gra:
## UAT — live stack, MCP Inspector + raw curl

Driven by an agent with no implementation context, against a `DEV_SLUG=742` stack.
Raw output throughout. One tooling note that matters: **the Inspector's SDK strips
`_meta` from `tools/list`**, so every `_meta.ui` claim below is from raw curl, not
the Inspector — otherwise you chase a server bug that isn't there.

### 1. Both `ui://` URIs read, and the api served what the build produced — PASS

`resources/read`'s `text` against what the web tier is serving:

```
account-brief:    read.text sha256=245e05e8…e037c59 len=11164
                  web fetch sha256=245e05e8…e037c59 len=11164   identical=True
relationship-map: read.text sha256=00820f4e…31753f len=11628
                  web fetch sha256=00820f4e…31753f len=11628   identical=True
```

Both under `mimeType: text/html;profile=mcp-app`, both carrying `_meta.ui` with
four **empty** allowlists and **no** `permissions` member:

```
"_meta": {"ui": {"csp": {"connectDomains": [], "resourceDomains": [],
                         "frameDomains": [], "baseUriDomains": []},
                 "prefersBorder": true}}
any null-valued member present: False
'permissions' appears anywhere: False
```

`[]` and not `null` is the load-bearing detail — a host reads an absent list as
"unspecified", which is where a permissive default lives.

### 2. Views origin stopped — the invariant two review rounds found broken — PASS

Vite killed, api restarted with its exact recorded argv:

```
ERROR msg="mcp apps: a view document is not being served" uri=ui://margince/account-brief.html err="… connect: connection refused"
INFO  msg="mcp apps: view documents primed" held=0 catalog=2 origin=http://localhost:8881
WARN  msg="mcp apps: some views are not served; the advertised set is fixed until this api restarts" held=0 catalog=2
```

**(a) unadvertised** — `resources/list` returns the query vocabulary alone;
`resources/read` of a view answers `-32602 no resource at ui://…`.

**(b) both tools still answer in text** — 18 tools listed, `read_brief` and
`who_knows` both present and both **callable**, returning the same answers they
gave while the views were held.

**(c) `_meta` absent from exactly those tools, extension unadvertised**:

```
tool count: 18
tools carrying _meta: []
  read_brief: listed=True has_meta=False
  who_knows:  listed=True has_meta=False
server/discover capabilities: "extensions": { "io.modelcontextprotocol/tasks": {} }
```

The App extension is gone from `server/discover` — a host that saw it advertised
would be entitled to prefetch a document, and there is none. Restored afterwards:
`held=2 catalog=2`, both URIs listed again.

### 3. A document with a `<link>` → refused, alarm raised, last-known-good retained — PASS

A stand-in origin served a bad `account-brief.html` and a verbatim-real
`relationship-map.html`:

```
ERROR msg="mcp apps: a view document is not being served" uri=ui://margince/account-brief.html
      err="crmapps: refused for a reason re-asking cannot change: the document reaches beyond what a view may: [<link]"
INFO  msg="mcp apps: view documents primed" held=1 catalog=2
```

The refusal names the token it found. Partial availability, exactly as designed:

```
resources: ['margince://schema/query', 'ui://margince/relationship-map.html']
read_brief -> _meta: null
who_knows  -> _meta: {"ui": {"resourceUri": "ui://margince/relationship-map.html", …}}
margince_mcp_app_view_held{uri="…account-brief.html"} 0
margince_mcp_app_view_held{uri="…relationship-map.html"} 1
margince_mcp_app_admission_failures_total 1
```

**Retention was observed, not inferred.** The *held* relationship-map was swapped
to a `<link>`-bearing copy at 14:28:35; the real 5-minute tick fired at 14:33:02:

```
14:33:02 WARN msg="mcp apps: a view could not be re-read; the last known-good document is still being served"
         uri=ui://margince/relationship-map.html err="… reaches beyond what a view may: [<link]"

served sha256: 00820f4e…31753f  len 11628   still last-known-good: True
'<link' present in served doc: False
margince_mcp_app_admission_failures_total 2   (1 -> 2)
```

The view stayed listed and readable, serving the pre-swap bytes — which is the
whole point: one bad response during a rolling deploy must not take a working
view down.

### 4. A deliberate build-stamp mismatch → reported, not refused — PASS

An origin serving both real documents with the stamp rewritten:

```
WARN msg="mcp apps: the served view was built from a different revision than this api;
     expected during a rollout, worth alerting on once one has finished"
     uri=ui://margince/account-brief.html
     api_revision=94844a98abb897ce4f24a779ad66620d42b72732
     document_revision=deadbeefdeadbeefdeadbeefdeadbeefdeadbeef
INFO msg="mcp apps: view documents primed" held=2 catalog=2
```

Both views **still served** (`resources/read` → HTTP 200), and the gauge is per
URI:

```
margince_mcp_app_view_held{uri="…account-brief.html"} 1
margince_mcp_app_build_skew{uri="…account-brief.html"} 1
margince_mcp_app_build_skew{uri="…relationship-map.html"} 1
```

`/metrics` gating behaved as documented: `401` without the Bearer, `404` on an api
with no `--metrics-token`.

### 5. Rendered in light and dark, before and after

The measured table is in the PR description. Both views were driven through the
real handshake as a top-level page (`window.parent === window`, so the sender
check passes on a message the page posts to itself) and screenshotted in all
eight combinations. The visible change: the blue accent becomes the product's
emerald, the dark ground becomes the product's dark green rather than a
slate-blue of its own, spacing lands on the 4/8/12/16/24 scale, and every warmth
band was re-measured against 4.5:1 on `--bgElevated` in both themes.

### Also proven, off the live stack

The Playwright zero-request lane was **verified to fail** when a computed sink —
`window["fet"+"ch"](url)` — is injected into a built document. A zero-request
assertion that has never been made to fail is indistinguishable from one that
does nothing.

tinnhlt-gra:
Thanks — six of the nine landed as fixes in `ba22644`, and two of them mattered a lot.

**Fixed**

- **`Resources` built a listing from two snapshots** (`apps.go`). It asked `Holds()` per catalog entry and each call re-loaded the pointer, so a refresh landing between iterations could advertise a pair no single immutable set ever contained. This is the exact property the snapshot exists to provide, so getting it wrong here was the whole design being wrong quietly. One load, whole answer derived from it.
- **The case rule covered half the language.** Keying case-insensitivity on a leading `<` or trailing `=` missed every CSS and bare-attribute token: `@IMPORT`, `URL(`, `IMAGE-SET(`, `SRCSET`, `HTTP-EQUIV`, `JAVASCRIPT:`. It now reads the **token**: an uppercase letter marks a JS identifier and is matched exactly; an all-lowercase token is markup or a URL scheme and is matched loosely. That distinction is load-bearing rather than tidy — `Function(` folded to lowercase matches every `function(` in every view, and the check would refuse everything. Both languages apply the identical rule and both now carry the test that proves the exception.
- **408 and 429 were permanent refusals** (`fetch.go`). A tier answering 429 while everything else boots would have left a view unadvertised for the process lifetime — precisely the failure the design says it must not have.
- **The document story could miss the handshake** (`story-hosts.tsx`). `srcDoc` as a prop let the frame load and announce before the effect installed the listener, and the bridge announces once and never retries. The document is loaded inside the effect now, after the listener.
- **`asWarnings` filtered rather than rebuilt** (`types.ts`), so extra members flowed through and `message` was typed `string` without being checked — the "types are not validation" mistake in the file written to prevent it.
- **`primeBriefly`'s 100 ms wall clock** is gone, along with every other one in the package: the tests whose view stays broken now answer a permanent 404 (so `Prime` stops immediately and needs no deadline), and the one test that is *about* the deadline cancels from inside the responder.

**Rejected, with reason**

- **Naming `make -C backend mcp-apps-vocab` in the `go:embed` comment.** That is not build-process narration in the sense the rule means — it is how a reader regenerates a generated file, which every generated file in this tree states. The rule targets review-ticket numbers and fix narration, and the failure message of the test that catches the drift says the same thing for the same reason.
- **The orientation half of the package comment.** "Authored in `frontend/src/mcp-apps`, fetched over HTTP, admitted, held" is what a reader needs to know before anything else in the package makes sense. You were right about what sat beside it, though: the narration of the rejected embedding alternative — node in CI jobs, build ordering, caches — belonged in the design doc rather than the source, and it is gone.

Issue #813, opened 2026-08-10
update_record's MCP write ignores the released approval pin the REST door forwards


tinnhlt-gra:
## What

A redeemed approval carries the version it was granted against, and the write it releases
is supposed to be conditioned on that version — because redemption commits its own
transaction and the handler then opens a fresh one, so the skew check inside redemption
proves the row was right *when the approval was consumed*, not when the effect lands.

Both doors implement that, differently, and `update_record` falls through the MCP one:

- **REST** forwards the pin as the request's own `If-Match`
  (`compose/agentgatestaging.go`, `redeemIfPresented`), so every operation behind that
  door gets it, `update_record` included.
- **MCP** carries it through `agents.pinForWrite`, which only three tools call:
  `advance_deal`, `progress_deal` and `advance_project_phase`.
  `updateRecord.applyRecord` passes `args.IfVersion` raw
  (`backend/internal/modules/agents/tools_update.go`), so a redeemed `update_record`
  retry that supplied no `if_version` writes unconditioned.

One credential, two doors, different guarantees — the asymmetry ADR-0055 exists to close.

## A sibling worth deciding at the same time

`disqualify_lead` pins `TargetVersion` at staging, but its seam — `DisqualifyLead(ctx, id)`
— cannot carry a version at all, so the pin it stages can never be applied. That is a
seam-shape question rather than a missed call site, and it belongs in the same change.

## Shape of a fix

Route every 🟡-stageable tool's write through `pinForWrite` rather than through the
caller's argument, and derive the obligation: *a tool that stages a target version applies
one*. A fitness test over the registry can hold that, the way
`stagingfitness_test.go` already walks the stageable set.

Predates #614; found by an independent review of that branch and verified against the tree.

dungbanguyen:
picking this up

Issue #812, opened 2026-08-10
The REST agent gate ignores a presented X-Approval-Token when a dynamic call resolves 🟢


tinnhlt-gra:
## What

The two doors disagree about what a presented approval token means on an
**auto-executed** call.

- **MCP** redeems it whatever the tier resolved to. `Registry.Invoke`'s 🟢 arm calls
  `runClaimed` → `redeemPresented`, and the comment says why: the retry of a per-field
  precedence staging admits at the auto-execute tier, so its asserted authority is
  consumed and validated against the identical-call hash, never ignored.
- **REST** only redeems inside `splitOrRedeemUpdate`, which runs for `update_record`
  alone. Every other tool's 🟢 arm forwards straight to the handler
  (`compose/agentgate.go`, `admitAgentCall`), so an `X-Approval-Token` on the request is
  neither validated nor consumed.

## Why it matters

A `advance_deal` staged as 🟡 and then re-issued after the deal's own state changed can
resolve 🟢 on the retry. On REST that call runs while:

- the approval it asserts stays **pending** — an authority object nobody consumed, still
  decidable, still visible in the inbox;
- an **invalid or replayed** token is accepted silently, because nothing looks at it.

Neither is an escalation on its own — the call was admitted at 🟢 on its own merits — but
one credential governed two ways is exactly what ADR-0055 exists to prevent, and a
pending approval that has already been acted on is a lie in a human's inbox.

## Evidence

- `backend/internal/compose/agentgate.go` — `admitAgentCall`'s `outcome.err == nil` arm.
- `backend/internal/compose/agentsplit.go` — `splitOrRedeemUpdate`, the only redeeming
  path on that arm.
- `backend/internal/modules/agents/registry.go` — the MCP arm that does redeem, with the
  comment stating the rule.

Predates #614; found by an independent review of that branch and verified against the tree.

Issue #811, opened 2026-08-10
advance_deal: the tier binds the deal's version but not the stage semantics it was also resolved from


tinnhlt-gra:
## What

#614 bound the auto-executed deal move to the **deal's** version: the tier gate reads the
deal to prove both endpoints of the move open, and the write is now conditioned on the
version that read saw. A second input to the same decision is not bound.

`agents.DealMoveTierInput` resolves the tier from three reads:

- the TARGET stage's semantic (`stages.StageSemantic(ctx, toStageID)`),
- the deal's current stage id,
- the SOURCE stage's semantic.

Only the deal's version travels forward. Stage semantics live on `stage` rows that are
mutable independently of any deal, so a stage-config change in the window is invisible to
the pin.

## The window

1. the gate reads the target stage as `open` and the deal as open at v7 → resolves 🟢;
2. an admin changes that target stage's semantic to `won`;
3. `deals.Store.AdvanceDeal` re-derives won/lost from the target stage's semantic **inside
   its own transaction** (`deal_advance.go`, `resolveAdvanceTarget`) and reads `won`;
4. the deal is still v7, so the version compare passes, and the deal is **closed**
   unattended — on a tier that was resolved for an open→open move.

The mirror case closes the loop: changing a closed deal's current stage semantic to `open`
lets the same path reopen it.

## Why it was not fixed with #614

The racing actor here is a **human admin**, not the agent — pipeline and stage
configuration is `x-agent-access: human-only`, so an agent cannot arrange either side of
this window. That is a materially different risk from #614, where the agent controlled
both sides, and the fix is a different mechanism: the deal's `version` column cannot bind
a `stage` row, so this needs either a stage-config version carried through
`TierResolverInput` or the semantics re-checked against what the gate saw, inside the
write's transaction.

## Evidence

- `backend/internal/modules/agents/dealmove.go` — `DealMoveTierInput` carries
  `ObservedVersion` for the deal only.
- `backend/internal/modules/deals/deal_advance.go` — `resolveAdvanceTarget` re-reads the
  target semantic in-transaction; `ApplyGuarded` compares only the deal version.
- `backend/internal/modules/deals/stages.go` — `UpdateStage` mutates `semantic` behind its
  own If-Match, touching no deal row.

Found by an independent review of the #614 branch, verified against the tree.

dungbanguyen:
picking this up

Issue #773, opened 2026-08-10
advance_deal's tool copy tells an agent only half the confirm-first rule


tinnhlt-gra:
The tool description an agent reads says:

> "Moving onto a stage that closes the deal as won or lost is a decision a person makes:
> it is staged for approval…"

Target-only. Since #610 the resolver charges a move with a terminal stage at EITHER end to
the confirm-first tier (`agents/dealmove.go:44-49`), so **reopening** a won deal is staged
too. An agent working from the copy has no way to anticipate that: it reads a won deal,
moves it back to an open stage expecting a 🟢 auto-execute, and gets an approval
requirement it was not told about.

The contract now describes the shipped rule (#624). The copy is the model-visible half and
is deliberately NOT changed there, because it is a different kind of change:

- `agents/toolcopy_records.go:126` (`advanceDealCopy`) is rendered into the tool listing,
  so any edit moves `prompt_version` and makes the current `agent_loop` records stale —
  the paid aicert lane is a merge condition, not a nicety.
- Tool copy on this surface is **measured, not reasoned**. A single added sentence has
  broken an adjacent certified scenario before. Whatever wording is chosen has to be run
  on the certified bindings and compared, not argued for.

`progressDealCopy` (`:140`) carries the same target-only framing and should move with it —
they are one rule stated twice.

Worth measuring at the same time: whether saying it at all is net-positive. The rule only
bites on a reopen, which is rare, and the listing budget is tight (~1,355 tokens of
headroom before this). A wording that costs a scenario elsewhere to describe a rare case
correctly is not obviously the right trade — which is exactly why it wants a measurement
rather than a judgement.

Found while fixing #624.

tinnhlt-gra:
**Making this the host for the whole model-visible copy batch, because every item in it is blocked on the same thing and none of them justifies the paid lane alone.**

Triaged the agent-surface cluster (30 issues, open and closed) while re-measuring #737. Three issues are the same defect class — copy or schema that a model reads and that no longer matches behaviour — and all three are blocked identically: an edit moves `prompt_version`, which stales the `agent_loop` records, which is #776's paid pass. Batching them is the only way any of them lands.

## The batch

**1. This issue — `advance_deal` / `progress_deal` confirm-first rule.** Verified still present on `main`: `toolcopy_records.go:128-132` says the id "of the stage you are moving TO comes from list_pipelines" and frames the approval around "moving onto a stage that closes the deal", which is the pre-#610 target-only rule.

[...]

PR #814, opened 2026-08-10
fix(agents): an auto-executed deal move is bound to the record its tier was read from (#614)

### LINKED ISSUE #614: advance_deal/progress_deal: a concurrent close between tier resolution and the write can still reopen a deal ungated
## The window

#610 closed the unconditional hole: the dynamic tier resolver read only the destination
stage, so an agent could move a won deal back to an open stage on its own authority. It now
resolves from **both** endpoints, and a terminal stage at either end raises to confirm-first.

A narrower window survives it. The source stage is read in `ResolverInput`, and the write
happens later in `Handle`:

1. the deal is open, and the move is open→open, so the gate resolves 🟢 and admits it;
2. another actor closes the deal as won;
3. `AdvanceDeal` executes the open target against a deal that is now terminal.

The result is a reopen with no human in the loop — the same outcome, reached through a race
rather than through the rule.

## Why it was not fixed in #610

Closing it means binding what the gate SAW to what the write does: the resolver would have
to carry the deal version it observed forward, and `Handle` would have to pin it as the
write's `if_version` when the caller supplied none. That changes the write path and changes
what `pinForWrite` means for an unapproved call — today it exists for the approved-retry
case — so it wants its own review rather than riding a governance fix as an afterthought.

#610 is still worth having on its own: before it, no concurrency was needed at all.

## Shape of a fix

- `DealMoveTierInput` returns the deal's version alongside its stage semantic, and
  `mcp.TierResolverInput` carries it.
- On an auto-executed deal move with no caller-supplied `if_version`, `Handle` pins the
  observed version, so a deal that moved in between fails as version skew rather than
  applying.
- The test to write first is the interleaving itself: resolve, close the deal underneath,
  then execute — and assert the write is refused rather than reopening.

Both dynamic tools go through one builder now, so this is one change rather than two.

Found by the review bots on #610 (cubic, twice, from both tools).

tinnhlt-gra:
Closes #614.

## The hole

`#610` made the deal-move tier turn on **both** endpoints, so reopening a closed deal
raises to confirm-first. It proves both endpoints open from a read that commits before
the write does, and on the auto-execute path nothing bound the two:

| step | where |
|---|---|
| 1 | `agents.DealMoveTierInput` reads the deal to resolve the SOURCE semantic, and discards the record |
| 2 | `auth.Gate.Admit` resolves 🟢 and returns |
| 3 | `Handle` writes with `pinForWrite(ctx, args.IfVersion)` — **nil**, because `releasedPinKey` is only ever set by `RedeemAndMark` |

A close landing between (1) and (3) reopens a won deal at the 🟢 tier: `closed_at`, the
lost reason and the FX rate frozen at close are all cleared, with no human in the loop.
`pinForWrite`'s own comment already named the generalisation — *the agent controls both
sides of that window* — and applied it to the redeemed path only.

Both doors had it: MCP through `registry.tierResolverFor`, REST through
`compose/agentgate.go`'s `advanceDealTierInput`. One builder, so one fix.

## The fix

**The read that decides the tier already holds the version. Carry it.**

- `mcp.TierResolverInput` gains `ObservedVersion *int64`, filled by `DealMoveTierInput`
  from the record `dealStageSemantic` already reads — one read, one moment, no extra query.
- `auth.Gate.Admit`, the one admission point both doors call, puts it in the returned
  context **on the auto-execute outcome and nowhere else**. Set there rather than in each
  dispatch layer, for the reason `RedeemAndMark`'s comment already gives for redemption:
  a rule spelled twice on this seam is how the tier itself came to judge both endpoints
  on one door and only the destination on the other.
- MCP consumes it in `pinForWrite`; REST forwards it as `If-Match`, which is
  `redeemIfPresented`'s existing forward one tier down.

The store already refuses the mismatch — `Patch.ApplyGuarded` inside
`deals.Store.AdvanceDeal`'s transaction, which answers the contract's `version_skew` 409.

### Two things review changed, and both were real

**A caller's own pin may not disagree with what the gate read.** The first draft preferred
the caller's `if_version` / `If-Match` unchecked, on the reasoning that it is "stricter or
equal in every case that matters". That comment was false, and it is the defect class this
repo hunts by name. The caller controls the argument, so a version nobody read is a
version nothing proved: an agent that names the version the racing close will **produce**
walks straight through, because the store's compare then passes on precisely the record
the tier decision does not describe. That converts a coin-toss race into an armable trap —
**reproduced live, 25 reopens in 30 attempts** (table below). A disagreement now answers
skew, which is also what a caller holding a genuinely stale version already got, one layer
down.

**The seam fails closed as well as binding.** A dynamic tier that resolves to auto-execute
without naming the record it was read from is now **raised**, not run. Without it the
guard was advisory: a dynamic tool reporting no version would be admitted 🟢 and written
unpinned — #614's exact shape, restored by omission. Raising is the shape the resolver
contract already takes (it may only ever raise), and it is the honest answer for a record
carrying no version of its own: an overlay mirror row answers `0`, and zero conditions no
write (the REST door would emit `If-Match: 0`, which the contract's own parser rejects as
a malformed header the caller never sent).

## The derived gate

Review-loop rule 2 — derive the obligation, don't maintain a list.

- **`TestEveryDynamicTierToolPinsTheReadItsTierTurnedOn`** walks the registered tool set
  and requires every `TierDynamic` spec's `ResolverInput` to report the version its tier
  was read from.
- **`TestADynamicTierThatNamesNoRecordVersionIsRaisedRatherThanRun`** holds the production
  half, so the rule does not depend on the test finding every registrar.
- **`TestAdmitPinsNothingItDidNotAdmitAtTheAutoExecuteTier`** holds the other side: no pin
  on a static tier, on a raised tier, or for a human principal.

**Every one of these was verified by breaking it** — each guard was re-run against the
defect it describes and confirmed red first.

### Unit and integration

The race is driven against **real Postgres** in
`compose/dealmovepin_integration_test.go`: the concurrent close is a real `AdvanceDeal`
committed through the real provider, inside the window between the gate's read and the
write it admits. Everything is real except the timing — a concurrent close is otherwise a
coin toss, and a test that reproduces a race one run in fifty is a test that reports green.

```
make check-backend      exit 0
make craft-static       PASS — 0 blocker, 0 major, 0 minor (backend, extensions, fixtures)
make test-integration   OK: integration passed with 0 skips (31 packages, parallel)
```

Coverage on the new code: `pinAdmittedWrite` 100%, `pinForWrite` 100%,
`dealStageSemantic` 100%, `AutoExecutePin` 100%, `withAutoExecutePin` 100%,
`DealMoveTierInput` 85.7%, `resolveTier` 87.5%, `Admit` 96.9%.

The `agent_loop` prompt-version gate is **byte-identical to `origin/main`**, so no
certification record went stale and the paid lane owes nothing. No tool description, input
schema or contract prose moved.

### Live UAT

Two racers fired at the same instant against a real stack (`DEV_SLUG=pin614`): an agent
passport doing an open→open 🟢 move on `POST /v1/deals/{id}/advance`, and a human closing
the same deal as won. A **reopen** is the defect signature — the close returns 200 and the
deal nevertheless ends in an open stage with no `closed_at`.

| build | agent's pin | runs | reopened | refused as skew |
|---|---|---|---|---|
| `origin/main` | none | 40 | **36** | 0 |
| first draft of this fix | predicts the post-close version as `If-Match` | 30 | **25** | 0 |
| this PR | predicts the post-close version as `If-Match` | 30 | **0** | 30 |
| this PR | none | 40 | **0** | 40 |

The window is genuinely being hit — the refusals are the race landing, not the race being
missed.

And the routine paths still work, on the same stack:

```
no race, no caller pin ............................. 200, stage moved, version 2
no race, caller pins the version it read (2) ....... 200
no race, caller pins a version nobody read (99) .... 409 version_skew
    "If-Match 99 is not the version this record was read at (3) — re-read it and retry"
```

Through the MCP surface, driven by the official Inspector CLI
(`npx @modelcontextprotocol/inspector --cli`, `protocolEra: modern`, passport bearer):

- `tools/list` — 30 tools, `advance_deal` and `progress_deal` present;
- routine open→open move — succeeds, answers the record at version 2;
- stale `if_version` — `isError: true` with an actionable message, *"The record changed
  since it was read; re-read it and retry with the new version."*;
- reopening the now-won deal — still 🟡, staged as an approval with the redemption
  instructions.

Three findings from the two reviews are real, out of this change's scope, and filed rather
than implied away:

- **#811** — the tier is also resolved from stage **semantics**, which are mutable
  independently of any deal, so the deal's version cannot bind them. The racing actor
  there is a human admin (stage config is `x-agent-access: human-only`), not the agent, so
  it is a different risk needing a different mechanism.
- **#812** — the REST door ignores a presented `X-Approval-Token` when a dynamic call
  resolves 🟢, where MCP redeems it. Predates this branch.
- **#813** — `update_record`'s MCP write ignores the released approval pin that the REST
  door forwards as `If-Match`. Predates this branch; `disqualify_lead`'s seam cannot carry
  a version at all and belongs in the same change.

No tool copy changed: `advance_deal`'s description is rendered into the certified
`agent_loop` prompt, so editing it moves `prompt_version` and owes the paid lane — and copy
on this surface is measured, not reasoned. #773 already holds that question.

## Reviews

Reviewed independently by **Fable** and **Codex** before pushing. Both found the
caller-pin hole from opposite directions, which is why it is fixed here rather than filed;
Fable additionally found the fail-open seam, the overlay zero-version edge, and the
`pinForWrite` comment that had begun to overclaim. All four are addressed in the diff.

---
## Summary by cubic
Bind auto-executed deal moves to the exact record version the tier was resolved from to prevent race-driven reopens. Fixes #614 by carrying the observed version through `auth.Gate.Admit` and enforcing it on both MCP and REST paths.

- **Bug Fixes**
  - Added `ObservedVersion` to `mcp.TierResolverInput`; `DealMoveTierInput` sets it from the source deal read (ignored if 0).
  - `auth.Gate.Admit` pins the observed version only for auto-execute and raises when a dynamic tool resolves 🟢 without a version.
  - REST: `pinAdmittedWrite` forwards the pin as `If-Match`; rejects a mismatched caller header with `409 version_skew` (keeps a matching caller pin). Static tiers forward no pin.
  - MCP: `pinForWrite` enforces pin precedence (caller → released approval → admitted) and returns `version_skew` if the caller pin disagrees with the gate’s read; applied to `advance_deal`, `progress_deal`, and `advance_project_phase`.
  - Added unit, REST door, and real Postgres integration tests; derived checks require every dynamic tool to report and bind the read version; added a REST test proving static-tier creates pin nothing.

- **Migration**
  - Dynamic tier resolvers must set `ObservedVersion` when the decision turns on a record; missing versions will raise to approval.
  - If a tool forwards a caller `if_version`, it must match the gate‑observed version when present; disagreements fail with `version_skew`.

<sup>Written for commit 5ec70843928f895255777d25f57d2b66537c1279. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/814?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented automatically executed updates from overwriting newer record changes.
  * Added version checks to reject stale or conflicting writes with a conflict response.
  * Preserved matching caller-provided version constraints.
  * Required approval when automatic execution cannot confirm the record version.
  * Prevented concurrent deal changes from being unintentionally reopened.

* **Tests**
  * Added coverage for version pinning, concurrent updates, dynamic actions, and approval fallbacks.

### DIFF
--- a/backend/internal/compose/agentgate.go
+++ b/backend/internal/compose/agentgate.go
@@ -28,6 +28,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"strconv"
 
 	"github.com/go-chi/chi/v5"
 
@@ -226,6 +227,9 @@ type admissionOutcome struct {
 func admitAgentCall(w http.ResponseWriter, r *http.Request, next http.Handler, outcome admissionOutcome) {
 	switch {
 	case outcome.err == nil:
+		if !pinAdmittedWrite(w, r) {
+			return
+		}
 		// The effect is charged on BOTH arms — the field split forwards to the
 		// same handler through its own path — and on NEITHER when the handler
 		// refused. A quota counts what an agent did, so a rejected mutation that
@@ -266,6 +270,51 @@ func admitAgentCall(w http.ResponseWriter, r *http.Request, next http.Handler, o
 			outcome.registry.ChargeEffect(r.Context(), outcome.spec)
 		}
 	}
+}
+
+// pinAdmittedWrite conditions an auto-executed agent write on the record state
+// its tier was decided from, by forwarding the gate's pin as the request's own
+// If-Match.
+//
+// This is redeemIfPresented's forward (agentgatestaging.go) one tier down, and
+// for the same reason: the gate resolved a dynamic tier by READING the record,
+// that read commits before the handler's transaction opens, and the agent
+// controls both sides of the window — its own 🟢 call can close a deal between
+// the two. Moving the compare inside the transaction that mutates is what makes
+// a record that changed lose to the version check instead of to timing.
+//
+// A caller that sent its own If-Match keeps it — but only if it names the
+// version the gate read, and that is CHECKED. The caller controls the header,
+// so a version the gate never saw is a version nothing proved: a caller naming
+// the version the racing close will PRODUCE walks straight through, because the
+// store's compare then passes on precisely the record the tier decision does not
+// describe. Preferring the caller unchecked would turn a coin-toss race into an
+// armable one. A disagreement is answered as skew, which is also what a caller
+// holding a genuinely stale version already gets, one layer down.
+//
+// It reports whether the request may proceed; a refusal has already been
+// written.
+func pinAdmittedWrite(w http.ResponseWriter, r *http.Request) bool {
+	version, pinned := auth.AutoExecutePin(r.Context())
+	if !pinned {
+		return true
+	}
+	if caller := r.Header.Get("If-Match"); caller != "" {
+		// Compared as the numbers they are: the contract's If-Match is a bare
+		// integer version, and two spellings of one number must not read as
+		// disagreement. A caller header this parser refuses is left for the
+		// handler's own IfMatchVersion to answer, which is where that message
+		// already lives.
+		if got, err := strconv.ParseInt(caller, 10, 64); err != nil || got == version {
+			return true
+		}
+		httperr.Write(w, r, fmt.Errorf(
+			"If-Match %s is not the version this record was read at (%d) — re-read it and retry: %w",
+			caller, version, apperrors.ErrVersionSkew))
+		return false
+	}
+	r.Header.Set("If-Match", strconv.FormatInt(version, 10))
+	return true
 }
 
 // effectRecorder answers whether the handler behind this door actually
--- a/backend/internal/compose/agentgatepin_test.go
+++ b/backend/internal/compose/agentgatepin_test.go
@@ -17,6 +17,7 @@ package compose
 // middleware produces rather than what a hand-set context claims.
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"net/http"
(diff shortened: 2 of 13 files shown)


tinnhlt-gra:
Addressed in `5ec70843`: `TestAWriteWithNoAdmittedPinIsForwardedUnconditioned` now resolves `create_record`'s own spec through the policy table, fails loudly if that spec is dynamic, and admits through the real `auth.Gate.Admit` — so the absent `If-Match` is the static tier's doing rather than a context no gate produced. The load-bearing static-tier assertion (an `ObservedVersion` present in the resolver input and still no pin) lives in `admitpin_test.go`'s "a static tier read no record to decide itself" case; this one is its end-to-end companion on the REST door.

An independent review raised the same weakness, which is why it was worth fixing rather than answering.

PR #819, opened 2026-08-10
fix(finance): the demo ledger keeps a real timeliness sample, and the suite reads it at the ledger's own clock (#786, #798)

### LINKED ISSUE #786: Flaky: the finance card test draws a random ledger archetype every run
`TestAfterASyncTheCardHasFiguresToShow` (`internal/modules/finance/sync_integration_test.go:252`) fails intermittently in the integration lane:

```
sync_integration_test.go:252: no payment-behaviour median after a sync over eighteen months
--- FAIL: TestAfterASyncTheCardHasFiguresToShow (0.06s)
```

It passes on a re-run, and passes when the package runs alone, so it reads as noise — it is not.

## Why it flakes

The offline generator is deterministic **per workspace**, and the workspace is the part the test randomizes.

- `offline.go:111` seeds the ledger RNG with `seedFor(p.workspace, externalCustomerID)`.
- `offline.go:112` then picks the customer's archetype straight off that seed: `kind := archetypes[rng.IntN(len(archetypes))]`.
- The test mints a **fresh** workspace every run — `sync_integration_test.go:59`, `ws: ids.NewV7()` — while the customer id is fixed (`"ACME-01"`).

So every run draws a different archetype and a different ledger. The archetypes differ in `openTail` and `disputeRate`, and FIN-FORM-3 refuses a median below five settled invoices inside the 180-day window. On the seeds that draw a slow/disputed archetype, the sample falls under the floor, `MedianDaysAfterDue` is legitimately nil, and the assertion fires.

The generator's own comment (`offline.go:106-110`) states the property the test depends on — *"must produce the same ledger for the same customer on every machine and every run"* — and it is true, but only for a **stable** workspace id. Production and the demo seed satisfy that; this test does not, and nothing says so.

The archetype comment at `offline.go:75-85` already anticipates the failure mode: a small tail "would leave a generated customer demonstrating the refusal rather than the figure".

## Fix shape

Pin the workspace the generator is seeded from, so the test asserts against one known ledger rather than a lottery — either a fixed UUID in `financeEnv`, or seed the provider from a constant rather than `e.ws`. Then the assertion is about the formula rather than about which archetype came up.

Worth checking the sibling assertions in the same file for the same dependency while in there; they assert weaker properties (`> 0`) and so survive the draw, which is why only this one shows.

## Not caused by the change that found it

Found while running the full lane on an unrelated branch (the MCP-SESS-READS charging fix). The finance package passes both clean at `20286c3f` and with that branch applied; the failure only appears in a full-lane run, which is the seed lottery rather than any interaction.

### LINKED ISSUE #798: finance: the offline ledger's fixed epoch drifts out of the 180-day timeliness window — intermittent now, permanent from 2026-09-01
`TestAfterASyncTheCardHasFiguresToShow` (`internal/modules/finance`) is failing
intermittently now and will fail **unconditionally from 2026-09-01**. Found while
investigating an unrelated PR's red integration shard; the assertion that trips is

```
sync_integration_test.go: no payment-behaviour median after a sync over eighteen months
```

which is `after.MedianDaysAfterDue == nil`, i.e. `TimelinessOver` reported
`InsufficientSample`.

## Why

Two facts that are fine apart and fatal together:

- **The generated ledger is anchored to a fixed date.** `offlineEpoch` is a
  constant, `2026-08-01`, and every invoice is issued at `epoch - N months` for
  N = 0..17. The ledger does not move.
- **The timeliness window is measured from `now`.** `TimelinessOver` keeps
  invoices whose `FullyPaidAt` falls in `[now-180d, now]`, and FIN-FORM-3 refuses
  a median below `MinTimelinessSample` (5).

So the window slides forward with wall-clock time while the ledger stays put, and
the number of settled invoices inside it shrinks month by month:

| date | window starts | generated invoices in window | minus the open tail (1) | vs floor (5) |
|---|---|---|---|---|
| 2026-08-10 | 2026-02-11 | 6 | **5** | exactly at it |
| 2026-09-01 | 2026-03-05 | 5 | **4** | below |
| 2026-10-01 | 2026-04-04 | 4 | **3** | below |
| 2026-11-01 | 2026-05-05 | 3 | **2** | below |

(Membership is by payment date rather than issue date, so the real boundary is a
few days off these, but the direction and the September crossing are not.)

**Today the margin is exactly zero**, which is why the failure is intermittent
rather than constant. `openTail: 1` leaves precisely five settled invoices in the
window — the floor — and the archetypes carry `disputeRate` 30–40 per thousand.
A single dispute inside the window costs one settled invoice and drops the sample
to four. The generator's own comment names this exact arithmetic:

> Monthly invoices mean the 180-day window holds six, and FIN-FORM-3 refuses a
> median below five settled ones … A dispute inside the window costs another

What makes it a coin flip per run rather than deterministic: the test seeds a
**fresh workspace uuid** on every run (`ws: ids.NewV7()`), and `seedFor` hashes
that uuid into the PCG seed — so the archetype and every dispute roll differ each
run. The generator is reproducible for a given workspace, which is what it was
designed for; the test just never uses the same one twice.

## What it is not

Not load or timing — the assertion is a plain nil check on a computed figure, and
it passes 3/3 locally today. Not a recent regression in the finance code.

## Suggested fix, roughly in order of preference

1. **Anchor the generator to `now` rather than a constant**, so the ledger keeps
   its intended relationship to the window. That is what the comment "eighteen
   months puts well over the five settled invoices the timeliness floor needs"
   assumes and no longer describes.
2. If the fixed epoch is deliberate (reproducible demo data across machines),
   give the test an injected clock pinned near the epoch, so it measures the
   formula rather than the calendar.
3. Either way, **restore a real margin**: at zero, any dispute fails the run. A
   fortnightly cadence, or a floor the generator clears by more than one invoice,
   removes the coin flip.

Worth doing before September; after that it is a permanent red on every PR that
touches the backend.

tinnhlt-gra:
Closes #798. Closes #786.

`TestAfterASyncTheCardHasFiguresToShow` failed intermittently and was on course
to fail unconditionally. Three separate causes, and all three had to go for the
test to be about the formulas again rather than about the calendar and the draw.

## The window slid off the ledger (#798)

The generator is anchored to a fixed epoch on purpose: a ledger keyed on today
produces different dates every morning, so every sync would rewrite every
mirrored row — the one thing the sync exists not to do, and the claim
`TestASecondSyncOverAnUnchangedSourceWritesNothing` is there to hold.

But the figures are folded over windows measured back from `now`. The ledger
stays put and the window walks away from it. Measured with the real
`TimelinessOver` over 20,000 seeds:

| as of | below FIN-PARAM-3's floor |
|---|---|
| 2026-08-01 (the epoch) | 0% |
| 2026-09-01 | 0.03% |
| 2026-10-01 | 62% |
| 2026-12-01 | 100% |

(The issue estimated the crossing as September from issue dates; membership is
by *payment* date, so it is really October. The direction and the outcome were
right.)

So **the suite reads FIGURES at a clock pinned to the epoch** — the `WithClock`
seam the store already offers, and the idiom
`TestCrossingADueDateDoesNotRewriteTheLedger` already uses in this file. The
**STATE** is deliberately still read at the real clock: staleness is a genuine
comparison against a `last_success_at` the database stamped, and pinning one
side of it would make that assertion pass for the wrong reason.

`TestASyncRepairsAnInvoiceThatIsMissingItsRate` gets the same treatment. It
would not have gone red — it would have gone *vacuous*, since an empty window
produces a zero total rather than the FIN-AC-6 refusal it is asserting the
absence of.

Anchoring the generator to `now`, which the issue suggested first, is the one
fix that is not available: it trades this failure for a mirror that rewrites
itself daily.

## The sample had no margin

Monthly billing puts barely six settled invoices inside FIN-PARAM-2's 180-day
window against FIN-PARAM-3's floor of five. The open tail took one and a dispute
took another, so **1.5% of generated customers demonstrated FIN-FORM-3's refusal
rather than the figure** — even at the epoch, with no calendar drift at all. The
generator's own comment claimed "well over the five settled invoices the
timeliness floor needs", which had not been true for some time.

Fortnightly billing over the same eighteen months fixes it. The **cadence**, not
the depth, decides how many settled invoices land inside a window measured in
days. The thinnest ledger the generator now produces carries eight, and the
archetypes can afford open tails that differ — which is one of the readings the
card exists to tell apart, and was previously flattened to 1 for all three by
the same arithmetic.

## The test drew a different ledger every run (#786)

The generator's seed is `sha256(workspace | customer)`, and the test minted a
fresh workspace on every run — so the archetype, the disputes and the amounts
were a lottery. The rows still live in a per-run workspace so parallel runs do
not share a tenant; only the generator's seed is pinned, and nothing but the
seed reads that string.

## The guards

Both were shown to fail against the defect before they were allowed to pass.

- **The floor guard** now reads the real FIN-FORM-3 fold. The spelling it
  replaces counted every payment after the window opened — no upper bound, no
  dispute exclusion, no floor — so it passed while production answered
  insufficient-sample over the very same ledger. Reverted to monthly billing it
  reports `sample of 5 clears the floor by 0, want a margin of 3`.
- **The epoch guard** now sweeps the seed space rather than checking one fixed
  ledger, and that is how it found a real defect nobody had filed: the credit
  note is dated *forward* from the invoice it reduces, so on **6% of seeds it
  was issued after the epoch** — a generated row that moved past the fixed point
  the whole ledger is measured back from. The note now targets an invoice old
  enough to land on or before it. Reverted, the guard names the row and the date.

Both sweep 64 workspace/customer pairs rather than four fixed names, because a
claim about "every generated customer" is a claim about the seed space.

## Verification

- `make check-backend` green.
- `make test-it DIR=backend/internal/modules/finance` green, unit + integration.
- **The drift claim, proved rather than argued:** with `offlineEpoch` moved back
  a year — i.e. a wall clock a year past it — the suite passes as written, and
  fails with `net invoiced = 0, want a positive figure` the moment the figures
  are read at the wall clock again.

No frontend file changed.

## Left open

The demo ledger itself still ages: an installation running well after the epoch
shows a card whose windows have moved off the generated data. That is a product
question about demo-data freshness rather than a test defect, and it is filed
separately.

---
## Summary by cubic
Stabilizes the finance tests and demo ledger by reading figures at the ledger’s epoch and adjusting the generator so every seed keeps a valid timeliness sample. Fixes intermittent and time-based failures in the sync suite.

- **Bug Fixes**
  - Read card figures at the ledger’s epoch using store `WithClock`; keep state at the real clock to test staleness correctly.
  - Pin the generator seed in integration tests (`ledgerSeed`) so archetypes don’t change per run; rows still live in per-run workspaces.
  - Ensure credit notes target sufficiently old invoices so their issue date never lands after `offlineEpoch`.

- **Refactors**
  - Switch the demo ledger to 18 months of fortnightly billing (39 invoices, 14-day cadence) and raise `openTail` for ordinary/slow archetypes to keep a real sample inside the 180-day window with a 3-invoice margin above the floor.
  - Add seed-sweeping guards that use the real timeliness fold to enforce the sample margin and verify no generated row is dated after `offlineEpoch`.

<sup>Written for commit 46648a2017795e899a33f53740e7573eabb94aa2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/819?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/internal/modules/finance/offline.go
+++ b/backend/internal/modules/finance/offline.go
@@ -37,6 +37,12 @@ const OfflineProviderName = "offline_demo"
 // NOT time.Now(). A generator keyed on today produces a different ledger every
 // day for a customer nobody touched, so every sync would rewrite every row —
 // and the sync's whole job is to notice what actually changed.
+//
+// The cost of that choice is that the ledger ages: the figures are computed
+// over windows measured back from NOW, so as the calendar leaves the epoch
+// behind, fewer and fewer generated invoices fall inside them. Anything that
+// asserts a FIGURE over this ledger therefore reads it at a clock pinned to
+// this epoch, or it is measuring how long ago the epoch was.
 var offlineEpoch = time.Date(2026, time.August, 1, 0, 0, 0, 0, time.UTC)
 
 // OfflineProvider generates one workspace's demonstration ledger.
@@ -74,29 +80,24 @@ type archetype struct {
 	// disputeRate is how often an invoice is disputed, per thousand.
 	disputeRate int
 	// openTail is how many of the most recent invoices are still unpaid — the
-	// open balance a real account always carries.
-	//
-	// Bounded by what the timeliness window needs. Monthly invoices mean the
-	// 180-day window holds six, and FIN-FORM-3 refuses a median below five
-	// settled ones — so a tail of three would leave a generated customer
-	// demonstrating the refusal rather than the figure, which is the opposite
-	// of what a demo ledger is for. A dispute inside the window costs another,
-	// which is why the slow archetype's tail is the smallest.
+	// open balance a real account always carries, and the reading the card
+	// leads with. It grows with the archetype because a slower payer sitting on
+	// more open money is one of the differences the card exists to show.
 	openTail int
 }
 
 var archetypes = []archetype{
 	{name: "prompt", lateLow: -6, lateHigh: 2, disputeRate: 0, openTail: 1},
-	{name: "ordinary", lateLow: -2, lateHigh: 12, disputeRate: 30, openTail: 1},
-	{name: "slow", lateLow: 5, lateHigh: 38, disputeRate: 40, openTail: 1},
+	{name: "ordinary", lateLow: -2, lateHigh: 12, disputeRate: 30, openTail: 2},
+	{name: "slow", lateLow: 5, lateHigh: 38, disputeRate: 40, openTail: 3},
 }
 
 // InvoicesFor generates one customer's ledger.
 //
-// Eighteen months of monthly invoices, with at least five settled inside the
-// timeliness window — which is the sample floor FIN-FORM-3 refuses below, so
-// a generated customer that fell short would demonstrate the refusal rather
-// than the figure.
+// Eighteen months of fortnightly invoices, which leaves about a dozen settled
+// inside the timeliness window — comfortably clear of the five FIN-FORM-3
+// refuses below, so a generated customer demonstrates the figure rather than
+// the refusal.
 func (p *OfflineProvider) InvoicesFor(
 	_ context.Context, externalCustomerID string,
 ) (SourceLedger, error) {
@@ -115,12 +116,12 @@ func (p *OfflineProvider) InvoicesFor(
 	var out SourceLedger
 	// Oldest first, so the settled ones the timeliness window reads sit at the
 	// front and the open tail is genuinely the recent end.
-	for month := offlineMonths - 1; month >= 0; month-- {
-		issued := offlineEpoch.AddDate(0, -month, 0)
+	for period := offlineInvoices - 1; period >= 0; period-- {
+		issued := offlineEpoch.AddDate(0, 0, -period*offlineCadenceDays)
 		due := issued.AddDate(0, 0, paymentTermDays)
 		invoice := SourceInvoice{
-			ExternalID: fmt.Sprintf("%s-INV-%04d", externalCustomerID, offlineMonths-month),
-			Number:     fmt.Sprintf("INV-%s-%03d", issued.Format("2006"), offlineMonths-month),
+			ExternalID: fmt.Sprintf("%s-INV-%04d", externalCustomerID, offlineInvoices-period),
+			Number:     fmt.Sprintf("INV-%s-%03d", issued.Format("2006"), offlineInvoices-period),
 			IssuedOn:   issued,
 			DueOn:      &due,
 			Currency:   currency,
@@ -130,7 +131,7 @@ func (p *OfflineProvider) InvoicesFor(
 		invoice.TaxMinor = net * vatPercent / 100
 		invoice.GrossMinor = invoice.NetMinor + invoice.TaxMinor
 
-		if month < kind.openTail {
+		if period < kind.openTail {
 			// The recent end is still open. This is what gives every account a
 			// non-zero open balance, which is the reading the card leads with.
 			out.Invoices = append(out.Invoices, invoice)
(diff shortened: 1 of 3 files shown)


tinnhlt-gra:
The demo-data freshness half is filed as #820 — why a fixed epoch is load-bearing, why anchoring to `now` or to a coarse boundary both reintroduce the defect they were meant to fix, and the stored-anchor shape that does work.

tinnhlt-gra:
@coderabbitai review

tinnhlt-gra:
@coderabbitai review

PR #865, opened 2026-08-11
feat(frontend): the app shell loses its top bar and gains a real page heading

luit-nfq:
## What this changes

The chrome after login carried a full-width top bar (search field, system-of-record chip, account avatar) above a sidebar that floated as a rounded card inset from the viewport. The bar spent the layout's scarcest space on things that belong to the session rather than the screen, and the page it framed had **no heading at all** — the only page-level name was a `<span>` in that bar.

**Sidebar** — now the left edge of the application: full height, square, a hairline against the content instead of a gap. It carries what is true for the whole session:

- **Search** is the first row under the brand. Still a button, never inline typing (AC-shell-7, one search affordance); the ⌘K cap sits at its right, and its hairline is an inset shadow so the glyph stays in the column of icons below it.
- **The account block** sits at the foot and prints who is signed in (avatar, name, address, chevron), with the menu opening upward. One Settings row became two — **Account settings** (`#/settings/account`) and **Settings** (`#/settings`) — with language, theme and sign-out unchanged. Expanded, the trigger's accessible name carries the visible name (WCAG 2.5.3); collapsed, the identity is described for a screen reader rather than left to a tooltip.
- **The Margince agent panel** left the sidebar for the top right of the content column, as a strip that reads in one line. It still claims configuration and never liveness, and its example-data marker never outlives the example values it covers at any width.

**Page head** — not a panel: no border, no shadow, no card. It establishes one rule, which the suites pin:

> **Exactly one `<h1>` per railed page.** On a route that names no record the shell mints it. On a record route the shell yields — printing only the trail that leads there — and the record surface's own name is the page's h1.

Three screens were reconciled to it: `dedupe` and `design` demoted their duplicate page titles, and the lead surface gained `SectionHeader level={1}` (it renders no record header and was the one railed page left with no heading of any level).

Fixed on the way: a screen's own id segment is no longer read as a record, so `#/settings/privacy` stopped rendering an h1 reading "privacy"; the phone sheet spells its group names out instead of inheriting the collapsed rail's hairlines; the sidebar's rows take the design system's focus ring.

## Verification

- `make check-fe` — 182 test files, 2240 tests, green; biome, tsc, build, and the DS purity / spacing / font / icon / native-control gates all pass.
- `pnpm playwright test e2e/ac.spec.ts e2e/company-record.spec.ts` — 64 passed, 14 skipped, including a new `AC-shell-1k` that pins the one-h1 rule across a list, a record and a settings tab.
- Walked in the browser at desktop, 620px and collapsed, in both themes: the account menu, the phone sheet, the palette from the sidebar's search row, and a record page's crumb-plus-name.

## Draft, on purpose

This is the first pass of a UI/UX iteration; more feedback lands on this branch before it is ready for review. One design call is still open — screens that centre a narrow reading column (Home, Reports) leave the heading at the page gutter while their content starts well right of it. Noted in STATUS.md.

---
## Summary by cubic
Replaced the top bar with a real page head, moved session controls into the sidebar, and added an agent dock. Unified all railed screens to one content column, improved navigation accessibility, made the phone sheet fully modal, and stabilized person record tests.

- **New Features**
  - Sidebar is full-height on the left; search stays a button under the brand; the account row moved to the foot, shows the signed-in user, and splits “Account settings” and “Settings” with WCAG 2.5.3-compliant labels.
  - Page head with a one-h1 rule. The shell mints h1 on non-record routes; record routes supply it via their surfaces (use `SectionHeader level={1}`).
  - Agent dock at the head’s right with rest/hover/panel states. Shows an approvals badge at rest and uses a shared popover-dismiss hook.
  - All railed and list screens use the shared 1280px “wrap” column; rail-less screens keep the narrow column. Scrollbar lanes are reserved on both sides; safe-area insets are enabled.

- **Bug Fixes**
  - Routing and headings: unknown routes show “Not found”; route id segments no longer read as records (e.g., settings/privacy).
  - Navigation: the active row adds a 3px accent bar and semibold weight with neutral text; focus ring is an inner outline; group headings move to 12px semibold; labels to 14px; collapse runs at 200ms, with reduced-motion keeping fades; ⌘B/Ctrl+B toggles the sidebar.
  - Popovers and sheet: shared dismissal ignores clicks inside, supports Escape/outside-click, and returns focus; phone sheet is fully modal, closes on breakpoint change, and has a proper header and “Close”.
  - Sidebar/account: language list opens inside the account menu; the sidebar foot rule spans the panel; spacing on the account block is corrected.
  - Layout and safe areas: removed contain:layout to avoid clipping fixed-position overlays; phone list gutter is restated; horizontal and bottom safe-area insets apply.
  - Tests: person record page mocks 360/brief/consent reads; in overlay mode, omitted sections are named instead of answered empty.

<sup>Written for commit 72a201fbed3657531b3dc69845920b529103ef4b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/865?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Introduced a persistent workspace sidebar with integrated search, navigation, account controls, and AI status access.
  * Added account menus with language, theme, settings, sign-out, keyboard, and responsive interaction support.
  * Added an AI status panel showing approvals, tools, activity, routing, and spending information.
  * Added responsive mobile navigation with safe-area support.

* **Improvements**
  * Standardized page layouts and heading hierarchy for clearer navigation and record pages.
  * Improved focus management, dismissal behavior, accessibility labels, and localized interface text.


luit-nfq:
All six review findings were real and are fixed — none dismissed.

| Finding | What changed |
|---|---|
| `popover.ts` dismissed on **every** document click, not only outside ones | It ignores clicks inside the panel now. The two places that must both act *and* close (a destination inside the phone sheet, the account menu's links) say so at the call site — a preference row inside a popover has to be able to act without taking the popover with it. |
| `contain: layout` on the content column | Removed. Containment makes that element the containing block for every fixed-position descendant, and the screens render toasts, the ask FAB and their modals inside it — all of which would have anchored to the column and been clipped by its `overflow: hidden`. 200ms of animation is not worth that. |
| An open sheet survived the window widening past 700px | It closes when the breakpoint stops matching. Left open, the page stayed scroll-locked and `inert` with no control on screen to release it. |
| `viewport-fit=cover` also exposes the horizontal insets | `.app` pads for `env(safe-area-inset-left/right)`. The full-height layout is the one that applies in landscape, where the notch sits over the sidebar's own edge. |
| The list surface kept the desktop gutter on a phone | Its rule lives outside the `max-width: 700px` block and outranks the `.wrap` rule inside it, so the phone gutter is restated for it. |
| The page-head test hand-rolled a `<button>` | It passes the design-system `Button` — a test that supplies its own version of production proves nothing about production. |

Two things found while getting the e2e green, both from `main` rather than from this branch:

1. The hermetic mock had **no handler** for `GET /people/{id}/360`, `/brief` or `/consent/guard`, so every spec that opened a person record rendered the error boundary. The 360's `person` is required by the contract and the fall-through answered a list-shaped body without one. The mock serves all three now and names the mirror-unserviceable sections in `sections_omitted` under overlay mode. That repaired `AC-shell-1k` and `PERF-1`, which had been asserting against a crash.
2. `AC-overlay-7` could not be repaired the same way: the person page V2 states a withheld section in its own vocabulary and no longer carries the SoR-specific state the other 360s use. Its person section now asserts what the page actually promises — that it renders, and that no panel degrades into an error box — and the missing distinction is filed as #882 rather than encoded here as if it were intended.

**`live-boot`, `deterministic-gates` and every integration shard fail on this branch for a reason that is not in it: `main` carries two migrations numbered `0209`** (`0209_person_record_page_v2` from #868 and `0209_drop_workspace_identity_columns` from #862), so `migrate` aborts with `pgmigrate: duplicate version 0209`. Main's own run is red on all of those jobs plus its frontend job (a biome formatting drift in `i18n/de.ts` and `vi.ts` that this branch already fixes). Nothing here can go green until the collision is renumbered.

luit-nfq:
The sixth finding (STATUS.md, "remove Home from the narrow reading-column note") no longer applies: that note is gone. The rebase replaced it with the `--pageColumn` paragraph, because the narrow column it described was removed — every railed screen, Home included, now reads one column from that token.

PR #883, opened 2026-08-11
fix(migrations): the repair coverage reads the verb-grant shape

dungbanguyen:
## main is red

`b443a0da` (#875 is on top of it, so `main` has been red since) fails three tests in `backend/migrations`:

```
FAIL: TestTheRepairsCoverEveryGuardedRBACBackfill/core
  0210_consumer_mail_create_grant writes role permissions 1 times but only 0 are
  in the guarded shape this test reads. The unread ones are invisible to the
  coverage check above, which would then report a repair complete while a
  backfill it cannot see goes unrepaired.
```

Reproduced on plain `origin/main` in a clean worktree. `97a80be9` sits between the breaking commit and now — it is docs-only, so its gates skipped and it reported green, which is why `main` looked healthy.

The test is right to fail. Its sibling assertions spell out the stakes: a backfill no repair re-applies leaves every installation that upgraded past it *"403ing on every route, permanently."*

## The migration is not at fault, and is not edited here

An applied core migration is immutable — additive-only. Correcting one in place changes what a **fresh** installation gets while every database that already applied it keeps the old behaviour, and the two diverge silently with nothing to reveal it. That holds without production: dev and CI databases have applied `0210`.

What `0210` does is legitimately a **second operation**:

| | `guardedBackfillPattern` | `0210` |
|---|---|---|
| Writes | `{objects,<object>}` whole | one verb *inside* an existing entry |
| Guarded on | that object's **absence** | that object's **presence** |

Forcing the second into the first would either overwrite the verbs already present or, under the absence guard, match nothing at all. Two shapes, because there are two operations — so this adds a pattern rather than stretching one past its meaning.

## Verb grants are reconciled, deliberately not collected

The coverage comparison keys on `(object, role)` and cannot carry a verb: two migrations granting different verbs on one object would collide on a single key and one would replace the other, unnoticed.

Rather than model a dimension nothing needs yet, `assertNoVerbGrantBelowTheCeiling` refuses the case that would matter:

- **Above the ceiling** — never shipped unbound, needs no repair. That is where the one verb grant that exists sits.
- **At or below it** — the comparison would silently skip a backfill, the exact outcome this file exists to prevent. It fails, and says what to extend.

A deliberate scope boundary, stated in the source rather than left implicit. If the verb dimension should be modelled properly instead, that is a contained follow-up to `rbacGrant` plus the sort and messages.

## Both directions mutation-proven

Neither gate passes vacuously:

- **Check every migration** instead of only those below the ceiling → the new assertion fires on `0210` by name, with the verb, object, roles and ceiling `0192`
- **Blind the verb pattern** → the reconciliation canary catches the write it can no longer read (`1 write, 0 in a guarded shape`)

## Verification

- `go test ./migrations/` — green, uncached
- The three previously failing tests — all pass

---
## Summary by cubic
Teach the RBAC repair coverage to read verb-level grants, fixing three failing migration tests and unblocking `main`. Adds a guarded verb-grant pattern and a check to prevent unread backfills.

- **Bug Fixes**
  - Added a `guardedVerbGrantPattern` to match verb-level updates guarded on object presence.
  - Updated coverage logic to count both guarded shapes and validate object/guard alignment, while not collecting verbs to avoid (object, role) collisions.
  - Introduced `assertNoVerbGrantBelowTheCeiling` to block verb grants at or below the ceiling, ensuring repairs can’t be skipped.

<sup>Written for commit bd295252707ace0674ebf7edb36787f3a5b55fb5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/883?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/migrations/rbacrepair_test.go
+++ b/backend/migrations/rbacrepair_test.go
@@ -50,6 +50,19 @@ var guardedBackfillPattern = regexp.MustCompile(
 		`WHERE \(is_system AND key (?:IN \(([^)]*)\)|= ('\w+'))\s*` +
 		`AND NOT permissions->'objects' \? '(\w+)'\)`)
 
+// The second legitimate shape: grant ONE verb on an object entry the role
+// ALREADY has, guarded on the object's presence. The pattern above cannot
+// express it and must not be stretched to — it seeds `{objects,<object>}` whole
+// and guards on that object's ABSENCE, so forcing this write into it would
+// either overwrite the verbs already there or, with the absence guard, match
+// nothing at all. Two shapes, because there are two operations.
+var guardedVerbGrantPattern = regexp.MustCompile(
+	`UPDATE role SET permissions = jsonb_set\(\s*` +
+		`permissions,\s*'\{objects,(\w+),(\w+)\}',\s*` +
+		`'([^']*)'::jsonb\)\s*` +
+		`WHERE is_system AND key (?:IN \(([^)]*)\)|= ('\w+'))\s*` +
+		`AND permissions->'objects' \? '(\w+)'`)
+
 // Every write to role.permissions, however it is spelled. This has to be
 // STRICTLY BROADER than guardedBackfillPattern, not a prefix of it: a canary
 // sharing the pattern's literal head would miss exactly the deviations that
@@ -96,6 +109,7 @@ func TestTheRepairsCoverEveryGuardedRBACBackfill(t *testing.T) {
 			}
 
 			assertNoUnnamedRepairAboveTheCeiling(t, namespace, newestRepair, backfills)
+			assertNoVerbGrantBelowTheCeiling(t, namespace, newestRepair)
 		})
 	}
 }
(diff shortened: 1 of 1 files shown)


dungbanguyen:
Superseded by #874, which landed the same conclusion and carried it further.

We independently agreed on the important call: `0210` is an applied core migration and must not be edited, so the fix belongs in the derivation, not the SQL. #874 then did the fuller version — `verb` added to `rbacGrant`, sorted and labelled, and verb grants **collected into** the coverage comparison.

This PR deliberately stopped short of that: it reconciled verb grants for the canary but refused them below the repair ceiling rather than modelling them. #874's approach is strictly better — a verb grant below the ceiling is now covered instead of rejected, so there is no case the derivation declines to handle.

Closing as redundant. `origin/main` verified green on `go test ./migrations/`.

PR #895, opened 2026-08-11
fix(migrations): the identity-column drop takes 0211, off a duplicate 0209

dungbanguyen:
**`main` does not build right now.** Two migrations hold version `0209`, so the core namespace fails to load and `make check-backend` is red for everyone:

- `0209_person_record_page_v2` — #868, merged 05:08 UTC
- `0209_drop_workspace_identity_columns` — #862, merged **06:31 UTC**

Mine merged last, into a `main` that already held the number, so mine is the one that moves — to **0211**, since `0210` was taken in between.

## How it got through

Not a numbering accident, a merge-order one. My branch had been rebased before #868 landed and was never brought forward again, and the merge went through on a `mergeStateStatus` of `UNKNOWN` rather than a verified up-to-date branch. Neither PR's checks ever saw the other's file: **each was green alone and broken together.** Worth knowing for the branch-protection settings — `strict: true` did not bite here.

## Why renumbering a merged migration is right this time

The rule against it exists because an applied version never re-runs, so editing one diverges fresh installations from deployed ones. That cannot happen here: this version has only ever applied in CI, because `main` has not been loadable since #862 merged.

The comments citing the version follow it.

## Two pre-existing breaks this does NOT fix

Fixing the duplicate makes the namespace loadable again, which reveals a second failure that was masked behind it — and there is a third on the craft gate. **Neither is mine, and both are on `main` independently of this PR:**

1. `TestTheRepairsCoverEveryGuardedRBACBackfill` — `0210_consumer_mail_create_grant` writes `jsonb_set(permissions, '{objects,capture_settings,create}', …)`, a three-level path the gate's guarded-shape pattern (`'{objects,(\w+)}'`) cannot read. The gate's complaint is legitimate: a write it cannot see is one the repair-coverage check would report complete while a backfill goes unrepaired. Whether the answer is an additive repair or a gate that reads the deeper path is that feature's call, not mine. Filed as an issue.
2. `craft static` — `internal/compose/deepread_integration_test.go` is 1013 lines against the 1000-line test ceiling, from #(company employee_range). Also filed.

So this PR's own CI will be red on (1) and (2). It still strictly reduces the breakage, and nothing here can merge until they are resolved.

---
## Summary by cubic
Renumbered the migration that drops workspace identity columns to 0211 to resolve a duplicate 0209. This unblocks the core migration loader and restores the `main` build.

- **Bug Fixes**
  - Renamed migration files to `0211_drop_workspace_identity_columns.up/down.sql` to avoid collision with `0209_person_record_page_v2`.
  - Updated references from 0209 to 0211 in `deals`, `identity`, and schema tests.
  - No SQL changes, only a version bump to fix the loader failure.

<sup>Written for commit da24f738a9ada323679222c7a57a37537dbe9274. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/895?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


dungbanguyen:
Closing — already fixed on `main`. `0211_drop_workspace_identity_columns` landed there while this branch was open, with the same rename and the same comment updates, so this PR is a no-op.

Two things from the review worth carrying forward, and I'm opening a separate PR for them:

**The ledger cannot see a renumber.** `dbmigrate.Up` writes `(version, name)` but `appliedVersions` reads back `version` only, so a recorded version under a *different* name is indistinguishable from a correct application. On any database that applied the drop as `0209` — this tree is shared with parallel sessions that switch branches — `0209_person_record_page_v2` is now **skipped silently and forever**, and the loud failure that follows (`column w.name does not exist` when `0211` re-runs) masks it. `Down` has the mirror problem: it would run `person_record_page_v2`'s down for a `0209` row that actually recorded the drop, succeed having reverted nothing, and delete the row.

**STATUS.md still cites 0209** for the identity drop, which now names an unrelated migration.

Also worth noting for branch protection: `required_status_checks.strict` is not enforcing up-to-date branches on `main` — that is how two PRs each claimed `0209` while each was green alone.

PR #896, opened 2026-08-11
fix(agents): the scope-filtered catalog refuses to be stored (#644); feat(deals): a forecast that moves without a stage change leaves a trace (#823)

### LINKED ISSUE #644: The scope-filtered tool catalog is served with no Cache-Control, against BYO-WIRE-1
## The pin

Spec **BYO-WIRE-1** states it directly:

> Scope-filtered catalogs cache **`private`** — a shared entry on a scope-filtered
> response is a disclosure that never reaches the server to be audited.

## What is missing

`tools/list` became scope-filtered in #636: the response body now varies by the
presenting passport's scopes. The hosted MCP transport sets `Content-Type`,
`WWW-Authenticate` and `Mcp-Session-Id` — and **no `Cache-Control` at all**.
There is no `Cache-Control` anywhere on the `/mcp` path, at the handler or at
the edge.

So a scope-varying response is served with no cache directive. A shared cache
that stores it can serve one passport's catalog to another, and because the
second request never reaches the server, nothing is audited — which is the
disclosure the pin names.

## Why this reads as a missed sibling rather than an oversight

The repo already knows this shape and applies it correctly elsewhere on
principal-varying responses:

- `internal/modules/identity/handlers_roster.go` — `private, no-store`
- `internal/modules/identity/handlers.go` — `private, no-store`
- `internal/modules/identity/handlers_users.go` — `no-store`

The MCP surface is the one that did not get it, and it acquired the
scope-varying property later (#636) than those handlers were written. That is
the "fixed the case under review, missed the sibling copy" shape from the
engineering rules.

## Mitigating factor, stated so it is not rediscovered as a reason to close

`/mcp` requires an OAuth bearer, and RFC 9111 §3.5 says a shared cache MUST NOT
store a response to a request with an `Authorization` header unless the response
explicitly permits it. A compliant shared cache therefore will not cache this
today. That lowers the practical severity — it does not settle it: the pin is
explicit, the header is one line, and "every intermediary is compliant" is not
an invariant this codebase gets to assume about an operator-hosted deployment
(A2 is hosted by the operator or a hosting partner, never Gradion — ADR-0027).

## Suggested fix

Set `Cache-Control: private, no-store` on the MCP JSON response path, matching
the identity handlers. `no-store` rather than bare `private` because a session's
catalog is re-derived per request against live authority — revocation binds
mid-session, so a cached catalog is a stale authority claim as well as a
cross-principal one.

Worth a fitness test rather than a point fix: any handler whose response varies
by principal carries a private/no-store directive. That derives the obligation
from the system instead of maintaining it as a list.

## Not caused by #643

#643 moves the scope filter into `Registry.Offered` and points `tools/list` at
it; it does not change whether the response is scope-varying. This gap predates
it and is filed separately rather than folded into an in-flight review.

### LINKED ISSUE #823: agents: explain_forecast_change cannot reconcile, because deal history is written on stage change only
Split out of Track G (commercial reads). The other three tools in that track —
`review_commitments`, `render_pipeline_review`, `prepare_handoff` — have their
deterministic substrate and are shipping. This one does not, and the reason is
worth stating precisely because the roadmap had it wrong.

## Not "the snapshots are missing"

The plan assumed historical forecast snapshots were absent. They are not.
`deal_stage_history` (core `0006`) exists and its own comment says what it is
for:

> Append-only stage-change snapshot: "pipeline as of date X" + conversion reports.

It freezes the values a reconstruction needs — `amount_minor_at_change`,
`currency_at_change`, and since core `0058` `win_probability_at_change` — and
every deal gets a birth row (`from_stage_id NULL`).

## What is actually missing

It is written in exactly two places, and both are stage transitions:

- `modules/deals/deal.go` — creation
- `modules/deals/deal_advance.go` — a move

`deal_update.go` writes **no** history row. So two edits leave no trace at all:

- an **amount revised in place**, and
- a **close date slipped** — which is the single most common reason a real
  forecast moves.

A forecast reconstructed from this table therefore reconciles over stage
movement and silently omits the rest, while presenting itself as the whole
answer. That is the failure `modules/finance/formulas.go` already refuses by
hand elsewhere: *"a partial sum wearing the label of a total"*. And it fails the
tool's own exit criterion — the answer must reconcile arithmetically to the
deals that moved — since the deals that moved most invisibly are the slipped
ones.

## Options

1. **Give the history an amount/close-date arm.** A row on any change to the
   fields a forecast reads, not just on a stage move. Most faithful; it is a
   write-path change on the hottest table in the product, so it wants care about
   volume and about what "changed" means for a nullable close date.
2. **Derive from `audit_log` instead.** Every mutation already commits an audit
   row in the same transaction (the write shape). Reconstruction is possible in
   principle and unpleasant in practice — the payload is a diff, not a
   projection, and the query would encode the shape of every past writer.
3. **Ship it scoped, and say so.** Answer only stage-driven movement, and have
   the tool's own answer state that boundary rather than implying completeness.
   Cheap and honest, but it answers a narrower question than the one users ask.

(1) is the real fix; (3) is a legitimate interim **only** if the narrowing is in
the answer text, not just in a comment.

## Not blocking the rest of the track

The other three tools do not read this table. Nothing here needs to land before
they do.

tinnhlt-gra:
Closes #644. Closes #823.

Two unrelated fixes in one branch at the author's request; they share no files and are separate commits.

## #644 — the scope-filtered catalog refuses to be stored

`tools/list` became scope-filtered in #636, so the body varies by the presenting passport, and the transport set `Content-Type`, `WWW-Authenticate` and `Mcp-Session-Id` and **no cache directive at all**. BYO-WIRE-1 states it directly: a scope-filtered catalog caches `private`, because a shared entry on a scope-filtered response is a disclosure that never reaches the server to be audited.

`private, no-store`, matching the three `identity` handlers that already answer this for their own principal-varying reads. `no-store` rather than a bare `private` for a reason those reads do not have: **every call here is re-authorized against live authority**, so revocation binds mid-session and a stored catalog is an outdated *authority* claim as well as another principal's surface.

Set above the framing branch, because it is a property of the answer rather than of how the client asked for it — the event-stream frame carries the same scope-filtered body.

**The test asserts the filtering first and the directive second, and that order is the point.** A test that checked only the header would pass just as well against a transport that had stopped filtering by scope altogether — the very state that makes the header necessary. Verified both ways:

| mutation | what failed |
|---|---|
| remove the header | `Cache-Control ""`, want `private, no-store` |
| give both tools the same required scope | `both passports were served the same 2 tools, so this response does not vary by scope` |

**Mitigating factor, recorded so it is not rediscovered as a reason this was unnecessary:** `/mcp` requires a bearer, and RFC 9111 §3.5 already stops a *compliant* shared cache from storing it. That lowers the severity and does not settle it — A2 is operator-hosted (ADR-0027), and "every intermediary is compliant" is not an invariant this codebase gets to assume about someone else's deployment.

## #823 — a forecast that moves without a stage change leaves a trace

`deal_stage_history` is written on creation and on a move, and nowhere else, so **an amount revised in place and a close date slipped leave no history at all** — and the second is the single most common reason a real forecast moves. Anything reconstructing a forecast from that table reconciles over stage movement and omits the rest while presenting itself as the whole answer.

### The design decision, and the constraint that drove it

A **second table**, not an arm on the existing one. `deal_stage_history`'s own comment is "append-only stage-change snapshot", and **five readers hold it to that**:

| reader | what it assumes |
|---|---|
| `deals/health.go:294` | `max(changed_at)` is "when did this deal last move" — stalled-deal detection |
| `automations_preview.go:181,199` | `count(*)` of rows is a count of stage movements |
| `org360/suggestionreads.go:236`, `viewbaseline.go:199` | same, for the account baseline |
| `deals/closedatesweep.go:213` | same |

A row that is not a transition would make every edited deal look freshly moved and inflate every movement count — silently, in stalled-deal detection, which nobody re-derives once it looks plausible. Two tables keep both meanings **by construction** rather than by every present and future reader remembering to filter.

### The rest of the shape

- **Snapshot, not diff**, mirroring the sibling, so a reconstruction unions the two tables in time order and learns one idiom instead of two.
- **The values are the state the deal is IN after the change** — what "the forecast as of date T" reads off the latest row at or before T. Recorded the other way round, every answer would be one edit stale.
- **No birth row needed**: every deal already gets one in `deal_stage_history` carrying amount and currency.
- **The patch decides whether to write.** It has already compared old against new, so a close date cleared or set for the first time counts as movement exactly as a revision does, and a rename writes nothing.

### Verified against a live database, four mutations

| mutation | what failed |
|---|---|
| never record | `an amount revision wrote 0 forecast rows, want 1` |
| record the pre-change value | `recorded amount = <nil>, want 250000` |
| record on any update | `a rename wrote 1 forecast rows, want 0` |
| write into `deal_stage_history` instead | the suite fails — this is the decision above, pinned |

**Scope**: this is the substrate #823 names. The tool that reads it (`explain_forecast_change`) does not exist anywhere in the tree, so there is no answer text to narrow and nothing here to gate — which is also why the issue's option 3 ("ship it scoped and say so in the answer") was not available.

## Gates

`make check-backend` green, `make craft-static` clean across backend/extensions/fixtures, and the new integration test run against a real Postgres. The migration is numbered `0215` against `origin/main` at push time — re-check before merging if main moves, since a duplicate version is a failure no branch-local gate can see.

---
## Summary by cubic
Sets correct caching for the scope-filtered MCP tools catalog and adds forecast history so deal amount/close-date changes are tracked without a stage move. Fixes #644 and delivers the data needed for #823.

- **Bug Fixes**
  - `tools/list` now sends Cache-Control: "private, no-store" for both JSON and event-stream responses to prevent shared-cache leaks across principals.
  - Test asserts scope-based filtering first, then the header, to catch any regression in filtering.

- **New Features**
  - Added `deal_forecast_history` with a write path that snapshots post-change amount, currency, and expected close date when they change without a stage transition.
  - Keeps `deal_stage_history` semantics intact; no rows on non-forecast edits (e.g., rename).
  - Includes integration test, migration `0215` (RLS + index), and schema fitness FK decision for `deal_forecast_history.deal_id`.

<sup>Written for commit 7162cd019a0e3360326dfc62cf8c1806f5f87cd8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/896?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/internal/compose/integration/dealforecasthistory_integration_test.go
+++ b/backend/internal/compose/integration/dealforecasthistory_integration_test.go
@@ -1,0 +1,123 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+//go:build integration
+
+package integration
+
+// A forecast moves for two reasons that leave no stage transition behind: an
+// amount revised in place, and a close date slipped. deal_stage_history is
+// written on creation and on a move and nowhere else, so a forecast
+// reconstructed from it reconciles over stage movement and silently omits the
+// rest — while presenting itself as the whole answer.
+//
+// The last assertion here is the one that pins a design decision rather than a
+// behaviour: these rows go to their own table BECAUSE five readers hold
+// deal_stage_history to its stated meaning, counting rows as movements and
+// taking max(changed_at) as "when did this deal last move". If a future change
+// folds this write back into that table, stalled-deal detection and every
+// automation preview start lying, and nothing else in the suite would notice.
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/jackc/pgx/v5"
+
+	"github.com/gradionhq/margince/backend/internal/modules/deals"
+	"github.com/gradionhq/margince/backend/internal/platform/database"
+	"github.com/gradionhq/margince/backend/internal/shared/kernel/ids"
+)
+
+func TestAForecastMovedWithoutAStageChangeIsRecorded(t *testing.T) {
+	e := Setup(t)
+	pipeline, open, _ := DealFixture(t, e)
+	admin := e.Admin()
+	deal := ids.From[ids.DealKind](e.SeedDeal(t, "Forecast probe", pipeline, open, &e.Rep1))
+
+	stageRows := func(t *testing.T) int {
+		t.Helper()
+		return countRows(admin, t, e, `SELECT count(*) FROM deal_stage_history WHERE deal_id = $1`, deal.UUID)
+	}
+	forecastRows := func(t *testing.T) int {
+		t.Helper()
+		return countRows(admin, t, e, `SELECT count(*) FROM deal_forecast_history WHERE deal_id = $1`, deal.UUID)
+	}
+
+	stageAtStart := stageRows(t)
+
+	// A field a forecast never reads must leave no trace here, or the table
+	// answers "the forecast moved" for every edit anyone makes to a deal.
+	renamed := "Forecast probe, renamed"
+	if _, err := e.Deals.UpdateDeal(admin, deal, deals.UpdateDealInput{Name: &renamed}); err != nil {
+		t.Fatalf("renaming the deal: %v", err)
+	}
+	if got := forecastRows(t); got != 0 {
+		t.Fatalf("a rename wrote %d forecast rows, want 0 — a name is not a forecast field", got)
+	}
+
+	// amount_minor and currency move together or not at all — a bare number is
+	// not money, and the store refuses one without the other.
+	amount, currency := int64(250_000), "EUR"
+	if _, err := e.Deals.UpdateDeal(admin, deal,
+		deals.UpdateDealInput{AmountMinor: &amount, Currency: &currency}); err != nil {
+		t.Fatalf("revising the amount: %v", err)
+	}
+	if got := forecastRows(t); got != 1 {
+		t.Fatalf("an amount revision wrote %d forecast rows, want 1", got)
+	}
+
+	// The value recorded is the state the deal is IN after the change, which is
+	// what a reconstruction asking "what was the forecast as of date T" reads
+	// off the latest row at or before T. Recorded the other way round, every
+	// answer would be one edit stale.
+	recorded := recordedForecastAmount(admin, t, e, deal.UUID)
+	if recorded == nil || *recorded != amount {
+		t.Fatalf("recorded amount = %v, want %d — the row must carry the amount the deal now has", recorded, amount)
+	}
+
+	// A close date set for the first time is a slip like any other: NULL and a
+	// date are different forecasts, and the move from one to the other is
+	// exactly what a reconstruction needs to see.
+	slipped := time.Now().UTC().AddDate(0, 1, 0)
+	if _, err := e.Deals.UpdateDeal(admin, deal, deals.UpdateDealInput{ExpectedClose: &slipped}); err != nil {
+		t.Fatalf("slipping the close date: %v", err)
+	}
+	if got := forecastRows(t); got != 2 {
+		t.Fatalf("a close-date change wrote %d forecast rows in total, want 2", got)
+	}
+
+	// And the reason this is a separate table at all.
+	if got := stageRows(t); got != stageAtStart {
+		t.Fatalf("deal_stage_history grew from %d to %d rows without a stage change — health.go reads "+
+			"max(changed_at) there as 'when did this deal last move' and the automation previews count "+
+			"those rows as movements, so both now answer wrongly", stageAtStart, got)
+	}
+}
+
+func countRows(ctx context.Context, t *testing.T, e *Env, query string, deal ids.UUID) int {
+	t.Helper()
+	var n int
+	if err := database.WithWorkspaceTx(ctx, e.Pool, func(tx pgx.Tx) error {
+		return tx.QueryRow(ctx, query, deal).Scan(&n)
+	}); err != nil {
+		t.Fatalf("counting rows for %q: %v", query, err)
+	}
+	return n
+}
+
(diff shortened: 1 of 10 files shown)


tinnhlt-gra:
Two shards failed; one was mine and one was not.

**Mine — `TestFK_rowScopedTargetsHaveVisibilityDecision`, fixed in `7162cd01`.** The new `deal_forecast_history.deal_id` FK had no visibility decision, which is the gate working: a reference to a row-scoped record is a read of it, so every such column carries an explicit classification and a new one nobody classified fails rather than passing quietly. Classified server-derived, with the reason rather than the category: `UpdateDeal` appends the row for the deal it has just written, after `auth.EnsureVisible` admitted that deal at the top of the same transaction, and the id arrives on the path rather than in a request body — the same reasoning the sibling `deal_stage_history.deal_id` entry records.

**Not mine — `TestAnEvidenceApplyClearingANameWaitsOnTheNameLock`**, which spent its full 60s `probeBudget` on ~109k probes and reported *"the writer neither reached the lock nor returned, so this run proved nothing"*. That is the second test on that helper to redden a PR today (#874 hit the `people` create-dedupe probe the same way), so it is filed as **#898** with the analysis — including that `probeBudget`'s own comment records this being fought once already, when it moved from a probe count to a duration.

The tell in both cases is duration: ~60s and a six-figure probe count means the budget expired. A genuine lock-ordering failure fails in milliseconds with a different message.

PR #907, opened 2026-08-11
feat(frontend): one card primitive, settings split into cards, and a second sidebar navigation level

luit-nfq:
## What this changes

Three UI/UX threads, all behind the login gate.

### One card, with a stacked header

The card was a component 31 callers used and a `className="card"` 106 others hand-rolled — the component could not express the element it wrapped, its own `style`, a test id, an accessible name, or a header, so 54 files each carried their own version of the same five chrome values.

`Card` now takes them: `as` (`section` by default, plus `div` / `article` / `form` / `li`), `inset`, `title` / `sub` / `actions`, `className` / `style` / `id` / `ariaLabel` / `testId` / `onSubmit`, and `role="status"` for the three advisory banners that must announce themselves. Every card site in the authenticated app is on it. `EmptyState` composes it instead of restating its classes, and the `.rates-head` and five `.list-head` wrappers that existed only because the header had no actions slot are gone.

`SectionHeader` stacks its title over its description at full width, with actions the only thing beside the pair. Side by side on one baseline the two competed for the same line: a description long enough to be worth writing pushed the title around, and at narrow widths one of them lost. The description carries a measure, so a wide card does not stretch it into a line the eye loses its place in.

Documented in `Design System/Atoms` (the `Cards` and `Sections` stories) and in the design-system README, which now names `Card` the one card surface.

### Dense settings pages become several cards

Five surfaces rendered everything they hold in one box. Extensions was the worst: a lead paragraph, the seat ceiling, every composed unit, each unit's routes and jobs, and a role-by-CRUD matrix per permission object, with the CSS growing a fake separator (`.ext-unit + .ext-unit { border-top }`) to stand in for the card borders that were missing.

- **Extensions** — lead card, then a card per unit (version and the unit's own page in the header actions), each object's matrix an inset card inside it.
- **Users & roles** — Invite (the card IS the form) and Members with a roster count.
- **Capture** — the domains that come from the company profile, read-only, apart from the ones an admin curates.
- **Installation** — Organization and Currency, sharing one save: the endpoint takes one sparse patch, and two buttons would promise two writes that do not exist.
- **Voice DNA** — status, derived voice, preferences, writing samples, builds; still ONE card on a first run, because five empty cards is not five facts.
- **Audit log** — filters and entries, with the six filters a grid of labelled cells rather than a page-tall form.

Permissions, requests and copy semantics are unchanged, and each page's non-admin surface is exactly what it was.

### A second sidebar navigation level

The settings sub-navigation floated in the content area — a column of tabs beside the cards it switched, belonging to neither the page nor the chrome. It is now a level of the sidebar, and the sidebar is a stack of levels rather than one rail with a special case for settings.

- `app/subnav.ts` is the contract, and it is data: an entry may publish `children`, a screen publishes a section, and `navTrail` derives the levels a route reaches from the segments the router already parses. A third level costs the `children` array that describes it.
- `app/navlevel.tsx` renders one level whatever its depth, so the ten destinations and a section's entries are the same rows — badges, tooltips, the active rule and the phone bar have one spelling.
- Visibility stays where the grants are: settings publishes a finished section through `useSettingsSection`, and the shell never learns what a grant is.
- On a settings route the rail shows that level in place of the destinations (64px cannot carry two, and 252px carrying both reads as twenty places to go), with the way back up above it, naming where it leads rather than saying "back". At phone width the bar becomes that control plus More, which opens the sheet at the level the panel is showing.
- The page head names the ENTRY, not the section: the level already says "Settings", and the page said it a second time above a heading that said it a third.
- Settings keeps its two existing groups (Your settings / Organization). Splitting them into separate menus would be a guess, and those pages are about to change heavily.

Stories: `App/Shell` gains the level expanded, the level collapsed, a third level, and the phone bar with its sheet.

### Smaller

- The account menu's first row reads **Account** rather than "Account settings".
- `check-space-tokens.sh` now derives its obligation for **every** custom property, not just `--space-*`. A fallback-less `var(--x)` that nothing declares resolves to nothing and the declaration is dropped, silently — it found seventeen such uses (`--border`, `--bg`, `--text`, `--radius-2`, `--radius-md`, `--fs-caption`, `--surfaceRaised`, `--focusRing`, the `--color-*` family, `--warning`, `--bgSubtle`, `--shadow-sm`), each a border, fill or size that had been rendering as nothing. `--ob-live-bleed` read an undefined property and no rule consumed it, so it is deleted.

## Verification

`make check-fe` green: 185 test files, 2291 tests, and all seven frontend gates (DS purity, font lock, icon glyph, spacing, custom properties, native controls, contract types) pass. Walked the settings surfaces in the running app at desktop width, expanded and collapsed, including the walk back up to the destinations.

## Not in this PR

Agent-dock work and the remaining layout feedback — following in the same branch.

---
## Summary by cubic
Unifies the app on a single `Card` with a stacked header, splits dense settings into focused cards, and adds a second sidebar navigation level with a route-aware Back. Also polishes phone/rail layout (consistent page gutter, no reserved scroller gutter), keeps search visible at all widths, and moves language/theme to Settings → Account.

- New Features
  - One card, everywhere: `Card` now supports `as`, `inset`, `title`/`sub`/`actions`, `className`/`style`/`id`/`testId`/`ariaLabel`, and `onSubmit`; `EmptyState` composes it. `SectionHeader` stacks title over description.
  - Settings split into smaller cards: Extensions per-unit; Users split into Invite and Members; Capture shows company vs curated domains; Installation has Organization and Currency (one save); Voice DNA split by subject but stays one card on first run; Audit Log becomes Filters + Entries.
  - Second sidebar level: data-driven in `app/subnav.ts`; `app/navlevel.tsx` renders any depth; Back navigates and remembers the parent (fallback Home). Rail adds a Settings row (hidden on phone and while inside a section). Search is a row at all widths. At phone width the bar always shows the four destinations and section entries move under the page head via a switcher. Level header and phone sheet refined.

- Refactors
  - Replaced hand-rolled `.card` and ad‑hoc headers with `Card` + `actions`; advisory banners use `Card inset` with `role="status"`.
  - Tightened CSS token gate: `scripts/check-space-tokens.sh` now fails on any fallback‑less `var(--x)` not defined; fixed undefined tokens and normalized to `--borderSubtle`, `--bgElevated`, `--textMeta`, etc.
  - Account menu simplified (Account, Settings, Sign out); preferences moved to Settings → Account; removed `LocaleMenu`; added `setTheme` and exported `THEMES`; phone sheet rows match the nav geometry.
  - Layout polish: unified phone page gutter for head/content/bar, centered the page head row at phone width, and removed `scrollbar-gutter` reservation to avoid permanent side lanes.

<sup>Written for commit 5cabbaaef87cd1ddb663208b248442b6ed3e0e18. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/907?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


luit-nfq:
Superseded by #997: renaming the branch closed this PR. Same work, rebased onto `main`.

PR #961, opened 2026-08-12
refactor(compose): the voice half of reply drafting is its own file

### LINKED ISSUE #953: main is red: replydraft.go crossed the 500-LOC cap in #952, blocking every PR
`deterministic-gates` is red on `main` at `66fdf6ed`:

```
FAIL: backend/internal/compose/replydraft.go is 539 lines (> 500) — split into
one-concept-per-file packages
```

The file crossed the cap in #952 (*"the Dossier field is fed, so a first touch can say
what the company IS"*): 448 lines before it, 539 after.

Because the cap is a required check, this blocks **every** PR, not only ones touching
that file — a PR is tested merged with `main`, so it inherits the failure. It is
already holding up #913.

## Why this is filed rather than fixed in passing

The gate names the remedy — split into one-concept-per-file — and that is a design
decision about a feature that landed minutes ago, in a subsystem whose author knows
what separates. The alternative, adding a line to `scripts/go-file-length-waivers.txt`,
is specifically what the ratchet exists to prevent for NEW growth: the waiver list is
for pre-existing offenders that may shrink and never grow. Quietly waiving it would
retire a gate the repo armed on purpose, which is not a call to make from an unrelated
PR.

## Suggested shape

`replydraft.go` currently holds the reply drafter, its voice loading, the deterministic
floor plumbing and the provenance assembly. The provenance/voice half looks separable
from the drafting half, but the author is better placed to say — and per the gate's own
wording the split should follow a concept, not a line count.

Note also that `main` has been red four times today in ways that block the integration
lane (#924 comms fixture, the `mcp-info` drift fixed by #946, the #548 contention flake,
and this). Each was found by a PR trying to merge rather than by the push that caused
it, which is worth a thought about whether the required set runs early enough on `main`
itself.

dungbanguyen:
`deterministic-gates` is **red on main** at `66fdf6ed`:

```
FAIL: backend/internal/compose/replydraft.go is 539 lines (> 500) — split into
one-concept-per-file packages
```

The file crossed the cap in #952 (448 → 539). Because that check is required, it blocks **every** PR, not only ones touching this file — a PR is tested merged with `main`, so it inherits the failure. Closes #953.

## Split, not waived

`scripts/go-file-length-waivers.txt` is for offenders that predate the cap and may only shrink. Adding new growth to it would retire the gate for the next file too, so the fix is the one the gate names.

## The seam was already there

`replydraft.go` is the drafting **call**: the prompt, the schema, the retry, the validation, the bounds. The voice layer is a profile laid **over** that call — load the actor's active profile, run the draft under it, check the result against the profile's avoid rules, record the outcome as a voice signal. That layer moves to `replydraftvoice.go`.

Result: **410 and 153 lines**, both well under the cap.

`voiceBlockFor` stays in `replydraft.go` deliberately — both system variants take it, so it belongs with the request builder that reads it rather than with the profile that fills it.

## Verification

- `make check-backend` — green. Worth stating explicitly: `new-from-merge-base` treats moved lines as new, so all 126 are re-gated by the full strict set and **pass unchanged**. No reshaping was needed to get them through, which is what a pure move should look like.
- `make go-file-length` — green, 0 waivers.
- `go test ./internal/compose/` — green.
- `make test-it DIR=backend/internal/compose` — green, 61.1s (this is the suite that drives the drafter end to end).

Found while trying to land #913, which cannot merge past a red required check and is unrelated to this file.

---
## Summary by cubic
Split the voice layer of reply drafting into its own file to bring `backend/internal/compose` back under the 500-LOC gate and restore green deterministic gates on main. No behavior changes; adds a missing warning log on violation-retry failure.

- **Refactors**
  - Moved voice profile loading, violation checks, retry/sanitize, and signal recording from `replydraft.go` to `replydraftvoice.go`.
  - Kept `voiceBlockFor` in `replydraft.go`; moved `replyDraftVoiceSystem` back beside `replyDraftSystem`.
  - Added a warning log when the violation retry inside `completeVoiced` fails.
  - File sizes: `replydraft.go` 419 lines; `replydraftvoice.go` 153 (both under cap).
  - All gates/tests green.

<sup>Written for commit 9a2ae86cdd864f0ca9b55a2504d06e2b09599485. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/961?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


dungbanguyen:
Superseded — `main` already split this file, and further than I did: `replydraft.go` is down to **222 lines** with both `replydraftvoice.go` and `replydraftmodel.go` alongside it. `deterministic-gates` no longer fails on the 500-LOC cap.

Closing rather than rebasing: the conflict is not incidental, it is that the same seam was cut twice. Whoever did it chose the same filename (`replydraftvoice.go`) and the same first boundary, then went on to separate the model call as well — which is the better split.

Nothing here is worth salvaging into that version except possibly two small things the review round on this PR surfaced, both about code that still exists on main:

1. **The violation-retry error inside `completeVoiced` is swallowed.** `retried, retryErr := d.complete(...); if retryErr == nil { draft = retried }` — `retryErr` reaches neither the log nor the caller, and the critic retry bypasses `completeChecked`, so `draftRetryLog` never sees it either. A draft that kept its violations then looks exactly like one that had none. T2.
2. **`completeVoiced`'s failure contract is easy to misread.** It returns its error at three of four exits, and the degrade belongs to `DraftEmailWithProvenance`, which answers with the DETERMINISTIC draft — not by retrying without the profile. I wrote a file comment claiming the opposite and both reviewers caught it; worth a sentence somewhere on main so the next reader does not conclude that error return is vestigial and delete the caller's degrade.

Filing those two as a follow-up rather than reopening this.

PR #1008, opened 2026-08-12
feat(company): the record reads as one page, with a daily brief above its tabs

joshbruh-lang:
The company record page reads as one page rather than a stack of separate cards. A full-width band, between the header and the tab bodies, carries the account's standing readings, the daily brief (what changed, what needs a person, what's due) and the tab bar itself — it stays on screen across every tab, since it describes the account rather than one view of it. The rail is one panel: a details grid over collapsible sections (health, people, tags), rather than a column of separate cards repeating what the grid already names. Every tab body now sits on the same shape — Panel, with full-bleed rows and a titled header — instead of the mix of naked sections and ad-hoc cards the page carried before. `Panel`, `FieldGrid` and `AvatarStack` are new in the design system for this.

**One editable control per field, one filing verb per capability.** Two controls for one field can disagree about what they last wrote, and a second "Add tag" strip meant the same account-level write had two homes. Lifecycle is set from the header, beside Owner, in the block a reader already uses to change things about the account; the rail's Details grid shows the value read-only, the way it already shows Owner. Tags and list memberships are added from the rail's Tags & lists section, next to the values they change — the overview column no longer carries a second "Add tag" / "Add to list" strip for the same write.

**The account's leading deal is pinned by its own refusals.** The last-offer reading picks one deal to represent the account's commercial standing — the biggest priced deal, tie-broken by id — and refuses to pick at all across a mixed-currency set or a truncated deals page, since either would present a wrong deal as a fact about the account. That selection and its refusals were unasserted; now: the single-currency, fully-priced, untruncated case names the biggest deal; an unpriced deal never outranks one that carries a real amount; a mixed-currency set names no leading deal; a truncated page names no leading deal; and the all-unpriced case, checked while writing these, turned out not to be a refusal at all — LastOffer names the deal, not its price, so a set with no priced deal to prefer still names one, deterministically.

**On size:** this PR is larger than it should be. The rail, the band, the brief and the tab bodies were rebuilt against each other and against three new design-system components, so there is no earlier point in this branch where the page both works and is smaller than this. Further work on this page ships as separate PRs from here.


joshbruh-lang:

PR #1011, opened 2026-08-12
fix(company): a section still loading is not one that failed

joshbruh-lang:
The composite 360 read hands every section the same undefined `view` for two different reasons: the read is still in flight, or the read failed. Only the second is a fact about the account, and `sectionState` could not tell them apart — every view-derived card on a record page (Health, People, the commercial picture, recent activity, tags, deals) read UNAVAILABLE, "some of this page could not be loaded", for as long as its own read was still running.

`sectionState` now takes the composite read's own pending flag and resolves an undefined `view` to LOADING while it is in flight, UNAVAILABLE once it has actually failed. The pending flag is threaded from `CompanyPage` down to every card that reads the composite view directly with no skeleton guard of its own: `PeopleCard`, `CommercialPanel`, `RecentActivityPanel`, `DealsCard`, `TagsCard`.

Two call sites in this repo are outside this change: the rail's own `HealthSection`/`PeopleSection` (companyrail.tsx) and its `TagsSection` (companyrailtags.tsx) still call `sectionState` without a `loading` argument, so they keep their present (unfixed) behavior. Both files sit under the rail's own ownership boundary — the parameter is optional and defaults to `false` precisely so those call sites keep compiling and behaving exactly as before rather than silently changing under a file I don't own. They need the same one-line fix once `CompanyRail` accepts a `loading` prop of its own.

Test: pending renders a skeleton, not the "could not be loaded" copy — and reaffirms that a failed read still does.

---
## Summary by cubic
Fixes section loading vs failure so users see a loading state instead of error copy, and rebuilds the company record into a unified, panel-based layout with a full-width band and a single, collapsible rail.

- **Bug Fixes**
  - sectionState now resolves undefined views to LOADING while the composite read is in flight, and to UNAVAILABLE only after a failure.
  - The composite read’s pending flag is threaded to view-derived cards (People, Commercial, Recent Activity, Deals, Tags) so they show skeletons while loading.
  - Tests updated to cover the new loading behavior and AA checks for the record route.

- **New Features**
  - Full-width band between the header and columns: standing readings, a merged “Today on this account” brief, and the tab bar.
  - Rail becomes one panel: a Details grid over collapsible Health, People, Signals, and Tags & Lists; lifecycle and owner edits live in the header; tagging/list verbs live in the rail.
  - Tabs reworked as panels: People, Deals, Tasks, Documents, Profile; tasks can be completed inline; capped deal lists explain the cap; dossier and documents use consistent panels.
  - Design system additions: Panel (titled card with full-bleed rows), FieldGrid (aligned label/value), AvatarStack (overlapping avatars); Meter gains a flat accent variant.
  - i18n updated for new tab labels and copy; styling refined for stat cards, band spacing, and accessibility.

<sup>Written for commit f26859e8ee9ecda5d2ae46c0815f9d134d5371c2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1011?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


joshbruh-lang:

PR #1017, opened 2026-08-12
fix(company): the way-in pulse names the contact, not their id

joshbruh-lang:
EntityRef's own fallback shows a record's raw id while its name is still loading and whenever the lookup fails to resolve one — collapsing "not answered yet" and "answered with nothing" into the same rendering, which is fine for a reference sitting among other content but wrong for the header's own pulse line, the first sentence a reader meets: a raw uuid there reads as a broken page rather than as a name one request away.

`StrengthPulse` now runs the name lookup itself, under EntityRef's own cache key (`fetchEntityName`, now exported), and draws a skeleton while it is pending rather than delegating to EntityRef's id fallback mid-flight. A contact whose name genuinely cannot be resolved still falls back to the id once the lookup has actually answered, same as any other EntityRef.

This is the same defect class as `sectionState` collapsing "loading" and "failed" into "unavailable" (fix/section-loading-vs-unavailable): an answer not yet arrived, rendered as though it had arrived empty. EntityRef itself is unchanged — it has 18 callers, and fixing it globally is a separate, larger change; the next place this pattern turns up should be looked at with that in mind rather than re-derived from scratch.

Carries the rail's own loading-vs-unavailable fix underneath it (`fix/rail-section-loading`, itself on top of `fix/section-loading-vs-unavailable`) since it was committed on the same shared branch; not rebased away.

---
## Summary by cubic
Rebuilds the company record with a full-width band for standing, daily brief, and tabs, and fixes the header pulse to show a contact’s name (not a UUID) while it loads. Adds `Panel`, `FieldGrid`, and `AvatarStack`, and clarifies loading vs unavailable states across 360 and the rail.

- **New Features**
  - Full-width band between identity and columns: readings, “Today on this account”, and the tab bar.
  - Rail becomes one panel with a Details grid and collapsible Health, People, Signals, and Tags & lists.
  - Tab bodies render as panels; tasks can be completed from the row.
  - Design system: `Panel`, `FieldGrid`, `AvatarStack`; `Meter` supports a flat accent.

- **Bug Fixes**
  - Header StrengthPulse looks up the contact name via exported `fetchEntityName`; shows a skeleton while pending; falls back to id only after a null result.
  - `sectionState` and rail sections now distinguish “loading” from “unavailable”, showing failure only when a request actually fails.

<sup>Written for commit 3c4bdbf5e18704bc9300eccb7a55de01ec29778d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1017?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


joshbruh-lang:
Superseded by #1052. The way-in display it restores is already on the page there (the header names the contact rather than their id). The one thing this branch carried beyond that — a skeleton while the contact name resolves, so a raw id never flashes mid-load — did not survive the header rewrite and is worth redoing against the new header as its own small change, branched off main.

Reviewing here: https://github.com/gradionhq/margince-poc-v1/pull/1052

PR #1018, opened 2026-08-12
fix(rail): thread the composite read's loading flag into Health and People

joshbruh-lang:
## Summary
- The rail's Health and People sections read the account's composite view directly, and while that read was still in flight they showed the same "could not be loaded" copy a real outage draws.
- Both now take the composite read's own pending flag (threaded from CompanyPage, same shape `fix/section-loading-vs-unavailable` already applies elsewhere) and pass it into `sectionState`, so they render the waiting shape while the read runs and the failure sentence only once it has actually failed.

## Test plan
- [x] `pnpm tsc -b` — clean
- [x] `pnpm vitest run src/screens/companyrail.test.tsx` — 12/12, including two tests pinning the distinction (pending → skeleton, not unavailable copy; failed/undefined+not-loading → unavailable copy), verified non-vacuous by stashing the source fix and confirming both fail
- [x] `make check-fe` from repo root — exit 0

---
## Summary by cubic
Fixes the rail so Health, People, and Tags show a loading skeleton while the composite view is fetching, and only show “could not be loaded” if the read actually fails. This removes the failure flash on page open.

- **Bug Fixes**
  - Thread `loading` from `CompanyPage` to `CompanyRail`, then into `HealthSection`, `PeopleSection`, and `TagsSection` via `sectionState`.
  - Distinguish “loading” vs “unavailable” for sections that read directly from the composite view.
  - Expanded tests to pin both cases (loading ⇒ skeleton; failed ⇒ unavailable).

<sup>Written for commit 8db791f72959e603a1dee372ba39e3f2fb2ec19c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1018?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/screens/companyrail.test.tsx
+++ b/frontend/src/screens/companyrail.test.tsx
@@ -115,14 +115,14 @@ function stub(overrides: Record<string, (req: Request) => Response> = {}) {
 describe("CompanyRail", () => {
   it("renders nothing while the composer holds the column", () => {
     stub();
-    render(<CompanyRail orgId="o-1" view={view()} withPeople composerOpen />);
+    render(<CompanyRail orgId="o-1" view={view()} withPeople loading={false} composerOpen />);
     expect(screen.queryByText("Details")).not.toBeInTheDocument();
   });
 
   it("draws the details grid from the fields the record actually carries", async () => {
     stub();
     render(
-      <CompanyRail orgId="o-1" view={view()} withPeople composerOpen={false} />,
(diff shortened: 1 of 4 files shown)


joshbruh-lang:

PR #1019, opened 2026-08-12
fix(dossier): a company's own name is not a line on its own profile

joshbruh-lang:
The Profile tab's dossier printed a raw field key and a doubled full stop, both from the same deterministic composer (the floor a deployment with no model configured writes from):

- "display name: Sontana." led every dossier, restating the account's name — already the page's own heading — under a label nobody wrote for a reader, because `display_name` had no entry in the label table and the composer fell back to the column name with its underscores opened out. It is now skipped, the same way an unmapped field is already skipped in the account brief this composer sits beside.
- Every other line ended in "..": the composer appended a full stop to a value that had frequently already written its own, doubling it. The account brief's own composer solved exactly this (a value's own terminator is kept, only a bare one gets a closing stop) in a function the dossier never called. That function is now `claims.TerminateSentence`, shared by both composers rather than duplicated between them.

The third defect (the citation row printing "profile field" ten times) was fixed earlier and is already in the redesign this branches from.

Scope: content defects only, no change to how the facts are presented.

---
## Summary by cubic
Fixes the dossier so it no longer restates a company’s own name and stops writing doubled punctuation, while also unifying the company record into a consistent, panel-based layout with a full-width band and a single rail. Adds a shared backend helper for sentence termination and introduces small design-system components used across the new layout.

- **Bug Fixes**
  - Skip unmapped labels in the dossier (e.g., `display_name`) so the company name isn’t repeated.
  - Add `claims.TerminateSentence` and use it in both brief and dossier to keep an author’s terminator and only add a period when needed.
  - Expand unit tests to cover termination and label mapping.

- **Refactors**
  - Rebuild the company record: a full-width band (standing readings, daily brief, tabs), a single rail panel (Details grid + collapsible sections), and tab bodies as panels.
  - Consolidate inline edits: one control per field; add/modify tags and lists from the rail only; tick tasks complete in-row.
  - Add `Panel`, `FieldGrid`, and `AvatarStack` to the design system and use them across the record.
  - Update i18n keys, tests (unit/e2e), and layout styles to the new structure.

<sup>Written for commit af9f02c94d0359b24cb41c22bb4dba906a424751. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1019?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


joshbruh-lang:

Issue #1039, opened 2026-08-12
overlay: Apply accepts two fields writing the same canonical target


dungbanguyen:
`overlay.Apply` (`backend/internal/modules/overlay/mapping.go`) checks that no two child rows of one parent claim the same position, but nothing checks that two FIELDS do not claim the same canonical target. Two shapes slip through:

- Two `TargetColumn` fields declaring the same `To`. The last one in declaration order wins and the earlier value is gone, with no error and nothing in the unmapped list.
- A `TargetColumn` whose `To` equals a `TargetChild`'s parent key (e.g. `To: "person_email"` alongside `To: "person_email.email"`). The column write and the child collection land on the same key, and which survives depends on field order.

Both are silent: the mapping is accepted, the record projects, and the loss shows up only as a canonical key that never carries what a reader expects. A `Const` target is already checked against a field's target, and a child target is already checked against another child's position — the field-vs-field case is the gap between those two.

Pre-existing; the child-row position check added by the core-field reconciliation work narrowed the class without closing it.

`checkChildRowDeclarations` is the obvious home: it already walks `m.Fields` once per `Apply`, independently of what any record carries, and already owns the "a declaration defect must not wait for the right data to surface" reasoning. Collecting each field's effective target key (the `To` for a column, the parent for a child) into one seen-set there would catch both shapes with the same walk.

Issue #1038, opened 2026-08-12
overlay: a mapping addition reaches only records re-synced after the deploy


dungbanguyen:
A mirror row is re-projected only when the incumbent's own baseline advances: `ingestSQL` rewrites `fields` under `EXCLUDED.updated_at_baseline > overlay_mirror.updated_at_baseline`, and `Backfill` short-circuits on a converged cursor. So a canonical key a mapping starts producing today reaches only the records the incumbent modifies after the deploy. A record nobody edits upstream again never carries it.

Two of this branch's additions are in that position, and both degrade gracefully rather than answering wrongly — which is why this is a bounded-coverage note, not a defect:

- **`created_at`** (person and organization). A pre-existing row carries no `created_at`, so `overlayTimeOr` falls back to the mirror's own sync instant. The record reads as created when we mirrored it, which is the honest claim the mirror can make for itself, but it is not the incumbent's create instant the binding advertises.
- **`person_phone`** (and the child-collection shape generally). A pre-existing row carries no `person_phone` rows, so the contact publishes an empty phones collection indefinitely. The numbers are absent, never wrong.

The consequence to record: the registry's `mapped` dispositions and the wire-coverage gate are claims about what the pipeline PRODUCES, not about what every stored row already holds. They are honestly true only for records re-synced after the mapping change that introduced the key. Any future mapping addition inherits the same bound.

What would close it: a re-projection pass that replays the current mapping over stored `raw` payloads without waiting on an incumbent edit, or a mapping-version watermark the backfill treats as a reason to revisit a converged cursor. Either would also serve #996, which is the flip-side of the same mechanism.

Raised from the core-field reconciliation review.

Issue #1032, opened 2026-08-12
overlay: confirm createdate on HubSpot companies against a live portal capture


dungbanguyen:
companiesMapping maps `createdate` to the canonical `created_at`, and the mapping comment states that `createdate` is the same property name on contacts and companies (unlike the last-modified property, which is `lastmodifieddate` on a contact and `hs_lastmodifieddate` on a company). That was confirmed against HubSpot's documentation only. The contact side rests on a live DACH portal capture (`rawContact`); the company side does not. Capture a live company and confirm.

This is a fidelity confirmation, not a correctness risk: `propertyNames` feeds only the `properties=` list on the read, never a filter, so a property name the portal does not know cannot 400 the request. If the property were absent, no canonical `created_at` key is landed and the wire falls back to the mirror's own sync instant — the pre-existing behaviour. The cost of being wrong is a mirrored company reporting when we mirrored it rather than when it was created.

dungbanguyen:
## Decision, 2026-08-31: fold this into the one shared company capture

Re-triaged against `origin/main` at `8ecb691f1`. Confirmed. The asymmetry is
visible in the fixtures themselves:

- `overlay/hubspot/mapping_hs_test.go:14` — `rawContact` is described as a
  **spike** fixture, i.e. taken from a live portal.
- `:216` — `rawCompany` is described only as *"a §9-shaped HubSpot company
  properties map"*, i.e. hand-built to the documented shape.

So the contact side rests on evidence and the company side rests on
documentation.

### The ruling

**One live company capture, answering this and the two sibling mappings at
once.**

https://github.com/margince/margince/issues/1028 (`lifecyclestage`) and
https://github.com/margince/margince/issues/1031 (`type`) were both decided today
to require a real portal capture before their transforms are written. This is a
third question the same capture answers, and capturing one company three times
would be silly.

So: take one capture of a real company, and from it —

1. confirm `createdate` is the property name on companies (and that the
   last-modified property really is `hs_lastmodifieddate` rather than
   `lastmodifieddate`, since that asymmetry is the reason to doubt the first);
2. record what `lifecyclestage` actually holds (#1028);
3. record what `type` actually holds (#1031).

**And replace the hand-built `rawCompany` fixture with the real capture**, so the
company tests rest on the same kind of evidence the contact tests do. That is the
durable half — it means the next company property confirmed against documentation
alone has somewhere honest to be checked.

### Severity, kept accurate

[...]

Issue #1031, opened 2026-08-12
overlay: map a mirrored company's type to relationship_types


dungbanguyen:
HubSpot's company object carries a standard `type` property (PROSPECT/PARTNER/RESELLER/VENDOR/OTHER), while Organization.relationship_types is this product's own multi-valued vocabulary (customer/partner/supplier/investor/portfolio_company/competitor/other). The two overlap only partially, and one value cannot be carried at all: an organization IS a partner iff it carries `partner` here AND has a `partner` extension row (Organization.relationship_types in the contract, migration `0175`), and a mirrored company has no extension row — so the transform has to DROP HubSpot's PARTNER rather than map it, which is a decision the mapping cannot make on its own. Folding a single-valued source into a multi-valued target needs a documented transform and a real portal capture of what `type` actually holds before it is written — the same flag-don't-invent deferral #1028 takes for `lifecycle`, the sibling field ADR-0079/A124 split out of the same retired `classification`. Recorded as deferred in overlay.FieldBindings().

dungbanguyen:
## Decision, 2026-08-31: carry the values that map; drop PARTNER deliberately and visibly

Re-triaged against `origin/main` at `8ecb691f1`. Confirmed as a declared deferral
— `backend/internal/modules/overlay/fieldbinding.go:226` points here, immediately
below its sibling `lifecycle`
(`:225` → https://github.com/margince/margince/issues/1028, still open). Both came
out of the same retired `classification` field, and both take the same
flag-don't-invent posture.

### The ruling

**Map the values that translate. Refuse PARTNER explicitly, with the reason
recorded, and surface it as unmapped rather than dropping it silently.**

PARTNER is the whole reason this needed a decision, and the answer follows from
the product's own rule: an organization **is** a partner only if it carries
`partner` in `relationship_types` **and** has a partner extension row. A mirrored
company has no extension row. So writing `partner` from HubSpot's `type` would make
the record assert a status the product's own definition says it does not hold —
which is worse than not carrying the value, because a half-partner is a record that
reads as a partner to every surface that checks only the vocabulary.

Refusing it is therefore not a limitation to apologise for; it is the mapping
declining to invent a fact. What matters is that the refusal is **visible**: the
value goes into Apply's unmapped-flag list so a user can see the incumbent said
something we did not take, and the transform states why in the code beside it.

### On the shape mismatch

Folding a single-valued source into a multi-valued target is fine and needs saying
[...]

PR #1044, opened 2026-08-12
feat(overlay): the mirror's field coverage becomes provable

dungbanguyen:
Three layers had to agree about a field — the incumbent mapping's target string, the mirror's jsonb key, and the wire assembly's field pick — and nothing connected them. Two defects followed, both live: the contacts mapping still described `person_phone` as a custom column it has not been since migration `0004`, and `address` was mapped into the mirror for both entities and picked up by nothing, so the value existed in storage while the slot a client reads stayed empty. Every layer was correct on its own.

## What this adds

A pure-data registry (`overlay/fieldbinding.go`) declaring, per contract field, whether the mirror carries it (`mapped`) or why not — `deferred` with a filed issue, `unmappable` with a reason, or `native_only` for what is derived or server-stamped. Three gates hold it honest, all in `make check`:

- a mapping target must name a registry key
- every contract field of an **armed** entity must be dispositioned, derived from `api/crm.yaml`
- every `mapped` binding must reach the slot a caller receives — asserted behaviourally, including against an empty-payload baseline, because a slot filled by a fallback passes a presence check

`person` and `organization` are armed. `deal`, `lead` and `activity` are declared `Armed: false`, so the remaining work is visible in code rather than in a backlog note.

## Coverage gained

A mirrored contact now publishes its emails, phones and address; a mirrored company its address. Both carry the incumbent's own created/updated instants instead of the sync instant — which had made "recently updated" mean "recently synced", the same answer for every record in the workspace.

The address transform normalizes HubSpot's property names to the contract's `Address` member names, and `TargetChild` mappings land a collection of rows rather than one nested object, so a contact's work and mobile numbers are separate typed rows.

## Bounded claim

The mirror's on-disk keys are unchanged, so no re-sync and no incumbent API budget is spent. But `ingestSQL` re-projects a row only when the incumbent's baseline advances, so a record never modified upstream keeps its stored payload indefinitely: the readers accept both the old and new shapes, while newly mapped fields reach only records re-synced after deploy (#1038).

## Not done, and tracked

`owner_id` is `deferred` on both entities — nothing resolves the incumbent owner to an `app_user` on the read path (#994). Organization's `description`, `website_url`, `linkedin_url`, `lifecycle`, `relationship_types`, `parent_org_id` and `archived_at` are each `deferred` with an issue, because each needs a product decision and a live-portal capture rather than a one-line mapping (#1023–#1028, #1031). Also filed: #996, #1004, #1006, #1016, #1032, #1039.

---
## Summary by cubic
Makes overlay field coverage explicit and provable. Previously, person/organization dropped mapped emails/phones/address and reported the mirror’s sync time as created/updated; now they publish those fields and use the incumbent’s timestamps, with child fields emitted as ordered collections with stable row ids.

- Registry: `overlay.FieldBindings()` dispositions every contract field (`mapped`, `deferred`, `unmappable`, `native_only`); `person` and `organization` are armed; `deal`, `lead`, and `activity` remain unarmed.
- Gates: mappings must reference a registry key; every contract field for armed entities is dispositioned from `api/crm.yaml`; each `mapped` binding must reach the client response (seeded via the real mapping pipeline); fixtures prove the declared incumbent properties exist.
- Child fields: `TargetChild` produces collections with declared attributes (`type`, `is_primary`, `position`), ordered by position, with stable synthetic row ids distinct per row; organization domains are read as a collection.
- Address: the transform writes contract member names; readers accept both old and normalized spellings to keep legacy rows readable.
- Timestamps: `created_at`/`updated_at` come from the incumbent (fallback to mirror sync time when absent).
- Structure: mirror-wide dispositions are defined once and composed; `workspace_id` is no longer bound because it is not published; `owner_id` is `deferred` until resolved to an `app_user`.
- Implementation: field extraction moved to overlaywirefields.go; coverage and wire gates added; Sonar config treats the field registry as a data table, not duplication.

**Migration**
- No mirror schema changes and no forced re-sync. Newly mapped fields appear on the next upstream change; readers accept both old and new payload shapes.

<sup>Written for commit dce5fd6c01171984ff0924451978ff914b80cd4a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1044?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Improved CRM synchronization for contact and company data.
  * Added structured support for addresses, emails, phone numbers, and company domains, including primary status and ordering.
  * Added support for creation and last-modified timestamps.
  * Improved address field normalization while preserving additional properties.

* **Bug Fixes**
  * Improved compatibility with both collection-based and legacy single-object data formats.
  * Added safeguards to ensure mapped fields are consistently included in synchronized records.

### DIFF
--- a/backend/.go-arch-lint.yml
+++ b/backend/.go-arch-lint.yml
@@ -158,11 +158,7 @@ deps:
     canUse: [pgx, oapi]
   overlay:
     mayDependOn: [overlay, contracts, platform]
-    # yaml: the field-coverage gate reads api/crm.yaml directly, the same
-    # way root's rbacvocabulary gate does, to derive the contract's field
-    # list without a generated intermediary the gate itself would have to
-    # trust.
-    canUse: [pgx, oapi, yaml]
+    canUse: [pgx, oapi]
   # The importer engine (import-export-migration): sources and native
   # writers are injected seams (compose), so the module itself sees only
   # the platform — never the record modules it feeds.
--- a/backend/.go-arch-lint.yml
+++ b/backend/.go-arch-lint.yml
@@ -158,7 +158,11 @@ deps:
     canUse: [pgx, oapi]
   overlay:
     mayDependOn: [overlay, contracts, platform]
-    canUse: [pgx, oapi]
+    # yaml: the field-coverage gate reads api/crm.yaml directly, the same
+    # way root's rbacvocabulary gate does, to derive the contract's field
+    # list without a generated intermediary the gate itself would have to
+    # trust.
+    canUse: [pgx, oapi, yaml]
   # The importer engine (import-export-migration): sources and native
   # writers are injected seams (compose), so the module itself sees only
   # the platform — never the record modules it feeds.
--- a/backend/gatecensus_test.go
+++ b/backend/gatecensus_test.go
@@ -50,7 +50,7 @@ const (
 	// Permitting the marker without counting it would let it spread quietly and
 	// reopen the class these rules close; pinned, every new one moves a number a
 	// reviewer sees.
-	wantFixtureAnnotations = 11
+	wantFixtureAnnotations = 12
 )
 
 // censusDecl is one package-level declaration this census governs — a map from
--- a/backend/internal/compose/flipchildrows_test.go
+++ b/backend/internal/compose/flipchildrows_test.go
@@ -1,0 +1,132 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+package compose
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/gradionhq/margince/backend/internal/modules/overlay"
+	"github.com/gradionhq/margince/backend/internal/modules/people"
+)
+
+// twoEmailRowsMapping declares a contact whose work and personal addresses are
+// separate rows of one collection — the shape a mapping reaches for the moment
+// an incumbent carries more than one address, and the shape the wire already
+// publishes whole.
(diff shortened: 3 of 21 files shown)


dungbanguyen:
Addressed the first half in 0f830fc4: `flipPersonEmails` and `flipOrgDomains` now iterate the whole child collection and emit one input per valid row, holding `email_type` to the contract enum per row rather than only for the first. `TestFlipCarriesEveryRowOfAChildCollection` builds the payload through the real `overlay.Apply` from mappings declaring two rows each, and re-adding a `break` after the first append makes it fail — so the iteration is pinned, not incidental. That closes the divergence you identified: the read wire published the whole collection while the flip wrote only the leading row, and flip writes are durable.

Not taking the second half — `CreatePersonInput.Phones` — deliberately. `people.parsePersonContacts` calls `values.ParsePhone`, which returns an error on a number it cannot normalize to E.164, and HubSpot sends numbers unnormalized (a live portal capture in the fixtures holds `"49 176 10042069"`). Passing the mapped rows straight through would turn a succeeding flip import into a hard failure for many contacts. Whether an unparseable number is dropped, disclosed as a skip the way `duplicate_email` already is, or fails the row is a product decision rather than a mechanical fix, so it is tracked in #1004 with that analysis rather than guessed at here.

PR #1058, opened 2026-08-12
feat(flip): a landing commits the record and its identity together

tinnhlt-gra:
## What was wrong

An imported record lands in two steps that are **not** one transaction. `flipreconcile.go`
opens by saying so:

> "A record lands in two steps that are NOT one transaction: the owning module store creates
> the native row … then the engine's identity map records what that external id became. A
> process that dies between them leaves a created record the resume cannot see, and the resume
> would create it again."

That file exists to repair it — a scan that adopts live rows carrying the reserved import
provenance the map does not know.

The shape was not a preference. A module store **could not** be called with a caller's
transaction; #1047 changed that.

## The change

The three people classes land in one transaction. `CreatePersonTx`, `CreateOrganizationTx`
and `CreateLeadTx` are the caller-opened twins of the creates they share a body with — same
gates, same order, only the transaction borrowed — and `migration.RecordIdentityTx` writes the
map row beside them. A landing that fails now leaves neither the record, nor its audit row,
nor a map entry.

Three details the shape turns on, each with its own test:

- **The run cache is written after the commit, never inside it.** `lookup` answers from that
  cache before it ever asks the map, so an entry for a rolled-back landing would make this
  run's later pages — and the association phase — resolve an id that does not exist.
- **A lead the store replays under its natural key aborts the transaction** rather than
  committing an identity row. That row would make the next attempt report the estate
  converged, turning a one-shot disclosure into none. Same rule as before, now enforced by the
  rollback instead of by an early return.
- **Custom fields are refused on a caller-opened create, not dropped.** The catalog they are
  matched against is read in a transaction of its own — precisely the second connection these
  seams exist to avoid — so a create that accepted them would answer success with the values
  missing and nothing saying why. The store-opened entry points carry them exactly as before.
  This is the design's own ruling (SP1 §3.3) rather than a new one.

`flipreconcile` stays. It adopts orphans from attempts that predate this, and from the deal
and activity classes, which keep the two-step landing for now: the deal because its landing
takes a second transaction anyway (born open, then advanced to its terminal stage), the
activity because it is simply next.

## Tests, each shown red against the defect it describes

`fliplanding_integration_test.go` drives the real writers. The forced failure is a production
path rather than a hook: the identity row's composite FK names the import run, so a landing
recorded against a run this workspace does not hold fails exactly where a crash would — after
the record is written, before the commit.

Against the **old two-transaction landing**, restored as a probe:

```
--- FAIL: TestAFailedIdentityWriteLeavesNoPersonBehind
    person rows = 1, want 0 — the record outlived the transaction that was supposed to carry its identity
    audit rows = 1, want 0 — the write shape's audit row committed without the record it describes
--- FAIL: TestAFailedIdentityWriteLeavesNoOrganizationBehind
--- FAIL: TestAFailedIdentityWriteLeavesNoLeadBehind
```

That probe left the cache suite green, so it was probed separately against **its** defect —
caching from inside the landing transaction:

```
--- FAIL: TestARolledBackLandingCachesNothing
    the run cache names a person the failed landing never committed, so the resume would skip creating it
```

Both green on this branch, alongside the replay arm (a replayed lead is skipped and left
unmapped) and three new arms in the pool-of-one suite: each caller-opened create runs with the
field catalog wired and an active custom column present — the arrangement under which a seam
that reached for the catalog would block — and each refuses custom-field values.

## Why this is one PR

Steps 2–4 of the extension core write port produce exported seams whose planned consumer is
`Tx.Core()`, several steps away. Shipping them alone would be exported API with no caller,
which T3 forbids and which both in-tree precedents avoided (`UpdateDealTx` and `LogActivityTx`
each arrived with `compose/extractionaccept`). The flip is a consumer that exists today and
wants exactly this.

---
## Summary by cubic
Flip imports now land every class — people, organizations, leads, activities, and deals — by committing the native record and its identity-map row in one transaction to prevent orphans and bad cache entries. Deal landings are atomic; the stage advance still happens afterward.

- **New Features**
  - Added `CreateDealTx`; wired `CreatePersonTx`, `CreateOrganizationTx`, `CreateLeadTx`; added `RecordIdentityTx` so the identity row writes in the same commit.
  - Flip landing uses a single atomic commit and updates the per-run cache only after commit.
  - Lead and activity replays under their natural keys roll back and return a skip; deals are born open and advanced later, with reconcile kept for pre-change or mapped-but-open deals.

- **Bug Fixes**
  - Eliminates orphaned records and stray audit rows across people, organizations, leads, activities, and deals.
  - Ensures no cache entries are written for rolled-back landings.
  - Caller-opened creates reject custom fields (`person`, `organization`, `lead`, `deal`) to avoid a second connection and silent drops.

<sup>Written for commit 00fed6aa008de784766383e68271ac3b1e5739ad. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1058?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Reliability**
  - Imports now create people, organizations, leads, and identity records atomically.
  - Failed identity updates automatically roll back related records, audit entries, and cached data.

- **Duplicate Handling**
  - Replayed leads with the same natural key are skipped cleanly without creating duplicate identity mappings.
  - Existing duplicate-person behavior remains supported.

- **Data Entry**
  - Transaction-based creation now supports people, organizations, and leads while preserving validation, deduplication, auditing, and event tracking.
  - Custom fields remain supported through the standard store-managed creation flow.


tinnhlt-gra:
The two classes this PR leaves on the two-transaction landing — activity and deal — are filed as #1059, with the reason each was held back: the activity is straightforward wiring around the `LogActivityTx` that already exists, while the deal needs the exported create seam **and** a decision about its second, by-design window (born open, then advanced), which `settleAdoptedDeal` compensates today.

tinnhlt-gra:
Both reviews are in, and they converged on the same finding from different directions: converting three of five classes left the sibling copies unfixed — the pattern CLAUDE.md's first review rule names — and the comment explaining the deferral was the rationalization its fifth rule forbids. The activity's seam (`LogActivityTx`) had existed all along, so "next" was a choice rather than a constraint.

**All five classes land atomically now** (00fed6aa). `deals.CreateDealTx` is the twin the deal landing needed; the deal keeps only the window it has by design — born open, then advanced — where `settleAdoptedDeal` still finishes a mapped-but-open deal. With every class converted, `remember` had no callers left and is gone, and the two comments still describing the whole writer surface as two transactions (`flipreconcile.go`'s header and `migration.Writers.ReconcileIdentities`' doc) now state what is true. #1059 is closed as done rather than deferred.

Three test corrections, each from the review:

- **The duplicate-email skip had no test at all** — and it is the one preserved behaviour that now depends on the store's error travelling out through a rolled-back landing. It has one, shown red against a landing wrapped without `%w`: precisely the refactor that would turn every such skip into a failed run with nothing going red.
- **The organization and lead failure arms accepted any error**, so they would have passed had the landing failed *before* the identity write ran. All five now require that refusal specifically.
- **The success arms counted a map row**, which proves less than it looks: `import_record_map` carries no FK to the native tables, so each arm now joins the record it names.

Also fixed: `ensurePerson` wrapped its error outside the landing while its siblings wrapped inside, reporting an identity-write failure as "creating person".

Local `make check` and the full integration lane (`0 skips`) are green on the amended branch.

PR #1063, opened 2026-08-12
feat(privacy): an admin can author the retention ladder, and an installation can refuse to delete

### LINKED ISSUE #695: Let an admin author retention policies, including a never-delete posture
## What is missing

The retention engine runs, but nobody can configure it. There is no screen and no API operation that creates or changes a retention policy — the rows are written once at workspace bootstrap and never again. The code says so itself in `backend/internal/modules/privacy/retention.go`.

Worse: the seeded defaults **delete** data. Activity transcripts and AI call payloads are set to erase after 365 days. A customer who told us "keep everything for seven years" would silently lose data on our defaults.

## Why it matters

A customer engagement in flight has exactly that requirement — seven years, no deletion. Today meeting it means a code or seed change by us, which is not a product. Any regulated customer will ask for the same control.

- The table: `backend/migrations/core/0010_consent_retention.up.sql` creates `retention_policy` (object type, category, retain days, action archive/anonymize/erase, lawful basis, enabled) plus legal-hold columns.
- The evaluator and the ladder: `backend/internal/modules/privacy/retention.go`, `retention_floor.go`, `retention_graph.go`.
- The nightly job is wired: `backend/internal/compose/jobs_privacyretention.go`.
- Legal holds are honoured — a row on hold is never auto-acted.
- The current defaults live in `backend/internal/modules/consent/retention.go`.

## What to build

1. Contract operations and a settings screen to list, create, edit, enable and disable retention policies.
2. A "retain only, never erase" posture at workspace level that no policy can override.
3. Fix a real design constraint noted in the code: the job's maximum pass duration is derived from the assumption of one row per selector. Adding rows breaks that bound, so the timeout maths has to change with this work.
4. Revisit the shipped defaults so a fresh install does not delete data a customer expected to keep.

## Done when

- [ ] An admin can set a seven-year retain-only policy from the UI, and it survives an upgrade.
- [ ] The nightly job honours the authored policies and still finishes inside its bound.
- [ ] Legal holds still win over everything.

**Size:** S–M — the engine is done, this is the doorway.
**Spec:** `gdpr-compliance-surfaces.md` — UC-GDPR-09 (author a custom retention policy) and UC-GDPR-03 (the nightly ladder).

tinnhlt-gra:
Closes #695. Wave 0 / PLAN R1 of the capability roadmap.

The retention engine has run since ADR-0011 and **nobody could configure it** — the rows were planted once by the workspace bootstrap and never written again. A customer needing seven-year retention needed a code change from us, which is not a product.

**Spec landed first** (contract-first, P3): gradionhq/margince-foundation#1286, merged. It was a hard prerequisite, not diligence — `GCS-GAP-2` and UC-GDPR-09 `F1`/`F2` each read *"flagged for reconciliation, do not invent behavior."*

## The surface

Six operations, admin/ops-only behind a new `retention_policy` RBAC object, every **write** `x-agent-access: human-only` because an agent never authors a deletion rule. Audit-only, no outbox event — the closed catalog has no retention-policy stream (`EVT-NOEVT-3`), and `retention.applied` stays the *engine's* statement about a record it acted on.

| | |
|---|---|
| `GET`/`POST /retention-policies` | list with live suppression state / author |
| `PATCH`/`DELETE /retention-policies/{id}` | sparse edit (`enabled:false` disables without deleting) / remove |
| `GET`/`PATCH /retention/settings` | the retain-only posture |

Plus a Retention card on the privacy settings tab: the posture toggle with its consequence spelled out, the ladder with a per-row **"Suppressed by retain-only"** reading, inline edit, and delete behind a confirm modal that says deleting ≠ pausing.

## Two invariants moved with it, because adding a writer is what made them reachable

**The `UNIQUE` constraint did not enforce what two documents claim.** `category` is nullable and Postgres counts NULLs as distinct, so any number of `('activity', NULL)` rows were legal — while UC-GDPR-09 step 2 asserts a second row for one scope is rejected, and `privacy.MaxPassDuration` (the nightly job's **declared timeout**) is derived from the scope count. The scheduler's cap was a fiction waiting for a second row. Core `0217` makes it `UNIQUE NULLS NOT DISTINCT`, and the test inserts **directly, bypassing the store** — a handler-level refusal would look identical while leaving every other writer free to break the bound.

**`MaxPassDuration` is now held by that constraint plus scope resolution**, not by the absence of a feature. It grew with the new `deal/won` selector, and the job census refused the mismatch until `api/jobs.yaml` moved with it — exactly the coupling #695 point 3 asked us to fix, announcing itself.

## The authorable set is derived, not listed

`GCS-PARAM-8` is the evaluator's own selector table, carried as one contract enum so a client cannot offer a scope that would be stored and then skipped, loudly, every pass forever — reporting governance the installation does not have. A scope outside it is `422` **naming what is authorable**, because a closed vocabulary the caller cannot enumerate is unfixable from the error alone.

That set gains **`deal/won`**, which UC-GDPR-09's main-success scenario authors and no selector answered: the worked example of the use case could not be performed.

## The keep-everything obligation, without editing the defaults

`privacy.retain_only` suppresses every destructive action — `anonymize` and `erase`, never `archive`, because archiving retains — **without touching a policy row**. The ladder stays visible, so an admin sees exactly what would happen if the obligation lapsed, and lifting the posture resumes enforcement with nothing to re-author. `seeds.retention.default_policy: retain_only` in `margince.yaml` declares it before first boot, closing the window between seeding the rows and the first admin sign-in — a key `single-organization-auth-concept.md` already drew and nothing implemented.

Suppression is **not audited per record**: `audit_log` records mutations and a suppression is the absence of one, so a row per skipped record would be ~1,400/night saying nothing happened. The posture *change* is the audited event, the per-pass skip is logged, and `suppressed_by_posture` reports the live state on the row itself.

**Both destructive seeded defaults stand.** #695 point 4 asked for a revisit. `DM-SEED-3` and `DM-SEED-6` are Art. 9 arguments, and DM-SEED-6's window is written into **ADR-0064/A110 (Accepted)** — amending it from a build convenience would overwrite a ratified decision. The reasoning, and what changing it would take (a founder decision, in an ADR), is recorded in the spec's Parameters appendix rather than left implicit. PLAN R1's *"a fresh bootstrap seeds no `erase` action"* is therefore **not met, on purpose** — the posture and the config key meet the obligation it names while every other installation keeps a storage-limitation default.

## Verification

- `make check` green; `make craft-static` green tree-wide; `make test-integration` green with **0 skips**.
- **New-file backend coverage 85.1%** (`retentionscope` 96%, `retentionpolicystore` 83%, `handlers_retention` 85%, `retentionposture` 83%). Frontend: `retention.logic.ts` 100%, `retention.tsx` 92.5%, `retentionpolicyform.tsx` 94.1%.
- The engine's existing legal-hold tests pass untouched — **holds still win over everything**.
- Governance check: all four writes generate as `human-only` with no MCP tool. The two reads carry no override, so an admin *may* lend an agent read access to the posture (capped by their own grant, ADR-0055) but no agent can ever change it.

Five fitness functions caught real obligations while building this — the job census, the `ServerInterface` compile assertion, the RBAC gate, the write-shape gate, and the RBAC-vocabulary gate. Each one is in the diff as a fix rather than a waiver, except the one audit-only waiver, which carries its reasoning in-source.

---
## Summary by cubic
Lets admins author the retention ladder and adds a retain‑only posture that suppresses destructive actions. Validates scope+action pairs so only executable policies are accepted, and protects the embed cascade while retain‑only is on.

- Adds GET/POST/PATCH/DELETE /retention-policies and GET/PATCH /retention/settings behind the `retention_policy` RBAC object; all operations are human‑only and writes are audit‑only (no outbox).
- Validates scope+action pairs against the executor table; rejects non‑executable combinations and names allowed actions for the chosen scope; any drifted row is skipped loudly rather than stopping the pass.
- Retain‑only suppresses anonymize and erase (not archive). Each policy reports suppressed_by_posture. The evaluator reads posture per destructive record, so mid‑pass changes stop destruction sooner. Legal holds still win. The embed sweep skips payload‑bearing embed calls while retain‑only holds, so the call → payload cascade cannot erase subject content.
- Authorable scopes derive from the evaluator’s selector table; adds deal/won. The server rejects unknown scopes and returns the authorable set; the UI offers exactly that set.
- Enforces one row per scope with UNIQUE NULLS NOT DISTINCT on (object_type, category). Increases the pass timeout to 5h5m0s and updates the job contract hash/specs.
- Update paths take a row lock (FOR UPDATE) to serialize concurrent patches.
- New Privacy → Retention screen: posture toggle; ladder with live suppression state; inline create/edit/disable; confirmed delete.
- Seeds support seeds.retention.default_policy in margince.yaml to set first‑boot posture without editing policy rows.

**Rollout/Migration**
- Run migrations; 0223 tightens uniqueness and installs the `retention_policy` object. Verify no duplicate (object_type, category) rows exist.
- If the installation must keep everything from first boot, set seeds.retention.default_policy: retain_only before provisioning. Existing installs can toggle the posture in Settings.
- Regenerate server/SDK code and job specs after pulling; the job contract hash and pass timeout changed.

<sup>Written for commit 36a4c20bb99bea46288767ab332b8f0c20bf4c87. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1063?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tinnhlt-gra:
## Review round applied — one high-severity data-integrity bug fixed

The mandated review round (craftsmanship + security redteam) found a **high-severity defect in this PR's own new surface**, plus three real hardening items. All applied in `c718d78b`; `eea735bd` carried an earlier round.

### The high one: an authorable pair that stops the nightly pass

`scope` and `action` are two independent enums on the wire, and the write path validated each half separately — but the evaluator dispatches on the **pair**, and only 7 of the 15 reachable combinations have an executor. So `POST {"scope":"deal/won","action":"erase"}` was a `201`, and on the next pass `apply` fell to its `default`, `EvaluateWorkspace` returned on the first due record, and the job stayed red **every night until somebody found the row**.

Worse: policies are ordered by `object_type`, so a poisoned `activity` row took every later policy with it. **One accepted request silently stopped storage limitation installation-wide** — and the create form made it a one-click accident.

Fixed by making the executor table the dispatch *and* the authorable set, one map, so they cannot drift. The refusal names the actions that scope does admit (which the two enums cannot express), a row that reached the table anyway is skipped loudly rather than aborting the pass, and a fitness test walks the table against the selectors in both directions. The integration test reproduces the original abort exactly when the guard is removed:

```
a policy with no executor aborted the whole pass:
  retention activity/ on 019ff6ae…: retention: no executor for activity/anonymize
```

### Also fixed

- **The two governance reads are now `human-only`.** I had called this defensible earlier; the reviewer's argument is better — every comparable read (`listRoles`, `listAuditLog`, `listAiCalls`) carries the annotation, and what an installation destroys and whether destruction is suspended is not reconnaissance to hand an agent, even one carrying an admin's passport.
- **The posture survives a data reset.** It was configuration-class, so a reset deleted it and re-ran the bootstrap seed — returning an installation that had opted in *through the screen* to the destructive ladder with the posture off. That is verbatim the state `seedRetentionPosture` exists to make unreachable.
- **The posture is re-read per record** for a destructive policy, not per stage. A stage is up to 200 records in 200 transactions, so a stage-level read alone let an admin who turned it on mid-stage watch 200 irreversible actions finish after the promise was made.

### One thing worth flagging in the fix itself

Converting the executors to named methods was not cosmetic. `updateguard_test.go` walks *named* functions to assert every by-id `UPDATE` carries a concurrency guard or is a ratified exception — **closures are invisible to it**, so the closure form I first wrote silently dropped three retention writes out of that gate's coverage. The single blanket waiver on `apply` is now three per-executor waivers, each stating what makes its absolute write correct.

### Verification

`make check` green · `make craft-static` green tree-wide · integration lane 37/38, the one failure being the pre-existing flake filed as #1064 (passes in isolation, passes on `main`, and the same ambient errors appear in its *passing* neighbours).

New-file coverage **86.0%**, every file ≥83% — and two previously untested destructive paths, `person/anonymize` (which deletes every satellite carrying the subject) and the bare activity archive, are now driven end to end. Those were pre-existing gaps, not introduced here, but an Art. 17 path nobody had watched run seemed worth closing while in the area.

tinnhlt-gra:
UAT proof of work: retention-authoring golden path recorded on this branch's dev stack (DEV_SLUG=retentionpr1063). Demonstrates: viewing the seeded retention ladder, creating a new `deal/won` policy, editing its window, and toggling the retain-only posture (destructive policies show "Suppressed by retain-only" while archive-only ones keep running).

retention-authoring-uat

Issue #1080, opened 2026-08-12
identity: the RBAC replay fixture and the published matrix are stale on main


LarsGradion:
`make check-backend` is red on a clean checkout of `origin/main`. Two tests in `internal/modules/identity` fail:

- `TestSeededRBACRoleDocumentsMatchTheCommittedReplayFixture` — `migrations/testdata/rbac_seeded_defaults.json`
- `TestPublishedRBACMatrixMatchesTheSeededRoleDocuments` — `docs/reference/rbac-matrix.md`

Both report the same divergence: the committed artifacts still carry a `lead` object where the server now seeds `integrations`.

## Reproduce

```
git switch main && git pull --ff-only
go test ./internal/modules/identity/ -count=1 -run RBAC
```

## Cause

A change moved the RBAC object matrix without regenerating the two artifacts that pin it. Both tests print the fix themselves:

```
go test ./internal/modules/identity/ -run RBAC -update
```

## Consequence

The merge gate is red for everyone, on every branch, regardless of what they changed. That trains people to read a red `check-backend` as noise, which is how a real failure gets waved through.

Whoever moved the matrix should regenerate and commit both artifacts — the values are theirs to confirm, not mine to guess.

PR #1084, opened 2026-08-12
chore(compose): move the two adapters out of server.go, under the size cap

LarsGradion:
The second half of main's red merge gate (the first is #1083).

`internal/compose/server.go` was **505 lines against a 500-line cap**, so `deterministic-gates` failed on a clean checkout of main and on every branch, regardless of what it changed:

```
FAIL: backend/internal/compose/server.go is 505 lines (> 500) — split into one-concept-per-file packages
```

## What moved

Two adapters, verbatim, into `serveradapters.go`:

- **`signalStrength`** — the arch-legal signals→people seam. signals declares its own seam type and the cross-module edge lives in compose, never as a signals→people import.
- **`paramParseError`** — maps a generated request-parameter parse failure onto the same 422 `validation_error` shape every other bad query input uses.

## Why these two

`server.go`'s job is to say **what this installation is wired out of** — the embedded handler sets, the `ServerInterface` assertion that makes a missing handler a compile error, and the wiring in `newServer`. A reader answering that question does not need either adapter body. Neither is part of what the composed Server *is*.

No behaviour change: both bodies are unchanged and their imports followed them. I did not take a waiver, because the file genuinely reads better split and a waiver would just move the problem.

## Gate

`make go-file-length` now passes (`no hand-written Go file exceeds 500 LOC, waivers: 0`), `go build ./...`, `go vet`, and `go test ./internal/compose/... .` all green. server.go is 465 lines.

---
## Summary by cubic
Moves two adapters from `backend/internal/compose/server.go` to `backend/internal/compose/serveradapters.go` to get under the 500-LOC cap and unblock `deterministic-gates`. No behavior change; code moved verbatim and imports followed.

- Moved `signalStrength` (signals→people seam) and `paramParseError` (maps generated param parse failures to 422 `validation_error`); `server.go` now focuses on wiring.
- `server.go` is 465 LOC; `make go-file-length` passes; `go build`, `go vet`, and `go test ./internal/compose/...` are green.

<sup>Written for commit 67c5c782565444dc92095b4855f413f7f3813f47. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1084?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/internal/compose/server.go
+++ b/backend/internal/compose/server.go
@@ -44,11 +44,9 @@ import (
 	"github.com/gradionhq/margince/backend/internal/platform/agentquota"
 	"github.com/gradionhq/margince/backend/internal/platform/auth"
 	"github.com/gradionhq/margince/backend/internal/platform/blobstore"
-	"github.com/gradionhq/margince/backend/internal/platform/httperr"
 	"github.com/gradionhq/margince/backend/internal/platform/httpserver"
 	"github.com/gradionhq/margince/backend/internal/platform/keyvault"
 	"github.com/gradionhq/margince/backend/internal/platform/overlaybudget"
-	"github.com/gradionhq/margince/backend/internal/shared/kernel/ids"
 )
 
 // Server satisfies crmcontracts.ServerInterface by embedding: every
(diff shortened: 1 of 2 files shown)


LarsGradion:
Superseded by #1089, which fixed the same breakage from the other end and split `server.go` the same way (it is 463 lines on main now).

Two sessions reached for the same red gate at the same time. Nothing here is worth rebasing on top of that — the adapters #1089 moved out are the same two this PR moved, so the diff is now empty in substance.

The other half of what I filed still stands: #1080 (the stale RBAC artifacts) merged as #1083 and is on main.

Issue #885, opened 2026-08-11
The reference extension ships in the default composition, so every production build serves the demo notepad


tinnhlt-gra:
`extensions/notes` (the reference unit, formerly `crm-demo`) is in the **default composition**, so every production build serves `/v1/ext/notes/*`: a reference notepad with a signing-key feature and a one-minute heartbeat.

Raised by @lstrojny in review of #659 (finding 12).

## Why it matters

The two acceptance-run defects on #659 were found **because that demo was running as product code**:

- a read seat could replace its signing key;
- a fresh installation produced a recurring heartbeat failure.

Both are fixed. They stand as evidence that the reference unit has production blast radius *only because it ships by default*.

A second symptom found while testing: nothing grants extension RBAC objects on a fresh install, so out of the box the unit's screen could only ever say "you do not hold access". #659 now adds an operator UI for that — but a demo unit that ships enabled and ungrantable is the shape of the problem.

## What to do

Move reference/demo units to `fixtures/`, or compose them only in a dedicated CI lane. Keep the default `extensions/` for capabilities intentionally shipped as part of the product.

**The machinery already exists**, which makes this small: `.github/workflows/ci.yml`'s `extension-reference` job already has an *"Enable the reference extension"* step that copies `fixtures/extensions/crm-hello` into `extensions/` for a composed build. So this is a `git mv` plus one `cp` line in a lane that already does exactly this for another unit.

`craft-static` already scans `fixtures/`, so the unit keeps its craftsmanship gate.

## The one entanglement

`notes` is the only unit with a frontend, so moving it out of the default composition also empties the composed SPA's unit-screen path and drops `extensions/*/frontend` from the pnpm workspace. That is why this waits on the frontend-slice split (#878) rather than landing with #659.

## Sequencing

After #659 and #878.

dungbanguyen:
Both sequencing dependencies are done — #659 and #878 are merged — and the entanglement this waited on is gone: `pnpm-workspace.yaml` no longer globs `extensions/*/frontend`, so moving the unit out no longer empties a workspace member.

I stopped short of doing it, because the move as specified would silently drop the unit's tests from CI and I cannot land the line that prevents that.

[...]

PR #1092, opened 2026-08-13
feat(extensions): a unit reaches core records through a governed port

tinnhlt-gra:
## What

`extension.Tx` grows one method — `Core()` — and a unit can file a note to a record: the unit's own row and a core activity, in one transaction.

This is **SP1 steps 7–11 + 13**. Step 12 landed separately in #1066.

## Why

A unit that wants a core record has SQL and the shared application role, so it can reach the table today. A write made that way lands with **no audit row, no outbox event, no RBAC check and no attribution** — every guarantee the product's own write shape exists to keep. The port is the same write the product makes, minus nothing.

```go
err := rt.Tx(ctx, func(ctx context.Context, tx extension.Tx) error {
    filed, err := tx.Core().Activities().Create(ctx, crm.CreateActivityRequest{
        Kind: crm.CreateActivityRequestKindNote, Body: &body, Source: "extension:notes",
    }.LinkTo(subject, subjectID))
    if err != nil { return err }
    _, err = tx.Exec(ctx, `INSERT INTO `+noteTable+` (…, filed_activity_id) VALUES (…, $5::uuid)`, …, filed.Id)
    return err
})
```

## The write is the product's own

`activities.LogActivityTx` already runs the RBAC gate, the per-link row-scope check, the insert, the audit row and the outbox event inside a caller's transaction. The port maps and delegates; it re-implements none of them.

**That also removed a step.** DESIGN §4 had filing call `Get` to "validate the caller-supplied subject exists and the caller may see it" — `insertActivityLinks` already calls `EnsureLinkTarget` → `EnsureVisibleLive`, the strict probe (exists, live, in the caller's row scope, existence check applying even to an unbounded actor). A `Get` repo would have been exported API with no caller, which is the mistake #1058's review caught once.

**And dissolved another.** Step 10 was "widen `BindExtensionRuntime` to reach the dispatcher". The port needs the *fresh* overlay answer, which is a workspace-row read, not a dispatcher method: it is now `overlayModeOf`, spelled once and called by both the dispatcher's uncached read and the port's guard. The binding keeps its two arguments and no role can wire the port half-way.

## What it refuses, and why each is a refusal rather than a silence

| | |
|---|---|
| **A job tick** | runs as the unit with no human behind it, and a core write is checked against the caller's live RBAC. The alternatives — a system principal (passes every check ever written) or resolving the agent seat's permissions (a governance surface of its own) — are features. `ErrForbidden`. A tick's own tables stay writable. |
| **Overlay mode** | the native tables this writes are not the live ones. `ErrOverlayUnsupported`, resolved **fresh** per write, never from the mode cache. |
| **Custom fields** | need the field catalog, and reading it takes a second connection inside a transaction the unit already holds — a deadlock shape. Refused, never accepted-and-dropped. |

Refusals are **mapped**, not wrapped: a unit is other people's code, so a table name, a constraint or a SQL state must not reach it. A fault with no class becomes "the core refused this write" and nothing else.

## Attribution

The actor stays the human — a unit can never exceed its caller — so "which unit wrote this" is a second dimension. `storekit` reads ctx-bound provenance and merges it under one **core-owned** `extension` evidence member; a caller supplying that member is an **error**, not a silent overwrite in either direction.

## The published shapes

`pkg/extension/crm` is the contract's own types, generated from a **subset**: `tools/contract-subset` cuts named schemas and their transitive refs out of `crm.yaml`, the existing `contract-overlay` downgrades it (a 3.1 subset does not generate), and oapi-codegen emits models with pruning **off** — with no operations, every schema has zero incoming refs and a pruning pass would empty the package. 381 lines importing only `time`; the whole contract would have been thousands of types that cannot satisfy `pkg-purity`.

The two type sets are bridged by JSON rather than a hand mapper. They come from the same schema, so a field added to the contract crosses automatically; a hand mapper would compile perfectly while dropping it.

## The exemplar

`POST /ext/notes/file`, gated on a **third RBAC object** `ext_notes_filing` — the reasoning that split the signing key out: a seat that may jot a note has no business writing the shared timeline. No seeded role holds it, so the capability is inert until an admin grants it deliberately (which is also why #885 stays out of scope). It needs `activity:create` too, and nothing declares that pairing — acknowledged, not solved.

The activity is created **first**, so the note's receipt column is written by the insert that creates the row; if that insert fails, the transaction takes the activity with it.

## Evidence

- **Unit**: the filing writes both and the timeline entry says what the note says; a refused core write leaves **no note** (asserted on the half that would otherwise be silent); every refusal class keeps its sentinel through the unit's wording; a subject the enum does not name and an id that is not one are refused before anything opens a transaction; authorship comes from the caller here too.
- **Compose**: the tick refusal holds with **no database at all** (a tick reaching a pool to be told no would mean the check ran too late); the mapper leaks no core error text; the JSON bridge carries a whole activity and a request with its links.
- **storekit**: the merge rule tested where it lives — bound, unbound, and a caller supplying the reserved member, in both directions.
- **Mutation-verified**: the tick refusal, the detail strip and the field-fault class each caught by the test that names them.

## Gates

`make check-backend` green (build, vet, golangci base **and** strict, arch-lint, unit tests, drift, composition), `make craft-static` green on all three trees. `origin/main` merged in.

## Review rounds folded in

Two independent adversarial reads, and they agreed on the two that mattered.

- **A core verb took the handler's context.** `Tx` re-binds the invocation's workspace, actor and correlation onto the callback's context — but `Create` accepted any context and passed it to `auth.Require`, the row-scope clauses and the audit row. A handler is handed a context per invocation, so it can keep one: an admin's, from an earlier call. Passing that during a lower-privileged call would have made the write the admin's. Every verb now re-binds through the Runtime's own `scoped`, and a Core built without that authority refuses rather than falling back.
- **Attribution shipped dead.** `provenance.WithExtension` had no non-test caller, so `ExtensionFrom` always answered false, every core write carried no `extension` evidence member, and three comments said otherwise. It is bound in `scoped` now — beside the actor, and for the same reason. The test that let it ship is the reason it shipped (storekit's positive case supplies the binding production never supplied); it stays, joined by one that calls `scoped` and reads what it produced.
- The overlay refusal answered **before** the caller's own `activity:create` grant was checked, so a seat that could not make the write at all learned whether the installation is in overlay mode. A failed workspace-mode read returned its own text — a relation name and a SQL state — to extension code. `crm.LinkTo` appended into shared spare capacity, so two links built from one base request could overwrite each other. All three fixed.
- The **second-acquire** — the mode read opened a connection inside the caller's open transaction, the deadlock shape this programme's own step 1 removed — was found and fixed mid-review. It is why #1095 exists: the static gate could not see it.

## Filed, not fixed here

- **#1094** — the port has no database-lane test. Atomicity in the reverse direction, the overlay refusal and the end-to-end attribution are proven by construction and by unit tests over fakes; the integration suite that would prove them is scoped in the issue.
- **#1095** — `txseamacquire_test.go` cannot see a transaction held in a struct field, which is how the second-acquire shipped green.
- **#1096** — `file_note` is not idempotent; a retry double-files.

## Not in this PR

**Step 14, the UI slice** — the Units nav group, the three surface exports (`problemMessageOf`, `isVersionSkew`, `ProblemError`), the filing control, and the three `sign` screen defects. It is additive affordance work over a capability that is already reachable over REST and MCP, and the nav group turns a fixed, order-pinned 10-item list into a variable-length one — its own slice, with its own tests.

---
## Summary by cubic
Adds a governed core port `extension.Tx.Core()` so a unit can create core CRM activities in the same transaction as its own rows. Fixes authority binding so core writes use the invocation’s identity, and stamps extension attribution; replaces unsafe direct SQL.

- Delegates to `activities.LogActivityTx`; shared mapping via `LogActivityInputFrom`; refuses job ticks (`ErrForbidden`), overlay workspaces (`ErrOverlayUnsupported`), and custom fields.
- Authority binding: core verbs re-bind workspace, actor, correlation, and extension provenance via `Runtime.scoped`. Handler-supplied contexts no longer affect identity; a Core built without binding now refuses.
- Attribution: `storekit.AuditWithEvidence` merges `provenance.WithExtension` and writes reserved `extension` evidence; caller-supplied `extension` errors.
- Overlay guard: `overlayModeOf` reads on the caller’s transaction after the `activity:create` grant check; dispatcher wraps reads with `WithInfraTx`. Read failures are logged and mapped, not leaked.
- Helpers: `crm.LinkTo` clones to avoid shared-capacity overwrites; `extensiontools` carries verb version into attribution; cursor and tx adapters moved to `backend/internal/compose/extensiontx.go`.
- Contract models: subset published in `backend/pkg/extension/crm` via `tools/contract-subset` and `oapi-codegen`.
- Notes exemplar: `POST /ext/notes/file` (gated by `ext_notes_filing`) creates the core activity first, then inserts the note with `filed_activity_id`; both commit atomically.

**Rollout and migration**
- Apply `extensions/notes/migrations/0004_note_filing.up.sql`.
- Grant `ext_notes_filing` and core `activity:create` to seats that may file notes; no seeded role includes `ext_notes_filing`.
- Overlay workspaces cannot use the port.
- Scheduled job ticks cannot use the port.

<sup>Written for commit a7b38ed44e49a4297d4c589a935b7a38b30cd7fd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1092?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tinnhlt-gra:
@coderabbitai review

tinnhlt-gra:
@coderabbitai full review

PR #1101, opened 2026-08-13
fix(authz): the compliance audit read and the subject-request queue are the admin's, and one predicate says so

luit-nfq:
Phase 2 of the settings restructure: one gating rule, and the permission defect that was hiding behind four of them.

## The defect

`ListAuditLog` gated on `auth.Unbounded(actor)`. Three seeded roles hold `row_scope = all` — `admin`, `ops` **and** `read_only` — so the check handed the full workspace governance trail to two roles the ratified governance matrix (AAD-ROLE-4, A91) denies, where it says a non-admin calling an admin-only governance op gets a 403.

It ran permissive in the direction that looks like nothing at all. A read-only member sees no button, so nobody notices — and a hidden button was never a control, so that member could call the operation directly. The audit read is oversight *of* ops' own machine-origin actions, which is why the matrix prints `—` for ops on that row while making ops identical to admin on every data object.

Row scope was the wrong predicate. The route now asks for the admin role, and the integration test asserts both halves: `ops` and `read_only` refused, the bounded rep and the agent principal still refused, the admin still served.

**Not** narrowed here: the DSR queue gates the same way, but its authority is genuinely unreconciled upstream — it ships as a 🟢 MCP tool with no admin pin, and "admin-mediated" settles who *fulfils* a request rather than who may see the queue. Acting on that would be acting on an open question. It is recorded in ADR-0102 §4 (margince-foundation#1285).

## The client had drifted the same way

Four call sites in three files asked which role the caller held, in two spellings, and the wider one — `admin || ops` — decided four settings tabs at once. One guess covering four different authorities:

| Tab | Server admits | Client showed | Result |
|---|---|---|---|
| Users & roles | admin | admin, ops | ops saw a tab that dead-ended on "admins only" |
| Extensions | admin | admin, ops | same |
| Audit log | *unbounded* → now admin | admin, ops | ops was served a read the matrix reserves |
| Privacy & consent | admin, ops | admin, ops | correct — consent config is genuinely Admin/Ops |

`useHoldsAdminRole` is now the single predicate, documented where it lives: reading a role key is the re-derivation the capability hooks exist to avoid, and it is admissible only for routes that gate on the role server-side and that no RBAC object describes. `useHoldsConsentAdminRole` stays separate because the authority differs — and it is interim in a way the admin one is not, since `consent_config` is a governed object upstream that is merely absent from the shipped vocabulary.

Neither consults the seat ceiling: an admin on a read seat still reads the roster and the trail, and each mutating control inside asks `useCanMutate` for itself.

## Tests

The nav test that asserted the old behaviour (`gives ops the three objectless tabs`) was documenting the bug. It now asserts the split from both sides — ops gets Privacy alone, admin gets all four — so a future widening fails here rather than in production.

## Verification

- `make check-fe` — green
- `make test-integration` — **2228 passed, 0 skips, 38 packages** (at `INTEGRATION_JOBS=4`; the default job count exhausts `max_connections=100` on this machine and fails unrelated packages on connect — environmental, worth its own issue)

---
## Summary by cubic
Restricts workspace governance reads to admins: the audit log and the data-subject-request (DSR) queue now require an admin human. Previously, `ops` and `read_only` could read via unbounded row scope; they now receive 403. Client gates Users, Extensions, and Audit for admins only; Privacy stays Admin/Ops, and the inbox withholds for non-admins.

- Backend
  - Audit log: `privacy.ListAuditLog` enforces human-only + `auth.RequireAdmin`; replaces the unbounded-scope check. Integration test updated to assert admin served; `ops`, `read_only`, bounded reps, and agents refused.
  - DSR queue: `consent` gates listing to human-only + admin; adds an integration test refusing `ops`, `read_only`, and bounded reps; admin reads.
  - SAR assembly: adds a human-only check; keeps unbounded-scope requirement for the cross-scope read.
  - Policy/tests/harness: pins that exactly three seeded roles are unbounded; moves permission fixtures to dedicated helpers and adds `OpsPerms` mirroring admin’s object grid without governance authority.

- Frontend and Rollout
  - Adds `useHoldsAdminRole` and `useHoldsConsentAdminRole`. Gates Users, Extensions, and Audit on admin; gates Privacy on admin or ops. `PrivacyInboxCard` disables the DSR fetch for non-admins and renders a withheld message. Tests and i18n updated.
  - No migrations. If any non-admin automation calls GET `/audit-log` or GET `/data-subject-requests`, it will now fail with 403; move those calls to an admin principal.

<sup>Written for commit eee91deff529867e6bccb7850ca35ed723990e8c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1101?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Access Control**
  * Restricted audit log access to authenticated administrators.
  * Limited Users & roles, Extensions, and Audit log settings to administrators.
  * Kept Privacy & consent available to administrators and operations staff.
* **Consistency**
  * Centralized role-based access checks across settings, extension access, and user administration.
* **Tests**
  * Added coverage confirming restricted audit log and settings visibility for non-administrator roles.

### DIFF
--- a/backend/internal/compose/integration/auditlog_integration_test.go
+++ b/backend/internal/compose/integration/auditlog_integration_test.go
@@ -34,17 +34,16 @@ func TestAuditLogReadRequiresAdminHuman(t *testing.T) {
 	}
 
 	// An unbounded row scope is NOT the predicate. ops and read_only both seed
-	// with scope `all`, and the governance matrix reserves the compliance read
-	// for the admin alone — the read is oversight of ops' own machine-origin
-	// actions, so it cannot sit with the role it oversees.
-	for _, role := range []string{"ops", "read_only"} {
-		unboundedCtx := e.As(e.Rep1, []ids.UUID{e.Team1}, principal.Permissions{
-			RoleKeys: []string{role},
-			Objects:  map[string]principal.ObjectGrant{"person": {Read: true}},
-			RowScope: principal.RowScopeAll,
-		})
-		if _, err := privacy.ListAuditLog(unboundedCtx, e.DB(), privacy.AuditFilter{}); !errors.Is(err, apperrors.ErrPermissionDenied) {
-			t.Fatalf("%s reads audit log: err=%v, want permission denied", role, err)
+	// with scope `all` — pinned by policy's own suite — and the governance
+	// matrix reserves the compliance read for the admin alone: it is oversight
+	// of ops' own machine-origin actions, so it cannot sit with the role it
+	// oversees. Ops carries the admin object grid here because production does,
+	// which is what makes its refusal evidence about the ROLE rather than about
+	// a missing grant.
+	for _, unbounded := range []principal.Permissions{OpsPerms, ReadOnlyPerms} {
+		ctx := e.As(ids.NewV7(), []ids.UUID{e.Team1}, unbounded)
+		if _, err := privacy.ListAuditLog(ctx, e.DB(), privacy.AuditFilter{}); !errors.Is(err, apperrors.ErrPermissionDenied) {
+			t.Fatalf("%v reads audit log: err=%v, want permission denied", unbounded.RoleKeys, err)
 		}
 	}
 
--- a/backend/internal/compose/integration/auditlog_integration_test.go
+++ b/backend/internal/compose/integration/auditlog_integration_test.go
@@ -5,11 +5,12 @@
 
 package integration
 
-// The audit-log governance read (GET /audit-log, feedback/13): admin
-// (unbounded human) reads the workspace trail newest-first with live
-// filters and a stable keyset walk; a bounded rep and an agent
-// principal are refused outright — the surface never narrows to a
-// misleading partial view.
+// The audit-log governance read (GET /audit-log): an admin human reads the
+// workspace trail newest-first with live filters and a stable keyset walk.
+// Everyone else is refused outright — a bounded rep, an agent principal, and
+// the two roles that hold an unbounded row scope without holding the
+// compliance authority. The surface never narrows to a misleading partial
+// view.
 
 import (
 	"errors"
@@ -21,7 +22,7 @@ import (
 	"github.com/gradionhq/margince/backend/internal/shared/kernel/principal"
 )
 
-func TestAuditLogReadRequiresUnboundedHuman(t *testing.T) {
+func TestAuditLogReadRequiresAdminHuman(t *testing.T) {
 	e := Setup(t)
 
 	e.SeedPerson(t, "Audit Subject", nil)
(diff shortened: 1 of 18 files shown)


luit-nfq:
## Review round applied — and it found a live one

The end-of-work review turned up the same retired predicate still standing on the **GDPR case queue**, and unlike the SAR case it was live rather than latent.

**Repro before this push:** sign in as a member holding the seeded `read_only` role — the least-privileged in the matrix — and `GET /v1/data-subject-requests` returns **200** with the whole queue: who filed an Art. 15/17 request, against whom, the deadline, the status, the resolution. `ops` additionally passed the update arm and could transition cases.

I had left it deliberately, on the grounds that its authority is unreconciled upstream because it ships as a 🟢 agent tool with no admin pin. **That reasoning does not survive contact with the function** — `requireDSRAdmin` refuses agents outright, so the annotation describes a tool that can never succeed and is not evidence about who may see the queue. The queue now takes the same shape as the audit read: human, then admin. Subject-access fulfilment is admin-mediated, and reading the queue is how it is mediated.

The client had claimed this was already true. The settings comment said the queue "gates itself" while nothing did; now the inbox reads the admin predicate, disables the fetch rather than issuing a call the server would only refuse, and renders the card **withheld in its natural position** rather than absent — an absent card on a tab an ops seat reaches for the consent registry would read as "no requests", a different claim entirely.

### Also applied

- **`AssembleSAR` had no human arm.** An agent under a passport carries the granting human's live grants, so an admin's read-scoped passport would have assembled a subject's entire Art. 15 package. Nothing routes it today; the arm goes in before someone wires it. Its comment also conflated admin with unbounded — the exact confusion being removed next door.
- **The premise was unproven.** "Three seeded roles hold row scope `all`" was asserted in comments and nowhere else. `TestExactlyThreeSeededRolesAreUnbounded` now derives it from the seed, so narrowing a role tomorrow fails there rather than silently thinning what the refusal tests cover.
- The audit-log **file header** still described the predicate this PR removed and contradicted the function 70 lines below it.
- The human refusal returned a **bare sentinel** beside a wrapped one.
- **`OpsPerms` / `ReadOnlyPerms`** join the harness so suites stop hand-rolling lookalikes, with the object grid cloned rather than aliased to the admin's. The fixtures moved to their own file, which is what kept `harness.go` under the length cap.

### Left alone, on purpose

Narrowing SAR's authority beyond the human arm. Its gate is `person.delete` + unbounded, which resolves to admin ∪ ops and already excludes `read_only`; no ratified row names subject-access export, so tightening it would be the unratified move I wrongly talked myself into for the DSR queue and correctly out of here.

### Verification

- `make check-fe` — green
- `make test-integration` — **0 skips, 38 packages** (at `INTEGRATION_JOBS=3`; the default exhausts `max_connections=100` on this machine and fails unrelated packages on connect)

PR #1143, opened 2026-08-13
fix(forecast): the weighted number is computed once, per deal, on the server

### LINKED ISSUE #719: Show the weighted forecast number, the slipped bucket, and add coverage
## What is missing

Three gaps on the forecast surface, all small, all visible to a user:

1. **The weighted number is computed and not shown.** The forecast report calculates both an unweighted and a weighted total on the server; the screen renders only the raw amount. The more useful number is thrown away at the last step.
2. **The slipped bucket renders nowhere.** The backend derives a "slipped" category for deals that are committed or best-case but past their close date. The frontend's category list does not include it, so those amounts silently appear in no tile.
3. **There is no coverage view.** Pipeline against target — the "do I have enough pipeline to make the number" question — does not exist anywhere, despite the forecasting chapter claiming it.

## Why it matters

Forecast is a screen people look at weekly, and two of these are pure display bugs on numbers we already have. The third is the question sales managers actually ask.

- `backend/internal/compose/report.go` — the forecast report, including the weighted sum and the derived category expression with the slipped bucket. The maths is done.
- `frontend/src/screens/reports.tsx` — the forecast tiles (this is where the raw value is used instead of the weighted one) and the category list that is missing "slipped".
- Quotas and attainment already exist: `backend/internal/modules/quotas/`, including a deliberate refusal rather than a divide-by-zero when a target is zero. Coverage builds on that.

## What to build

1. Render the weighted number on the forecast tiles — decide whether to show both or make it switchable.
2. Add the slipped category to the frontend so those deals appear somewhere.
3. A coverage calculation (open pipeline against the target for the period) and a place to show it.

## Done when

- [ ] The weighted forecast is visible without doing arithmetic by hand.
- [ ] No deal's value disappears from every tile.
- [ ] A manager can see pipeline coverage against target.

**Size:** S for the first two, S–M with coverage.
**Spec:** `forecasting.md` (its status claims coverage exists — this closes the gap or the chapter needs correcting) and `reporting.md`, which admits the coverage surface is not built.

tinnhlt-gra:
## Summary

Closes #719, rewritten per `.tmp/capability-roadmap/PROGRESS.md`/`PLAN.md` (F2)
— it is a correctness bug, not a missing display feature.

- The board's stage columns (`deals.tsx`) and the reports screen's
  deals-by-stage table both computed the weighted figure as
  `round(Σamount × probability / 100)` — rounding the column's SUM once.
  The `forecast` report already computes `Σround(amount × probability / 100)`
  (AC-F1: round PER DEAL, then sum) — a different number, off by the rounding
  residue of every deal in the group.
- Fix: `deals-by-stage`'s reportSpec gains a `win_probability` dimension and
  a `weighted_amount_minor` measure, sharing the exact SQL expression
  `forecast` already used (one spelling, not two). The board sums each
  deal's own rounded weighted value (it already has every deal's
  `amount_minor` locally); the reports screen now requests and renders the
  server's `weighted_minor` aggregate instead of re-deriving it client-side.
- The forecast tab now requests and renders the weighted figure it always
  computed server-side but never showed, and gains a `slipped` tile — a
  real server-derived category (`forecastCategoryExpr`) that nothing
  rendered before, so those deals' amounts appeared in no tile at all.
- The forecast banner (en/de/vi) described the pre-fix screen; rewritten to
  match what the tiles now actually show.

**Out of scope, on purpose:** the original issue's third bullet (pipeline
coverage vs. target) is a distinct, much larger capability —
`rolling-coverage` (REPORT-KEY-5) and the whole S-E09.6 coverage-risk
screen in `margince-foundation`, neither of which exist and neither of
which this roadmap schedules under this wave. Filing a follow-up issue for
it rather than growing this PR.

## Test plan

- [x] TDD throughout: every change has a red-first test (1 new Go
  integration test file, 2 Vitest files touched).
- [x] `make check-backend` — green (build, vet, lint incl. strict new-code
  gate, unit tests).
- [x] `make check-fe` — green (biome, vitest, tsc, build).
- [x] `make test-integration` — green, 0 skips (an unrelated
  `internal/modules/overlay` flake on the first run reproduced 5/5 in
  isolation and the very next full run was clean — not this change).
- [x] `craft static --strict` diff-scoped — 0 blocker/major/minor.
- [x] Independent Fable review pass — 2 real findings fixed (a test fixture
  that couldn't actually discriminate the bug it was written to catch, and
  a stale banner contradicting the new UI) before push.
- [ ] Manual UAT on a live dev stack (forecast tab shows weighted + slipped;
  board and reports-table weighted columns match the report engine) —
  running now, report to follow.

---
## Summary by cubic
Make weighted totals consistent by rounding per deal on the server and using that figure everywhere. Previously the board and Deals by stage rounded the column sum once; now both read or compute Σround(amount × probability / 100), and Forecast tiles show both raw and weighted totals plus the server-derived “slipped” category.

- Backend: `deals-by-stage` joins `stage` for `win_probability` and adds a `weighted_amount_minor` measure shared with Forecast; the SQL casts `amount_minor` to numeric before multiplying to prevent bigint overflow. Drops the unused `stage_id` filter. Adds integration tests for per‑deal rounding, “Explain This Number” reconciliation, and row‑scope. Generalizes report test helpers.
- Frontend: Board sums each deal’s rounded weighted value. Reports request and render the server’s `weighted_amount_minor` for Deals by stage and Forecast; Forecast tiles include raw and weighted totals and the “slipped” bucket. Copy updated in `en`, `de`, `vi`. Tests assert aggregate requests and per‑deal rounding.
- Docs: `run_report` catalog documents `deals-by-stage` support for `win_probability` and `weighted_amount_minor`. No breaking changes; clients may use the new measure.

<sup>Written for commit f656f41b6f7c6b0f6583c44ead6d2a8eb00aeca7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1143?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/internal/compose/derivation_test.go
+++ b/backend/internal/compose/derivation_test.go
@@ -185,7 +185,7 @@ func TestForecastSpecShape(t *testing.T) {
 	if got := spec.fromClause(); got != "deal t JOIN stage s ON s.id = t.stage_id" {
 		t.Errorf("fromClause = %q", got)
 	}
-	if spec.measures["weighted_amount_minor"] != "round((t.amount_minor * s.win_probability) / 100.0)::bigint" {
+	if spec.measures["weighted_amount_minor"] != "round((t.amount_minor::numeric * s.win_probability) / 100.0)::bigint" {
 		t.Errorf("weighted measure = %q", spec.measures["weighted_amount_minor"])
 	}
 	if spec.measures["amount_minor"] != "t.amount_minor" {
--- a/backend/internal/compose/report.go
+++ b/backend/internal/compose/report.go
@@ -35,8 +35,12 @@ const (
 	whereArchivedNull = "t.archived_at IS NULL"
 
 	// joinStageForWinProbability is the one join a spec adds when it needs the
-	// deal's current stage for win_probability — a to-one lookup (deal.stage_id
-	// is NOT NULL, stage.id is its PK), so it never widens the row grain.
+	// deal's current stage for win_probability. It is safe from BOTH directions
+	// a join can go wrong: it cannot MULTIPLY rows (a to-one lookup — deal.stage_id
+	// is NOT NULL, stage.id is its PK) and it cannot DROP one under the workspace
+	// GUC either — every deal's stage_id is validated against the SAME workspace
+	// at write time (modules/deals writes inside the RLS tx), so a deal's stage row
+	// is never invisible to the deal's own scope.
 	joinStageForWinProbability = "JOIN stage s ON s.id = t.stage_id"
 	colWinProbability          = "s.win_probability"
 
@@ -44,11 +48,23 @@ const (
 	// §6, AC-F1): round PER DEAL, half away from zero, so a roll-up over it
 	// equals the sum of its own rows exactly. Shared by every spec that joins
 	// stage for win_probability, so the forecast report and any other
-	// per-stage weighted figure cannot drift apart (review-loop rule 1).
-	weightedAmountMinorExpr = "round((t.amount_minor * s.win_probability) / 100.0)::bigint"
+	// per-stage weighted figure cannot drift apart. The multiply casts to
+	// numeric first — amount_minor is an unbounded bigint, and bigint × smallint
+	// overflows before the /100.0 below would otherwise widen it.
+	weightedAmountMinorExpr = "round((t.amount_minor::numeric * s.win_probability) / 100.0)::bigint"
 	// fieldWeightedAmountMinor is the API-facing measure name every spec that
 	// defines weightedAmountMinorExpr registers it under.
 	fieldWeightedAmountMinor = "weighted_amount_minor"
+
+	// fieldStageID, fieldStatus and fieldWinProbability are report-vocabulary
+	// field NAMES (map keys) — distinct from the col* constants above, which
+	// are the SQL expressions those names resolve to. Declared here, not
+	// borrowed from an unrelated surface's vocabulary (overlay's query-param
+	// names happen to share these spellings, but renaming one must never
+	// rename the other).
+	fieldStageID        = "stage_id"
+	fieldStatus         = "status"
+	fieldWinProbability = "win_probability"
 )
 
 type reportAggregate struct {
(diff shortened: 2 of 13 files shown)


tinnhlt-gra:
## Manual UAT — live dev stack

Ran \`make dev-fresh\` + \`make seed-dev\`, logged in as the demo admin, then created three deals via the API to exercise the exact rounding-divergence case and the slipped bucket:

- Two Discovery-stage (25% win probability) deals at €1,234.50 each. `Σround(amount×p/100)` = round(30862.5)×2 = **€617.26**; the old bug, `round(Σamount×p/100)`, would have shown **€617.25** — a real one-cent divergence, confirmed first at the API (`POST /reports/deals-by-stage`) and then in the UI.
- One `commit`-category deal with no close date set, to land in the server-derived `slipped` bucket.

**Deals-by-stage table** (Reports → Deals by stage): Discovery shows weighted **€617.26** — the correct per-deal-rounded figure, not €617.25.

**Board** (Pipeline): Discovery column also shows weighted **€617.26**, same figure — board and reports table now agree because both come off the same server measure.

**Forecast tab**: the `Slipped` tile renders (previously nothing rendered it) showing €900.00 raw / €90.00 weighted, and the banner text now matches what's on screen (both raw and weighted are shown per tile).

Screenshot (Forecast tab, all five categories including Slipped):

forecast tab with slipped tile

(Screenshot captured locally during the session — happy to attach via a follow-up comment if GitHub upload is needed; the numbers above were independently verified through both the raw API response and the rendered DOM.)

Test deals archived and dev stack torn down (`make dev-stop`) afterward.

tinnhlt-gra:
Correction: the screenshot link in my previous comment was a placeholder and doesn't resolve — GitHub's CLI has no image-upload endpoint, so I can't attach the capture directly. The numbers I reported (Discovery weighted €617.26 on both the board and the deals-by-stage table, Slipped tile at €900.00/€90.00) were read directly off the rendered page during the session and cross-checked against the raw API response (`POST /reports/deals-by-stage` and `POST /reports/forecast`), included below verbatim.

deals-by-stage (Discovery, filtered):
```json
{"deals":2,"raw_minor":246900,"stage_id":"...","weighted_minor":61726}
```

forecast (grouped by forecast_category):
```json
{"deal_count":1,"forecast_category":"slipped","raw_minor":90000,"weighted_minor":9000}
```

Happy to attach an actual screenshot via the GitHub web UI if that's wanted for the record — let me know.

tinnhlt-gra:
Screenshot for the UAT comment above (Forecast tab, all five categories including the new Slipped tile):

screenshot-1786608738582-0

PR #1163, opened 2026-08-13
feat(extensions): a unit picks a record by name, not by UUID

tinnhlt-gra:
Follow-up to #1142. The filing control on `#/ext/notes` asked for a **Record id** and took a raw UUID in a text input — the one place the reference extension was worse than the product it exists to demonstrate. Nobody has a record id; they have a name.

## What it uses

`RecordPicker` already solved this: the debounced search → candidate list → pick control `offers.tsx` uses twice (buyer org, product), carrying **no transport of its own** — the caller passes `searchTargets`.

What stopped a unit using it was the published surface, which named Badge, Button, Card, EmptyState, Field, SectionHeader, TextInput and Select and nothing else. So this **widens the published frontend surface**, which that file's own doc calls a reviewed act: *"Adding a name here says the core will keep rendering it for units it did not write."*

The promise is worth making. Picking a record is the first thing any unit that writes a core record needs, and the alternative is every unit hand-rolling a combobox against a design system that already refuses a bare `<select>`. What is published is a **rendering** promise and no data one — a unit still reaches its candidates through the api surface, under the caller's own RBAC, exactly as a core screen does.

## Why `/search` and not the list endpoints

`GET /search?q=…&types=<type>` is cross-object, RBAC-scoped, and returns `{type, id, title}`.

The obvious alternative — one list endpoint per subject type — does not work, and the reason is worth recording rather than rediscovering: **`/deals` takes no `q`.** `/people`, `/organizations` and `/leads` do. A picker built on the list endpoints would have covered three of the four record types the filing accepts and left `deal` typing a UUID, which is the state this change removes.

Being RBAC-scoped matters twice: a seat is offered only what filing would actually be allowed to name, and the picker never becomes a way to learn which records exist.

## The two behaviours worth knowing

- **The search is scoped to the record type the form is on.** Unscoped, a query for "acme" while the form says Person would offer the Acme organization, and filing it would name a person id that is an organization's.
- **Changing the type clears the pick.** A contact chosen while the form said Person is not a deal; a stale pick would file against whatever record holds that id — or, far more often, against nothing, with a refusal nobody can read.

A hit whose `title` is null is dropped rather than rendered as an empty row: the contract types it nullable, and a candidate nobody can read is one nobody can choose on purpose.

## Tests

Three new cases in the unit's own suite, both new guards **mutation-verified**:

- the filing sends the id the picker resolved, and nobody types one (mutating the scoping breaks it);
- the search carries `types=<selected type>`;
- changing the type disables filing again, because the record it would have named is gone;
- a titleless hit offers no candidate — asserted against the `/search` call having happened, since "no row appeared" is also what a search that never ran looks like.

## Not in scope, deliberately

- No endpoint, no contract change, no core screen change.
- No prompt, AI task or corpus is touched, so **no aicert re-run is owed**.
- The MCP tool surface is unchanged: `file_note` still takes `subject_type` + `subject_id`, so an agent calling it is unaffected. This is the human screen only.

---
## Summary by cubic
Let users file notes by picking a record by name instead of pasting a UUID. Replaces the “Record id” input with a searchable `RecordPicker`; changing the record type clears the selected record and any visible candidates to avoid filing against the wrong record.

- Uses `/search` with `types=[subjectType]` and `limit=10`; drops hits with null `title`. Submit enables only with a note body and a picked record; request sends `subject_type` and `subject_id` from the pick.
- Fix: `RecordPicker` now empties its candidate list immediately when `searchTargets` changes; stale answers are ignored (requests are not aborted). The design-system suite pins clearing candidates and removing a failed-search message on search-space change.
- Publishes `RecordPicker` and `RecordPickerCandidate` from `frontend/src/surface/design-system.ts`; surface test updated. No API or contract changes.
- UI copy: new `extNotes.filing.subjectSearchLabel` in `en`, `de`, and `vi`.
- Tests cover id-from-pick, `types` scoping, clearing on type and search-space change, null-title filtering, and mutation variables; `/search` checks `response.ok`.

<sup>Written for commit 7e5030280011f3fcc3012fc381750a14be237707. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1163?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/extensions/notes/frontend/i18n/de.json
+++ b/extensions/notes/frontend/i18n/de.json
@@ -8,7 +8,7 @@
   "extNotes.filing.noGrant": "Ihnen wurde das Heften nicht gewährt.",
   "extNotes.filing.organization": "Unternehmen",
   "extNotes.filing.person": "Person",
-  "extNotes.filing.subjectIdLabel": "Datensatz-ID",
+  "extNotes.filing.subjectSearchLabel": "Datensatz über den Namen suchen",
   "extNotes.filing.subjectTypeLabel": "Datensatztyp",
   "extNotes.filing.title": "Notiz an einen Datensatz heften",
   "extNotes.notes.add": "Hinzufügen",
--- a/extensions/notes/frontend/i18n/de.json
+++ b/extensions/notes/frontend/i18n/de.json
@@ -8,7 +8,7 @@
   "extNotes.filing.noGrant": "Ihnen wurde das Heften nicht gewährt.",
   "extNotes.filing.organization": "Unternehmen",
   "extNotes.filing.person": "Person",
-  "extNotes.filing.subjectSearchLabel": "Datensatz über den Namen suchen",
+  "extNotes.filing.subjectSearchLabel": "Datensatz nach Namen suchen",
   "extNotes.filing.subjectTypeLabel": "Datensatztyp",
   "extNotes.filing.title": "Notiz an einen Datensatz heften",
   "extNotes.notes.add": "Hinzufügen",
--- a/extensions/notes/frontend/i18n/en.json
+++ b/extensions/notes/frontend/i18n/en.json
@@ -8,7 +8,7 @@
   "extNotes.filing.noGrant": "You have not been granted filing.",
   "extNotes.filing.organization": "Company",
   "extNotes.filing.person": "Person",
-  "extNotes.filing.subjectIdLabel": "Record id",
+  "extNotes.filing.subjectSearchLabel": "Find the record by name",
   "extNotes.filing.subjectTypeLabel": "Record type",
   "extNotes.filing.title": "File a note to a record",
   "extNotes.notes.add": "Add",
--- a/extensions/notes/frontend/i18n/vi.json
+++ b/extensions/notes/frontend/i18n/vi.json
@@ -8,7 +8,7 @@
   "extNotes.filing.noGrant": "Bạn chưa được cấp quyền đính ghi chú.",
   "extNotes.filing.organization": "Công ty",
   "extNotes.filing.person": "Người",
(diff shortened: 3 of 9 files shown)


tinnhlt-gra:
## UAT — recorded on a live stack

Captions are on the frames. Fixture: one contact *Acme Renewal Contact*, one company *Acme Industrial* — deliberately sharing a word, because that is what makes the scoping visible.

Picking a record by name, scoped to its type

1. **The record is found by name.** There is no id field on the card any more; nothing to paste into.
2. **Switching the record type drops the candidate already on screen.** This is the wrong-record path both reviewers found: the debounce cancels the *next* answer, not the one already rendered, so before the fix the person row stayed clickable under `subject_type: deal` for 250 ms plus a round trip.
3. **The same query answers with the company.** Same word, different search space — `/search?q=acme&types=organization` — so a contact cannot be filed as a company by accident.
4. **Filed**, and the form clears.

What the database says about that filing — the id came from the pick, not from anybody's clipboard:

```
audit_log
  create | ext_notes_note | via tool/file_note
  evidence.extension.detail = {"subject_type":"organization","subject_id":"019ffa85-5891-…"}

organization 019ffa85-5891-…  →  "Acme Industrial"
```

The run was made against this branch after the review fixes, on `DEV_SLUG=pick` with api and worker up.

tinnhlt-gra:
@coderabbitai review

PR #1174, opened 2026-08-13
feat(deals): the board's column totals are computed over every deal

### LINKED ISSUE #718: Compute pipeline board column totals on the server
## What is missing

The Kanban board's per-column numbers — deal count, raw value, weighted value — are calculated in the browser from the deals it happens to have loaded, and the board loads at most 100 deals with no paging. On a pipeline with more than 100 open deals, **the totals shown are wrong** and nothing tells the user.

## Why it matters

A wrong number presented confidently is worse than no number. This is a live risk for any real customer dataset and for demos loaded with imported data.

- `frontend/src/screens/deals.tsx` — the board, its column building, and the 100-deal cap (which is documented in the file). The weighted calculation itself is correct: raw value times the stage's win probability. Mixed-currency columns already refuse to sum rather than adding different currencies together — keep that behaviour.
- `backend/internal/modules/deals/` — the list and read paths to extend.

## What to build

1. A server-side aggregate per stage for the current filter: count, raw sum, weighted sum, with the same mixed-currency rule.
2. Have the board render those totals rather than adding up the loaded cards.
3. Keep the cards paged or virtualised for the column contents, now that the totals no longer depend on having every card in the browser.

## Done when

- [ ] Column totals are correct on a pipeline with several hundred deals.
- [ ] Totals respect the active filters.
- [ ] A mixed-currency column still declines to invent a total.

**Size:** S.
**Spec:** `deals-and-pipeline.md` S-E03.1 (per-column count, raw and weighted sums).

tinnhlt-gra:
## Summary

Closes #718 (roadmap F3) — blocked on #719/F2 (merged as PR #1143, same
file, same numbers), now clear.

- The Kanban board's per-column numbers (deal count, raw sum, weighted
  sum) were computed in the browser over whatever page of cards happened
  to load — an honest, documented 100-deal cap on the CARDS, but the
  totals built from that page were presented as column totals. On a
  pipeline with more open deals than that, the totals were confidently
  wrong.
- Fix: a per-stage server aggregate (`deals-by-stage`, grouped by
  `[stage_id, currency]`) supplies the totals; the board keeps its
  one-screenful card cap unchanged. A mixed-currency stage arrives as
  more than one row and still declines to sum (the existing, correct
  rule, now proven server-side instead of only over the loaded page).
- `deals-by-stage`'s filter vocabulary grows to match the board's own
  dials: `organization_id`, `partner_sourced`, `stalled` — the last two
  are boolean-valued expressions the engine's existing `expr = $n`
  filter rendering already handles with no special-casing.
- `partner_sourced` and `stalled` are now exported from `modules/deals`
  as `PartnerSourcedSQL(alias)` / `StalledSQL(alias)` instead of being
  hand-copied into `compose` — `stalled` in particular was bare, unaliased
  SQL that is safe in `deal_read.go`'s own query but ambiguous the moment
  a query joins another table sharing a column name, which
  `deals-by-stage`'s stage join is exactly. An integration test proves
  the regression (reverting the fix locally 500s with "column reference
  ... is ambiguous" rather than returning a wrong number), and a second
  test proves the Go predicate (`IsStalled`) and the SQL predicate agree.
- The board's "N deals" count now comes from the same aggregate
  (`BoardColumn.count`), not `deals.length` — the count was wrong for the
  identical reason the sums were.

## Test plan

- [x] TDD throughout: every change has a red-first test (new integration
  tests for the report vocabulary + the ambiguity regression + the
  Go/SQL agreement check; new unit tests for `buildStageTotals`/
  `buildColumns`/`PipelineBoard`'s count fallback).
- [x] `make check-backend` — green (build, vet, strict new-code lint,
  unit tests).
- [x] `make check-fe` — green (biome, vitest, tsc, build).
- [x] `make test-integration` — green, 0 skips. (Two runs hit the known
  `internal/modules/overlay` `TestSweepBackoffGatesDueOverlayConnections`
  flake under full parallel load — passes reliably in isolation; a third
  full run was clean. Same flake class already logged from the #719 PR,
  unrelated to this diff.)
- [x] `craft static --strict` diff-scoped — 0 blocker/major/minor.
- [x] Independent craft-reviewer + security-redteam pass — real findings
  fixed before push: a hand-copied `partner_sourced` predicate that only
  agreed with the module's real one by comment (now shares one spelling
  with `stalled`'s existing pattern), a half-literal/half-constant field
  vocabulary in `deals-by-stage` (finished consistently across all three
  deal-report specs), a comment claiming an agreement test that didn't
  exist (now does), and an ignored type assertion in a new test.
- [x] Manual UAT on a live dev stack: seeded 105 deals in one stage (past
  the 100-card cap) plus a mixed-currency deal, confirmed the board and
  the Reports screen's deals-by-stage table show the identical, correct
  total (€1,050.00 raw / €105.00 weighted / **105** deals — not the
  page-capped 100), confirmed the mixed-currency stage still hides its
  sum, and confirmed the three new filters (`organization_id`,
  `partner_sourced`, `stalled`) answer correctly. Recording attached
  below.

---
## Summary by cubic
Computes pipeline board column totals (count, raw, weighted) on the server over all matching deals and renders them in the board. Previously, the board summed only the loaded page of cards (100 cap) and used `deals.length` for counts; now it uses the `deals-by-stage` aggregate so totals and counts are correct on large pipelines.

- Board behavior: reads totals grouped by `stage_id` and `currency`; mixed-currency columns still hide sums; `BoardColumn.count` shows the true count and falls back to `deals.length` while totals load; the one-screenful card cap is unchanged.
- Report changes: `deals-by-stage` adds filters to match the board’s dials (`organization_id`, `partner_sourced`, `stalled`, `currency`) and accepts `currency` in `group_by`. Filters reuse `deals.PartnerSourcedSQL(alias)` and `deals.StalledSQL(alias)` to avoid ambiguous columns under the stage join; field/column names are de-duped via constants.
- Frontend: `DealsScreen` calls `/reports/deals-by-stage` with the active board filters, `buildStageTotals` shapes rows into a per-stage map, and `buildColumns` renders server totals instead of summing cards.
- Tests/docs: integration tests cover currency grouping, filter correctness, the stalled aliasing regression, and Go/SQL agreement; UI tests verify totals and count rendering; `mcp-info.json`/`.md` document the new `run_report` shapes for `deals-by-stage`.

<sup>Written for commit 9135e92407965b4a6355f9e818744fa9d09b22b2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1174?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tinnhlt-gra:
Manual UAT recording: seeded 105 deals in one stage (past the 100-card cap) plus a mixed-currency deal, walked the board then Reports -> Deals by stage. Both show 105 deals / EUR 1,050.00 / weighted EUR 105.00 -- the true total, matching the API response directly, not the page-capped 100.

board-column-totals-uat

PR #1183, opened 2026-08-13
feat(overlay): a flip refuses a projection an older mapping produced

### LINKED ISSUE #996: overlay: a flip freezes whatever projection the mirror happens to hold, with no re-projection gate
The address transform normalizes the incumbent's own property names (`address`/`state`/`zip`) onto the contract's Address member names (`line1`/`region`/`postal_code`). A mirror row written before that change still carries the incumbent spelling.

**Correction to this issue's original framing.** It claimed the read path was self-healing and scoped itself to the flip on that premise. That premise is false. `ingestSQL` rewrites a row's `fields` only when the incumbent's own baseline advances (`EXCLUDED.updated_at_baseline > overlay_mirror.updated_at_baseline`), and `Backfill` short-circuits on a converged cursor. A record the incumbent never modifies again is therefore never re-projected — not by the poller, not by re-running the backfill — so its stored payload keeps the old member names indefinitely, on the read path exactly as much as on the flip path.

**Read path: fixed.** `overlayAddress` (`backend/internal/compose/overlaywirefields.go`) now reads both spellings, the way `overlayChildRows` already read both child-row shapes and for the same stated reason. Normalization stays in the transform, so new writes remain canonical; the reader owns the compatibility. `TestOverlayAddressReadsTheIncumbentMemberSpelling` holds it. The flip reads through that same helper, so the specific defect this issue was opened for — a flipped native row with an empty street, region and postcode — is fixed too.

**What remains, and what this issue now tracks.** The flip is the one path where a stale projection becomes permanent. It writes durable native rows and freezes the mirror, and `FlipChecks.ForceFreshDone` only requires that no row is stale and that the backfill has converged. A record synced before a mapping change is neither stale nor re-backfilled, so it passes the preflight while carrying a payload the current mapping never produced. Whatever the mirror holds at flip time becomes the native row, with no second chance — the address rename is simply the first instance of that class, and every future mapping change inherits it.

Options: have the flip preflight require a re-sync watermark newer than the last mapping change, or force a full re-projection of the mirror before a flip is permitted.

Raised from the core-field reconciliation work.

### LINKED ISSUE #1038: overlay: a mapping addition reaches only records re-synced after the deploy

dungbanguyen:
The overlay mirror stores a **projection** of an incumbent record — the result of running it through a declared mapping — not the record itself. A row is re-projected only when the incumbent's own baseline advances (`ingestSQL` guards on `EXCLUDED.updated_at_baseline >`), and `Backfill` short-circuits on a converged cursor. So a record the incumbent never touches again is never re-projected, by the poller or by re-running the backfill.

That means a **mapping** change strands every already-mirrored row holding a payload the current mapping would never produce. On the read path that is survivable — readers accept both shapes. On the overlay→native flip it is permanent: the flip writes durable native rows and freezes the mirror, so whatever a row holds at flip time becomes the native row, and `ForceFreshDone` only ever asked whether rows were stale and the backfill converged. A record synced before a mapping change is neither.

The address rename that prompted #996 was the first instance. Every future mapping change inherited the class.

## What this adds

A **fingerprint of each mapping declaration**, stamped on every mirror row at ingest. Staleness is computed by comparing a row's stored fingerprint against the current one for its class — no marker pass, correct the instant new code deploys. The flip folds that into its existing stale count, so the existing `ForceFreshIncomplete` reason covers it: **no contract change, no `crm.yaml` edit**. The background sweep gains a final phase that re-fetches stale rows, yielding to watermark work so an estate under budget pressure stays *fresh* while it converges slowly.

Rows written before the column carry NULL, which reads as stale and re-projects — which is why this also closes #1038.

## Three decisions worth knowing

**The ingest guard admits a re-projection at the same baseline only.** A re-fetch of an untouched record does not advance the baseline, so the write would otherwise be refused and nothing would ever converge. But an *older* read stays refused whatever its fingerprint says: a stale page carries a stale projection too, and admitting it on a fingerprint difference let laggard sweep pages overwrite fresher rows.

**The current-fingerprint map is keyed by incumbent class, not canonical.** `activity` is backed by five declarations at once and the fingerprint hashes the source, so a canonical→one-fingerprint map is unsatisfiable — four fifths of every activity estate would be permanently stale.

**The sweep attributes a stale row by its mirror key, not its payload.** Attribution by the declaration's constants is self-defeating, because the fingerprint hashes those constants: changing one makes every row of that class stale while the filter searches for the new value and matches nothing. The OVA-MAP-7 id namespace (`calls:123`) is invariant under declaration edits.

## Bounded claim

The first deploy converges the whole estate, bounded per sweep pass, spending metered incumbent API budget proportional to the mirror. That is the intended trade: a blocked flip beats a permanently wrong one.

Closes #996, closes #1038.

---
## Summary by cubic
Blocks overlay→native flips when mirror rows were projected by older mappings and re-projects them so native rows don’t freeze outdated projections. Previously the flip ignored mapping drift; now each row stores a mapping fingerprint, preflight treats NULL/mismatch as stale, and the sweep clears them.

- Schema: add nullable `projection_fingerprint` to `overlay_mirror`; NULL or mismatch marks the row stale.
- Ingest/adapters: `hubspot` stamps `overlay.Fingerprint(m)`; the fake adapter stamps a constant. The staleness guard admits a same-baseline re-projection but still rejects older reads. Fingerprint and owner inputs are pointers to encode SQL NULL for pre-existing rows.
- Flip and sync status: `ForceFreshDone` now also requires zero stale projections; classes with no current declaration are excluded; sync status reports stale-by-class.
- Sweep: adds a post-watermark re-projection phase that selects stale rows via the shared predicate, attributes by namespaced mirror key, enqueues `overlay_refetch` through the job’s ambient River client, and respects the budget meter.
- Jobs/wiring: the re-fetch worker now builds a workspace-bound mirror store per job (not shared) to prevent panics and unmetered retries; the budget meter remains shared. Compose injects current projection fingerprints (keyed by incumbent class) into `overlay.Service`. List paging moves to `mirrorpage.go`.

**Rollout**
- Run the migration. The first deploy will re-fetch NULL/mismatched rows and may increase incumbent API usage proportional to mirror size. No config changes or `crm.yaml` edits required.

<sup>Written for commit 3170e46c3e868a4cd482bbd8601fd8708044e81b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1183?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added projection tracking to identify mirror records created by outdated mappings.
  - Automatically queues bounded re-projection and refetch work during overlay sweeps.
  - Re-projection retries stale records and coalesces duplicate jobs.
- **Bug Fixes**
  - Flip preflight and sync status now correctly flag outdated or missing projections.
  - Re-projected records recover cleanly while preserving activity attribution.
  - Improved handling for unmapped classes and shared canonical targets.
- **Tests**
  - Added comprehensive integration coverage for stale detection, re-projection, retries, and job enqueueing.

### DIFF
--- a/backend/internal/compose/capturejobs_integration_test.go
+++ b/backend/internal/compose/capturejobs_integration_test.go
@@ -20,6 +20,7 @@ import (
 	"time"
 
 	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgxpool"
 	"github.com/riverqueue/river"
 
 	"github.com/gradionhq/margince/backend/internal/platform/jobs"
@@ -77,7 +78,7 @@ func TestBackfillCompletionBuildsTheDigest(t *testing.T) {
 	// The boot pass placed its own dispatcher row, so "no dispatcher exists" is
 	// not the claim to make — "the completion added none" is. Read the count
 	// before the backfill runs and hold it to that afterwards.
-	dispatchersBefore := countJobsOfKind(ctx, t, b, CaptureDigestArgs{}.Kind())
+	dispatchersBefore := countJobsOfKind(ctx, t, b.env.Pool, CaptureDigestArgs{}.Kind())
 
 	// Now schedule the backfill; the worker pages it to done and enqueues the
 	// same-day digest off the completion edge.
@@ -98,7 +99,7 @@ func TestBackfillCompletionBuildsTheDigest(t *testing.T) {
 
 	// A dispatcher row added here would be one workspace's backfill running
 	// the digest for every workspace in the installation.
-	if after := countJobsOfKind(ctx, t, b, CaptureDigestArgs{}.Kind()); after != dispatchersBefore {
+	if after := countJobsOfKind(ctx, t, b.env.Pool, CaptureDigestArgs{}.Kind()); after != dispatchersBefore {
 		t.Errorf("capture_digest rows went %d -> %d across the backfill; a completion must enqueue the "+
 			"CHILD kind for its own workspace, never the dispatcher that fans out over the whole fleet",
 			dispatchersBefore, after)
@@ -142,10 +143,12 @@ func TestBackfillCompletionBuildsTheDigest(t *testing.T) {
 }
 
 // countJobsOfKind answers how many river_job rows of a kind exist right now.
-func countJobsOfKind(ctx context.Context, t *testing.T, b *backfillWireEnv, kind string) int {
+// Every suite resets the database before it runs, so the count is this test's
+// own inserts and nothing else.
+func countJobsOfKind(ctx context.Context, t *testing.T, pool *pgxpool.Pool, kind string) int {
 	t.Helper()
 	var n int
-	if err := b.env.Pool.QueryRow(ctx, `SELECT count(*) FROM river_job WHERE kind = $1`, kind).Scan(&n); err != nil {
+	if err := pool.QueryRow(ctx, `SELECT count(*) FROM river_job WHERE kind = $1`, kind).Scan(&n); err != nil {
 		t.Fatalf("counting %s rows: %v", kind, err)
 	}
 	return n
--- a/backend/internal/compose/integration/overlay/overlay_e2e_integration_test.go
+++ b/backend/internal/compose/integration/overlay/overlay_e2e_integration_test.go
@@ -75,6 +75,25 @@ type budgetWire struct {
 	Band     string  `json:"band"`
 }
 
+// asCurrentProjection stamps rec with the fingerprint of the declaration the
+// composed server projects incumbentClass through today (asked of compose,
+// which is the one package sanctioned to name the concrete incumbent). The
+// fake stands in for that adapter's OUTPUT, so a fixture that means "a row the
+// current mapping produced" must say so with the registry's own digest: the
+// server compares every mirror row against it, and a row no current
+// declaration accounts for reads as stale — which is the whole point of the
+// comparison, and would otherwise hold every flip fixture short of ready. A
+// fixture that means the opposite sets the field itself.
+func asCurrentProjection(t *testing.T, rec overlaymod.Record, incumbentClass string) overlaymod.Record {
+	t.Helper()
+	fingerprint, ok := compose.OverlayProjectionFingerprints()[incumbentClass]
+	if !ok {
+		t.Fatalf("no current declaration for incumbent class %q — the fixture names a class the registry never declared", incumbentClass)
+	}
+	rec.ProjectionFingerprint = fingerprint
+	return rec
+}
+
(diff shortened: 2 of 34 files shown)


dungbanguyen:
Taken in 6a62bab9 — `fingerprintArg` is now a `*string`, so the SQL NULL the column's meaning depends on is stated in the type rather than left to an untyped local. pgx maps a nil pointer to NULL, and `TestFlipChecksCountARowThatRecordsNoDeclarationAsStale` asserts the stored column is genuinely NULL (not `''`) before asserting the block, so that behaviour is pinned rather than assumed.

One note for the record: `ownerArg`, two lines above, uses the same `var x any` shape — so the original matched its immediate neighbour rather than being an oversight. I've left `ownerArg` alone because it is outside this change's subject and works correctly; the local inconsistency is deliberate rather than missed, and worth a separate pass if the pattern is unwanted generally.

Issue #1160, opened 2026-08-13
Settings UI/UX: the Phase 5 list, from a full audit of the eleven entries


luit-nfq:
Everything the settings audit found that is real UX work rather than a defect. Ordered by impact within each group. Phase 5 (the layout/UI-UX branch) is where these belong.

## Structure

- **`connections` straddles both groups.** It is the one org entry with no grant predicate, and the test proving it is named "collapses to Connections alone for a principal holding neither an org grant nor an admin role" — so a rep sees a heading reading **Organization** over exactly one row whose content is *their own* mailbox, *their own* LinkedIn import and reach. The ungating is a patch over a misfiled entry, and it is the real answer to the third-group question: the seam is not a missing group, it is one entry belonging to both. Proposal: `You → Your inboxes` (connectors, LinkedIn) and `Organization → Integrations` (provider, webhooks, HubSpot mirror + user map), each with an honest predicate.
- **The Data model object switcher reads as a page filter.** Deal/Company/Contact/Lead chips sit directly under the page h1 but scope only the custom-fields section, not Products or Offer templates below. New misread: `#/custom-fields` used to be its own page.
- **`data-model` is the heaviest page in the product** — a field list, an add form, an audit rail, a pipeline editor and **two full `ListTable` surfaces**, each with its own search/filter/columns/pagination, ~4 screens tall. And the two toolbars differ (Products has a search box, Offer templates does not). Also: "Data model" is engineer vocabulary, and a branded PDF layout is not a data model.
- **No settings page has page-level sub-copy and there is no slot for one** — eight org pages stack two to four unlike cards under a single word. `NavSection` would need a `descriptionKey`.
- **`ai` is ordered backwards**: spend → per-call trace → prices → the automations that *cause* the spend. Put the automations first and the debug trace last.
- **`general` separates the currency pair** with the company profile between base currency and the rate sheet — the header comment claims they are adjacent.
- **`people` opens on an empty invite form** when the common reason to visit is to look at somebody's role. And the concept is named three ways: the row "People & access", the card "Users & roles", the roster "Members".
- **The AI model-costs table is ~1100px with no cap** — you scroll past every model price to reach Automations.

## Naming and copy

- **Three card titles collide with the heading that contains them**: `settings.identity` = "You" inside the group heading "You"; `installationSettings.orgTitle` = "Organization" inside "Organization"; `captureSettings.title` = "Capture" on the Capture page — the merge of "two surfaces both called Capture" left one still called Capture.
[...]

Issue #1159, opened 2026-08-13
Three server surfaces in settings' territory still have no UI


luit-nfq:
Diffed all contract paths against frontend literals while auditing the settings IA. `GET /admin/job-health` was exactly this shape and now has a card on Maintenance; these three remain.

**1. `/capture/blocked-domains`** (`listBlockedDomains`, `setBlockedDomain`) — zero frontend references. "The domains refused a company, and why… so an operator can tell an automatic refusal from somebody's deliberate one." Every human role may read; admin/ops write. The spec even writes the operator story (a newsletter publisher that became a client). Its home is obvious: Capture, as a third card beside the consumer-mail list. Highest-value of the three.

**2. Role→object grants are editable only for extension objects.** `GET /roles` returns every role with every object grant and `PATCH /roles/{key}/objects/{object}` sets any of them — a fully generic RBAC matrix. The only UI renders it per composed extension unit (`extension-access.tsx`). Meanwhile the settings IA gates eleven entries on `useHoldsWriteGrant(...)` for eight objects. So an admin can observe which nav rows a grant opens and cannot edit a single one of those grants in the product. Promote the matrix to a first-class card on People & access, with extension units as one section inside it.

**3. The overlay→native flip is curl-only** (`preflightOverlayFlip`, `executeOverlayFlip`). The one irreversible migration in the product, with a dedicated dry-run endpoint built for a human to read, has no surface — while the *connect* half renders on Connections. It belongs under `OverlayCard`, or on Maintenance beside the danger zone, which is already ordered by ascending consequence.

Also worth noting from the same sweep: `GET /passports` returns the whole workspace's passports to an admin (deliberate, per `identity/passport_list.go`), but the contract summarises the operation as "the caller's own … the passports the caller has minted". Contract-vs-implementation drift in the over-disclosure direction — worth reconciling upstream.

luit-nfq:
The **blocked-domains** surface (a) is built in #1835. The other two named here are untouched, so this stays open.

Notes from doing (a), in case they help whoever takes the rest:

- It follows `OwnDomainsCard` / `ConsumerMailDomainsCard` rather than introducing a third spelling of a domain-list card, and it sits last on the capture tab because it is the *outcome* of the three rules above it.
[...]

Issue #1135, opened 2026-08-13
settings.test.tsx is 2057 lines — split it along its three subjects


luit-nfq:
The settings suite is well past the 1000-line ceiling the craftsmanship rules cite for test files. It was already ~1.8× over before the information-architecture merge, which grew it a further ~270 lines.

The file now covers three separable subjects, and the cut follows them:

- **entry visibility** — the whole-list `navTabs()` contract, the per-role and per-grant expectations, the group headings, the fallback for an unknown or forbidden id;
- **agent credentials** — the passport card, revoke, the tool console and its passport scoping;
- **the audit log** — the filter row, the entries card, the diff expansion.

Nothing about the current file is wrong; it has simply stopped being navigable, which is the trigger the rules name. Splitting it is mechanical and was deliberately kept out of the IA merge, which was already a large diff.

Worth doing while the IA is fresh, since the per-seat expectations in the visibility half are the part a future change is most likely to need to read.

dungbanguyen:
picking this up

Issue #1014, opened 2026-08-12
Card and Panel are two primitives for one job


joshbruh-lang:
\`Card\` (design-system/atoms.tsx) and \`Panel\` (design-system/panel.tsx) both answer "a card that carries a header". They arrived in parallel — Card gained title/sub/actions in #997, Panel was added for the company record's titled surfaces at the same time, and neither knew about the other. A caller building a titled card today has to pick, and nothing says which.

What each has that the other lacks: Card has an \`as\` element choice, \`inset\`, a \`sub\` line, a live-region mode, and style/id/role/testId passthroughs. Panel has a fixed-height header (a title alone and a title with a button read the same height), \`PanelRow\` children that run full-bleed against the panel's own edges, and \`PanelBody\` for padded content so one panel can hold both shapes as siblings.

The full-bleed row is the part Card has no answer for and the reason Panel exists: a list inside a card wants to touch the card's edges and highlight across its full width, which a padded body cannot do.

Deferred deliberately rather than resolved inside the company-record work, so that change did not rewrite a design-system component that had just landed from another author. Whoever takes this decides whether the row and footer shapes fold into Card as further props, or the two stay distinct with the README naming which is for which — then migrates the callers.

dungbanguyen:
## Assigned to @luit-nfq for the decision, 2026-08-31

Re-triaged against `origin/main` at `8ecb691f1` during the
`status: needs-decision` sweep. **Both primitives still exist and both are in
use**, so the choice this ticket describes is still one every caller has to make.

Confirmed unchanged:

- `frontend/src/design-system/atoms.tsx` — `Card`, with the `as` element choice,
  `inset`, the `sub` line, the live-region mode and the style/id/role/testId
  passthroughs.
- `frontend/src/design-system/panel.tsx` — `Panel`, with the fixed-height header
  (a title alone and a title-with-a-button read the same height), `PanelRow`
  running full-bleed against the panel's own edges, and `PanelBody` for padded
  content so one panel holds both shapes as siblings.

The full-bleed row remains the thing `Card` has no answer for, and the reason
`Panel` exists.

### The three options, so the decision is a choice rather than a design session

1. **One primitive — fold `Panel`'s shapes into `Card`.** Add the fixed-height
   header, the full-bleed row and the padded body as further props; migrate the
   callers; delete `Panel`. Card is the atom the design-system catalog opens with
   and already carries the passthroughs. Cost is the migration, and it only grows.
[...]

PR #1185, opened 2026-08-13
feat(settings): opening a page is a read, and a withheld surface says so

### LINKED ISSUE #1157: Settings: four cards vanish on a permission denial where their neighbours say why
Found auditing the eleven-entry settings surface. The tree currently holds two contradictory doctrines about a withheld surface, and after the merge both now appear **on the same page**.

`STATE-4a` decides hide-vs-disable-vs-withheld **by cause**: absent because it does not apply, disabled because a precondition the reader can fix is unmet, withheld because a permission denies it. A withheld surface says so.

Stating it correctly: `JobHealthCard`, `PrivacyInboxCard`, `ExtensionAccessCard`, `UsersAdminCard`, `CaptureSettingsCard`.

Returning `null` on a denial:

- `frontend/src/screens/retention.tsx` — `if (!canRead) return null`, on Privacy & audit **beside** `PrivacyInboxCard`, which renders "Only an admin can see subject requests".
- `frontend/src/screens/embedreindex.tsx` — same shape, on Maintenance **beside** `JobHealthCard`, which renders its withheld body.

So for an ops seat the Privacy page now shows all three behaviours at once: the consent registry works, retention vanishes, the subject queue explains itself, and (until the fix on this branch) the audit log showed a red failure. Making that one page internally consistent is the highest-value single change.

Section-level vanish-on-denial, same violation expressed as a conditional render: the AI spend cards (`settings.tsx`), `overlay-health.tsx` action row, `rates.tsx` write affordances, `installation-settings.tsx` (fields disabled with **no reason**, while `capture-settings.tsx` passes `Switch`'s `reason` — and its own comment claims parity).

Four of those are test-pinned as intended hiding, so this needs a decision rather than a sweep: pick one doctrine, write it down, and make the pinning tests agree.

### LINKED ISSUE #1158: Settings: a read-only seat is hidden from pages the server serves it 200 on
Measured against the live API, six principals (admin, manager, rep, read_only, ops, and a manager on a `read` seat), one cookie jar each.

**No over-admission anywhere** — every entry the client shows has at least one card the server answers 200 on, so no seat lands on a page of pure refusals. Every disagreement runs the other way.

| entry | seat | client | server |
|---|---|---|---|
| `general` | read_only | hidden | `GET /installation/settings` 200, `GET /company` 200 (only `/fx-rates` 403) |
| `capture` | read_only | hidden | all four capture reads 200 |
| `data-model` | read_only | hidden | custom-fields, pipelines, stages, products, offer-templates all 200 |
| `privacy` | manager, rep | hidden | `GET /consent-purposes` 200 — and `POST /data-subject-requests` **201** |
| `people` | every non-admin | hidden | `GET /users` 200 (rest correctly 403) |

The `read_only` row is the systematic one: entry predicates are built from **write** grants, because they were written to answer "can you *use* this". A read-only seat holds no writes, so eight of eleven entries disappear — including three surfaces it could open as routes of their own before the merge (`#/products`, `#/offer-templates`, `#/custom-fields` were all ungated).

The `privacy` row is the sharpest: the server executes a DSR filing (201) that the client refuses at two levels.

This is a real IA decision, not a bug to sweep: **should a seat that can only read reach a configuration page read-only, or is a config page a place you go to change something?** Whichever way it goes, it should be one rule applied to all eleven, written down where `useOrgTabVisibility` can cite it — today it is an accident of each predicate being a write grant.

Note the seat-folding half of this was fixed on `feat/settings-ia`: opening a page is a read, so no entry predicate may consult the licensing seat (`capture` did, via `useCanWrite`).

luit-nfq:
Phase 5 of the settings restructure — layout, UI/UX and global chrome — plus the settings-tagged issues in the tracker. Closes #1157, closes #1158. Substantially advances #1160.

## Two product decisions

**#1158 — entry predicates asked for WRITE grants.** Each was written to answer "can you *use* this", and measured against the live API that hid a read-only seat from **eight of eleven entries the server answers 200 on**, including three surfaces that were ungated routes of their own before the merge. Opening a page is a read, so every predicate asks for one now, and the write affordances inside say for themselves who may use them.

The predicates were rebuilt from the **seeded grant matrix read straight out of a seeded database** rather than guessed:

```sql
select key, permissions from role;   -- unrolled with jsonb_each(permissions->'objects')
```

| object | admin | manager | ops | read_only | rep |
|---|---|---|---|---|---|
| automation | rcud | r--- | rcud | r--- | r--- |
| custom_field | rcud | r--- | rcud | r--- | r--- |
| embedding_reindex | r-u- | ---- | r-u- | ---- | ---- |
| fx_rate | rcu- | ---- | rcu- | ---- | ---- |
| installation_settings | r-u- | r--- | r-u- | r--- | r--- |
| overlay_connection | rcud | r--- | rcud | r--- | r--- |
| retention_policy | rcud | ---- | rcud | ---- | ---- |

**The consequence, stated because it is the cost of the decision**: on a fresh install every role now reaches eleven of twelve entries and only Maintenance narrows. The nav no longer describes the seat — what distinguishes seats has moved *inside* the pages. That is exactly why #1157 had to land in the same PR.

**#1160's structural headline — `connections` split in two.** One Organization row held both a reader's own mailbox and LinkedIn network *and* the installation's contact-data credential, webhooks and HubSpot mirror. That is **why** it had no predicate: any honest gate took a personal task away from whoever it hid it from.

```
You                    Organization
  Account                General
  Writing voice          People & access
  Agents                 Integrations   ← provider, webhooks,
  Connections   ←          Capture        HubSpot, user map
    connectors             Data model
    LinkedIn               AI · Privacy & audit · Maintenance
```

The entry with no predicate **disappeared rather than moving**. Keeping the `connections` id for the personal half was load-bearing: the OAuth callback route in `internal/compose/connectors_outcome.go`, the compose "connect a mailbox" link and the home link all still resolve. Only the system-of-record chip followed the mirror.

## #1157 — the doctrine is now written down

`STATE-4a` **appears nowhere in this repo** — it lives only in the spec, so every screen re-derived it. It is in `frontend/src/design-system/README.md` now: absent / disabled / withheld, decided by **cause**.

Two cards returned `null` on a permission denial beside neighbours on the same page that explained themselves, so for an ops seat the Privacy page showed the registry working, retention silently gone and the subject queue saying why. An absent retention card does not read as "not yours" — it reads as "this installation keeps nothing". Both state the denial now and **still ask the server for nothing**, because the answer was already known.

## Also fixed

- **The scroll position survived a route change** — reproduced and then verified fixed in a browser: 1831px carried into the next page before, 0 after.
- **No skip link existed anywhere** (WCAG 2.4.1). A button, not `<a href="#content">`, because the app is hash-routed and a fragment link would navigate. Verified live: first tab stop, reveals on focus, moves focus to `main#content`.
- **Entering the settings level dropped focus to `<body>`** — the walk out was pinned, the walk in was not, and the guard meant to arm it fires only for a nav row with children, which nothing in the tree produces. Two comments claimed both directions were covered.
- **Modal focus restore ran against openers the mutation had removed** — passport revoke, member deactivate, connection end, DSR transition all landed on `<body>`.
- **Every card's load failure was announced to nobody** — one fix at `QueryStates` covers most surfaces, plus the shared create/edit and archive errors.
- **The tool console dimmed unreachable rows to `opacity: 0.4`**, taking the "scope not granted" caption that is meant to be the text equivalent below the AA floor — while the passport list 150 lines above it had chosen a strikethrough for exactly that reason. Every credit-pool meter was also announced as an anonymous number triple (`label=""`).
- **Twenty-one German strings addressed the reader as *Sie*** against `de.ts`'s own mandated informal *du*, every one on a Connections surface — the register flipped exactly where a rep does their own personal setup.
- Retention wrote on flip through a `Checkbox` and collapsed its own edit panel; the member roster was hidden from seats the server serves it to; three pages were ordered against the reader; headings were styled spans; passport scopes rendered raw protocol words.

## A measurement worth reading

The collapsed rail cannot scroll below ~670px of viewport height, and the obvious fix — `overflow-x: clip` + `overflow-clip-margin` — is **silently wrong**. Chrome computes `clip` to `hidden` as soon as the other axis scrolls, so the clip margin is ignored and every tooltip stops painting, **while the tooltip's rect still reports as escaping the box**. A geometry assertion would have called it fixed; a paint hit-test (`elementFromPoint`) shows the pixel is not the tooltip. Reverted, with the full disproof recorded on #1160.

## Verification

- `make check-backend` → **0** (no Go changed; run after rebasing onto the `/imports` merge, which the breaking-change gate reads as a removal when a branch is behind)
- `make check-fe` → **0**, 199 files / 2488 tests
- `pnpm e2e` → **75 passed, 0 failed**

Adding `data-model`, `integrations` and `people` to the axe and 390px sweeps — two settings pages of twelve were covered, both short ones — immediately found a real pre-existing AA violation: `.cf-hint` sets 11.5px prose in `--textTertiary` (~2.2:1 on white) where `tokens.css` names `--textMeta` the canonical AA small-text role.

## Deliberately deferred

#1135 (splitting the 2000-line settings suite — this diff is already large), #1159's three missing UIs, #1014 (Card vs Panel), a `NavSection` page sub-copy slot, and the data-model page's four-surface weight.

---
## Summary by cubic
Settings entries now open on READ, and pages you can’t write to render as withheld with a stated cause. Previously entries gated on WRITE, which hid read-only seats from pages they could safely read; now all roles reach 11/12 entries (Maintenance still restricted), with write controls gated in-page. Connections is split: per-user mail/network stay in Connections; installation-wide wiring (providers, webhooks, mirror) moves to Integrations; the system‑of‑record chip links to `#/settings/integrations`. Closes #1157 and #1158; advances #1160.

- Privacy & audit entry requires `person:read`; provider Connect/Disconnect/Delete, company profile Save, and Products/Offer templates New/Edit/Archive are gated by their write verbs. FX and AI model rate sheets gate the status read, show read‑only posture when write is withheld, and don’t fire list queries without a read grant. AI spend/call trace and overlay actions keep their place and explain withheld; auto‑enrich is a disabled Switch with a reason.
- Accessibility and UX: add a Skip to content control, restore focus when a modal removes its opener, announce loading via a status region and failures via role=alert, adjust heading levels, and fix small‑text contrast on data‑model hints.
- Navigation and links: overlay chip targets `#/settings/integrations`; Home’s connector‑health deep link points to `#/settings/connections`; add a link from Own domains to `#/settings/general`. Remove `useHoldsWriteGrant`; add a nav walk claim to focus the arriving settings level.
- Tests and seeds: expand e2e/axe across `settings/data-model`, `settings/integrations`, `settings/people`; add coverage for focus restore, withheld/read‑only states, dialog semantics; seed `person:read` for Privacy.

**Review and rollout**
- Expect broader settings nav visibility for non‑admins; verify per‑page writes still 403 and withheld/read‑only surfaces state their posture.
- Update docs and deep links: installation‑wide settings are under Integrations; personal setup remains under Connections; ensure chips/links point at `#/settings/integrations` or `#/settings/connections` as noted.
- Spot‑check keyboard and screen‑reader flows: skip link, focus on entering settings, modal focus after destructive actions, loading/alert announcements, and withheld messaging.

<sup>Written for commit b0b3a80070bb70666487a05c3e6941df901188fd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1185?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


luit-nfq:
### Review round: the read-grant flip exposed six real defects, all fixed

CodeRabbit reported itself **rate limited** on this PR — it passed its check without reviewing — so I ran an adversarial pass over the diff myself before merging. It found things the green gates could not, and the branch is three commits longer for it.

**Two pages rendered a permanent 403 with a dead Retry.** Both rate sheets asked only a WRITE question (`useCanUpsert`) and then fired their list query unconditionally; `fx_rate` and `ai_model_rate` are admin/ops-only. So a `read_only` seat opening **General** met a red failure — directly beneath the read-only caption this branch had just added above it, announced assertively by the `role="alert"` this branch had just added to it. Two contradictory explanations of one denial, stacked, plus an announcement on arrival. Both cards now gate the read, ask the server for nothing when it is denied, and say they are withheld.

**Three surfaces rendered controls the server refuses.** `ProviderCard` offered Connect, Disconnect and **Delete data** with no capability call anywhere in the file — and a comment I wrote in this branch endorsed that in prose, which is the house rule against rationalizing a gap instead of closing it. The company profile's Save and New/Edit/Archive on both list surfaces were ungated the same way. All newly reachable by `read_only`, because the predicates opened those entries. (Connect asks for `create` alone rather than an upsert: `Store.Connect` gates on `create` even for a replacement key.)

**`privacy: true` was a stand-in for a grant that exists.** `ListPurposes` demands `person:read`, not a role — so an installation that edited that grant away would have been handed a page of four refusals with the nav still offering it. The predicate asks for it now, which is the rule stated forty lines above it in the same file. The e2e seed's own comment predicted the catch: *"a spec that reaches a surface this list forgot fails loudly here"* — it forgot `person`.

**Four surfaces dropped every write affordance in silence**, where the doctrine permits that only when the surface states its read-only posture once: the consent registry, pipelines, webhooks, and the overlay action row. The **AI spend and call trace vanished** on a denial — a claim that the installation spent nothing and made no model calls — and now keep their place and say whose figures they are. Auto-enrich turned out to be the *stateful* case, the only place a reader sees whether enrichment is on, so it became a disabled `Switch` carrying its `reason`; the README now names that case, and the pure-action case the danger zone occupies.

**Fourteen comments in this branch described the old behaviour**, including one that called absence "withheld" in the same commit that defined the difference. Also fixed: a story fixture granting writes without reads (so it captured the Account fallback — nothing asserts on a story), `useHoldsWriteGrant` left with no caller, an `id` with no reader, and nine German strings still addressing the reader as *Sie* on surfaces the split touched.

Two judgements from the fixes, both deliberate: `NotConfiguredState` beats the webhooks posture line when delivery is off, because that cause withholds those controls from an admin too; and `overlay.adminOnly` was generalised from "connect HubSpot" rather than reused as-is, because it now also labels a withheld *Sync/Disconnect* row.

**Verification after the round:** `make check-backend` 0, `make check-fe` 0 (201 files), `pnpm e2e` 75 passed / 0 failed, all six DS gate scripts PASS, rebased onto current `origin/main`.

Issue #1190, opened 2026-08-13
license: check the bundled entitlement token offline, then enforce the seat grant


lstrojny:
An installation currently has no way to prove its entitlement. `margince-constellation`
publishes a license-validation WebAssembly module that checks a signed license token
against a keyset it embeds, offline, with no callout. This issue tracks bundling that
module here and using it: first the check, then the seat grant it carries.

This realizes the offline-entitlement path the spec already asks for (B-EP08.21;
UC-E11-05 E1 "air-gapped entitlement"; UC-ADMIN-03 E3). ADR-0029 today says entitlement
rides the release service and does not yet name an offline module — that amendment is a
follow-up, filed when slice 1 lands, not a blocker for it.

## The upstream contract

`host.Check(ctx, moduleBytes, issuer, product, generation, token) -> (grants, error)`,
about 110 lines over one dependency (`github.com/tetratelabs/wazero`, pure Go, no cgo).

- The host gunzips the module itself, so the published `.gz` is embedded as-is.
- The issuer is derived per environment upstream and never configured; a shipped product
  pins the production issuer.
- Product is `margince`, generation `0` — the only generation issued today, and the
  module rejects a mismatch.
- Grants are free-form int/bool attributes. `seats` is the one this product acts on.
- The 90-day grace past expiry is baked into the module, not a local choice.

Two constraints shape everything below:

1. `margince-constellation` is private and this repo is public. The host package cannot
   be imported — a source install could never resolve the module path — so it is vendored
   as source, and the module cannot be fetched at build time either.
2. The published module trusts only the production keyset. Upstream gets a passing-check
   path by compiling with `-tags licensetest`; bundling only the release module means we
   can test rejection paths and not acceptance.

### Vendor the module, verifiably

`internal/platform/licensecheck/module/` holds the committed `licensecheck.wasm.gz`
(1.66 MB), the pinned `sha-<commit>` release tag it came from, and its SHA-256.
`make license-module` downloads that pinned tag, verifies the asset against the release's
`SHA256SUMS`, and rewrites both files — so a refresh is a reviewable diff naming the
upstream commit. A test asserts the committed blob still hashes to the committed digest,
so a corrupted or hand-swapped blob fails `make check` instead of failing at boot on an
operator's machine.

Package named `licensecheck` after its upstream counterpart, and to stay clear of
`backend/license_test.go`, which is the SPDX-header gate.

### The host

Upstream's host file copied in under this repo's BUSL header, with a comment naming the
[...]

PR #1191, opened 2026-08-13
feat(license): an installation proves its entitlement offline, before it serves

lstrojny:
Slice 1 of #1190: the entitlement check. Seat enforcement is deliberately **not** here — slice 2 needs two decisions first, listed at the bottom.

## What this does

The license-validation WebAssembly module `margince-constellation` publishes is bundled at `backend/internal/platform/licensecheck/module/` and run in-process through wazero. The module embeds the public keyset it trusts and verifies the operator's token against it with **no callout of any kind**, so an air-gapped installation proves its entitlement exactly the way a connected one does (the path UC-E11-05 E1 / B-EP08.21 already asked for).

Issuer, product and generation are pinned constants, not settings: the bundled module trusts only the production keyset, so a token from any other issuer could never verify whatever this side passed.

## Two constraints that decided the shape

Constellation is **private** and this repository is **public**. So:

- The upstream host package cannot be imported — a source installation could not resolve a private module path — and it is vendored (~110 lines, `host.go`, kept in step by hand, naming the upstream commit).
- The module cannot be fetched at build time either, because no source install and none of this repo's CI jobs hold constellation credentials. The 1.66 MB blob is committed, and `make license-module` is the only way it is replaced.

The pin is an immutable `sha-<commit>` tag, never rolling `latest`. The download is verified against the digest **GitHub itself reports** for the stored asset — the release's own `SHA256SUMS` covers the four CLI binaries and *not* the wasm module, which is worth asking upstream to fix. The recorded digest is then held to the blob by `TestBundledModuleMatchesItsRecordedDigest`, so a swapped or truncated module fails `make check` rather than a boot.

## Posture, and what each one does

| posture | boot | reported as |
|---|---|---|
| no token configured | boots, with a warning naming the bundled module | `margince_license_posture{state="absent"} 1` |
| token verified | boots | `state="valid"`, plus `margince_license_seats` when the license grants a count |
| token refused | **refuses to boot**, api and worker alike | — |

An unlicensed installation is a supported state — every dev and CI process here runs in it, and P7 forbids a silent lockout. A token that is present and refused is an operator mistake nobody downstream can distinguish from a deliberate downgrade, so it stops the boot. A module that cannot **run** at all is refused the same way: a validation module the build cannot execute is a packaging fault. A configured `token_file` that is missing or empty is likewise a boot error — that path and "unlicensed" are indistinguishable downstream, so a typo must not read as a decision.

Nothing stops a process that is already serving. The api re-checks daily and its `/metrics` posture degrades; a licensing edge case must not take a customer's CRM offline mid-month with no human in the loop.

## Verified by running it

- Both roles log the posture with the module tag: `msg="no license configured; the installation is running unlicensed" state=absent module=sha-9e9c638b8d03`
- `/metrics` serves the section live (three-series enum, seat gauge omitted rather than zeroed when nothing is granted).
- A bad token refuses the boot after the installation binds, carrying the module's own reason **and** the setting to correct: `api: licensecheck: the license was refused by the bundled validation module (sha-9e9c638b8d03): license rejected: … — correct or remove license.token_file (or MARGINCE_LICENSE) and start again`

## Note for the reviewer

`cmd/api`'s `run` was already at the 80-line craft ceiling, so this wiring adds **zero** lines to it: entitlement folded into `bindInstallation` (which settles what the installation *is* — the same question at the same moment, and both refuse the boot on an answer this build cannot honor) and the option into `baseComposeOptions`, beside the other always-applied options.

## Known gaps, both tracked on #1190

1. **The success path is unproven by tests.** The published module trusts only the production keyset, so no token this repository can mint is ever accepted; the tests cover the rejection paths, the host plumbing, and the posture state machine. Upstream publishing its `-tags licensetest` build would close it, and that should land before seat enforcement is built on that read.
2. **Slice 2 needs two decisions.** There is no right error sentinel for a seat-ceiling refusal (`apperrors` is a fixed registry: `ErrBudgetExceeded` maps to 429, `ErrSeatTierInsufficient` means "your seat is too low"), so it is a spec-first addition or `ErrConflict`. And what an *unlicensed* installation is entitled to is business policy (A36's free ten, or uncapped), not an implementation detail.

ADR-0029 still says entitlement rides the release service and does not know about a bundled module; that amendment is on #1190 as well.

---
## Summary by cubic
Verify an installation’s license offline before either role serves, using a bundled WebAssembly validator. Previously API and worker always started; now absent or valid licenses boot, while a rejected license blocks startup in both roles. Running processes are not stopped; the API rechecks daily and keeps the last good posture on module faults or panics.

- Boot wiring: API resolves entitlement during installation bind and threads posture via `compose.WithLicensePosture`; worker calls `compose.EnsureLicense` before opening a DB pool.
- Runtime and module: `backend/internal/platform/licensecheck/` runs a bundled validator via `wazero`; vendored host (`host.go`) autodetects raw, gzip, and brotli. Artifact lives at `backend/internal/platform/licensecheck/module/` with pinned `VERSION` and `.sha256`; a test holds the digest to the blob.
- Metrics and logs: `/metrics` exposes `margince_license_posture{state="valid|absent|rejected"}` and `margince_license_seats` when granted. Module faults are not reported as rejections. Rejection reasons are single-line and size-bounded to prevent log forgery.
- Config and precedence: new `license:` block with `token_file`. `MARGINCE_LICENSE` overrides only when non-empty; empty falls through to the file. An unreadable, empty, or oversized (>64 KiB) `token_file` blocks boot; no token boots with a warning. Boot logs name the token source.
- Watcher behavior: daily recheck re-reads the token, survives panics, logs faults, and preserves the last good posture; the API is the single RUNNING posture reporter.
- Module refresh: `make license-module` pins an immutable `sha-<commit>` tag, resolves the tag to its commit, verifies the download against GitHub’s per-asset digest, rejects releases publishing multiple framings, and rewrites `VERSION` and `.sha256`. No rolling tags. `.go-arch-lint.yml` allows `wazero` and `brotli`.
- Tests: isolate from a real `MARGINCE_LICENSE`, cover gzip/brotli framing, token precedence, file bounds, and the posture state machine. Success path awaits a test-authority module (tracked in #1190).

- Rollout: no action needed to keep running unlicensed. To license, place the token in a file and set `license.token_file`, or set a non-empty `MARGINCE_LICENSE`. A rejected token or bad `token_file` prevents both API and worker from booting. Seat enforcement is out of scope for this PR.

<sup>Written for commit 8fd2ff9f9d33f5f918b106e5230273dcccb03ae9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1191?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added optional offline license validation using a configured token file or `MARGINCE_LICENSE`.
  * Unlicensed installations continue to operate by default; rejected configured licenses prevent API and worker startup.
  * Added daily license rechecks to detect entitlement changes during runtime.
  * Added license posture and optional seat-limit metrics.

* **Documentation**
  * Added configuration guidance, environment-variable details, startup behavior, runtime checks, and license-module maintenance instructions.

* **Chores**
  * Added safeguards to keep local license credentials and validation assets out of version control.


lstrojny:
Let’s move the refresh make target into constellation to remove traces of the constellation repo

lstrojny:
Let’s also add a simple license screen to the admin area that show the number of seats + the number of used seats. Let’s also add an error bar when the license limit is hit

lstrojny:
Compression changed to brotli, see https://github.com/gradionhq/margince-constellation/pull/290

Issue #1188, opened 2026-08-13
listCustomFields' object query enum has no "project" entry


tinnhlt-gra:
## What's missing

\`GET /custom-fields\`'s \`object\` query parameter enum (\`crm.yaml\`) is:

\`\`\`
enum: [person, organization, deal, lead, activity]
\`\`\`

There is no \`project\` entry, even though \`customfields.FieldObjects\` (backend) includes \`project\` as a real target for custom fields — a project-owned custom field can be created and populated, but this listing endpoint cannot enumerate it by object.

## Where this surfaced

While building #706 (the \`date_field_approaching\` automation trigger), the frontend's date-field picker (\`frontend/src/screens/automations.datefield.tsx\`) needs to list a workspace's active date-typed custom fields for whichever object a \`renewal_reminder\` instance targets. \`renewal_reminder\`'s own \`object\` vocabulary (\`renewalReminderObjects\`, mirroring \`customfields.FieldObjects\`) includes \`project\`, but the picker has no way to query \`project\`'s custom fields through this endpoint — it degrades to a disabled, hinted state for that one object rather than crashing, which is documented in the picker's own comment, but the underlying gap is real.

## Impact

A workspace cannot configure a \`renewal_reminder\` (or any future feature reusing this listing endpoint) against a project's own custom date field through the UI, even though the field itself is fully supported server-side.

## Fix

Add \`project\` to \`listCustomFields\`'s \`object\` query enum in \`crm.yaml\`, and confirm the handler's implementation already handles it (it likely does, since \`ActiveColumns\` is object-generic) — this is a contract widening, not new engine work.

tinnhlt-gra:
Folded into #1520, which reports the same defect one layer wider: the contracts custom-field `object` enums are behind `customfields.FieldObjects`, which admits `project` and excludes `activity`. #1520 covers `CustomField.object` and `CreateCustomFieldRequest.object`; this issue covers the `listCustomFields` query parameter. All three are one contract change against one authority, and #1520 proposes the parity gate that keeps them together.

The scope statement on #1520 has been updated to include this query enum explicitly, and the `renewal_reminder` date-field picker (`automations.datefield.tsx`) has been carried across as the concrete blocked consumer. Closing; track it on #1520.

PR #1192, opened 2026-08-13
feat(automation): wire the date-field trigger, including yearly recurrence

### LINKED ISSUE #706: Wire the date-field automation trigger, including yearly recurrence
## What is missing

The automation trigger "a date field is approaching" is registered in the catalogue and **can never fire**. Nothing feeds it candidates. The code says so plainly: the scan's handler map contains only two activity-based handlers, anything else is skipped, and the preview is stubbed out with the same reason.

Two further gaps behind it: the automation cannot say *which* date field it watches (the one catalogue entry hardcodes a single knob), and there is no yearly recurrence — the current model anchors on one absolute date, so a birthday would fire once and never again.

## Why it matters

"Remind me before a birthday / a contract renewal / a licence expiry" is a small, obvious, frequently requested feature. Right now the option appears in the product and silently does nothing, which is worse than not offering it.

- The trigger kind and the closed catalogue: `backend/internal/modules/automation/catalog_triggers.go`.
- The clock scan that drives time-based automations: `backend/internal/modules/automation/timescan.go` — read its comments, they describe exactly this gap.
- The clock handlers, including the one catalogue automation on this trigger: `backend/internal/modules/automation/handlers_clock.go`.
- Preview: `backend/internal/modules/automation/automations_preview.go`.
- Date custom fields are real indexed columns, and the field-catalog seam can enumerate them.

## What to build

1. A candidate source for the clock scan that finds records whose chosen date field falls within N days, across the record types that support custom fields.
2. Register it in the scan's handler map so it actually runs.
3. A parameter letting the user pick which date field to watch.
4. Yearly recurrence, with an idempotency key that includes the year so it fires once per year rather than once ever.
5. Un-stub the preview so the user can see who would be matched before enabling it.

## Done when

- [ ] A user creates a birthday reminder on a date field, and it fires this year and again next year.
- [ ] The preview shows the matching records before the automation is switched on.
- [ ] Nothing fires twice for the same record in the same year.

**Size:** S.
**Spec:** `automation.md` AUTO-PARAM-1. Note the pin currently scopes this trigger to close and renewal dates, so widening it to any date field may need a wording change.

tinnhlt-gra:
## Summary

Closes #706 (plan item C3, `.tmp/capability-roadmap/PLAN.md` wave 2). `renewal_reminder` was fully built (Match/Plan/IdempotencyKey, tested) but had no candidate source — the code's own doc explained why: no seam reached an arbitrary workspace custom field's value. This PR wires it end to end.

- **Candidate seam**: `automation.DateFieldScan`, mirroring the existing `ActivityScan` seam — declared in `automation` with only `ids`/`datasource`/stdlib types, implemented by `customfields.Service.DateFieldCandidates` (validates `(object, column)` against the workspace's own active-field catalog before any identifier reaches SQL — literal `BETWEEN` for a one-time field, MMDD-based matching with year-boundary wraparound for `recurs_yearly`), wired via a compose adapter.
- **Recurrence**: a recurring field's occurrence date is projected onto the current scan window's year entirely inside the candidate source, so `renewalReminder`'s existing `Match`/`Plan`/`IdempotencyKey` needed **zero changes** — proven directly by a fixed-clock, two-simulated-year test.
- **Params**: `object`, `date_field`, `recurs_yearly` join the existing `days_before` on the catalog schema (with a closed `enum` for `object`).
- **Preview**: `renewal_reminder` gets a real dry-run (built dynamically per-instance, since its table/column are workspace-configured rather than static) instead of the `previewNotYetSupported` stub. A `recurs_yearly` instance refuses preview honestly rather than answering an almost-always-misleading zero-match snapshot.
- **Frontend**: the automation editor's schema-driven form gained a boolean checkbox, an enum-driven picker, and a date-field picker scoped to the chosen object (sourced from the real `/custom-fields` catalog, filtered to active date-typed columns).

## Review trail

Two independent automated reviews (security-redteam, craft-reviewer) plus a Fable design/code pass ran over the full diff before this PR opened, on top of the deterministic gates. All findings were fixed, not just filed:

- A misconfigured `renewal_reminder` instance (a retired or never-created custom field) used to abort the **entire workspace's** clock-scan pass, taking a healthy `no_activity_reminder`/`check_in_cadence` instance down with it in the same tick. Fixed with an `ErrDateFieldUnavailable` sentinel crossing the seam, translated at the one point both error types are in scope (compose's adapter) — now a per-instance skip.
- Preview 500'd on an unknown/wrong-typed `date_field` instead of the 422 every other invalid param on this surface gets. `AutomationStore` gained a `WithFieldCatalog` seam (mirroring `deals.Store`/`people.Store`) so preview validates against the live catalog before building SQL.
- A `days_before: 365` recurring window (the allowed maximum) silently matched almost nothing — the MMDD boundary comparison took the wrong branch when the window's bounds coincided. Fixed with a one-character change (`<=` → `<`) that lets the existing wraparound OR-formula's tautology handle the full-year case correctly, no special case needed.
- Nine smaller quality findings (comment residue, a weakened fitness test, an untested pure function, naming, a stale doc comment, etc.) — all fixed.

## Test plan

- [x] `make check-backend` — green (build, vet, both golangci-lint configs incl. the strict new-code gate, unit tests, drift, composition)
- [x] `make test-integration` — green, 0 skips, 40 packages (real Postgres)
- [x] `make frontend-check` — green (design-system gates, lint, 2497 unit tests, build)
- [x] `craft static --strict` (the exact pre-push hook) — 0 blocker, 0 major, 0 minor
- [x] Manual UAT on a live `make dev` stack (browser-driven): custom field creation, seeding, the automation form's enum/date-field/checkbox controls, a working non-recurring preview (`Matches now: 1`), and the recurring-preview honest refusal (422, legible message, confirmed via network + console) — full pass, no crashes or console errors
- [x] New-code coverage: unit tests for every new pure function (`projectOccurrence`, `validateDateColumn`), a fitness test binding `automation`'s object vocabulary to `customfields.FieldObjects`, and four new real-Postgres integration tests (in-window/out-of-window firing, two-year recurrence re-arm, misconfigured-instance isolation, preview-matches-real-rows)

## Follow-ups filed, not fixed here (documented, in scope decisions)

- #1188 — `/custom-fields`'s `object` query enum has no `project` entry, found while building the frontend date-field picker (pre-existing contract gap, unrelated to this feature's own object vocabulary).
- A recurring instance's preview is out of scope (DESIGN.md's own call) — it refuses honestly rather than fabricating an answer; a follow-up could project a recurring window into preview if that's wanted later.
- UAT noted the preview panel shows a bare match count, not record names/links — pre-existing behavior of the shared preview component this PR didn't touch, worth a follow-up polish.

---
## Summary by cubic
Wires `renewal_reminder` to real custom date-field candidates and previews, adds yearly recurrence, and fixes date and exclusion edge cases. Previously same-day dates could scan but never fire and archived records could mint tasks; now scan, match, and preview align on dates and archived exclusions.

- Candidate source: `DateFieldScan` via `customfields.Service.DateFieldCandidates`; validates `(object, date_field)` against the live catalog, excludes archived rows, binds DATE-only window bounds, and maps unknown columns to `automation.ErrDateFieldUnavailable` for a per-instance skip instead of aborting the workspace scan.
- Recurrence and matching: projects by MMDD with a strict “<” fix for 365-day windows; `Match` truncates `OccurredAt` to UTC date so “due today” fires.
- Preview: builds a dynamic predicate from instance params, validates against `fieldcatalog.Reader`, and refuses recurring previews with 422; fixes the “would have fired” estimate to use span-overlap logic.
- Params/UI: adds `object`, `date_field`, `recurs_yearly` beside `days_before`; editor renders an enum picker for `object` and a scoped date-field picker with empty/error guidance. Instances must set `object` and `date_field` to activate scanning.
- Tests: real-Postgres integration for scan and preview, parity/fitness tests for object vocabulary and archived-at, and unit tests for occurrence projection and param validation.

<sup>Written for commit bffec9fcb7f8e33a502c5f85b0136a7e94518b28. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1192?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added renewal reminders based on custom date fields, with configurable lead time and yearly recurrence.
  * Added date-field selection to automation creation and editing, including object-specific filtering and empty-state guidance.
  * Added dynamic previews showing matching records and reminder timing.
* **Bug Fixes**
  * Invalid or unavailable date fields no longer interrupt processing of other automations.
  * Improved handling of recurring dates across year boundaries.
* **Localization**
  * Added date-field selector translations in English, German, and Vietnamese.


tinnhlt-gra:
## UAT — live `make dev` stack, browser-driven

Full golden path walked end to end on a fresh cold-boot stack:

1. Settings → Data model: created a `date`-typed custom field (`Renewal Date`) on Person.
2. Seeded a person with that field ~15 days out.
3. Automations screen → "Use template" on `renewal_reminder`: confirmed `object` renders as a real picker (not free text), `date_field` is disabled until an object is chosen and then queries/offers the real custom field by human label, `days_before` is a number input, `recurs_yearly` is a checkbox. Saved (created paused, as designed).
4. Preview: `Matches now: 1` for the seeded person — proves the whole chain end to end (schema → params → live custom-field validation → real SQL predicate).
5. Created a second, `recurs_yearly` instance over a real birth date. Preview refused honestly: *"preview is not yet supported for a recurring (recurs_yearly) renewal_reminder instance…"* (422) — the deliberate design decision this PR makes, rather than a silently misleading zero-match snapshot.
6. Console clean throughout (only the two expected 422 `ProblemError` logs from step 5).

No manual clock-scan trigger exists in the Maintenance page, so the actual scan/task-creation mechanics are proven separately by the real-Postgres integration suite (`TestRenewalReminderFiresWithinWindowButNotBeyondIt`, `TestRenewalReminderRecurringAnchorReArmsEachYear`, `TestRenewalReminderMisconfiguredInstanceDoesNotAbortTheWorkspacePass`, `TestRenewalReminderPreviewMatchesTheRealSeededRows`) — this pass proved the human-facing UI wiring.

UAT recording

Minor non-blocking UX note from the pass: the preview panel shows only a match count, not record names/links — pre-existing behavior of the shared preview component this PR didn't touch, worth a follow-up polish.

tinnhlt-gra:
## Response to CodeRabbit's review

Fixed 6 of the 7 actionable findings, pushed in \`1bab59a31\`/\`bffec9fcb\`:

- **firedCount days_before shift** (automations_preview_renewal.go): fixed, but not with the exact suggested diff — I independently re-derived the correct window via interval-overlap math and found the suggested symmetric shift (\`[since+days_before, now+days_before]\`) has an edge-case gap (it drops an already-past-due record whose active-match span closed before \`since+days_before\` but still overlapped the trailing window). Applied the mathematically correct \`[since, now+days_before]\` instead (only the upper bound shifts), proven by a new \`WouldHaveFired\` assertion in the integration suite.
- **Archived-row exclusion** (candidates.go): fixed, both query shapes now exclude \`archived_at IS NULL\` rows, with a new integration test proving it.
- **DATE-precision binding** (candidates.go): fixed, \`queryLiteralCandidates\` now binds DATE-only string bounds instead of raw \`time.Time\`. This surfaced a *second*, related bug our own end-of-session review round caught: \`renewalReminder.Match\` was still comparing the DATE-typed anchor against the scanner's wall-clock \`OccurredAt\`, so a renewal due exactly today was scanned but then silently discarded — fixed in \`bffec9fcb\` by truncating to UTC calendar-date on both sides of the comparison.
- **Fixed clock in test, recording fake instead of nil, frontend \`as\`-cast removal, frontend \`isError\` handling**: all fixed as suggested.

**Not addressed**: the suggestion to replace the catalog's \`map[string]any\` schema/params shape with typed values. That pattern predates this PR — every other catalog entry (\`singleIntParamSchema\`, \`noParamsSchema\`, the lead-routing schema) uses the identical idiom, so typing only the \`renewal_reminder\` addition would introduce a second, inconsistent convention alongside the existing one rather than fixing anything this PR is responsible for. I think this is a legitimate architectural improvement, but it's a dedicated cross-cutting refactor across the whole automation catalog, not a scope fit for this PR — happy to file it as a follow-up issue if that's useful.

Issue #1230, opened 2026-08-14
Home page's Morning brief widget leaks a raw 'not yet implemented' error


tinnhlt-gra:
## What's wrong

The Home page's "Morning brief" widget shows a raw server error to the end user instead of a graceful empty/loading state:

> Couldn't load this view. operation GetMorningDigest is specified but not yet implemented

This looks like a contract operation (`GetMorningDigest`) that exists in the OpenAPI spec but has no real handler behind it yet (falling through to the generated `Unimplemented`/stub response), with the frontend surfacing that raw message directly rather than handling the "not built yet" case.

## Repro

Log into a fresh dev install and land on the Home screen — the Morning brief card shows the error above instead of content or a sensible placeholder.

Found incidentally during UAT of an unrelated feature (#723); filed rather than fixed there since it's unrelated to that change.

Issue #1229, opened 2026-08-14
Import undo/resume is unreachable once you navigate away from Settings → Maintenance


tinnhlt-gra:
## What's missing

The import screen's completed-run state (`run`/`report` in `useImportFlow`) lives only in frontend React state, reset on remount. Once a CSV import has run (dry-run, commit, or now undo — #723), navigating away from Settings → Maintenance and back loses all of it: the screen resets to the blank "Choose a file" state even though the run and its rows are still live server-side.

This affects two documented flows, not just one:
- **Resuming a failed commit** (IEM-WIRE-6 promises a failed run is "a resumable state, not a dead end") — unreachable once you've navigated away.
- **Undoing a committed import** (#723, A93) — same problem. In particular, the natural "edit one row you don't want reversed, then undo the rest" flow requires leaving the Maintenance screen to edit the row (e.g. via the Leads list) and coming back, which loses the undo affordance entirely.

## Repro

1. Settings → Maintenance → import a CSV, commit it.
2. Navigate to Leads (or any other screen).
3. Navigate back to Settings → Maintenance.
4. The import card is back to "Choose a file" — no sign the run exists, no way to undo or resume it from the UI (only a fresh upload).

## Suggested fix

`GET /imports/{id}` and `GET /imports/{id}/report` already exist and answer for a run by id regardless of frontend state. The screen needs either:
- a "reopen the last run" affordance backed by `RunStore.Latest` (already exists at the store level, not currently wired to any handler), or
- a persisted reference (e.g. the run id in `localStorage`/URL) the screen re-hydrates from on mount.

Found during UAT of #723; filed rather than fixed there since it's a pre-existing limitation of the screen's single-active-run model, not something that PR introduced.

Issue #690, opened 2026-08-09
Build the CSV import path: upload a file, map the columns, dry run, commit


LarsGradion:
## What is missing

We cannot import a customer's data file. There is no CSV reading, no column mapping, no import API operation, and no import screen. A customer who wants to bring their existing CRM or ERP export into Margince has no way to do it in the product.

## Why it matters

This is the first thing every new customer needs and the last thing we can do. Two customer engagements in flight both start with "here is our export, load it". Right now the only route is a hand-written script run by us, which does not scale and cannot be sold as product.

## What already exists — do not rebuild this

The engine underneath is real and good. Reuse it:

- `backend/internal/modules/migration/engine.go` — the generic import engine: a `Source` seam (objects, counts, rows, associations), a `Writers` seam, `DryRun` producing per-object will-create / will-update / skipped counts, and a checkpointed resumable `Run`.
- `backend/internal/modules/migration/run.go` — the `import_run` store, audit entries, checkpoints.
- `backend/migrations/custom/20260730120000_flip_and_import_run.up.sql` — the `import_run` table. Its `connector` CHECK already allows `csv`, `hubspot`, `salesforce`.
- `backend/migrations/custom/20260730140000_import_record_map.up.sql` — the external-id to native-id map, so re-delivering the same file does not duplicate rows.
- Two `Source` implementations exist as worked examples: `backend/internal/compose/flipsource.go` (overlay mirror) and `flipbundle.go` (export bundle restore).
- CSV *writing* already exists in `backend/internal/compose/filteredexport.go` and `exportbundle.go` — read those for how we handle CSV encoding.

## What to build

1. A CSV `Source` implementation that reads an uploaded file and yields rows. Add the `csv` connector constant next to `ConnectorMirror` / `ConnectorBundle` in `engine.go`.
2. A column-mapping layer. The engine expects rows already keyed by native column names, so this is the genuinely new part: read the file's header row, suggest a mapping to native fields, let the user correct it, and store the mapping on the import run.
3. The HTTP surface. There are zero import operations in `backend/api/crm.yaml` today. Per the spec chapter this needs a contract change first (upload, preview/dry-run, start, status, and later undo).
4. The screens: upload a file, review and fix the suggested mapping, read the dry-run report, then commit.
5. Custom fields must be mappable targets too, not just core columns.

## Done when

- [ ] A user can upload a CSV of organizations or people, map its columns, see a dry-run report, and commit it.
[...]

PR #1231, opened 2026-08-14
feat(import): undo reverses a committed csv run, and names what it kept

### LINKED ISSUE #723: Build the undo-this-import operation
## What is missing

Imports are supposed to be reversible as a whole batch. The bookkeeping for it exists — the import run, the checkpoints, the map from external ids to the rows we created, and even an `import_undo` value in the audit action vocabulary — but there is **no operation that actually undoes an import**. The audit vocabulary names an action nothing can perform.

## Why it matters

Reversibility is what makes a customer brave enough to try an import at all. It is also the honest half of "dry run, then commit": if the dry run was wrong, they need a way back. This should land with, or immediately after, the CSV import path.

- `backend/internal/modules/migration/run.go` — the import run store, audit entries and checkpoints.
- `backend/migrations/custom/20260730140000_import_record_map.up.sql` — the external-to-native id map, which is precisely the list of what an undo has to remove or revert.
- `backend/internal/modules/migration/engine.go` — the engine, including the create-versus-update classification an undo needs to respect.

## What to build

1. An undo operation for an import run, using the record map.
2. Correct handling of the two cases: rows the import **created** can be removed; rows it **updated** need their prior values restored, which means the undo has to be able to find them — check what the audit rows already give us before designing new storage.
3. A refusal when undo is no longer safe — for example a row a human has since edited. Refuse and explain rather than quietly reverting someone's work.
4. The audit trail for the undo itself.

## Done when

- [ ] A user can undo an import batch and the workspace looks as it did before.
- [ ] Rows touched by a human since the import are not silently reverted; the user is told.
- [ ] The undo is audited like any other write.

**Size:** S–M.
**Spec:** `import-export-migration.md` S-E11.6 (whole-batch reversible) and IEM-AC-9.

tinnhlt-gra:
## Summary

Closes #723 ("Build the undo-this-import operation"), the next pickup on `.tmp/capability-roadmap` wave 5 now that #690 (CSV import) has shipped.

`import_record_map` already recorded which native rows a run created; the operation to reverse them never existed. Contract-first: **IEM-WIRE-9** is minted upstream in three small `margince-foundation` PRs (merged before this one) —
- [#1289](https://github.com/gradionhq/margince-foundation/pull/1289) — mints `undoImportRun`, `POST /imports/{id}/undo`
- [#1290](https://github.com/gradionhq/margince-foundation/pull/1290) — fixes the `undo` field's null shape for the codegen it feeds
- [#1291](https://github.com/gradionhq/margince-foundation/pull/1291) — adds the `errored` bucket (see "what review found" below)

### What it does

- `POST /imports/{id}/undo`, valid only from `complete` and only for the `csv` connector (`hubspot`/`salesforce` unbuilt; `mirror`/`bundle` are the ADR-0071 overlay flip, not a customer import).
- Every row the run created lands in exactly one bucket: **reversed** (archived through `DisqualifyLead`/`ArchiveOrganization`), **kept** (a human has touched the row since it was created — A93's "kept, you edited these"), or **errored** (could not be reversed — a business-rule refusal or a row-scope miss — named with why, never a reason to abort the rest of the run).
- Checkpointed, paged (200/page, matching the forward engine's own page size) and resumable on the run's own `checkpoint` column.
- A concurrent second `undo` call on the same run is refused (a session-scoped Postgres advisory lock held for the call), not a second pass over the same rows.
- Frontend: the post-commit outcome gains an "Undo this import (N rows)" action, reads back the reversed/kept/errored split, and offers to continue (not restart) an undo interrupted mid-page.

### What review found and fixed before this was ready

Three independent passes (a craftsmanship review, an adversarial security review, and an independent Codex pass) on the first cut found real bugs, not style nits — all fixed, see the second and fourth commits:
- The human-touch reference instant was the *run's* completion time, not each *row's* own creation time — fixed by joining `import_record_map.created_at` per row in SQL.
- The check only matched `action='update'`; a human who independently archived a row (audited as `'archive'`) was invisible to it. Now any human-actor audit row on the entity counts.
- A single unreversible row aborted every row after it and wedged the run in `undoing` permanently. Spec PR #1291 adds the `errored` bucket; the loop now records and continues.
- Two concurrent `undo` calls on the same run both passed the lifecycle check and both looped over the same rows (the row lock only covered its own transaction). Fixed with a session-scoped advisory lock.
- Unbounded row read + a full-report rewrite on every single row. Now paged.

## UAT

Manually driven end to end against a running dev stack (upload → dry-run → commit → undo), confirmed via the Leads list that reversed rows are genuinely `disqualified` (not deleted). Recording attached below.

Found two **pre-existing, unrelated** gaps along the way, filed rather than fixed here:
- #1229 — the import screen's completed-run state isn't durable across navigation (affects the existing forward-resume flow too, not just undo)
- #1230 — the Home page's Morning brief widget leaks a raw "not yet implemented" error

## Test plan

- [x] `make check` (backend + frontend) green
- [x] `make craft-static` — 0 blocker/major/minor
- [x] Store-level integration tests (`internal/modules/migration`): kept/reversed/errored split, connector/status gates, concurrency refusal, cross-page resume, tenant fencing
- [x] Compose-level integration tests (`internal/compose/integration`): full upload→approve→undo over real HTTP, organizations branch, human-disqualified-outside-the-import branch
- [x] Frontend: `import.test.tsx` covers the undo button, kept disclosure, errored disclosure, and the interrupted/continue state
- [x] Manual UAT against a running dev stack (see recording)

---
## Summary by cubic
Adds undo for committed CSV imports so operators can reverse a run without overwriting later human edits. Previously imports were irreversible; now undo archives rows the run created, keeps rows a human touched since creation, records per-row refusals, and can resume if interrupted.

- Endpoint: POST `/imports/{id}/undo`. Allowed from `complete` (fresh) or `undoing` (resume) and only for `csv`. Returns 202 with the run; 409 on pending/validating/running/failed, non-`csv`, already `undone`, or a concurrent call.
- Run lifecycle and report: new statuses `undoing` and `undone`. Once `undone`, the run report includes an `UndoReport` with `reversed_count`, `kept` (object+id), and `errored` (object+id+reason).
- Behavior: walks `import_record_map`; archives created rows via idempotent `csvWriters.Reverse` (leads/organizations); keeps any row with a human-audit entry after that row’s `created_at` (any human action counts); records only per-row domain refusals (`permission denied`/`not found`/`conflict`) in `errored`; other failures stop the pass and leave the run `undoing`.
- Resilience: paged at 200 with checkpoint-based resume; persists `undo_report` and checkpoint per page; session-scoped Postgres advisory lock blocks concurrent undo; all reads/writes repeat tenant predicates.
- Frontend: shows “Undo this import (N rows)” (created rows only), “Continue the undo” when interrupted, and the reversed/kept/errored split.

**Review and rollout**
- Migration adds `undo_report jsonb` and permits statuses `undoing`/`undone`; down migration folds them to `complete` and resets `checkpoint` to 0. Apply DB migrations before deploy.
- Review focus: `RunStore.Undo` paging + resume, human-touch SQL (per-row `created_at`, any human action), row-refusal classification into `errored` vs. stop-and-resume, advisory lock behavior, and `csvWriters.Reverse` for leads/organizations. Other connectors (`hubspot`, `salesforce`, `mirror`, `bundle`) remain unsupported and return 409.

<sup>Written for commit 9d8fc1e3f10d067bb570c68e537db02501606397. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1231?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added the ability to undo completed CSV imports.
  * Undo operations can resume after interruption and show progress.
  * Human-edited or disqualified records are preserved.
  * Completion reports show reversed, retained, and failed records.
  * Added English, German, and Vietnamese translations.

* **Bug Fixes**
  * Prevented invalid, duplicate, concurrent, or unsupported undo requests.
  * Added safeguards to maintain data isolation and reliable reversal results.


tinnhlt-gra:
## UAT — manually driven end to end against a running dev stack

*(Note: a screen recording exists locally at `import-undo-uat.gif`, but neither `gh` nor the GitHub REST API expose a way to attach a binary asset to a PR comment from the CLI — GitHub's drag-and-drop upload is a web-UI-only flow. Posting the full written report here instead.)*

**Flow exercised:** login → Settings → Maintenance → "Import a file" → upload a 2-row lead CSV (`Email,Full Name`) → column mapping auto-detected (`Email→email`, `Full Name→full_name`) → "Check what this will do" → dry-run showed `Create: 2, Update: 0, Unchanged: 0, Skipped: 0` → "Import 2 rows" → **"The import finished."** with a new **"Undo this import (2 rows)"** button → clicked it → **"The import was undone." / "2 rows reversed."**

**Verified in the estate, not just the report:** opened the Leads list with "Show archived" on — both rows now show status `disqualified` with an orange badge, confirming a real reversal (archive), not a delete.

**Two pre-existing, unrelated gaps found along the way** (filed, not fixed here — out of scope for this PR):
- #1229 — the import screen's completed-run state isn't durable across navigation (a pre-existing limitation of the single-active-run screen model, affecting the existing forward-resume flow too — not something this PR introduced, but it does mean undo is only reachable in the same browser session immediately after commit)
- #1230 — the Home page's Morning brief widget leaks a raw "not yet implemented" error

No kept/errored-bucket UI check was possible in this pass because of #1229 (editing a row to test "kept" requires navigating away from the Maintenance screen, which loses the undo affordance) — that path is covered instead by the automated integration tests (`TestCSVImportUndoReversesUntouchedAndKeepsEdited`, `TestCSVImportUndoKeepsARowAHumanDisqualifiedOutsideTheImport`) which exercise it directly over HTTP without the UI's navigation constraint.

tinnhlt-gra:
## Addressed CodeRabbit's review

Fixed:
- **Major** — infrastructure errors were being routed into the `errored` bucket alongside genuine per-row refusals, letting a transient failure (dropped connection, timeout) finish the run as `undone` with rows never actually reversed and no way to resume. Now only a real row refusal (permission denied / not found / conflict) lands in `errored`; anything else stops the pass and leaves the run `undoing` so a later call resumes it. Added a store-level test for this exact path.
- **Major** — the undo lock's connection was released back to the pool even when the unlock itself failed, which could leave the session-level lock held by an idle pooled connection. Now follows the existing `claimFlip` pattern: hijack and close the connection on a failed unlock instead.
- `toContractImportReport`'s `committed` check excluded `undone`/`undoing`, so the report for an undone run showed dry-run predicted counts instead of what the commit actually did.
- The undo button's row count included `updated` rows; undo only ever reverses rows the run *created* (there's no `import_record_map` entry for a mere update).
- `clearAnswers` didn't reset the `undo` mutation, so a stale error/pending state could leak into the next file's outcome.
- Wording: "unreversible" → "irreversible"; the operation and 409/checkpoint docs now say `undoing` is a valid start state (resuming) — mirrored upstream in margince-foundation#1292 (merged).
- Test quality: the "no kept rows" assertion queried an untranslated i18n key (can never match rendered output); a test payload now uses the existing `apptest.AnyMap` convention instead of a bare `map[string]any`.

**Not fixed here, filed as #1233**: the human-touch check and the archive aren't atomic — a millisecond-scale TOCTOU window remains between checking `humanTouchedSince` and calling `Reverse`. A proper fix means passing the audit cutoff into `DisqualifyLead`/`ArchiveOrganization` and rechecking it inside the same transaction as the archive — those methods are called throughout `modules/people`, so that's a real cross-module change, not a contained fix within this PR.

tinnhlt-gra:
import-undo-uat

PR #1265, opened 2026-08-14
docs(deployment): system requirements documentation

lstrojny:
Adds `docs/reference/system-requirements.md`. `deployment.md` covered how to run the three roles and said nothing about what to run them on; this is the missing half — the software floor, sizing for the single-node and separate-node shapes, the database settings that matter, and the network and availability constraints.

The page is deliberately **recommendations only**: an operator gets the numbers, not our derivation. Everything about how they were arrived at lives here instead.

## Where the numbers come from

Reproduce with `make db-up` then `make bench-perf`, which seeds the mid-market tier and runs the canonical queries.

Measurement host: Apple M3 Max (16 cores / 64 GB), Docker VM at 16 vCPU / 7.75 GB, Postgres 16 + pgvector at stock configuration apart from `max_locks_per_transaction=512`.

**Footprint at the mid-market tier** (250k persons / 500k activities), 156 tables and 509 indexes:

| | Size |
|---|---|
| Database total | 430 MB |
| of which table data | 217 MB |
| of which indexes | 195 MB |

Per row including indexes: activity ≈ 370 B, person ≈ 350 B, activity_link ≈ 250 B. An empty migrated schema is ≈ 20 MB.

**Embeddings dominate.** Measured separately — 20,000 distinct random `vector(1024)` rows in a probe table (107 MB) plus an HNSW index over them (156 MB):

| | Per chunk |
|---|---|
| Table (heap + TOAST) | 5.4 KB |
| HNSW index | 7.8 KB |
| **Total** | **≈ 13.2 KB** |

A vector is past the TOAST threshold and random floats do not compress, so there is no saving below the theoretical size, and the index costs more than the row. At mid-market with one chunk per person / organization / activity — ~760k chunks — that is **~10 GB against 430 MB of CRM data**. Hence the page splitting its tables on whether retrieval is bound rather than on tier alone.

**Latency**, mid-market tier, 20 samples:

| Query | p50 | p95 | p99 | Budget |
|---|---|---|---|---|
| `search_fts` | 121ms | 127ms | 130ms | 200ms |
| `context_graph` | 15ms | 16ms | 16ms | 300ms |

Both pass with wide margin, but against a fully page-cached 430 MB database — which is exactly the condition the embedding footprint makes unlikely in production. The page therefore states no latency guarantee.

## Constraints read out of the code

Each is invisible from the outside and expensive to learn in production:

- **Pool math.** 16 connections per process, plus the api's owner pool of 3, plus River's — one api and one worker is ~40 backends, every api replica +19. And since the RLS binding is `set_config(..., true)`, a pooler must run in **transaction** mode: statement pooling splits a transaction across backends and drops the tenant GUC the policies key on.
- **Both roles scale horizontally.** The relay claims outbox rows `FOR UPDATE SKIP LOCKED`, migrations serialize on an advisory lock, and the worker's periodic jobs run under River's leader election.
- **api-to-database latency is a requirement.** Graph assembly is recursive CTEs over multiple round trips against a 300ms budget.

## Known gaps in the sizing

Not stated on the page, but real:

- **Local inference is unsized.** The page says only that it needs more memory and possibly a GPU. It deserves its own pass.
- **api and worker process memory** are estimates from the process shape, not measurements under load.
- **Redis worst case** is uncomputed: cap × stream count × envelope size, where each stream caps at 2^17 entries. The 2 GB figure is a placeholder.
- **Restore time at mid-market volume** is unmeasured — and it is the number that decides whether the single-node shape is acceptable at that tier.

## Notes for review

- Docs-only. `ci-doc-parity`, `check-craft-doc` and `make-target-parity` pass; the pre-push craft gate reports no changed Go files.
- The page is **not yet linked** from `docs/README.md` or `deployment.md`. Say the word and I will add both here.
- Unrelated, surfaced while writing the storage section: `config/ai-routing.example.yaml` binds `gemini-embedding-001` at `dimensions: 1536`, while `migrations/core/0022_embeddings` declares `embedding vector(1024)` and no migration alters that column. The width guard in `search/embedding.go` validates against the *configured* width, so a 1536-wide vector would pass it and then fail at the INSERT. Untested — but if it holds, a deployment copying the shipped example gets an installation whose embeddings never write.

## Summary by CodeRabbit

* **Documentation**
  * Updated system requirements guidance for deployment models, supported software, sizing, networking, storage, retention, scaling, and availability.
  * Clarified operational guidance for offline environments, connection pooling, cache services, and embedding storage.
  * Simplified several explanations and removed outdated or unnecessary references and requirements.

### DIFF
--- a/docs/reference/system-requirements.md
+++ b/docs/reference/system-requirements.md
@@ -38,7 +38,11 @@ state is in Postgres, Redis and the object store.
 Pick the tier by contact count. The second question is whether AI-powered
 search and retrieval is enabled: the embedding store is roughly twenty times
 the size of the CRM data itself, so it moves the numbers more than the tier
-does.
+does. That ratio assumes the default 1536-dimension embedding width; the store
+scales linearly with the configured `dimensions`. Switching the embedding
+model or width rewrites the store in place through a fleet-wide re-embed
+([configuration.md](configuration.md#embedding-binding-changes--reindex)) —
+it costs provider calls, not extra disk.
 
 ## Single node
 
--- a/docs/reference/system-requirements.md
+++ b/docs/reference/system-requirements.md
@@ -1,0 +1,115 @@
+# System requirements
+
+What an installation needs to run. Two deployment shapes are supported:
+**single node** (every service on one host) and **separate nodes** (api /
+worker / web / database on hosts of their own). Both run the same images and
+the same configuration.
+
+The sizing below assumes the AI lanes are bound to a cloud provider. Running a
+model on the installation's own hardware
+([how-to/enrich-with-a-local-llm.md](../how-to/enrich-with-a-local-llm.md))
+needs substantially more memory and possibly a GPU, and is not sized here.
+
+Deployment mechanics — roles, database bootstrap, routing, health checks — are
+in [deployment.md](../deployment.md).
+
+## Software
+
+| Requirement | Notes |
+|---|---|
+| **Postgres 16** with **pgvector** ≥ 0.5.0 | The `vector` extension is not trusted, so a superuser installs it once (`scripts/deploy/db-bootstrap.sql`). |
+| **Two non-superuser Postgres roles** | `margince_owner` (DDL) and `margince_app` (runtime). Neither may hold `SUPERUSER` or `BYPASSRLS` — both defeat the `FORCE ROW LEVEL SECURITY` tenant boundary. |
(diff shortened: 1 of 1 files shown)


tinnhlt-gra:
@lstrojny the embedding index become unbound at [migration 0114](backend/migrations/core/0114_embedding_identity.up.sql)

- When user update the embedding model / size -> must go to setting and re-embeding

PR #1269, opened 2026-08-14
build(gates): lint and gofmt the 11 Go modules outside backend's ./...

### LINKED ISSUE #1270: backend/tools: bring the module under the full baseline lint config (26 findings)
Split out of #1269, which put a tree-wide **gofmt** floor under every Go module but deliberately stopped short of the full baseline lint config.

`backend/tools` is a separate module, so `golangci-lint run ./...` from `backend/` never reaches it. Running the baseline config against it today:

```
$ golangci-lint run ./tools/...
26 issues:
* errcheck: 2
* forbidigo: 5
* gosec: 8
* misspell: 2
* staticcheck: 9
```

Bringing the module in needs two separable pieces of work.

## 1. A config carve-out for a build-time CLI (~13 of the 26)

These are the baseline config being wrong for this module, not the code being wrong:

- **`forbidigo` ×5** — "use slog, not fmt.Print*" on `fmt.Printf`/`Println` in `gen-agentpolicy`, `gen-payloads`, `gen-workflow`, `extmigrategate`. Each prints its own completion summary to stdout. That rule is a server rule; a generator's stdout *is* its interface.
- **`gosec` ×8** — G304 (file inclusion via variable), G306/G301 (file/dir perms), G204 (subprocess with tainted args), G703 (path traversal via taint) across `gen-composition`, `contract-overlay`, `gen-payloads`, `gen-recordfields`. A build-time generator reads and writes paths it computes itself and shells out to `go list`. The taint is the job description.

The fix is a per-path override in `.golangci.yml` (or a `backend/tools/.golangci.yml`) that disables these for the module — with a comment saying why, so the carve-out is not read later as an oversight.

## 2. A small real burn-down (~11 of the 26)

- **`errcheck` ×2** — `tools/gen-composition/unitmanifest.go:212,214`: `defer os.Remove(tmpName)` and `tmp.Close()`. A genuine T2 violation (never swallow an error) that the module boundary has been hiding.
- **`misspell` ×2** — `tools/gen-jobs/main_test.go:353,362`: `postive` → `positive`. Note this is fixture text asserted against, so both sides change together.
- **`staticcheck` ×7** — deprecations: `parser.ParseDir` (astreader.go:41, unitmanifest.go:283), `ast.Package` (astreader.go:82,149, astreader_liveness.go:38), `runtime.GOROOT` (main.go:203), plus `QF1001` De Morgan in extmigrations.go:336. The `ast.Package`/`ParseDir` cluster is one migration to `golang.org/x/tools/go/packages`, not six edits — and it is a real behavioural improvement, since `ParseDir` ignores build tags when associating files with packages.

## Not a finding: the two `SA4004`

Recording this so the next person does not re-derive it. `tools/extmigrategate/inventory.go:210,246` ("the surrounding loop is unconditionally terminated") are **correct as written**. `assertNoTriggers`/`assertNoRules` are fail-closed allowlist gates: they return an error on the first offending row by design, because the caller needs one clear refusal, not an enumeration. These want a reasoned suppression, not a rewrite.

## Also worth deciding in the same pass

`cli/craft` and `fixtures` are outside the lint lane for the identical reason. If a per-module config lands here, it should answer for them too rather than being repeated three times.

Scope note: `arch-lint` should **not** be extended — its DAG is the product module's (`shared → platform → modules → compose → cmd`) and means nothing for tooling.

lstrojny:
## Problem

`backend/tools/gen-composition/extjobs.go` was not gofmt-formatted and was merged to `main` with `make check` green.

The cause: `make -C backend lint` runs `golangci-lint run ./...` from `backend/`. The `./...` pattern stops at module boundaries. The repo has 12 Go modules, so this covers 1 of them.

| Module | Covered by `backend`'s `./...` |
|---|---|
| `backend/` | yes |
| `backend/tools/` | no |
| `cli/craft/` | no |
| `composition/` | no |
| `extensions/{de,dispact-connector,notes,yogi}` | no |
| `fixtures/extensions/*` (4 modules) | no |

The other 11 modules were only checked by the pre-commit hook, which runs on staged files for the person who staged them. `backend`'s `test` target already works around the same boundary manually (`go test ./tools/...`), but the lint and format passes never got the same treatment.

## Changes

Two new gates, both added to `check-backend`, so CI runs them without a workflow change.

**1. `make gofmt`** (`scripts/check-gofmt.sh`)

Checks every tracked Go file with gofmt. The file list comes from `git ls-files`, so there is no module list to maintain. Generated files (`*_gen.go`, `*.gen.go`) are skipped because the drift gate already verifies them and they must not be hand-edited.

**2. `make lint-modules`** (`scripts/check-lint-modules.sh`)

Runs golangci-lint on every module except `backend`, using the same `backend/.golangci.yml`. Takes about 6 seconds warm.

One shared config instead of one config per module: golangci-lint has no config inheritance, so a per-module config would have to copy the baseline, and copies drift.

### Why both gates are needed

golangci-lint reports gofmt violations, so `lint-modules` covers formatting for 7 of the 12 modules on its own. `gofmt` is still needed for two reasons:

- golangci-lint requires a type-checkable package. The `fixtures/extensions/*` units import the product module without declaring a `require` for it, on purpose, so they only resolve inside the test harness that composes them. gofmt only needs to parse a file, so it works there.
- It runs in under a second and keeps working if a module becomes temporarily unlintable.

### Three problems found while building `lint-modules`

- **Workspace membership is read from the generated `go.work`, not assumed.** If a module is not listed in the workspace it is run under, every package fails to type-check. golangci-lint then prints `0 issues` and exits non-zero. That looks like a passing gate but nothing was checked.
- **The run is uncapped** (`--max-same-issues=0`). The default cap is 3. After fixing the 3 visible `G304` findings, 4 more appeared that the cap had hidden. The original estimate of 26 findings was wrong for this reason.
- **`fixtures/` is excluded** for the type-checking reason above, not to make the gate pass. Those files are still covered by the craft gate, the license header test, and `gofmt`.

## Findings fixed

- `writeFileAtomic` discarded a `Close()` error when the write failed. Both errors are now returned via `errors.Join`. This is a real violation of the "never swallow an error" rule.
- `gen-recordfields` wrote output with mode `0o644`. Its three sibling generators use `0o600`. Changed to match.
- Two `os.WriteFile` calls in craft's tests had unchecked return values, so the tests could pass even if the write failed. Now checked with `t.Fatal`.
- `dollarTag` had an inline predicate with four negations. Extracted as `isDollarTagChar`.
- `parser.ParseDir` and `ast.Package` are both deprecated. Replaced with a local `parseDirByPackage`. The suggested replacement, `go/packages`, does not work here: it resolves and type-checks a build, but `gen-composition` runs before a resolvable build exists, since it generates the workspace the build uses. It also needs to read source without applying build tags, so the generated manifest is identical on every host.
- The two `misspell` findings are in test fixture data. `postive` is the key the parser is supposed to reject, so fixing the spelling would make the test pass for the wrong reason. Waived on the test with that explanation.

## Findings waived

Each waiver is on the line it applies to, with a reason.

- **`gosec`, 10 findings** (G304, G306, G301, G204, G703). A build-time generator reading and writing paths it computed itself. Uses the `#nosec` form already used 13 times in this module.
- **`staticcheck` SA4004, 2 findings.** `assertNoTriggers` and `assertNoRules` in `extmigrategate` return on the first row on purpose. They are fail-closed allowlist checks, and the caller needs one clear failure, not a list.
- **`staticcheck` SA1019.** `runtime.GOROOT()`. The suggested replacement is `go env GOROOT`, which requires finding a `go` binary on `PATH`. That code exists specifically to avoid trusting `PATH`.
- **`gosec` G404.** `extensions/yogi` picks a random entry from a public list of sayings. Not a security decision.

## `fmt.Print*` in the CLI modules

The `forbidigo` rule `use slog, not fmt.Print*` exists because a `fmt.Print` line has no level, no `correlation_id`, and no fields, so it cannot be correlated with audit rows or bus events. That reasoning applies to the server.

Two of the newly covered modules are build-time CLIs where it does not apply. `gen-jobs` prints `63 kinds, 9 queues generated` for the make lane. `gen-workflow` prints the files it scaffolded and the next steps. `craft` prints its findings table. For these tools, stdout is the output interface, not a log.

This would have needed 9 in-source waivers: 8 new ones in `backend/tools`, plus one that has been on `cli/craft/main.go:100` since the craft gate was added. Instead, two exclusion rules in `backend/.golangci.yml` state it once:

```yaml
- path: '^tools/'
  linters: [forbidigo]
  text: 'use slog, not fmt\.Print'
- path: '^\.\./cli/craft/'
  linters: [forbidigo]
  text: 'use slog, not fmt\.Print'
```

Two details matter here:

- **`text:` limits this to the `fmt.Print*` pattern.** `forbidigo` also enforces the two river registration rules and `http.Error`. Excluding the whole linter by path would disable those too, and would disable any pattern added later.
- **`path:` is relative to the config file's directory**, which is how golangci-lint reports issue paths. So `backend/tools` appears as `tools/` and `cli/craft` as `../cli/craft/`. Neither pattern can match the product module, whose files appear as `internal/`, `cmd/`, or `pkg/`. If a golangci-lint upgrade changed this convention, the patterns would stop matching and the findings would reappear, failing the gate rather than silently widening it.

The extension units are not included. They are libraries loaded into the server, where the rule applies as written.

**Trade-off, documented in the config:** a `fmt.Print` left in a tool after debugging is now invisible to the gate. An in-source waiver would have required someone to write a reason for it.

## Verification

- Both gates were run against a deliberately broken file and fail on it, then pass once fixed.
- The `forbidigo` exclusion was checked three ways: the gate passes with the rules in place; an `http.Error` added to `backend/tools` is still reported, confirming `text:` limits the scope; and with the two rules deleted, all 9 findings reappear and `lint-modules` fails.
- `make check-backend` passes, including both new gates.
- `make gofmt`: all 3081 hand-written Go files across every module are formatted.
- `make lint-modules`: 7 modules outside `backend/` are clean.
- `craft static --strict` on all changed Go files: 0 findings.

Closes #1270.


lstrojny:
Follow-up for the deferred half filed as #1270 (config carve-out + the ~11 real findings, and a note on why the two `SA4004` are not findings).

lstrojny:
The red checks here are **not** from this change. `main` at 165e5f3e cannot load its own core migration sequence — `pgmigrate: duplicate version 0240`, because #1251 and #1264 each added a different migration numbered 0240 and both merged. Filed as #1271.

This PR adds no migrations; its `make check-backend` is green locally against its base, and CI fails only because it tests the merge with current `main`. Every open PR is red on the same two tests. Blocked until #1271 lands, then this needs a rebase and a re-run.

lstrojny:
## The printf problem — three options, none picked

Eight of the waivers this PR adds are the same sentence:

```go
//nolint:forbidigo // a CLI reporting its own result to stdout, not a server logging an event
fmt.Printf("%d kinds, %d queues generated\n", len(c.Kinds), len(c.Queues))
```

across `gen-jobs`, `gen-aitasks`, `gen-agentpolicy`, `gen-payloads`, `gen-stubs`, `gen-workflow` (×2) and `extmigrategate`. The rule is `.golangci.yml`'s `^fmt\.Print.*$` → *"use slog, not fmt.Print*"*, whose stated reason is that a `fmt.Print` line carries no level, no `correlation_id` and no fields. That reason is about the **server**. A generator's stdout *is* its interface to the make lane that ran it.

Three ways to answer it. I checked all three against the real config rather than reasoning from the docs.

### A — in-source waiver, one per line (what this branch does)

**Cost:** the same justification written eight times. The real risk is not verbosity, it's that a repeated waiver reads as boilerplate, and the ninth `fmt.Print` — the one that is genuinely wrong — gets the same rubber stamp on review.

**In its favour:** it binds the line, not the file. And it is already the tree's idiom — `cli/craft/main.go:100` on `main` today carries `//nolint:forbidigo // CLI stdout output, not logging`. Option A is the status quo, not a new invention.

### B — a path-scoped exclusion in `backend/.golangci.yml`

```yaml
exclusions:
  rules:
    - path: '^tools/'
      linters: [forbidigo]
```

**This works, and the reason it works is not obvious.** golangci v2 reports paths relative to the *config file's* directory, not the working directory. `check-lint-modules.sh` `cd`s into each module and passes `--config backend/.golangci.yml`, so a finding in `backend/tools/gen-jobs/main.go` is reported as `tools/gen-jobs/main.go` and one in `cli/craft/main.go` as `../cli/craft/main.go`. `^tools/` therefore matches the tools module and nothing in the product module — there is no `backend/tools` inside `./...` to collide with. Verified: with that rule and the `//nolint` deleted, `gen-jobs` reports `0 issues`.

**Cost, and why I did not just do it:**

1. It exempts the **whole linter**, not the pattern. `forbidigo` also carries the two `river.AddWorker` / `PeriodicJobBundle` rules and `http.Error`. Harmless for a codegen module that imports none of them — but it is a wider hole than the one being cut, and nothing fails if that stops being true.
2. It exempts the whole **tree**, so a `fmt.Print` added to a tool for debugging and forgotten is invisible. The config's own comment (lines 154-158) argues against exactly this for the sanctioned `river.AddWorker`: *"exempting the whole file would also exempt a second, unsanctioned registration added to it later."*
3. It puts the justification in a path list that outlives whatever it was written for, rather than next to the code.
4. It answers for `backend/tools` only. `cli/craft` needs `../cli/craft/` and the extension units another entry — three path rules restating one idea, which is the thing the single-config decision was trying to avoid.

### C — stop calling `fmt.Print*`

The rule's pattern is `^fmt\.Print.*$`. `fmt.Fprintf(os.Stdout, ...)` does not match it, so this needs no waiver anywhere.

Done as a lint-dodge that is worse than the waiver. Done properly it is the only option that leaves the tree better:

```go
func main() { os.Exit(run(os.Args[1:], os.Stdout, os.Stderr)) }

func run(args []string, stdout, stderr io.Writer) int {
    ...
    fmt.Fprintf(stdout, "%d kinds, %d queues generated\n", len(c.Kinds), len(c.Queues))
    return 0
}
```

**In its favour:** the generators become testable end-to-end — today most of them are `func main()` with `log.Fatalf`, so their summary line and their exit behaviour are asserted by nothing. Writing to an injected `io.Writer` is not a workaround for the linter, it is what makes the output an interface instead of a side effect. `gen-stubs` and `gen-composition` already have a `run() error` shape to extend; `gen-jobs`, `gen-aitasks`, `gen-agentpolicy`, `gen-payloads` and `gen-recordfields` do not.

**Cost:** it is a refactor of seven `main()` functions plus the tests worth writing once they are testable, and it does nothing for the ten `#nosec` waivers, which stay either way. That is a change with its own review, and folding it into a gate PR would bury it.

### Recommendation

**A now, C as a follow-up, not B.** A is the tree's existing answer and this PR does not have to relitigate it. B trades eight honest local statements for one silent regional hole and buys only brevity. C is the change that would actually make the waivers unnecessary rather than invisible — and its real payoff is the testability, not the linter, which is why it deserves its own PR rather than a paragraph in this one.

If the eight repeated lines are the objection, the cheapest honest fix inside A is to write the *reason* once — as a package comment in each tool, or a single line in `docs/reference/` the waivers point at — and shorten the waivers to `//nolint:forbidigo // generator stdout; see <ref>`.

lstrojny:
Decision: **B**, with the two narrowings the original objections asked for.

`text: 'use slog, not fmt\.Print'` answers objection 1 — the carve-out is the `fmt.Print*` pattern, not the linter, so the river-registration and `http.Error` rules stay armed in `backend/tools`. Verified by planting an `http.Error` there: still reported.

Scope is `backend/tools` **and** `cli/craft`, not just the former. Doing only `backend/tools` would have left the repo with two different answers to one question — a path rule here, `//nolint:forbidigo` on `cli/craft/main.go:100` there — which is the drift the single-shared-config decision exists to avoid. That pre-existing waiver is removed too. The extension units stay under the rule: they are libraries loaded into the server.

Objection 2 stands and is unfixable within B, so it is written into the config instead of left to be rediscovered: a stray `fmt.Print` left in a tool after a debugging session is now invisible to the gate. That is the accepted cost.

9a19295..6009f64

PR #1274, opened 2026-08-14
feat(license): show the licensee, the seats and the expiry date

lstrojny:
This PR adds a License page to Settings. The page shows four things:

- who holds the license
- how long the license is valid
- the number of seats that the license grants, and the number of seats in use
- a warning when the installation uses more seats than the license grants

This is the read part of #1190. The next slice adds the enforcement. **This PR blocks nothing.** The workspace continues to operate when the seats are more than the limit, and when the license is past its expiry date.

## The two cards

**Licensed to** shows the organization, the contact person and address, the installation handle (`subject`), the expiry date, and the license id. The module reports these values; the product does not store them.

Each licensee row appears only when the token contains that claim. A license from before these claims existed is valid in the same way, and the page then shows the rows that it has. An empty row tells the reader that something is missing from their license, which is not correct.

**License and seats** shows the seat count and the meter.

| The license | The page |
|---|---|
| grants a seat limit | shows the used seats, the granted seats, and a meter |
| grants no seat limit | shows "No limit", and shows no meter |
| is not configured | shows "No license configured" |
| grants fewer seats than the installation uses | shows a warning, and marks the used-seats slot |

The server does not send `seats_granted` when no limit applies. It does not send zero. A client that changes this absent value to zero makes two incorrect statements: that the license permits no seats, and that each seat is more than the limit. A test prevents this behavior.

## The warnings, in the order of their cost

| State | The page |
|---|---|
| more than 90 days before the expiry date | shows the date only |
| 90 days or less before the expiry date | shows an amber notice. It does not interrupt. |
| after the expiry date, in the grace period | shows a red notice. It interrupts. |
| more seats than the limit | shows a red notice on the card below |

An expired license stops the installation after the grace period. Too many seats stop nothing. Therefore the expiry notice is above the seat notice.

The renewal window is 90 days. The installation gets a warning for the same number of days before the expiry date as it continues to operate after that date. This number is equal to the grace period of the module. It is not the same constant: the module keeps its grace period and does not report it. If the module changes that period, this number no longer agrees, and no test can find it. Issue #1190 records a request for a `grace_until` field.

The server sends the two decisions (`in_grace`, `renewal_due`) and not only the dates. A client cannot then make a different decision from the one that the installation makes. `renewal_due` is true when `in_grace` is true, thus a client that reads one field gives a warning.

## Which license authority the installation accepts

A production installation accepts one authority: `margince-license-authority`.

This is necessary. The first version of this code gave a different reason for the same rule:

> the bundled module trusts only the production keyset, so a token from any other issuer could never verify against it

That statement is not correct. A license from the test authority has a valid signature, because the test licenser uses a key that the published keyset contains. A test license was refused on the issuer name alone. The issuer is thus the only control that stops a test license from licensing a customer.

An installation with `MARGINCE_ENV` set to `dev`, `staging` or `test` also accepts `margince-license-authority-test` and `margince-license-authority-dev`. A developer can then use a test license. `MARGINCE_ENV` is fail-closed: an unknown value or no value is production.

A production installation makes one call to the module, as before. The first refusal is the one that the product reports. An expired production license reports that it is expired, and not that the issuer is incorrect.

## Permissions

Only the admin role and the ops role can read this page. A seat meter shows the commercial position of the installation. UC-ADMIN-03 F1 gives a rep the seat of the rep, but not the entitlement of the workspace.

This needs a new `license` RBAC object, and a new object needs a policy backfill migration (core 0258). Migrations 0191 and 0121 did the same. The code writes each role document one time, when it creates the workspace, and does not write it again. An object in `coreObjects` alone thus operates on a new database and gives a 403 error on all older installations.

**Read is the only permitted operation, also for the admin.** The server reads the token from the deployment file at start, thus the API has no write operation for a permission to control.

The operation is `human-only` (`x-agent-access`), as all related governance reads are.

The response never contains the token. The token is a credential, and the response goes to a browser.

## The count of the seats

The count includes each full seat that is not deactivated, and it includes the agent seats.

- It does not include read seats. Read seats have no limit (A62/ADR-0047).
- It does not include a suspended seat. The installation removed this access before.
- It includes an agent seat. The constraint `app_user_agent_is_full` makes each agent a full seat.

This count is different from the "active" count in the specification (a sign-in in the last 30 days, UC-ADMIN-03 precondition 3). The two meters thus do not agree, and issue #1190 records the difference.

## Notes for the reviewer

**The code connects the handler at one point only.** `WithLicensePosture` builds it. The assembly runs before the options, thus a handler that the assembly builds gets a nil posture and gives a 501 error for the full life of the process. I made this error first, and a test now prevents it.

**Two items need a second opinion:**

1. The meter looks the same at 10 of 10 and at 11 of 10. The meter stops at 100 percent and keeps its colors. The notice and the numbers show the difference. I did not use `tone="danger"`: the `Meter` contract keeps the tones for values where a low value is bad. A change to the bar is a change to the design system.
2. Nobody examined the narrow story. `fe-uat` ignores the viewport value of a story, and its wide image and narrow image are equal. Only the viewport control in Storybook shows the page at 390 px.

**Two gates found errors in the writing, not in the code.** The `migration-versions` gate found that main took 0257 for `channel_provider_display`; this migration is now 0258. SonarCloud reported 34 parameters on `objects()`, which had 33 before this branch. The waiver names the rule and the reason, as `mailer.go` does. The parameter list is the mechanism of that function, and a change to it is a refactor of the RBAC seed.

**Not verified against a running server.** Fixtures supply each state. A test license is now possible in a development installation, thus a person can do this verification.

**One gate fails and this PR did not cause it.** `make lint-modules` reports `fmt.Printf` in `backend/tools`. The same failure occurs on main, and another branch corrects it.


lstrojny:
`core/0260` collided again: `main` took it for `scheduled_send_agent_provenance` after this branch claimed it for `license_rbac`. The green `deterministic-gates` above predates that merge — the next push here will fail `migration-versions`, and the stacked #1364 already does, since it carries this branch's migration.

Not renumbering it from #1364, since this branch is where the file lives and it is being worked on.

Issue #696, opened 2026-08-09
Add time periods to reporting (revenue by year, closed-won by quarter)


LarsGradion:
## What is missing

The reporting engine has **no concept of time periods at all**. There are four fixed reports and their allowed dimensions contain no year, quarter or month. You cannot ask "revenue by year", "closed-won per quarter", or "activity per month". Custom fields are also not available to any report.

## Why it matters

A customer engagement needs to show revenue history per account and a consolidated number over a corporate group. Neither is expressible today, whichever way the underlying data is modelled. It is also the foundation for the "KPIs per site, per industry, per period" ambition further out.

## What already exists — do not rebuild this

- `backend/internal/compose/report.go` holds the four prebuilt reports: `open-deals-per-company`, `deals-by-stage`, `activities-by-kind`, `forecast`. Each compiles to a validated plan against an allowlisted vocabulary; anything outside it is refused with a clear error.
- `backend/internal/compose/reportsql.go` and `reportcatalog.go` are the compiler and catalogue.
- "Explain this number" already works — every report run can return the executed plan and the source rows. Keep that for the new reports.
- There is an ad-hoc planner (`runAdHocPlan`) that derives dimensions from schema descriptors, but it is deliberately not reachable over HTTP — it serves internal tooling and conformance tests.
- The corporate-group rollup endpoint returns exactly three fixed measures (weighted open pipeline, closed-won for the current quarter, 30-day activity count) with no period parameter.

## What to build

1. A date/period dimension in the report vocabulary (year, quarter, month buckets over a chosen date column).
2. A `revenue-by-year` style prebuilt report using it.
3. Allow custom field columns as report dimensions and measures.
4. A period parameter on the corporate-group rollup, so "the whole group in year X" is answerable.

## Done when

- [ ] A user can run a revenue-by-year report and drill into the rows behind any number.
- [ ] The same report works across a corporate parent tree, not only a single company.
- [ ] Out-of-vocabulary requests still fail with the existing clear error rather than running.

**Size:** S–M for the first report, M with the tree and custom fields.
**Spec:** `reporting.md` (the chapter notes the report compiler and runner exist but the coverage surface does not) and `records-depth.md` RD-FORM-1 for the rollup measures.

tinnhlt-gra:
**Items 1 and 2 shipped** in #1281 (`80a96ca93`), spec-first via margince-foundation#1298 which pinned REPORT-PARAM-4..8 and REPORT-VOCAB-1.

[...]

Issue #693, opened 2026-08-09
Make custom fields, tags and city usable as list and segment filters


LarsGradion:
## What is missing

You can store a value in a custom field, and then you cannot filter by it. The filter vocabulary that powers dynamic lists, saved views and exports is a fixed set of core columns. It contains no custom fields, no tags, and no city.

The practical effect: a customer sets up their own category field, fills it in, and cannot build a target list from it.

## Why it matters

This is the difference between "you can add your own fields" and "you can work with your own fields". Both customer engagements hit it immediately — one to build category-based target lists, one to slice imported data. It also blocks anything downstream that starts from a list.

## What already exists — do not rebuild this

- `backend/internal/modules/collections/vocab.go` is the single closed filter vocabulary. Today: organization has `owner_id`, `industry`, `size_band`, `lifecycle`, `classification` (retired); person has only `owner_id`; deal has pipeline, stage, owner, organization, partner, project, status, forecast category.
- The rest of the machinery is built and stays: dynamic and static lists, saved views, live re-evaluation, filtered export.
- Custom field values are **real indexed columns** (`cf_*`), not a key-value blob — see `backend/internal/modules/customfields/`. The seam that lists a record type's active custom columns already exists: `backend/internal/shared/ports/fieldcatalog/fieldcatalog.go` (`ActiveColumns`). That is the bridge to use.
- Tags exist as objects (`backend/internal/modules/collections/tags.go`), they are simply not filter criteria.

## What to build

1. A bridge that injects the active `cf_*` columns into the filter vocabulary per record type, using the field-catalog seam so it stays correct when fields are added or retired.
2. Tags as a filter criterion ("has tag X").
3. City as a filterable field. Today the address is unstructured JSON on both organization and person (`backend/migrations/core/0005_organizations.up.sql`, `0004_people.up.sql`) and there is no city column at all — so this is either a promoted, indexed column plus a backfill, or an indexed expression over the JSON. Promoting the column is the cleaner call.
4. More filter criteria on person, which today can only be filtered by owner.

## Done when

- [ ] A user can build a dynamic list from a custom field value, a tag, and a city.
- [ ] Retiring a custom field does not leave a broken saved segment.
- [ ] The filters work the same way in lists, saved views and filtered export.

**Size:** M.
**Spec:** `lists-views-segmentation.md` — AC-filters-and-views-3/4 already require filters to reach custom fields, so the build is behind the spec here, not ahead of it.

dungbanguyen:
[...]

PR #1281, opened 2026-08-14
feat(reports): a report can group by year, quarter or month, and win-loss reads it

tinnhlt-gra:
Closes the period half of **#696** (`Add time periods to reporting`). Roadmap row F1, wave 1 — the last row in that wave.

**Spec first:** [margince-foundation#1298](https://github.com/gradionhq/margince-foundation/pull/1298) pins REPORT-PARAM-4..7 and REPORT-VOCAB-1. Reporting's own **REPORT-N-1** states the per-key vocabularies *"must be pinned at ticket time before the catalog is buildable"*, so that lands first. **Merge that before this.**

## What ships

`win-loss` (**REPORT-KEY-8**, under its corpus name) is the first key to carry the period grains:

| | |
|---|---|
| base set | unarchived deals with status `won` or `lost` |
| dimensions | `status`, `source`, `owner_id`, `pipeline_id`, `organization_id`, `currency`, `period_year`, `period_quarter`, `period_month` |
| measures | `amount_minor` |
| filters | every dimension |
| defaults | grouped by `status`, count + sum |

An **open deal is absent** from it, not a zero in it — it has not been won or lost, and a zero would read as a fact.

## The design decision, and why it is not the one the roadmap predicted

The roadmap said the hard half would be that *"a bucket is a range, not the equality predicate the drill-through inverts."* That turned out not to be the problem. `derivationWhere` already binds `expr = $n` against the **dimension expression**, and `forecastCategoryExpr` has been proving a computed `CASE` dimension round-trips for months.

The real hazard is one layer down, in **serialization**. A handle carries a group key out through a URL query string and binds it back; `predicateString` renders a non-string via `fmt.Sprint`. So a timestamp bucket produces **two renderings of one value**:

```
wire cell:     period_year: 2025-01-01T00:00:00Z
its own handle: period_year=2025-01-01+00%3A00%3A00+%2B0000+UTC
```

The second is Go's default `time.Time` format, which Postgres will not parse. Every period report's "Explain This Number" would 500 — at exactly the moment a reader stops trusting a figure and asks what is behind it.

`to_char` makes the round trip **total instead of incidental**, and pays twice more: `'2025-Q4' < '2026-Q1'` sorts chronologically as plain text, so a period report needs no separate sort key, and the value a reader sees is the value that travels.

**That claim is mutation-checked, not asserted.** Swapping `to_char` for `date_trunc` makes `TestWinLossPeriodDerivationReconcilesExactly` fail on exactly the mismatch above.

## Timezone

Buckets are cut in the installation's reporting zone (data-semantics §2 r4) — the same zone the forecast's "today" uses, through the `reportZoneToken` bind that `bindInstallationZone` was already written to serve (its doc comment literally anticipates a report that *"buckets by date"*).

A deal closing **00:30 UTC on 1 Jan 2026 is a 2025 deal in New York**, and `TestWinLossPeriodBucketsFollowTheInstallationZone` pins it. Without that test the defect is invisible on every UTC installation — which is most fixtures.

## Constants lifted out of the surrounding code

Per rule 1 (*fix the invariant, not the call site*), rather than spell these a second time in the new file: `fieldAmountMinor`, the `aggFn*` vocabulary `aggregateSelect` switches on, and `aliasDeals`. `tableDeal` already existed in `replayscope.go` under a header reading *"one spelling each"*, so this uses it instead of declaring a fourth.

## Published surface

`docs/reference/mcp-info.{json,md}` regenerated: `run_report` derives its report enum and vocabulary description from this catalog, so it grew by one entry (**4.1 KB → 4.5 KB**, tool count unchanged at 38). The `TestPublishedMCPSurfaceMatchesWhatAClientIsServed` gate caught this and is why it is in the same commit.

## Deliberately NOT in this PR

- **Custom fields as report dimensions/measures** (#696 item 3) — that seam belongs to **#693**, and a second agent is near it.
- **The period parameter on the corporate-group rollup** (#696 item 4) — follow-up.
- **A re-record of the `agent_loop` aicert site.** The catalog change restamps `prompt_version`. Measured before and after: that site was **already stale on `origin/main`** with identical figures, and readiness is **16 of 27 both ways** — so this does not degrade certification state. Filed as a follow-up rather than folded in.

## Gates

- `make check` — **green**
- `make craft-static` — **0 blocker, 0 major, 0 minor** across `backend/`, `extensions/`, `fixtures/`
- full `internal/compose` integration package — **green** (44s), run whole rather than `RUN=`-filtered because the registry census gates cannot be seen by a subset
- new-code coverage — **100%** on all three new functions from unit tests alone

No migration, so no collision with concurrent work on migration numbering.

---
## Summary by cubic
Reports can now group by year, quarter, or month; `win-loss` uses these period buckets anchored on `closed_at`. Buckets render as canonical text in the installation timezone so drill-through round-trips cleanly and sorts chronologically; open deals remain excluded.

- Win–loss adds dimensions `period_year`, `period_quarter`, `period_month` and `source` (all also filters). Defaults: grouped by `status` and `currency`; aggregates `count` as `deals` and `sum(amount_minor)` as `amount_minor_sum`. Open deals are absent, not zeroes.
- Derivation handles now encode unset columns via an `isnull` key, so empty text is not conflated with NULL; the plain-language definition says “is not set.” Reserved handle keys are enforced.
- `run_report` filter semantics: NULL filters bind as `IS NULL` (aligned with drill-through). Only string and boolean filter values are accepted; numbers/objects/lists are refused with 422 and advice to quote periods (e.g., "2026").
- Bucket values are generated via `to_char(timezone(...))` (text) instead of timestamps, fixing handle round-trip and enabling lexical chronological sort. Integration tests pin round-trip integrity and timezone boundaries.
- `backend/internal/compose/reportperiod.go` adds the period vocabulary and `winLossSpec`; `report.go` lifts shared constants (`aggFn*`, `aliasDeals`, `fieldAmountMinor`, `tableDeal`). `run_report` schema lists `win-loss`; docs regenerated in `docs/reference/mcp-info.{json,md}`. No migration required.

<sup>Written for commit 034bc5a12b23d2f8c4161e3fb7992c194daaed39. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1281?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tinnhlt-gra:
## UAT — verified on a live stack, not only in tests

Run against an isolated `make dev DEV_SLUG=rperiods` stack (api `:18326`) rebuilt from this branch's final commit, seeded with won/lost deals across 2025 and 2026.

696-win-loss-period-mcp-uat

The recording is the **agent-facing** surface, which is the one this change moves: MCP Inspector connecting to `/mcp` with a `read`-scoped passport, `win-loss` appearing in the `run_report` enum beside the four existing keys, and the call returning revenue by year.

### 1. Revenue by year — the issue's headline ask

```
2025    2 deals      14300000 minor
2026    2 deals      15800000 minor
```

### 2. Closed-won by quarter — its second phrasing

```
2025-Q1     4500000
2025-Q3     9800000
2026-Q1    12500000
2026-Q2     3300000
```

Returned in chronological order **with no sort key** — the text buckets sort themselves.

### 3. Explain This Number, on a period bucket

```
definition: Over live (unarchived) deals that have been won or lost, bucketed by when they
            closed in the installation's reporting timezone (an open deal is absent from this
            report, not a zero in it), filtered to status = "won", within the group where
            period_year = "2025": ...
source rows : 4500000 + 9800000
recomputed  : {"revenue_minor": 14300000}
```

The drill-through agrees with the number it explains.

### 4. Refusals — a caller mistake must not read as an outage

| call | before | after |
|---|---|---|
| `group_by: ["period_week"]` | 422 | 422, naming the grains that do work |
| `filters: {"period_year": 2026}` (a number) | **opaque 500** | **422** — *"cannot compare a number against `period_year` — quote the value as text"* |

The second one matters most on the tool surface: an agent told "internal error, try again" retries something that can never succeed.

### 5. Both doors agree

| probe | before | after |
|---|---|---|
| `filters: {"owner_id": null}` vs the handle it mints | 0 vs 6 | **6 vs 6** |
| empty-`source` bucket vs the handle it mints | 1 vs 0 | **1 vs 1** |
| default plan across EUR + USD | one unit-less `30160000` | split per currency |

---

### One process note worth recording

The first pass of this UAT reported the numeric filter still 500ing **after** the fix. It wasn't — `make dev DEV_SLUG=<slug>` does not replace an already-running slug stack (the sweep deliberately spares slug stacks), so I was probing a binary 15 minutes older than the fix. `make dev-stop DEV_SLUG=<slug>` first, then `make dev DEV_SLUG=<slug>`. Exactly the "a stale api is indistinguishable from a broken feature" trap CLAUDE.md warns about, met from the one direction the warning doesn't cover.

tinnhlt-gra:
@coderabbitai full review

PR #1293, opened 2026-08-14
refactor(activities): a message records the transport that carried it, not its kind

tinnhlt-gra:
## Summary

`activity.kind` has been answering two questions at once: **what sort of interaction happened**
(email, call, meeting, note, task) and **which messaging transport carried it** (telegram). The send
path recovered the second by reading the first back as a provider name:

```go
provider := string(anchor.Kind)   // activities/channelsend.go
```

That equivalence is a coincidence, not a rule. It holds only while every transport is *also* spelled
as a kind, and it makes a transport that names no interaction kind unreachable — the derivation
answers "not a channel conversation" for a conversation that plainly is one.

So the transport becomes its own column: `activity.channel_provider`, referencing the
`channel_provider` registry. Every writer records it, and the send path reads it.

**`kind` is deliberately UNCHANGED.** It still carries `'telegram'`, every reader of it still works,
no contract enum moves, there is no frontend change and no new endpoint. Narrowing `kind` to its six
semantic members is a separate PR that depends on this column already being correct for every row —
which is why the migration backfills it *from the registry* and an integration assertion proves no
channel activity lacks one.

This is slice 1a of `.tmp/ext-core-write-port/DESIGN-SP5-channel-axis.md` (plan:
`PLAN-05-channel-axis.md`). It supersedes `DESIGN-SP4.md`, whose §4 shipped as #1264.

## What is in it

- **Migration `0247_activity_channel_provider`** — the nullable column with its FK into the registry;
  a backfill derived from the registry (`WHERE kind IN (SELECT provider FROM channel_provider)`),
  exact today *precisely because* the old model stored the provider in the kind; and the
  `^[a-z][a-z0-9_]*$` + 32-char grammar CHECK that `ProviderRef` already documented but the column
  never enforced.
- **`channel_provider_provider_fkey` dropped.** `channel_provider.provider REFERENCES
  activity_kind (kind)` is the conflation in schema form — it asserts every transport is also an
  interaction kind, which refuses any provider that names no kind.
- **The boot reconcile stops minting `activity_kind` rows.** It only ever did so to satisfy that FK.
  Left in place, the next boot after the kind-narrowing migration would put the removed rows straight
  back, so that slice's fitness assertion would pass or fail depending on whether a boot had run.
- **The write spine** — `capture.ActivityFields` and `activities.LogActivityInput` gain
  `ChannelProvider`; both INSERTs carry it; telegram's normalizer, the ingest bridge, the outbound
  reply, and the REST/agent mapping all set it.
- **The read side** — `provider := string(anchor.Kind)` is gone, replaced by `channelProviderOf`, a
  narrow row-scope-gated read. An empty provider answers the same `NotAChannelConversationError` as
  before, reported by the anchor's **kind**, because that is the word a rep sees on the record.

## Review

Reviewed by **Fable and Codex** independently, against the diff. Both raised the same central
finding, and it was a real behavioural regression the tests had not caught:

- **Writers could still mint a channel-kind row with a NULL transport.** `POST /v1/activities` and the
  `log_activity` agent tool both accept `kind: "telegram"` (`recordshapes_gen.go:32` advertises it),
  and extension ingress passes a unit-supplied kind through. A row written that way would have been
  unrepliable, while an identical row written *before* this migration stayed repliable because the
  backfill covered it — same data, different behaviour decided by write date. Fixed by recording the
  transport at the write, in the one place REST and the agent surface share
  (`LogActivityInputFrom`) plus the ingress bridge, with mirrored tests.

Fable additionally found, all fixed:

- **The migration could have bricked a deploy.** An earlier draft asserted zero `whatsapp` rows on the
  premise that nothing writes that kind. False — it is contract-valid on `POST /v1/activities` and
  advertised to the agent, so any installation that had ever hand-logged one would fail the migration
  with no repair path. The assertion also bought nothing: `whatsapp` has never been a
  `channel_provider` row, so such an activity was not a channel conversation before this change and
  is not one after. Removed, with the reasoning recorded in the migration.
- **Two comments asserted the opposite of the shipped code**, and the index they justified had no
  reader — the send path resolves by primary key and capture's reply-match still selects on
  `(thread_key, kind)`. Comments corrected and the index dropped; the slice that moves the
  reply-match onto the column is the one that should add it, with the query that uses it.
- **A doc-comment split** — `channelProviderOf` had been inserted between `resolveConversation`'s doc
  block and its function, so godoc would attribute the wrong comment. Moved.
- A stale `datasweep.go` comment naming the dropped FK, and an FK test that asserted only `err != nil`
  while using a real extension unit's name as its "unregistered" provider.

Codex additionally found:

- **A migration-number collision** with `0246_agents_drop_workspace` on `origin/main` (this is the
  third collision this branch has hit — `0244`, `0246`, now settled at `0247`).
- **The down migration restored the provider→kind FK blindly.** Since the reconcile no longer mints
  kind rows, any installation whose composed set grew a provider would fail rollback with a raw
  constraint error. It now re-creates the kind rows for exactly those providers first. Verified by
  reproducing the failing state (a registered provider with no matching kind) and rolling back
  successfully.

## Gates

- `make check-backend` — green (its earlier `contract-breaking-check` failure was purely a stale base;
  `origin/main` had moved).
- `make test-integration` — **OK: integration passed with 0 skips (2440 tests, 40 packages)**.
- `make craft-static` — 0 blocker, 0 major, 0 minor across `backend/`, `extensions/`, `fixtures/`.
- Pre-push hook — craft-static, `rls-store-path`, `no-jurisdiction` all OK.
- **New-code coverage: 92.5%** (99/107 changed statement lines), measured against the merge-base with
  the integration lane's own covdata merged with a unit profile. The 8 uncovered lines are error paths
  in `channelProviderOf`, including a defensive row-scope re-probe that `GetActivity` already refuses
  ahead of — matching `resolveConversation`'s documented re-probe idiom. Reaching them would require
  bypassing the real call order.
- `make e2e-ai-report` — this diff touches no prompt, corpus or tool-schema file, so no certification
  record can be staled by it. (The 17 stale / 7 absent records are pre-existing on `origin/main`.)

## UAT

On a live isolated stack (`DEV_SLUG=sp5axis`, fresh database), through HTTP and psql:

```
1. the migration is applied on this live installation
  PASS activity.channel_provider exists
  PASS channel_provider carries its grammar CHECK
  PASS the provider->kind FK is gone (a transport is not an interaction kind)

2. boot registered the transport and minted NO interaction kind for it
     channel_provider: telegram
     activity_kind:    call,email,meeting,note,task,telegram,whatsapp
  PASS telegram is registered as a transport
  PASS activity_kind is exactly the migration's seed — boot added nothing

3. logging in
  PASS authenticated as the bootstrap admin (session cookie)

4. a hand-logged CHANNEL activity records its transport
     activity 01a0012a-e07e-... -> kind=telegram channel_provider=telegram
  PASS the transport was recorded at the write (this is the reviewed regression, closed)

5. a hand-logged NON-channel activity records no transport
     activity 01a0012a-e116-... -> kind=note channel_provider=<NULL>
  PASS no transport recorded, as a note never travelled on one

6. the send path READS the column: a reply on an anchor with no transport is refused
     POST /v1/activities/01a0012a-e116-.../send-message -> HTTP 422 code=not_a_channel_conversation
     detail: a note activity is not a messaging-channel conversation; reply on the channel the
             conversation was held on
  PASS refused as not a channel conversation, decided by the column
  PASS the refusal names the anchor's kind, so the rep can act on it

RESULT: ALL CHECKS PASSED
```

Script and captured output: `.tmp/ext-core-write-port/uat/uat-sp5-1a.{sh,out}`.
Fresh-install boot recording: `.tmp/ext-core-write-port/uat/sp5-1a-fresh-install-boot-uat.gif`.

**A live Telegram send/receive is NOT part of this UAT** — no bot credentials exist in this
environment, the same limit #1264 recorded. What is proven above is the migration on a real boot, the
reconcile's new behaviour, the write-side recording, and the live refusal path. A real
unit-supplied-transport send is the acceptance test of the slice that builds one.

## Deliberate deviations from the plan, with reasons

- **Capture's reply-match stays on `kind`.** The plan said move it to the provider. Reading the code
  disproved that: `kind = $2` is a **security control**. `thread_key` is one flat namespace holding
  both a channel's `<provider>:<bot>:<chat>` key and a mail thread root, and the mail half is
  *attacker-supplied* (the sender's own `References` root) — the medium match is what stops a forged
  header manufacturing a reply fact against a Telegram conversation. A straight swap also silently
  breaks mail, which has no provider. It moves in the narrowing slice, where it is forced and testable.
- **`activityColumns` does not select the column.** Nothing reads the provider off the contract
  `Activity` shape until that shape has a field for it. Selecting an unconsumed column in the module's
  hottest read would be dead weight.

## Notes

- The behavioural gate on the old derivation is a test, not a grep:
  `TestSendMessageRefusesAChannelKindThatCarriesNoProvider` seeds `kind='telegram'` with a NULL
  provider and requires a refusal. **Mutation-verified** — restoring `provider := string(anchor.Kind)`
  fails it with its own diagnostic.
- The "every channel activity carries a transport" assertion lives *inside* the shared
  `capturedMessage` read-back helper, so every existing telegram ingest test proves it through the
  real writer rather than a hand-inserted row.
- `compose/extingressdrift_test.go` gains a waiver path in its core→published direction, which it did
  not have. Publishing `ChannelProvider` on the extension surface now would put a field there no unit
  can meaningfully set until it has a declared channel; the waiver records why the empty value is the
  accurate answer, and commits a future channel-supplying unit to publishing it.

---
## Summary by cubic
Records each message’s transport in `activity.channel_provider` and makes the send path read it. Previously the send path derived transport from `activity.kind`; now it refuses replies when `channel_provider` is empty. `kind` is unchanged.

- Adds `activity.channel_provider` referencing `channel_provider`; backfills from the registry (`UPDATE activity SET channel_provider = kind WHERE kind IN (SELECT provider FROM channel_provider)`).
- Drops `channel_provider.provider → activity_kind(kind)` FK; adds `channel_provider` name grammar CHECK (`^[a-z][a-z0-9_]*$`, ≤32 chars).
- Replaces `provider := string(anchor.Kind)` with a row-scoped `channelProviderOf` read; empty provider returns `NotAChannelConversationError` named by the anchor’s kind.
- Writers now record the transport:
  - `activities.LogActivityInput` and `capture.ActivityFields` gain `ChannelProvider`.
  - REST/agent mapping sets it via `activities.ChannelProviderForKind(req.Kind)`.
  - Telegram normalize and compose ingress set `ChannelProvider`; `capture.Sink` inserts it (NULL when empty).
- Boot reconcile stops inserting `activity_kind` rows; it only upserts `channel_provider`.

**Rollout/Migration**
- Run core migration `0247_activity_channel_provider` (adds column, backfills, adds grammar, drops FK). Down migration recreates missing `activity_kind` rows before restoring the FK.
- No API or UI change; `kind` continues to carry channel names until a later narrowing slice.
- Action required only if composed provider names violate the new grammar; rename providers before boot or the reconcile will fail.

<sup>Written for commit c8ee530fe16d6edc3906bf54907fd5be5cf9f203. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1293?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/internal/compose/channelprovider.go
+++ b/backend/internal/compose/channelprovider.go
@@ -3,11 +3,16 @@
 
 package compose
 
-// The boot-time reconcile for the derived channel vocabulary (DESIGN-SP4 §4):
-// the ONE place activity_kind and channel_provider are written, and the ONE
-// place activities.SetChannelProviders and comms.SetChannelProviders are
-// called — so the DB registry and both packages' in-memory snapshots cannot
-// be set two different ways.
+// The boot-time reconcile for the derived channel vocabulary: the ONE place
+// channel_provider is written, and the ONE place
+// activities.SetChannelProviders and comms.SetChannelProviders are called — so
+// the DB registry and both packages' in-memory snapshots cannot be set two
+// different ways.
+//
+// It does NOT write activity_kind. A provider names a transport; an activity
+// kind names what sort of interaction happened. Those are different axes, and
+// the vocabulary of interaction kinds is fixed by the contract and seeded by
+// the core migration, so boot has nothing to add to it.
 //
 // Called from NewCaptureRegistry, which this codebase already constructs more
 // than once per process (a role-specific alternate wiring path, the worker's
@@ -25,10 +30,9 @@ import (
 	"github.com/gradionhq/margince/backend/internal/platform/database"
 )
 
-// reconcileChannelProviders inserts an activity_kind row and a matching
-// channel_provider row (transport='core') for every name in providers not
-// already present, then sets both packages' in-memory snapshots to exactly
-// the composed set passed in.
+// reconcileChannelProviders inserts a channel_provider row (transport='core')
+// for every name in providers not already present, then sets both packages'
+// in-memory snapshots to exactly the composed set passed in.
 //
 // It runs over database.WithInfraTx, not the workspace-bound database.DB.Tx:
 // activity_kind and channel_provider carry no workspace_id, so binding a
@@ -37,12 +41,10 @@ import (
 // with no organization bootstrapped yet, which is exactly when a process
 // first constructs this registry.
 //
-// activity_kind is inserted FIRST, in the same transaction: channel_provider
-// FKs into it, so a provider this binary just compiled in but that the core
-// migration's fixed seed never anticipated (a future core channel connector
-// beyond telegram) would otherwise fail this very insert with a foreign-key
-// violation — the exact deployment fact this reconcile exists to establish,
-// self-inflicted.
+// A provider name has to satisfy channel_provider's own grammar constraint,
+// which is where an unusable name is refused — not here. The alternative, a
+// check in Go beside the insert, would be a second spelling of the rule that
+// could disagree with the column's.
 //
 // It never DELETEs. A provider whose supplier is gone on a later boot keeps
 // its row — activity and person_channel_identity rows still reference it, the
@@ -51,11 +53,6 @@ import (
 func reconcileChannelProviders(ctx context.Context, pool *pgxpool.Pool, providers []string) error {
 	err := database.WithInfraTx(ctx, pool, func(tx pgx.Tx) error {
 		for _, provider := range providers {
-			if _, err := tx.Exec(ctx, `
-				INSERT INTO activity_kind (kind) VALUES ($1)
-				ON CONFLICT (kind) DO NOTHING`, provider); err != nil {
-				return err
-			}
 			if _, err := tx.Exec(ctx, `
 				INSERT INTO channel_provider (provider, transport) VALUES ($1, 'core')
 				ON CONFLICT (provider) DO NOTHING`, provider); err != nil {
--- a/backend/internal/compose/channelprovider_axis_integration_test.go
+++ b/backend/internal/compose/channelprovider_axis_integration_test.go
@@ -1,0 +1,107 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+//go:build integration
+
+package compose
+
+// The transport axis is separate from the interaction axis, asserted against the
+// schema rather than against anyone's memory of it. The behavioural halves live
+// where the behaviour is — the send path's refusal in
+// internal/modules/activities, the boot reconcile's in
+// channelprovider_integration_test.go — and what is left here is the shape those
+// depend on.
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/jackc/pgx/v5/pgconn"
+
+	"github.com/gradionhq/margince/backend/internal/compose/integration"
+)
+
+// A provider is a transport and names no interaction kind, so channel_provider
+// must not reference activity_kind. The FK it originally shipped with asserted
+// the opposite, and it is the reason boot had to mint an activity_kind row per
+// provider — which would silently restore whatever a kind-narrowing migration
+// removed, on the next boot.
+//
+// Introspection rather than a behavioural probe because the failure this guards
+// is a RE-ADD: an FK restored by a later migration breaks nothing until an
+// extension registers a provider that is not also a kind, which is far too late
+// to learn it.
+func TestChannelProviderDoesNotReferenceActivityKind(t *testing.T) {
+	integration.Setup(t) // triggers EnsureSchema — the migrations this test asserts on
+	owner := integration.OwnerConn(t)
+
+	var referenced []string
+	rows, err := owner.Query(context.Background(), `
+		SELECT confrelid::regclass::text
+		FROM pg_constraint
+		WHERE conrelid = 'channel_provider'::regclass AND contype = 'f'`)
+	if err != nil {
+		t.Fatalf("querying channel_provider's foreign keys: %v", err)
+	}
+	defer rows.Close()
+	for rows.Next() {
+		var target string
+		if err := rows.Scan(&target); err != nil {
+			t.Fatalf("scanning a foreign key row: %v", err)
+		}
+		referenced = append(referenced, target)
+	}
+	if err := rows.Err(); err != nil {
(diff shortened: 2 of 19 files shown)


tinnhlt-gra:
@coderabbitai review

tinnhlt-gra:
**CodeRabbit did not review this PR**, and its green check does not say so.

Both its automatic pass and an explicit `@coderabbitai review` returned `Review rate limited` / `Review limit reached` — an account-level quota, not something this branch can clear. The check reports `pass` because the bot exited cleanly, not because it read anything.

Recording it so the review coverage on this change is not overstated. What did review it:

- **Fable** and **Codex**, independently, against the diff — both found the same central defect (writers could still mint a channel-kind row with a NULL transport, a real behavioural regression), plus a migration that could have bricked a deploy on contract-valid `whatsapp` data, a blindly-restored FK in the down migration, a third migration-number collision, and a doc-comment split. All fixed in the pushed commit; the findings and their fixes are itemised in the PR description.
- Every deterministic gate: `check-backend`, `craft-static` (0/0/0), the full integration lane (2440 tests, 0 skips), `deterministic-gates`, `live-boot`, `integration (RLS + erasure + HTTP e2e)`, `govulncheck`, `secret-scan`, `dco`, SonarCloud quality gate.
- New-code coverage 92.5%, measured against the merge-base rather than `origin/main`'s moving tip.

Merging on that basis. If CodeRabbit's quota frees up and it flags something here later, it should be handled as a follow-up rather than left unread.

PR #1296, opened 2026-08-14
feat(automation): a drafted email waits in the inbox until a human sends it

### LINKED ISSUE #707: Give automation-drafted emails a real held-drafts inbox
## What is missing

When an automation drafts an email, the draft is recorded as JSON on the automation's run record. There is no drafts queue a human opens, reads, edits and releases. "We drafted it for you" is currently only visible if you go looking at automation runs.

## Why it matters

Draft-never-send is our whole posture for outbound automation, and it only pays off if the human actually finds the draft. Several requested features ("draft a notification and hold it for release") assume this queue exists.

- The action itself: `backend/internal/modules/automation/handlers_actions.go` (`applyDraftEmail`) — it composes through the comms seam and never sends. Keep that.
- Approval staging, decisions, edit scope and single-use decision tokens: `backend/internal/modules/approvals/`. If a draft is modelled as a staged item, most of the release machinery already exists.
- Email drafting, sending and the transmit-time gates: `backend/internal/modules/comms/` and `backend/internal/modules/activities/email.go`.
- The approval inbox screen: `frontend/src/screens/inbox.tsx`.

## What to build

1. A first-class held draft: durable, listable, editable, with the record it belongs to and the automation that produced it.
2. A drafts surface — either its own queue or a section of the existing inbox; prefer reusing the inbox rather than inventing a second place for waiting work.
3. Edit-then-release, with the send going through the normal gates and audit.
4. Discard, with a reason, so we learn what the automation gets wrong.

## Done when

- [ ] An automation-drafted email appears somewhere a human naturally looks.
- [ ] They can edit it and release it, and the send is audited like any other send.
- [ ] Nothing is ever sent without that human action.

**Size:** S–M.
**Spec:** `automation.md` AUTO-PARAM-2 (draft, never auto-send) and `drafting.md`.

tinnhlt-gra:
Closes #707.

An automation's `draft_email` action composed a message and left it as JSON on the run record. Nothing opened it, so *"we drafted it for you"* was true only for someone who went looking at automation runs.

## Where it goes was already decided

Four ratified pins put a held draft in the approval inbox, so this builds no new queue:

| Pin | Says |
|---|---|
| `automation.md` AUTO-PARAM-4 | 🟢 auto-run: … **draft** … · 🟡 held for approval: **send** … |
| `automation.md` AUTO-AC-1 | the 🟡 ones are *"held in the approval inbox"* |
| S-E15.2 / AUTO-AC-7 | *"its draft is created (🟢) and the send waits for my approval (🟡)"* |
| `drafting.md` acceptance | *"an agent's proposed send waits in the approval inbox"* |

AUTO-NOTE-1 names the object outright: *"a send is the approval-gated completion of a draft-email action."* So drafting still auto-runs and now hands the send it proposes to the staging path `ActionSendEmail` already used.

## Releasing it is the ordinary send

The effect calls the same store entry point `POST /activities/{id}/send-email` calls, so consent (default-deny per purpose), recipients, attachments, the sign-off, deliverability, the audit row and the outbound activity are the ones any send produces. A released draft is not a special kind of mail, and anything derived here would be a way for it to differ from the message the same human could have typed. `crm.yaml:16973` already documented this as the release path.

**Three things commit together or none do:** the single-use redemption that spends the human's authority, the send, and the parked run's completion. Each pair alone is a defect this codebase can already name — redemption without the send is the consumed approval whose effect never ran (unrecoverable for mail); the send without the run transition leaves history claiming an automation still waits for a decision that went out days ago. A gate that refuses rolls the redemption back with it, so the approval stays **unconsumed and releasable** once the cause is fixed.

## Read this bit — it is not drafts work

**`recordApplyOutcome`'s staged arm wrote only `detail`, never `applied`.** The moment `draft_email` began staging, the draft it had just composed would have been silently dropped: run history reporting that an automation drafted a reply while holding nothing to show for it. Fixed here, and the integration test asserts it through real Postgres.

**A locked re-read of the anchor now lives on the shared write path.** The version pin never covered an archive landing between validation and the write, and `threading()` does not filter archived rows — so a reply could join a thread whose root was gone. The immediate send and the scheduled fire get it too (rule 1: fix the invariant, not the call site).

## What the issue did not mention, and this needed

- **A deliverable address.** Every other caller supplies `to`, because the MCP flow has an agent in the middle; an automation has nobody. `ReplyAddressFor` reads the counterparty off the thread's participants and **excludes our own users** — on an outbound anchor the `from` participant is us, and a rank that missed that would mail the message back to its sender. It refuses where the greeting resolver beside it shrugs.
- **A consent purpose**, declared per automation and never defaulted in code: sending is default-deny per purpose, so a default here would be this process picking a lawful basis for the operator. The seeded recap template names `business_correspondence` — the purpose ADR-0098 D1 added for answering somebody individually.
- **Its own staging kind.** Sharing `send_email` would waive that kind's target version pin for **agent** sends too. `held_draft` takes the waiver alone, with the rationale the three existing waivers use.
- **Late registration.** `applySendPath`, not the effect list — the send store is assembled from server options applied *after* the approvals surface is built, so registering it beside the others binds a store with no signature and no unsubscribe linker. Mail that goes out worse than the same human's own, with nothing failing to say so. The version-pin census reads the same table the wiring does, so the two cannot drift.

## Deliberately not here

- **The TTL.** `stagingTTL` is 24h against APPR-PARAM-1's *"72h default, per-item override"*. Changing the constant alone is a half-fix: without an expiry sweeper that persists the terminal status, emits the system audit/event, releases quota and blocks the linked run, it only widens the window in which an expired draft parks its run forever. Filing whole.
- **The generic redemption slice** (`automation/seams.go:43`). Closed here for this kind only; every other 🟡 automation action still never executes on approval. Filing.
- **Erasure/SAR reach into approval payloads.** Pre-existing, widened by this. Filing.

## Verification

| Gate | Result |
|---|---|
| `make check` | **exit 0** |
| `make craft-static` | **exit 0**, no findings |
| `make test-integration` | **exit 0** — `OK: integration passed with 0 skips (40 packages)` |
| Contract / migrations | **none** — release rides the existing decide endpoint, `expires_at` already exists |

Six new integration tests against real Postgres: release sends + completes the parked run; a second approve produces no second message; a refusal leaves the approval unconsumed and sends nothing; and the three address-resolution cases including the mail-yourself one.

Design notes and both adversarial design reviews (Fable + Codex, run before implementation) are in `.tmp/held-drafts/DESIGN.md`.

---
## Summary by cubic
Drafted emails from automations now wait in the approval inbox as `held_draft` and send only after a human releases them. Previously `draft_email` wrote a draft with no send path; release now uses the ordinary send flow with atomic redemption, send, and run completion, and refusals leave the approval pending.

- Staging: `draft_email` stages `held_draft` with `JoinPending`; approvers need the same grant as `send_email`.
- Release: late-registered effect in `applySendPath` calls `activities.PrepareSend` then `activities.SendPreparedTx`; a precheck runs before the decision to catch consent, mailbox, or archived-thread refusals and keeps the approval releasable.
- Addressing: `activities.ReplyAddressFor` resolves the counterparty (excludes our users) and prefers a person’s primary email; compose forwards via `automation.Comms.ReplyAddress`.
- Drafting: `draft_email` now requires `consent_purpose` and records its artifact on `applied` even when staged; the seeded post‑meeting recap uses `business_correspondence`.
- Send-path hardening: `SendOrigin.lockAnchorLive` re‑reads the anchor with a share lock; send core is exported as `PrepareSend` and `SendPreparedTx` and used by immediate and scheduled sends.
- Editing: approvers may edit only subject and body; `to` and `consent_purpose` cannot change, and the approvals API rejects retargeting edits.
- UI: the inbox renders `held_draft`; subject and body are editable, with the body in a textarea; i18n labels added.

**Rollout**
- No database migrations or API shape changes.
- Automations using `draft_email` must provide `consent_purpose` (e.g., `business_correspondence`).
- Ensure email send configuration is set (public base URL, signature, unsubscribe).

<sup>Written for commit 54699cc6f1a6d40d99ec44dd4f5bc0fc0d029dda. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1296?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tinnhlt-gra:
## UAT — live stack, real automation, real database

Not a hand-staged approval. A meeting was logged through `POST /v1/activities`, which fired the seeded `post_meeting_recap` automation: it composed the draft, resolved the counterparty off the thread's participants, and staged a `held_draft` summarised as `an automation drafted a reply to anna@example.com - read it before it goes`.

### 1. The inbox renders it, edits it, and refuses honestly

held draft: edit and release

Subject renders as a line and **Message as a textarea** carrying the whole composed draft. The fields that are not the question — `to`, `consent_purpose`, `anchor_activity_id`, `intent` — are not offered. I edited the body and pressed **Approve edited**.

It then refused: **reconnect your mailbox to enable sending**. This cold stack has no mailbox connected, so the preflight caught it — and the draft is **still Pending**, with Accept/Edit/Reject intact and the badge still at 1.

That is the blocker both reviews found, demonstrated live. Before the preflight, this same refusal committed the decision first and left an approved row nothing could re-drive: the message would have been gone. It was unplanned, and it proves the fix better than a happy path would.

### 2. Discard with a reason

held draft: discard with a reason

### 3. What the database holds afterwards

```
approval      status=rejected  decision_reason='Too soon after the meeting - I will follow up myself next week.'
workflow_run  post_meeting_recap  status=blocked  reason='approval ... was rejected by the deciding human'  has_artifact=t
workflow_run  post_meeting_recap  status=skipped  (the email activity was not a meeting - Match works)
```

**Discard with a reason lands** — issue item 4, verbatim in `decision_reason`. **The run terminates on reject** instead of parking forever. And **`has_artifact=t` on that blocked run**: the composed draft survived onto `workflow_run.applied` through parking AND rejection — the regression the first review found, where the staged arm wrote only `detail` and would have dropped the artifact silently. Shown here on a real row.

### Not covered here, and why

The successful transmit needs a send-capable mailbox (real OAuth), which a cold dev stack cannot have. That half is covered in the integration lane against real Postgres: one outbound activity, one delivery row, the parked run completing, single-use on a second approve, and full rollback on a refusal. `make test-integration` → `OK: integration passed with 0 skips (40 packages, parallel)`.

tinnhlt-gra:
@coderabbitai review

tinnhlt-gra:
Follow-ups filed for what this PR deliberately left out, each with the evidence from the two reviews:

- **#1303** — approval TTL is 24h with no override and no expiry transition (APPR-PARAM-1 says 72h). Deliberately not fixed here: changing the constant alone widens a window in which an expired staging still parks its run forever, because expiry is computed lazily and never persisted. Worth knowing for this feature: a draft composed Friday afternoon expires before Monday.
- **#1304** — an approved 🟡 automation action never executes and its run parks forever. Closed here for `held_draft` only; generalising it would change behaviour for `advance_deal`, `send_email`, `emit_flow_event` and at-scale `assign_owner`.
- **#1305** — erasure and subject-access do not reach approval payloads or workflow runs. Pre-existing, but this PR makes such rows routine and they now carry an addressee and a full message body.

tinnhlt-gra:
@coderabbitai review

Issue #1298, opened 2026-08-14
A seat downgrade leaves standing write record_grants, and nothing will notice when a seat-change surface arrives


tinnhlt-gra:
## The invariant, and its single guard

`AAD-AC-4` says a read-seat user may not RECEIVE a write `record_grant`. That is now enforced —
`refuseWriteGrantToReadSeat` in `backend/internal/modules/identity/grants.go`, shipped with
#711 — but **only at grant creation**.

Nothing revokes an existing write grant when a seat is downgraded. `app_user.seat_type` carries only
the generic timestamp trigger, and the row-scope predicate
(`backend/internal/platform/auth/rowscope.go`) treats any unexpired grant as live without consulting
the subject's current seat.

## Why this is not a live defect today

- No online surface writes `seat_type` — only the bootstrap and the test harness. There is no
  downgrade to race.
- While the seat is `read` the stale grant is **inert**: REST mutations die at the ceiling
  (`identity/handlers.go`), agent calls at `platform/auth/admit.go`.

So no authority leaks now. What survives a downgrade→upgrade cycle is a grant that was legitimately
created under a full seat, and arguably should survive.

## Why it still deserves an issue

The invariant has exactly one guard, at one door, and the second door is coming: the moment a
seat-change endpoint lands, whoever writes it has to decide — revoke write grants on downgrade, or
state in writing that a downgrade parks them and an upgrade restores them. That is review-loop rule 1
(fix the invariant, not the call site), and the cheapest time to decide it is before the endpoint
exists rather than in the PR that adds it.

**Done when:** either a downgrade revokes (or expires) the subject's write grants and a test proves
it, or the parked-and-restored reading is written into `access-and-admin.md` beside AAD-AC-4 and a
test pins THAT.

dungbanguyen:
## Decision, 2026-08-31: a seat downgrade revokes standing write grants

Re-triaged against `origin/main` at `8ecb691f1`. Confirmed:
`refuseWriteGrantToReadSeat` exists at
`backend/internal/modules/identity/grants.go:291-307` and is called from `:196` —
**at grant creation only**. Nothing revokes an existing write grant when a seat
changes, and the row-scope predicate treats any unexpired grant as live without
consulting the subject's current seat.

Also confirmed not a live defect: no online surface writes `seat_type`, and while
the seat is `read` the grant is inert — REST mutations die at the seat ceiling,
agent calls at the admission gate.

### The ruling, taken now rather than by whoever ships the endpoint

[...]

Issue #1297, opened 2026-08-14
createRecordGrant is documented as idempotent but answers 409 on re-assert


tinnhlt-gra:
## What the contract says

`crm.yaml` `createRecordGrant`:

> Idempotent on `(record_type, record_id, subject_type, subject_id)` — re-asserting upgrades/downgrades `access` and resets `expires_at`.

The operation also declares the `Idempotency-Key` parameter.

## What the code does

`backend/internal/modules/identity/grants.go` `CreateRecordGrant` runs a plain `INSERT` with **no
`ON CONFLICT DO UPDATE`**. A re-assert of the same tuple hits the unique constraint
(`record_grant_unique`, restated in `migrations/core/0225_collapse_composite_keys.up.sql`) and is
mapped to `apperrors.ErrConflict` → **409**.

So the documented upgrade/downgrade path does not exist. A client following the contract — "share
this record again with write to widen it" — gets a conflict, and the only way to change an access
level today is revoke-then-re-grant.

## Why it is worth fixing rather than re-documenting

The sharing UI has no other way to change an access level, and 409 is not a refusal a caller can act
on here: nothing about the request was wrong, and the tuple the caller must first revoke is not named
in the response.

Found while building the seat × role × action matrix
(gradionhq/margince-poc-v1#711): the matrix's share cell silently 409'd for every role after the
first, which is also why that test now resets its fixtures between cells.

## Deciding first

Which half is right — the contract's upsert or the code's conflict — is a product call about what
re-sharing means. If the contract wins, the fix is an `ON CONFLICT (record_type, record_id,
subject_type, subject_id) DO UPDATE`, and the **seat check must stay ahead of it** so a read seat
cannot be upgraded into a write grant by a second call. If the code wins, the contract prose and the
`Idempotency-Key` parameter both come off.

PR #1299, opened 2026-08-14
test(auth): the seat ceiling is proven over every seeded role, and the refusal it exposed now exists

### LINKED ISSUE #711: Return the seat_tier_insufficient error, and test the seat x role x action matrix
## What is missing

Read seats are enforced — a read-seat user genuinely cannot write — but two specified pieces are absent:

1. The specific error code `seat_tier_insufficient` does not exist anywhere in the backend. A client cannot tell "your seat does not allow this" apart from other permission failures.
2. There is no test covering the seat type × role × action matrix. Role permissions are tested; the seat ceiling on top of them is not.

## Why it matters

The seat ceiling is a commercial boundary as well as a security one, and it sits *below* role permissions — even an admin on a read seat must be refused. That is exactly the kind of rule that quietly breaks and is exactly the kind of test we do not have. The error code matters because a UI cannot show the right message without it.

- The seat column and its constraint: `backend/migrations/core/0002_identity.up.sql` (`seat_type` in read/full, plus a constraint that an agent user must hold a full seat).
- Seat resolution at the single admission point: `backend/internal/platform/auth/admit.go`.
- The architecture test that pins the layering: `backend/rbacgate_test.go`.
- The contract already carries the seat type with fail-closed semantics (an unknown value must be treated as a read seat).

## What to build

1. Return the specified error code on a seat-tier refusal, with the right HTTP status.
2. A matrix test: for each seat type, each role, and each action class (mutate, send, advance, approve, export), assert the outcome — including that an admin on a read seat is still refused.
3. Cover the two related refusals named in the spec: a read-seat user receiving a write record-grant, and an agent credential being bound above the granting human's seat.

## Done when

- [ ] A read-seat user attempting a write gets the specific documented error.
- [ ] The matrix test fails if anyone weakens the ceiling.

**Size:** S.
**Spec:** `access-and-admin.md` AAD-AC-4 and the error registry entry it names.

tinnhlt-gra:
Closes #711.

## What was actually missing

The issue's item 1 — the `seat_tier_insufficient` code — **already ships**: `httperr.go:50` maps the
sentinel and `identity/handlers.go:334` raises it for every mutating REST method on a read seat. What
was owed is the matrix, and what building the matrix found is a refusal that is specified and was
absent.

## The matrix

`AAD-AC-4` asks for `seat_type × role × action`. What existed was two verbs on one role, which cannot
state the claim that matters: **the ceiling sits BELOW role permissions** — a read seat is refused
whatever its role grants, and a full seat then falls through to RBAC.

The rows come from the database — `select key, permissions from role` over the seeded system roles,
the technique #1158 proved — and the **full-seat expectation is derived from that same grant
document**, so re-granting a role or seeding another moves the test with it rather than leaving a
literal behind.

**5 roles × 5 action classes** (create, update, advance, export, share):

| leg | cells | assertion |
|---|---|---|
| read seat | 25 | `403 seat_tier_insufficient`, every role, every action |
| full seat, granted by the role doc | 21 | **succeeds** (2xx) |
| full seat, not granted | 4 | `permission_denied`, never the seat code |

Both counts are asserted: a matrix that fell entirely on one side would prove half of what it claims,
and would narrow silently the day a policy changed.

## The refusal it exposed

`AAD-AC-4` says a read-seat user **receiving** a write `record_grant` is rejected with the seat
sentinel. `CreateRecordGrant` checked the record type, the grantor's RBAC, the record's visibility and
the subject's existence — and never the subject's seat. The word did not appear in the file. So a
full-seat admin could hand a read-seat colleague a write grant that widened their scope onto a record
every write to which the ceiling then refused: **authority that reads as real and cannot be
exercised.**

`refuseWriteGrantToReadSeat` closes it. A `read` grant to a read seat stays legal — that is the
authority their licence already carries. A `team` subject is deliberately not filtered: a team is not
a seat, and refusing a whole team because one member reads is a wider rule than the AC states; its
read members are still refused at their own admission.

## What review changed

Both reviewers independently found the matrix could go green for the wrong reason, and they were
right on every count:

- the full-seat leg asserted only "not the seat code", never success — so the **export cell was
  `422` for all five roles** (a filterless object export is refused before RBAC runs) and counted as
  a pass, and the **share cell `409`'d** for every role after the first on the unique tuple;
- fixtures carried one role's successful write into the next role's cell;
- `SetWorkspaceSeat` flips *every* human, recipient included, so the share cell's read-seat leg was
  ambiguous between the actor's ceiling and the recipient's — the same sentinel from two rules.

Fixed: granted cells must succeed, the export sends a real `owner_id` filter through a CSV-safe
reader, fixtures reset between cells, and the share recipient is pinned to a full seat so each cell
is a statement about the actor.

## Raised, not patched

- **#1297** — `createRecordGrant` is documented as idempotent (re-assert upgrades access) and answers
  `409`; there is no upsert path. Which half is right is a product call, and the seat check must stay
  ahead of whichever wins.
- **#1298** — a seat downgrade leaves standing write grants. Inert today (nothing writes `seat_type`
  online, and the ceiling blocks the writes anyway), but the invariant has one guard at one door and
  the second door is coming.
- `AAD-AC-4`'s last clause — a Passport bound above the `{read}` ceiling is `422 scope_exceeds_grantor`
  — is **unreachable as specified**: minting is a `POST`, so the method ceiling already refuses a read
  seat *any* passport, including a read-only one. That is stricter than the spec, not weaker; making
  the 422 reachable is a product decision about delegation, not a defect to patch.

## Gates

`make check` (backend + frontend), `make craft-static`, and the full integration lane —
`OK: integration passed with 0 skips (40 packages)` — green on this branch, rebased on current `main`.

---
## Summary by cubic
Proves the seat ceiling ordering across all seeded roles and actions, and closes a validation gap: write `record_grant`s to users on a read seat now return 403 `seat_tier_insufficient`. Previously the grant could be created but was unusable; now we fail fast. Read grants remain allowed; `team` subjects are unchanged.

- Adds a seat_type × role × action integration matrix in `backend/internal/compose/integration/seatmatrix_integration_test.go`. The test derives seeded roles from the DB and reads RBAC expectations from each role’s permission document. It asserts: read seat → 403 `seat_tier_insufficient` for all actions; full seat → success when granted, `permission_denied` otherwise. Fixtures reset per cell; the export action uses an `owner_id` filter; share recipients are pinned to a full seat.
- Enforces the recipient’s seat in `CreateRecordGrant` via `refuseWriteGrantToReadSeat` in `backend/internal/modules/identity/grants.go`: refuse write grants to read-seat users; allow read; ignore `team` subjects.
- Documents the recipient seat rule in `backend/api/crm.yaml` and `frontend/src/api/schema.d.ts`.
- Callers that create write grants must handle 403 `seat_tier_insufficient` when the recipient holds a read seat. No schema or migration changes.

<sup>Written for commit 8b7c42d83e7eb4a05ec1dcc649e6071fffcd3b12. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1299?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/api/crm.yaml
+++ b/backend/api/crm.yaml
@@ -7955,6 +7955,12 @@ paths:
         re-asserting upgrades/downgrades `access` and resets `expires_at`. A grant can never exceed the
         granting principal's own access (scope-intersection). Audited (`action: record_share`). Bounded:
         flat explicit grants only — no sharing hierarchies, criteria-rules, or grant-of-grant delegation.
+
+        The seat ceiling binds the RECIPIENT as well (AAD-AC-4): a `write` grant to a user on a `read`
+        seat is refused `403 seat_tier_insufficient`, because it would widen that user's scope onto a
+        record every write to which the ceiling then refuses. A `read` grant to a read seat is the
+        authority their licence already carries and is accepted. A `team` subject is not a seat and is
+        not filtered — the read seats inside it are still refused every write at their own admission.
       parameters:
         - $ref: '#/components/parameters/IdempotencyKey'
       x-agent-access: human-only   # the grant verbs reject any non-human principal
--- a/backend/internal/compose/integration/seatmatrix_integration_test.go
+++ b/backend/internal/compose/integration/seatmatrix_integration_test.go
@@ -1,0 +1,431 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+//go:build integration
+
+package integration
+
+// AAD-AC-4 as a matrix rather than a sample: seat_type × role × action.
+//
+// The claim under test is an ORDERING one — the seat ceiling sits BELOW
+// role permissions, so a read seat is refused whatever its role grants,
+// and a full seat then falls through to RBAC. Two verbs on one role
+// cannot state that; only every seeded role against every action class
+// can, and only if the expectation for the full seat is read out of the
+// same grant document the server enforces rather than written down here
+// a second time.
+//
+// So the rows come from the DATABASE (`select key, permissions from
+// role`, the seeded system roles), and the full-seat verdict is derived
+// from that JSON. Re-granting a role, or seeding another, moves this
+// test with it; a hand-listed table would have gone on passing.
+
+import (
+	"bytes"
+	"encoding/json"
+	"net/http"
+	"testing"
+
+	"github.com/gradionhq/margince/backend/internal/compose/integration/apptest"
+	"github.com/gradionhq/margince/backend/internal/shared/kernel/ids"
+)
+
+// seatAction is one of AAD-AC-4's action classes, bound to the endpoint
+// that performs it and to the grant that decides it for a full seat.
+type seatAction struct {
+	class string
+	// object/verb name the RBAC grant the server checks, so the expected
+	// full-seat outcome can be read out of the role's own permissions.
+	object string
+	verb   string
+	// call issues the request and answers the status and the problem code.
+	call func(t *testing.T, e *apptest.AppEnv, f seatFixtures) (int, string)
+}
+
+// seatFixtures are the records the matrix acts on, all created once by
+// the bootstrap admin on a full seat before any role is swapped in.
+type seatFixtures struct {
+	personID   string
+	dealID     string
+	pipelineID string
+	birthStage string
+	nextStage  string
+	colleague  string
+}
+
+func seatActions() []seatAction {
+	return []seatAction{
+		{
+			class: "mutate (create)", object: "person", verb: "create",
+			call: func(t *testing.T, e *apptest.AppEnv, _ seatFixtures) (int, string) {
+				return callForCode(t, e, "POST", "/v1/people", apptest.AnyMap{
+					"full_name": "Matrix Probe", "source": "manual",
+				})
+			},
+		},
+		{
+			class: "mutate (update)", object: "person", verb: "update",
+			call: func(t *testing.T, e *apptest.AppEnv, f seatFixtures) (int, string) {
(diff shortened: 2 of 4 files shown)


tinnhlt-gra:
## UAT — one admin, one action, two seats

Live `DEV_SLUG=seatm` stack built from this branch (api `:18477`), signed in as `admin@demo.test` — **the admin role throughout**. Only the seat changes.

seat-ceiling-uat

1. **Read seat.** New contact → *"Grace Hopper"* → Create. The dialog stays open with **`seat tier insufficient`**, and the list still holds one contact. The role is `admin`; the ceiling does not care.
2. **Full seat, same admin, same click.** The contact is created and the 360 opens. That is the AC's second half — a full-seat user in the same role succeeds, subject to RBAC.

The same pair over the API, against the same stack:

```
# read seat
$ POST /v1/people        → 403  seat_tier_insufficient

# the refusal this PR adds — a write grant to a read-seat colleague
$ POST /v1/record-grants  {"access":"write", subject = a read-seat user}
  → 403  seat_tier_insufficient
     "a write grant needs a full seat; the subject holds \"read\""

# and the read grant to the same colleague, which their licence does carry
$ POST /v1/record-grants  {"access":"read",  same subject}   → 201
```

Before this branch the second call answered **201**: the colleague got a write grant that widened their scope onto a record every write to which the ceiling then refused.

PR #1319, opened 2026-08-15
fix(send): a scheduled message nothing will wake is found and re-armed

### LINKED ISSUE #1257: A scheduled send whose timer exhausts can be lost, or can hold a newer reschedule
Found by review during ADR-0104 (scheduled send).

## What is wrong

On the timer's last attempt, the fire transaction rolls back and releases its lock, then a **separate** transaction holds whichever `scheduled` row it finds. That handoff is neither durable nor bound to the version the failing attempt saw.

Two consequences:

- **A message can be lost.** If a database outage makes both the fire and the hold fail on the final attempt, River discards the job and the row stays `scheduled` with no live timer. Nothing will ever wake it, and nobody is told.
- **A newer reschedule can be held.** A rep can move the message after the fire rolled back but before the stale job's hold runs. The hold then lands on the rescheduled version, and its fresh timer later no-ops.

## Fix direction

Bind the exhaustion hold to the version and due instant the failing attempt observed, so a stale job cannot hold a newer intention. Add a periodic reconciliation for rows left `scheduled` with no live job, so a failed final hold is retried once infrastructure recovers.

## Why it was deferred

Both paths need infrastructure to fail at a specific moment to reach, and the recovery sweep is its own piece of work. The ordinary refusals — consent, seat, missed window — hold correctly today and are covered by tests.

LarsGradion:
Closes #1257. Two defects, both in what happens when a timer does not come back.

## A message could wait forever with nobody told

A scheduled row is armed by a River job and by **nothing else**. That is the right design — the row is the schedule, the job is a dumb alarm — but it has one failure the send path cannot see: a job discarded after exhausting its attempts, or lost to an outage spanning the whole ladder, leaves the row `scheduled` with no timer. Nothing wakes it, and the rep is told nothing, because *being told* is what the fire path does and the fire path never runs.

A pass now finds those every 15 minutes and **re-arms** them — not sends, not holds. It puts back the alarm that was lost, and the ordinary fire path makes the ordinary decision, including holding it as a missed window (the honest outcome for a message an outage carried past its time). A sweep that decided anything itself would be a second send path, and this feature's whole design is that there is one.

Each recovered message binds its **own** workspace, read from the row. The pass has no tenant — it sweeps the installation for rows nobody is watching — and one taken from its context could only ever recover a single workspace's messages.

## A stale attempt could stop a message the rep had just rescheduled

The exhaustion hold took whatever row it found. So: the fire rolls back, the rep moves the message, and the stale job holds their **fresh** intention under a verdict about the version before theirs.

The claim now reads the row version, the fire reports it back even when it fails (recorded outside the transaction, because a rollback would take the observation with it), and the hold declines when the row has moved. Zero means *"this claim is the observation"*, not *"skip the check"* — an unconditional arm would put the defect back, and the recovery pass cannot repair it because that pass only scans rows still `scheduled`; a wrongly-held row is invisible to it.

## Operational note

**This adds a new periodic job to every installation.** Cheap by design — one indexed read of overdue rows, and on a healthy system it finds nothing and logs nothing. It logs at WARN only when it actually finds a lost timer, which is something an operator wants to know even though the message is now recovered.

Its fault policy is declared in `jobs.yaml`: one unrecoverable message logs and the pass carries on, because the **cadence is the retry** — failing the whole job would re-run messages that already succeeded, and one permanently unreadable row would block every recovery behind it.

## Tests

Two integration cases over a real Postgres, both mutation-checked:

- a stale attempt declines rather than holding a rescheduled message;
- recovery re-arms a stranded message — asserted by counting the **River alarms** for it, because an earlier version of this test drove the fire path directly and would have passed whether or not anything was enqueued.

Full local gate, craft gate, RLS and jurisdiction greps green; pre-push hook run rather than bypassed.

---
## Summary by cubic
Recovers stranded scheduled sends and makes exhaustion holds version-safe so stale attempts cannot stop a rescheduled message. Before: exhausted or lost timers left rows scheduled with no alarm, and a stale attempt could hold a just‑rescheduled message. Now: a periodic recovery re‑arms missing alarms, and holds bind to the attempt’s observed row version and decline if the row moved or was never read.

- Adds periodic job `comms_scheduled_send_recovery` (15m; registers when `SendRegistry` and `SendDelivery` are present). It resolves the installation’s workspace, explicitly filters by `workspace_id`, scans for ≥30m‑overdue scheduled rows, re‑enqueues the alarm due now (no send/hold), logs WARN with found/rearmed/failed, and continues on per‑row errors.
- Fire/Hold contract: `FireScheduledSend` returns the observed row version even on failure. `HoldScheduledSend` requires an observation and no‑ops if the row moved or the attempt never read the row; `UnobservedVersion` (−1) means “use this hold’s claim,” while 0 means “attempt never claimed,” which declines. The worker passes Observed on timer exhaustion and `UnobservedVersion` for inactive‑sender holds.
- Security/fitness: recovery reads under the installation workspace; RBAC waivers mark `OverdueScheduledSends`/`RearmScheduledSend` as worker‑only. Adds a test proving a rep cannot score a lead they cannot see (returns NotFound).

<sup>Written for commit 35e77715e6fcf6ccc7ff33a5490a2e42e6b8ad2b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1319?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


LarsGradion:
@coderabbitai full review

Worth a close look at two things:

1. **A new periodic job on every installation** (`comms_scheduled_send_recovery`, 15-minute cadence). It reads overdue `scheduled_send` rows on the installation pool and re-arms them. I would like a second opinion on the cost of that read at scale and on the grace window (30 minutes) that keeps it off messages still being worked by the normal retry ladder.

2. **The fault policy** declared in `jobs.yaml`: one message that cannot be re-armed is logged and the pass carries on, on the reasoning that the cadence IS the retry. The alternative — failing the whole job — would re-run messages that already succeeded and let one permanently bad row block every recovery behind it. Reasonable people could disagree; please push back if you would.

Also: `OverdueScheduledSends` deliberately reads on the installation pool rather than a workspace-bound transaction, because the pass has no tenant of its own and binds each message's own workspace before touching it. That is the one place this diff steps outside the usual tenant-scoped read, and it is the part I would most want challenged.

LarsGradion:
Two additional fixes pushed, both for failures that were already red on `main` rather than introduced here:

**1. Frontend types carried a contract the backend had already moved past.** The `sent` status landed in `crm.yaml` with the released-to-sent work, but `make gen-types` was never run beside it, so `schema.d.ts` declared a status set the backend no longer serves. This is what `fe-quality` and `frontend` were failing on — the regenerated file is committed unedited.

**2. The two lead-score tables had no visibility decision.** `TestFK_rowScopedTargetsHaveVisibilityDecision` has been failing on main since the lead-score work merged, taking the whole integration lane down for every branch. `lead_manual_signal.lead_id` and `lead_score_history.lead_id` are FKs to a row-scoped record and the fitness gate requires each to be explicitly classified.

Both are safe, and the entries now say why:

- `lead_manual_signal.lead_id` **is** client-supplied and is gated — `SetLeadManualSignal` calls `auth.EnsureVisibleLive` in the same transaction as the insert.
- `lead_score_history.lead_id` carries no client-named lead; `recomputeLeadScoreTx` appends it from the lead its enclosing gated write is already working on.

`TestARepCannotScoreALeadTheyCannotSee` makes the first entry a proven claim rather than an asserted one, and it is mutation-checked: delete the `EnsureVisibleLive` call and it fails with "the visibility probe is not running".

Worth reviewing carefully: the first draft of that test **passed against no gate at all**. It seeded a lead with no owner, and `OwnerPredicate` reads `owner_id IS NULL OR = me` — an unowned record is visible to every scope tier by design, so the probe allowed it honestly. The committed test gives the lead a different owner.

LarsGradion:
@coderabbitai The index finding is correct and now fixed in core `0251`. I measured it rather than taking it on reasoning, and the numbers are worse than the review suggested.

On 200k scheduled rows with 5k genuinely overdue, the query as written:

```
Parallel Seq Scan on scheduled_send
  Rows Removed by Filter: 195000
  Buffers: shared hit=4556          Execution Time: 7.452 ms
```

Postgres does not fall back to a seq scan only at scale — with no overdue rows it *does* use `idx_scheduled_send_due`, but as a full index scan followed by a sort, because `workspace_id` leads and the sweep never constrains it. It can never walk that index in `scheduled_at` order and stop at the limit.

With the new index, same data and same query:

```
Index Only Scan using idx_scheduled_send_recovery_due
  Heap Fetches: 0
  Buffers: shared hit=1 read=3      Execution Time: 0.079 ms
```

I took your `INCLUDE (id, workspace_id)` suggestion as written. `Heap Fetches: 0` is what it buys: those are the only two columns the sweep selects, so carrying them in the leaf keeps the heap out of the plan entirely. Partial on `'scheduled'` for the same reason 0242's is, so the index stays proportional to the pending backlog rather than the send history. 0242's index is untouched — it remains right for every workspace-scoped read.

The measurement matters beyond this PR: this pass runs every 15 minutes on every installation and recovers nothing on almost every run, so the bad plan would have been paid forever and shown up as nothing more specific than background load.

Two other notes on this PR, both for defects that were already on `main` rather than introduced here: the regenerated frontend types (`fe-drift`), and the two lead-score tables' visibility classification with a mutation-checked test behind it. The migration-sequence half is #1321, not this PR.

LarsGradion:
End-of-work review round applied. The security pass found a real defect that changed the design, so this is worth reading before merging.

## The sweep could transmit mail for an archived organization

The recovery read was installation-wide — no workspace predicate, on the infra pool. Core 0217's tenant-scope sweep leaves an archived workspace's rows in place rather than deleting them, so the pass could find a scheduled message belonging to an organization somebody switched off, re-arm it, and send it. The fire path behind it does not catch this: it re-checks the sender's seat against `app_user` rows that are still active in the archived tenant. Before this PR those rows were inert because nothing enumerated them; the sweep is what made them reachable.

## The per-row workspace binding did not do what its comment claimed

`RearmScheduledSend` goes through `sendStore`, whose handle is `InstallationDB(pool)`, and `DB.Tx` resolves that handle's workspace via `identity.Service.InstallationWorkspace` — which **ignores the context**. So `principal.WithWorkspaceID(ctx, row.Workspace)` set the context while the transaction's `app.workspace_id` GUC stayed the installation's. They agreed by accident for live rows and diverged silently for archived ones. `storekit.LockWriteIdentity` keys on the GUC and `storekit.MustWorkspace` keys on ctx, so any writer later added inside that transaction would lock under one workspace and stamp under the other.

## One fix for both

The sweep now reads its own workspace like every other read in the module. ADR-0061 gives an installation exactly one live organization, so there is nothing a cross-tenant read should reach that a bound one does not. The worker resolves it once through `installationJobCtx` — the spelling every other tenant-less pass already uses — so ctx and GUC cannot drift.

## That retires core 0251

The index existed only because the unbounded query could not use `idx_scheduled_send_due` (which leads with `workspace_id`). Bound, the query is exactly what 0242's index was built for: index scan with both columns in the `Index Cond`, no sort, on the same 200k-row data.

Worth noting for anyone tempted to keep both: with 0251 also present the planner **preferred the worse plan** — scanning `scheduled_at` alone and filtering other workspaces out. It measured equal only because the test database has one workspace. So the index would have looked harmless in benchmarks and cost real work in production. @coderabbitai — your finding was correct about the query as it stood; fixing the tenant scope removed the need for the index rather than confirming it.

## Also applied

- The job now registers on `SendRegistry` **and** `SendDelivery`, matching the worker it enqueues into. Gating on one would file alarms nothing consumes, and `ScheduledSendArgs` has no `UniqueOpts`, so they would pile up in `river_job` every 15 minutes.
- The WARN line ran *before* the loop and reported rows read as rows re-armed — if every re-arm failed it claimed a recovery that never happened. It now runs after, counting `found`/`rearmed`/`failed`. The `failed` count is the only place a deterministically-unrecoverable row becomes visible; the cadence retries transient faults forever and cannot heal those.
- Comment cleanups: fix narration and `#1257a`/`#1257b` ticket labels removed, the triplicated "a row is armed by a River job" paragraph reduced to one copy, and the row-scope test's caller renamed so `base.owner` cannot be misread as the lead's owner.

Local: `make check` green apart from the two inherited `duplicate version 0248` failures, `make craft-static` clean, and the compose integration suite green. Mutation-checked: disabling the re-arm loop fails `TestARecoveryPassReArmsAMessageWhoseAlarmIsGone`.

Still blocked on #1321 landing first.

PR #1321, opened 2026-08-15
fix(migrations): main has two 0248s, and the frontend types drifted

tinnhlt-gra:
`main` is red on two counts, independently. Neither can be merged past, so every PR in the repo is blocked until this lands.

## Two migrations claim core `0248`

- #1301 → `0248_backfill_window_reach`
- #1309 → `0248_scheduled_send_references`

Both merged, so `pgmigrate` refuses to load the core namespace at all: every embedded-migration test fails and **no installation, fresh or existing, can migrate**.

The backfill window renumbers, for two reasons — it landed last, and it is main's own tip, so nothing has been built on the number yet. `0249` was already taken by the lead-score work, so it becomes **`0250`**. Nothing else in the tree names the version (the filename is the whole reference); the header comment moves with it so the file does not describe itself by a number it no longer has.

This is the sixth such collision in three weeks and the second this week. The number is picked when a branch starts and claimed when it merges, and nothing in between notices — worth a gate of its own, filed separately if it recurs.

## The frontend types drifted

`ScheduledSend.status` gained `sent` in `crm.yaml` and `frontend/src/api/schema.d.ts` never followed, so `fe-drift` fails on a clean checkout of main. Regenerated here — the file is generated, so the diff is the generator's.

## Verification

On a worktree branched from `origin/main`, `make check` fails on **both** counts before this change:

```
--- FAIL: TestEmbeddedMigrationNamespacesLoad/core
        pgmigrate: duplicate version 0248
make[2]: *** [fe-drift] Error 1
```

and after it: `make check` → **exit 0**, `make craft-static` → **exit 0**.

Found while merging `main` into #1296; that PR is blocked behind this.

---
## Summary by cubic
Unblocks main by fixing a duplicate core migration, regenerating scheduled-send types, and classifying two lead FKs for the schema fitness gate. Old behavior: `pgmigrate` refused to load due to two 0248s; `fe-drift` failed because `sent` was missing; the integration lane failed on unclassified `lead_manual_signal.lead_id` and `lead_score_history.lead_id`. New behavior: the backfill window migration is 0250; types include `sent`; FK gates are classified; checks pass.

- Core migrations: renamed `0248_backfill_window_reach` to `0250_backfill_window_reach` and updated the header. No runtime or data change. No manual action; run standard migrations.
- Frontend schema: regenerated `frontend/src/api/schema.d.ts` to add `sent` to `ScheduledSend.status` and the corresponding query filter.
- Schema fitness: added `rowScopedFKDecisions` entries — `lead_manual_signal.lead_id` is gated (checked via auth in the same transaction); `lead_score_history.lead_id` is a child row written only through the lead’s gated write paths.

<sup>Written for commit f998ef433dbc2f23bfefb61822887c596b508d13. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1321?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


LarsGradion:
I independently hit this and wrote the same fix before spotting yours — mine (#1322) is closed in favour of this one, since yours was opened first and the diffs are byte-identical. Two sessions landing two 0250s is how the earlier 0242 fix went wrong (#1284 duplicated #1275 and had to be reverted in #1285), so only this one should merge.

Two things I verified locally that may save you a round:

**Your shard 3 failure is not yours.** It is `TestNoDealLandsOnAStageBeingRemoved`:

```
stage_removal_race_integration_test.go:74: no stage was removed in any round — the race never ran
```

That is the load-sensitive concurrency test failing to schedule its race under a loaded runner, not a defect in a migration renumber. My #1322, with the identical diff, went green on the same checks — so a re-run should clear it.

**There is a second, separate defect still on main after this merges.** `TestFK_rowScopedTargetsHaveVisibilityDecision` fails because the two lead-score tables (`lead_manual_signal.lead_id`, `lead_score_history.lead_id`) carry no visibility classification. That is fixed in #1319, not here.

I checked the combination directly: your renumber plus #1319's classification plus #1319's new core 0251, all applied together — `make check` green, and the migrations integration lane green including the fitness test. The two PRs do not conflict, and 0251 is numbered to sit above your 0250.

LarsGradion:
Confirming why this PR cannot go green on its own, and what unblocks it.

Its remaining failures split in two:

- `integration shard (1/12)` and the `integration` job — `TestFK_rowScopedTargetsHaveVisibilityDecision`, the two lead-score tables with no visibility classification. Already on main, fixed only in #1319.
- `integration shard (3/12)` — `TestNoDealLandsOnAStageBeingRemoved` ("the race never ran"), the load-sensitive concurrency test. Not a real defect; the identical diff in my now-closed #1322 went green on the same check.

So the two open PRs each need the other's fix: this one needs #1319's classification to pass the fitness gate, and #1319 needs this renumber to load the migration sequence at all. Neither is green alone, and that is not going to resolve by re-running.

I verified the combination locally — this renumber, plus #1319's classification, plus #1319's new core 0251 index, all applied together:

- `make check` green
- `make test-it DIR=backend/migrations` green, including `TestFK_rowScopedTargetsHaveVisibilityDecision` and the migration sequence

The numbering does not conflict: this takes 0250, #1319's index takes 0251.

Someone with merge rights has to break the tie, and this PR is the one to land first — it is the smaller change, and it is what makes `make dev` work again for everyone. Its fitness failure is fixed by the PR immediately behind it. I am not merging over a red check myself.

PR #1322, opened 2026-08-15
fix(migrations): two PRs both claimed core 0248, so no installation can migrate

LarsGradion:
`main` cannot load its own core migration sequence:

```
migrate: pgmigrate: duplicate version 0248
```

#1309 added `0248_scheduled_send_references` and #1301 added `0248_backfill_window_reach`; each was numbered against a `main` that did not yet have the other, and both merged. The loader rejects the sequence **before applying anything**, so this is not a partially-migrated database — it is no migration at all. Every DB-backed lane fails, and so does `make dev`.

Reproduced locally on a clean checkout of `origin/main`:

```
make migrate
migrate: pgmigrate: duplicate version 0248
```

## The fix

#1309 merged first and keeps 0248; the backfill-window half becomes **0250** (0249 is the lead-score explainability migration). Its header comment moves with it. Nothing outside the two files referenced the number — checked across Go, YAML, SQL and docs.

Re-running is safe for any database that somehow recorded the old version: the migration drops and re-adds one CHECK constraint, so applying it twice leaves the same constraint. In practice none can have, because the duplicate version stopped the whole sequence from loading.

## Also in this PR

The regenerated frontend types. `make gen-types` was never run after the `sent` scheduled-send status landed in `crm.yaml`, so `schema.d.ts` declared a status set the backend no longer serves, and `fe-drift` was red on main for that reason. The file is generator output, committed unedited.

## Verified

- `make migrate` reaches head after the renumber
- `make check` green (was red on `fe-drift` before the type commit)

One caveat for the reviewer: the migrations integration lane still fails on this branch, at `TestFK_rowScopedTargetsHaveVisibilityDecision`. That is a **separate** defect also already on main (the lead-score tables carry no visibility classification), fixed in #1319 rather than here to keep this PR to the thing blocking everyone.

This is the second 0240-shaped collision in as many days (#1275 was the first). The check that catches it is cheap and has to happen twice: read `origin/main`'s highest core migration number immediately before merging, not only when the branch is created.

---
## Summary by cubic
Fixes a duplicate core migration version that blocked startup. Main previously failed to load due to two 0248s; it now loads by keeping `0248_scheduled_send_references` and renumbering `backfill_window_reach` to 0250. Also updates FE types to include the `sent` scheduled-send status.

**Review and rollout**
- Renames to `backend/migrations/core/0250_backfill_window_reach.{up,down}.sql` (SQL unchanged; header comment updated).
- Safe to apply even if an environment recorded the old step; the CHECK constraint ends identical.
- No code references the numeric version; only filenames/comments changed.
- Regenerates `frontend/src/api/schema.d.ts` to add `sent`; removes FE schema drift.
- Action: run database migrations normally; no manual steps.

<sup>Written for commit dff43481df47acba150ccc3a0fcf364c209dd6ee. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1322?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


LarsGradion:
Closing in favour of #1321, which is the same fix opened two minutes earlier. The two diffs are byte-identical: `0248_backfill_window_reach` → 0250, the header comment moved with it, and the regenerated `schema.d.ts`.

Two sessions independently diagnosed `pgmigrate: duplicate version 0248` and wrote the same patch. Landing both would leave main with two 0250s and recreate the exact failure — that is how the earlier 0242 fix went wrong (#1284 duplicated #1275 and had to be reverted in #1285).

#1321 has it. One note for whoever merges it: the migrations integration lane still fails afterwards at `TestFK_rowScopedTargetsHaveVisibilityDecision`, which is a **separate** defect also already on main — the two lead-score tables carry no visibility classification. That half is fixed in #1319.

Issue #1324, opened 2026-08-15
An openai_compatible binding cannot carry a document, so BYOK operators lose every document feature


tinnhlt-gra:
## What is missing

`model.Request.Attachments` is the cross-provider input part (image / PDF / audio), and each adapter is required to map it or refuse it — never drop it silently. Two adapters map it today:

- `gemini.go:244` — carries `image/*` and `application/pdf`
- `openai.go:197` — carries `image/*` and `application/pdf`

`openaicompat.go:159` refuses every attachment, and says why in its own comment:

> This text-only chat wire carries no attachment parts, so reject every attachment rather than accept-then-drop it (spec §3.8 map-or-reject). The generic endpoint's multimodal support is model-dependent; mapping images to image_url content parts on this shared wire is a follow-up, and until it lands an honest rejection beats a silent drop.

This is that follow-up. The consequence is concrete: an operator whose routing binds `openai_compatible` — which is how the repo's own OpenRouter example config binds **every** tier — cannot use any feature that sends a document to a model, however capable the model they bound actually is.

## Why it is not a one-line adapter change

Multimodal support on the OpenAI-compatible wire is a property of the **binding**, not of the adapter. OpenRouter carries images for vision-capable models and not for text-only ones; a self-hosted vLLM carries them only when the served model is a vision model. The adapter cannot answer "can I carry an image?" from its own code, because the answer depends on which model the operator bound to that tier.

So it needs a declaration in the routing config, and `config/ai-routing.schema.json` is generated with `additionalProperties: false` — which makes this three coordinated changes, not one:

1. a per-binding capability field in `config/ai-routing.yaml`
2. the matching change in `tools/gen-aitasks`, which generates the schema
3. `image_url` content parts on the `openai_compatible` chat wire, gated on that declaration

## How other open BYOK systems solve it

Surveyed while designing #701, because the same question decides what a document extractor can be fed:

| system | wire normalization | capability detection | provider-independent lane |
|---|---|---|---|
| **LiteLLM** | one shape (`{type: file, file: {file_data: "data:…;base64,…"}}`) mapped per provider | a **model capability registry** — `supports_pdf_input()`, `supports_vision()` — checked by the caller; no automatic fallback | a separate OCR lane (`/ocr`, Mistral OCR, Azure Document Intelligence) as explicit pre-processing |
| **Vercel AI SDK** | `FilePart{mediaType, data}` | none — "developer responsibility"; errors surface from the provider | none, beyond an `experimental_download` hook |
[...]

PR #1331, opened 2026-08-15
feat(documents): a real extractor reads an attached document, and the accept writes the reading a human was shown

### LINKED ISSUE #701: Build a real document and image extractor behind the extraction seam
## What is missing

The interface for "read a document and give me structured fields with evidence" exists and is well designed. The only implementation in production is a **no-op** that returns nothing. We cannot read a scanned form, a PDF, or a photographed business card.

## Why it matters

A customer engagement needs completed paper forms turned into per-question values, with the original image kept as proof. Another needs offer documents made searchable. Both are blocked on the same missing piece.

- `backend/internal/shared/ports/extraction/extraction.go` — the seam. An extractor takes an attachment id and returns fields, each with the value, the **quote it came from**, the page or section, a confidence, and an explicit "omitted, and why" option. That evidence-or-omit shape is exactly our house rule; keep it.
- `NoOpExtractor` is the production default today; `FixtureExtractor` serves tests.
- How AI work is declared here: tasks are declared contract-first in `backend/api/ai-tasks.yaml`, generated into the AI module, wired in composition, and each task must ship a certification corpus. The walkthrough is `docs/how-to/add-an-ai-task.md`. Follow it — do not hand-roll a model call.
- Attachment ingestion is a hard prerequisite (see the sibling issue); without it there is nothing to extract from.

## What to build

1. A new AI task declaration for document extraction, including a model binding capable of reading images.
2. A real `Extractor` implementation behind the existing seam.
3. A certification corpus of real sample documents with expected values — this is how we hold quality and how we set the confidence floor.
4. A staging path: low-confidence values are proposed for a human to confirm, never written silently.

## Done when

- [ ] A scanned form attached to a mail produces structured values, each carrying the quote or region it came from.
- [ ] Values below the confidence floor are staged, not written.
- [ ] The certification corpus runs in the normal lane and fails the build if quality regresses.

**Size:** M.
**Spec:** `documents-and-files.md`, the extraction seam, and ADR-0074 for AI task certification.

tinnhlt-gra:
Closes #701.

## What was actually missing

`ports/extraction.NoOpExtractor` has been *"the production default"* since RD-T10 shipped the seam, answering honestly empty. The roadmap's read of this row was "only the provider is missing", and that is true of the Go seam and false of the feature — building the provider found the seam was the wrong shape twice over, and both are fixed here.

**A model call cannot happen inside the request that asks for it.** `GetAttachmentExtraction` called `Extract` inline: free under a no-op, multi-second under a vision model, and able to fail. Migration `0245_transcript_read` states the rule in its own words, and the deep read is the same shape for the same reason. So a reading becomes a durable record behind a 202, polled until terminal — new table `attachment_extraction` (`0251`), owned by `activities`, with that file's lease, in-flight uniqueness and abandoned-reading recovery mirrored deliberately rather than reinvented.

**The accept re-ran the extractor to validate.** Sound while two runs agree by construction; with a real model the human accepts the amount they were shown and a different one lands on the deal, carrying an audit note that quotes evidence they never saw — and it bills a second call to do it.

## The invariant this PR is really about

The reading stores what it grounded, and the accept resolves it **by id**.

Storing the fields was not enough on its own. Resolving *the newest stored reading* has the same failure with a quieter cause: a colleague presses the button, or a re-upload triggers a re-read, while the panel is open — both readings are real, neither looks wrong, and the number written is not the one anyone agreed to. So `AcceptExtractionRequest` gains a required `extraction_id`, and the panel sends the reading it rendered.

`TestAcceptAttachmentExtractionWritesTheReadingItWasShownNotTheNewest` fails against **both** shapes. Mutation-verified in both directions: with the accept switched to `LatestExtractionRead` it fails naming the wrong amount, and passes on restore.

## What the model is fed, and how that was decided

`model.Capabilities` gains `AttachmentMIMEs`, reported by each adapter from **the same list its wire enforces** — so a binding cannot advertise a media type it then refuses. The reading picks its lane before spending a call:

1. a text-native type goes as **text**, where the source quote is checked verbatim against the document
2. a type the binding declares goes as **bytes**
3. neither finishes the reading with a stated reason and **no model call at all**

Lane 2 carries a real weakness and the code says so rather than papering over it: an image has no text to search, so its quote cannot be substring-checked, and grounding there rests on the prompt and the corpus rubric.

I surveyed how other open BYOK systems solve this, because it decides what a document extractor can be fed:

| system | wire normalization | capability detection | provider-independent lane |
|---|---|---|---|
| **LiteLLM** | one shape mapped per provider | a **model capability registry** (`supports_pdf_input`, `supports_vision`), checked by the caller | a separate OCR lane (`/ocr`, Mistral OCR, Azure Document Intelligence) |
| **Vercel AI SDK** | `FilePart{mediaType, data}` | none — "developer responsibility" | none |
| **LibreChat** | MIME sniff → per-kind route | none — the **human picks** | OCR → speech-to-text → text parse |
| **Open WebUI** | — | none | a pluggable content-extraction engine (Tika, Docling, …) |

Nobody makes every provider carry bytes, and capability is declared per binding rather than learned by being refused. Margince already had the better half in `ErrAttachmentUnsupported`; what was missing was the declaration.

**Not in this PR:** the `openai_compatible` `image_url` mapping, so an OpenRouter-bound operator can read documents too. It is per-*binding* rather than per-adapter, so it needs a routing-config field and a `gen-aitasks` change — filed as #1324 with the survey attached, rather than buried in a feature PR.

## Also here

- The `Extractor` interface is **gone**. With a durable reading there is no call left to abstract, and a no-op beside a real engine would leave two answers to "what does this document say". The package keeps `ExtractedField`, which is the one thing two modules must agree on.
- `FakeClient` now obeys map-or-reject like every real adapter, and its recorded payload carries attachments. It was the one client that silently dropped an input — exactly where a test would fail to notice.
- The scan gate **moved to where bytes reach a model**: the start of a reading, not the poll. Asserting it on an operation that touches no bytes would go on passing after it had been removed from the path that matters.
- `values.MinorUnits`, the inverse of the `MajorUnits` already there, so the amount is scaled in Go from the figure the document prints. A model asked to multiply by a hundred can be wrong by a hundred, and that is the error nobody catches on a scan.
- A second omission reason, `not_confidently_stated`. With only `not_stated_in_file` every under-confident reading presented as a document that says nothing — a claim about the *document* made on the strength of a fact about the *reading*.

## The panel (AC-attachments-4/-5)

Specified since the corpus was written; nothing in `frontend/` had ever called either endpoint. It offers each grounded field with the text it was read from, renders what the document does **not** say as the answer it is, and keeps the three states apart in the words as well as the data — not answered yet, read it and it states none of them, could not read it at all.

## Certification

Four scenarios: a complete order form, one that states three fields and deliberately not the fourth (a validity window is not a close date), one that states none at all, and a PDF read on the bytes lane. The paid gemini run and its record follow in a second commit — the corpus is finalized first, since a later corpus fix invalidates every record already paid for.

## Gates

`make check-backend`, `make check-fe` and `make test-integration` all green over the merged tree, re-run after resolving the `origin/main` merge. The one breaking contract change is ratified as an exact blob pair in `scripts/contract-breaking-allowlist.txt` rather than run under `CONTRACT_STABILITY=pre-live`, so the exception cannot mask the next contract edit.

Spec landed first: gradionhq/margince-foundation#1307 (RD-DDL-4, RD-PARAM-N-2..N-6, RD-WIRE-N-1..N-3, RD-AC-N-2..N-9).

---
## Summary by cubic
Queues and stores a durable reading of one attached document and requires accept by reading id. Old: GET ran a no-op extractor inline and accept re-ran it. New: a background reading is persisted and polled; GET returns 404 when never read; accept resolves by that reading id to prevent mismatched writes and long request latency.

- Persists readings in `attachment_extraction` (migration `0252`) and exposes: POST `/v1/attachments/{id}/extraction` → 202 with reading id; GET `/v1/attachments/{id}/extraction` → `queued`/`running`/`done`/`failed` plus grounded fields and honest omissions; unread attachments return 404.
- Adds job kind `document_extract` on the `transcript_read` queue; wires `DocumentExtractBrain`; removes the `Extractor` interface; readings store grounded fields and accept writes exactly what was shown.
- Validates replies on the shipped path: uses the structured lane when available and always re-validates; quotes must support their values; filename and media type are fenced in the prompt.
- Hardens lifecycle: finish is claim-scoped; retries hand the claim back; job de-dup covers active states only; unreadable/scan‑blocked files close as `failed` (the panel offers retry).
- Updates accept: `AcceptExtractionRequest` requires `extraction_id`; amount renders/edits in major units and converts on write; amount must include currency; rejects JSON numbers too large to be safe. Implements #701’s invariant.
- Declares and enforces `model.Capabilities.AttachmentMIMEs`; picks a lane per file (text vs bytes); unsupported types complete with a stated reason and no model call; adapters advertise carriage; the fake client can simulate text‑only.
- Adds `values.MinorUnits` for safe parsing; frontend panel shows the three states, displays grounded quotes, renders money correctly, and sends the `extraction_id` it rendered.
- Cert corpus and tests cover full, partial, empty, and PDF cases; fixes include capability‑set intersection and stale‑edit handling.

**Rollout and migration**
- Run DB migration `0252_attachment_extraction`.
- Update clients to include `extraction_id` in `acceptAttachmentExtraction` requests.
- Ensure the worker config binds a provider for `document_extract` and a blobstore (`SendBlob`) so it can read bytes; to use the bytes lane, bind a model that declares image/PDF support, otherwise readings will omit or fail by design.
- No payloads are captured for this task; task is premium‑only.

<sup>Written for commit 9c3d263561a8f4a47e36b59fb1cfaef9cb936130. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1331?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tinnhlt-gra:
## UAT — live stack, real model

Recorded against a live `DEV_SLUG=docx` stack (api :18574, fe :8574) with a real gemini binding, driving the whole flow through the UI.

701-document-extraction-uat

**What the recording shows, in order.** The account's Documents tab carries two copies of the same order form. The first has never been read and offers *"Nobody has read this file for deal fields yet."* — pressing **Read this file** answers 202 and the panel says it is reading. When the worker finishes, the panel becomes *"AI read this file — 4 fields it can ground, staged for your record (accept to persist)"*, each field carrying the section it was read from. Clicking the amount opens the evidence mark: provenance `agent:document-extractor`, confidence `high`, and the verbatim quote `148,500.00`. **Accept 4 fields** flips the panel to *"4 fields accepted to the deal — original snippets retained"* with the controls gone.

**The amount is the thing to watch.** The document prints `EUR 148,500.00`; the deal stores `14850000` minor units; the panel shows **€148,500.00**. Those are three correct renderings of one number, and an earlier revision of this branch showed the middle one on screen — under a label reading "Amount", beside a snippet quoting the first, with an edit box under it.

### The API half, for the record

```
POST /v1/attachments/{id}/extraction      → 202 {"id":"…","joined":false,"status":"queued"}
GET  /v1/attachments/{id}/extraction      → done, 4 grounded fields, 0 omitted
```

```json
{"field":"amount_minor","value":"14850000","source_quote":"148,500.00",
 "page_or_section":"2. COMMERCIAL TERMS","confidence":"high"}
{"field":"expected_close_date","value":"2027-01-31","source_quote":"31 January 2027",
 "page_or_section":"3. ACCEPTANCE","confidence":"high"}
```

The close date is the second thing worth noting: the form carries an issue date (14 October 2026) and an acceptance deadline (31 January 2027), and the reading took the deadline. The accept then wrote all four onto the deal, and the record's header reads €148,500.00 / 31/01/2027.

### Certification

```
document_extract: certified (reliability=1.00 score_p50=100)
```

gemini-3.5-flash, four scenarios: a complete order form, one that states three fields and deliberately not the fourth, one that states none at all, and a PDF on the bytes lane. Record committed.

**Three paid runs, and the first two each found a real defect** rather than a corpus problem:

1. The scenario graded the free-text deal **name** by exact string, which grades phrasing rather than reading — an order form headed *"Order Form — Pallet Handling Programme, Graz site"* whose scope paragraph says *"the pooled pallet programme"* can honestly be named from either.
2. **The field's own name was telling the model to convert the amount.** Two runs of the identical prompt split: one returned `148500.00`, the figure the form prints, the other returned `14850000` — already in minor units, which this system would have scaled *again* into a deal worth €14.85 million. The value-supported-by-its-quote check is what caught it; nothing else in the stack could tell that from a large but honest amount. The model is now asked for `amount`, and the column keeps its own name on this side of the boundary.

### And one defect only the live stack could find

The first UAT attempt failed with *"this installation stores no document bytes, so there is nothing to read"* — a true sentence about the store the worker was built with, and a false one about the installation, where the bytes were sitting the whole time. The worker role configures a blobstore; this lane's store was constructed without it. Invisible to every unit test, because the store's honest refusal is indistinguishable from a role that genuinely has none.

Issue #1305, opened 2026-08-14
Erasure and subject-access do not reach approval payloads or workflow runs, which hold personal data


tinnhlt-gra:
## What is wrong

`approval.proposed_change` is persistent JSON with no foreign key to the subject it describes, and `workflow_run.planned` / `workflow_run.applied` are the same. Neither is reached by Art. 17 erasure or by subject-access assembly: there is no approval or workflow-run handling in `backend/internal/modules/privacy/`.

A staged approval routinely holds a person's name in `summary` and their data in the payload. A held draft (#707) holds their **email address and the full body of a message written to them**.

## Why it matters now

This is pre-existing — agent stagings have always carried payloads — but #707 (#1296) makes such rows routine rather than occasional: every automation-composed reply now stages one, carrying an addressee and prose about a named person, and it sits there for the life of the approval.

The precedent is already in the tree and points the other way. `privacy/erasure.go` explicitly scrubs **scheduled sends** for exactly this reason — a `scheduled_send` row holds an unsent body and addresses before any activity exists, so ADR-0104's own cost section says *"a scheduled send is not a place personal data may hide."* A staged held draft is the same object one step earlier in its life, and nothing reaches it.

## What to build

1. Approval rows and workflow runs join the erasure cascade, the way `scheduled_send` does.
2. Subject-access export includes them (`privacy/sar.go` reaches transmitted comms today; a message staged and never released is invisible).
3. A pending approval naming an erased subject must be invalidated **before** redaction, not left decidable against a scrubbed record.

Worth checking while in there: whether raw bodies belong in approval events and audit summaries at all, or only their identifiers.

## Found by

Both adversarial reviews of #1296, independently.

PR #1332, opened 2026-08-15
refactor(activities): the kind says what happened, the provider says what carried it

tinnhlt-gra:
Carries **ADR-0107/A158** into the build. Upstream landed first (contract-first, P3): the ADR was ratified in margince-foundation#1301, `specs/contract/crm.yaml` took the narrowing in #1305, and #1306 reconciled the chapter prose.

## The problem

`activity.kind` answered two unrelated questions in one column — *what sort of interaction happened*, and, for two of its values, *which messaging transport carried it*. The send path read the second back out of the first (`provider := string(anchor.Kind)`), which holds only while every transport is also spelled as a kind. A transport naming no interaction kind was unreachable, so **an extension unit could not file a single message without a spec PR first**.

## What changes

- `telegram` and `whatsapp` leave the kind vocabulary; **`message`** enters it. The transport is `activity.channel_provider`, a `ProviderRef` — a pattern-constrained string, not an enum, because which providers exist is a *deployment* fact.
- **Core `0251`** backfills both kinds, registers `whatsapp` as a transport (so hand-logged rows keep their meaning — it is a transport with no sender yet, which is honest while A103's connector is still coming), and adds the both-directions CHECK tying `kind='message'` to a non-null provider.
- The **down migration refuses rather than corrupts**: a message on a provider registered *after* the narrowing has no pre-narrowing spelling, so it names the offending transports and stops.
- ~15 readers moved, the frontend routes on the right axis, and the `?channel_provider=` filter is the only way left to ask what `kind=telegram` used to answer.

## Two production bugs the integration lane caught

Unit tests were blind to both: the **Telegram normalizer** and the **outbound reply writer** were still filing under the retired kind. The narrowed schema rejects that outright — every capture and every rep reply would have failed in production.

## What the review rounds found

**Codex found an exploitable hole.** A unit has *two* write doors, and the "a unit may not file a message" refusal sat on only one — while this very diff gave the second door the `channel_provider` field. A composed unit could have written `kind=message, channel_provider=telegram` and minted a valid **send anchor**: a rep or approved agent replying on it would transmit a real message from the workspace bot to whoever the unit linked. It would also have inherited the GoBD six-year floor. Now one refusal, called by both doors, with a fitness test that fails if either stops calling it.

**Fable found a cross-tenant hole** in the reply-match: no `workspace_id` predicate, so a forged mail `References` root could match an outbound leg in another workspace. Pre-existing on `main` rather than new — but this diff rewrites that statement, RLS was retired at `0217` so nothing else supplied the scope, and the new index leads with `workspace_id`. Fixed.

Both reviewers independently confirmed the reply-match predicate itself is equivalent-or-stronger than before.

Also fixed: the flip importer bypassed the new invariant; the contract promised `field_not_valid_for_kind` that no Go code emitted; the agent was told `channel_provider?: value` because `gen-recordfields` could not unwrap a one-element `allOf`; and the axis fitness test matched only flow-style enums, silently skipping four block-style ones.

## Gates

| Gate | Result |
|---|---|
| `make check` | exit 0 |
| `make test-integration` | **OK, 0 skips, 40 packages** |
| `make craft-static` | 0 blocker / 0 major / 0 minor |
| frontend vitest | 2649 passed |
| **new-code coverage** | **94.9%** (149/157 changed statement lines, measured against the merge-base) |
| `agent_loop` aicert | re-certified → `current` |

Both new guards are **mutation-verified**: removing either fails its own test with its own diagnostic.

**aicert honesty:** `agent_loop` was already stale on all three bindings on the baseline — verified, not assumed. I re-certified the gemini binding (the one this installation routes to); deepseek and mistral remain stale exactly as they were on `main`.

**Contract resync:** recorded as a ratified blob pair in `scripts/contract-breaking-allowlist.txt` rather than leaving the gate relaxed — any later contract edit restores the strict check automatically.

## UAT

Run against a live isolated stack (`DEV_SLUG=sp5axis`), **all checks passed** — script, output and recording attached in the following comment.

---
## Summary by cubic
Narrows activity taxonomy so kind states the interaction and channel_provider states the transport. Before: `kind=telegram|whatsapp` named the carrier. Now: all channel messages use `kind=message` and set `channel_provider`; non-messages must not set it.

- API/schema: `crm.yaml` drops `telegram`/`whatsapp` and adds `message`; ListActivities accepts `channel_provider`; public `activity.captured` events include `channel_provider` when `kind=message`.
- Database: core migration `0251` backfills provider-named kinds to `message`, registers `whatsapp` as a transport, and adds a CHECK enforcing: message ↔ provider present; non-message ↔ provider absent. Adds a workspace-leading index for reply-match; the down migration refuses if any message uses a provider created after this change.
- Send/capture and guards: send path reads transport from `channel_provider`; Telegram normalizer files `kind=message` + provider; a shared refusal blocks extensions from logging `kind=message` on a core-owned transport; reply-match now scopes by workspace and medium.
- UI/search: timelines and filters treat channel conversations as `message`; transport-specific queries must use `channel_provider`; titles fold the provider in to disambiguate.
- Migration actions:
  - Run core migration `0251` in order; do not attempt a down migration if any post-narrowing providers have messages.
  - Register every deployed transport so inserts with `channel_provider` satisfy the FK.
  - Update clients and automations: stop sending `kind=telegram|whatsapp`; send `kind=message` and set `channel_provider`; use the `channel_provider` filter instead of `kind` to ask “which transport?”; event consumers must read `channel_provider`.
  - Delivery semantics unchanged; retention now protects all channel messages via `kind != note|task` instead of transport name lists.

<sup>Written for commit 35f7bef25c71af41dc5607bad92887c064eff693. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1332?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tinnhlt-gra:
## UAT — live stack, all checks passed

Run against an isolated dev stack (`DEV_SLUG=sp5axis`, api `:18948`, db `margince_dev_sp5axis`) — a real boot with the migration applied, not a test template. Script: `.tmp/ext-core-write-port/uat/uat-sp5-1b.sh`.

What it proves, and why each check is there rather than being a restatement of a unit test:

1. **The narrowing is live** — `activity_kind` holds six interaction kinds and no transport; both transports are registered.
2. **The two axes cannot disagree** — both directions of the CHECK exercised against the live database: a message with no transport is refused, *and* an email carrying one is refused. The reverse direction is the one a suite forgets.
3. **A hand-logged message stores the pair** and `GET /v1/activities/{id}` serves `channel_provider` back — a field written but not readable is half a change.
4. **A message with no transport is a 422 naming the field** (`field_not_valid_for_kind` on `channel_provider`), never a 500 from a constraint the caller cannot see.
5. **The retired kinds are gone from the write surface** — `kind=telegram` and `kind=whatsapp` both 422. A 5xx here would mean the value reached the database.
6. **The send path still reads the column**, not the kind.
7. **The provider filter narrows by transport** — telegram returns rows, whatsapp returns none. A filter that answered everything would also pass a bare count check.

```
$(cat /tmp/uat-clean.txt)
```

PR #1344, opened 2026-08-15
fix(flakes): the sweep is scheduled by one clock, and a tree scan gets a tree-sized budget

### LINKED ISSUE #1329: flake: TestSweepBackoffGatesDueOverlayConnections races the real clock under parallel load
## What happens

`TestSweepBackoffGatesDueOverlayConnections`
(`backend/internal/modules/overlay/syncbackoff_integration_test.go`) fails
intermittently in the full `make test-integration` lane:

```
syncbackoff_integration_test.go:76: after a successful sweep the workspace must be due again
--- FAIL: TestSweepBackoffGatesDueOverlayConnections (0.06s)
```

It passes reliably in isolation (`-count=3` green), and the same lane passed on
a re-run with no code change, so it is load-dependent rather than a real
regression.

## Why

The test drives the backoff with the wall clock:

```go
store.RecordSweepFailure(ctx, apperrors.ErrIncumbentBudgetExhausted, time.Now())
// ...
store.RecordSweepSuccess(ctx, time.Now())
if !isDue(ctx, t, pool, ws) { ... }   // <- fails here
```

`isDue` compares `next_sweep_at` against the database's `now()`, while the
success is recorded against the *test process's* `time.Now()`. Under parallel
load the gap between those two readings widens, and the "due again" window the
assertion depends on can close before the read happens.

This is the `time.Sleep` / real-clock flakiness T11 names: the clock is a
boundary and wants injecting, not sampling.

## Suggested fix

Inject a `Clock` into the mirror store the way other suites do, and drive
`RecordSweepFailure` / `RecordSweepSuccess` and the due-gate from the same
source, so the assertion tests the backoff arithmetic rather than the skew
between two clocks. Backdating `next_sweep_at` through the owner connection —
the shape `approvalexpiry_integration_test.go` uses — is the cheaper variant if
a full clock injection is more than this deserves.

## Provenance

Found while running the lane for #1328, which does not touch `internal/modules/overlay`.
Not fixed there: unrelated subsystem, and the fix is a design choice about how
that store takes its clock.

### LINKED ISSUE #1340: flake: design-system conformance emoji scan times out at 5s under CI load
## What happens

`src/design-system/conformance.test.ts:138` — *"uses no emoji glyphs in source
strings — Lucide only (§2b)"* — intermittently fails the `fe-unit` job:

```
##[error]Error: Test timed out in 5000ms.
 ❯ src/design-system/conformance.test.ts:138:3
```

It passes locally (2669/2669) and passed on a plain re-run of the same commit
with no code change, so it is load-dependent rather than a real regression.

## Why

The assertion walks **every** `.ts`/`.tsx` file in the tree and regex-scans its
contents, against vitest's default 5s per-test ceiling. That budget is a
function of runner load and of how much source exists — both of which only ever
grow — so the margin shrinks with every file added, and any PR can be the one
that tips it.

The failure is also maximally confusing: the job is named
`fe-unit (vitest + coverage)`, so it reads as a coverage-threshold failure until
you open the log.

## Suggested fix

Give the whole-tree scans in this file an explicit timeout that reflects what
they do (`it(..., { timeout: 30_000 })`), rather than inheriting a per-unit-test
default they were never sized for. The scan is I/O over the repo, not a unit
test, and sizing it honestly is cheaper than re-running the lane.

Worth checking the sibling scans in the same file for the same exposure while
in there — they have the same shape.

## Provenance

Blocked #1328 once and cleared on re-run. That PR's entire frontend diff is two
generated API files, so it did not cause this.

tinnhlt-gra:
Closes #1329. Closes #1340.

Neither flake was a regression. Both were a measurement taken against something
that was never going to hold still.

## #1329 — the overlay sweep was scheduled by two clocks

The due-scan gates on `COALESCE(s.next_sweep_at, now()) <= now()`, so
`overlay_sync_state.next_sweep_at` is only ever compared against the clock
*inside* Postgres. `RequestSweep` and the flip seal already wrote it with that
clock; `RecordSweepSuccess` and `RecordSweepFailure` took a `time.Time` from the
caller and wrote the app process's clock instead.

The failure path has minutes of margin and never noticed. The success path means
*due immediately* and has none: whenever the app's clock ran ahead of the
database's, the reset landed in the future — a healed connection stayed backed
off for as long as the skew lasted, and nothing anywhere said so. In the
integration lane the same skew read as
`TestSweepBackoffGatesDueOverlayConnections` failing under load.

Both functions now take no clock at all, so the column is written by the one
clock that reads it, and no future caller can reintroduce a second one. Fixing
it at the API shape rather than in the test is deliberate: a test-side backdate
would have made the lane green and left the production skew in place.

**Proof of the mechanism**, from a throwaway integration probe (not committed):

```
zzprobe_integration_test.go:35: database-clock reset -> due (the fixed shape)
zzprobe_integration_test.go:49: app-clock reset under +2s skew -> NOT due (the flake)
--- PASS: TestProbeAppClockResetIsNotDue (0.09s)
```

A `next_sweep_at` stamped 2s ahead of the database's clock is not due; the
database-clock reset is. That is #1329's failure mode, reproduced on demand.

## #1340 — a repo sweep sized like a unit test

The design-system conformance gates read, and mostly TypeScript-parse, the whole
source tree against vitest's 5s per-test default. Measured per leg, warm, under
the coverage instrumentation `fe-unit` actually runs with:

| gate | cost |
|---|---|
| emoji glyphs (§2b) | 1058 ms |
| literal colours | 922 ms |
| hard-coded copy | 766 ms |
| type families | 487 ms |

That is already a fifth of the ceiling on a quiet laptop, and it had blown 5s on
a loaded CI runner — reported against a job named `fe-unit (vitest + coverage)`,
so it read as a coverage-threshold failure until you opened the log.

The budget is declared once on the suite now, an order of magnitude above the
worst observed cost. These scans are synchronous file I/O: there is no hang for
a tight ceiling to catch, so a generous one costs nothing, and every gate the
file grows inherits it.

## Two stale comments went with them

- The sweep test justified using the real clock because a reset "is always
  now-or-past" — that was the assumption that broke.
- `isDue` justified its membership filter with intra-package parallelism the
  lane does not have (`test-integration-parallel.sh`: packages run in parallel,
  tests inside one still run `-p 1`). The filter is right; the reason was not —
  several tests here seed more than one workspace, and the fleet-wide scan sees
  them all.

## Verification

- `make check` — EXIT=0.
- `make test-it` green on all three affected integration packages:
  `internal/modules/overlay`, `internal/compose/integration/overlay`,
  `internal/compose`.
- `pnpm vitest run` — 2669 passed, 0 failed.
- Pre-push gate: `craft static` 0 blocker / 0 major / 0 minor; rls-store-path
  and no-jurisdiction both OK.

---
## Summary by cubic
Eliminates two CI flakes by scheduling overlay sweeps with the database clock and giving the design‑system repo scan a suite‑level timeout. Old: app clock wrote `next_sweep_at` and Vitest’s 5s default timed out; new: all sweep scheduling uses Postgres `now()` and the suite declares a 60s budget. Side effect: healed connections become due immediately under skew; backoff floors still apply. No DB migration.

- Backend: `RecordSweepSuccess` and `RecordSweepFailure` drop the caller clock; all writes to `overlay_sync_state.next_sweep_at` use `now()` and `now() + make_interval(secs => ...)`. Adds a package gate (`syncclock_test.go`) that enforces the database‑clock rule for assignments and INSERT positions, plus an integration assertion that verifies pacing delay units. Review `backend/internal/modules/overlay/syncbackoff.go` and updated call sites in `backend/internal/compose/jobs_overlay.go` and integration tests.
- Frontend: `frontend/src/design-system/conformance.test.ts` adds a suite‑level `{ timeout: 60_000 }`; removes per‑test 5s timeouts without changing logic or coverage.

<sup>Written for commit 2aad001bbcfdaa4b26f9e2454b16873196fbd8b1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1344?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/STATUS-ARCHIVE.md
+++ b/STATUS-ARCHIVE.md
@@ -42,6 +42,23 @@ functions now take no clock at all, so the column is written by the one clock
 that reads it, and no future caller can reintroduce a second one. The test needs
 no clock either — the sleep-free, injection-free version is now the honest one.
 
+The rule is a gate rather than a paragraph (`syncclock_test.go`), because the
+defect is invisible at runtime on any machine whose two clocks agree — which is
+every CI runner and every laptop, so no test that exercised the store could have
+failed against it. It derives both write spellings from the package source: an
+assignment's RHS must start at `now()`, and the INSERT column-list position must
+be `now()` outright, with a subject count so a renamed column fails the gate
+instead of emptying it. Beside it, an integration assertion now measures how far
+out the store actually paced the sweep, which due/not-due cannot see: an interval
+handed to Postgres in the wrong unit reads as "not due" exactly like a correct
+one. Both were verified against the defects they describe before being trusted.
+
+The same two clocks are still running in capture's ADR-0063 sync-state, filed as
+[#1346](https://github.com/gradionhq/margince-poc-v1/issues/1346) rather than
+fixed here — a connector-pacing change does not belong behind a flake fix. It is
+the more dangerous half: capture's skew can run the other way and *shorten* a
+backoff, and its rate-limited branch paces on a provider's own `Retry-After`.
+
 **#1340 — a repo sweep sized like a unit test.** The design-system conformance
 gates read, and mostly TypeScript-parse, the whole source tree, against vitest's
 5s per-test default. The heaviest leg measures ~1.1s locally under the coverage
--- a/STATUS-ARCHIVE.md
+++ b/STATUS-ARCHIVE.md
@@ -21,6 +21,37 @@
 > [CHANGELOG.md](CHANGELOG.md) and [README.md → *What works
 > today*](README.md#what-works-today).
 
+## 2026-08-15 — two flakes: one clock too many, one budget too small (PR #1344)
+
+#1329 and #1340, the two load-dependent flakes #1328 filed on its way through.
+Neither was a regression. Both were a measurement taken against something that
+was never going to hold still.
+
+**#1329 — the overlay sweep was scheduled by two clocks.** The due-scan gates on
+`COALESCE(s.next_sweep_at, now()) <= now()`, so `overlay_sync_state.next_sweep_at`
+is only ever compared against the clock *inside* Postgres. `RequestSweep` and the
+flip seal already wrote it with that clock; `RecordSweepSuccess` and
+`RecordSweepFailure` took a `time.Time` from the caller and wrote the app
+process's clock instead. The failure path has minutes of margin and never
+noticed. The success path means *due immediately* and has none: whenever the
+app's clock ran ahead of the database's, the reset landed in the future — a
+healed connection stayed backed off for as long as the skew lasted, and nothing
+anywhere said so. In the integration lane the same skew read as
+`TestSweepBackoffGatesDueOverlayConnections` failing under parallel load. Both
+functions now take no clock at all, so the column is written by the one clock
+that reads it, and no future caller can reintroduce a second one. The test needs
+no clock either — the sleep-free, injection-free version is now the honest one.
+
+**#1340 — a repo sweep sized like a unit test.** The design-system conformance
+gates read, and mostly TypeScript-parse, the whole source tree, against vitest's
+5s per-test default. The heaviest leg measures ~1.1s locally under the coverage
+instrumentation `fe-unit` runs with, and had already blown 5s on a loaded CI
+runner — reported against a job named `fe-unit (vitest + coverage)`, so it read
+as a coverage-threshold failure until you opened the log. The budget is declared
+once on the suite now, an order of magnitude above the worst observed cost:
+these scans are synchronous file I/O, so there is no hang for a tight ceiling to
+catch and a generous one costs nothing. Every gate the file grows inherits it.
+
 ## 2026-08-15 — the three approval-lifecycle gaps #1296 left open (PR #1328)
 
 Issues #1303, #1305 and #1304 — the follow-ups #1296's review filed — shipped as
--- a/STATUS.md
+++ b/STATUS.md
@@ -30,15 +30,12 @@ needs the whole file to start a session.
   inbox and decidable by nobody. The spec PR carries the `expired` verdict and an
   optional `decided_by` into the event catalog (which still declares
   `approved | rejected`, so the build currently diverges), and discharges
-  AUTO-NOTE-1 with AUTO-PARAM-7. **Known follow-ups, all filed**:
-  [#1329](https://github.com/gradionhq/margince-poc-v1/issues/1329) an overlay
-  sync-backoff test races the real clock under parallel load;
+  AUTO-NOTE-1 with AUTO-PARAM-7. **One follow-up still open**:
   [#1335](https://github.com/gradionhq/margince-poc-v1/issues/1335) an
   approved-but-unredeemed agent staging survives erasure for the redemption
-  window (its payload is emptied, its token is not);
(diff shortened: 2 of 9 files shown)


tinnhlt-gra:
## Review round applied (2aad001)

Two reviewers on the backend diff. The security pass came back clean on all four angles it was pointed at — and verified `make_interval`'s parameter typing against the live PG16 rather than reasoning about it. The craft pass found one thing that mattered, and it was the right thing.

### The invariant had no gate

The parent commit stated "every `next_sweep_at` write takes Postgres's clock" in a doc comment and enforced it with nothing. Worse, the defect is **invisible at runtime on any machine whose two clocks agree** — every CI runner, every laptop — so no test that exercised the store could have failed against the bug the comment describes. The tree has now written that same rationale twice (capture's `MarkQueued` carries it for `next_attempt_at`) and both times habit was the only thing holding it.

`syncclock_test.go` derives the rule from the package source: an assignment's RHS must start at `now()`, and the column's position in an INSERT column list must be `now()` outright — matched positionally, the way Postgres matches it, so a statement whose columns and values drifted out of step can't slip through. Scope is this package because `tableownership_test.go` pins `overlay_sync_state` here.

**Verified against the defects it describes**, not just observed green:

| mutation | result |
|---|---|
| `next_sweep_at = EXCLUDED.next_sweep_at` (pre-fix reset) | caught |
| `next_sweep_at = $1` (pre-fix pacing) | caught |
| INSERT stamps `$1` positionally | caught |
| column renamed → gate iterates an empty set | caught (subject count) |

That last row is the one I'd have shipped broken. Both checks key on a column *name*, so a rename would have left them reporting success over nothing.

### A units hole the flake fix opened

`assertPacedWithin` measures how far out the store actually paced the sweep. Due/not-due cannot see this: an interval handed to Postgres in the wrong unit reads as "not due" exactly like a correct one. Mutating `secs =>` to `mins =>`:

```
next sweep paced 9h59m59.997843s out, want within [9m0s, 10m0s]
```

A 60x backoff that every existing assertion would have called correct.

### Smaller findings, applied

- The clock rationale moved to the file header, where every writer of the column is in view — a reader of `RecordSweepFailure` shouldn't have to go read another function's doc for the file's rule.
- Dropped the redundant `::double precision`; `make_interval` has one signature and the in-tree precedent (`capture/autoenrich.go:234`) binds a bare float.
- `isDue`'s comment: my rewrite swapped one false premise for another. "Tests here seed more than one workspace" is true of the package and false of this file. It now attributes the membership filter to the harness resetting once per test, which is checkable.

### Filed, not fixed: #1346

Both reviewers flagged capture's ADR-0063 sync-state running the same two clocks, and the security pass showed my earlier read of it as harmless was wrong. Its skew can run the *other* way and **shorten** a backoff: at ≥96s the pacing is fully defeated and the dispatcher re-syncs a refusing provider every tick. Its rate-limited branch paces on the provider's own `Retry-After`, so that gets silently under-honored too.

Not fixed here — a connector-pacing change across a different module, with its own integration lane to re-verify, does not belong behind a flake fix.

`make check` EXIT=0; overlay integration green; craft static 0/0/0.

Issue #1349, opened 2026-08-15
check-lint-modules.sh lints nothing on macOS: BSD sed leaves go.mod on the path


tinnhlt-gra:
`make check` fails on macOS for every branch, including a clean `origin/main`, with:

```
scripts/check-lint-modules.sh: line 105: cd: backend/go.mod: Not a directory
scripts/check-lint-modules.sh: line 105: cd: cli/craft/go.mod: Not a directory
...
FAIL — golangci-lint findings in: backend/go.mod backend/tools/go.mod cli/craft/go.mod ...
```

Verified against a fresh worktree of `origin/main`, so it is not branch-specific.

## Cause

The module list strips the filename with a GNU-only construct:

```sh
modules="$(git ls-files '*go.mod' | sed 's#/\?go\.mod$##' | ...)"
```

`\?` is a GNU extension. BSD `sed` — the default on macOS — treats it literally, so nothing is stripped and every entry stays `<dir>/go.mod`. Two consequences follow:

1. `cd "$mod"` fails for every module, so **no module is linted**.
2. Because the `cd` failure makes the `if !` branch true, every module lands in `$failed` and the gate reports "golangci-lint findings" — naming files it never opened.

The `grep -v '^backend$'` filter below it silently stops matching for the same reason, which is why `backend/go.mod` appears in a list that is supposed to exclude it.

## Why it matters more than a broken local gate

The failure mode is inverted from the usual one: it reports findings that do not exist rather than missing findings that do. An engineer on macOS sees a red gate naming eight modules, goes looking for lint errors in files the script never read, and finds nothing — which is the worst kind of wasted hour. It also trains people to treat this gate's red as noise.

CI is unaffected (GNU sed), so nothing is being missed on the merge path.

## Fix

A portable spelling, e.g.:

```sh
| sed -e 's#^go\.mod$#.#' -e 's#/go\.mod$##'
```

or `dirname` per line. Worth a portability note beside it: this script is run by humans on their own machines as well as by CI, so a GNU-only construct here is a different risk from one inside a container.

Found while running `make check` on a feature branch; the gate's own diagnostics made it look like the branch's fault until a clean `origin/main` worktree reproduced it.

dungbanguyen:
Closing as fixed.

This reported `scripts/check-lint-modules.sh` failing on macOS because BSD `sed` left `go.mod` on the path, so the script `cd`-ed into a file and lint effectively ran nowhere.

[...]

PR #1350, opened 2026-08-15
feat(capture): a member can see what the pipeline did with their messages

tinnhlt-gra:
The capture pipeline is fully instrumented for writes and invisible to the people it acts for. `audit_log` records the activity that WAS created; the decisions that created nothing — the internal-only drop, the transactional suppression, the deferral, the verdict — are `system_log` breadcrumbs carrying no member, no read surface, and no retention at all. A rep whose message never appeared had no way to ask why.

This adds **Settings → Capture activity**: the last 24 hours of that member's own connections, as a funnel of five outcomes and a row per message. Plus a `/metrics` counter family, because until now the endpoint said nothing about capture at all.

## What it looks like

Attached below: a UAT recording against a real stack — the funnel, the list, both scope toggles, and the operator posture flipped on.

## The shape

**One row per message, five outcomes that partition the window** — `captured` · `internal` · `suppressed` · `deferred` · `fault`. Writing one row per *gate* instead would count a suppressed message twice, and a member reading "what happened to my mail" would see more outcomes than they have messages.

**Visibility is derived, not declared.** `user_id` set means a personal connection and the row is that member's alone; NULL means a workspace-owned channel binding (a Telegram bot, a Zalo OA) whose traffic belongs to no member, and that is what the new `capture_trace` RBAC object reads. **No grant reaches a colleague's mailbox.** The two are separate operations for that reason rather than one with a parameter.

**Reasons that change what an outcome means.** A ceiling-refused deferral is not waiting for a verdict — none is coming. Mail from an already-judged-noise sender lands and is then archived, so "captured" answers "why did this not appear?" with "it did". Both say so.

**Verdict progress is joined from the disposition ledger**, through the activity's counterparty address — so one sender's answer covers every message they sent, and it works with payload capture off, because the address lives on the activity rather than on the trace.

## `capture.trace_payloads` — read this part

Off by default. With it on, each traced message additionally keeps its **sender and a bounded subject for 24 hours, including for messages dropped because every party was internal**. That is the case an operator turns it on to diagnose, and it retains correspondence the CRM otherwise refuses to store.

It is settable **only in `margince.yaml`** — no API, no in-app switch — because that is the installation operator's decision, not the mailbox owner's.

**Two things a reviewer should weigh deliberately:**

1. `backend/api/ai-tasks.yaml` pins `capture_counterparty_verdict` with `no_payload: true` (ADR-0074): content from that task must never reach `ai_call_payload` *whatever the deployment posture says*. This feature writes nothing there and does not change the AI lane, so the pin's letter holds — but it is in tension with its spirit, and that was surfaced and decided rather than discovered. ADR-0082's letter is untouched: the `system_log` breadcrumb still carries connector, source system and external id and nothing else.
2. An **erased subject's address is never written**, whatever the posture. `Trace` consults the same suppression list `recordDisposition` does, for the reason its comment gives — a diagnostic table in payload mode is exactly where an erased address would otherwise come back. Erasure also reaches rows already written, through the real `privacy.Eraser`.

A **channel record's `source_id` is the customer's account id**, which this codebase already treats as personal data, so it is stored hashed. Dedupe is equality and a hash equals itself, so the unique index is unaffected while an erasure inside the window has nothing there to reach. ADR-0082's permission to record an external id was written about mail, where that id names a message rather than a person.

## Scope

Captured **messages**. `Sink.Upsert`'s lead arm has four outcomes of its own and none is a message; the screen and the docs say so rather than leaving a connector silently missing. The funnel starts at the ingress gate — what a connector filtered on its own side is not comparable between connectors, and the page says that too.

## Reviews and what they changed

Two adversarial design reviews before implementation and two code reviews after. The design changed materially as a result:

- **The verdict engine is not a trace writer.** It works on a ledger row keyed by *address* with no natural key, so a trace row would have borrowed the first triggering activity's key — turning "what happened to this message" into "what happened to this sender, filed under one of their messages" — and then collided with itself the moment a sender was re-judged inside the window, silently dropping the newer answer. The read joins the ledger instead.
- **A raw i18n key** would have reached members on every suppressed row: the registry answers `transactional_infra:<domain>` and the catalog falls back to the key itself. The trace stores the class now, and the `as never` that silenced the compile error is gone — an unknown reason renders nothing.
- **An unattributable row carried content into the manager's view.** A personal connector arriving without a member writes `user_id NULL`, the workspace-readable class; such a row now carries no payload at all.
- **`hideUnreadableLinks` asked for the generic row-scope clause**, which refuses `activity` outright — the personal read would have 500'd for any member with a captured message. Found by writing the test for it.

## Verification

- `make check` ✅ · `make test-integration` ✅ (40 packages, 0 skips)
- **New-code coverage 85.9%** (82.5% counting two files this PR only moves)
- AI certification untouched: no prompt, validator or certified pin is in this diff
- UAT against a real stack: all six claims plus the payload flip, recording attached

One pre-existing failure is **not** from this branch: `check-lint-modules.sh` uses a GNU-only `sed` construct and reports phantom findings on macOS, on `origin/main` too — filed as #1349.

Migrations are **0257/0258**: a parallel session held 0251/0252 and `main` gained 0255/0256 mid-branch, so the first numbering would never have applied to a database already past them.

---
## Summary by cubic
Makes capture outcomes visible for the last 24 hours and enforces that window. Previously invisible decisions now surface in Settings and via API, with per-process outcome counters and an hourly sweep.

- `GET /v1/capture/activity` returns the caller’s own rows; `GET /v1/capture/activity/workspace` returns shared‑channel rows gated by `capture_trace`. The API returns a funnel and paginated entries and reports whether trace payloads are enabled.
- Optional `capture.trace_payloads` (default false) keeps sender and a bounded subject for 24 hours; erasure deletes both. Unattributable rows carry no payload, and channel `source_id` is hashed and bounded.
- Counters expose `margince_capture_outcomes_total{outcome=...}` as process counters. Counts measure decisions, not a partition of messages.
- Correctness: idempotent writes via an expression‑index natural key; explicit workspace predicates (no RLS); a scoped LATERAL join prevents cross‑tenant verdict leakage and duplicates.

**Rollout and migrations**
- Apply migrations `0258_capture_trace` and `0259_capture_trace_rbac` (seeds read‑only `capture_trace` to admin/manager/ops).
- Enable the job runner for `capture_trace_sweep` and `capture_trace_sweep_workspace` (hourly; includes archived workspaces). Prometheus will pick up the new counters.
- No config change required unless enabling `capture.trace_payloads` in `config/margince.yaml`.

<sup>Written for commit a0a6723374998f1aaad6ed2225fc832f5e326186. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1350?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added a **Capture activity** settings tab showing personal capture outcomes from the last 24 hours.
  * Authorized viewers can review shared workspace activity, with pagination, outcome summaries, reasons, timestamps, and optional payload details.
  * Added configurable payload visibility, disabled by default.
  * Added capture outcome metrics and automatic 24-hour trace cleanup.
  * Added privacy protections, including subject erasure handling and scoped activity links.

* **Documentation**
  * Documented capture activity visibility, retention, configuration, privacy behavior, and permissions.
  * Added localized interface text in English, German, and Vietnamese.


tinnhlt-gra:
## UAT — Settings → Capture activity

Run against a real stack booted from this branch (`make dev`), signed in as the bootstrap admin. The pipeline's write path is covered by the integration tests; what needed a human eye was the screen, so the trace rows were seeded directly to cover every state it must tell apart.

**Recording:** 26 frames — login, the funnel, the list, both scope toggles, and the reload after the operator posture was flipped on.

### Claims verified

| Claim | Result |
|---|---|
| Funnel shows five outcomes, and a zero renders as `0` rather than being omitted | **PASS** |
| Every row reads "content not stored" (payload capture off by default) | **PASS** |
| A capped deferral says a verdict is **not** coming, not merely "waiting" | **PASS** (see below) |
| A `noise_prior` row says it landed but will be archived — not a plain success | **PASS** |
| The Shared channels toggle appears for a grant holder, and shows the `user_id IS NULL` row and **none** of the admin's personal rows | **PASS** |
| The page states where the numbers start | **PASS** |
| `capture.trace_payloads: true` → sender and subject render; rows without them still say so | **PASS** |

The zero case is worth naming: the API omits zero outcomes from the JSON entirely (`"funnel": {"captured": 1}`), and the screen still draws all five tiles with `0`. "Nothing was dropped as internal today" is exactly what somebody opens this page to confirm, so the tile has to be there.

The payload flip also confirmed the distinction that matters: with capture **on**, rows that genuinely carried nothing switch from *content not stored* to *no sender recorded* — the screen never blames the operator's posture for a deletion somebody requested.

### Two defects it found, both fixed in this PR

1. **A raw i18n key rendered at a member.** The `transactional_infra` reason had no catalog entry in any of the three locales, and the catalog falls back to the key itself, so the row read `captureActivity.reason.transactional_infra` verbatim. The `as never` on that lookup is what suppressed the compile error that would have caught it. Fixed: the keys exist, the cast is gone, and an unknown reason now renders nothing rather than an identifier.
2. **The outcome argued with its own explanation.** A capped deferral read "Waiting on a verdict" above "no verdict is coming". It reads "Not queued" now.

Both have regression tests.

### Not verified

Narrow-width layout. `resize_window` reported success but kept returning the 1512px viewport, so it was left alone rather than ground at. The CSS folds the row grid to a stack at 48rem; worth a manual pass at ~800px before anyone relies on it.

tinnhlt-gra:
**On the recording:** the GIF is at `.tmp/capture-trace/uat-capture-activity.gif` (2.3 MB, 26 frames, 1512×830) on the machine that ran the UAT. It is not attached here and not committed: GitHub only accepts image uploads on comments through the web UI, and `CLAUDE.md` treats captures as debris that must stay out of the repo. Drag the file onto this comment box if you want it inline on the PR.

Correcting my own wording in the comment above — "Recording: 26 frames" reads as if it were attached. It is not; the path is the artifact.

tinnhlt-gra:
@coderabbitai review

tinnhlt-gra:
## SonarCloud: two fixed, one pushed back

Quality gate passed (84.8% coverage on new code, 0% duplication), with three new issues. Two were fair and are fixed in `87c11a0`:

- **`Upsert` cognitive complexity 17 > 15.** This branch is what pushed it over, so the arm it added — the doomed-transaction trace — is now its own function. That is where its reasoning belongs anyway: a trace written inside a transaction that is about to roll back goes with it, and this is the outcome a member most needs explained.
- **Empty `FleetWide()`.** Its body now says it is a marker method, satisfied rather than called.

**Pushing back on the third: `policy.go:207` — "34 parameters, greater than the 7 authorized".**

Sonar is right about the shape and wrong about the remedy here. `objects()` takes one grant per RBAC object as a positional parameter, and this PR added the 34th because it adds an object. Fixing it means changing the signature and all five seeded role documents in the same edit — in the file that decides what every role in every workspace may do, with a merge-blocking parity test, a committed replay fixture and a published matrix hanging off it.

That is a refactor worth doing and worth doing **alone**, where the diff is legible and a reviewer can see that no grant moved. Bundling it into a feature PR is how a permission quietly changes under cover of something else. Filed as a follow-up rather than done here.

The finding is also pre-existing in substance: the function was over the threshold at 33 parameters and at 20. This PR made it one worse, not newly bad.

tinnhlt-gra:
## Review round: all 11 threads addressed, plus two internal reviewers

Four reviewers ran over this diff — CodeRabbit, an adversarial security pass, a craftsmanship pass, and a UAT against a real stack. Between them they found **two defects that would have shipped as data-protection bugs**, and I want the first one read carefully rather than skimmed.

### Cross-tenant read (CodeRabbit 🔴 + security pass, same line)

The resolution join keyed on an address alone:

```sql
LEFT JOIN capture_pending_counterparty d ON d.email = a.counterparty_email
```

No workspace predicate — and an address is not unique across tenants, so a member could be shown **another installation's verdict about their own sender**. There is no RLS behind these tables since `0217`, which makes the predicate the entire isolation, and this query spelled it for the trace row and forgot it for both joins.

The same line also fanned out: the ledger keeps a row per address per state, so a sender judged twice made one message appear twice — inflating the page past its own `LIMIT` and corrupting the keyset cursor.

Now a scoped `LATERAL` taking one row, newest first with an open question ahead of a resolved one.

### The sweep never ran on a multi-workspace database

The child bound `InstallationDB`, which answers `ErrMultipleWorkspaces` as soon as more than one live tenant exists. Every child would have failed, exhausted its attempts and deleted nothing — while the read kept filtering to 24 hours and the table grew invisibly behind it. Under the payload posture those rows are addresses and subject lines. It binds the tenant its args name now. It also skipped **archived** workspaces, which is wrong for a retention pass: storage limitation does not pause because a tenant stopped logging in.

### The rest

- **Erasure reached the address but not the subject.** A subject line routinely names somebody ("Re: intro — alice@acme.test"), so a message from a *different* sender kept her address after her request completed — and a trace written afterwards would have added it back. Two lanes now.
- **`source_id` was unbounded.** For mail it is the sender's own Message-ID, and the natural key is a btree over it: an oversized header raises `index row size exceeds maximum`, and since `Trace` runs on the capture transaction, that error fails the **capture**.
- **I claimed something untrue in three places** — that the five outcomes partition the window. They do not: `outcome` is in the unique key, so a message that landed and was later refused on a replay holds both `captured` and `fault`, and my own rollback test relies on that. The comments now say these count decisions, not messages.
- **Frontend:** both `as` casts removed (each route fetched on its own branch, so the response is typed by the route that produced it); `resolution.status` now takes the same closed-set guard the reasons do, since it carried the identical raw-key exposure; the suppressed label no longer names one cause for an outcome with several.

### The tests

Two new ones exist because **a single-workspace fixture cannot fail**: a second tenant, and a sender with a history. Both were run against the unscoped join *before* the fix and both failed with exactly the right diagnosis — the neighbour's `Status:real` arriving as mine, and one message returning as two. The readable-link test was also proving nothing (`RowScopeAll` short-circuits the probe before it runs), so it uses a bounded reader now. Pagination has a test at all for the first time.

`make check` and `make test-integration` green. The one red gate is #1349, which fails identically on `origin/main`.

tinnhlt-gra:
uat-capture-activity

PR #1354, opened 2026-08-15
feat(channels): the installation says which transports it has, and what to call them

tinnhlt-gra:
Slice 1c of the SP5 channel-axis arc, on top of #1332. Carries **ADR-0107/A158**'s discovery endpoint; upstream landed first (margince-foundation #1305 for the operation, #1309 naming its response shapes so the generated Go reads `ChannelProviderDirectory` rather than `ListChannelProviders200JSONResponseBodyData…`).

## Why

`ProviderRef` published an invariant with no resolver: the contract says a provider names a transport *this installation* registered, and gave a client no way to find out which those are. After the narrowing, the raw id is all a timeline has — the kind no longer carries the name.

## Two distinctions the endpoint exists to keep

**REGISTERED is not SENDABLE.** `whatsapp` is registered with no composed connector, so it is published with `supplies_transport=false` rather than hidden. The snapshot therefore holds two sets: every registered provider (what a message MAY name) and the composed subset (what a reply CAN leave on).

**A DEPLOYMENT fact is not a TENANT fact.** `supplies_transport` says what this binary composed; whether *this workspace* can send reads `channel_connection` and stays in the send pre-flight.

## What the security review caught

**The endpoint failed quiet, by default.** The snapshot was written only inside `NewCaptureRegistry`, which the api role builds *only when a keyvault root key is configured*. On a vault-less install it answered **200 with an empty list** — every timeline rendering raw ids, and an agent told the vocabulary is empty while `log_activity` still demands a value from it. It now loads from the registry table at server assembly, for every role serving `/v1`, which also makes the new columns load-bearing rather than write-only.

Sharper still: **my own test could not have caught it**, because it read the same process-global another test in the package had populated. The replacement clears the snapshot and reloads from the table — the state a registry-less role actually starts in.

Two things I had written that were simply false, both fixed: the migration claimed the columns exist because `SendScopeFor` needs the credential model (it does not — it derives from an in-memory set), and the test holding the seeded label against the booted one passed **vacuously** (SQL `initcap` vs a Go word-break rule differ only on a multi-word id, and every provider today is one word).

## A decision, made rather than inherited

The review asked whether a unit's provider id — an operator's choice of what to install — should reach every authenticated seat. **Yes**, and deliberately: a member whose timeline shows a message that arrived on a unit's channel needs to know what to call it. What is gated is the **act**, not the **name** — sending is an RBAC question answered by the unit's own `ext_<unit>_<object>` grants and the send pre-flight. Disclosure and capability are separate axes.

## Gates

| Gate | Result |
|---|---|
| `make check` | exit 0 |
| `make test-integration` | **OK, 0 skips, 40 packages** |
| `make craft-static` | 3/3 PASS (0 blocker / 0 major / 0 minor) |
| UAT, live stack | **ALL CHECKS PASSED** — see the comment below |

Migration renumbered `0252` → `0257` (main took 0252–0256 while this was open); the comments that named the number no longer do, per this repo's own trap list.

---
## Summary by cubic
Introduces `GET /v1/channel-providers` and the `list_channel_providers` agent tool to resolve provider names and capabilities at runtime. Previously clients printed raw provider ids; now the directory returns a display `label`, a `credential_model`, and whether the provider `supplies_transport`. The endpoint reads the boot snapshot loaded from the registry table, fixing the silent-empty behavior on vault-less installs.

- Registered is not sendable: the directory publishes every registered provider and marks the composed subset via `supplies_transport`. Sendability for a workspace still comes from `channel_connection` during send pre‑flight.
- Deployment fact is not tenant fact: `label` and `credential_model` are properties of the composed connector; the directory is readable by any authenticated seat and contains no operator-configured data.
- Server/agents/frontend:
  - Composition root upserts display facts into `channel_provider`, snapshots the composed set, and loads the directory from the table at server assembly for all roles.
  - Agents expose `list_channel_providers`; schemas and conformance tests updated.
  - Frontend adds `useChannelProviders` and `useProviderLabel` to render labels, falling back to the raw id if unknown or on fetch failure.
- Tests ensure the directory matches the registry, seeded labels match boot labels (including underscore word‑breaks), and the load path works without a capture registry.

**Rollout/Migration**
- Apply migration `0257`: adds `label`, `credential_model`, and `supplies_transport` to `channel_provider`. No operator action needed; reconcile refreshes values on boot.
- Client change required: use `/v1/channel-providers` (or the agent tool) to render provider names. Do not infer send capability from registration; rely on send pre‑flight for tenant gating.

<sup>Written for commit 00234851b4fb228662cf45807abd9267e5e93f0c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1354?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tinnhlt-gra:
## UAT — live stack, all checks passed

Run against an isolated dev stack (`DEV_SLUG=sp5axis`, api `:18948`, db `margince_dev_sp5axis`) — a real boot with the migration applied. Script: `.tmp/ext-core-write-port/uat/uat-sp5-1c.sh`.

Each check is there because it can fail independently of the others:

1. **Boot wrote the three facts** into the registry — not just that the columns exist, but that something filled them.
2. **Every registered transport is published** — the count is compared against the registry, so publishing a subset fails. This is the check that caught the original defect, where only the *composed* set was published and `whatsapp` vanished.
3. **Registered is not sendable** — `telegram` true, `whatsapp` false. A version that reported everything sendable would promise a reply with no path to happening.
4. **The label is the human spelling** — `WhatsApp`, not the title-cased `Whatsapp`. This is the pair most likely to drift, because the SQL seed and the Go reconcile hold the same exception separately.
5. **Anonymous is refused (401), an authenticated seat is served (200)** — the authorization decision, exercised from both sides rather than asserted.

```
$(cat /tmp/uat1c-clean.txt)
```

tinnhlt-gra:
**New-code coverage: 87.4%** (146/167 changed statement lines), measured against the merge-base rather than `origin/main`'s tip — the tip moves mid-flight and attributes other people's merged files to the diff.

Union of the unit profile and the integration lane's per-package covdata. The uncovered lines are the HTTP handler body and the agent tool's error arms — both reached only through a live request or a directory read that fails, and the shaping functions they delegate to (`publishedChannelProviders`, `channelProviderFactsFor`) are covered directly. That split is deliberate: the handler is three lines of wiring around a function the tests exercise, so covering it would mean asserting the wiring rather than the behaviour.

PR #1357, opened 2026-08-15
fix(send): an agent-scheduled message fires as the agent that scheduled it

### LINKED ISSUE #1258: An agent-scheduled send fires under a fabricated agent identity
Found by review during ADR-0104 (scheduled send).

## What is wrong

`scheduled_send` stores the authorizing human's id and a `principal_kind`. At fire, an agent execution is reconstructed as `agent:<human-user-id>` with no passport id and no on-behalf-of. That is **not** the agent principal that scheduled the message.

## Why it matters

Two different agents, or two different passports, acting for the same human collapse into one invented actor. The release audit row, the activity's `captured_by` and the outbox envelope then cannot say which agent produced the message, and the agent-under-human attribution chain ADR-0055 depends on is broken exactly where an incident investigation would look.

The part that IS correct today: `principal_kind` is preserved, so an agent-scheduled message still fires without the human's signature or display name. That was the defect worth blocking on and it is covered.

## Fix direction

Persist immutable scheduling-time provenance — the original agent id, the passport id, the on-behalf-of human — separately from the live authority the fire re-derives. The live authority stays the ceiling; the stored provenance is what the audit names.

## Why it was deferred

It is an attribution defect in the audit trail, not an authorization one: the message still fires under a ceiling the human currently holds, and still renders as an agent's message.

LarsGradion:
Closes #1258.

## The defect

A `scheduled_send` row stored the authorizing human and whether the actor was a human or an agent — and nothing else about the actor. At fire the worker rebuilt an agent as `agent:<the human's id>`: an identity no agent ever had.

Every agent, and every passport, acting for one person collapsed into that single invented actor. `audit_log.actor_id`, the activity's `captured_by` and the outbox envelope then could not say which agent produced a message — the attribution chain ADR-0055 rests on, broken exactly where an incident investigation looks.

## The fix

Core `0260` stores the scheduling-time record — the agent's own principal id, its passport, and the human it acted for — and the fire path reads those instead of deriving one.

`agent_actor_id` is the anchor: the one fact the fire path cannot reconstruct. The passport and the human are independently optional beside it, so each NULL means something specific rather than "missing" — no passport presented, or no human named. The constraint forbids provenance with no actor, and any agent provenance on a human-scheduled row.

## Authority is re-derived, not restored

Two separate gates, both live at fire:

1. **The human's seat and grants** are re-read through `EffectiveAuthority` and remain the ceiling. "agent ≤ human" — a stored identity names the actor without widening what it may do.
2. **The passport is re-authenticated** through `AuthenticateAgentByID`, which re-runs revocation, expiry, the granting human's status, and the connection it belongs to. It must also still belong to the human the row is charged to.

The second gate exists because the first cannot see it: a revoked passport leaves the human active with the same grants, so without it a message an operator revoked an agent's credential to stop would send anyway. That was a real defect in this PR's first three rounds, found by CodeRabbit.

A message refused by either gate is **held**, never dropped, under a reason naming what a human can act on — `passport_revoked` is its own reason rather than `sender_inactive`, because the human's account is fine and sending them to check it would waste their time.

## Rows written before this migration

They keep NULL provenance and fall back to the old derived id. A backfill would have to invent the very identity this change exists to stop inventing — those rows genuinely cannot say which agent it was, and NULL is the honest record of that.

## Verification

Three integration tests, each mutation-checked:

- `TestAnAgentScheduledSendFiresUnderTheAgentThatScheduledIt` — schedules under a real passport, fires through the production worker, asserts the audit names that agent. Restore the derived id and it fails.
- `TestAnAgentWithNoHumanBehindItStillRecordsWhichAgent` — an agent naming no human records its own id and no human. Store nothing instead and it fails.
- `TestARevokedPassportHoldsTheMessageItWasScheduledUnder` — revokes the passport, asserts the message is held with no delivery staged. Skip the re-authentication and it sends.

## One note for the reviewer

`make lint-modules` fails on this branch **and on clean `origin/main`** — `scripts/check-lint-modules.sh:105` does `cd "$mod"` where `$mod` is a `go.mod` file path. Verified against a detached worktree of main; filed as [#1358](https://github.com/gradionhq/margince-poc-v1/issues/1358) and left alone rather than fixing another gate inside this PR.

Everything else passes: build, vet, lint, arch-lint, go-file-length, drift, test, check-composition, migration-versions, both integration lanes, and `craft-static`.


LarsGradion:
@coderabbitai full review

Three decisions here are judgment calls rather than obvious, and I would rather have them challenged now than discover the disagreement in production.

**1. The CHECK constraint deliberately does NOT require every agent row to name its agent.** It says a row names it *completely or not at all*:

```sql
(principal_kind = 'agent' AND agent_actor_id IS NOT NULL AND agent_on_behalf_of IS NOT NULL)
OR (agent_actor_id IS NULL AND agent_passport_id IS NULL AND agent_on_behalf_of IS NULL)
```

The all-NULL arm is what keeps pre-migration agent rows valid, since they cannot be backfilled — inventing an identity is the exact defect being fixed. The cost is that the schema alone cannot tell a legacy row from a writer that forgot to populate the columns. I think the writer plus the test is the right place for that, but argue the other side if you see a way to distinguish them in the schema.

**2. `agent_passport_id` is nullable while the other two are effectively required for an agent row.** An in-process workflow agent legitimately has no passport, so NULL means "no passport was involved" rather than "we lost it". Check whether that is actually true of every agent path that can reach `SendOrSchedule`, and whether a passport-less agent should even be allowed to schedule mail.

**3. `provenanceOf` falls back from `OnBehalfOf` to `UserID`** when the former is zero. My reasoning is that both name the same human and the column has to be non-NULL for the agent arm. If there is a path where those two are legitimately DIFFERENT people, that fallback silently records the wrong human, and I would want to know.

Also worth a look: the fired principal keeps the **human's** grants while carrying the **agent's** identity. That is intentional — "agent ≤ human" is the ceiling and a stored identity must not widen it — but it means the actor id and the permissions now come from different sources, so please check I have not created a path where the identity is used for an authorization decision somewhere downstream.

LarsGradion:
I checked two of my own three questions rather than leaving them open. Both answers are now in the code comments (commit `113eb4ede`), because both comments previously stated things I had not verified.

**Q3 — can `OnBehalfOf` and `UserID` name different people?** No, not on any real agent path, and the reason is stronger than my original comment said. `identity.AgentIdentity.Principal()` ([passport.go:300](backend/internal/modules/identity/passport.go#L300)) sets **both** fields from the same `a.OnBehalfOf`, and `auth.Admit` ([admit.go:98](backend/internal/platform/auth/admit.go#L98)) derives the agent's seat and RBAC from `OnBehalfOf`. An agent whose two fields disagreed would already be acting under a ceiling belonging to somebody other than its `UserID` — a much bigger problem than this row. The fallback is safe.

**Q2 — can a passport-less agent reach a send?** Not today, but the shape is real and my comment cited the wrong example. I wrote "an in-process workflow", which was invention. The actual passport-less agent principal is [extjobsrun.go:279](backend/internal/compose/extjobsrun.go#L279), which mints one for an extension tick with an agent's own `app_user` id as `UserID` and no `OnBehalfOf`. That path cannot send mail — I checked, there is no `SendOrSchedule` or `SendEmail` call reachable from the extension runtime — so every agent that reaches scheduling today carries a passport. The nullable column stays conditional rather than assuming.

**Q1 — the CHECK constraint's all-NULL arm — is still genuinely open** and is the one I would most like challenged. The schema cannot distinguish a legacy row from a writer that forgot to populate the columns, and I could not find a way to make it without breaking the pre-migration rows.

Separately: `deterministic-gates` is red on this PR. I expect it is [#1358](https://github.com/gradionhq/margince-poc-v1/issues/1358) — `check-lint-modules.sh` `cd`s into a `go.mod` file, which fails on clean `origin/main` too — but the run was still queued when I looked, so I have not confirmed it from the CI log yet and am not claiming it until I have.

LarsGradion:
**A stop-time Codex review caught a real defect I introduced, and it is worth reading — I made the same mistake I was fixing, one layer down.**

Codex flagged that my two comments contradicted each other. Following it found a bug, not just wording.

The first comment claimed a UserID-only agent "still names a real human". The second said [extjobsrun.go:279](backend/internal/compose/extjobsrun.go#L279) mints exactly that shape with an **agent's** own `app_user` id as `UserID` — the query is `WHERE id = $1 AND is_agent`. Both cannot be true, and the second is the accurate one.

So my `OnBehalfOf → UserID` fallback was writing **an agent's id into `agent_on_behalf_of`**, a column that means "the human behind this agent". The fire path hands that to `actor.OnBehalfOf`, and `auth.Admit` ([admit.go:98](backend/internal/platform/auth/admit.go#L98)) derives seat and RBAC from `OnBehalfOf`. That is a fabricated authority — the same class of defect this whole record exists to end. I introduced it while fixing the identical shape one layer up.

**Fix:** an agent with no `OnBehalfOf` now stores no provenance at all. That row keeps the pre-migration behaviour, which is the honest outcome for an actor that never named a human, and the CHECK's all-or-nothing shape is what makes it representable. `TestAnAgentWithNoHumanBehindItStoresNoProvenance` covers it, mutation-checked: restore the fallback and it fails.

To be precise about severity: **no agent that can reach a send carries this shape today.** The tool surface is the only agent door to `SendOrSchedule` and its principals are passport-minted with both fields set from one value. This is a guard on the writer, not a live bug that was reachable.

Two other things since the last push:

- Rebased onto current main, which took `0257` with #1354, so the migration is now **`0258`**. The new `check-migration-versions.sh` caught that collision locally before the push — second time today it has earned its keep.
- `contract-breaking-check` was red before the rebase purely because main had added an API path my branch predated. Clean after rebasing; my branch does not touch `crm.yaml`.

Local state: every `make check-backend` target passes except `lint-modules`, which is [#1358](https://github.com/gradionhq/margince-poc-v1/issues/1358) and fails on clean main too. Both integration lanes green, `craft-static` clean.

LarsGradion:
**Second Codex stop-gate finding, and it was the original bug returning through the side I had just closed.**

Making a passport-less agent store *nothing* left a new row indistinguishable from a pre-migration one. The fire path reads a NULL actor id as "this row cannot say which agent it was" and falls back to the derived `agent:<human-uuid>` — so the invented identity came back for exactly the actor whose real id had been in hand at schedule time.

The mistake was upstream of both attempts. My CHECK required the actor id and the human to travel together, which forced a choice between **fabricating a human** and **discarding a known agent**. Neither is acceptable, and neither had to be — that was a constraint I wrote, not a fact about the domain.

`agent_actor_id` is now the anchor: the one fact the fire path cannot reconstruct. The passport and the human are independently optional beside it, so each NULL means something specific rather than "missing":

- no passport → none was presented (an in-process agent)
- no human → the actor named none

The constraint still forbids provenance with no actor, and any agent provenance on a human-scheduled row. The all-NULL arm now covers only pre-existing rows — which is what makes the fire path's reading of a NULL actor id unambiguous, because after this migration an agent row always names its agent.

`TestAnAgentWithNoHumanBehindItStillRecordsWhichAgent` asserts both halves: the row records the agent and no human, **and** the fired message is audited under that agent rather than an invented one. Mutation-checked: restore the store-nothing behaviour and it fails.

The test file crossed the 1000-line ceiling again with the second assertion, so both agent-provenance tests and their helpers moved to `scheduledsendagent_integration_test.go`. Three files now, each under the ceiling and each one concept.

---

Worth saying plainly, since it is three rounds on the same few lines: **the first two attempts were both wrong, in opposite directions, and neither was caught by CI.** All checks were green on the version that fabricated a human, and green again on the version that discarded the agent id. The gates prove the code runs; they had nothing to say about whether the record was true. That is what the stop-gate review caught both times.

LarsGradion:
@coderabbitai Your blocking finding was correct, and it contradicted a claim I made in this PR's own description. **"A revoked passport still stops the send" was not true.**

`fireAs` re-read only the human's authority. It copied the stored passport onto the fired principal and never checked it — so a passport revoked between scheduling and firing changed nothing. The human is still active with the same grants, so the message went out under a credential nobody honours. Revoking an agent's passport is exactly how an operator stops that agent from acting, and for scheduled mail it did not.

**Fixed as you described.** The passport is re-authenticated at fire through `AuthenticateAgentByID`, which re-runs the same liveness rules the token path runs — revocation, expiry, the granting human's status, the connection it belongs to. Its own doc comment turned out to describe this case exactly: *"a parked overnight job wakes up with exactly the authority the passport still has, not the authority it had when enqueued."* I also added your ownership check: a passport that has moved to another human is not the credential this message was scheduled under. The human's live grants stay the ceiling; this is a second gate, not a replacement.

One departure from your suggestion, deliberately: the refused message is **held under its own reason** rather than the existing `sender_inactive`. That reason would have been a lie — the human's account is fine — and a rep sent to check their account when their agent credential was revoked looks in the wrong place. `fireAs` now returns which refusal it found, and `passport_revoked` carries a sentence saying what happened and that they can send it themselves.

`TestARevokedPassportHoldsTheMessageItWasScheduledUnder` is the test you asked for: schedules under a real passport row, revokes it, asserts the message is held with no delivery staged. Mutation-checked — skip the re-authentication and it sends.

**A side effect worth reporting:** my existing attribution test started failing, because it scheduled under an *invented* passport id that was never issued. The new check correctly refused it. That test now issues a real passport — which is the case it meant to exercise all along, and its failure was the check doing its job rather than a regression.

On your other three answers: agreed on all counts, and thank you for checking #2 and #3 against the real paths rather than taking my framing. The `provenance_state` column in #1 is a good idea I am not taking in this PR — it would make the legacy/recorded distinction schema-visible, and I have opened no issue for it yet because I would rather hear whether you think it is worth the column before I do.

Local: every `make check-backend` target green (including `lint`, which caught the nesting this added), both integration lanes green, `craft-static` clean. The migration is now `0260` — main took `0258` while this was in review.

LarsGradion:
Following up on your judgment-call #1 — the `provenance_state` column — with a fact I should have checked before deferring it.

**The legacy arm may have close to zero members.** `scheduled_send` first reached main on 2026-08-14 (it rode in on the 0240 collision fix, #1275), so the window in which an agent could schedule a message *without* provenance is roughly one day wide. My dev database has zero `scheduled_send` rows of any kind, agent or human.

That cuts both ways, and I want to be honest about which way I read it:

**For your suggestion:** if the legacy population is empty or near-empty, then the all-NULL arm is protecting almost nothing while permanently weakening the constraint for every future row. A `provenance_state` column initialized to `legacy` for existing rows and then defaulted-away would give the schema a real "this row was written after 0260 and must be complete" rule — and a future writer that forgot the columns would be rejected rather than silently classified as legacy. That is a genuine improvement over what I have.

**Against, and why I am still not doing it here:** it needs a real count from a deployed installation, not my dev box, and I do not have one. If some installation *does* hold agent-scheduled rows from that day, an unconditional `NOT NULL` state would need the same care I have been paying all along. Adding a column and a state machine on the strength of "my laptop says zero" is the kind of confident move that has already cost this PR three rounds.

So: filed as [#1362](https://github.com/gradionhq/margince-poc-v1/issues/1362) rather than dropped, with your reasoning and this timing evidence in it. If you think the one-day window is decisive enough to do it now, say so and I will — I would rather be argued into it than guess.

Also correcting something from my previous comment: I said `deterministic-gates` was red on the `lint-modules` bug. **That was wrong.** I pulled the log instead of assuming, and it was `contract-breaking-check` — main added `/capture/activity` paths in #1350 after my branch point, so the checker read them as removed by me. My branch does not touch `crm.yaml`. Rebasing cleared it, and that check now passes locally too. The `lint-modules` breakage ([#1358](https://github.com/gradionhq/margince-poc-v1/issues/1358)) is real but separate, and does not gate CI.

PR #1364, opened 2026-08-15
feat(license): require a license in production, and refuse a seat past the licensed count

lstrojny:
This PR makes the license binding. It adds two rules:

1. A production installation does not start without a license.
2. An installation does not create a new seat past the number the license grants.

This is slice 2 of #1190. It is stacked on #1274, which adds the license screen and the seat meter. **#1274 must merge first.**

## Rule 1: the boot needs a license

Before this PR, an installation with no license started everywhere. It printed a warning. Now the deployment posture decides.

| Posture | No license | License refused |
|---|---|---|
| production (`MARGINCE_ENV` unset or unknown) | boot fails | boot fails |
| `dev`, `staging`, `test` | boots, logs a warning | boot fails |

`MARGINCE_ENV` is fail-closed. An installation that sets nothing is production.

Development and CI need no change. `backend/Makefile` exports `MARGINCE_ENV ?= dev`. `scripts/dev.sh`, the integration runners and the `live-boot` job set it too. Production containers set nothing.

The error names both fixes:

```
no license is configured and this installation is production: point
license.token_file (or MARGINCE_LICENSE) at the license token issued for this
installation, or, if this is a development or test installation, set
MARGINCE_ENV=dev
```

Both are named because the reader is in one of two situations. Either the installation is licensed and lost the token reference in a redeploy. Or it is a development installation that never declared itself one. The process cannot tell these apart.

A refused license fails the boot in every posture. A non-production posture is a statement that there is no license. It is not a way to run a license the module rejected.

## Rule 2: the seat limit stops the next seat

Two operations create a full seat. Both are refused when all licensed seats are in use.

| Operation | Refused when |
|---|---|
| `POST /v1/users` (invite) | all licensed full seats are in use |
| `POST /v1/users/{id}/reactivate` | the seat is `full` and all licensed full seats are in use |

The response is `403` with code `seat_limit_reached`:

```
seat limit reached: this installation's license grants 10 full seats and 10 are
in use; deactivate a member, or raise the licensed seat count
```

The counts are in the message. An admin needs them to choose between the two fixes.

What the limit never does:

- It never demotes a seat.
- It never ends a session.
- It never deactivates an account.
- An installation over its limit keeps every user working. Only the next seat is refused (P7).

What frees a seat: deactivation, and suspension. Read seats are not counted at all.

What is not capped: a license with no seat count, and an unlicensed development installation. Bootstrap is also not gated. The installation writes its own admin and Agent Runner seat before any check could refuse them.

## Two decisions from #1190

**The error.** New sentinel `apperrors.ErrSeatLimitReached`, mapped to `403 seat_limit_reached`. The existing sentinels do not fit. `ErrBudgetExceeded` is `429` and means "retry later", which is never true here. `ErrSeatTierInsufficient` is about the caller's own seat. The spec change is margince-foundation#1312 and **must merge first**, because `apperrors` is the fixed registry of `interfaces.md §0`.

**What an unlicensed installation may do.** The question no longer applies in production, because production always has a verified license. The remaining uncapped cases are a license without a seat count and a development installation. Both cap nothing.

## The seat count

The meter and the limit now run one query, `fullSeatsInUseQuery`. The number on the screen is the number the refusal uses.

Writing that query showed a defect in #1274. The comment said a suspended seat is not counted. The query counted every seat that was not deactivated, so it did count suspended seats. No test covered suspension.

The query now reads `status NOT IN ('suspended','deactivated')`. It names the withdrawn states, not the live one. A status added later is then counted until someone decides otherwise. The reverse would stop metering a new state silently, and the installation would issue seats the license never granted.

## Concurrency

The check runs inside the transaction of the writer, under one advisory lock per installation. Without the lock, two admins inviting at the same time both read one free seat and both take it. The installation ends up one seat over the limit, and nothing brings it back.

## Wiring

`WithLicensePosture` is the only wiring point. It already built the license screen handler and the `/metrics` section. It now also binds the seat limit, from the same posture, read at call time. A role that wires no posture enforces no limit. A role that reports a limit cannot fail to enforce it.

## Text changes

The screen said "nothing is blocked". That was true before this PR. The text now states both facts: nobody loses access, and no new member can be invited. Changed in `en`, `de` and `vi`. The test asserts both statements.

The `seats_used` description in `crm.yaml` and the counting note on the screen also said "not deactivated". Both now say suspended seats are excluded as well.

## Tests

`backend/internal/modules/identity/seatceiling_integration_test.go`, against a real database, through the real writers:

- refused at the limit, with both counts in the message, and no row written
- the last licensed seat is admitted, the next one is refused
- a suspended seat and a deactivated seat each free one seat
- reactivation is refused and the member stays deactivated, then succeeds when a seat is licensed
- a read seat is never refused
- no limit and no wiring both admit every seat

`backend/internal/compose/integration/seatceiling_http_integration_test.go`, against the composed api:

- the number from `GET /v1/installation/license` is the number `POST /v1/users` is refused against
- `403 seat_limit_reached`, and the usage does not change after the refusal
- one more licensed seat, and the same request returns `201`
- the next invite is then refused in the same process, without a restart

`backend/internal/compose/license_test.go`: production without a license fails and names both fixes; a non-production installation boots and logs it; a refused license fails in both postures.

## Gates

Green locally: `make -C backend check`, `make check-fe`, `craft static --strict` over the diff, and both integration packages.

Red on CI for one reason that is not in this diff: `main` took `core/0260` for `scheduled_send_agent_provenance` while #1274 uses `0260_license_rbac`. The renumber belongs to #1274.

No STATUS.md entry. #1274 carries the narrative for this work.

### DIFF
--- a/backend/api/crm.yaml
+++ b/backend/api/crm.yaml
@@ -6663,21 +6663,24 @@ paths:
         the process runs, so this reports what the installation last resolved rather than
         asking anything.
 
-        `state` is the posture: `valid` (a license the module honored), `absent` (none
-        configured — a supported state that runs), or `rejected` (one the module refused; an
-        installation in that state does not serve, so a client never sees it from a live
-        server).
+        `state` is the posture: `valid` (a license the module honored), `absent` (no license
+        configured, which only a non-production installation may run as), or `rejected` (one
+        the module refused). A serving process boots on a license or does not boot, so a client
+        reading this from a live server sees `valid` — or `absent` on a development
+        installation.
 
         `seats_granted` is ABSENT when there is no seat cap to report — no license, or one
         whose grant carries no seat count. That is not the same as a grant of zero seats, and a
         client that renders a missing value as `0` would tell an admin their license permits
-        nobody. `seats_used` counts full seats that are not deactivated; read seats are
-        unlimited and never metered (A62/ADR-0047).
+        nobody. `seats_used` counts every full seat the installation has not withdrawn —
+        neither deactivated nor suspended; read seats are unlimited and never metered
+        (A62/ADR-0047).
 
         `over_limit` is the server's own verdict on the pair, so a client cannot arrive at a
-        different answer than the one the installation acts on. Being over the limit is
-        reported, never enforced here: the workspace keeps working, which is P7's
-        warning-then-grace rather than a silent mid-month lockout.
+        different answer than the one the installation acts on. Nobody is ever demoted or
+        evicted by it and no session ends — what an exhausted entitlement stops is the NEXT
+        full seat, refused at `POST /users` and at reactivation with `seat_limit_reached`.
+        That is P7's warning-then-refusal rather than a silent mid-month lockout.
 
         Admin/ops-only, read included: a seat meter is the installation's commercial standing,
         and a rep reads their own seat elsewhere (UC-ADMIN-03 F1). Governed by the `license`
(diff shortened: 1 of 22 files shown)


lstrojny:
Heads-up on the inherited red: `migration-versions` fails on `core/0260`, which `main` has just taken for `scheduled_send_agent_provenance` while this stack's base carries `0260_license_rbac`. The file belongs to #1274, and that branch is being worked on right now, so the renumber belongs there rather than here — this branch picks it up on the next rebase. Nothing in this PR's own diff touches migrations.

lstrojny:
Rebased onto `main` now that #1274 has merged; the base of this PR is `main` and the diff is one commit.

Still red for a reason outside this diff, and it is worse than the earlier one: `main` itself carries **two** migrations at `core/0260` (`scheduled_send_agent_provenance` from #1357, `license_rbac` from #1274), and pgmigrate refuses the whole namespace on a duplicate, so nothing migrates anywhere. #1386 repairs it. With that commit applied locally, `make -C backend check` on this branch is green.

Issue #1367, opened 2026-08-15
Unit-authored message bodies inherit the statutory retention floor, with nothing bounding who may create them


tinnhlt-gra:
## What

`internal/modules/privacy/retention_floor.go` shields every activity whose `kind NOT IN ('task','note')` — which includes `message` — from retention expiry and from the Art. 17 person-erase cascade, for the jurisdiction's statutory period.

Before SP5 slice 2, an extension unit could not file `kind=message` at all: both write doors refused it outright. That change turns the blanket refusal into a bounded permission — a unit may file a message on a transport it DECLARED — which is the right bound for *whose transport*, and no bound at all on *how much*.

So third-party text can now become content the workspace's own retention policy and erasure cascade will not remove for the statutory period, and nothing gates the volume.

## Why it is not fixed here

It is a consequence of the feature working as designed rather than a defect in it: a captured chat message IS correspondence, and correspondence is what the floor exists to pin. A core connector (telegram) has exactly the same property today. What changed is only WHO can produce such content — the extension tier widened it from first-party connectors to any installed unit.

Worth deciding deliberately rather than inheriting:

- whether the floor should distinguish a core-captured message from a unit-captured one (probably not — the legal character is the same);
- whether an operator should be able to see, per unit, how much floor-pinned content it has produced;
- whether the answer is a bound at all, or simply disclosure.

Raised by the security review of SP5 slice 2 (ADR-0107/A158) as informational rather than as a defect in that change.

tinnhlt-gra:
Reviewed and deliberately not built — this is a decision rather than a defect, as the issue itself says. Recording a recommendation so whoever decides has one to react to.

**1. Should the floor distinguish core-captured from unit-captured messages?** **No.** The legal character of the content is identical, and the transport's supplier is not a fact the statute cares about. Splitting the floor by supplier would mean an operator's install choice silently changes how long correspondence survives, which is the opposite of what a retention policy is for.

**2. Should an operator see, per unit, how much floor-pinned content it has produced?** **Yes.** This is the honest answer to the actual risk. The operator installed the unit and can uninstall it, but today has no way to know what it has accumulated that they cannot delete. A per-unit count on the surface the unit's other facts already live on turns an invisible liability into a visible one.

[...]

Issue #1366, opened 2026-08-15
A unit's channel transport registers only when the capture registry is built, which is config-gated


tinnhlt-gra:
## What

`reconcileChannelProviders` — which since the SP5 slice-2 change also writes a `channel_provider` row for every transport a composed EXTENSION UNIT declares, and runs the core-vs-unit collision check — is called from `NewCaptureRegistry` (`internal/compose/capture.go`). The api role builds that registry only when a keyvault root key is configured.

So on a vault-less install:

- a unit's declared transport is never registered, and its captured messages fail `activity.channel_provider`'s foreign key as an unattributable 500;
- the collision check that refuses a unit shadowing a core transport never runs at all.

## Why it is not fixed in the change that introduced it

The coupling is real but currently self-limiting: a channel-supplying unit needs the vault for its own member secrets, so without one it cannot capture or send anyway. Fixing it properly means separating the unit half of the reconcile from core-capture construction, and that runs into an ordering question worth deciding deliberately rather than in passing — both halves publish into one in-memory "sendable" set (`activities.SetChannelProviders` / `comms.SetChannelProviders`), and a naive split lets whichever half runs second wipe the other's entries. On a role that composes no capture registry, publishing a union with an empty core half would deregister the core's own transport.

`LoadChannelProviderDirectory` in the same file already documents this exact class of defect for the READ side ("a defect this arc shipped once already") and was moved to run for every role that serves `/v1`. The write side deserves the same treatment.

## Suggested shape

Give the unit reconcile its own boot step next to `RegisterExtensions`/`LoadChannelProviderDirectory`, and hold the core and unit sendable halves separately so either order converges on their union rather than on the last writer.

Found while implementing SP5 slice 2 (ADR-0107/A158); reported by the craftsmanship review of that change.

tinnhlt-gra:
Reopening: the defect this describes is still live on `origin/main` (`0b7e41374`). The issue was auto-closed by the merge of #1368 — the PR it was *found in*, not one that fixed it.

Re-verified against `origin/main`:

- `reconcileChannelProviders` still has exactly **one** non-test caller — `backend/internal/compose/capture.go:163`, inside `NewCaptureRegistry`. No separate boot step was added next to `RegisterExtensions` / `LoadChannelProviderDirectory`; `server.go:362` calls only `loadChannelProviderDirectoryOrLog`, which *reads* the table.
[...]

PR #1368, opened 2026-08-15
feat(extensions): a unit supplies a real messaging transport, end to end

tinnhlt-gra:
## What

An extension unit can now supply a **real messaging transport**: declare a channel, capture a message on it, and have a rep's reply leave through it — on the member's own credential, through the product's ordinary reply path.

SP5 slice 2 (ADR-0107/A158). Slices 1a/1b/1c are merged (#1293, #1332, #1354); this completes the arc.

## The shape, because it is the decision worth reading

**A unit never sends. It DECLARES a transport and the core calls it.**

```
timeline reply box
  → activities.Store.SendMessage      (auth, consent, recipient from the conversation's links)
  → StageChannelTx
  → comms Dispatcher
  → commsResolver.ResolveChannel(member, provider)
  → the unit's Channel.Send
```

So the extension tier's outbound refusals stand untouched — a unit still may not spend the outbound cap from a tool or a job tick. What changed is that the human staged the message, the seat gate re-read them, and the core hands the unit a message to carry. That is the confirm-first staging the served surfaces could not do for themselves.

There is deliberately **no send verb and no send box** on the extension screen. `commschannel.go` already named this gap ("a unit-supplied provider's own resolve path… out of scope for this change"); this PR is that scope.

## What it took

- **`extension.Channel{Provider, Send, Live}`** on the published surface. `Live` is REQUIRED whenever `Send` is set — a transport that can transmit and cannot say whether it still may leaves the core guessing at the one moment guessing is unrecoverable. A nil `Send` is the documented capture-only case.
- **The reconcile registers unit transports** as their own registry rows (`transport='unit'`, `credential_model='per_member'`) and puts every sendable one into the set `SendMessage`'s pre-flight reads — without which a unit channel would register, publish, capture, and park every reply under "this installation cannot send on that".
- **The core-collision check lands in the reconcile**, where both sets exist. A unit shadowing a core transport fails the boot.
- **`ResolveChannel` branches** on whether a unit supplies the provider, and asks the unit's `Live` BEFORE handing over a message. A member who disconnected parks where a human can see it; a provider that could not answer is retried.
- **The blanket "a unit may not file a message" refusal becomes a bounded permission** — a unit may name a transport it DECLARED and no other, at both write doors. The other direction is refused too: a non-message naming a transport is a note the reply path would answer on.
- **dispact-connector** files `kind=message` on its own transport and binds the sender's account id.

## What the review round found — the value of the round

Three review passes (two Fable-side, one Codex) plus a parallel session reading the design against the branch. Each found something the gates could not see:

| Finding | Why it mattered |
|---|---|
| **A unit could seize `whatsapp`** | The collision check compared against composed *senders* (`{telegram}`), not the registry's reserved core rows. `whatsapp` is registered by migration with no Go connector behind it, so a unit declaring it passed, the upsert re-pointed the core row, and every hand-logged WhatsApp conversation became one the unit transmits. Fixed twice: the reserved set is read from `channel_provider WHERE transport='core'` inside the write transaction, AND the upsert carries `WHERE channel_provider.transport = EXCLUDED.transport` so the write itself cannot change a supplier. |
| **A unit could not report an unknown send outcome** | The channel seam marks a transmission in flight BEFORE the call and treats every error that is not `ErrSendOutcomeUnknown` as proof nothing went out. A POST whose answer was lost therefore delivered the rep's message **twice**. `extension.ErrSendOutcomeUnknown` now exists and is mapped. |
| **dispact bound a conversation id where a person belongs** | The core keys the person on `(provider, channel_user_id)`; that row is what an erasure suppresses. The same colleague DM-ing two members became two people, and an erasure covered one chat. Now binds the sender's account and resolves the conversation at send time. |
| **The request-time pre-flight refused every unit send** | `SendCapable` read `channel_connection` — the workspace-bot table a unit never writes. A rep would have been told at the composer to have an admin bind a bot with nothing to do with the transport. DESIGN-SP5 §8.1 warns about exactly this; the plan missed it. |

Plus two the published surface owed rather than the code: `Record.validateAddresses` now accepts an empty set whenever the counterparty names no address (§11.3 — a chat message has none anywhere in it), and the published `Counterparty` now refuses being named BOTH by an address and by a channel account, which the core has always refused and the surface let through as an unattributable failure.

## Proof

- `make check-backend` — exit 0
- `make craft-static` — 0 blocker / 0 major / 0 minor across all three roots
- `make test-integration` — OK, **0 skips**
- **New-code coverage: 89.4%** (449/502 changed statement lines), measured against the merge-base
- No prompt, corpus or tool-schema file is touched, so no aicert can be staled by this diff

## Filed rather than fixed

- #1366 — the unit reconcile rides the keyvault-gated capture registry
- #1367 — unit-authored message bodies inherit the statutory retention floor, with nothing bounding volume

## Follow-up already claimed

The Zalo OA connector is being built on this seam by a parallel session. `credential_model` derives from the unit's declaration in their follow-up, with a real second caller to justify it.

---
## Summary by cubic
Units can now declare a real messaging transport and carry replies over it on the member’s credential via the normal reply path. Previously, units could only capture; now a rep’s reply on a captured conversation can leave through the unit-declared provider, while tools/jobs still cannot send.

- Adds `extension.Channel{Provider, Send, Live}`; `Live` is required when `Send` is set; the manifest now includes `channels`.
- Reconcile registers unit transports in `channel_provider` (`transport='unit'`, `credential_model='per_member'`), includes sendable ones in send preflight, and blocks collisions with core transports.
- Channel resolution branches for unit providers and calls `Live` before send; disconnected members park; “could not answer” is retried; unknown send outcomes are mapped.
- Composer preflight asks the unit path first, avoiding `channel_connection` checks that don’t apply to unit transports.
- Units may name only transports they declared; non-message records naming a transport are refused.
- Published surface grows `ActivityKindMessage`, `ChannelProvider`, and `Counterparty.ChannelIdentity`; address/identity exclusivity is enforced.
- `extensions/dispact-connector` implements send + live checks, files `kind=message` on `dispact`, binds sender account, and reports unknown outcomes.

**Rollout notes**
- Units that declare `Channel.Send` must implement `Channel.Live`.
- Provider names use the `channel_provider` grammar (snake_case) and appear in the manifest’s `channels`.
- No behavior change to tools/jobs: outbound remains disallowed outside human-staged replies.

<sup>Written for commit 21d77ac0f45698ff7ad6afe1171b6ba5fa7460a8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1368?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/internal/compose/channelprovider.go
+++ b/backend/internal/compose/channelprovider.go
@@ -33,10 +33,14 @@ import (
 	"github.com/gradionhq/margince/backend/internal/platform/database"
 )
 
-// reconcileChannelProviders upserts a channel_provider row (transport='core')
-// for every composed provider, carrying the display facts the discovery
-// endpoint publishes, then sets both packages' in-memory snapshots to exactly
-// the composed set passed in.
+// reconcileChannelProviders upserts a channel_provider row for every transport
+// this binary composed — the core connectors passed in, plus every channel the
+// composed UNITS declare — carrying the display facts the discovery endpoint
+// publishes, then sets both packages' in-memory snapshots to the subset a reply
+// can actually leave on.
+//
+// It is also where a unit shadowing a core connector is refused, because this is
+// the first point at which both sets exist (unitChannelFacts).
 //
 // It runs over database.WithInfraTx, not the workspace-bound database.DB.Tx:
 // activity_kind and channel_provider carry no workspace_id, so binding a
@@ -55,35 +59,57 @@ import (
 // FK would refuse the delete anyway, and ErrConnectorNotConfigured already
 // parks a send against it rather than needing the row gone.
 func reconcileChannelProviders(ctx context.Context, pool *pgxpool.Pool, providers []string) error {
-	err := database.WithInfraTx(ctx, pool, func(tx pgx.Tx) error {
-		for _, facts := range channelProviderFactsFor(providers, providers) {
+	units, err := unitChannelFacts(providers)
+	if err != nil {
+		return err
+	}
+	rows := append(channelProviderFactsFor(providers, providers), units...)
+	if err := database.WithInfraTx(ctx, pool, func(tx pgx.Tx) error {
+		for _, facts := range rows {
 			// The display facts are UPSERTED, not left alone on conflict: they
 			// describe the composed connector, so the running binary is their
 			// only source of truth and a row written by an older build must be
 			// corrected rather than preserved. The provider itself still lands
 			// once — the primary key sees to that.
+			//
+			// `transport` is upserted with them, and it has to be: a provider
+			// that moves from a core connector to a unit (or back) is exactly
+			// the case where a stale value would tell the send path to resolve
+			// the wrong credential.
 			if _, err := tx.Exec(ctx, `
 				INSERT INTO channel_provider (provider, transport, label, credential_model, supplies_transport)
-				VALUES ($1, 'core', $2, $3, $4)
+				VALUES ($1, $2, $3, $4, $5)
 				ON CONFLICT (provider) DO UPDATE SET
+					transport         = EXCLUDED.transport,
 					label             = EXCLUDED.label,
 					credential_model  = EXCLUDED.credential_model,
 					supplies_transport = EXCLUDED.supplies_transport`,
-				facts.provider, facts.label, facts.credentialModel, facts.suppliesTransport); err != nil {
+				facts.provider, facts.transport, facts.label, facts.credentialModel, facts.suppliesTransport); err != nil {
 				return err
 			}
 		}
 		return nil
-	})
-	if err != nil {
+	}); err != nil {
 		return err
 	}
 	// These two take the COMPOSED set: they answer "can a reply leave this
 	// installation", which is a question about what was compiled in. What a
 	// message may NAME is a different set, loaded from the registry itself by
 	// LoadChannelProviderDirectory — see there for why it is not written here.
-	activities.SetChannelProviders(providers)
-	comms.SetChannelProviders(providers)
+	//
+	// A unit transport belongs in it exactly when the unit declared a Send. The
+	// send path's own pre-flight reads this set, so a unit channel left out of
+	// it would register, publish, capture — and park every reply a rep wrote
+	// under "this installation cannot send on that", which is the one failure a
+	// rep cannot tell from a broken provider.
+	sendable := slices.Clone(providers)
+	for _, unit := range units {
+		if unit.suppliesTransport {
+			sendable = append(sendable, unit.provider)
+		}
+	}
+	activities.SetChannelProviders(sendable)
+	comms.SetChannelProviders(sendable)
 	return nil
 }
 
--- a/backend/internal/compose/channelprovider.go
+++ b/backend/internal/compose/channelprovider.go
@@ -21,6 +21,7 @@ package compose
 
 import (
 	"context"
+	"fmt"
 	"log/slog"
 	"slices"
 	"sync"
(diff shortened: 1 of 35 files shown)


tinnhlt-gra:
## UAT — live, against the real provider

Run on an isolated dev stack (`DEV_SLUG=sp5axis`) with the `dispact-connector` unit connected to a live `workspace.gradion.com` account. Both halves of the arc exercised end to end.

**Message bodies and subjects are deliberately omitted throughout.** The poll backfilled real direct messages belonging to named colleagues out of a live workspace; every claim below is proven by structural columns, and none of it needs the message text. This repo is public.

### 1. The transport becomes a real registry row

```
 provider | transport |  label   | credential_model | supplies_transport
----------+-----------+----------+------------------+--------------------
 dispact  | unit      | Dispact  | per_member       | t
 telegram | core      | Telegram | workspace_bot    | t
 whatsapp | core      | WhatsApp | workspace_bot    | f
```

`transport='unit'` and `credential_model='per_member'` are what the send path branches on; `whatsapp` stays `core`/unsendable beside it, which is the row a unit was able to seize before the security fix in this PR.

Extension objects are fail-closed as designed: `connect` refused until `ext_dispact_connector_connection` was granted on the admin role.

### 2. Capture — a message, on its own transport

The scheduled poll ran on its own 120s cadence (`state=completed`, no forcing) and landed 39 activities across 4 conversations. The aggregate is unanimous:

```
  kind   | channel_provider | direction | count
---------+------------------+-----------+-------
 message | dispact          | inbound   |    39
```

**Zero `note` rows** — the pre-slice behaviour is gone, not merely supplemented. Three `person_channel_identity` rows were bound, each keyed on the sender's own account id (not the conversation id — the defect a review pass caught), and each links to an auto-created person.

### 3. The reply — resolved, gated, transmitted

Replying on a captured conversation via `POST /v1/activities/{id}/send-message`:

- The recipient was **resolved from the conversation**, never named by the caller — `activity → activity_link → person → person_channel_identity → channel_user_id`.
- The **consent gate ran and allowed on a derived basis**, and wrote its own proof:
  ```
  consent_qualifying_event: kind=inbound_message | source=derived
  ```
  Worth stating plainly: there is no explicit consent grant for this person. `business_correspondence` is Art 6(1)(f), and the gate derives the qualifying event from an inbound message — so **the capture this slice adds is what makes the reply lawful**, and the system wrote down why rather than assuming it.
- The outbound activity landed as `kind=message`, `channel_provider=dispact`, `direction=outbound`, on the anchor's own `thread_key`.
- The delivery **transmitted**:
  ```
  status | provider | attempts | provider_message_id
  -------+----------+----------+--------------------------------------
  sent   | dispact  |        1 | 84e5f402-eb92-4e3e-9d8e-0dba1a2653d5
  ```

One attempt, no park, and a provider message id returned by the live workspace. The message arrived.

### What this proves that the tests cannot

The unit and integration suites prove the seam's logic. This proves the **composition**: a third-party unit's `Send`, invoked by the core's own delivery dispatcher, under the sending member's sealed credential, against a provider nobody stubbed.

### Gates

| | |
|---|---|
| `make craft-static` | 0 blocker / 0 major / 0 minor, all three roots |
| `make test-integration` | OK, 0 skips |
| New-code coverage | **89.4%** (449/502 changed statement lines, measured against the merge-base) |

### Notes

- A screen recording of the walkthrough is being reviewed separately before publication, for the same disclosure reason the bodies are omitted here.
- The dev database used for this run holds real colleagues' messages and is being torn down; it is not part of the repo.
- `TestNoDealLandsOnAStageBeingRemoved` failed one local lane run at its own anti-vacuity guard. Reproduced on `origin/main` in a clean worktree, so it is pre-existing and machine-dependent, not from this change — filed as #1372. All 12 CI integration shards pass here.

Issue #1369, opened 2026-08-15
The share picker disables an already-granted subject, so an access level still cannot be changed in the UI


tinnhlt-gra:
## What happens

`frontend/src/screens/share.tsx` renders each roster candidate with `disabled={already}` where `already = grantedSubjectIds.has(candidate.id)`, and labels the row `share.alreadyGranted` ("already has a grant"). So once a colleague holds any grant on a record, the only way to change what they hold is to revoke it and share again.

## Why it is worth a ticket now

That workaround existed *because* the API refused a re-assert with a 409. gradionhq/margince-poc-v1#1297 fixed the API half: `POST /v1/record-grants` is now idempotent on `(record_type, record_id, subject_type, subject_id)` exactly as `crm.yaml` documents, so re-asserting a tuple upgrades or downgrades `access` and resets `expires_at` on the same row.

The backend can now do it and the UI still cannot ask for it.

## What it is not

Not a one-line flip of `disabled`. Picking an already-granted subject has to say something about what they hold today, and the interaction wants deciding before it is built:

- does the row show the current access, so the picker is not a blind overwrite?
- is a **downgrade** confirmed, given it silently reduces a colleague's authority? (The audit trail now records it as a `record_share` with a before-image, so it is legible after the fact either way.)
- does the existing per-row access control on the grants list become the edit affordance instead, making the picker's job "add", not "change"?

## Constraints already settled, so they are not re-derived

- The seat ceiling refuses a `write` grant to a read seat on the re-assert path exactly as on the first call (`TestAReAssertCannotWalkAWriteGrantPastTheSeatCeiling`), so the UI does not need to pre-filter for it — but it should render the 403 `seat_tier_insufficient` rather than a generic failure.
- gradionhq/margince-poc-v1#1247 is still open underneath this screen: `useRoster` reads one 200-user page.

luit-nfq:
Not attempted in the frontend sweep on #1739 — the interaction genuinely needs deciding first, and the issue already frames the three questions well (does the row show what the subject holds today, is a downgrade confirmed, or does the grants list become the edit affordance and the picker stay "add").

One note the sweep can add: #1247, which this issue names as sitting underneath the screen, is now confirmed to have **no client-side exit**. `/users` accepts no `ids` parameter, so the roster cap is an additive contract change. If the answer here turns out to be "the picker becomes search-driven", that is the same missing server-side read — worth deciding the two together rather than building a picker on top of one page of two hundred.

[...]

Issue #1341, opened 2026-08-15
TestNoDealLandsOnAStageBeingRemoved is flaky (~58%): the race can legally never run


dungbanguyen:
## Symptom

`TestNoDealLandsOnAStageBeingRemoved` (`backend/internal/compose/integration/stage_removal_race_integration_test.go`) fails roughly three times in five, on `main`, with:

```
stage_removal_race_integration_test.go:74: no stage was removed in any round — the race never ran
```

Measured locally, twelve consecutive runs of just that test against a real Postgres: **5 pass, 7 fail**. It also failed CI on an unrelated PR (#1336, whose diff touches neither deals nor stages), taking the whole `integration` gate red.

## Why it fails

Each round starts two goroutines on one deal — `AdvanceDeal` onto a target stage, and `ArchiveStage` of that same stage — and waits for both:

```go
go func() { _, advanceErr = e.Deals.AdvanceDeal(...) }()
go func() { removeErr = e.Deals.ArchiveStage(admin, targetID, nil) }()
wg.Wait()
assertRemovalOutcome(t, round, advanceErr, removeErr)
```

Per round, either outcome is legitimate and `assertRemovalOutcome` accepts both. The final assertion is the problem — it requires that **at least one round's removal won**:

```go
if n := e.WsCount(t, `SELECT count(*) FROM stage WHERE pipeline_id = $1 AND archived_at IS NOT NULL`,
    pipeline); n == 0 {
    t.Fatal("no stage was removed in any round — the race never ran")
}
```

Nothing makes the removal win any round. It is decided by goroutine scheduling and by which transaction reaches its lock first, so a run where the advance happens to win every round is a legal outcome of the code under test and a failure of the test.

The intent behind that assertion is right and worth keeping: without it a guard that refused every removal would satisfy the invariant while proving nothing. The mechanism is what does not hold.

## Suggested direction

The property is "when the removal wins, no live deal is left on a removed stage". That does not need to be discovered by racing:

- Drive at least one round **deterministically** — remove with no advance in flight — so the "a removal can win" half is proven rather than hoped for, and keep the racing rounds for the concurrency half.
- Or retry rounds until a removal wins, bounded, and report honestly if the bound is hit rather than asserting on the aggregate.

Either keeps the guard-is-not-vacuous check without making the suite's green depend on the scheduler.

## Notes

Landed with #1292. This is a test defect, not a product one: the invariant it guards held in every run above — only the "the race actually ran" precondition failed.

[...]

PR #1370, opened 2026-08-15
fix(identity): re-asserting a record grant updates it, as the contract says

### LINKED ISSUE #1297: createRecordGrant is documented as idempotent but answers 409 on re-assert

tinnhlt-gra:
`createRecordGrant` is documented **"Idempotent on `(record_type, record_id, subject_type, subject_id)` — re-asserting upgrades/downgrades `access` and resets `expires_at`"**, and its `201` is documented as *"Grant created (or updated)"*. The code ran a plain `INSERT` and mapped the unique violation to **409 — a status the operation does not declare at all**. So the documented way to change an access level did not exist, and the only route was revoke-then-re-grant.

Contract-first decides which half moves, and there is no spec work: `margince-foundation` `specs/contract/crm.yaml` carries the same sentence verbatim. This is code catching up to a ratified contract.

## What was NOT broken, which is most of the scope

`Idempotency-Key` is fully wired — `POST /v1/record-grants` is an entry in `compose/replayscope.go`'s `replayableOperations`, derived from the contract by `TestIdempotentOperationsMirrorTheContract`. Two mechanisms share the word "idempotent" here and only one was broken:

- **`Idempotency-Key`** — "this is a retry of the same request" → replay the recorded answer for 24h. **Worked.**
- **Idempotent on the natural key** — "re-asserting this tuple with a different access is an upsert" → **missing.**

A caller changing an access level sends a *new* request, not a retry, so the replay gate correctly passed it through — into the 409.

## The upsert

Follows `deals/fxrate_store.go`, the tree's worked audited upsert, rather than inventing a second shape: read the displaced row, `ON CONFLICT DO UPDATE`, audit with both images.

| column | on conflict | why |
|---|---|---|
| `access` | `EXCLUDED` | the contract's "upgrades/downgrades" |
| `expires_at` | `EXCLUDED` | the contract says a re-assert **resets** it. `COALESCE` would make an expiry unclearable — mutation-tested |
| `reason` | `EXCLUDED` | belongs to the assertion in force, not the first one ever made |
| `granted_by` | `EXCLUDED` | the row answers "who is accountable for this access". Keeping the original grantor would attribute Bob's upgrade to Alice |
| `version` | `+ 1` | the column has existed on the table since `0011` for exactly this |
| `id`, `created_at` | **untouched** | a re-assert is the same sharing relationship — which is what makes it idempotent rather than a duplicate under a new name |

The audit verb stays `record_share` in both directions, downgrades included: the contract pins it and `record_unshare` is revocation's word. Which way the access moved is the **image pair's** job, so the image widened to carry `expires_at` and `reason` — an image that omits a field the upsert can change reports "nothing moved" when something did. `privacy/recordhistory.go` already serves these images on the record timeline, so a downgrade now reads as one there.

## What three reviews changed, because it is most of this PR now

The first commit was the idempotency fix. Two independent reviews and an
adversarial pass then found the same shape from different angles, and the
second commit is the answer to it.

**The gate that admits a re-assert is `EnsureLinkTarget`, and the grant arm
SATISFIES it.** So the person a record was shared with passes it — and the
upsert let them administer the very share that admitted them. Reproduced
against a running installation, all three:

| attack | before this PR | with the upsert alone |
|---|---|---|
| clear the `expires_at` that time-boxed the share | 409 | **201, share is permanent** |
| widen one's own `read` share to `write` | 409 | **201** |
| pass the record to a colleague the sharer never named | allowed (pre-existing) | allowed |

The **time** one is the sharpest and I had not considered it: a one-hour share,
re-asserted by its beneficiary with no `expires_at`, comes back permanent. The
409 was load-bearing by accident, and the upsert removes it.

**The rule was already in the contract and nothing made it operative** —
*"flat explicit grants only … no grant-of-grant delegation."* `auth.EnsureCanShare`
makes it operative: administering a record's grants requires that record in the
caller's **own** scope, so the grant arm that widens every read path is
deliberately not consulted.

**It is deliberately NOT written as access-level intersection**, which was my
first fix and is weaker. `record_grant.access` is consulted by no authorization
path in this tree — a `read` grant confers write, filed as
[#1373](https://github.com/gradionhq/margince-poc-v1/issues/1373) — so a rule
resting on that distinction refuses `write` and waves the identical authority
through as `read`. The redteam demonstrated exactly that against my first
attempt. Scope is real today; the access distinction is not yet, and the rule
that shipped does not depend on it.

Three more findings, all real:

- **The concurrency comment was false.** `FOR UPDATE` cannot serialize two
  concurrent *creates* — an absent row locks nothing — so the loser audited a
  first share over a row it displaced. Now `storekit.LockWriteIdentity` before
  the precondition read, which is what `fxrate_store.go` takes for the same
  reason. An intermediate revision of mine "solved" this with a version oracle
  that answered **409**, on the one endpoint whose premise is that it declares
  none; that is gone.
- **The audit image omitted `granted_by`** — under a comment claiming it omitted
  nothing the upsert can change. `RevokeRecordGrant` hand-built a third shape;
  both now render through `grantImage`.
- **The seam fix falsified two more sibling comments** the first pass missed
  (`capture/freemaildomain.go`, `identity/onboarding.go`).

## The security question this turns on

**A re-assert must not become a second door onto an authority the first door refuses.** The unique constraint used to refuse a second call outright, so `refuseWriteGrantToReadSeat` (AAD-AC-4) only ever ran on the create path. The upsert turns the second call into a real write, so the ceiling's job got wider.

`TestAReAssertCannotWalkAWriteGrantPastTheSeatCeiling` shares `read` with a read seat (allowed — it is the authority the licence already carries), re-asserts the same tuple as `write`, and requires **403 `seat_tier_insufficient` *and* the stored access still reading `read`**. `TestAShareIsNotALicenceToReShare` drives a real team-scoped principal through all three attacks above and reads the row back after each refusal, with an allow arm (the owner may still administer the share) so it cannot be satisfied by a probe that refuses everyone.

**Everything is mutation-tested, and two mutants corrected my own reasoning.**
Moving the seat guard *behind* the write still passes — the transaction rolls it
back, so ordering is not what protects it, and I rewrote a comment that claimed
it was. Replacing `OwnerPredicate` with `VisiblePredicate` in `EnsureCanShare`
(my weaker first fix) fails all three attack arms, which is the proof that the
rule is *own scope* rather than visibility.

## Also: the audit seam, fixed at the seam

Writing the before-image exposed a live defect one layer down. `storekit.marshalOrNil` wrote SQL NULL only for an **untyped** nil, so a caller that builds its image in a `map[string]any` and leaves it nil hands it an interface carrying a **typed** nil — which marshals to JSON `null` and is invisible to every `WHERE before IS NULL` query in the tree.

This has now cost three call sites:

- `capture/owndomainstore.go` hit it, and worked around it **locally** with `var beforeImage any` and a comment naming the trap.
- `deals/fxrate_store.go` has carried it **live** ever since — `replacedFxRate` returns a nil `map[string]any`.
- This change reproduced it from scratch, a module away, and its own test caught it.

Per the repo's rule 1 (fix the invariant, not the call site) the nil check moved into the seam, which fixes all three and the next one. `TestAnAbsentImageIsSQLNullWhateverKindOfNilCarriesIt` holds it across every nilable kind, with `TestAnEmptyImageIsStillAnImage` as the mirror so the rule cannot be satisfied by returning nil for everything falsy. capture's local workaround and its now-stale comment are retired.

## Deliberately out of scope, so it is not read as an oversight

**The SPA still cannot reach this.** `share.tsx` **disables** an already-granted subject in the picker and labels it `share.alreadyGranted`, so there is no UI path to change an access level — the workaround *is* the symptom the issue describes. Making it reachable is a real interaction decision (does the row show its current access? is a downgrade confirmed?) and belongs in its own PR. Filed as #1369.

## Verification

- `make check` (root, backend + frontend) — **green**
- `make test-integration` — run in full twice, because this changes two platform seams. First run (before the auth work): **`OK: integration passed with 0 skips (40 packages, parallel)`**. Second run: one failure, `TestNoDealLandsOnAStageBeingRemoved`, which is **[#1341](https://github.com/gradionhq/margince-poc-v1/issues/1341)** — a documented ~58% flake on `main` with the identical message (*"no stage was removed in any round — the race never ran"*), reproduced here as 1 fail / 2 pass in three consecutive runs of that test alone. Nothing else red
- `craft static --strict` on every changed file — 0 findings
- **UAT against a live stack** (`DEV_SLUG=grant`) — video below. Six steps: create, share `read` with an expiry, re-assert as `write` with none → **201, same grant id, same `created_at`, access upgraded, expiry reset**; one row not two; audit trail reading `(none) -> read` then `read -> write`; and the read-seat re-assert refused with no residue.
- **No AI-certification impact**: `createRecordGrant` is `x-agent-access: human-only` with no `x-mcp-tool`, no prompt/corpus/fixture is touched, and `make check` (which fails on a stale aicert record) is green.

## Upstream

**Contract-first: [margince-foundation#1313](https://github.com/gradionhq/margince-foundation/pull/1313) carries the spec half** and should land first. It pins what the old sentence left open — that a re-assert *restates* the grant (so `reason` and `granted_by` move too, and `id`/`created_at` do not), and that administering a share requires own-scope. The `crm.yaml` change here mirrors that text.

Closes #1297

### DIFF
--- a/backend/api/crm.yaml
+++ b/backend/api/crm.yaml
@@ -8169,8 +8169,14 @@ paths:
         Grants `read` or `write` on a single record to a user or team otherwise out of their own/team/all
         scope. Enforced by widening BOTH the application visibility WHERE clause AND the RLS backstop
         policy (`data-model §1.3c/§2.5`). Idempotent on `(record_type, record_id, subject_type, subject_id)` —
-        re-asserting upgrades/downgrades `access` and resets `expires_at`. A grant can never exceed the
-        granting principal's own access (scope-intersection). Audited (`action: record_share`). Bounded:
+        a re-assert RESTATES the grant rather than amending it: `access`, `expires_at`, `reason` and
+        `granted_by` all take the values of the new request, so a field the caller omits is cleared and
+        accountability moves to the re-asserting principal. `id` and `created_at` do not move — it is the
+        same share, not a new one. A grant can never exceed the granting principal's own access
+        (scope-intersection), and a grant is not itself a licence to share: administering a record's
+        grants requires that record in the caller's OWN scope (owner/team/all), because a caller whose
+        only claim to it is a share would otherwise be exercising the grant-of-grant delegation this
+        operation excludes. Audited (`action: record_share`). Bounded:
         flat explicit grants only — no sharing hierarchies, criteria-rules, or grant-of-grant delegation.
 
         The seat ceiling binds the RECIPIENT as well (AAD-AC-4): a `write` grant to a user on a `read`
--- a/backend/internal/compose/integration/grants_integration_test.go
+++ b/backend/internal/compose/integration/grants_integration_test.go
@@ -14,6 +14,7 @@ import (
 	"context"
 	"net/http"
 	"testing"
+	"time"
 
 	"github.com/jackc/pgx/v5"
 
@@ -97,14 +98,6 @@ func TestRecordGrantHTTPLifecycle(t *testing.T) {
 	}, nil, &grant); status != http.StatusCreated {
 		t.Fatalf("create grant → %d", status)
 	}
-	// Duplicate share → 409.
-	if status := e.Call(t, "POST", "/v1/record-grants", apptest.AnyMap{
-		"record_type": "person", "record_id": e.personID,
-		"subject_type": "user", "subject_id": admin.User.ID,
-		"access": "read",
-	}, nil, nil); status != http.StatusConflict {
-		t.Fatalf("duplicate grant → %d, want 409", status)
-	}
 
 	var listed struct {
 		Data []struct {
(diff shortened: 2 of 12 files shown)


tinnhlt-gra:
## UAT — against a live stack, not a test harness

Recorded against `DEV_SLUG=grant` (api `:19009`, db `margince_dev_grant`), driving the real HTTP surface as an authenticated human. Script kept at `.tmp/record-grant-upsert/uat.sh`.

**On the instrument:** MCP Inspector does not apply here — `createRecordGrant` is `x-agent-access: human-only` and carries no `x-mcp-tool`, so there is no agent surface to inspect. The honest instrument is an authenticated REST session, which is what this is.

https://github.com/user-attachments/assets/e7472353-4cff-4368-8f8a-986c9b75fadc

What the recording shows, in order:

**1.** Share `read` with a reason and an expiry → `201`.

**2.** Re-assert the same tuple as `write`, supplying no expiry and no reason → **`201`, where this used to be `409`**. Same `id`, same `created_at`, `access` upgraded, `reason` and `expires_at` reset to null. That expiry reset is the half a `COALESCE`-shaped upsert would silently refuse.

**3.** One row, not two — `GET /record-grants` returns a single grant.

**4.** The audit trail says what moved: `record_share: (none) -> read`, then `record_share: read -> write`. The first image is SQL NULL, which is the `marshalOrNil` half of this PR — without it that row stores JSON `null` and is invisible to `WHERE before IS NULL`.

**5.** The seat ceiling still refuses on the SECOND call: a read-seat colleague accepts a `read` grant (the authority their licence already carries), then the same tuple re-asserted as `write` is refused `seat_tier_insufficient` — and the stored access is read back as `read`, so the refusal left no residue.

Gates alongside it: `make check` green, `make test-integration` green in full (`OK: integration passed with 0 skips (40 packages, parallel)`) because this touches a platform seam ~40 audit writers share, and `craft static --strict` clean on every changed file.

tinnhlt-gra:
## UAT, re-recorded against the reviewed code

The first recording was against the idempotency commit only. This one runs against the merged branch — same live slug stack (`DEV_SLUG=grant`, api `:19009`), rebuilt after the auth change, with a seventh step for the new rule.

https://github.com/user-attachments/assets/db4b2709-679f-44ac-95d1-17f25bc45a85

Steps 1–6 are unchanged and still pass: the re-assert returns **201 where it used to return 409**, on the same `id` and `created_at`, with `access` upgraded and `expires_at` reset; one row, not two; the audit trail reading `(none) -> read` then `read -> write`; and the seat ceiling refusing on the second call with no residue.

**Step 7 is honest about what a UAT can and cannot show here.** The deny arm of `EnsureCanShare` needs a **team-scoped** caller, and the dev session is the bootstrap admin — an unbounded principal short-circuits the probe by design, so no amount of clicking from this session can exercise the refusal. That arm is proven in the lane by `TestAShareIsNotALicenceToReShare`, which drives a real team-scoped principal through all three attacks and reads the row back after each one. What the recording *can* show is the allow arm, and does: the owner re-boxes an expiry that the share's beneficiary is now refused.

That split is the point rather than a gap — a UAT that appeared to prove the refusal from an admin session would be proving nothing, because the admin is exactly the caller the rule waves through.

PR #1377, opened 2026-08-15
build(gates): the module lint says which of its three failures happened (#1347)

### LINKED ISSUE #1347: make check is broken on macOS: check-lint-modules.sh uses a GNU-only sed extension and reports the failed cd as lint findings
Filed while rebasing #1345 onto `484f30a49`. Not a defect in that branch — reproduced on a clean checkout of `main`.

## What happens

`make check` fails on macOS at the `lint-modules` gate added by #1269:

```
./scripts/check-lint-modules.sh: line 105: cd: backend/go.mod: Not a directory
./scripts/check-lint-modules.sh: line 105: cd: backend/tools/go.mod: Not a directory
... (all 8 modules)

FAIL — golangci-lint findings in: backend/go.mod backend/tools/go.mod cli/craft/go.mod
composition/go.mod extensions/de/go.mod extensions/dispact-connector/go.mod
extensions/notes/go.mod extensions/yogi/go.mod
```

Every gate before it passes; `make -C backend check` and `make check-fe` both pass on their own.

## Cause

`scripts/check-lint-modules.sh:70` strips the filename with

```sh
modules="$(git ls-files "*go.mod" | sed "s#/\?go\.mod\$##" | ...)"
```

`\?` (optional-previous-atom) is a **GNU sed extension**. BSD sed — macOS `/usr/bin/sed` — reads it as a literal `?`, so the pattern never matches and nothing is stripped:

```
$ echo backend/go.mod | sed "s#/\?go\.mod\$##"
backend/go.mod            # BSD: unchanged
$ echo backend/go.mod | sed "s#/\{0,1\}go\.mod\$##"
backend                   # POSIX form, correct on both
```

`$mod` therefore stays a file path, `cd "$mod"` fails, and the non-zero exit is collected into `failed`.

## Why it is worth more than a one-line fix

**The failure reports itself as something it is not.** A failed `cd` is announced as *"golangci-lint findings in: backend/go.mod …"*, so the first read is "my branch has lint findings in eight modules" — and the follow-on text explains how to fix findings that do not exist. The gate cannot currently distinguish "this module has findings" from "I could not enter this module", and those need different answers.

Linux CI is unaffected (GNU sed), so this reads as green upstream and red only on a developer machine — the direction that erodes trust in the local gate fastest.

## Suggested exits

- The stripping is portable as `sed "s#/\{0,1\}go\.mod\$##"`, or by dropping sed for `dirname`.
- Separately: have the loop distinguish "could not enter" from "found issues", so a broken invocation can never again be reported as a finding. A gate that mislabels its own failure is the shape this repo already fixed once in the migration-collision work.
- Worth a portability sweep of `scripts/*.sh` for other GNU-only constructs while someone is in there.

tinnhlt-gra:
Closes #1347.

## What was already fixed, and what was not

The headline — `make check` broken on macOS — is **already resolved on `main`**: the GNU-only `\?` in the module-list sed landed as two portable substitutions in #1359, and `make check-backend` is green here on BSD sed. I verified that before touching anything.

What that fix did not touch is the half #1347 calls out as *"worth more than a one-line fix"*: the loop could not tell its own broken invocation from the thing it reports.

```sh
if ! (cd "$mod" && golangci-lint run …); then failed="$failed $mod"; fi
```

`cd` failing exits **1**. golangci finding issues exits **1**. Both landed in one bucket, printed as *"golangci-lint findings in: …"* and followed by advice on how to fix findings that did not exist. That is how a path bug came to report eight modules' worth of lint findings in modules nothing had ever entered.

## Three outcomes, kept apart

| Class | How it is detected | What the message says |
|---|---|---|
| **Could not enter** | `cd` exits a reserved status (90), well outside golangci's range | the directory is missing from the working tree, **nothing was linted**, check the checkout — not the code |
| **Could not complete** | golangci exits non-1 | the tool refused to run (unreadable config, unresolvable workspace, internal error); its own error is above |
| **Findings** | golangci exits 1, having run | the original message, unchanged |

The `cd` gets its own status rather than a pre-flight `[[ -d ]]` check, so the classification cannot drift from the thing it classifies: there is no window between testing the directory and entering it.

## All three induced and observed

Hiding one module (`mv extensions/notes …`) produces both of the non-finding classes at once, which is the clearest possible demonstration:

```
FAIL: lint-modules could not enter: extensions/notes
  Those directories come from tracked go.mod files that the working tree does
  not have, so NOTHING was linted in them. This is not a lint finding…

FAIL: golangci-lint could not complete in: backend/tools(exit 3) cli/craft(exit 3)
  extensions/de(exit 3) extensions/dispact-connector(exit 3) extensions/yogi(exit 3)
  A non-1 exit is the tool refusing to run…
```

**Six modules the old code would have called lint findings, none of which were linted at all.** Injecting a real violation into `extensions/notes` still reports `FAIL — golangci-lint findings in: extensions/notes`, unchanged. Clean tree: `OK: lint-modules — 7 module(s) outside backend/ lint clean`.

No automated guard: there is no shell-test harness in the tree (33 scripts, none tested, no shellcheck gate), so adding one is its own change rather than a rider on this. Stated plainly rather than implied — the classification above is proven by induced runs, not by a test that will re-run tomorrow.

## The portability sweep the issue also asks for

Done, and it found nothing. Across `scripts/*.sh` and `.githooks/`: no `\?`, `\+` or `\|` in a BRE, no `grep -P`, no `readlink -f`, no `date -d`, no GNU-form `sed -i`, no `stat -c`, `find -printf`, `tac`, `xargs -r`, `head -n -N` or `seq -w`. The one remaining `\+` is inside a `sed -E` expression matching `// +build`, where an escaped plus is a literal in both dialects.

Also tightened the surviving comment on that sed line: it narrated the past defect ("on a developer laptop every path kept its suffix… the gate reported findings in eight modules"), which is history git already holds. It now states the invariant instead.

## Gates

`make check-backend` green locally on macOS — which is the thing #1347 says was broken.

## Found but not fixed

While inducing the failure classes, the gate reported findings in `../../margince-next-erase/backend/tools/…` — a **deleted** sibling worktree — because golangci's cache keys results to absolute paths that no longer exist. It reads exactly like "your branch has findings in eight modules", is cleared by `golangci-lint cache clean`, and cost time twice today. That is a fourth way this gate's output can mislead, owned by the tool rather than by this script; filed separately rather than papered over here.

---
## Summary by cubic
Makes the module lint gate report which failure happened instead of lumping all errors under “golangci-lint findings.” Previously, a failed cd or tool error appeared as findings; now each outcome is distinct, preventing misleading failures and clarifying remedies.

- Distinguishes three outcomes: could not enter (cd fails), could not complete (golangci-lint exits non‑1), and findings (golangci-lint exits 1).
- Reserves exit 90 for cd failures and reads the actual lint exit code; no preflight dir check, so classification cannot drift.
- Prints a separate list and guidance for each outcome; still exits 1 if any list is non‑empty.
- No caller changes. `make check-backend` stays green on macOS.

<sup>Written for commit f59046a51e17e835246737b4744c3f7269662226. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1377?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/scripts/check-lint-modules.sh
+++ b/scripts/check-lint-modules.sh
@@ -67,11 +67,11 @@ fi
 # a type-checkable package; it has nothing to say here. They are NOT unchecked
 # code: the craft gate, the license header test and the tree-wide gofmt gate all
 # cover fixtures/, and none of the three needs to typecheck to do its job.
-# `s#/\?go\.mod$##` reads as an optional slash under GNU sed and as a LITERAL
-# backslash-slash under the BSD sed macOS ships, so on a developer laptop every
-# path kept its `/go.mod` suffix, `cd` failed with "Not a directory", and the
-# gate reported findings in eight modules it had never entered. Two plain
-# substitutions say the same thing in both dialects.
+
+# Two plain substitutions rather than one optional-slash pattern: `\?` is a GNU
+# extension, and the BSD sed macOS ships reads it as a literal `?`, so a pattern
+# that strips the suffix on CI leaves it in place on a laptop. Both dialects
+# agree on these two.
 modules="$(git ls-files '*go.mod' \
   | sed -e 's#/go\.mod$##' -e 's#^go\.mod$#.#' \
   | grep -v '^backend$' \
@@ -93,7 +93,22 @@ fi
 # looks like it passed and did not run.
 members="$(sed -n 's#^[[:space:]]*\.\./\.\./##p' "$COMPOSED_WORK")"
 
-failed=""
+# THREE ways this gate can fail, kept apart, because they are three different
+# questions for whoever reads the output and only one of them is about the code.
+#
+# golangci exits 1 when it RAN and found issues, and non-1 when it could not run
+# at all — an unreadable config, a workspace that does not resolve, an internal
+# error. A failing `cd` is a third thing again, and a bare `cd "$mod" && …` exits
+# 1 for that too, which is how one path bug came to announce itself as findings
+# in eight modules nothing had ever entered, under a remedy explaining how to fix
+# findings that did not exist. So the cd gets a reserved status of its own, well
+# outside the range golangci uses, and the run's exit is read rather than merely
+# tested.
+readonly CANNOT_ENTER=90
+
+findings=""
+unenterable=""
+broken=""
 count=0
 for mod in $modules; do
   count=$((count + 1))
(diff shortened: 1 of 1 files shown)


tinnhlt-gra:
@coderabbitai review

Issue #1382, opened 2026-08-15
A channel-captured person is minted with no email, so the same human becomes a second contact


tinnhlt-gra:
## What

A person created from a captured channel message carries **no email address**, even when the connector knows one. Because the CRM's dedupe spine matches on email, that person cannot be recognised as an existing mail-captured contact — so one human ends up on two records.

Observed on a live UAT run of `dispact-connector` (#1368):

```
    full_name     | emails | provider
------------------+--------+----------
 Dung Nguyen      |      0 | dispact
 Tuyen Dinh Quang |      0 | dispact
 Luu Nguyen Thanh |      0 | dispact
```

The connector had every one of those addresses: `recordFor` refuses a sender with no address outright (`extensions/dispact-connector/record.go`), and the provider's `/api/users/batch` returns `email` for each. The address is dropped on purpose, and that is where the chain starts.

## The chain, because no single link is wrong

1. `connector.Counterparty` is named by an address **or** by a channel identity, never both — `counterpartyShapeOf` returns `shapeAmbiguous` and `admitCounterpartyShape` refuses it (`capture/sinkchannel.go`). The published `extension.Counterparty` restates the same rule.
2. So a unit filing a channel message must drop the address it holds. dispact does (`counterpartyOf`).
3. So `EnsureChannelCounterpartyInput` never receives one.
4. So `resolveChannelPerson` calls `DedupePerson` with `PersonCandidate{FullName, ChannelIdentities}` and **no `Emails`** (`people/ensurechannel.go`).
5. So the exact-email tier never runs, and a contact who already exists from mail capture is not matched.

`PersonCandidate.Emails` already exists and is documented as "checked in full against the exact tier" — the channel path simply never populates it.

## Why this matters more now

Before the extension tier, the only channel connector was Telegram, where a bot genuinely has no address for the sender — dropping an address you never had costs nothing. A unit-supplied transport can be different: `dispact` is a workspace product whose API returns the member's corporate email, and the same is likely true of any transport built on a business platform. So the rule now discards evidence the system actually has.

The Duplicates surface is the designed safety net, and it works — but it turns every colleague who both emails and DMs into a manual merge, and the merge is only possible if a human notices.

## What this is NOT

Not a request to relax the counterparty rule. The exclusivity is load-bearing: an address and a channel account resolve through different ladders (the mail ladder defers a first-time corporate address to the pending inbox; the channel ladder mints and binds), and a record naming both is genuinely asking which one it meant.

[...]

PR #1383, opened 2026-08-15
fix(frontend): a captured channel conversation can be answered from the UI

tinnhlt-gra:
## What

#1368 shipped the whole channel send path and proved it against a live provider. **It shipped with no way for a rep to reach it.**

The composer was there and correctly wired — `TimelineActions` renders Reply for a `kind === "message"` row and posts `/activities/{id}/send-message`. But it is mounted on the **deal**, **company** and **person-list** timelines, and a captured channel message links to a **person**, whose only routed surface is the contact page — whose conversation-memory card mounted no actions at all. So for exactly the conversations that slice creates, the button existed nowhere. The UAT sent through the API and never noticed.

Found by the user opening the contact and asking where the send button was.

## Two ways in

**Reply, on the row.** The memory card now carries the reply — a deliberate exception to "summary, not ledger", and *only* the reply: Relink is a raw-ledger act and stays on the ledger. Extracted as `ChannelReplyAction` so the two surfaces that offer a reply cannot come to offer different things.

**The header composer, made general — and adaptive.** It lists every way this person can be reached and asks only when there is a choice:

| Contact | Composer |
|---|---|
| Address only | unchanged — no picker |
| Address + a Dispact conversation | picker: Email, Dispact |
| No address, one conversation | opens on that channel, no picker |

A picker with one option is a control that cannot be used, so it doesn't render. The case that motivated it: a contact with **no address at all**, where the header offered mail to somebody with no mailbox while a live conversation sat below.

## What a channel transport can and cannot do, stated rather than implied

It **continues** a conversation and never **opens** one. `send-message` resolves the recipient from the conversation being answered, and there is deliberately no account-started twin (`sendAccountEmail` exists; `sendAccountMessage` does not). So a transport is offered only when the person already has a conversation on it **and** the identity is still reachable — two different questions, and a transport with one and not the other would be a choice that fails at the send. Both withholding cases are tested.

Mail's subject, bcc, send-later and AI draft are withheld on a channel, each for its own reason: no subject line, `send-message` takes no schedule, and the model lane drafts mail only. Rendering them disabled would promise a capability that is coming.

## Structure

`PersonComposer` was at the complexity ceiling before this and over it after, so the pieces came out as their own units rather than as a lint suppression: `useTransportChoice`, `TransportPicker`, `ToLine`, `MailOnlyFields`, `canSendOn`, `sendChannelReply`.

## Proof

- `make frontend-check` — exit 0, **2692 tests**
- `make check-backend` — exit 0
- `make craft-static` — 0/0/0 across all three roots
- Five new tests pin the adaptive rule and both withholding cases

## Related

- Follows #1368 (SP5 slice 2)
- **#1382** — filed while testing this: a channel-captured person is minted with no email, so the same human becomes a second contact. The connector *has* the address and the counterparty rule forces it to drop it. Spec discrepancy, raised not worked around.

---
## Summary by cubic
Lets reps reply to captured channel conversations from the contact page and fixes the person timeline to include each message’s `channel_provider`. Previously the send path existed but was unreachable in the UI and timeline rows reported a null provider, so channels could not be named or selected.

- Reply on memory card: adds Reply on message rows using the same gate as timeline actions; only Reply is shown (Relink stays on the ledger). Shared via ChannelReplyAction to keep behavior consistent across surfaces.
- Header composer: offers Email and any reachable channel that has a conversation anchor; Email leads; shows a transport picker only when there’s a real choice. With a single available transport, it opens directly on that one.
- Channel semantics: continues an existing conversation only (anchors on the activity); offered only when the identity is reachable and a conversation exists; hides mail-only fields (subject, bcc, send‑later, model draft); To is read‑only and describes the continued conversation. Switching person/transport clears drafted content. Send gating moved to canSendOn.
- Backend: person360 timeline now selects/scans `channel_provider`; adds an integration test asserting the provider round‑trips so the memory card can label transports and the composer can offer them.
- Implementation/tests/i18n: extracted useTransportChoice, TransportPicker, ToLine, MailOnlyFields, sendChannelReply; added `persondrawers.transport.test.tsx` covering adaptive choice and withholding cases; added i18n strings (en/de/vi); memory layout adds a reply column without reserving space when absent.

<sup>Written for commit ed7cdad9ce7c405ebac9bb8b1fb2b4375534ff3a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1383?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/internal/compose/integration/person_relationship_room_integration_test.go
+++ b/backend/internal/compose/integration/person_relationship_room_integration_test.go
@@ -580,3 +580,52 @@ func TestMergeCarriesTheEnrichmentSidecarToTheSurvivor(t *testing.T) {
 		t.Errorf("%d row(s) stayed on the merged-away person, invisible to every read", left)
 	}
 }
+
+// A captured channel message reaches the page NAMING THE TRANSPORT THAT CARRIED
+// IT, and this test exists because it did not.
+//
+// Since ADR-0107/A158 the kind says only that an interaction was a message —
+// which transport carried it is `channel_provider`, a separate column. The 360's
+// timeline read is a hand-written sibling of activities.activityColumns, and
+// when the narrowing added the column there, nothing pointed at the copy. So
+// every activity on every person page reported a null provider: the memory card
+// rendered "message" where it should have said the transport's name, and the
+// composer could not tell a contact reachable on a unit's channel from one
+// reachable nowhere.
+//
+// The guard is a ROW-TO-PAYLOAD assertion rather than a source grep, because
+// what regressed was a SELECT list — a grep for the column name would have
+// passed on a query that selected it into a variable nobody scanned.
+func TestThePerson360TimelineNamesTheTransportThatCarriedAMessage(t *testing.T) {
+	e := Setup(t)
+	owner := OwnerConn(t)
+	mine := e.SeedPerson(t, "Luu Nguyen Thanh", &e.Rep1)
+
+	// telegram, because it is registered by the core migration on every
+	// installation — the FK on activity.channel_provider means an unregistered
+	// name could not be seeded at all, and this test is about the read.
+	message := SeedRow(t, owner, `INSERT INTO activity
+		(id, workspace_id, kind, channel_provider, body, occurred_at, direction, source, captured_by)
+		VALUES ($1, $2, 'message', 'telegram', 'they wrote', '2026-08-01T09:00:00Z',
+		        'inbound', 'manual', 'human:x')`, e.WS)
+	LinkActivity(t, owner, e.WS, message, "person", mine)
+
+	rep := e.As(e.Rep1, []ids.UUID{e.Team1}, roomPerms)
+	page, err := personRoomService(e).Assemble(rep, ids.From[ids.PersonKind](mine))
+	if err != nil {
+		t.Fatalf("Assemble: %v", err)
+	}
+	if page.Activities == nil || len(page.Activities.Data) != 1 {
+		t.Fatalf("the timeline holds %v rows, want the one captured message", page.Activities)
+	}
+	got := page.Activities.Data[0]
+	if got.Kind != "message" {
+		t.Errorf("kind = %q, want message", got.Kind)
+	}
+	if got.ChannelProvider == nil {
+		t.Fatal("the message reached the page with no transport; the timeline renders it as the bare word \"message\" and every transport looks alike")
+	}
+	if string(*got.ChannelProvider) != "telegram" {
+		t.Errorf("channel_provider = %q, want telegram", string(*got.ChannelProvider))
+	}
+}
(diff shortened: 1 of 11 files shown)


tinnhlt-gra:
## UAT — the reply, from the UI, against the live provider

Run on an isolated dev stack (`DEV_SLUG=sp5axis`) with `dispact-connector` connected to a live workspace account. Message bodies are omitted for the same reason as #1368: the captured rows are real colleagues' direct messages and this repo is public.

### It found a second bug before it proved anything

Pointing the new composer at a real contact, the transport picker resolved **no channel** for somebody with a live Dispact conversation. The payload it reads had already dropped the provider:

```
GET /v1/people/{id}/360  →  activities[].channel_provider: null
```

…while the row itself held `channel_provider = 'dispact'`.

`person360`'s timeline read is a hand-written `SELECT` beside `activities.activityColumns`, and when ADR-0107/A158's narrowing added the column to the shared one, nothing pointed at the copy. **The column has been missing from every person page since 1b.** That is repo rule 1 — fix the invariant, not the call site — landing on the exact failure it describes.

Fixed in `ed7cdad`, with a row-to-payload guard rather than a source grep: what regressed was a SELECT list, and a grep for the column name would pass on a query that selects it into a variable nobody scans. Mutation-verified — replacing the column with `NULL::text` fails it with its own diagnostic.

### Then the walkthrough

**The composer adapts.** Luu has no address and one live Dispact conversation, so the header opens with **no picker** and in channel form: To reads *"Continues your Dispact conversation"*, and Subject, Bcc, Send-later and Draft-with-AI are all absent — each withheld for its own reason.

**The consent gate explains itself.** Choosing *Business correspondence* turned the verdict line green with **"they wrote to you on 15 Aug"** — the derived Art 6(1)(f) basis, rendered to the rep. There is no consent grant on file for this person; the capture is what makes the reply lawful, and the UI says so in words rather than just permitting it.

**The pre-flight refuses a dead credential, correctly.** A first attempt was refused with *"this workspace has no connected dispact bot to send through"* — the connection was in `reauth_required` after a bad token was pasted. Not a defect: the request-time pre-flight asking the unit's `Live` is exactly the §8.1 path #1368 added, doing its job before anything staged. Reconnecting cleared it.

**The send transmitted.**

```
status | provider | attempts | provider_message_id
-------+----------+----------+--------------------------------------
sent   | dispact  |        1 | 9959e749-cda7-4642-9be9-2e3686d094a4
```

One attempt, no park, a real id from the live workspace. The recipient was resolved from the conversation and never named by the caller.

**And the timeline reads correctly afterwards** — the channel column now says **Dispact** where it said the bare word "message", every row carries a **Reply**, and the sent message is on the record.

### Gates

| | |
|---|---|
| `make frontend-check` | exit 0, 2692 tests |
| `make craft-static` | 0/0/0, all three roots |
| New tests | 5 frontend (the adaptive rule + both withholding cases), 1 integration (row-to-payload, mutation-verified) |

A screen recording of the walkthrough exists and is being reviewed before publication, for the disclosure reason above.

PR #1387, opened 2026-08-15
fix(channels): the transport registry is written at boot, not when a vault is (#1366)

### LINKED ISSUE #1366: A unit's channel transport registers only when the capture registry is built, which is config-gated

tinnhlt-gra:
Closes #1366 — which I **reopened** first: it had been auto-closed as completed by the merge of #1368, the PR it was *found in*, while the defect was untouched. Re-verified live on `origin/main` before starting.

## The defect

`reconcileChannelProviders` is the ONE writer of `channel_provider` — core connectors plus every channel a composed unit declares — the ONE place a unit shadowing a core transport is refused, and the ONE caller of `activities.SetChannelProviders` / `comms.SetChannelProviders`. Its only non-test caller was `NewCaptureRegistry`, and every non-test construction of that registry sits behind a keyvault gate (`cmd/api/keyvault.go:40`, `cmd/worker/boot.go:420`).

So a vault-less install registered none of it:

1. A unit's declared transport was never written. The ingress door validates against the **declared** set, not the registry (`extingress.go:251`), so a unit-filed `kind=message` is accepted and then violates `activity → channel_provider`'s foreign key — a 500 that names nothing.
2. The collision check never ran, so a unit shadowing `telegram`/`whatsapp` booted clean instead of boot-failing.
3. Both in-memory snapshots kept their compile-time `telegram`-only default (`activities/channelvocabulary.go:33`, `comms/seams.go:302`), so a unit channel was silently unsendable — the one failure a rep cannot tell from a broken provider.

`LoadChannelProviderDirectory` already documents this exact class of defect for the READ side and was moved off that path. The write side now gets the same treatment.

## One correction to the issue's suggested shape

The issue warns of a "two halves, last writer wins" hazard and proposes holding the core and unit sendable sets separately. **That hazard does not exist as merged** — `reconcileChannelProviders` computes the union in one place (`channelprovider.go:96-101`) with a single call site for both setters. Fixing the non-problem would have added a synchronisation concern the code does not have. What it needed was the one writer moved off the vault path.

## The change

- The reconcile is **removed** from `NewCaptureRegistry` rather than left there and duplicated, so the config coupling goes away instead of gaining a second trigger.
- New `compose.RecordComposition` folds the two boot-time facts that follow from the composed extension set — the extension inventory and the channel vocabulary its units declare — into one ordered step. One function because the ordering is not a caller's to choose, and a role that wired only the inventory is exactly the broken state this fixes.
- It **returns its error** instead of halting. A unit colliding with a core transport is a refusal an operator must act on; the panic was only ever a consequence of having nowhere to report from inside a constructor. Both halves are wrapped with the step that failed.
- The core half is **derived, not restated**: `CoreChannelProviders()` enumerates off a pool-less, vault-less registry construction. A hand-kept list would be a second answer to "what did this binary compile in".

Ordering, both satisfied by the call site: **after** `RegisterExtensions` (the unit half reads `ComposedExtensions()`) and **before** `compose.New` (assembly loads the transport directory FROM these rows). That second one also closes a live secondary gap — today a unit channel first composed in this build lands in the table but is missing from *this* process's directory snapshot until the next boot.

## What holds it

| gate | what it pins |
|---|---|
| `TestEveryRoleThatComposesExtensionsRecordsWhatItComposed` | **The wiring itself.** Walks `cmd/` with `go/ast`: every role that calls `compose.RegisterExtensions` must also call `compose.RecordComposition`. Derived, not listed — a new role arrives covered. |
| `verify-boot.sh` step 3/4 | **The boot proof.** CI's live-boot lane already boots with no `MARGINCE_KEYVAULT_ROOT_KEY`; it now asserts the running api publishes every transport the composed manifests declare. Expectation derived from `extensions/*/manifest.generated.json`, so a unit that starts or stops declaring a channel needs no edit here. |
| `TestCoreChannelProvidersEnumeratesEveryComposedCoreTransport` | The enumeration is honest against the richest core composition (`CaptureSyncRegistry`, both OAuth apps). A connector that starts supplying a transport fails here rather than going unregistered on every install with no Google app. |
| `TestTheBootStepRegistersAComposedUnitTransportWithNoCaptureRegistry` | The regression itself, on a role that builds no capture registry. |

The seven existing reconcile/collision integration tests are untouched — they drive `reconcileChannelProviders` directly.

## Proof

**Mutation-tested.** A code review caught that my first version of the boot-step tests would pass against a no-op: they asserted only `telegram`, whose row is seeded by `migrations/core/0240` (and `testdb` deliberately does not reset `channel_provider`) and which both packages carry as a compile-time default — so every assertion was already true before the act. They now empty the sendable sets first (failing loudly if the emptying did not take) and assert a **unit** row, which only the boot step can have written. With `ReconcileChannelProviders` stubbed to `return nil`:

```
channelprovider_integration_test.go:355: reading back the unit transport the boot step should have registered: no rows in result set
--- FAIL: TestTheBootStepRegistersAComposedUnitTransportWithNoCaptureRegistry
--- FAIL: TestTheBootStepReconcilesBeforeTheInstallationIsBootstrapped
```

The AST gate too — dropping `RecordComposition` from `cmd/worker`:

```
--- FAIL: TestEveryRoleThatComposesExtensionsRecordsWhatItComposed
    cmd/worker registers the composed extension set but never calls compose.RecordComposition
```

**UAT — the real binary, booted vault-less.** Throwaway database and its own Redis; `make dev` never run. Full transcript in the working notes.

Baseline (`0b7e41374`, no part of the fix), vault-less boot confirmed by the *absence* of `api connector-credential vault enabled` in the log:

```
 provider | transport |  label   | credential_model | supplies_transport
----------+-----------+----------+------------------+--------------------
 telegram | core      | Telegram | workspace_bot    | t
 whatsapp | core      | WhatsApp | workspace_bot    | f
```

The `dispact-connector` unit composes (`extensions=4`), its migrations apply, its routes mount — and the channel it declares is simply absent. Nothing in the boot log hints at the gap, which is what made this quiet.

Fixed binary, database reset to byte-identical migrated state, same vault-less boot:

```
 provider | transport |  label   | credential_model | supplies_transport
----------+-----------+----------+------------------+--------------------
 dispact  | unit      | Dispact  | per_member       | t
 telegram | core      | Telegram | workspace_bot    | t
 whatsapp | core      | WhatsApp | workspace_bot    | f
```

Booted twice **with** the vault configured (log confirms the vault path engaged): the same three rows after each boot — no regression, and the write is a genuine upsert rather than an append.

## Gates

- `make test-it DIR=backend/internal/compose` — green (80s). `make test-it DIR=backend/internal/compose/integration` — green (150s).
- Pre-push `craft static --strict`: `PASS — 0 blocker, 0 major, 0 minor`.

## Notes

- `internal/compose/integration/apptest` composes a server without this boot step, so it runs on the compile-time default. It registers no extensions, so there is nothing for the step to write there and the divergence is inert today — left alone rather than changed as a drive-by. The AST gate covers `cmd/`, which is where a role can actually get this wrong.
- Rebased onto `1837a13a5`, after #1386 fixed the duplicate `core/0260` that had `main` red.

---
## Summary by cubic
Writes the channel registry at boot for all roles instead of only when a keyvault is configured. Previously vault-less installs skipped registering unit transports and collision checks and kept default in-memory sendable sets; now the registry is recorded after `compose.RegisterExtensions` and before server assembly so unit channels are sendable immediately.

- Review
  - Move reconcile out of `NewCaptureRegistry` into `compose.RecordComposition`, which writes the extension inventory and calls `compose.ReconcileChannelProviders`; add `compose.CoreChannelProviders` (derived from a pool-less registry).
  - `compose.RecordComposition` returns errors; callers updated in `cmd/api` and `cmd/worker`.
  - Gates: AST check in `backend/bootcomposition_test.go` enforces `RecordComposition` when roles call `RegisterExtensions`; integration tests cover unit transport registration without a capture registry and pre-bootstrap; `scripts/verify-boot.sh` asserts the API publishes composed transports.

- Rollout
  - No migration required; writes are idempotent upserts.
  - Operators must resolve any unit that collides with a core transport; boot surfaces the error.
  - Ensure each role calls `compose.RecordComposition` after `compose.RegisterExtensions` and before server assembly.

<sup>Written for commit cdc20856f4815d788cc876ff663e958d169b0154. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1387?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/bootcomposition_test.go
+++ b/backend/bootcomposition_test.go
@@ -1,0 +1,85 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+package backendarch
+
+// The fitness gate on the boot sequence: a process role that composes
+// extensions also records what it composed.
+//
+// It is derived rather than listed. Naming cmd/api and cmd/worker here would
+// leave the next role to be remembered, and the whole failure this guards is a
+// role that quietly does not do something — nothing in a boot log says a
+// transport went unregistered. Asking "which roles register extensions" of the
+// tree means a new one arrives already covered.
+
+import (
+	"go/ast"
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+// The two compose entry points that must travel together. RegisterExtensions
+// declares what this binary composed; RecordComposition is what writes the
+// consequences down — the extension inventory, and the channel vocabulary those
+// units declare. A role with the first and not the second boots with units
+// installed and their transports unregistered, which is the state a captured
+// message on such a transport fails on, several layers away from the cause.
+const (
+	registersExtensions = "RegisterExtensions"
+	recordsComposition  = "RecordComposition"
+)
+
+func TestEveryRoleThatComposesExtensionsRecordsWhatItComposed(t *testing.T) {
+	roles, err := os.ReadDir("cmd")
+	if err != nil {
+		t.Fatalf("reading the process roles under cmd/: %v", err)
+	}
+	composing := 0
+	for _, role := range roles {
+		if !role.IsDir() {
+			continue
+		}
+		called := composeCallsIn(t, filepath.Join("cmd", role.Name()))
+		if !called[registersExtensions] {
+			continue
+		}
+		composing++
+		if !called[recordsComposition] {
+			t.Errorf("cmd/%s registers the composed extension set but never calls compose.%s — "+
+				"it boots with units installed and the transports they declare unregistered, "+
+				"and nothing in its log says so", role.Name(), recordsComposition)
+		}
+	}
+	if composing == 0 {
+		t.Fatal("no process role calls compose." + registersExtensions +
+			" — this gate compared nothing, so it cannot have held anything")
+	}
+}
+
+// composeCallsIn reports which compose.<Name> functions the role's package
+// calls. Selector-matched on the `compose` package name, which every role
+// imports unaliased; an aliased import would read as no call at all, so the
+// zero-roles guard above is what keeps that from passing quietly.
+func composeCallsIn(t *testing.T, dir string) map[string]bool {
+	t.Helper()
+	_, files := parseGoFilesUnder(t, dir)
+	called := map[string]bool{}
+	for _, file := range files {
+		ast.Inspect(file, func(n ast.Node) bool {
+			call, ok := n.(*ast.CallExpr)
+			if !ok {
+				return true
+			}
+			sel, ok := call.Fun.(*ast.SelectorExpr)
+			if !ok {
+				return true
+			}
+			if pkg, ok := sel.X.(*ast.Ident); ok && pkg.Name == "compose" {
+				called[sel.Sel.Name] = true
+			}
+			return true
+		})
+	}
+	return called
+}
--- a/backend/cmd/api/main.go
+++ b/backend/cmd/api/main.go
@@ -92,10 +92,12 @@ func run(ctx context.Context, args []string, stdout io.Writer) error {
 	if err != nil {
 		return err
 	}
-	// Record the composed extension set when it changed since the last
-	// boot — install/upgrade/removal happen in source, so this is where
-	// they become observable (ADR-0069 §5).
-	if err := compose.ObserveExtensionInventory(ctx, pool, logger, extensions); err != nil {
+	// What this binary composed, recorded before it serves: the extension
+	// inventory (install/upgrade/removal happen in source, so this is where
+	// they become observable — ADR-0069 §5) and the channel vocabulary those
+	// units declare. Before the server is assembled, which loads the transport
+	// directory FROM the rows this writes.
+	if err := compose.RecordComposition(ctx, pool, logger, extensions); err != nil {
 		return err
 	}
 
--- a/backend/cmd/worker/main.go
+++ b/backend/cmd/worker/main.go
@@ -86,10 +86,14 @@ func run(ctx context.Context, args []string, stdout io.Writer) error {
 		return err
 	}
 
-	// Record the composed extension set when it changed since the last boot
-	// (ADR-0069 §5); pre-bootstrap it skips — the api records the first
-	// observation once it has bootstrapped the installation.
-	if err := compose.ObserveExtensionInventory(ctx, pool, logger, boot.extensions); err != nil {
+	// What this binary composed (ADR-0069 §5); pre-bootstrap the inventory half
+	// skips — the api records the first observation once it has bootstrapped
+	// the installation. The worker composes the same units the api does and
+	// runs the send path, so it registers the same channel vocabulary and
+	// answers "can a reply leave this installation" from the same set; the
+	// write is an idempotent upsert, so whichever role boots second changes
+	// nothing.
+	if err := compose.RecordComposition(ctx, pool, logger, boot.extensions); err != nil {
 		return err
 	}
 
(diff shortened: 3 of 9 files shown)


tinnhlt-gra:
Pushed a coverage fix after Sonar's gate failed at **54.5% on new code**.

The gap was `compose.RecordComposition` itself — the function the process roles actually call. My integration tests reached past it to `ReconcileChannelProviders`, which left the step every role wires untested and the ordering it owns unproven. That is the same "test the real entry point" problem the review had already caught one layer down, so the fix is the honest one rather than a coverage patch:

- `TestTheBootStepRegistersAComposedUnitTransportWithNoCaptureRegistry` now drives `RecordComposition`, and additionally asserts the **inventory** half landed — both facts follow from the same composed set, and a step that recorded one of them is exactly the half-wired boot this change exists to make unreachable.
- `TestTheBootStepReconcilesBeforeTheInstallationIsBootstrapped` likewise.
- New `TestRecordingTheCompositionFailsTheBootWhenAUnitShadowsACoreTransport` covers the refusal path: the boot aborts, and the message says **which** of the two halves failed, so an operator reading one line can tell a transport collision from an inventory problem.

`backend/cmd/**` is already in `sonar.coverage.exclusions`, so the two call sites in the mains were never the measured gap.

Re-ran locally: `make check-backend` green, `make test-it DIR=backend/internal/compose` green (84s), pre-push `craft static --strict` `PASS — 0 blocker, 0 major, 0 minor`.

Issue #1434, opened 2026-08-16
pipeline trace: company triage and material events are registered but not yet reported


tinnhlt-gra:
The capture pipeline trace registers every stage a captured message passes through, and each stage must declare how it answers: `stored`, `derived`, or — when it answers nothing yet — an explicit absence with a reason. Two stages ship in the `planned` state, meaning **the stage runs, but the trace does not report it yet**.

| Stage | Subject | Where its state lives |
|---|---|---|
| `company_triage` | domain | the triage queue and its sweep reconciler |
| `material_events` | thread | `signal`, `signal_thread_scan` |

`planned` exists precisely so this is tracked rather than invisible: the registry's gates refuse a stage that neither answers nor explains itself, and `planned` requires this issue reference. A member currently sees these two rungs render "runs, but is not reported here yet" instead of a state.

## What each needs

**`company_triage`** — derivable from the triage queue plus the domain's not-a-company marking. The open question is whether it belongs on a per-message ladder at all: its subject is a *domain*, and "this domain was queued for triage" on a message rung may confuse more than it explains. It may be better placed on the company surface.

**`material_events`** — derivable from `signal_thread_scan` (settling, parked with a refusal count, scanned) plus whether any `signal` cites this activity. Worth doing together with #1433, since for chat transports the honest answer is `transport_not_read` and the derivation would be reporting a gap rather than a state.

Neither is blocking: both are additive registry changes plus one derivation read each, with no schema or contract change.

tinnhlt-gra:
## Triaged 2026-08-16 — one decision owed, one item blocked on #1433; assigned for tomorrow morning

Not in today's fix batch because both stages carry an open question rather than missing work.

**`company_triage` — the decision is whether this rung belongs on a per-message ladder at all**, and my read is that it does not. Its subject is a *domain*. A per-message ladder that reports "this domain was queued for triage" is answering a question the reader did not ask on a surface about one message, and the trace's own stage vocabulary makes subject an explicit property precisely so a rung cannot quietly change what it is a claim about. Recommend **moving it to the company surface and retiring the `planned` registration**, rather than deriving it where it currently sits. That is a registry deletion plus a new rung elsewhere, not a widening.

[...]

Issue #1433, opened 2026-08-16
capture: the attention classifier and the conversation reader are email-only, so chat-transport messages are silently never read


tinnhlt-gra:
A member sent a chat message reading *"Are you free at 8:00 PM Monday?"*. It was captured and appeared on the timeline. Nothing else happened — no `meeting` label, no commitment, no open loop — and no surface said why.

## Cause

A message captured over a chat transport is stored with `kind = 'message'` (ADR-0107: `channel_provider` is non-null exactly when `kind = 'message'`). Two AI passes that would have read it are hard-wired to `kind = 'email'`:

| Pass | Predicate | Effect |
|---|---|---|
| Attention classification | `backend/internal/modules/activities/capturelabel.go` — `capture_label IS NULL AND captured_by LIKE 'connector:%' AND kind = 'email'` | the message is never in the backlog, so it is never labelled `commitment \| meeting \| noise` |
| Material-event extraction | `backend/internal/compose/signalextractread.go` — `a.thread_key IS NOT NULL AND a.kind = 'email'` | excluded twice: by kind, and because a chat message usually carries no `thread_key` |

A third, `backend/internal/compose/signalscan.go`, is `a.kind IN ('email','call','meeting')` — also excluding `message`, though that pass is account-level rather than per-message.

## Scope of the decision needed

This is not a bug with an obvious fix. Widening the predicate is a product question per pass:

- **Attention labels** — the classifier's prompt is written about mail ("You label captured emails for attention routing"). Chat messages are shorter, less formal and arrive far more frequently; the batch economics and the prompt both assume mail. Widening the predicate without revisiting either would produce low-quality labels at higher cost.
- **Material events** — the extractor waits six hours for a *thread* to settle and reads the last six messages of it. A transport with no thread key has no unit of work to settle, so this needs a different grouping (per counterparty? per day?) before the predicate can widen.

## Not a workaround

Related: #1382 (a channel-captured person is minted with no email), same theme — channel-captured traffic is second-class to mail across the pipeline.

The visibility half is being handled separately: the capture pipeline trace now reports this as `attention_label: skipped — transport_not_read` rather than leaving the member with silence. That makes the gap legible; it does not close it.

tinnhlt-gra:
## Triaged 2026-08-16 — product decision owed per pass, assigned for tomorrow morning

Agreed this is not a widen-the-predicate fix, and the issue's own framing is why. A recommendation per pass so tomorrow is a decision rather than a re-derivation.

[...]

PR #1441, opened 2026-08-16
feat(capture): a member can read every step their message took, and why a step declined

tinnhlt-gra:
A member sent a chat message reading *"Are you free at 8:00 PM Monday?"*. It was captured. Nothing else happened — no meeting, no commitment, no open loop — and **Settings → Capture activity said `Captured`, and stopped there**.

Producing the answer took a code read: the attention classifier's backlog is `WHERE capture_label IS NULL AND captured_by LIKE 'connector:%' AND kind = 'email'`, the message arrived over Dispact, so it is `kind = 'message'` and was never eligible. Nothing in the product said so.

The defect this PR fixes is not the email-only predicate (that is #1433). It is that **a pipeline stage can decline to run and leave nothing a member can read.**

## What a member gets

The five counters become the filter — "26 waiting on a verdict" is the question people arrive with, and there was no way to ask *which 26*. Clicking a row opens the whole ladder: every stage the message passed, in order, with what each did and **why** a stage that declined declined.

Transport ids finally resolve to names. `useProviderLabel()` has existed since #1354 and this screen never called it, so it printed `ext:dispact-connector:dispact` at a member — against the contract's own instruction that a label is resolved, never stored.

## The shape, and why it is this small

The first two drafts of this design were much bigger: a new event table, a hashed join key, an event emitted per stage. Review deleted a write path each time. What shipped is **one column, no new writer, and a read that joins what the product already stores**.

Every stage declares one answer source:

| Source | Stages |
|---|---|
| **stored** (a `capture_trace` row) | internal-only drop, activity write's one refusal, tier ladder — all three already written today |
| **derived** (durable state a module owns) | activity write, person creation, sender verdict, attention label |
| **declared absent, with a reason** | connector filter (`by_design`), ingress gate (`by_design`), erasure check (`untraceable`), company triage + material events (`planned`, #1434), claim extraction (`not_built`) |

Copying a durable record into a trace creates a second source that can disagree. Migration 0258 says so outright about verdict outcomes — a copy "would collide with itself the moment a sender were re-judged inside the window" — and a job-written copy also lands under no member, where the ingest-written row was that member's alone.

## Distinctions the rungs keep apart

These are the substance; conflating any pair would make the surface state something false.

- **`not_applicable` vs `unknown`** — one says the stage did not apply, the other says we can no longer tell. Past the 24h window only the second is honest, and a message with no rows at all cannot claim the first.
- **`withheld` renders unconditionally** for a non-owner, whether or not a row exists. Withholding only when there *is* one is a row-existence oracle: a colleague comparing two shared messages would learn which hit a replay fault on somebody else's mailbox.
- **Four kinds of absence** — never shown by design, untraceable without breaching an erasure, not read yet, and a step that does not exist — are four different facts about the product.

## Two things made structural rather than conventional

**The funnel metric.** `countTraced` now fires only for registry-opted-in funnel stages. Every stage writing through `Trace()` is a funnel stage *today*, so the unconditional count was correct today and silently wrong the first time it wasn't — inflating `margince_capture_outcomes_total` with no diff to any metric code and no test failing.

**The eligibility predicate.** `ClassifyBacklogPredicate` is ONE string, used verbatim by the classifier's backlog query and by the reader that explains it. A reader with its own copy is precisely the defect this surface exists to expose; reproducing it inside the surface would have been the worst possible bug. An integration test seeds one row per exclusion and asserts both halves agree.

## The registry, and why it is not a list

Three stages shipped invisible because nothing asked. The registry's gates refuse a stage that neither answers nor explains its own silence, so a future stage cannot ship dark. Gate 3 is per-stage-per-status rather than a grep for a call site — a stage recording only `done` would satisfy a grep while its skip paths, the entire motivating class, stayed dark.

## Growth without a frontend release

`PipelineLadder` holds no list of stages; it walks whatever the server sends. A stage this build has never heard of renders from the server's own `label` and `reason_text` rather than vanishing or printing a raw key. That is a deliberate departure from the closed-set rule stated two files away — right for a five-value enum, wrong for a vocabulary designed to grow, where an omission is indistinguishable from "nothing happened". A test asserts an unknown stage still renders and that no raw `pipeline.` key reaches the DOM.

## Verification

- `make check-backend`, `make check-fe`, `make test-integration` (**0 skips**) all green, re-run after rebasing onto `main`.
- Frontend suite: 2,719 tests / 215 files.
- `craft static --strict` clean; the registry gates were mutation-tested (removing one `AbsentReason` correctly fails).
- Migration 0270 applied against a rebuilt database. It was 0269 until the rebase — `0269_person_email_correspondence_evidence` landed on main mid-flight, which is why the number is claimed last.

## Notes for review

- **The migration backfills by CASE, not a flat default.** An `internal` row *is* the internal-only drop (that gate returns before the ladder runs), and an `invisible_incumbent` fault is the activity write refusing a replay. A default would mislabel every in-window row and undercount the funnel's `internal`.
- **Deploy order matters more than usual**: after `SET NOT NULL`, an old binary's insert fails, and per 0258 a trace error on the capture transaction fails the capture. Nothing is lost — a connector sees an error rather than a skip, so its cursor holds — but this is migrate-then-restart.
- **A rebase conflict nearly ate an upstream fix.** Main added `AND a.channel_provider IS NULL` to the disposition join while this branch extracted that join into a shared const. The guard and its rationale are carried onto the shared const.
- `server.go` sat exactly on the 500-line ceiling, so the next handler set embedded was always going to break the gate. The tool-registry rebuild moved to its own file.

## Follow-ups filed

- #1433 — the email-only predicates themselves. This PR makes the gap legible; it does not close it.
- #1434 — company triage and material events, registered `planned`.
- A spec-repo section for this surface is owed after merge: `capture_trace` appears nowhere under `specs/` today, so the surface being extended is itself unspec'd. Deferred deliberately by the owner so the shape is proven before it is normative.

---
## Summary by cubic
Lets members open a per‑message ladder that lists every ingress stage in order, what each did, and why a stage declined — fixing the gap where “Captured” was the last visible step and email‑only stages gave no reason.

- Adds `GET /v1/activities/{id}/pipeline` (human‑only) that returns ordered rungs for all registered stages; unknown stages render from server `label`/`reason_text`.
- Introduces `pipelinetrace` registry (stage order, labels, reason catalog, funnel opt‑in). Only three stages store rows; others derive from durable state to avoid copies that drift.
- Persists `stage` on `capture_trace` and uses it to place rungs; window endpoints still filter to funnel stages while the ladder returns all rungs for one message. `countTraced` now increments only for registry‑opt‑in funnel stages.
- Unifies the attention classifier’s eligibility with a single `ClassifyBacklogPredicate` used by both the backlog and the explanation; integration and unit tests assert agreement.
- Secures the drawer: the assembler checks activity access and omits `activity_id` when the caller lacks the grant. Past retention and non‑owners both render `unknown` with `record_not_available`; `withheld`/`expired` states are removed. HTTP doors 503 when the assembler is not composed, the wire enforces the payload posture, and tests cover doors, wire, auth predicates, and unknown‑stage rendering.
- Frontend: counters act as a client‑side filter; clicking a row opens the ladder. New `PipelineLadder` renders whatever the server sends and resolves provider ids via `useProviderLabel`.

**Rollout**
- Run migration 0271 before restarting application processes. After `SET NOT NULL` on `capture_trace.stage`, old binaries’ inserts will fail and cause capture to error; connectors will hold cursors. Backfill assigns `stage` by CASE; no manual fixes required.

<sup>Written for commit b6497ecfac874f06df9d9a2effb99cdb4fe6310f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1441?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tinnhlt-gra:
@coderabbitai review

tinnhlt-gra:
## UAT — driven against a real stack

Slug stack on :8183, five seeded fixtures, one per shape the ladder must tell apart. Recording below.

**1. The motivating case.** The Dispact message is Captured, and the ladder says why nothing followed:

> **Attention label** · Skipped — *this step reads email only, and the message arrived over another transport*

That is the sentence that took a code read to produce when this started.

**2. The funnel filters, and stays honest.** Clicking a counter narrows the rows and prints "Showing 1 of 1 Waiting on a verdict in this window" — both numbers, because the counters count the window and the filter narrows what is loaded.

**3. The internal-only drop has no activity row at all**, which is why the trace-keyed door exists. Its ladder reads Internal-only check *Skipped — every party was on your own domains*, then **Did not apply** for saved-to-timeline, contact decision and contact created. It never reached those steps, and it does not claim to have.

**4. The four absences stay distinct**, as designed: connector filtering (*not counted here — the numbers mean different things per connector*), admission check (*a fault of the connection, not of one message*), erasure check (*reporting this would restore data an erasure removed*), company check and conversation reading (*this step runs, but is not reported here yet* — #1434), commitments and open loops (*this step does not exist yet*).

**5. Per-sender rungs say so.** Contact decision, contact created and sender verdict each carry *about the sender, not this message alone*; company check carries *about the sender's domain*.

**6. The payload posture is stated once**, at the foot of the ladder, rather than repeated per rung.

**Transport labels**: the Telegram row resolves to **Telegram** through the channel-provider registry. The gmail rows stay `gmail` — mail connectors are not registered channel providers, so `useProviderLabel` falls back to the raw id, which is the documented behaviour rather than a miss.

capture-pipeline-trace-uat

tinnhlt-gra:
@coderabbitai full review

tinnhlt-gra:
## UAT 2 — a real Dispact account, both payload postures

Not fixtures this time. A personal access token deposited against `https://workspace.gradion.com`, the connector authenticating as **Tin Nguyen**, resolving the provider workspace and walking its cursor to high-water-mark 781983. 15 captured and 20 awaiting a verdict, all real.

**Default posture — no raw data.** Rows read `content not stored`; the ladder carries outcomes, decisions, reason classes and ids only, and states the posture ONCE at its foot rather than per rung:

> No sender or subject is stored for any step: this deployment did not turn payload capture on.

**With `capture.trace_payloads: true`.** Same messages, re-walked after resetting the cursor. Sender and subject now appear on the row — `luu.nguyen@gradion.com` / `Luu Nguyen Thanh` — and the posture note is gone.

The part worth checking closely: under the payload posture the sender appears on **Contact decision** and nowhere else. That is the stored rung. Every derived rung beside it — contact created, sender verdict, company check — carries status and reason only, because a derived rung has no payload of its own to show. The design's rule holding on real data rather than on a fixture built to demonstrate it.

**And the motivating case, on a genuinely captured Dispact message:**

> **Attention label** · Skipped — *this step reads email only, and the message arrived over another transport*

That is the question this PR opened with, answered in the product, from a real credential and a real message.

**One thing the run exposed that is not mine to fix here.** The transport column resolves `dispact` → **Dispact** through the channel-provider registry, but a row whose provider is the extension-scoped `ext:dispact-connector:dispact` stays raw — the registry carries no entry for that form, so `useProviderLabel` falls back to the id, which is its documented behaviour. It is the string from the original bug report and it is still ugly. Filed separately rather than widened into this change.

dispact-real-uat-both-postures

PR #1448, opened 2026-08-16
feat(ai): a binding may narrow a capable provider downward

### LINKED ISSUE #1424: A binding cannot narrow a capable provider downward — input: [text] on gemini is refused
## What is missing

#1324 (PR #1422) added `input:` to a routing binding, and restricted it to `openai_compatible` and
`vllm` — the providers whose carriage depends on which model was bound rather than on the wire.
On every other provider the field is a **startup error**.

That is the right default, and it forecloses one legitimate intent: **narrowing a capable provider
downward.**

```yaml
premium:
  provider: gemini
  model: gemini-3.1-flash-lite
  input: [text]      # refused today
```

Gemini carries `image/*` and `application/pdf` from code (`gemini.go:238`). An operator who wants
scanned invoices to stay out of an egressing model — while keeping that model for text — has no
way to say so. The location ladder (`profile:`) is all-or-nothing for the deployment; this is
per-tier and per-modality.

DeepSeek Harness supports exactly this shape, as `modelOverrides: { <model>: { input: [text] } }`
layered over a provider's own declaration.

## Why it was left out

Rejecting is forward-compatible and accept-and-ignore is not: a rejected config can be permitted
later without breaking anyone, while a field that silently did nothing could never be tightened
once operators had written it. So #1324 rejects, and its error deliberately says *not supported on
this provider yet* rather than implying the field is meaningless — this issue is the "yet".

- `validateInput` (`internal/modules/ai/inputmodality.go`) — the whole rule set, including the
  `inputProviders` restriction this issue relaxes.
- `carriageFor` — modality words → media types.
- `model.CarriesMIME` — pattern-aware membership, which is what an intersection needs.

## What to build

The narrowing must be an **intersection with the adapter's own carriage**, never a replacement: a
binding must not be able to widen `ollama` into carrying images it cannot send. So a declaration
on a native provider means *at most this*, and `Caps()` returns
`intersect(adapterCarriage, carriageFor(input))`.

Note the ladder interaction already covered in `docs/reference/configuration.md`: a task's carriage
is the intersection over its bound rungs, so narrowing one rung narrows the task. That is the
correct behaviour here — it is the point — but it should be stated where an operator will read it.

## Watch out

The privacy framing is the reason to build it, so do not let it read as a security control. It
governs what *this product* sends; it is not a guarantee about an endpoint an operator chose. The
existing `profile:` wording (`ai-operational-spec` §1.4) draws that line carefully — match it.

## Done when

`input:` is accepted on every provider, narrowing-only, with a test that a declaration can shrink
a native provider's carriage and cannot widen it.

tinnhlt-gra:
Closes #1424. Spec half (contract-first, P3): gradionhq/margince-foundation#1321 and #1322, both **merged**.

## The gap

`input:` was accepted only on `openai_compatible` and `vllm`, where carriage is genuinely unknowable from code. That was the right default, and it forecloses a legitimate intent:

```yaml
premium:
  provider: gemini
  model: gemini-3.1-flash-lite
  input: [text]      # refused before this
```

An operator who wants scanned invoices kept out of an egressing model — while keeping that model for text — had no way to say so. `profile:` is all-or-nothing for the deployment; this is per tier and per modality.

The rejection it replaces was written to be relaxed: its error said *not supported on this provider **yet***, never *meaningless*. So no existing config changes meaning, and an omitted `input:` still gets exactly what that provider always carried.

## What this does

The field is accepted everywhere and does two jobs:

- On `openai_compatible`/`vllm` it **is** the carriage — unchanged from #1324.
- Everywhere else it **narrows**: `Caps().AttachmentMIMEs` is the intersection of the adapter's wire carriage with what the binding declared.

So `input: [text]` on a gemini tier sends that tier no attachment at all, and **no declaration can hand a wire a lane it lacks** — the intersection is what guarantees that, rather than a rule somebody has to remember.

Each native adapter now holds the binding's list as a field and reads it in **both** `Caps()` and its send-time gate — the shape `openAICompatClient` already had. One list, two uses: a binding cannot advertise what its own gate refuses.

## Found in review, fixed here

- **`WithFakeClient` bypassed the whole feature.** It swapped the caller's fake in *after* `buildClients` had applied the narrowing, so a `{provider: fake, input: [text]}` binding got a client carrying documents anyway. The injected fake now takes the intersection of what the bindings it stands in for declared, and is left alone when none declared anything (a caller's `CarryingNothing()` must survive).
- **The fake's `Stream` skipped the attachment gate its own `Complete` runs.** A lane that refuses on one method and carries on the other has a `Caps()` that answers for neither.
- **A narrowed refusal said only "cannot carry"**, which after this change is wrong twice: the wire *does* carry it, and the operator's own `input:` line is what stopped it. Narrowed refusals name that line now; refusals the wire decides still do not, because pointing at a knob that cannot help is its own misdirection.
- **The widening fitness test kept its own list of wires.** It derives from `knownProviders` now and states the invariant that holds for both kinds of provider: a declaration gets *at most what it asked for*, and *at most what the wire has* wherever the wire has an answer of its own.

## Test-double debt paid

The gemini and openai suites built their clients by struct literal. After this change that is a second, differently-configured production — carriage is decided in `SelectBrain` — so both helpers go through `SelectBrain` like the anthropic one already did (review-loop rule 6).

## Verification

- `make check-backend` + `make craft-static` green; `main` merged in before the gate run
- **100% coverage** on every new function (`narrowedCarriage`, `refuseNarrowedAttachments`, `sharedFakeCarriage`, `carrying`)
- `config/ai-routing.schema.json` regenerated: the conditional that forbade `input` outside two providers is gone from the generator, the committed schema, and the drift assertion. `airoutingschema_test.go` runs a real JSON Schema validator and asserts schema and parser agree per case — including the new `narrowed to text on gemini`.
- No UAT video: the observable behaviour is a `Caps()` answer and a refusal, both driven end-to-end in tests through the real `SelectBrain`/`ParseRouting`. The UAT that mattered for this family shipped on #1437.

---
## Summary by cubic
Accepts `input:` on all providers to control what a binding can be given. Previously rejected on native providers; now it narrows their wire carriage, while `openai_compatible`/`vllm` continue using `input:` as the full carriage declaration. Defaults are unchanged.

- Caps and send-time enforcement now read the same narrowed list across adapters, preventing advertise-vs-gate mismatches.
- Native providers (`openai`, `gemini`, `anthropic`, `ollama`) intersect the wire’s carriage with the binding’s `input:`; `input: [text]` sends no attachments, and no declaration can add a lane a wire lacks.
- Injected `fake` client honors narrowing across tiers via an intersection; preserves `CarryingNothing()` when no binding declares `input:`. Streaming now runs the attachment gate.
- `SelectBrain` populates per-binding carriage for all clients; tests for `openai`/`gemini` build clients through `SelectBrain`.
- Schema and docs updated: `input:` allowed everywhere; omit to take the provider’s default, or set `[text]` to disable attachments. Error messages for narrowed refusals now name the `input:` line.
- No migration required; existing configs behave the same. Optional: add `input:` per tier to narrow carriage where needed.

<sup>Written for commit 0ebf90fb75a0a9a3606a670e09b132f5284830cc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1448?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added configurable input modality support across all chat providers.
  * Added attachment capability narrowing for images and PDFs.
  * Added clearer guidance when attachments are unsupported or restricted.
  * Applied modality restrictions consistently across routing tiers and streaming requests.

* **Documentation**
  * Updated configuration examples, schema descriptions, and reference documentation to explain provider-specific input behavior and capability narrowing.


tinnhlt-gra:
@coderabbitai full review

tinnhlt-gra:
@coderabbitai full review

Issue #1449, opened 2026-08-16
Attachment bytes egress unexamined: no sniff before the carriage decision, no scan on the model path, and the StripReport is discarded


tinnhlt-gra:
Follow-up to #1428, which states the scope. This one is about whether to narrow it.

## Where we are after #1428

The product now says plainly that (a) an attachment's media type is the uploader's claim, not a finding, and (b) the secret stripper does not reach inside an attachment's base64. Both are stated in `ai-operational-spec` §4.2, `docs/reference/configuration.md`, and at the two code sites that decide them.

Saying it does not change what happens. These are the three things that would.

## 1. Sniff for the carriage decision only

Verify the bytes really are the image kind claimed before building the wire part — **without** displacing the stored content type as the authority for anything else (that separation is what `documentextractrun.go` deliberately protects today). Cheap, and it closes "an 8 MB blob labelled `image/png`" for the image lanes. It does nothing for a real image with a credential rendered in it.

Watch out: `image/svg+xml` matches `image/*` by prefix and is a text format that no vision model reads as an image — a sniff that only confirms "is it the claimed kind" still admits it. The lane's admitted set may want narrowing at the same time (see #1439, which is about `image/*` being wider than any vendor decodes).

## 2. Scan attachment bytes before egress

The real answer for content, and real work. There is a seam to build on: the existing `scan_pending` / `attachment_blocked` gates already stand between a file and every read of it — but they are about the file's own state, not about what a model is allowed to be shown.

## 3. Make a redaction that fires inside an attachment observable

`payload, _, err := sendablePayload(...)` — every adapter discards the `StripReport`. No security consequence today, and it means "the stripper touched this call" is unobservable on the real wire (the fake records it; nothing else does). If either of the above lands, that report is the signal you would want, and it is also the cheapest of the three: the router already traces a call, and the report is one more field on that trace.

## Done when

A decision is on record for each of the three, with the ones we take built and the ones we decline written into §4.2 beside the scope statement — so the next reader finds the reasoning rather than the gap.

tinnhlt-gra:
## Triaged 2026-08-16 — decision owed on all three, assigned for tomorrow morning

The issue asks for a decision on each of the three; here is a recommendation per item so tomorrow is a yes/no rather than a re-read.

[...]

PR #1450, opened 2026-08-16
docs(ai): say what an attachment lane does and does not guarantee

### LINKED ISSUE #1428: An attachment's MIME is the uploader's claim, and base64 hides its bytes from the SecretStripper
Found during the security review of #1422; **pre-existing on the `openai` and `gemini` bindings,
which have carried attachments since #701.** #1422 extends the same behaviour to the generic
OpenAI-compatible endpoint, which is why it is worth writing down now.

Two facts that are individually defensible and together decide what an attachment lane means.

## 1. The MIME is the uploader's claim, never sniffed

`documentextractrun.go:224` lowercases and trims `meta.ContentType` and uses it as the routing
authority. `documentextract.go:205` says so deliberately — sniffing here would be a second
content-type authority beside DOC-PARAM-9's, and that was the right call for the reason it gives.

The consequence, once a lane carries bytes: anyone who can attach a file — **including an external
counterparty on a captured email** — can label an arbitrary 8 MB blob `image/png` and have its
bytes egress to the bound endpoint. `model.CarriesMIME` matches `image/*` by prefix, so
`image/svg+xml` is reachable too.

So `input: [image]` is a **carriage declaration, not a content filter**, and nothing in the tree
currently claims otherwise — but it is the kind of thing a reader assumes.

## 2. Base64 makes the bytes invisible to the SecretStripper

The stripper runs over the marshalled wire body (`outbound.go:118`), which is correct and
unbypassed. But the nine rules are anchored on characters `base64.StdEncoding` cannot emit — `-`,
`_`, `.`, `:`, `@`, whitespace — so a credential inside an attached file is structurally
unmatchable once it is a `data:` URL. The same file arriving on the **text** lane would be
scrubbed.

The spec's "secrets never appear in a model-bound payload" (§4.2 / A8) is therefore a text-lane
guarantee. That is a reasonable place to land, but it should be a stated scope rather than an
assumption a reader carries over from the text lane.

## What to do

Not obviously a code change — possibly just an honest scope statement. Options, roughly in order of
cost:

1. **State the scope.** Say in the spec and in `docs/reference/configuration.md` that attachment
   bytes are outside the stripper's reach and that a MIME declaration is not a content check.
   Cheapest, and it stops the assumption.
2. **Sniff for the carriage decision only** — verify the bytes really are the image kind claimed
   before building the wire part, without displacing DOC-PARAM-9 as the *stored* content-type
   authority. Contains #1 for images; does nothing for #2.
3. **Scan attachment bytes before egress.** Real work, and the existing `scan_pending` /
   `attachment_blocked` gates are about the file's own state, so there is a seam to build on.

## Also worth a line

`StripReport` is discarded on every adapter's send path (`payload, _, err := sendablePayload(...)`),
so a redaction that fires inside an attachment is silent. No security consequence — but it means
"the stripper touched this call" is unobservable, and if #1 or #2 above ever land, that report is
the signal you would want.

## Done when

The product says plainly what an attachment lane does and does not guarantee — and if we decide it
should guarantee more, that is a follow-up with a scan seam behind it.

tinnhlt-gra:
Closes #1428. Spec half (contract-first, P3): gradionhq/margince-foundation#1323, **merged**.

Found during the security review of #1422; **pre-existing** on the `openai` and `gemini` bindings, which have carried attachments since #701. #1422 extended the same behaviour to the generic OpenAI-compatible wire, which is why it was worth writing down.

## The two facts

Each is defensible on its own. Together they decide what carrying a document means, and both read the other way round to someone who has only met the text lane.

**The secret stripper does not reach inside an attachment.** It runs over the marshalled body — the right place, and unbypassable — but an attachment rides that body as base64, and every rule is anchored on characters base64 cannot emit (`-`, `_`, `.`, `:`, `@`, whitespace). A credential inside an attached *file* is unmatchable by construction; the same file arriving as text is scrubbed. Reaching it would mean decoding and re-encoding every attachment on every call.

**The media type that picks the lane is the uploader's claim.** It is read from what capture recorded and never sniffed — deliberately, because a second content-type authority beside the stored one is its own defect. So whoever attached the file, **an external counterparty on a captured email included**, chooses which lane its bytes take, and `image/*` matches by prefix.

## Where it is said

In the three places a reader actually meets them, rather than one canonical paragraph nobody is standing next to when it matters:

- `stripper.go` — the pass's own doc comment, next to the rules that make it true
- `documentextractrun.go` — at the lane decision, beside the existing note on why nothing sniffs
- `docs/reference/configuration.md` — the operator-facing carriage section, with what to reach for instead (`profile:`, or narrowing the tier with `input:`)
- `docs/explanation/agent-surface.md` — where the stripper is introduced as covering "*every* outbound payload"

## What this is not

Not a decision to sniff, scan, or narrow. Those have real cost and are #1449, filed rather than folded in — sniffing for the carriage decision only, scanning bytes before egress, and surfacing the `StripReport` every adapter currently discards. Stating the scope is what lets those be judged on their merits instead of discovered by surprise.

No behaviour changes here, so there is nothing to UAT: `make check-backend` and `make craft-static` are green, and the diff is comments and prose.

---
## Summary by cubic
Clarifies what the attachment lane does and does not guarantee: the lane is chosen by the content type recorded at ingress (sniffed for captured attachments; uploader-declared for direct API uploads), and the secret stripper guarantees only the text lane and does not inspect encoded attachments. This prevents assuming attachments are sniffed or scrubbed.

- No behavior changes; updates doc comments in `backend/internal/modules/ai/stripper.go` and `backend/internal/compose/documentextractrun.go`, and clarifies operator docs in `docs/reference/configuration.md` and `docs/explanation/agent-surface.md`.
- Lane decision: the AI path adds no second content-type authority; `image/*` matches by prefix; carriage is not a content check; `input:` states what we carry, not what the bytes are.
- Stripper scope: runs over the marshalled body; attachments ride base64, rules match plaintext, so secrets inside an attached file are out of scope; notes the hazard that alphanumeric rules could coincidentally match inside a blob.
- Rollout: none. If stricter handling is needed, use `profile:` or narrow the tier with `input:`.

<sup>Written for commit 66d27cd3e01f68468d3c23ca0804eb89ad37981d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1450?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/internal/compose/documentextractrun.go
+++ b/backend/internal/compose/documentextractrun.go
@@ -223,6 +223,13 @@ func (d *DocumentExtractor) sourceFor(
 	// ContentType is optional on the row; a document that never declared one
 	// takes neither lane, which is the honest answer — the alternative is
 	// sniffing bytes here, a second content-type authority beside DOC-PARAM-9's.
+	//
+	// What that decides, stated because it is easy to read the other way
+	// (ai-operational-spec §4.2): this type is the UPLOADER's claim, so whoever
+	// attached the file — an external counterparty on a captured email
+	// included — chooses which lane its bytes take. Carriage is not a content
+	// check, and a binding's `input:` says what this product will send, not what
+	// the bytes are.
 	mime := ""
 	if meta.ContentType != nil {
 		mime = strings.ToLower(strings.TrimSpace(*meta.ContentType))
--- a/backend/internal/compose/documentextractrun.go
+++ b/backend/internal/compose/documentextractrun.go
@@ -224,12 +224,13 @@ func (d *DocumentExtractor) sourceFor(
 	// takes neither lane, which is the honest answer — the alternative is
 	// sniffing bytes here, a second content-type authority beside DOC-PARAM-9's.
 	//
-	// What that decides, stated because it is easy to read the other way
-	// (ai-operational-spec §4.2): this type is the UPLOADER's claim, so whoever
-	// attached the file — an external counterparty on a captured email
-	// included — chooses which lane its bytes take. Carriage is not a content
-	// check, and a binding's `input:` says what this product will send, not what
-	// the bytes are.
+	// So the lane is decided by what INGRESS recorded, and what that means
+	// depends on how the file arrived (ai-operational-spec §4.2): a captured
+	// attachment carries the type sniffed from its bytes, with a disagreeing
+	// sender claim recorded rather than obeyed, while a file uploaded through
+	// the API carries its uploading principal's declared type unsniffed. Either
+	// way carriage is not a content check — a binding's `input:` says what this
+	// product will send, never what the bytes are.
 	mime := ""
 	if meta.ContentType != nil {
 		mime = strings.ToLower(strings.TrimSpace(*meta.ContentType))
--- a/backend/internal/modules/ai/stripper.go
+++ b/backend/internal/modules/ai/stripper.go
@@ -18,6 +18,16 @@ import (
 // pass through untouched, because privacy is the location ladder (A8
 // revised), and pretending a regex protects PII would be a false
 // guarantee.
+//
+// It is a TEXT-lane guarantee, and the spec says so in the same breath as the
+// rule itself (§4.2). The pass runs over the marshalled body, which is the last
+// point before egress and cannot be bypassed — but an attachment rides that body
+// as base64, and every rule below is anchored on characters base64 cannot emit
+// (`-`, `_`, `.`, `:`, `@`, whitespace). A credential inside an attached FILE is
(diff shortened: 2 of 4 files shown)


tinnhlt-gra:
@coderabbitai full review

PR #1481, opened 2026-08-16
feat(seed-demo): surfaces beyond records, FX rates, and paper in every language

LarsGradion:
Follows #1445. Fills the screens a demo reaches after the obvious ones, and
fixes the reason the Vietnamese half was invisible.

## What a seeded installation now holds

Verified against a live stack with 171 companies: 803 people, 54 deals, 48
contracts, 81 documents, 11,015 facts, 200 invoices — plus **4 tags applied
157 times, 2 dynamic lists, 18 projects, 6 quotas, 2 FX rates**.

## FX was not optional

Winning a deal FREEZES its rate, so the moment Korean companies started
billing in USD every close was refused:

```
no fx_rate from USD to EUR to freeze at close — an admin must load the rate
for this currency pair before this close can succeed
```

The phase runs before deals close, which is the whole point of it existing.

## The Vietnamese half was invisible

The coverage matrix guarantees counts, never *which* company holds them, and
there are two Vietnamese companies among 171. The hash reliably made both a
prospect and a target — so every Vietnamese contract, document and dong
invoice existed in code and in tests and nowhere a demo could reach. That is
the same as not having built them.

`ensureLocaleIsVisible` promotes one company per non-German language to
customer: after coverage so it cannot pull a cell below its floor, and only
where that language has no customer already. A floor, not a quota — the
Korean half already had two and was correctly left alone.

The database now holds `VULETECH — Hop dong khung` in VND with a Vietnamese
PDF and a `Thoa thuan bao mat` NDA beside it, `엠비젼 — Master agreement` in
USD, and German paper for the K5 half.

## Custom fields are not seeded, and cannot be

`POST /v1/custom-fields` answers **501** — "operation custom-field schema
changes is specified but not yet implemented". The contract describes the
add-field engine in full and the handler is a stub, so that screen stays empty
in the demo because it is empty in the product. The code that would fill it is
deleted rather than commented out; the note in `surfaces.go` is the record.

`GET /v1/custom-fields` also needed the lesson `/v1/attachments` taught:
`object` is a required query parameter with no installation-wide form, so
asking without it is a 422 rather than an empty page.

## Verification

`make craft-static` and the seeder tests pass. Seeded repeatedly against an
isolated `DEV_SLUG` stack; the verify pass is green and a second run creates
nothing.

---
## Summary by cubic
Seeds workspace surfaces (tags, lists, projects, quotas), loads fixed FX rates for non‑EUR currencies, ensures a visible customer per supported language, and scales generated amounts into each account’s currency. Previously non‑EUR deals could not close, Vietnamese assets were unreachable, and VN/US PDFs printed euro‑sized numbers; now close succeeds, localized paper appears, and amounts look plausible.

- FX runs first and loads fixed demo rates for USD, VND, GBP, CHF to EUR; if a new currency appears in the plan, add it to `demoFxRates`.
- Surfaces include tags (applied idempotently), dynamic organization lists, projects advanced through phases, and quarterly seller quotas; the report now includes FX and surface counts.
- Promotes one company per non‑DE locale to customer only when missing, then ensures at least one contract and loose document so each language renders paper; includes a test for customer presence per language.
- Scales generated amounts into the account currency (e.g., VND uses round multiples; USD near EUR at this precision) so PDFs show realistic numbers.
- Custom fields are not seeded (`POST /v1/custom-fields` returns 501; `GET` requires an `object` query).

<sup>Written for commit 36cf4e3c5519ec5fd6ed7146a164ff73fe25684d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1481?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Demo data now includes fixed foreign-exchange rates for non-EUR currencies.
  * Demo workspaces can include tags, dynamic lists, projects, and quarterly seller quotas.
  * CRM demo data now provides customer coverage across supported non-German languages, including contracts and related records.
* **Improvements**
  * Demo seeding reports expanded totals and handles existing data and preview runs more reliably.
  * Currency and CRM data creation is resilient to incomplete or concurrent setup conditions.

### DIFF
--- a/backend/tools/seed-demo/fx.go
+++ b/backend/tools/seed-demo/fx.go
@@ -1,0 +1,105 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+package main
+
+// The conversion rates a multi-currency pipeline cannot close without.
+//
+// Winning a deal FREEZES its rate to the base currency, so a deal in USD or
+// VND is refused outright until a rate exists:
+//
+//	no fx_rate from USD to EUR to freeze at close — an admin must load the
+//	rate for this currency pair before this close can succeed
+//
+// That refusal is correct — a won number nobody can convert is a forecast
+// nobody can add up — and it is what made this phase necessary the moment the
+// dataset stopped being all-euro. Korean exhibitors bill in USD and the
+// Vietnamese ones in dong.
+//
+// The rates are fixed rather than fetched. A demo that reached out to a rate
+// provider would need a key, a network and a story about what happens when it
+// is down; a demo that states its rates is reproducible and cannot drift.
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+// demoFxRates are the rates the seeded currencies convert at, to the euro
+// base. Approximately mid-2026 and deliberately round: these exist so a deal
+// can close and a total can be added up, not so anybody can price a trade.
+var demoFxRates = map[string]string{
+	"USD": "0.92",
+	"VND": "0.000035",
+	"GBP": "1.17",
+	"CHF": "1.05",
+}
+
+// seedFxRates loads a rate for every currency the plan will actually use.
+//
+// It must run BEFORE any deal closes. The rate is frozen at the close, so a
+// missing one is not a warning the seeder can carry past — the advance is
+// refused and the run stops.
+func seedFxRates(c *client, refs pipelineRefs, plan map[string]profile, mode runMode) (int, error) {
+	wanted := map[string]bool{}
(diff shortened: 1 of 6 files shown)


LarsGradion:
@coderabbitai review

PR #1507, opened 2026-08-17
docs(agents): an issue you file says how bad it is and where it lives — and a docs edit stops booting twelve shards

dungbanguyen:
The rulebooks told an agent to file an issue and stopped there. Issues therefore arrived unlabeled — at 297 open, **211 (71%) carried no labels at all**, which left no way to distinguish a finding nobody had triaged from one somebody had read and judged ordinary.

Both docs now state the three axes a new issue carries, and the reasoning behind each rather than just the list.

## What is added

- **Exactly one `priority:` and exactly one `area:`**, with the invariant named outright: unlabeled means untriaged, so filing without labels tells the next reader a falsehood about your own finding.
- **Priority is severity, never schedule.** This is the part most likely to drift, so it is stated as a rule rather than implied: a milestone carries the schedule, so demoting a real defect because it is not this week's work makes the label lie. The qualifying test for `critical` is given as "does this stop somebody else working, or is somebody's data wrong right now" — which is why a flaky gate earns it despite no user ever seeing one.
- **One area, the place the fix lives**, so a filter never double-counts. A doc that is wrong about a subsystem takes that subsystem's area rather than a documentation area, because it belongs beside the code it misleads about.
- **`status: needs-decision` / `status: spec-change`**, which keep unactionable work out of somebody's queue — the second because contract-first (P3) puts the change upstream first.
- **Provenance labels kept rather than tidied**: `capability-gap` and `fast-track-debt` record *why an issue exists*, which is the one thing nobody can reconstruct later.
- **Check for an existing parent tracker** before filing a sibling.

## Split between the two files

CLAUDE.md takes the long form (the qualifying test per level, the area rationale); AGENTS.md takes the digest. Neither paragraph is one of the three parity-gated sections, so they differ on purpose — as the CLAUDE.md header says most sections do.

## Gates

- `TestSharedRulebookSectionsAreIdenticalInBothDocs` — PASS (the gated sections are untouched)
- `make check-craft-doc` — `OK: AGENTS.md ## Craftsmanship present`
- Docs-only diff (`CLAUDE.md`, `AGENTS.md`), so the pre-push Go gates do not apply

## Known limitation, filed separately

The area list is now hand-maintained in two documents and in the repo's actual label set, with nothing holding the three together. Adding a sixteenth area means remembering all three places. That is exactly the "prefer fitness functions over point fixes" rule this repo binds itself to, and it is not fixed here — a follow-up issue records it.

---
## Summary by cubic
Standardizes issue labeling in `AGENTS.md` and `CLAUDE.md` and splits CI scopes to avoid booting twelve Postgres shards on docs-only edits. Previously issues arrived unlabeled and `CLAUDE.md` edits could break the parity unit test without running Go lanes; now each issue carries one `priority:` and one `area:`, and CI separates `backend_db` from `backend` so unit tests run while integration shards skip.

- Require exactly one `priority:` (severity, not schedule) and one `area:` (where the fix lives); optional `status: needs-decision` and `status: spec-change`; keep provenance. `security` is for hardening only — exploitable findings follow `SECURITY.md` via private advisories.
- Docs about a subsystem use that subsystem’s `area:`; check for an existing parent tracker and attach as a sub-issue.
- CI: add `backend_db`; derive `backend` from it by YAML anchor plus `AGENTS.md`/`CLAUDE.md`; include `CLAUDE.md` in the classifier; run the parity unit test on rulebook edits; gate integration shards and the `integration` fan-in on `backend_db` only; update `infra/ci-pipeline.md` to document the split.
- Parity-gated sections remain unchanged.

<sup>Written for commit cb0c2b0f339482451c07f31831602d40ed45d06c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1507?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Documentation**
  - Added standardized issue-management guidance requiring one priority and one area label per issue.
  - Documented priority levels, area categories, severity criteria, optional status and provenance labels, and security-advisory reporting.
  - Added guidance for identifying parent trackers and inheriting priorities where applicable.

- **CI**
  - Improved change detection so documentation-only updates avoid unnecessary database infrastructure while still running relevant backend validation.
  - Database-dependent integration checks now run only when applicable changes are detected.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,6 +67,9 @@ jobs:
       contents: read
     outputs:
       backend: ${{ steps.filter.outputs.backend }}
+      # The database-bearing subset of `backend`. See the filter below for why
+      # the two differ, and by exactly what.
+      backend_db: ${{ steps.filter.outputs.backend_db }}
       frontend: ${{ steps.filter.outputs.frontend }}
       e2e: ${{ steps.filter.outputs.e2e }}
       deps: ${{ steps.filter.outputs.deps }}
@@ -78,7 +81,19 @@ jobs:
         id: filter
         with:
           filters: |
-            backend:
+            # `backend_db` is every path that can change what a DATABASE-backed
+            # lane proves; `backend` is that set PLUS the agent rulebooks, which
+            # a Go unit test reads and no integration test does. The anchor makes
+            # the second derive from the first — two hand-maintained copies of
+            # one list is the shape this repo refuses everywhere else, and it
+            # would rot in the direction that silently under-tests.
+            #
+            # The split exists because one flag was driving two unrelated
+            # things: run the Go unit gates, and boot twelve Postgres shards.
+            # A rulebook edit needs the first and cannot affect the second —
+            # the same reasoning already applied to `infra/**/!(*.md)` below,
+            # applied to the last two entries as well.
+            backend_db: &backend_db
               - 'backend/**'
               # The MCP App admission vocabulary is AUTHORED under frontend/ and
               # copied into the Go package, where a byte-equality test gates the
@@ -111,8 +126,20 @@ jobs:
               # every Go job restores, so an edit to it changes what the gates
               # compile — the same claim on this scope that ci.yml itself has.
               - '.github/actions/**'
-              - 'AGENTS.md'
               - 'sonar-project.properties'
+            # The rulebooks. `backend/agentsdocparity_test.go` reads BOTH and
+            # fails when the three shared sections drift, so a unit lane has to
+            # run for either — but no integration test opens them, so twelve
+            # Postgres shards do not.
+            #
+            # CLAUDE.md was absent from this classifier entirely until now,
+            # which meant an edit to it alone ran no Go lane at all and could
+            # break that parity test with nothing to catch it. Listing it here
+            # is the fix for that, not a convenience.
+            backend:
+              - *backend_db
+              - 'AGENTS.md'
+              - 'CLAUDE.md'
             frontend:
               - 'frontend/**'
               - 'backend/api/**'
@@ -508,7 +535,11 @@ jobs:
   integration-shards:
     name: integration shard (${{ matrix.shard }}/${{ strategy.job-total }})
     needs: [changes]
-    if: needs.changes.outputs.backend == 'true' && (github.event_name != 'pull_request' || github.event.pull_request.draft == false)
+    # backend_db, not backend: this is the twelve-Postgres lane, and the two
+    # paths that differ (the agent rulebooks) cannot change what it proves.
+    # It moves in lockstep with the `integration` fan-in below, which asserts
+    # on this job's result and would read a skip as a failure if only one moved.
+    if: needs.changes.outputs.backend_db == 'true' && (github.event_name != 'pull_request' || github.event.pull_request.draft == false)
     runs-on: ubuntu-latest
     permissions:
       contents: read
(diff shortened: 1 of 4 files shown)


dungbanguyen:
Both review findings accepted and fixed in `d709768a`.

**`security` provenance (CodeRabbit, Minor)** — correct, and the stronger version of the point is that `SECURITY.md` already forbids what my text implicitly permitted: *"Do not open a public issue or pull request for a security finding — a public report before a fix ships puts every deployment at risk."* Listing `security` among the provenance labels with no caveat, in a public repo, invited exactly that. Both docs now carry the carve-out, with the test `SECURITY.md` itself implies — **if you can write the reproduction, it belongs in an advisory** — and the in-scope list enumerated so the judgement is not left to feel.

**`status: spec-change` (CodeRabbit, Major)** — also correct. My wording ("does not belong in this repo's work queue yet") reads as parking the issue, and it contradicts the rule four paragraphs above it in the same file: *contract-first means the spec change LANDS first, not that somebody else writes it.* It now says the issue stays here, linked to the upstream change, as the implementation follow-up — the label sequences the work rather than handing it away.

I took the reasoning rather than the literal suggested diff, so the wording ties back to the existing rule instead of restating it in isolation.

Doc gates re-run green: `TestSharedRulebookSectionsAreIdenticalInBothDocs` PASS, `make check-craft-doc` OK.

dungbanguyen:
Scope extended at the maintainer's request: this PR now also fixes #1509, the classifier defect it was itself paying for.

## What changed

`backend` was one flag driving two unrelated things — run the Go unit gates, and boot twelve Postgres databases. A rulebook edit needs the first and cannot affect the second, since no integration test opens either document.

- **`backend_db`** is now the database-bearing set. **`backend`** derives from it by YAML anchor, plus the two rulebooks. Derived rather than copied, because a second hand-maintained list is the shape this repo refuses elsewhere and it would rot toward under-testing.
- Only the two lanes that open a database move: `integration-shards` and the `integration` fan-in. **They move together deliberately** — the fan-in asserts `success` from the shards, so skipping one alone would report a docs PR as a broken integration lane.
- `sonarcloud` already admits a skipped `integration` by design, and `live-boot` gates on `e2e`, so neither needed touching.

## A real gap found while doing it

**`CLAUDE.md` was absent from the classifier entirely.** An edit to it alone ran no Go lane at all — and `TestSharedRulebookSectionsAreIdenticalInBothDocs`, which such an edit can break, lives in one. It is listed now because of that, not for convenience.

## Verification

The anchor produces a nested list, so I checked rather than trusted the README: `paths-filter`'s `parseFilterItemYaml` recursively flattens (`src/filter.ts:118-119` at the pinned SHA `0e4a8c6`). Confirmed against the pinned revision, not `main`.

Gate assignment after the change:

| flag | jobs |
|---|---|
| `backend` | deterministic-gates · extension-reference · craftsmanship · integration-unit-coverage · govulncheck |
| `backend_db` | integration shards (×12) · integration (RLS + erasure + HTTP e2e) |

## What this PR still costs

It edits `.github/workflows/ci.yml`, which is in `backend_db` — correctly, since it changes the gate. So **this** run is the expensive one. The saving lands on the next docs PR.

Closes #1509.

PR #1513, opened 2026-08-17
fix(design-system): one avatar, so a record is the same mark on every screen

luit-nfq:
Second of four changes from the UI audit. PR #1506 made the controls one size; this makes the identity chip one mark.

## What was wrong

**The tint was opt-in.** Nine callers asked for it, three did not — so a company was a coloured monogram in the list it was found in and a neutral accent chip on the record page that list opened. One click apart, same company. The buying-committee chips on a person page and the colleague chips on the rail beside them disagreed the same way. A mark that identifies a record on one screen and not on the next identifies nothing.

**The tint was keyed on the displayed name**, which is not stable. Renaming a person or a company silently moved them to a different colour on every screen at once, which reads as a different record rather than as a rename.

**Four sizes were being rendered for a prop that admitted two.** 28px and 44px came from `size`; 20px came from `.lt-table.dense .avatar` in `listtable.css` and 76px from `.record-head-wide .avatar` in `composed.css`. A chip's size was decided by a class on its parent while the `size` prop said something that was not true.

**Three initials rules existed.** The design system split on whitespace and took two letters. The shell's own hand-rolled chip split on address punctuation too, so it read `jane.doe@example.com` as "JD" where the design system gave "J" — and every colleague with a J-address got the same chip. The onboarding rail's hand-rolled chip took one letter.

**Two hand-rolls and an SVG re-implementation.** `.acctavatar` in the rail's account block had its own initials rule, its own type size and a flat accent fill. `.mw-person-avatar` at the onboarding rail's foot took one letter on a hard-coded `--mono0Fill`, so every reader was the same colour — while the transcript three columns away drew the same person with the design-system chip and carried a comment claiming the two matched.

**`AvatarStack`'s separating ring was a border**, so a stacked face was 32px while the identical lone chip beside it stayed 28px, and the two sat on different centre lines in one row.

## What this does

- The tint is unconditional. `identity` is what the tone derives from where a caller has something better than the displayed name — a record id, an address — and the name remains the fallback.
- `size` is the whole scale: `xs` dense table, `sm` list row (default), `md` record header, `lg` wide record header. `RecordHead` picks its rung from the `wide` flag it already had, and the `composed.css` override is gone. The dense-table override stays and is commented as the one exception, because density is the table's decision and the chip arrives already built from a caller's cell renderer.
- The monogram splits on address punctuation as well as whitespace — the shell's rule, promoted, because it was the right one.
- Both hand-rolls are deleted. The rail, the onboarding transcript, the onboarding rail foot and Settings → Account are now one mark, all keyed on the address, so they agree with each other and survive a rename.
- The stack's ring becomes a `box-shadow`, and its overlap halves to compensate — the ring stopped costing the chip 4px, and the same pull would have left less of each face showing than before.

## Stories

`make fe-uat` matches a changed component to a story that imports it directly, and four of the touched components had none — two had never had a story at all (`src/design-system/README.md` already listed `MarginceWorkbench` as story-less), two were covered only transitively through `entries.tsx`. So this adds `Shell/Account block`, `Design System/Workbench`, `Onboarding/Company confirm card` and `Onboarding/Conversation workbench`, each covering the states its component has rather than the happy one alone — collapsed rails, a session with no display name, a probe that has not answered, a refused save.

The `Avatars` story itself was three default-size untinted chips, which could not have shown any of the above. It is now the scale, the six tones, the names a monogram rule usually gets wrong (an address, a one-word name, a hyphenated name, a single CJK character), a rename holding its colour, a logo painting over the monogram and a broken one leaving it standing, and a stack beside a lone chip.

## Verification

`make check-fe` green (221 files, 2826 tests). `make fe-uat` green — 141 stories captured, no unclean render. Reviewed in Storybook in both themes; the six tones stay distinct in dark, and the four rungs measure 20/28/44/76.

One thing only the visual pass caught: making the ring a shadow removed 4px from each stacked chip, so the faces tightened and the monograms began colliding. The tests could not see it — jsdom computes no layout.

## Not in this change

Still open from the audit, landing next: the second `Field` exported from `auth.tsx` and imported cross-screen, the change-password inputs that carry no CSS class at all, `MIN_PASSWORD` spelled three times with two different counting rules, and the settings separator/card sweep starting with `capture-activity`'s missing `Panel title`. The `connections.tsx` ego-graph still re-implements the mark as an SVG node with no initials floor; it is a different medium and gets its own pass.

---
## Summary by cubic
Unifies `Avatar` so the same record renders the same mark everywhere. Old behavior: tint was opt‑in, keyed on the displayed name, sizes leaked from CSS, and initials rules differed; stacks drew larger. New behavior: tint is unconditional and keyed on `identity` (name fallback), `size` is explicit (`xs|sm|md|lg`), one monogram rule splits on whitespace and email punctuation, and `AvatarStack` uses a `box-shadow` ring. Chips keep their color across renames, email‑only users get two‑letter monograms, and stacked faces align with lone chips.

- API: remove `tinted`; add `identity?: string`; `size?: "xs"|"sm"|"md"|"lg"`. `RecordHead` chooses `md`/`lg` from `wide`. Onboarding’s company identity card uses `md` (was `lg`).
- CSS: avatar rungs live in `atoms.css`; delete `.record-head-wide .avatar`; keep the dense‑table 20px exception in `listtable.css`; `AvatarStack` ring switches to `box-shadow` and overlap is reduced.
- Call sites: replace hand‑rolled chips with `Avatar` in the rail account block, onboarding workbench foot, and settings identity card; pass `identity=user.email` for the account/settings and onboarding transcript; pass `identity=org.id` in organizations; update company rail/contact rows and connections list.
- Stories/tests: add Storybook coverage for Shell/Account block, Design System/Workbench, Onboarding/Company confirm card, and Onboarding/Conversation workbench; tests cover identity‑keyed tone stability, monogram edge cases, and size rungs.

Migration
- Remove all `tinted` props.
- Pass `identity` where available (record id or email) to keep color stable across renames.
- Set `size` explicitly: `xs` dense tables, `sm` lists/default, `md` record headers, `lg` wide headers.
- Delete hand‑rolled avatar UI; use `Avatar` and `AvatarStack`.
- Do not rely on parent CSS to size avatars; the dense table override is the only exception.

<sup>Written for commit 7b4412df8431d7aaa99b82d01daea93b8ce81de8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1513?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


luit-nfq:
@coderabbitai review

Issue #1130, opened 2026-08-13
Nothing requires CI to pass before a merge into main: the required-status-checks ruleset (and with it the merge queue) is disabled


dungbanguyen:
## What happened

Three PRs broke `main` today, all the same shape.

1. **#1114 and #1104 both took core migration `0224`.** `pgmigrate` refuses to load a namespace holding a duplicate version, so every lane that touches a database failed before its first statement — integration shards, live-boot, the migration unit gates. Fixed in #1117.
2. **The same pair, second failure.** #1114's domain-admission write says `ON CONFLICT (workspace_id, domain)`; #1104 collapsed that unique to `(domain)` an hour later. Nine integration tests failed with `no unique or exclusion constraint matching the ON CONFLICT specification`. Fixed in the same PR.
3. **#1124 then collided on `0226`** with a migration that landed while it was open, and had to renumber twice more.

Each PR was green. Nothing was reviewed carelessly. The two changes were simply never compiled against each other.

## Why the gates cannot see it

`pull_request` runs check out the merge of the PR head with `main` **as it stood when the run started**. A PR that goes green and then waits while another merges is approved against a base that no longer exists.

The repository's ruleset makes that outcome reachable:

```
$ gh api repos/<org>/<repo>/rulesets/<id> \
    --jq '.rules[]|select(.type=="required_status_checks")|.parameters.strict_required_status_checks_policy'
false
```

`strict: false` means a branch never has to be up to date with `main` before merging.

## The two candidate fixes

- **Set `strict: true`** on the required-status-checks ruleset. Exact and complete for this whole class — it is the only thing that guarantees the green run saw the base it merges into. The cost is real: every PR re-runs CI after a rebase, which serializes merges when the queue is busy.
- **A merge queue**, which buys the same guarantee while keeping PRs mergeable in parallel, at the cost of setup and of queue-time failures being less obvious to attribute.

Not proposed: a "does this migration number collide" script. It would catch case 1 and neither 2 nor 3 — the conflict-target break is invisible to any check that does not build both changes together, which is precisely what strict mode does.

This is a repository-configuration decision with a real throughput cost, so it wants a human's call rather than a quiet flag flip.

tinnhlt-gra:
**Re-scoped, not closed — and the reason is the opposite of what was expected.** Verified against `origin/main` `0a475573f` and the live repository configuration.

[...]

PR #1516, opened 2026-08-17
fix(migrations): two migrations claimed 0282, so none of them applied

dungbanguyen:
**main is broken right now.** #1491 and #1487 were both open against a main where `0282` was free, and both merged. A namespace holding a duplicate version does not load a subset — it refuses outright, so `migrate up` fails before its first statement and every lane that needs a database fails with it. main has been in that state since 060f2755.

The offline-demo pair moves to `0283`, which keeps the ordering the merges established: #1491 landed first.

## Why editing a shipped migration is right here

It is normally forbidden, and for a good reason: an applied version never re-runs, so editing one changes what fresh installations get while deployed databases keep the old behaviour, and the two diverge silently.

That cannot happen here. No installation can have applied *either* of these, because the namespace they live in has never loaded. There is no deployed database whose recorded history this contradicts and nothing to repair additively — which is the same reasoning #1117 used for the same failure.

## The recurrence

This is the third time (#1117, then #1130's report, now this). The mechanism is `strict_required_status_checks_policy: false` on the branch protection: a PR whose checks went green against an older main may merge without re-running them against the current one, and migration numbering is precisely the kind of conflict git cannot see. Tracked in #1130.

## Verification

`migrate up` against an empty database: applied 292 core+custom+extension + 7 river migrations, schema at head.

---
## Summary by cubic
Renumbers the offline-demo capture migration pair from 0282 to 0283 to resolve a duplicate version that blocked `migrate up`. Previously, two migrations both claimed 0282 and the loader refused the namespace; now the offline-demo pair is 0283, preserving merge order and allowing migrations to run.

- Diff only renames the offline-demo `*.up.sql`/`*.down.sql` files to 0283 and updates the header comment; no schema or data change.
- Safe to edit the version: the namespace never loaded, so no installation could have applied either 0282; renumbering cannot diverge deployed databases.
- Required actions: none. Merge and run `migrate up` as usual.
- Verified on a fresh database: all migrations apply and the schema reaches head.

<sup>Written for commit 87a9b349da44055ca9c9ba4336eab66c7769ae33. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1516?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


LarsGradion:
Already fixed on main in #1512 — `capture_offline_demo` moved to 0283 and `capture_drop_workspace` kept 0282, which is the direction that does not strand installations that already applied it. `make migration-versions` is green on main now.

The collision was my doing: I admin-merged #1487 past the CI run that would have caught it while #1491 was landing. Sorry for the duplicated work here.

This one can close unless it carries something #1512 missed.

PR #1518, opened 2026-08-17
fix(design-system): one field system, so a password field is the same field everywhere

luit-nfq:
Third of the UI-audit series. #1506 made the controls one size, #1513 made the identity chip one mark; this makes the labelled control one component.

## What was wrong

**A second `Field` lived in `auth.tsx` and was exported from it** — a screen file holding a primitive, imported cross-screen by `setupclaim.tsx`. It existed for a reason: the DS `Field` could not carry a leading glyph and a trailing control inside one outline, and a password reveal has to sit inside the focus ring. `.input-icon` could do the left side and nothing else. So the fork grew, and with it a label size (11.5px uppercase vs 12.5px), a field gap (8px vs 12px), a control height (46px vs 40px), a radius (12px vs 8px) and a hint that was never wired to `aria-describedby` at all.

**Refusals rode the `hint` slot.** "Too short" and "these two don't match" rendered in `--textMeta` — the same grey as the rule they broke, and on `passwordcard.tsx` the same grey as the SUCCESS line four elements above. The only thing distinguishing a failed password change from a successful one was reading the sentence.

**`passwordcard.tsx`'s three inputs carried no CSS class at all.** `.input` never applied and Tailwind's preflight had already stripped the UA chrome: no border, no background, no radius, no focus ring, ~21px tall instead of 40. On the one screen where a signed-in person rotates their own credential. It also wasn't a `<form>`, and its button had no `type` — `Button` defaults to `type="button"` — so Enter did nothing.

**`MIN_PASSWORD = 12` was declared three times and the three disagreed on what a character is.** `auth.tsx:927` counted `password.length` (UTF-16 code units); `passwordcard.tsx` and `setupclaim.tsx` counted `[...password].length` (code points). A password with an emoji passed on the reset screen and was refused on the change screen.

**`setupclaim.tsx` spelled five ids twice** — hardcoded on both `Field` and its input, not `useId`-derived — and had `required` on five inputs with no required marker anywhere.

**The reveal existed on two of five password fields**, while the fork's own docstring said every password field carries one.

**Every form in the product zoomed iOS Safari on focus.** Safari zooms below 16px; `.input` is 13.5px. The sign-in screens were the only surface that had ever fixed it, in their own sheet.

## What this does

- `Field` grows `icon`, `trailing`, `labelEnd` and `error`. The `.field-shell` renders **only** when `icon` or `trailing` is passed, so the ~200 plain fields in the tree emit identical markup.
- `error` announces, sets `aria-invalid`, reads in the danger tone, and draws a danger boundary — keyed on the attribute rather than a class, so it reaches a shell-less field too.
- `usePasswordReveal` returns the button and the input's `type` together. Every password field has one now.
- One `passwordrule.ts`, counting code points.
- The auth fork is deleted; `auth.tsx`, `setupclaim.tsx` and `passwordcard.tsx` all use the one field.
- `--control-h` under `pointer: coarse` rises to 44px, which is where SC 2.5.5's 44×44 was actually aimed. AA's own SC 2.5.8 asks 24×24. The auth surface's 46px went, and `.auth-social` came down with it so the sign-in column is one set of controls again.
- `hintLive` keeps the caps-lock warning announcing, because caps lock is pressed *while* typing.

## A gate, because this nearly shipped broken

Writing the `pointer: coarse` block **inside** `:root` closed that block early and stranded the entire `--space-*` scale, the type scale and the font families in a query a mouse never matches. The visible symptom was the sign-in wordmark sitting on top of its own heading, because `top: var(--authPadBlock)` resolved to nothing and the element fell back to its static position.

Nothing in the tree could see it. The properties were still declared, so `check-space-tokens.sh` passed. The file still balanced, so biome passed. jsdom resolves no custom properties, so all 2826 unit tests passed. The type checker does not read CSS. `make check-fe` was green. It took loading the page.

So `tokens.test.ts` now derives the obligation from the file: `:root` must close before any media query opens, and the scales every surface measures itself in must be declared unconditionally. Both were verified to fail on the real break before being kept.

## Verification

`make check-fe` green (222 files, 2839 tests). `make fe-uat` green — 115 stories, no unclean render. Sign-in, setup and the change-password card reviewed in Storybook in both themes; the sign-in column measures one control height throughout, and both refusal states carry the boundary and the tone.

New: `field.test.tsx` (11 cases covering the shell, the description wiring when error and hint are both present, and the reveal's type/pressed pairing) and `passwordcard.stories.tsx`, which the card has never had.

## Not in this change

Still carried from the audit: the two controls that render unstyled (`persongraph.tsx:218` uses a `.btn-small` that exists in no stylesheet; `person360.tsx:107` uses a bare `.btn` with no variant), the ten hand-rolled Button duplicates, `Button` having no `loading` state, the `SETTINGS_TABS` 12/13/14 disagreement, and the `connections.tsx` SVG ego-graph that re-implements the avatar. Next PR is the settings separator rule and the first tabs — `capture-activity`'s missing `Panel title` and the `StatStrip` slot contract.

---
## Summary by cubic
Unifies all labeled inputs under one `Field` and standardizes password behavior across the product. Old: forked auth field with ad‑hoc shells, refusals in grey hints, and inconsistent reveals; New: `Field` adds in‑outline affordances, first‑class `error` with `aria-invalid`, optional live hints, and a shared reveal — plus global touch height and iOS zoom fixes.

- `Field` now supports `icon`, `trailing`, `labelEnd`, `error`, and `hintLive`. It spreads `{ id, required, aria-describedby, aria-invalid }` to the control. The `.field-shell` renders only when `icon` or `trailing` is present; `aria-invalid` drives the danger boundary for both shelled and plain fields.
- `usePasswordReveal` returns the trailing button and input `type` together; all password inputs reveal inside the outline.
- `passwordrule.ts` centralizes `MIN_PASSWORD = 12` and counts code points; all flows agree on length.
- Tokens: `--control-h` rises to 44px under `@media (pointer: coarse)`; `.input` uses 16px font under 35rem to prevent iOS Safari zoom. `tokens.test.ts` gates `:root` closing before any media query and verifies required scales are unconditional.
- Migrations applied: `auth.tsx`, `setupclaim.tsx`, and `passwordcard.tsx` use `Field`/`TextInput`; the auth‑only `Field` and custom shells are removed. `passwordcard.tsx` is a real `<form>`, uses `Callout` tones, and adds reveals to all three fields. Stories/tests cover `Field`, reveal, and the password card.

- Required migration actions:
  - Use `Field.error` for refusals; keep neutral rules in `hint` (set `hintLive` only when the hint changes while typing).
  - For password inputs, call `usePasswordReveal`, pass `trailing={reveal.trailing}`, and set `type={reveal.type}`.
  - Stop minting control ids; spread the control from `Field` to get `{ id, required, aria-describedby, aria-invalid }`.
  - Remove custom shells and height overrides; rely on `icon`/`trailing` and tokenized control heights.

<sup>Written for commit c17db03fcfdfacf93c216fd4cdb14ce632bcd131. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1518?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


luit-nfq:
@coderabbitai review

Issue #1465, opened 2026-08-16
capture trace: a channel record's ChannelIdentity flag is read off the counterparty, so a message id gets hashed as if it were an account id


tinnhlt-gra:
Found while fixing #1462; not fixed there because it changes no rendered output and deserves its own decision.

## What the flag is for

`TraceEntry.ChannelIdentity` documents itself precisely (`capture/trace.go`):

> ChannelIdentity reports that SourceID is a provider ACCOUNT id rather than a message id — which is personal data, and is hashed on write.

`traceSourceID` acts on it: true → `sha256:…`, false → kept verbatim. The rationale for keeping mail ids verbatim (ADR-0082 §1) is that the id identifies a message rather than a person, and it is what makes a support question answerable.

## What it is actually set from

`sinktrace.go`, `traceEntry`:

```go
channel := rec.Counterparty.ChannelIdentity.Provider != ""
```

That asks how the record NAMES ITS HUMAN, not what its `SourceID` is. Those are different questions, and the dispact connector shows them coming apart: it builds ONE natural key for every notification it lands —

```go
Key: fmt.Sprintf("%s:%d", providerWorkspace, item.ID)   // a notification id
```

— while naming direct messages by a channel account and mentions by an address. So on one connector, with identical key semantics, the direct messages are hashed and the mentions are not. Observed on a dev database:

| outcome | source_id hashed | rows |
|---|---|---|
| captured (DMs) | yes | 15 |
| deferred (mentions) | no | 31 |

## Why it matters

Neither direction is a leak — hashing a message id is conservative, and the un-hashed half is un-hashed correctly. The cost is diagnostic: half of one connector's traces can no longer be looked up by the provider's own message id, which is the thing ADR-0082 §1 kept the id for. It also means the flag's name and its doc comment do not describe what it holds, so the next person reading either will be misled.

## Suggested direction

The record should say what its key is rather than have capture infer it. Either an explicit field on the natural key, or — if the inference is kept — derive it from something that actually tracks the key's shape, not from the counterparty. Whichever way it goes, the flag's doc comment and its source should agree.

Related: the same wrong-proxy pattern in the same file and in `tracestore.go` is what #1462 fixes.

PR #1524, opened 2026-08-17
feat(extensions): a Zalo Official Account's conversations, and replies that leave as the account

tinnhlt-gra:
`extensions/zalo-oa` connects one Zalo Official Account to an installation. A
scheduled pull lands its conversations on the CRM timeline through the core's own
capture pipeline, and a rep's reply leaves back out through the same account.
Route C from the design — an extension on the SP5 channel seam (#1368), not a
core module.

**It has been driven against the real account.** It connected, the poll landed
real conversations, and a reply went out and came back. Three of the defects below
were found that way and could not have been found any other way.

## Two decisions the rest follows from

**1. The credential belongs to the administrator who connected it.** The design
specified a *workspace-level* credential. That cannot ingest:
`compose/extingressauthority.go:108` admits an ingest only for a member holding
one of the unit's **declared user-scoped** secrets — depositing one IS the consent
— so a workspace-only credential is refused on every record it ever tries to land.
Zalo already models the answer: there is no client-credentials grant, an OA admin
approves in a browser, and the pair that comes back is that human's grant. Still
**one account, one connection, every rep replying through it**; only custody is
per-admin, and when that admin leaves the connection parks visibly.

**2. Connecting takes a token pair rather than conducting the browser flow.** That
trip mints the FIRST pair and nothing else; from then on the pair renews itself
unattended. Running it once with a tool removes the one genuinely ambiguous
parameter in this provider's documentation (`code_challenge` base64 vs base64URL —
they differ most of the time, so the wrong choice fails *intermittently* against a
ten-minute code), a screen that had to ask an admin to paste a challenge back into
the developer console, and a redirect that had to match a console setting in two
places. **The cost, stated rather than discovered: onboarding is not self-serve.**

The pasted pair is **renewed immediately, once** — proving the refresh token works
in front of the human who brought it, replacing an expiry nobody knows with one
the provider stated, and taking the credential out of the custody of whatever
produced it.

## The provider's traps, each handled where it bites

- **Every response is HTTP 200**, including an expired token and a package too
  low. Nothing classifies on a status line.
- **`-224` vs `-212`** are never collapsed: one is 2.500.000 đ a year, the other a
  free console toggle. The screen renders the server's own sentence.
- **The tier gate is a capability probe.** `package_name` is a localized display
  string, stored as evidence and compared against nothing — and it is not
  one-shot: a package expiring under a working connection parks at `tier_lapsed`.
- **The refresh token is single-use.** One sealed document, a compare-and-set
  lease, the pair re-read *under* the claim, persist-before-use, and an unkept
  rotation parks with the reason named.
- **Account ids are OA-scoped**, so every identity key is namespaced by the
  account — read from the *credential*, never a request — and the poll reconciles
  the two every tick.
- **A send whose answer never came back** reports `ErrSendOutcomeUnknown`, and
  only that: Zalo takes no idempotency key, so a retry messages a customer twice.

## What review and live use found

Three independent reviews (craft, security redteam, Codex) ran twice. Two of them
**independently found the same worst defect from opposite directions**: the tier
gate ran on the pasted access token while the pair actually sealed came from a
separate exchange, so a row could name account X over a credential answering for Y
— and that id is the namespace every person binding is written under.

Also fixed: a renewal that could spend a single-use token twice; a tier refusal
*after* the renewal that destroyed the pair and then said "connect again"; a
resume that sealed an app secret unable to renew the pair it resumed; a
before-image read outside the row lock that orphaned a live credential; a backfill
that threw away its page progress; a screen that discarded the server's refusal;
and an app secret surviving a disconnect.

Live use added three more: a pair over a day old could not connect at all; `oa_id`
was ambiguous in the upsert (`ON CONFLICT DO UPDATE` — a bare column is ambiguous
with `EXCLUDED`, and the comment asserted the opposite); and **a reply the CRM
sent was captured back as a second copy**, because `listrecentchat` is a global
walk including the account's own outbound and the core writes the sent activity
with no provider id. The unit now remembers what it sent rather than dropping
outbound capture, which would make every timeline one-sided.

## Two core changes, both caused by this unit

**`gen-composition` resolves a mixed secret scope instead of refusing it.** #1523
refused a unit whose secrets span both scopes. This unit inverts the reasoning:
its token is user-scoped because the ingress port requires it, its app secret
describes the installation, and forced into one scope it landed on **Connections**
— *"yours alone; nobody else sees it, and disconnecting it affects only you"* —
about an account every rep replies through. The scope of a key says which
NAMESPACE it lives in; the presence of an installation credential says which page
the unit belongs on. Nothing moves for any other unit in the tree.

**`capture_trace` records a counterparty named by account.** The capture-activity
screen said "no sender recorded" for every channel message from a connector with
no addresses, three lines above "Contact created — Done — about the sender". The
payload was keyed on an address, so an empty one dropped the subject too. The
channel half uses `storekit.ChannelIdentitySuppressed` — an erased channel
identity is under a different kind, and the email check would have answered "not
suppressed" for every erased person and written the name the erasure removed.

## A finding that reopens a design decision

**`type: "nosupport"` is settled** (GUIDE §2): such a row carries no `message`, no
`url`, no `thumb`, no `links`, no `location` — only the parties, a time and an id.
**Three of four messages on a real account came back that way.** DESIGN §1 chose
pull-only accepting the loss of attachments and link unfurls; the measured loss is
the *content of most ordinary messages*. That is a product call this evidence
reopens, not one this PR makes.

## Gates

`make check` exit 0 on current `main` · `craft static --strict` 0/0/0 · golangci
0 issues · `make check-ext-migrations` (both migrations apply and revert as the
restricted `ext_zalo_oa` role) · capture integration suite green · unit suite
**87.2%** of statements · 50 frontend tests.

## Not in this PR, each for a stated reason

Roster/`user/detail` enrichment (it forces the `display_name` vs `user_alias`
product call), a bounded history import, the webhook (pull-only — but see the
finding above), and `user_id_by_app` (kept in the raw evidence, nothing wired).
Inherited rather than decided: `message` carries the six-year GoBD retention floor
by omission, and `message` is absent from `relstrength/interactionkind.go`. And
#1465 — `ChannelIdentity` is derived from the counterparty carrying a provider, so
a message-id natural key is hashed as if it were an account id; the error is
toward hashing and the fix belongs with the natural key.


tinnhlt-gra:
## UAT — Zalo Official Account connector, driven against the live account

**READ-ONLY.** Nothing was connected, replaced or disconnected. This installation holds the *only* live credential for a real Official Account belonging to another company, so `Connect`, `Replace the credentials` and `Disconnect` were never pressed and `DELETE /v1/ext/zalo-oa/disconnect` was never called. Every step below is a read.

### What the walkthrough shows

1. **Settings → Integrations** lists `zalo-oa` under **Installation add-ons** ("Add-ons this installation runs with one shared credential. What you set here applies to everybody."). This is the fix for it having been misfiled under *Your other accounts* on Connections, where the copy says an account is yours alone — wrong for a single shared installation credential.
2. **The Zalo Official Account screen** in its connected state: `Connected` badge, account name **NFQ**, package *Tăng trưởng — 12/08/2027*, last checked, and when the credential renews. No credential field is rendered — the four inputs are replaced by a **Replace the credentials** button, so the App Secret and both tokens are never displayed.
3. **Settings → Capture activity** with two `Zalo Oa` rows visible together: the 17:56 row now carries the **sender name** (`TinNguyen` / "Zalo message from TinNguyen"), while the older 17:38 row, captured before the fix, still reads *no sender recorded*. That side-by-side is the thing under test.
4. **"How this message was handled"** for the 17:56 row: the capture pipeline steps in order, with *Saved to the timeline*, *Contact decision* and *Contact created* all **Done**, and the contact decision naming the sender.
5. **The contact created from the Zalo conversation** (`TinNguyen`), whose timeline carries the Zalo Oa messages in both directions — inbound *Zalo message from TinNguyen*, outbound *Zalo message to TinNguyen* / *Zalo Oa* — which is capture and reply end to end.

### What it does NOT prove

* No write path was exercised: no connect, no credential replacement, no disconnect, no reply sent from the UI. The outbound messages on the timeline are from an earlier session, not from this run.
* It does not prove token refresh works — *Credential renews by* is a displayed value, not an observed refresh.
* It does not prove the backfill of the older rows; pre-fix rows still say *no sender recorded* by design, so this shows the fix going forward only.
* It is one account, one conversation, one installation — no multi-conversation, error-path, or revocation coverage.
zalo-oa-uat

PR #1529, opened 2026-08-17
fix(settings): one seam rule, and the cards that were drawing their own

luit-nfq:
Fourth of the UI-audit series, covering the founder's items 5 and 6. Two reports, one cause: the separators on the settings pages disagreed with each other, and the cards behind them were each solving the same layout problem privately.

## The seam

An audit found **twenty-odd independent `border-top: 1px solid var(--borderSubtle)` declarations** across the screen sheets — seven drawn as `border-bottom`, six full-bleed, seven inset by the body's own 20px, with a `:last-child` reset on about half. The same card could show a line reaching its edges above one section and a line stopping 20px short above the next.

One rule now, in `panel.css`:

> A boundary between two **sections** of a card is a **full-bleed hairline**. A section that wants air instead takes **spacing and no line at all**.

There is no third option, because the third option is what the tree had. Declared as adjacency (`.panel-body + .panel-body`, `.panel-row + .panel-body`) rather than an opt-in class, so a caller cannot forget it. Two spellings of exactly this already existed in screen sheets — two screens independently discovering the same missing rule.

`.panel-body > .section-header:first-child { margin-top: 0 }` is the reset `.card` has always had and `.panel` never got. Unreset, a rule followed by a level-3 heading measured **45px** on the connections tab, **41px** on general and **8px** on people — three distances for one relationship. The reported "Telegram bot has a large additional margin top" was that.

`.table` gains a last-row reset for the same reason at a smaller scale: a rule divides rows, so the last row has nothing to divide from. Without it every table in the tree ended on a hairline pointing at the card edge a few pixels below. Three screens had already reset it privately.

## Capture activity

`<Panel>` with **no `title`**, and a `SectionHeader` nested inside it. So the heading landed as a direct child of `.panel` — which has zero padding — and the card's own name sat flush against its left border while everything under it was indented 20px by `PanelBody`. No header band, no rule, and a title at 15px/600 where the other thirteen settings tabs read `--fs-sm`/`fw-medium`. One prop.

Its funnel passed `<button>` wrappers to `StatStrip`, which documents `StatCard` children — so the plate's slot reset landed on the button and the cards kept their own borders and 12px radius: **five bordered rounded boxes nested inside a plate built to draw none**. The screen answered with `all: unset`, which ties `.stat-strip > *` on specificity, so which rule drew the slot was decided by **stylesheet import order**. `StatStrip` owns a slot that is a control now, and the screen has nothing left to unset. The pressed fill lands on the slot rather than on a wrapper the `StatCard` repainted over — which is why selecting a filter used to show almost nothing.

Its rows were an 8px-gapped stack where each `border-bottom` floated in the middle of the space between two rows rather than dividing them, and the last drew one with nothing under it. Its body took one `form-stack` gap, replacing intervals of 0 / 12 / 8 / 0 / 0 that left Load-more butted against the last row.

## Connected inboxes

The failure of a connect attempt rendered in a `PanelBody` of its own, **two sections below the buttons that caused it**, past a section rule and in the gap before an unrelated heading.

It renders beside the strip that owns it now — and there are two, because one mutation drives both the add block's provider picks and a roster row's Reconnect. `connect.variables` decides which strip owes the explanation; moving it wholesale would have parked a failed *reconnect* under four unrelated *Add* buttons.

Three hand-styled `style={{ color: "var(--danger)" }}` lines in this card and its two connect forms become `Callout tone="danger"`, removing the only place in the tree where the danger colour was spelled by hand.

**Button widths stay shrink-wrapped, deliberately.** `.btn-sm` sets the width floor back to 0 because the small rung lives in dense rows, and a floor would not do what the report wants anyway: side by side, every label longer than the floor still sets its own width, so it would widen "Gmail" toward "Google Calendar" without lining anything up. The real defect was that the two containers disagreed about `align-items`.

## What the new rule broke, and what fixed it

The adjacency rule is a behaviour change, so it found two live regressions on `users-admin`: `.users-intro` and `.users-roster` are adjacent bodies, so the seam is now drawn there — and the roster's first `<li>` drew its own `border-top` one pixel under it, doubling the hairline, while the skeleton/empty surface that renders instead of rows had no top padding and sat flush against the new line. Both fixed; `.users-row` now spells `PanelRow`'s shape exactly, reset included.

## Verification

`make check-fe` green (221 files, 2822 tests). `make fe-uat` green — 47 stories captured, no unclean render. Capture activity and Connected inboxes reviewed in Storybook in both themes, including the pressed-filter state and both new failure placements.

New: two `play`-driven stories for the connect failures, which were only reachable by interaction and had no story in either placement.

## Deliberately not done

`consent.css`'s rows live in a `Card`, which has no full-bleed row slot, so they stay inset by the card's padding — reaching past it with a negative margin would be a third spelling. Converting that surface to a `Panel` is the real answer and is its own change. Same for `company-context`'s groups, where making each its own `PanelBody` means lifting a `QueryGate` out of the body; they took the sanctioned spacing-and-no-line branch instead.

## Still carried from the audit

The two controls that render unstyled (`persongraph.tsx:218`'s `.btn-small`, which exists in no stylesheet; `person360.tsx:107`'s bare `.btn`), the ten hand-rolled `Button` duplicates, `Button` having no `loading` state, the `SETTINGS_TABS` 12/13/14 disagreement, and the `connections.tsx` SVG ego-graph that re-implements the avatar. Next is the agent passport card — the create form moving to a drawer.

---
## Summary by cubic
Unifies panel seams under one adjacency rule and renders connect failures inline under the buttons that caused them. Old: screens drew their own inset/bleed rules, tables ended on dangling hairlines, and connect errors rendered in a separate section. New: panel seams are full‑bleed via adjacency, tables reset their last row, StatStrip slots behave as controls with pressed/focus chrome, and errors render inline as `Callout tone="danger"`.

- Panels: seam policy in `panel.css` (`.panel-body + .panel-body`, `.panel-row + .panel-body`), heading margin reset for first `SectionHeader`.
- Capture Activity: uses `Panel title`, one `form-stack` body, list rows switch to top-border adjacency, StatStrip slot reset removes nested card borders and shows pressed/focus states.
- Connectors: failure reason attaches to its strip (Add or Reconnect) as ARIA `alert`; actions alignment unified; small buttons remain shrink‑wrapped; refactor into `ConnectorRow`; new Storybook plays and tests cover both failure paths.
- Tables: global last-row border reset in `.table`; confirmed in `extension-access` and `import`.
- IMAP/Telegram forms: inline danger styles replaced with `Callout`.
- Consumer mail domains: baseline becomes its own `PanelBody` band with the panel seam above it.
- Consent: Card-inset rows draw a top border with first-row reset.
- Users-admin: roster first-row reset avoids doubled seam; intro padding corrected.

<sup>Written for commit c34014fe45b599566926ba26940285142f771f39. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1529?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/design-system/atoms.css
+++ b/frontend/src/design-system/atoms.css
@@ -81,6 +81,18 @@
   border-bottom: 1px solid var(--borderSubtle);
   color: var(--textContent);
 }
+/* A rule DIVIDES rows, so the last row has nothing to divide from and draws
+   none. Without this every table in the tree ended on a hairline pointing at
+   whatever sat under it — usually the card's own edge a few pixels below, which
+   reads as a doubled border. Two screens had already noticed and reset it in
+   their own sheets (`.provider-spend-table`, and now `.ext-matrix` and
+   `.import__table`); this is the same obligation derived from the component
+   instead of maintained as a list of the tables somebody remembered. `th` as
+   well as `td`, because a table with row headers puts one in the last row. */
+.table tbody tr:last-child td,
+.table tbody tr:last-child th {
+  border-bottom: 0;
+}
 .table tbody tr {
   cursor: default;
 }
--- a/frontend/src/design-system/panel.css
+++ b/frontend/src/design-system/panel.css
@@ -35,6 +35,36 @@
   padding: var(--space-5);
 }
 
+/* THE SEAM RULE, and it is one rule: a boundary between two sections of a card
+ * is a full-bleed hairline, and a section that wants air instead takes spacing
+ * and no line at all. There is no third option, because the third option is
+ * what the tree had — twenty-odd independent `border-top: 1px solid
+ * var(--borderSubtle)` declarations across the screen sheets, seven of them
+ * drawn as `border-bottom` instead, six full-bleed and seven inset by the
+ * body's own 20px, with a `:last-child` reset on about half. The same card
+ * could therefore show a line that reached its edges above one section and a
+ * line that stopped 20px short above the next.
+ *
+ * Declared as adjacency rather than as an opt-in class so a caller cannot
+ * forget it, and so two sections that ARE adjacent cannot disagree about
+ * whether they are. Two spellings of exactly this already existed — an opt-in
+ * `.connector-add` with no reset, and `.cf-screen > .panel-body + .panel-body`
+ * — which is two screens independently discovering the same missing rule. */
+.panel-body + .panel-body,
+.panel-row + .panel-body {
+  border-top: 1px solid var(--borderSubtle);
+}
+
+/* A section heading INSIDE a body has a body's padding above it already, so it
+ * does not also pay `.section-header`'s own 24px top margin. Unreset, a rule
+ * followed by a level-3 heading measured 45px from line to text on the
+ * connections tab, 41px on general and 8px on people — three distances for one
+ * relationship, none of them chosen. `.card` has carried this reset since it
+ * was written; `.panel` never got it. */
+.panel-body > .section-header:first-child {
+  margin-top: 0;
+}
+
 /* A hairline row that runs full-bleed against the panel's own edges and
  * highlights on hover — the shape every list inside a panel wants. No border
  * on the first row: the header above it already draws that line. */
--- a/frontend/src/design-system/statstrip.css
+++ b/frontend/src/design-system/statstrip.css
@@ -53,6 +53,49 @@
     0 -1px 0 var(--borderSubtle);
 }
 
+/* A SLOT MAY BE A CONTROL. A strip of filters is still one comparison — the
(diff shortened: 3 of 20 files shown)


luit-nfq:
@coderabbitai review

PR #1532, opened 2026-08-17
fix(settings): the passport card lists, and minting is a drawer

luit-nfq:
Last of the founder's six items. #1506 made the controls one size, #1513 the identity chip one mark, #1518 the labelled control one component, #1529 the separators one rule — this one moves a form out of a row it never fit in.

## What was wrong

Minting a passport was a row inside the card: a `t-label` span, a name field, five scope checkboxes and the submit, **all direct children of one flex line with a single 8px gap between every pair**. Nothing said where the field ended and the choices began, or that the five ticks were choices *for* the field beside them rather than for each other. The file's own CSS comment states the intent — *"a label, a name field, five scope ticks and the button, on as few lines as the width allows"* — which is the shape being fixed.

Four things were wrong beyond the layout:

**The label was not a label.** `<span className="t-label" id>` + `aria-labelledby` announces correctly and is *not* a label: clicking the words does not focus the control, and the browser's own form semantics never engage. `Field`'s docblock names this exact anti-pattern as the reason it exists and counted eleven call sites when it was written. This was the twelfth.

**The scopes were not a group.** Five sibling checkboxes, no grouping element, no accessible name — nothing said what they were choices for. `.field-multiselect` has existed in `atoms.css` for exactly this, and `create.tsx` has used it since it was written.

**The refusal said nothing.** A failed mint rendered as a `t-caption` two blocks below the button, past the token region, with no `role="alert"` — so it announced nothing and sat where nothing had been pressed.

**Nothing had a test.** A credential-issuing flow whose result is disclosed exactly once shipped with zero coverage of minting.

## What this does

The card **lists** what exists and carries one verb in its header band. The form is a **drawer** (`Modal placement="right"`), which the product already uses for surfaces a reader works in rather than glances at.

- `Field` for the name — a real `<label htmlFor>`.
- `fieldset` + `legend` for the scopes, with a hint the group is described by.
- `Callout tone="danger" live="alert"` inside the form, under the button.
- An empty scope set refuses through `Button`'s `reason`, so the disabled control says why: a passport with no scope is a credential that can do nothing.
- **The drawer does not close itself on success.** Closing would take the only sight of the credential with it — the list carries metadata and the server will not re-disclose a token. The live region stays mounted for the drawer's whole life rather than appearing with the token in it (a region inserted at the same moment as its content is not reliably announced), and focus still moves to it. Closing resets the attempt, so re-opening starts clean.

## The sibling call site

`AuditLogFilterFields` hand-rolled the same span-plus-`aria-labelledby`, and its comment justified it: *a real `<label>` per control would wrap every field onto its own line.* That is not what happens — `.field` **is** a flex column and the grid around it decides the layout either way, so the span bought nothing and cost the click target. Converted, and `.audit-filter` (a byte-for-byte copy of `.field`) deleted.

Fixing the invariant rather than the call site under review is house rule 1.

## Tests

Five new, since the flow had none:
- the form is in a dialog, with a named scope group and a label that focuses its control
- an empty scope set cannot be submitted past, and the reason is attached to the control rather than left in a `title` no screen reader announces on a disabled button
- the token is announced, focus lands in its region, and the dialog stays open around it
- a refusal is announced inside the form
- re-opening starts clean rather than showing the previous mint's token

Plus two `play`-driven stories (light and dark) for the drawer, and new copy in all three catalogs — `en`, `de`, `vi` — with key parity held by `i18n.test.ts`.

## Verification

`make check-fe` green (221 files, 2827 tests). `make fe-uat` green — 16 stories captured, no unclean render. Drawer and card reviewed in Storybook in both themes.

## One finding, not fixed

**`Button` cannot take a `ref`.** It types its props as `ButtonHTMLAttributes` where `TextInput` uses `ComponentPropsWithRef` and documents why ("a caller that opens this field itself needs the node, and React 19 passes `ref` as an ordinary prop"). I wanted one here for focus restoration and did not need it in the end — `Modal` restores focus to its opener by default and the Mint button does not disappear — so this is reported rather than fixed.

## Still carried from the audit

`persongraph.tsx:218` uses a `.btn-small` that exists in no stylesheet and `person360.tsx:107` a bare `.btn` with no variant — both render unstyled today. Ten hand-rolled `Button` duplicates, four inside the design system. `Button` has no `loading` state, so 100 sites hand-roll it with no `aria-busy`. `SETTINGS_TABS` registers 14 entries while the file's prose says 12 in two places and 13 in another. And `connections.tsx:173` re-implements the avatar as an SVG node with no initials floor.

---
## Summary by cubic
Moves passport minting from an inline row to a right-side drawer to protect the one-time token and fix form semantics. Old: a row with a pseudo-label and ungrouped checkboxes; New: a header action opens a drawer that keeps the token visible, prevents mid-flight closes, and uses accessible labels/groups.

- Drawer: right-side `Modal`; cannot close while mint is pending (Cancel disabled; Esc ignored); stays open on success; Done/Cancel close; closing resets label/scopes and the mutation.
- Form: real `Field` label for the agent name; `fieldset`/`legend` for scopes with a hint; submit disabled with a reason when no scopes; shows “Minting…” while pending.
- Token and errors: token in an always-mounted `role="status"` region that receives focus; refusals render a `Callout` with `live="alert"` under actions.
- Audit Log filters: replace span+`aria-labelledby` with `Field` labels; remove `.audit-filter`.
- Styles, stories, i18n, tests: remove `.passport-mint`; add light/dark drawer stories; add i18n keys `settings.minting`, `settings.mintCancel`, `settings.mintDone`, `settings.passportScopes*` in `en`, `de`, `vi`; tests moved to `frontend/src/screens/settings-passports.test.tsx` cover accessibility, disabled-with-reason, token announce/focus, refusal, clean reopen, and “cannot close mid-flight”.

**Migration**
- No backend changes. Delete downstream uses of `.passport-mint` or `.audit-filter` and use `Field`, `fieldset`/`legend`, and `.field-multiselect`.

<sup>Written for commit 390900863bcf1d5606edff1e0479c4385d8c95f2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1532?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Passport creation is now handled in a right-side modal with accessible scope selection and validation.
  * Successful creation displays the one-time token with improved focus and announcement behavior.
  * Creation errors are shown inline, and the modal resets cleanly when reopened.
  * Audit-log filters now provide clearer labeled controls.

* **Localization**
  * Added English, German, and Vietnamese translations for passport creation progress, completion, cancellation, scope guidance, and validation.

* **Bug Fixes**
  * Improved handling of refused passport creation and prevented submissions without required scopes.


luit-nfq:
@coderabbitai review

luit-nfq:
All three addressed in `e7c0dd5`:

1. **Oversized test file** — the mint suite and `mintBackend` moved out to `frontend/src/screens/settings-passports.test.tsx`. `settings.test.tsx` is back to its prior size rather than 157 lines larger.

2. **`userEvent.setup()` per test** — each test now creates one instance and drives every interaction through it, including `openDrawer`, which takes the test's instance rather than making its own.

3. **A pending mint could be closed** — this was the real one, and worse than a stray refetch. `mint.reset()` detaches the observer but does not cancel the POST, so a drawer closed mid-flight still created the passport server-side — and its token, disclosed exactly once and never re-served, went with the drawer. The list carries metadata only, so there was no way back to it. `closeMint` now returns early while `mint.isPending`, which covers Escape and the backdrop as well as Cancel, and Cancel is disabled alongside it.

The new test hangs the request rather than racing a real one, and I checked it against the unfixed code first — it fails there, so it is actually holding the behaviour.

PR #1554, opened 2026-08-17
fix(jobs): a wait owns its own deadline, so three awaits cannot share one budget

### LINKED ISSUE #1538: flaky: TestCaptureOvernightJobsRegisterAndRun shares one 60s deadline across three sequential job awaits
## What happened

`integration shard (12/12)` failed on an unrelated PR with:

```
capturejobs_integration_test.go:194: timed out waiting for "capture_enrich" to complete: context deadline exceeded
--- FAIL: TestCaptureOvernightJobsRegisterAndRun (60.20s)
```

The PR that hit it changed only `backend/tools/gen-composition` (its own module, no product code) and frontend files — nothing reachable from `internal/compose`.

## Why it is fragile, not broken

`TestCaptureOvernightJobsRegisterAndRun` starts a real River runner and then awaits three job kinds against **one shared deadline**:

```go
waitCtx, cancel := context.WithTimeout(ctx, 60*time.Second)
for _, kind := range []string{"capture_classify", "capture_enrich", "capture_digest"} {
    awaitKindCompleted(waitCtx, t, sub, kind)
}
```

So the budget is not 60s per kind — it is 60s for all three, and whatever the first await spends is taken from the second. On a contended runner the later kinds starve, which is exactly the observed shape (it timed out on the **second** kind).

Measured locally on the same commit, against the lane's own Postgres:

- single run: **3.59s**
- `-count=3`: **7.4s total**, 3/3 pass

That is ~17x headroom against the ceiling, and it still blew the budget in CI — so the ceiling is not the problem, the *sharing* is. The run in question had also been queued for roughly an hour behind a runner backlog.

## Suggested fix

Give each kind its own deadline derived from the work it waits on, rather than slicing one budget three ways — the same rule the repo already applies elsewhere (a timeout should be derived from the bound it must outlast, not a round number shared by unrelated waits). A per-kind `context.WithTimeout` inside the loop is the one-line version.

Worth checking the sibling awaits in the same file for the same shape while there.

## Not fixed here

Found while shipping an unrelated change; filed rather than worked around. The affected job passes on re-run.

tinnhlt-gra:
Closes #1538.

## The defect

`TestCaptureOvernightJobsRegisterAndRun` awaited three job kinds under a **single** 60s context:

```go
waitCtx, cancel := context.WithTimeout(ctx, 60*time.Second)
for _, kind := range []string{"capture_classify", "capture_enrich", "capture_digest"} {
    awaitKindCompleted(waitCtx, t, sub, kind)
}
```

That is not 60s each — it is 60s *between* them, and whatever the first await spends the second goes without. On a contended CI runner the second kind starved and the job failed on `capture_enrich`; the same commit passes in ~3s on an idle machine.

Its sibling `TestBackfillCompletionBuildsTheDigest` had the same shape and was **closer to the edge than it looked**: its three awaits take ~11s together, all drawn from one budget, with unrelated work interleaved between them.

## The fix

Removing the parameter rather than re-slicing it. `awaitKindCompleted` now derives its own deadline, so there is nothing left for a caller to share and **the shape cannot be written again** — the compiler refuses it, which is a stronger guard than an assertion about it would be.

That also puts the helper back in line with its two siblings: `awaitRows` and `waitUntil` have always owned their own clocks. It was the only wait in the package taking a caller's context, and the only one with the defect.

Three smaller decisions:

- **The budget is spelled once** as `awaitBudget`, read by all three helpers, so they cannot drift into three answers to one question.
- **The `runner.Stop()` graces keep their own literal.** A shutdown grace is a different bound from a wait for a job; I folded them together in a first pass and undid it — that would have been the same mistake in miniature.
- **`mustDeadline` is deleted.** It existed only to hand this helper a fresh context and went dead with its last caller.

## Sweep — no sibling left to fix

Every other await site was checked for the same shape. All are **1:1**, one deadline per await:

| file | deadlines | awaits |
|---|---|---|
| no_activity_reminder_workqueue | 1 | 1 |
| embedreindex_fanout | 2 | 2 |
| embedreindex | 3 | 3 |
| capture_auto_enrich_sweep | 1 | 1 |
| webhookretry_pass | 3 | 3 |
| agentscheduler | 2 | 2 |
| privacyretention | 4 | 4 |

The two `capturejobs` tests were 3:1 and 3:1 — the only instances.

`jobtest.AwaitKindsCompleted` is deliberately **untouched** despite taking plural kinds under one context: its kinds are awaited *concurrently* in one select loop, so a single deadline for the set is correct rather than shared. Same for the exported `integration.AwaitKindCompleted`, whose every call site already derives its own context — changing that signature would churn six packages to fix nothing.

## Verification

- Both affected tests pass: `TestBackfillCompletionBuildsTheDigest` 10.80s, `TestCaptureOvernightJobsRegisterAndRun` 2.92s.
- **Mutation-checked**: forcing `awaitBudget` to `1ms` fails with `timed out waiting for "capture_classify" to complete within 1ms`, proving the per-call deadline is wired rather than ignored. The message now names the budget it blew, which the old one did not.
- Full lane green: `OK: integration passed with 0 skips (41 packages, 2666 tests)`.
- `make check-backend` green.

No product code changes — test helpers only.

---
## Summary by cubic
Makes `awaitKindCompleted` own its deadline so sequential waits no longer share one budget. Old: three awaits could run under one 60s context and starve on contended runners. New: each await gets its own 30s `awaitBudget`. Side effects: the helper’s signature drops the context parameter and timeout messages include the budget.

- Updated call sites in `backend/internal/compose/capturejobs_integration_test.go`, `extjobs_integration_test.go`, and `jobs_integration_test.go` to call `awaitKindCompleted(t, sub, kind)`; removed shared `context.WithTimeout` and `mustDeadline`.
- Added `awaitBudget` and reused it in `awaitRows` and `waitUntil`; kept `runner.Stop()` grace separate.
- Left `jobtest.AwaitKindsCompleted` and exported `integration.AwaitKindCompleted` unchanged (concurrent waits or per-call contexts already correct). No product code changes or external migration required.

<sup>Written for commit d4917cfc03df469c92dc68ce0949f619b8e19cc9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1554?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Tests**
  - Improved integration-test reliability by giving each completion wait its own timeout.
  - Standardized test wait durations and timeout handling.
  - Reduced failures caused by sequential waits sharing an exhausted deadline.
  - Updated boot, backfill, follow-up digest, overnight, and external job completion checks.

### DIFF
--- a/backend/internal/compose/capturejobs_integration_test.go
+++ b/backend/internal/compose/capturejobs_integration_test.go
@@ -70,10 +70,8 @@ func TestBackfillCompletionBuildsTheDigest(t *testing.T) {
 		}
 	}()
 
-	waitCtx, cancel := context.WithTimeout(ctx, 60*time.Second)
-	defer cancel()
 	// Drain the boot digest so the next one cannot be it — nor deduped by it.
-	awaitKindCompleted(waitCtx, t, sub, CaptureDigestWorkspaceArgs{}.Kind())
+	awaitKindCompleted(t, sub, CaptureDigestWorkspaceArgs{}.Kind())
 
 	// The boot pass placed its own dispatcher row, so "no dispatcher exists" is
 	// not the claim to make — "the completion added none" is. Read the count
@@ -87,9 +85,9 @@ func TestBackfillCompletionBuildsTheDigest(t *testing.T) {
 	}, &river.InsertOpts{UniqueOpts: river.UniqueOpts{ByArgs: true, ByState: activeSweepStates}}); err != nil {
 		t.Fatalf("enqueue backfill: %v", err)
 	}
-	awaitKindCompleted(waitCtx, t, sub, "capture_backfill")
+	awaitKindCompleted(t, sub, "capture_backfill")
 	// The digest that follows the completed backfill is the payoff wiring.
-	awaitKindCompleted(waitCtx, t, sub, CaptureDigestWorkspaceArgs{}.Kind())
+	awaitKindCompleted(t, sub, CaptureDigestWorkspaceArgs{}.Kind())
 
 	// The waits above are satisfied by a digest of the right KIND, which the
 	// boot pass also produces — they prove the wiring fires, not WHAT it
@@ -188,9 +186,7 @@ func TestCaptureOvernightJobsRegisterAndRun(t *testing.T) {
 		}
 	}()
 
-	waitCtx, cancel := context.WithTimeout(ctx, 60*time.Second)
-	defer cancel()
 	for _, kind := range []string{"capture_classify", "capture_enrich", "capture_digest"} {
-		awaitKindCompleted(waitCtx, t, sub, kind)
+		awaitKindCompleted(t, sub, kind)
 	}
 }
--- a/backend/internal/compose/extjobs_integration_test.go
+++ b/backend/internal/compose/extjobs_integration_test.go
@@ -150,7 +150,7 @@ func seedAgentSeat(t *testing.T, ws ids.UUID) ids.UUID {
 // reports the ones that also finished.
 func awaitRows(t *testing.T, pool *pgxpool.Pool, kind string, want int) int {
 	t.Helper()
-	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), awaitBudget)
 	defer cancel()
 	var got int
 	for {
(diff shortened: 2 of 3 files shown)


tinnhlt-gra:
@coderabbitai full review

Issue #1588, opened 2026-08-17
Two schema fitness gates fail on main: activity_retention_evidence is unclassified for both visibility and the data reset


tinnhlt-gra:
## What

Two integration fitness gates fail on a clean `origin/main` (`cd7f91fd7`), both about the `activity_retention_evidence` table introduced by the recent privacy work (#1571 / #1578 / #1582):

**`TestFK_rowScopedTargetsHaveVisibilityDecision`** (`backend/migrations/schema_fitness_integration_test.go:345`)

```
FK activity_retention_evidence.activity_id -> activity has no visibility decision
FK activity_retention_evidence.deal_id    -> deal     has no visibility decision
```

**`TestSweepTargetsCarryNoDeleteBlockingTrigger`** (`backend/internal/compose/datareset_integration_test.go:322`)

```
sweep target "activity"                      carries DELETE-firing trigger "activity_refuse_restricted_mutation"
sweep target "activity_retention_evidence"   carries DELETE-firing trigger "activity_retention_evidence_is_frozen"
```

## Reproduce

```
git worktree add /tmp/probe origin/main --detach && cd /tmp/probe/backend
go test -tags integration ./migrations/  -run TestFK_rowScopedTargetsHaveVisibilityDecision -count=1
go test -tags integration ./internal/compose/ -run TestSweepTargetsCarryNoDeleteBlockingTrigger -count=1
```

Both fail on a detached `origin/main` worktree with no local changes.

## Why it matters

These are the two gates that exist so a new table cannot quietly acquire either property:

- The FK gate holds "anything that references a row-scoped record is a read of it" — an unclassified FK means nothing has decided whether reading through it needs `auth.EnsureLinkTarget`.
- The sweep gate holds that a protected/append-only table is listed in `preservedResetTables` rather than swept — as written, the data reset would attempt a DELETE the trigger refuses, so the reset fails partway.

Each gate wants either a real gate at the reference or an explicit classification. Neither is a test-only fix.

## Not caused by the change that found it

Found while running the lane for an unrelated upload-ceiling branch. Verified upstream by re-running both against a clean `origin/main` worktree before filing.

Third failure in the same lane, already tracked separately: #1572.

tinnhlt-gra:
Fixed in #1589. The two gates wanted different answers, which this issue's text did not anticipate — recording the reasoning here since it changes what a reader should expect to find.

[...]

Issue #1581, opened 2026-08-17
An attachment upload can leave an orphan object when the row transaction fails


tinnhlt-gra:
## What

`Store.UploadAttachment` puts the object **before** the row commits, deliberately: "a committed row always has its bytes; a failed write leaves at worst an orphan object, never a row promising bytes that are not there" (`backend/internal/modules/activities/attachmentupload.go`).

That ordering is the right one. What is missing is the other half — nothing ever removes the orphan. `blob.Delete` is not attempted on the second transaction's error paths, and no sweep looks for objects with no row.

The window is not hypothetical. The write transaction re-checks `ensureContractFileable` before inserting, so a contract archived or re-anchored **during** the upload deterministically produces a stored blob with no row referencing it.

## Why it matters

Orphaned objects accumulate silently in the bucket. They are invisible to the product (no row lists them), so they are also invisible to the retention engine and to an Art. 17 erasure cascade, which both walk rows. An orphan produced from a subject's document therefore outlives an erasure request that believed it had reached everything.

## Not caused by the ceiling work

Found while reviewing #1539 and confirmed pre-existing: the ordering and the in-transaction re-check are both present at that PR's base commit. Filed rather than fixed there, because the fix is a decision about which mechanism owns cleanup — a compensating delete on the error path, or a sweep that reconciles bucket against table — and that is bigger than the change that surfaced it.

## Notes

Two reviewers raised it independently on #1539. House rule 5 (never rationalize a known gap in a comment) is the reason it should not simply keep its current comment: the comment states the trade-off honestly but nothing gates it.

tinnhlt-gra:
Duplicate of #663, which reports the same defect (both attachment writers put the object before the row commits, nothing ever reclaims the orphan, and Art. 17 erasure reads `storage_key` off the row so it can never reach one) and is the broader of the two — it covers `RecordCapturedFiles` as well as `UploadAttachment`, cites DOC-AC-9 requiring an *enqueued* cleanup rather than a best-effort in-request delete, and already weighs sweep-and-reap against a compensating delete. It is also already a child of tracker #1495.

The detail unique to this issue — that the in-transaction `ensureContractFileable` re-check makes the orphan a deterministic outcome, not only a crash-window one — has been carried across as a comment there. Closing here; track it on #663.

Issue #1572, opened 2026-08-17
TestADismissalHoldsUntilTheEvidenceMoves fails deterministically on main


dungbanguyen:
## What

`TestADismissalHoldsUntilTheEvidenceMoves` (`backend/internal/compose/integration/person_relationship_room_integration_test.go:413`) fails on a clean checkout of `main`, deterministically — 4/4 runs.

```
person_relationship_room_integration_test.go:413: the dismissed moment
"moment:nothing_needed" came back against unchanged evidence
```

## Reproduce

```
make db-up && make migrate
cd backend
export MARGINCE_TEST_DSN=... MARGINCE_TEST_APP_DSN=...
go test -tags integration -count=1 \
  -run TestADismissalHoldsUntilTheEvidenceMoves ./internal/compose/integration/
```

Confirmed against `origin/main` with no local changes, so it is not a branch artefact. Found while running `make test-integration` on an unrelated config branch.

## Why it matters

The property under test is a real one: a moment a human dismissed must not reappear while the evidence behind it is unchanged. Either the product regressed and the test is right, or the test's notion of "unchanged evidence" drifted from the product's. Both readings need someone who knows the relationship-room evidence model — I did not want to guess at it inside an unrelated change.

Worth noting it does **not** fail in CI's sharded lane on recent merges, so the shard split or its fixture ordering may be masking it. That is itself worth understanding: a test that passes only when sharded is not gating what it claims to.

tinnhlt-gra:
## It is not the dismissal — the rung never fires

Diagnosis while this was blocking an unrelated PR (#1589). Recording it so whoever picks this up does not repeat the hour.

The failing assertion is at `person_relationship_room_integration_test.go:413`, which reads as "the dismissed moment came back". That framing sends you into the dismissal keying, and the dismissal is fine.

The real failure is one step earlier. The test seeds a 20-day-old unanswered **outbound** activity and expects the gone-quiet rung. The first `Assemble` already returns **`moment:nothing_needed`** — so `dismissed.ClaimKey` is `moment:nothing_needed`, the dismissal correctly suppresses that key, and the page then re-derives the same `nothing_needed` because nothing else qualifies. The assertion fires against a dismissal that worked.

So the question to chase is why `goneQuietMoment` (`internal/compose/person360/moments.go:329`) returns false for that fixture. It bails on exactly two conditions:

- `page.LastOutboundAt == nil`
- an inbound at or after the outbound

[...]

PR #1589, opened 2026-08-17
feat(uploads): how large a file this installation takes is the operator's number, not the build's

tinnhlt-gra:
## The gap

#1539 gave the Documents card an upload button and, to make it work at all, replaced the chassis's single 1 MiB body bound with a route-scoped ceiling. That ceiling is **25 MB because a constant says so**.

Whoever runs the installation is the only one who knows whether that is right. Scanned contracts on a machine with disk to spare want a different answer than a container with a small ephemeral volume — and today the answer requires a rebuild.

Reviewing the code that shipped it turned up eight more defects, six of which live in the same lines this change touches.

## What this adds

`uploads:` in `margince.yaml`, per route, over compiled-in defaults (OPS-CFG-12, DOC-PARAM-11 — [margince-foundation#1340](https://github.com/gradionhq/margince-foundation/pull/1340), merged first):

```yaml
uploads:
  attachment_mb: 25        # POST /v1/attachments
  csv_import_mb: 10        # POST /v1/imports/sources
  linkedin_import_mb: 8    # POST /v1/me/linkedin-connections
```

**Per route**, because a CSV import and a signed PDF are different resource questions: raising what a contract may weigh should not raise what a parse will attempt.

**Decimal MB**, because the config key, the server's refusal and the upload form's hint all state this number and have to state the same one. `25 << 20` reads as "25 MB" in a sentence while admitting 26.2 million bytes. That tightens the two import caps by 4.8% — the code moving to match refusal messages that have always said "10 MB" and "8 MB".

**What the file does not decide is which routes may carry a file.** That list stays in source, and it is the whole security property: several handlers decode `r.Body` with no bound of their own and two of those routes are unauthenticated, so a route carrying no file must remain unable to obtain the file bound by claiming a media type. An operator picks the number; nobody picks the list at run time.

The number is published read-only on `GET /installation/settings`, which every role may read, so the upload form states and enforces **this** installation's limit and refuses an oversize file before it is sent. With no answer yet the upload proceeds and the server refuses — a limit the client invented is one the reader cannot argue with.

## Eight defects, fixed where they live

**Found reviewing #1539's code:**

1. **The route census compared counts** — `len(parsers) != len(routes)` — so a mistyped path left the real route parsing under the 1 MiB JSON bound while the totals still agreed. Both gates now compare sets, one derived from the contract and one from the tree.
2. **The 413 truncated the limit it named.** `limit/1_000_000` read an 8,388,608-byte ceiling as "8 MB" and refused an 8.2 MB file the sentence said would fit; anything under a megabyte read as "0 MB".
3. **The attachment handler blamed the caller for its own I/O.** `ParseMultipartForm` has already consumed the body, so what reached the `io.ReadAll` branch was a spill-file read failure — answered as `422 file/too_large`.
4. **Nothing ever spilled.** The parse threshold was pinned at the whole ceiling, so resident memory scaled with concurrent upload size, twice over once the bytes were copied for the blob store.
5. **The contract and the LinkedIn how-to still promised fixed numbers** — "the 10 MB import body cap", "bounded at 8 MB" — the exact drift this change exists to remove, in the two documents a caller actually reads. (The how-to also called an oversize upload a 422; it has been a 413 since the refusal learned to tell the two apart.)

**Found by the reviewers, in my own new code:**

6. **My rewritten census was package-scoped**, and `internal/compose` already serves an upload route — so a second undeclared multipart parse anywhere in that package read as covered. The same hole as the one I had just fixed, one level down. Every parse site now names its route.
7. **The contract harvest read only `post`, and only inline request bodies.** A multipart PUT or a `$ref`'d body dropped out of the obligation set silently, so the *reader* half of the gate was never armed.
8. **The CSV import route was the one upload site without the unset-ceiling guard** — two of three call sites fixed. With no ceiling it answered `413 "the upload exceeds the 0 MB limit"`, blaming a good file for a number this composition never set.

Also **streamed the attachment write path**: 1 MiB spill threshold, `Content io.ReadSeeker` hashed and measured in one pass then rewound for `blob.Put`. Size and checksum come from the same pass, so they cannot describe different content.

## Every gate mutation-verified

A guard whose failure has never been observed is a guard nobody has checked. Each of these was made to fail first:

| Gate | Mutation | Result |
|---|---|---|
| `TestTheDeclarationGateCanActuallyFail` | mistyped path, ghost route, missing route | all three reported |
| `TestEveryMultipartParseNamesItsRoute` | undeclared `ParseMultipartForm` added to `internal/compose` | red |
| `TestAStoredUploadKeepsItsOwnSizeAndChecksum` | `digest` returns a constant | red |
| `TestTheLinkedInRouteRidesItsOwnConfiguredCeiling` | deleted the `peopleHandlers.WithUploadLimit` wiring line | red |
| `TestTheContractHarvestSeesWhatItCannotAccountFor` | multipart PUT, `$ref` body fixtures | both reported |

## UAT — live stack, non-default ceiling, driven through the UI

Recorded against a `DEV_SLUG=upcfg` stack booted with **`attachment_mb: 3`**, so the recording proves the file is read rather than that the default still works. Video in the comment below.

- `GET /installation/settings` reported `max_upload_bytes: 25000000`, then **3000000** after the config change alone.
- The dialog's hint rendered **"Up to 3 MB."** — the configured number.
- A 4 MB file was **refused in the form** ("That file is larger than 3 MB, which is the most this installation accepts") with **zero requests to `/v1/attachments`**, confirmed by network inspection.
- A 2 MB file uploaded and landed as a row — a file the old 1 MiB bound would have refused outright.

Over the API on the same stack:

```
2 MB → 201  byte_size 2000000  checksum b3fcf913…88fb4f
4 MB → 413  {"code":"body_too_large","detail":"the upload exceeds the 3 MB limit"}
```

`shasum -a 256` of the local file matches the stored checksum byte for byte, through the new streaming path.

## Review

Reviewed by Codex and Fable before opening; findings 1–8 above are theirs and mine combined. Two review points were **answered rather than changed**, and say so in the source:

- The client compares the **file** against a ceiling that bounds the whole **request**, so a file within a few hundred bytes of the limit is still refused by the server. Deliberate: subtracting a margin would refuse files the installation accepts, over a number the reader was never shown. The test now claims only what it proves.
- "Never fully resident" was overstated — a file under the 1 MiB spill threshold is held in memory by design, which the comment now says.

One boundary worth naming, since it looks like a missed call site: the contract form's signed-PDF dropzone uploads through the same endpoint and does **not** pre-check size. It also states no limit, so nothing there contradicts anything — the server's refusal is the honest path. The pre-check belongs where a number is promised, and only the Documents dialog promises one.

Both reviewers independently flagged a pre-existing orphan-object gap (blob stored, row transaction fails, nothing cleans up). Confirmed present at #1539's base commit, so filed as **#1581** rather than widened into this PR.

## Gates

- `make check-backend` and `make check-fe` — green on the rebased tree.
- New-code coverage: `uploadlimit.ts` 100%, `adddocument.tsx` 99.2%, `deployconfig/uploads.go` and the census gates covered by their own suites.
- **No AI certification is affected**: no prompt, corpus, fixture or AI task is touched. Verified rather than assumed — `make check` carries the record-currency gate.
- Three tests fail in the integration lane, and **all three fail identically on a clean `origin/main` worktree** — verified by re-running each against a detached `origin/main` before filing, not assumed:
  - `TestADismissalHoldsUntilTheEvidenceMoves` — already tracked as **#1572**.
  - `TestFK_rowScopedTargetsHaveVisibilityDecision` and `TestSweepTargetsCarryNoDeleteBlockingTrigger` — both about `activity_retention_evidence` from the recent privacy work; filed as **#1588**.

  Everything else in the lane is green, including the five new upload-ceiling suites.

## Note on lineage

This began as work on #1539, which merged while it was in progress. The commits were moved onto a fresh branch off `main`; the content is unchanged.

---
## Summary by cubic
Makes upload ceilings operator-configurable per route and publishes the active limit to clients. Previously all multipart uploads rode a compiled 25 MB cap; now each upload route enforces an installation-set limit (defaults: 25 MB attachments, 10 MB CSV import, 8 MB LinkedIn), and the Documents dialog pre-refuses files over this installation’s limit. Refusals name the exact decimal limit; an unwired ceiling is a 500, not “exceeds 0 MB”.

- Config: add `uploads:` to `margince.yaml` (decimal MB, 1–100). `deployconfig` validates and resolves defaults via `EffectiveUploads`. The active attachment limit is exposed as `max_upload_bytes` on `GET /installation/settings`; docs and example config state defaults.
- Server: new `compose.WithUploadLimits` injects per-route ceilings once and wires them to the chassis and handlers; limits apply to `activities` (attachments) and `people` (LinkedIn import). Upload-capable routes remain hardcoded. Multipart parsing now spills at 1 MiB; attachment writes stream, computing size and checksum in one pass. 413 messages use exact decimal MB.
- Client: the Add Document dialog reads `max_upload_bytes` via a shared hook, shows “Up to {size}”, and blocks oversize files locally. If the settings read fails or omits the field, the client does not invent a limit and the server enforces.
- Correctness/safety: compare declared upload routes as sets from the contract and the tree (all methods, resolves `$ref`); detect undeclared multipart parses; guard unset ceilings with 500; ensure LinkedIn and CSV import parse under the injected limits. New unit and integration tests cover wiring end to end.

**Migration**
- Operators: optionally set per‑route limits under `uploads:`; omitting keys keeps defaults.
- Internal callers: `activities.UploadAttachment` now accepts streaming content (`Content io.ReadSeeker`) instead of `Body []byte`. Update call sites.

<sup>Written for commit 4d3e7eefd234b27c489d338a699e19f6924a0ccb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1589?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/api/crm.yaml
+++ b/backend/api/crm.yaml
@@ -12858,7 +12858,7 @@ components:
       description: |
         The installation's identity and reporting basis (ADR-0090/A135). Read by every role,
         changed only by admin/ops.
-      required: [name, timezone, base_currency, base_currency_locked]
+      required: [name, timezone, base_currency, base_currency_locked, max_upload_bytes]
       properties:
         name:
           type: string
@@ -12881,6 +12881,18 @@ components:
           description: |
             Why the currency is locked, naming how many deals have already converted against
             it. Absent when it is still changeable.
+        max_upload_bytes:
+          type: integer
+          format: int64
+          description: |
+            The largest upload request this installation accepts, in bytes — set by whoever
+            operates it, not compiled into the build (OPS-CFG-12, DOC-PARAM-11). Read-only:
+            it is a deployment fact, so PATCH does not carry it.
+
+            An upload surface states and enforces THIS number rather than one of its own, so
+            a file too large is refused before it is sent instead of after. The ceiling bounds
+            the whole request, part framing included, which is why a client should leave a
+            little room rather than compare a file against it exactly.
     UpdateInstallationSettingsRequest:
       type: object
       description: A sparse installation-settings patch (admin/ops, human-only).
--- a/backend/api/crm.yaml
+++ b/backend/api/crm.yaml
@@ -11734,7 +11734,8 @@ paths:
   # Every one is human-only (IEM-WIRE-7 / A139): an import is the least
   # reversible write in the product, and a confirm-first gate over a
   # thousand-row dry-run report is a rubber stamp that manufactures evidence of
-  # review. Rate class RL-BULK-WRITE; body cap CAP-BODY (10 MB, import only).
+  # review. Rate class RL-BULK-WRITE; body cap CAP-BODY (import only, 10 MB by
+  # default and set per installation — `uploads.csv_import_mb`, OPS-CFG-12).
   # =========================================================================
   /imports/sources:
     post:
@@ -11777,7 +11778,11 @@ paths:
         '401': { $ref: '#/components/responses/Unauthorized' }
         '403': { $ref: '#/components/responses/PermissionDenied' }
         '413':
-          description: 'The upload exceeds the 10 MB import body cap (CAP-BODY). A distinct refusal, never a truncated read.'
+          description: |
+            The upload exceeds this installation's import body cap (CAP-BODY) — 10 MB by default,
+            set by whoever operates it (`uploads.csv_import_mb`, OPS-CFG-12). The refusal NAMES the
+            configured number rather than one compiled in, and it is a distinct refusal, never a
+            truncated read that imports half an estate and reports success.
           content:
             application/problem+json:
               schema: { $ref: '#/components/schemas/Problem' }
--- a/backend/cmd/api/boot.go
+++ b/backend/cmd/api/boot.go
@@ -116,6 +116,11 @@ func declaredSurfaceOptions(cfg apiConfig, deployCfg deployconfig.Config, pool, 
 		// zero value: an unconfigured `seeds` section means built-in defaults
 		// on both provisioning paths, which is the behaviour to preserve.
 		compose.WithBootstrapSeeds(deployCfg.Seeds),
+		// The per-route upload ceilings (OPS-CFG-12). Always applied, including
+		// for a file with no `uploads` section: EffectiveUploads resolves the
+		// compiled-in defaults, so this stays the one place the numbers are
+		// decided either way.
+		compose.WithUploadLimits(deployCfg.EffectiveUploads()),
 	}
 	reset, err := newResetLane(allowDataReset, pool, rdb, logger)
 	if err != nil {
--- a/backend/internal/compose/bodyceiling.go
+++ b/backend/internal/compose/bodyceiling.go
@@ -6,9 +6,10 @@ package compose
 import (
 	"mime"
 	"net/http"
-	"strings"
 
+	"github.com/gradionhq/margince/backend/internal/platform/deployconfig"
 	"github.com/gradionhq/margince/backend/internal/platform/httperr"
+	"github.com/gradionhq/margince/backend/internal/platform/httpserver"
 )
 
 // Which routes carry FILES, and may therefore read a body wider than the JSON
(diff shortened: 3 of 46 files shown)


tinnhlt-gra:
## UAT — the ceiling is the operator's, proven live

Recorded on a `DEV_SLUG=upcfg` stack booted with **`attachment_mb: 3`** in `config/margince.yaml` — deliberately not the 25 MB default, so this shows the file being read rather than a constant still working.

In order: the Documents card, **Add a document**, the hint reading **"Up to 3 MB."** (the configured number), a 4 MB file **refused in the form** with Upload disabled and the reason naming the same 3 MB — no request left the browser, confirmed by network inspection — then a 2 MB file chosen, titled and uploaded, landing as a row. That 2 MB upload is the case the old 1 MiB chassis bound refused outright.

upload-ceiling-uat

Issue #1566, opened 2026-08-17
gen-composition writes a tool's fragment_hash without the sha256: prefix its job entries carry


tinnhlt-gra:
In a generated `manifest.generated.json`, the `fragment_hash` of an **agent tool**
entry is a bare hex digest while the **job** entry in the same file carries a
`sha256:` prefix, and every `digest` in the file carries one:

```
  "id": "job/poll_chats",
  "fragment_hash": "sha256:90cc1fc8eb…"
  ...
  "id": "tool/zalo_oa_connect",
  "fragment_hash": "7f84402cff…"
```

The manifest is the record an operator reads to resolve what a unit requests, and
a field whose encoding depends on which kind of entry it sits on is one a reader
has to special-case. Both spellings appear in every unit that declares tools and
jobs, so this is not new to any one of them.

Noticed while reviewing #1524; the generator's output, not a unit's.

Issue #1563, opened 2026-08-17
Consent & Channels shows no row for a channel a contact is actually reached on


tinnhlt-gra:
Found by UAT of the Zalo Official Account connector (#1524), and not specific to
it.

A contact created from a chat conversation has no email and no phone. The
**Consent & Channels** block on their page shows:

- Email — **Allowed**
- Phone — Unknown

and **no row for the channel they are actually reached on**, which is the only
transport the CRM can message them over.

Two things are wrong at once. Consent is asserted for a channel the person does
not have, and no consent is represented for the one in use — so the block answers
a question nobody asked and stays silent on the one that governs whether a rep may
write to them.

The binding exists: `person_channel_identity` carries `(provider,
channel_user_id)`, which is what the reply path already resolves a recipient from
and what an erasure suppresses.

PR #1590, opened 2026-08-17
fix: a chat message is not mail, and a manifest hash says which algorithm made it

### LINKED ISSUE #1562: A chat message reads as mail on the contact page: envelope icon and "Email thread"
Found by UAT of the Zalo Official Account connector (#1524), but it is not
specific to that transport — it will read the same way for any channel provider.

## What it looks like

On a contact whose conversations arrived over a chat channel, the timeline rows
are drawn with an **envelope/mail icon**, and the relationship brief labels the
conversation an **"Email thread"**. The contact in question has no email address
at all.

## Why it is the chrome and not the data

The backend is explicit that it knows the difference. The same message's "How
this message was handled" panel reports:

> **Attention label** — Skipped — *"this step reads email only, and the message
> arrived over another transport"*

So the pipeline distinguishes transports correctly and the contact-page chrome
assumes mail.

## Suggested shape

The activity carries `channel_provider`, and `GET /v1/channel-providers` already
resolves a provider key to a label — the same pair the timeline uses elsewhere to
name the transport that carried a message.

tinnhlt-gra:
Closes #1562, #1563, #1566.

Three findings from the Zalo OA connector UAT (#1524), none of them specific to
that transport.

## A chat message read as mail (#1562)

Since ADR-0107/A158 the interaction **kind** and the **transport** that carried
it are separate axes, and the person page read the second off the first
everywhere. On a contact captured over a chat channel — no address, no number —
that produced a page claiming mail three times over:

- the **timeline** keyed its icon off the display channel (the *provider id*),
  which fell through the switch to the default envelope;
- the **relationship brief** labelled every cited activity an "Email thread",
  though a citation carries a record type and an id and says nothing at all
  about a transport;
- **Today's card** drew an envelope on its activity evidence and on its reply
  verb, for the same reason.

`frontend/src/screens/interactionchrome.tsx` now owns both halves in one place:
the icon comes from the kind, the name comes from the installation's transport
directory (`GET /v1/channel-providers`), and neither is inferred from the other.
A citation the page cannot resolve is named for the one thing it certainly is —
a conversation — rather than for a transport nobody checked.

## Consent & Channels answered backwards (#1563)

The block reported a verdict for mail and phone whatever the record carried, so
a contact with neither read as consented on both, while the transport the CRM
can actually reach them on had no row at all.

Rows now state **reachability before consent**, and the channel rows come from
`person.reachability` — the record's own read of `person_channel_identity`, the
same binding the reply path resolves a recipient from. They carry the
correspondence verdict rather than one of their own because the send gate is per
**purpose**: `consent.grantedForRecipient` resolves the person and asks
`VerdictForPerson`, so the transport never enters the answer. A blocked identity
keeps its row and says it cannot be delivered to — hiding it would answer "can I
write to them" by pretending the conversation never happened.

Rendered, for a contact reached only over chat:

```
Email      No address on file
Phone      No number on file
Zalo OA    Allowed
Dispact    Not deliverable
She wrote to you on 1 Aug 2026.
```

No contract change was needed: `Person.reachability` is already on the wire.

## fragment_hash was spelled two ways (#1566)

A tool entry's `fragment_hash` was a bare hex digest while the job entry beside
it, and every `digest` in the file, carried the `sha256:` prefix. The manifest is
what an operator reads to resolve what a unit requests, and a field whose
encoding depends on which kind of entry it sits on has to be special-cased by
every reader — one that did not would compare two spellings of the same digest
and see a change that never happened. `operationHash` now hashes through
`digestBytes`, the helper the job side already used; the committed manifests are
regenerated.

## Gates

- `TestEveryManifestHashIsAlgorithmPrefixed` is a fitness function, not a case:
  it walks every committed manifest in the tree and holds every field *named*
  like a hash — anything ending in `_hash`, plus `digest` — to one encoding, so a
  field added later is covered the moment it is named like its neighbours. It
  fails when the walk finds nothing, because a walk reading no manifests looks
  exactly like a clean tree.
- `persontransport.test.tsx` (9 tests) asserts **in both directions**: a page
  that stopped drawing envelopes by drawing nothing would pass a one-directional
  suite, and a mail contact still owes the reader an envelope. Each test was
  mutation-verified against the defect it describes — reverting the icon table
  fails exactly the three chat assertions and none of the mail ones; reverting
  the rail fails exactly the three channel assertions.
- Two new stories (`RailChannelReached`, `OverviewChannelConversation`) plus a
  catalog for the chrome itself. Writing the catalog is what exposed that `task`
  rendered as the raw lowercase wire value beside six translated labels — the
  same gap as the envelope, one column over — so it is named too.

`make check` green; `make fe-uat` green (27 stories).

---
## Summary by cubic
Stops the person page from treating chat messages as mail and makes manifest hashes algorithm‑prefixed. Previously chat rows drew an envelope and “Email thread,” and the consent rail reported email/phone verdicts without addresses while omitting the channel; now icons come from interaction kind, labels come from the provider directory, and reachability precedes consent. All manifest hashes now spell as `sha256:<hex>` so readers no longer special‑case fields.

- Interaction chrome and consent rail
  - New `interactionchrome.tsx`: icon from kind, label via the directory; unknown providers fall back to id; unknown kinds render a generic record icon.
  - Timeline, brief chips, and Today use this chrome. The reply verb shows “send,” moment evidence uses a neutral record icon, and `task` now has a label.
  - Consent & Channels reads `person.reachability` for channel rows and shows the correspondence verdict. Email/phone show “No address/number on file” when missing; reached chat channels render their own rows; blocked identities show “Not deliverable.”
  - Stories catalog the vocabulary; tests assert both chat and mail cases; i18n updated.

- Manifest hash encoding
  - `operationHash` now uses `digestBytes`; regenerated manifests prefix `fragment_hash` and all digest fields with `sha256:`.
  - A root fitness test (`backend/manifestdigest_test.go`) walks committed manifests in `extensions/` and `fixtures/`, decodes JSON, enforces algorithm‑prefixed strings for any field ending in `_hash` or `digest`, requires each risk‑tier entry to carry both `fragment_hash` and `digest`, and fails empty roots; it skips `node_modules` and non‑regular files.
  - Migration: regenerate manifests so all “*_hash” and “*digest” fields are `sha256:`‑prefixed.

<sup>Written for commit 4944365f7307f0cb13499088bec9f1d34c5c2525. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1590?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/manifestdigest_test.go
+++ b/backend/manifestdigest_test.go
@@ -1,0 +1,188 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+package backendarch
+
+// Manifest-hash encoding fitness function (ADR-0069 §7): every hash a generated
+// unit manifest publishes says which algorithm produced it.
+//
+// A manifest is the record an operator reads to resolve what a unit requests,
+// and a field whose encoding depends on which kind of entry carries it is one
+// every reader has to special-case. A reader that did not would compare two
+// spellings of the same digest and see a change that never happened.
+//
+// It lives here rather than beside the generator because it reads files OUTSIDE
+// its own module, which Go's test cache cannot key: in the tools module a
+// manifest edited to the old spelling replays the previous pass and reports ok.
+// The root fitness lane runs `-count=1` for exactly this reason.
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+	"testing"
+)
+
+const unitManifestFile = "manifest.generated.json"
+
+// manifestDigestEncoding is the ONE spelling every hash in a manifest carries.
+var manifestDigestEncoding = regexp.MustCompile(`^sha256:[0-9a-f]{64}$`)
+
+// isManifestHashField reports whether a field NAME is one that carries a hash.
+// Derived from the name rather than from a maintained list, so a field added
+// later is covered the moment it is named like the ones beside it — which is
+// the point, because the encoding is a property of the FILE and not of the
+// entries a reader happens to know about. The generator already computes a tree
+// digest and a manifest digest it does not yet publish; either would arrive
+// under a name this predicate already admits.
+func isManifestHashField(name string) bool {
+	return strings.HasSuffix(name, "hash") || strings.HasSuffix(name, "digest")
+}
+
+// manifestTrees are the two roots a unit manifest can live under, and whether
+// one is guaranteed to hold a unit that requests authority. extensions/ may
+// legitimately hold none — presence under it IS enablement, so the vanilla lane
+// composes an empty tree on purpose, and `de` is a jurisdiction pack that
+// requests no risk tier at all.
+var manifestTrees = []struct {
+	root           string
+	mustHoldTiered bool
+}{
+	{root: "../extensions"},
+	{root: "../fixtures", mustHoldTiered: true},
+}
+
+func TestEveryGeneratedManifestHashNamesItsAlgorithm(t *testing.T) {
+	for _, tree := range manifestTrees {
+		var tiered int
+		for _, path := range committedManifests(t, tree.root) {
+			tiered += assertManifestHashes(t, path)
+		}
+		// A root that reads nothing passes exactly like a clean one, and an
+		// aggregate count across both roots cannot catch it: extensions/ alone
+		// would carry the total forever while a fixture tree the walk never
+		// reached escaped the rule.
+		if tree.mustHoldTiered && tiered == 0 {
+			t.Errorf("%s yielded no risk-tier entry — the walk is reading no manifest that owes a hash", tree.root)
+		}
+	}
+}
+
+// assertManifestHashes holds one committed manifest to the encoding and returns
+// how many risk-tier entries it carried.
+//
+// It DECODES rather than scanning the text, and that is the load-bearing part.
+// A key written `fragment_hash`, a value that is null or a number, and a
+// compact `":"` separator all name the same field to every real reader of this
+// file; a text scan matches none of them and skips each one in silence.
+func assertManifestHashes(t *testing.T, path string) int {
+	t.Helper()
+	content, err := os.ReadFile(path) // #nosec G304 -- path comes from walking the trusted source tree
+	if err != nil {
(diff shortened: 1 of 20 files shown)


tinnhlt-gra:
## The red on this PR is main's, not this branch's

25 checks pass. The four that do not are all pre-existing and already tracked.

**Three test failures, all failing identically on `main`:**

| Test | Tracked as |
|---|---|
| `TestSweepTargetsCarryNoDeleteBlockingTrigger` | #1588 (`priority: critical`) |
| `TestFK_rowScopedTargetsHaveVisibilityDecision` | #1588 |
| `TestADismissalHoldsUntilTheEvidenceMoves` | #1572 |

Evidence, rather than an assertion:

- The same three tests fail on `main`'s own CI at `42bde22a` (run 32034202294)
  and at `a82b1cc6` (run 32029108670), which is where
  `activity_retention_evidence` arrived.
- `main`'s integration shards report `failure` on **8 of the last 12 runs where
  they actually ran**. The runs that read green are the ones whose change
  classifier skipped the shards entirely (docs-only pushes), so a green rollup
  on `main` is not evidence the lane passes.
- This branch changes no SQL, no migration and no reset-table list. `git diff
  --name-only $(git merge-base HEAD origin/main) HEAD` is nineteen files: three
  in `backend/tools/gen-composition`, one new `backend/manifestdigest_test.go`,
  five generated manifests, and ten under `frontend/src`. The failing assertions
  are about triggers on `activity` and FKs on `activity_retention_evidence`.

**SonarCloud is downstream of the above.** The `sonarcloud` job admits `success`
OR `skipped` from its upstreams but not `failure`; the integration fan-in failed,
so the scan job skipped and the external "SonarCloud Code Analysis" check has no
analysis to report. It will clear on its own once the lane is green.

Rebased onto `origin/main@06637648` to rule out a stale base. Locally on that
base: `make check` green, `make fe-uat` green (27 stories), `craft static
--strict` clean on all four changed Go files.

This is not mergeable until #1588 and #1572 land — flagging rather than working
around it.

Issue #1561, opened 2026-08-17
Zalo OA: the pull API withholds the content of most messages, which reopens the pull-only decision


tinnhlt-gra:
Found while driving `extensions/zalo-oa` against a live Official Account (#1524).

## What was measured

A message row of `type: "nosupport"` carries **no `message`, no `url`, no
`thumb`, no `links` and no `location`** — only `src`, `time`, `sent_time`,
`message_id`, both parties' ids, avatars and display names. It is not a type the
connector failed to render; it is the API declining to hand the content over.

**Three of the four messages on a real account came back that way.** Not an edge
case — the majority.

## Why it is a decision and not a bug

The connector's design chose **pull-only**, explicitly accepting the loss of file
attachments and link unfurls in exchange for not running a webhook. That trade
was made on the understanding that ordinary messages render.

The measured loss is much larger: the **content of most ordinary messages**. A
pull-only timeline records that a conversation happened, with whom and when, and
not what was said. The connector renders such a message as "(Zalo did not provide
the content of this message)" so a reader is not misled, but that is honesty about
the gap rather than a fix for it.

The webhook is the only known way to obtain the bodies, and it was the design's
own alternative.

## What has to be decided

Whether a channel that reliably captures *who and when* but not *what* is worth
having as it stands, or whether the webhook comes back — for content, not just for
attachments. Either answer is defensible; the current state is the one that was
chosen on an assumption the evidence has since contradicted.

Evidence and provenance are recorded in the connector's working notes (GUIDE §2,
"`type: nosupport` — SETTLED"). No connector change is proposed here.

dungbanguyen:
## Closed 2026-09-01 — out of scope for this repository

This concerns a Zalo connector unit, and **that unit is not part of this product.**
Verified against `origin/main` at `d42887998` with git rather than the filesystem:

```
$ git ls-tree --name-only origin/main extensions/
extensions/.gitkeep
extensions/de
extensions/openchannel
```

Zalo appears in this tree only as an *example* of a channel provider — the
snake-versus-kebab naming rule in `compose/channelproviderfacts.go:88-89` uses
`zalo_oa` / `zalo-oa` to illustrate itself, and two contract descriptions mention
"a Telegram bot, a Zalo OA" as examples of a workspace-owned binding. There is no
connector code, and per the rulebook presence under `extensions/` **is** the
enablement — so nothing here composes it.

The units are Gradion-only, so their decisions are not this repository's to make and
[...]

PR #1591, opened 2026-08-17
fix(zalo-oa): the tick's request ceiling comes from the account, and the screen shows it

tinnhlt-gra:
## Why

`maxPagesPerPoll` fixed every installation at six pages of ten. That number was
chosen for a walk that turned out not to exist — `listrecentchat` is a
conversation index, not a global message walk — and it is the wrong *kind* of
number regardless: the right ceiling is a property of the account, and Zalo
publishes no per-account rate limit in any response header, so it can only be
chosen, never discovered.

Found while investigating #1561 against the live NFQ Official Account. The walk
itself is the next change; this one moves the ceiling onto the row and puts it in
front of the operator who has to choose it.

## What changed

- **`poll_request_budget` on `ext_zalo_oa_connection`** (migration `0003`),
  default 40, `CHECK BETWEEN 2 AND 200`. Below 2 a tick could not both read the
  account and read a page; above 200 one tick would outrun the provider's
  per-minute ceiling on its own.
- **The tick spends it**, with the account read coming off the top — a ceiling
  that ignored that call would spend one request more than the operator
  authorized. `maxPagesPerPoll` is gone; `firstPollPages` still overrides it
  downward, because connecting an account brings what arrives from now on.
- **`zaloOaStatus` returns it and the unit screen shows it first**, ahead of the
  package. It is the one number on that card that governs how much of a busy
  account each check keeps up with, and a ceiling nobody can read is a ceiling
  nobody can choose.

Live-verified on the dev installation: the existing connection took the default
without touching its credential, and the tick after the restart spent the budget
with no error class.

## Tests

- The tick reads three pages on a budget of four requests, and reads the account
  exactly once — the assertion that the ceiling is the row's and that the account
  call is paid for out of it.
- The default reads past the old six-page cap and leaves no backlog behind.
- `TestABacklogThatOutlastsTheBudgetHoldsTheFloor` now sets the budget it means to
  outlast, so it no longer depends on a constant defined elsewhere.
- Status reports the budget on the row; the screen renders it first.

## Follow-up, not in this PR

There is no settings PATCH on this unit, so the value is set by its column
default today. A per-installation editor is worth filing separately, together
with the walk fix that this budget exists to bound.

---
## Summary by cubic
Moves Zalo OA polling from a fixed six‑page cap to a per‑account request budget stored on the connection and shown on the card. The tick now spends the row’s budget (default 40, 2–200), counts the account re‑read, stops before a page it cannot finish within the job window, and uses what remains to fill any backlog.

- DB/contract: add poll_request_budget to the connection (DEFAULT 40; CHECK 2–200). Connect and status return it. A test enforces the same bounds in the migration, both contract copies, and the screen constant.
- Poller/walk: read the page budget from the row; the first poll still uses firstPollPages. The walk refuses to start a page it cannot finish before the deadline, counts requests actually spent, and hands the backlog only the unused budget. Cadence and timeout are unchanged; jobs.yaml now explains deadline‑stop.
- Frontend/surface: the card shows “Requests per check” as “n / 200” with a note that Zalo publishes no per‑account limit. Publish FactList on the extension surface and add a `.card-actions` row; both the Zalo OA and `dispact-connector` screens adopt them.
- Rollout/tests: no action required; existing connections take the default. Until a settings editor exists, change the budget in the DB if needed. Tests cover row‑budget spend (account read included), draining a 90‑message feed by default, deadline‑stop, bounds consistency, UI rendering, and the surface export pin.

<sup>Written for commit b8736f827aa2c3347be21e00691adf5970c1f6b3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1591?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tinnhlt-gra:
## UAT — run against the live NFQ Official Account

Driven on a dev stack serving this branch against the real `margince` database, so
the connection, the credential and the poll are the live ones rather than fixtures.

zalo-oa-request-budget-uat

**What the walkthrough shows**

1. The card, connected: `Requests per check — 40 / 200` as the first row, with the
   qualifier under it. The ratio rather than a bare `40`, because the number alone
   answers "how many" and not "how much headroom".
2. The fact rows aligned in two columns with rules between them, and the two verbs
   in one row with air above them and a gap between them. Before this branch the
   list rendered with the browser's `dd` indent and the buttons sat against the
   last line of text, touching each other.
3. `Replace the credentials` opens the deposit form; its own submit row is
   right-aligned with `Cancel` beside `Replace`. `Cancel` returns to the described
   state. No credential was typed and nothing was submitted.

**Verified beyond the recording**

- The existing connection took the new column's default without touching its
  sealed credential: `poll_request_budget = 40`, `status = connected`.
- The tick then ran clean against Zalo — `last_error_class` null and
  `last_polled_at` advancing — so the row's ceiling is what the poll now spends.
- One failure was found and diagnosed during UAT and is **not** a defect in this
  branch: the first run showed `The last check did not finish` because a fresh
  worktree seeds its own `.env.local`, so its keyvault root key could not decrypt
  the token sealed by the other tree's key. Aligning the key cleared it.
- The form survives a status refetch: held open past the 20s poll of
  `/ext/zalo-oa/status` without collapsing.

**Gates, local, on the rebased branch**

`check-backend` ✅ · `check-fe` ✅ · `craft-static` ✅ (0/0/0) · `test-extensions` ✅ ·
`fe-test-ext` ✅ 52/52 · `check-ext-migrations` ✅ (applies as the restricted
`ext_zalo_oa` role, reverts clean) · coverage 87.4% on the unit, with every
function touched in `walk.go` at 100%.

PR #1593, opened 2026-08-17
feat(ai): the anthropic wire carries a document, because it does

tinnhlt-gra:
## What was wrong

An operator who binds a tier to `anthropic` and hands it a scanned invoice gets `ErrAttachmentUnsupported`. The declaration behind that is `carriesImages` — `image/*` and nothing else — shared with `ollama`, with this rationale in `openaicompat.go`:

> a PDF is model-dependent on the first wire and absent from the second, so the honest declaration stops at images rather than claiming a document lane one bound model may not serve.

That was a fair reading of the Messages API once. It is not the wire today.

## What the wire actually is

Per [PDF support](https://platform.claude.com/docs/en/build-with-claude/pdf-support): **all active models** accept a `document` content block, sourced as inline base64 with `media_type: application/pdf` or as a URL the API fetches itself. Neither needs a beta header. (The `file_id` source does; that stays refused, unchanged — this adapter does not send the Files beta header, and `anthropicRefuseAttachments` already says so.)

So the caution was protecting nobody from anything, and cost the document lane on the tier most deployments bind.

## The change

- `anthropic` joins `gemini` and `openai` on `carriesImagesAndPDF`.
- A PDF maps to the block kind the API has — `document`, picked off the media type — not an image block wearing a different `media_type`. The vendor paginates a document and would refuse the same bytes announced as an image, so the block type is as much part of "carried" as the bytes are, and the test asserts the literal wire fragment for both kinds.
- `anthropicImageSource` → `anthropicSource`. Both block kinds spell the source identically (`base64` + `media_type` + `data`, or `url`); a twin struct would be an invariant to hold rather than one it documents.
- `ollama` keeps `carriesImages`, now for the reason that actually holds: its chat API has a per-message `images` array and no document part at all, so there is no lane there to give or take.

## What does not change

The narrowing rules. `input:` still only takes carriage away, so an operator keeping scanned documents off an egressing anthropic tier writes `input: [text, image]` exactly as before — and that case is now pinned in `narrowing_test.go` rather than left to the shared `narrowedCarriage` path. `pdf` remains a refused modality word.

## Spec

Contract-first (P3): `ai-operational-spec` §1.4 named Gemini and OpenAI as the native document wires. The spec change is **gradionhq/margince-foundation#1343** and lands first; this PR is its implementation follow-up.

## Gates

`make check-backend` green (exit 0), `craft static --strict` clean on the diff, full `go test ./...` green.

---
## Summary by cubic
Allows PDFs on `anthropic` by mapping `application/pdf` to the Messages API `document` block; previously `anthropic` only carried `image/*` and returned ErrAttachmentUnsupported for PDFs. This exposes the vendor-supported document lane without changing narrowing behavior.

**Review notes**
- Capability change: `SelectBrain` now advertises `image/*` and `application/pdf` for `anthropic`; `ollama` remains `image/*` only.
- Wire mapping: choose block type by MIME (`document` for PDFs, `image` for images); both use a shared `anthropicSource` (`base64` or `url`). Tests assert literal wire fragments for both kinds.
- Gates and refusals: Files beta (`file_id`/`source.type: "file"`) stay refused; https URLs are carried; non-carried types (e.g., `audio/*`) remain refused. `ollama` has no document lane.
- Required operator action (only if you relied on default refusal): to keep PDFs off an `anthropic` tier, set `input: [text, image]`.

<sup>Written for commit 985c8e7aa635c03dc9759b62e2c96325f0e4789c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1593?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Anthropic integrations now support PDF attachments alongside images.
  - PDFs can be provided through URLs or inline content and are sent using the appropriate document format.
  - Configured input restrictions now correctly preserve or exclude PDF attachments.

- **Documentation**
  - Updated configuration guidance to show Anthropic support for image and PDF inputs.

### DIFF
--- a/backend/internal/modules/ai/anthropicparts.go
+++ b/backend/internal/modules/ai/anthropicparts.go
@@ -5,7 +5,7 @@ package ai
 
 // The Messages API's message body, which is two JSON shapes wearing one name: a
 // bare string for an ordinary turn, an ordered array of typed blocks once the
-// turn carries an image.
+// turn carries an attachment.
 //
 // Both shapes are spelled here for the same reason the OpenAI-compatible wire
 // spells both (openaicompatparts.go): a text-only call must keep marshalling to
@@ -26,8 +26,13 @@ import (
 const (
 	anthropicBlockText  = "text"
 	anthropicBlockImage = "image"
-	// The two spellings an image block's source takes: bytes the request
-	// carries, or a URL the API fetches for itself.
+	// A PDF is its own block type on this wire rather than an image with a
+	// different media type: the API paginates a document and an image has no
+	// pages, so the two are not one part kind spelled twice.
+	anthropicBlockDocument = "document"
+	// The two spellings an attachment block's source takes, the same pair for
+	// either block type: bytes the request carries, or a URL the API fetches
+	// for itself.
 	anthropicSourceBase64 = "base64"
 	anthropicSourceURL    = "url"
 )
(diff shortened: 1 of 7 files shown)


tinnhlt-gra:
Blocked on #1588, not on this diff.

Integration shards 6 and 9 and the `integration (RLS + erasure + HTTP e2e)` job fail here with:

- `TestSweepTargetsCarryNoDeleteBlockingTrigger` — `activity` and `activity_retention_evidence` carry DELETE-firing triggers while listed as sweep targets
- `TestFK_rowScopedTargetsHaveVisibilityDecision`

Both are byte-identical to the failures on main's own last run that actually exercised the integration lane ([run 32034202294](https://github.com/gradionhq/margince-poc-v1/actions/runs/32034202294), `42bde22a`). Main's newest run looks green only because it was a docs-only commit and the classifier skipped the whole integration lane.

This branch is based on `06637648` and touches `backend/internal/modules/ai/` plus one docs page — nothing that reaches the retention triggers or the FK visibility registry.

Rebasing won't clear it; #1588 has to land first.

tinnhlt-gra:
@coderabbitai review

tinnhlt-gra:
@coderabbitai full review

PR #1598, opened 2026-08-17
feat(lists): who works here, what industry, what size — and a gate that keeps one page size

### LINKED ISSUE #1585: The list page-size picker offers 100, but four screens were only fixed to follow it recently — no test pins the others
## What is wrong

`ListQuery.perPage` is now the single source for both the rendered page size and the server `limit`. Six screens read it (`people`, `organizations`, `leads`, `partners`, `products`, `offertemplates`), but only the contract at the shared wrapper is tested (`listquery.test.tsx`, "hands every fetcher the page size the reader picked") plus a screen-level assertion on companies.

A seventh screen added later that keeps a literal `limit: 50` would reproduce the original bug — the screen renders 100 rows per page while the server returns 50 — and no gate would notice.

## Why it was deferred

Found during the review of #1577; the four live screens were fixed in that PR. A derived fitness test (every `useListQuery` fetcher must reference `query.perPage`) is the real fix and is bigger than the change under review.

## What it means if nobody fixes it

A new list screen silently ships the "1-25 of 50 loaded so far" mismatch, which reads to a user as a broken pager rather than as a wrong limit.

Found by Codex review on #1577.

### LINKED ISSUE #1586: listPeople still lacks organization_id, and listOrganizations lacks industry/size_band
## What is missing

`specs/architecture/data-model.md` DM-VOCAB-1 specifies `organization_id` (via employment) as a **filter** on `listPeople`. It is not in either contract.

DM-VOCAB-2 specifies `industry` and `size_band` as filters on `listOrganizations`. Both columns exist on the table; neither is a query parameter.

The spec side was reconciled in foundation#1342 — that PR closed the org half of PO-EXT-4 (lifecycle, relationship_type, include_anchor) and left these three recorded as genuinely open in PO-N-VOCAB.

## Why it was deferred

#1583 added the two owner dials a rep asks for first (`owner_team_id`, `unassigned`). These three are the same shape of work and would have widened that PR without changing what it proves.

## What it means if nobody fixes it

"Which people work at this company?" cannot be asked of the people list — only through the company 360. Industry and size are visible on a company row and cannot be filtered on, which reads as a broken column.

## The work

Contract → handler binding → store predicate → overlay forward/refuse entry → declaredfilters census self-test → HTTP + SQL integration proof → frontend chip. `organization_id` needs a link predicate over `relationship` (kind=employment), not a column.

### LINKED ISSUE #1587: The "My team" list filter is withheld when a user belongs to more than one team
## What it does today

`useOwnerChips` (`frontend/src/screens/listquery.tsx`) offers "My team" only when `/me` reports exactly one team. With two or more the option is withheld.

## Why

The wire takes one `owner_team_id`. With several teams there is no non-arbitrary single answer, and picking one for the reader answers a question they did not ask. Withholding is the honest default and the reversible one.

## What it means if nobody fixes it

A user on two teams cannot reach the team filter from the UI at all, though the API serves it. Uncommon today (the demo seed has one team per rep) but wrong for a real org.

## Options, for whoever picks this up

1. A team picker (a `dataChips` entry with the viewer's teams as options) — most honest, most UI.
2. `owner_team_id` accepting a repeated parameter, matching the union the `team` row scope already means. That is a contract change and needs the spec first.

Not a defect in what shipped; a deliberate narrowing recorded so it is not mistaken for one.

LarsGradion:
Closes #1585, closes #1586, closes #1587 — three issues filed against the list work, closed together because they are one surface.

## #1586 — the filter vocabulary the spec asked for

**`organization_id` on the people list** answers *"who works at this account"*, by the **current primary** employment edge. A past employer does not match: "who works there" and "who has ever worked there" are different questions, and handing a rep the leavers beside the staff is the wrong answer wearing the right shape.

It is an `EXISTS`, not a join — a join multiplies a person by their matching edges and the keyset cursor would page over the copies as though they were separate people.

**`industry` and `size_band` on the company list** were visible on every row and filterable by nobody, which reads as a broken column. Industry is free text and matched as written. `size_band` is checked against the contract's enum, because an unknown band would answer an **empty page**, and empty reads as "no accounts that size" rather than "that is not a size".

All three are **refused in overlay**: employment is our edge under our ids, and firmographics are ours to hold, so forwarding them would answer the unfiltered list wearing a filtered list's shape.

## #1585 — a gate so the page size cannot split again

Four screens carried a literal `limit: 50` for a release after the shared wrapper was fixed, because nothing derived the obligation from the tree.

`list-page-size-coverage.test.ts` does: **the screen list comes from the directory**, so a new list screen is covered the day it is written rather than the day somebody remembers to add it. Mutation-checked — putting the literal back in `leads.tsx` fails it.

## #1587 — the team dial names teams

It withheld itself from anyone in more than one team, leaving them unable to ask a question the API answers. Each team the viewer belongs to is now its own option, so picking one is **picking a team** rather than accepting a guess about which was meant.

## Two gates caught real things

- The **agent tool surface** publishes a per-type filter vocabulary and picked all three filters up by itself. Its test pinned whichever filter sorted *first* alphabetically, so it failed on a correct change — it now anchors on the type heading and the filters themselves.
- **golangci** caught a bare `"industry"` string beside an existing `fieldIndustry` constant.

## Proven against Postgres

| Test | Asserts |
|---|---|
| employer filter | the current employee; **not** the leaver at the same account, **not** the person elsewhere |
| industry | both software accounts |
| size band | only the small one |
| unknown band | refused, not an empty page |

**Mutation-checked**: dropping `is_current_primary` from the predicate fails the first test — the check that it tests the query rather than the response shape.

Fixtures go through the **real writers** (`CreateRelationship` for the edge, `UpdateOrganization` for the firmographics), so the rows under test are the rows production makes.

## Gates

`make check-backend` green · `make check-fe` green (2897 tests) · `make craft-static` 0 findings · integration lane green apart from `TestADismissalHoldsUntilTheEvidenceMoves`, which **fails identically on unmodified main** (verified by stashing this branch) and is unrelated — person-moment dismissals.

---
## Summary by cubic
Adds `organization_id` on people, `industry` and `size_band` on companies, enforces one page size across list screens, shows all of the viewer’s teams in owner chips, and extracts the view-highlight rule to a `useActiveView` hook. This removes unfilterable columns, prevents pager mismatches, lets multi‑team users pick the team they mean, and reduces complexity without changing behavior. Addresses #1585, #1586, #1587.

- People list: `organization_id` returns only CURRENT PRIMARY employees via an EXISTS; refused in overlay to avoid forwarding our edge.
- Company list: `industry` matches exactly; `size_band` is enum‑validated and refused if unknown; both refused in overlay.
- Page size gate: all list fetchers send `limit: query.perPage`; `frontend/src/screens/list-page-size-coverage.test.ts` auto‑discovers list screens and guards the shared fetcher shape.
- Owner chips: each of the viewer’s teams is selectable; chips sourced via `useRoster` and passed through `dataChips` on `people` and `organizations`.
- View highlight: `useActiveView` in `frontend/src/screens/listquery.tsx` moves the tab‑lighting logic out of `ListTable`; behavior unchanged and mutation‑checked.

**Review focus**
- Store predicates: employer filter in `backend/internal/modules/people/person_list.go` (`personEmployerClause`); firmographics and enum validation in `backend/internal/modules/people/organization_list.go`.
- Overlay refusals: `backend/internal/compose/overlayread.go`.
- Page-size gate and list fetcher shape: `frontend/src/screens/list-page-size-coverage.test.ts` and `frontend/src/screens/listquery.tsx`.
- Owner chips and roster sourcing: `frontend/src/screens/listquery.tsx`, `frontend/src/screens/people.tsx`, `frontend/src/screens/organizations.tsx`.
- Generated API/types: `backend/api/crm.yaml`, `backend/internal/contracts/api_gen.go`, `frontend/src/api/schema.d.ts`.

<sup>Written for commit 4156b817b9d1869559e904c4682ec588bc4f81d6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1598?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/api/crm.yaml
+++ b/backend/api/crm.yaml
@@ -683,6 +683,14 @@ paths:
           in: query
           schema: { type: string }
           description: Filter by tag name.
+        - name: organization_id
+          in: query
+          schema: { type: string, format: uuid }
+          description: |
+            People who work at this account, by their CURRENT PRIMARY employment edge
+            (`relationship` kind `employment`, DM-VOCAB-1). A past employer does not match:
+            "who works there" and "who has ever worked there" are different questions, and the
+            list answers the first.
       responses:
         '200':
           description: A page of people.
@@ -1620,6 +1628,18 @@ paths:
             `true` returns only rows with no owner. Unassigned rows are visible at every row scope
             (AAD-ROLE-2), so this names the unowned queue rather than widening what the caller sees.
             Mutually exclusive with `owner_id` and `owner_team_id`; combining them is `422`.
+        - name: industry
+          in: query
+          schema: { type: string }
+          description: |
+            Accounts in one industry, matched exactly (DM-VOCAB-2). The column is free text rather
+            than an enum, so this is the value as it was written.
+        - name: size_band
+          in: query
+          description: How many people work there (DM-VOCAB-2).
+          schema:
+            type: string
+            enum: ['1-10', '11-50', '51-200', '201-500', '501-1000', '1001-5000', '5000+']
         - name: q
           in: query
           schema: { type: string }
--- a/backend/declaredfilters_test.go
+++ b/backend/declaredfilters_test.go
@@ -278,7 +278,7 @@ func TestTheDeclaredFilterCensusReadsTheGeneratedShape(t *testing.T) {
 	}
 	// The person list declares exactly these, and the three paging components
 	// it also declares are not among them.
-	want := "ai_written,captured_by_kind,include_archived,owner_id,owner_team_id,q,tag,unassigned"
+	want := "ai_written,captured_by_kind,include_archived,organization_id,owner_id,owner_team_id,q,tag,unassigned"
 	if got := strings.Join(wire, ","); got != want {
 		t.Errorf("listPeople's narrowing parameters = %q, want %q", got, want)
 	}
--- a/backend/internal/compose/integration/declaredfilters_integration_test.go
+++ b/backend/internal/compose/integration/declaredfilters_integration_test.go
@@ -379,3 +379,93 @@ func TestTheOrganizationListNarrowsToOneTeamAndToTheUnownedQueue(t *testing.T) {
 		t.Fatalf("unassigned=true returned %d organizations, want only the unowned one", len(queue))
 	}
 }
+
+func TestThePersonListNarrowsToOneEmployer(t *testing.T) {
+	e := Setup(t)
+	acme := e.SeedOrg(t, "Acme", nil)
+	other := e.SeedOrg(t, "Other", nil)
+	staff := e.SeedPerson(t, "Works At Acme", nil)
+	leaver := e.SeedPerson(t, "Left Acme", nil)
+	elsewhere := e.SeedPerson(t, "Works Elsewhere", nil)
+
+	// Seeded through the real writer, so the edge carries whatever that writer
+	// stamps: a hand-inserted row proves nothing about the rows production makes.
+	employ := func(person ids.UUID, org ids.UUID, current bool) {
+		t.Helper()
+		personID := ids.From[ids.PersonKind](person)
+		orgID := ids.From[ids.OrganizationKind](org)
+		if _, err := e.People.CreateRelationship(e.Admin(), people.CreateRelationshipInput{
+			Kind:             "employment",
+			PersonID:         &personID,
+			OrganizationID:   &orgID,
+			IsCurrentPrimary: current,
+			Source:           "manual",
+		}); err != nil {
+			t.Fatalf("seeding the employment edge: %v", err)
+		}
+	}
+	employ(staff, acme, true)
+	// A past employer at the same account: the filter answers who works there,
+	// not who has ever worked there, and returning the leaver beside the staff
+	// is the wrong answer wearing the right shape.
+	employ(leaver, acme, false)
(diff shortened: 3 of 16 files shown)


LarsGradion:
Merge order: this goes **after** #1597. Both touch `listquery.tsx`, `people.tsx` and `organizations.tsx`, so this branch rebases once saved views land.

The three integration shards that went red earlier never ran — GitHub's `codeload` was returning 429/503 during the outage and the jobs failed in `Set up job` before `setup-go` was downloaded. Re-run and green.

Separately: three integration tests fail on unmodified `main` (issue #1599). Verified by stashing this branch and running the lane on a clean tree — `TestADismissalHoldsUntilTheEvidenceMoves`, `TestFK_rowScopedTargetsHaveVisibilityDecision` and `TestSweepTargetsCarryNoDeleteBlockingTrigger`. Not caused here, but worth knowing before reading a red lane.

Issue #1557, opened 2026-08-17
Implement A165: the retention floor covers only Handelsbriefe, and what it covers is restricted


LarsGradion:
The spec for this is complete and merged upstream. Nothing in this repo implements it yet.

**What is wrong today.** `correspondenceFloorPredicate` (`backend/internal/modules/privacy/retention_floor.go:55`) shields every activity whose `kind NOT IN ('task','note')` from Art. 17 erasure for six years, and leaves what it shields fully readable in the product. Both halves are wrong in opposite directions: §257 HGB / §147 AO oblige retention only of a *Handelsbrief* — correspondence concerning an actual commercial transaction — so scheduling mail and marketing enquiries are being over-retained; and for the records genuinely covered, the obligation justifies storage, not a sales team continuing to read the correspondence of somebody who asked to be erased.

**The spec, all merged in `margince-foundation`:**

- **ADR-0114 / A165** — the floor covers correspondence linked to a deal that is `won` or carries an offer past `draft`; the class is stamped when first earned and never re-derived; what the floor covers is restricted rather than retained; redaction is per-datum and reaches the delivery row; release and pin are audited controller decisions.
- **ADR-0115 / A166** — the GDPR chapter's absolute-erasure and retain-only guarantees, and OPS-DR-6's restore replay.
- **ADR-0116 / A167** — the DDL: the class stamp, the restriction columns and their constraints, the immutability trigger, `activity_retention_evidence`, `redacted_fields`, the `comms_outbound` DDL, and `ErrRetentionHold` in the sentinel registry.
- **ADR-0117 / A168** — the event catalog and the closed `retention.applied` action enum.

**Sequencing that matters.** Narrowing the floor makes previously-shielded correspondence erasable, and `privacy_retention` is a scheduled worker acting on the due-list unattended. Between a deploy that narrows the floor and a deploy that adds the restriction, the nightly pass would erase newly-unshielded mail with no restriction path and no way back. The narrowing and the restriction ship together, or the narrowing lands inert behind the class stamp.

**Known traps, from the spec's own consequences sections and the planning review:**
- The retention engine's ordinary selectors must exclude restricted rows, or the immutability trigger fails the nightly pass on a row it may not touch, and one stuck row stops every later scope every night.
- The trigger must `RETURN OLD` on a permitted DELETE — `NEW` is null there.
- ~33 readers of the `activity` table bypass the shared scope clause; a fitness test is the only honest way to hold that.
[...]

PR #1600, opened 2026-08-17
fix(main): the integration lane is green again — a clock that drifted, and two gates nobody had answered

tinnhlt-gra:
## Main is red on the integration lane, for two unrelated reasons

Neither is visible on `main`'s own CI yet — its runs are sitting in `status=queued` behind backed-up runners — so this surfaced on unrelated PRs instead. Every branch cut from `main` currently fails three integration tests. This clears all three.

## 1. A fixture that froze the clock and then seeded against the database's (#1572)

`TestADismissalHoldsUntilTheEvidenceMoves` went red **on a date**, with nothing changed but the date.

The suite drives its services from a frozen `roomFixedNow` of 2026-08-04 and seeded its activity at the **database's** `now() - interval '20 days'`. Those are two clocks, and they drift apart by a day for every real day that passes:

```
frozen now   = 2026-08-04 12:00
seeded row   = 2026-07-28 14:30   (database now() − 20 days, observed)
gap          = 165.5h  →  waiting = int(6.895) = 6 days
rule needs   ≥ 7       →  the gone-quiet rung never fires
```

The row was nearly 20 days old to that clock the week it was written, and **6.9 days old on 2026-08-17** — one hour under the seven-day rule. The test passes while real now ≤ 2026-08-17 and fails from then on.

**Why the failure pointed at the wrong place.** The assertion reads *"the dismissed moment came back against unchanged evidence"*, which sends a reader straight into the dismissal keying. The dismissal was working perfectly. Because `gone_quiet` had already declined, the moment being dismissed was `nothing_needed` — and `nothing_needed` is what came back. Nothing about the dismissal was wrong.

**Fixed at the invariant, not the call site.** `SeedRow` takes parameters now, so a fixture whose subject is a *time* passes it rather than writing arithmetic against a clock its service never reads; `roomAgo`/`roomAhead` derive those timestamps from the same frozen value. The `meetingbrief` suite had five more of these and is converted with it — it passes today, but its meetings were seeded fourteen days beyond the seventy-two-hour horizon the prep rung applies, so it was passing *around* its fixture rather than because of it.

**The gate is the point.** This class is invisible to review — it passes the day it is written and fails on a date nobody can connect to a commit. `TestAFrozenClockFixtureDoesNotSeedAgainstTheDatabaseClock` refuses any suite that freezes a clock and then seeds against the database's. It caught its own doc comment on the first run, and it goes red when the original seed is put back.

## 2. Two fitness gates the retention-floor schema never answered (#1588)

`TestFK_rowScopedTargetsHaveVisibilityDecision` and `TestSweepTargetsCarryNoDeleteBlockingTrigger` have been failing since `activity_retention_evidence` landed. Both were asking a question nobody had answered, and the answers differ.

**The FK gate.** Both columns are server-derived and neither has a writer — the table shipped ahead of the pass that fills it (#1557). Classified as the obligation that writer inherits, in the same form as the `finance_*` entries beside them.

**The sweep gate.** Its message was right about one table and wrong about the other:

- `activity_retention_evidence` **is** a protected store. A direct DELETE is refused outright; the row goes only with the activity it substantiates, through the CASCADE. Sweeping it directly aborts the reset on the first row. Preserved now — which means *not a target*, not *kept*: the reset still clears it via the cascade from `activity`.
- `activity` is **not**. Its trigger refuses a delete only while a row carries `restricted_at` — a statutory hold on one record. Preserving `activity` would leave every conversation behind on a reset whose entire purpose is to clear them.

So the gate now separates the two shapes: an append-only store is preserved, a row-conditional guard is classified with what the reset owes it. What it owes is real but not yet due — nothing can set `restricted_at` today, because the trigger demands evidence first and no writer exists. When #1557 lands, the reset must lift the restrictions it is entitled to clear before deleting, or it aborts on the first held row. That is written where it will go red if forgotten.

## Verification

Every gate in this PR was made to fail before being trusted:

| Gate | Mutation | Result |
|---|---|---|
| `TestAFrozenClockFixtureDoesNotSeedAgainstTheDatabaseClock` | put the original `now() - interval '20 days'` seed back | red, naming the file |
| same | (unprompted) its own doc comment quoting the pattern | red on first run — now excluded, as the one file that must name it |
| `TestSweepTargetsCarryNoDeleteBlockingTrigger` | remove the preserved entry | red, naming the evidence table |
| same | mis-key the classification | red on `activity` **and** on the stale waiver |

`make check-backend` and `make test-integration` green on this branch. The lane's zero-skip guard passes.

## Note

The `activity_retention_evidence` half of this also exists as a commit on #1589, where it was written; it is included here so **one** merge makes `main` green rather than two that depend on each other. #1589 rebases cleanly once this lands.

---
## Summary by cubic
Unblocks the integration lane by fixing a frozen-clock fixture that drifted with real time and by classifying two retention-floor gates. Tests no longer seed against the database clock when a suite freezes time; the data reset now preserves evidence and distinguishes row-conditional holds.

- Frozen-clock fixtures
  - Old: suites froze a clock but seeded with `now() ± interval`, causing date-dependent failures. New: `SeedRow` accepts extra args; tests pass timestamps derived from the same frozen clock via `roomAgo`/`roomTomorrow`.
  - Adds a gate that fails any file that both freezes a clock and uses `now() ± interval`.
  - Migration: when a test freezes time, replace `now()` arithmetic with explicit timestamps and pass them to `SeedRow`.

- Retention-floor gates
  - Sweep gate: preserves `activity_retention_evidence` in `preservedResetTables` (direct DELETE refused; cleared via cascade from `activity`) and adds `deleteGuardedSweepTargets` to record that `activity` has a row-conditional hold (`restricted_at`) and remains a sweep target.
  - FK visibility gate: adds decisions for `activity_retention_evidence.activity_id` and `.deal_id` as server-derived, writer-pending obligations.
  - Forward requirement: when the writer lands, the reset must lift eligible `restricted_at` holds before sweeping `activity`.

<sup>Written for commit 5b075a97ce4a4d71dceb323477ab79882e2c958d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1600?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/internal/compose/datareset_integration_test.go
+++ b/backend/internal/compose/datareset_integration_test.go
@@ -20,6 +20,7 @@ import (
 	"github.com/gradionhq/margince/backend/internal/platform/deployconfig"
 	"github.com/gradionhq/margince/backend/internal/platform/keyvault"
 	"github.com/gradionhq/margince/backend/internal/platform/overlaybudget/budgettest"
+	"github.com/gradionhq/margince/backend/internal/shared/gatekit"
 	"github.com/gradionhq/margince/backend/internal/shared/kernel/ids"
 	"github.com/gradionhq/margince/backend/internal/shared/kernel/principal"
 )
@@ -287,7 +288,31 @@ func TestDropResetCustomFieldColumns(t *testing.T) {
 // and is not added to preservedResetTables, it would either abort the sweep at
 // runtime or be wiped against its guard's intent. This turns that silent
 // runtime hazard into a test failure that forces a conscious classification.
+// deleteGuardedSweepTargets are the sweep targets whose DELETE-firing trigger
+// does NOT protect the table as a whole, and why sweeping them is still safe.
+//
+// The gate below cannot tell the two shapes apart from the catalog alone. An
+// append-only ledger refuses every delete, and belongs in preservedResetTables.
+// A row-conditional hold refuses deletes for SOME rows and is not a protected
+// store — preserving it would exempt a table the reset exists to clear. So the
+// second shape is classified here, with what the reset owes it.
+var deleteGuardedSweepTargets = gatekit.Waive(map[string]string{
+	// Not a protected table: the guard refuses a DELETE only while a row carries
+	// `restricted_at`, which is a statutory hold on that one record. Preserving
+	// `activity` would leave every conversation behind on a reset whose whole
+	// purpose is to clear them.
+	//
+	// Nothing can set `restricted_at` today — the trigger requires evidence in
+	// activity_retention_evidence first, and no writer for that table exists yet
+	// (#1557) — so the sweep cannot currently meet a held row. When that writer
+	// lands, the reset must LIFT the restrictions it is entitled to clear before
+	// deleting, or the sweep aborts on the first held activity. This entry is
+	// that obligation, not a record of one already met.
+	"activity": "a row-conditional statutory hold, not a protected store: preserving it would leave every activity behind on a reset meant to clear them, and no writer can set the restriction yet (#1557) — when one lands the reset must lift before it sweeps",
+})
+
 func TestSweepTargetsCarryNoDeleteBlockingTrigger(t *testing.T) {
+	defer deleteGuardedSweepTargets.AssertAllMatched(t)
 	e := integration.Setup(t)
 	ctx := e.Admin()
 	if err := database.WithWorkspaceTx(ctx, e.Pool, func(tx pgx.Tx) error {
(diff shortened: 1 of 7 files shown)


tinnhlt-gra:
Reviewed against my own independent fix for the same two defects (I had both
green locally before handing over). The approach holds up, and two things are
better than what I had: the `fixtureclock` gate catches the whole class rather
than the two files that happened to drift, and `#1557` is the right issue for
the missing evidence writer.

I verified the load-bearing premise rather than taking it on trust — the waiver
says nothing can set `restricted_at` today, and that is true: there is no Go
writer of the column anywhere in `internal/`, and the only mention in
`migrations/` is the explanatory comment in `0290`. So deferring the sweep fix
is a recorded obligation, not a shipped defect.

**One non-blocking hardening note.** The waiver is keyed on the TABLE:

```go
if targetSet[table] && !deleteGuardedSweepTargets.Waived(t, table) {
```

so the entry for `activity` ratifies the table, not the specific guard that was
judged. A second DELETE-firing trigger added to `activity` later — an
unconditional one, the shape that genuinely belongs in `preservedResetTables` —
would be waived silently by an entry written about `activity_refuse_restricted_
mutation`. Keying on `table+"."+trigger` keeps the gate sharp for exactly the
case it exists to catch, and the reason text already names the trigger it is
about.

Not worth holding main red for; a follow-up is fine.

PR #1636, opened 2026-08-18
fix: main is green again — a budget that moved, and a 404 the contract never had

### LINKED ISSUE #1634: main is red: org360 query budget exceeded, and activity link-scope read not found
## What

Two integration tests fail on a clean checkout of `origin/main` (`c2d0f4d8`), reproduced locally with no branch changes applied:

```
internal/compose/integration/org360/org360_querycount_integration_test.go:168
    one 360 issued 34 queries, budget is 32
--- FAIL: TestOrganization360CostDoesNotGrowWithTheAccount

internal/compose/integration/lifecycle_invariants_integration_test.go:217: not found
--- FAIL: TestActivityReadsAreScopedThroughLinks
```

## Reproduce

```
make db-up && make migrate
cd backend
export MARGINCE_TEST_DSN=... MARGINCE_TEST_APP_DSN=...
go test -tags integration -count=1 -run TestOrganization360CostDoesNotGrowWithTheAccount \
  ./internal/compose/integration/org360/
go test -tags integration -count=1 -run TestActivityReadsAreScopedThroughLinks \
  ./internal/compose/integration/
```

Found because they failed CI shards 4 and 6 on an unrelated config branch; confirmed against `origin/main` with nothing local.

## Why each matters

**The query budget** is a standing guard against per-record fan-out: the point of the test is that a 360 view costs the same whether an account has three related records or three thousand. Two extra queries is small in itself, but the budget exists precisely so the growth is caught while it is small.

**The link-scope read** is a tenant-isolation property — an activity is reachable only through a link the caller may already see. `not found` where the test expects a hit means either the scoping narrowed too far (a legitimate read now refused) or the fixture no longer matches how links are written. The first would be a visible bug; the second hides the gate.

I have not attributed either to a specific commit. `main` gained 34 commits in the last day, several touching the org rollup and the activity spine, so a bisect by someone who knows those areas will be quicker and safer than my guess.

dungbanguyen:
Closes #1634. Two integration tests fail on a clean checkout of `main`, for unrelated reasons. Reproduced against a freshly created database, so neither is local-state drift.

## 1. The org360 query budget — raise it, with the reason

#1621 put `contact_count` and `open_deal_count` on the companies list, and attached them to the **single** organization read as well, *"so the page a row opens into agrees with the row"*. The 360 assembles through that read, so it now issues two more queries.

Both are **flat** — one per count for the whole page, which is one organization here — and the flatness assertion immediately above the ceiling still passes. That assertion is the property this test exists to protect; only the absolute ceiling moved. `32 → 34`, with the reason recorded, as the test's own comment asks of whoever raises it.

## 2. The timeline's narrowing gate — keep the fix, correct the shape

#1604 added something real, and it stays: the activity scope is an **any-link** rule, so an activity linked to both a visible person *and* a lead the caller may not read passes it. Narrowing to that lead's id would then answer *"this lead exists, and here is what happened on it."*

What it got wrong is the shape of the refusal. It answered `ErrNotFound`, but **`listActivities` documents 200 and 403 and no 404** — so a routine filter produced a response the contract never admitted. It also broke `TestActivityReadsAreScopedThroughLinks`, which has asserted the empty-page shape since #21.

The reasoning #1604 recorded does not hold:

> Any other answer, including an empty page, tells the caller the lead is there.

An empty page is exactly what an id naming **nothing** returns, and what a **readable** lead with no activities returns. The caller cannot tell the three apart — which is what existence-hiding means. Both shapes hide existence equally; only one of them is in the contract.

So the timeline answers empty, and its test now says why. **`ListOpenTasks` keeps `ErrNotFound`** — it feeds an agent seam with no published response set, so the difference is the *surface*, not the rule. The shared gate answers the visibility question; each surface answers in the shape its own callers are promised.

**Both properties are still proven.** The timeline case keeps its activity linked through a visible person, so an implementation that skipped the gate would hand it back and the test would fail.

## A judgment call worth flagging

I am reversing a deliberate decision from #1604 on the strength of the contract and the older test. If 404 is the intended answer for a narrowed list, the fix is the other one — document it in `crm.yaml` (spec-first, per P3) and update the #21 test — and I would rather be told that than have guessed silently. Either way `main` should not stay red while it is decided.

## Also worth knowing

`main`'s own CI shows only *Publish release* failing, not the integration lane — so main has been red in a way its own pipeline did not report. That is a separate question about which lanes run on push versus PR, and I have not touched it.

## Verification

`make check` · `craft-static` (3 roots, 0 findings) · `make test-integration` — *OK: integration passed with 0 skips (41 packages)*, against a fresh database.

---
## Summary by cubic
Fixes two unrelated integration failures on main: raises the org360 query budget after roll-up columns added two queries, and makes timeline narrowing return an empty page (200) instead of 404 to match the documented contract. This restores green main without changing visibility rules.

- org360: increase the query budget from 32 to 34. Flatness is still asserted; only the ceiling moved due to the new `contact_count` and `open_deal_count` columns.
- Timeline: when the narrowing target is not visible, `listActivities` now returns an empty page. The API admits 200/403 and no 404 in `crm.yaml`. Tests updated to assert emptiness.
- Gate reuse: introduce a shared `narrowingTargetVisible` check and call it from both surfaces.
- Tasks: `ListOpenTasks` keeps `ErrNotFound` for out-of-scope targets (agent seam; no published HTTP shape).

<sup>Written for commit f4c1afc52bbb36a2af74f5d5375a52e39fb1b312. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1636?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/internal/compose/integration/org360/org360_querycount_integration_test.go
+++ b/backend/internal/compose/integration/org360/org360_querycount_integration_test.go
@@ -163,7 +163,14 @@ func TestOrganization360CostDoesNotGrowWithTheAccount(t *testing.T) {
 	// the unbounded principal — the availability test (a held record reads
 	// as gone, admin included) runs for everyone. One indexed read by primary
 	// key, flat in the size of the account.
-	const budget = 32
+	// 34 since the companies list gained its two roll-up columns (#1621): the
+	// single organization read attaches contact_count and open_deal_count so
+	// the page a row opens into agrees with the row it came from, and the 360
+	// assembles through that read. Two queries, and both are FLAT — one per
+	// count for the whole page, which is one organization here — so the
+	// property this budget protects is untouched. The flatness assertion above
+	// is what proves that, and it passes; only the ceiling moved.
+	const budget = 34
 	if smallCost > budget {
 		t.Errorf("one 360 issued %d queries, budget is %d", smallCost, budget)
 	}
--- a/backend/internal/modules/activities/activity.go
+++ b/backend/internal/modules/activities/activity.go
@@ -280,8 +280,13 @@ func ListActivitiesTx(ctx context.Context, tx pgx.Tx, in ListActivitiesInput) ([
 	// visible person and a lead the caller may not read passes it — and
 	// filtering on that lead's id would then answer "this lead exists, and here
 	// is what happened on it" to someone with no right to either fact.
-	if err := ensureNarrowingTargetVisible(ctx, tx, in.EntityType, in.EntityID); err != nil {
+	switch visible, err := narrowingTargetVisible(ctx, tx, in.EntityType, in.EntityID); {
+	case err != nil:
 		return nil, storekit.Page{}, err
+	case !visible:
+		// Nothing the caller may see is linked to it, which is what an empty
+		// page says — and all it says.
+		return nil, storekit.Page{}, nil
 	}
 	limit := storekit.ClampLimit(in.Limit)
 	join, where, args, err := listActivitiesFilter(ctx, in)
--- a/backend/internal/modules/activities/commitments.go
+++ b/backend/internal/modules/activities/commitments.go
@@ -38,6 +38,7 @@ import (
 
 	crmcontracts "github.com/gradionhq/margince/backend/internal/contracts"
 	"github.com/gradionhq/margince/backend/internal/platform/auth"
+	"github.com/gradionhq/margince/backend/internal/shared/apperrors"
 	"github.com/gradionhq/margince/backend/internal/shared/kernel/ids"
 	"github.com/gradionhq/margince/backend/internal/shared/kernel/principal"
 )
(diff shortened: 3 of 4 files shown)


dungbanguyen:
Superseded by #1633, which fixed both failures first — I was working the same two in parallel and did not see it land.

Worth recording that #1633 resolved the timeline gate the **opposite** way to this PR: it kept #1604's `ErrNotFound` and updated the older test, on the reasoning that filtering BY a record is a read OF it. That is a coherent reading and it is now the settled behaviour, so I am not re-litigating it.

One finding does survive their fix, and I have opened it separately: `listActivities` returns that 404 while the contract documents only 200/401/403. Whichever behaviour is right, the contract should say what the endpoint does.

PR #1660, opened 2026-08-18
feat(records): a person and an account remember when something last happened, and the lists sort by it

### LINKED ISSUE #1131: last_touch_at on the companies list needs the activity link-walk gate and live-employment semantics
The companies list wants a "last contact" column. A first attempt (PR #1129, closed unmerged) computed it in the people module and adversarial review found the derivation unsound in four ways. Recording them so the next attempt starts from the right shape rather than rediscovering these.

## Why it cannot live in `people`

The correct predicate already exists: `activities.OrgReachSet()` walks activity → organization through the link, the deal and the employment arms, with `ended_at IS NULL AND archived_at IS NULL`. A module never imports a sibling, so `people` cannot call it — which is precisely why the first attempt hand-rolled a second spelling of the same walk and got it wrong. **The derivation belongs in `compose`**, where `OrgReachSet` is already used (`signalscan.go`, `signalextractread.go`), or on the activities side of the seam.

## What the hand-rolled version got wrong

1. **No activity authorization.** The aggregation checked `organization:read` only. Neither `activity:read` nor `auth.ActivityScopeClause` was applied, so a caller who cannot see an owner-private employee could still learn the exact timestamp of that person's mail. Workspace binding is not the gate here — object RBAC and the link-walk row scope are.

2. **Archived and ended employment counted.** The employment arm had no `archived_at IS NULL` and did not exclude ended employments, so an ex-employee's mail would keep determining the company's last contact indefinitely.

3. **`is_current_primary` was the wrong test.** The established walk includes every live ongoing employment. A consultant employed by two companies has their correspondence on both timelines, but the column would report it for the primary employer only and show the other as never contacted.

4. **Unbounded work behind a bounded response.** One query per page, but `max()` still scans every historical activity link for every organization and employee on the page. A page of active accounts can scan a large slice of the activity graph on each list refresh. Needs either an index that serves the aggregate or a maintained column updated on the activity write.

## Contract shape

`last_touch_at` was optional and nullable, with no way to distinguish "computed, never contacted" from "not computed for this record". Overlay-mirrored companies legitimately have no value, and the UI rendered every absent value as "No contact yet" — stating as fact something the server never claimed. Either the response distinguishes the two, or the column is omitted where it cannot be computed.

## Definition (decided, keep it)

Most recent activity linked to the company **or** to anyone currently employed there, in either direction, excluding archived activities. Tasks and notes do not count: something written *about* a company is not contact *with* it.

LarsGradion:
PO-DDL-1/-4 as amended (foundation#1349): last_activity_at is STORED on
person and organization, maintained by the activity write exactly as
deal.last_activity_at is — advanceLastActivity in activities moves the
linked record's clock and every account the link reaches (its deal's
account, a live employer), the same three arms OrgReachSet walks for the
account timeline, each as a single-statement monotone max so a back-dated
capture never moves a clock backwards. The migration backfills both from the
timeline and adds the sort indexes in the ORDER BY's own shape.

DM-VOCAB-1/2 gain the sort; both lists show a Last activity column that
sorts on the server. Proven against Postgres: a note on an employed contact
moves the contact and the employer, a note on a deal moves its account, an
untouched account stays NULL and sorts last, and the test fails with the
reach update removed.

Spec: gradionhq/margince-foundation#1349 (PO-DDL-1/-4, DM-VOCAB-1/2). Closes #1131.

---
## Summary by cubic
Adds a read‑only last activity clock to `person` and `organization` and enables server‑side sorting by it in Contacts and Companies. Before, only deals tracked last activity and lists could not sort by it; now both records store `last_activity_at` (NULL until the first linked activity), and activities advance these clocks on write. Back‑dated captures never move clocks backwards, and reachable accounts update via deals and live employments.

- Reviewer notes:
  - Migration adds `last_activity_at` to `person` and `organization`, backfills from the timeline, and creates partial keyset indexes: `(last_activity_at DESC NULLS LAST, created_at DESC, id DESC) WHERE archived_at IS NULL`.
  - The activity write path now calls `advanceLastActivity`: updates the directly linked record and any reached organizations (deal’s account, current employer) using a single‑statement GREATEST to ensure monotone max under concurrency.
  - API surfaces read‑only `last_activity_at` for both models; overlay marks it native‑only. Frontend shows a “Last activity” column on People and Companies and requests `sort=last_activity_at`, triggering a fresh keyset walk.
  - New integration test proves propagation and ordering against Postgres.

- Rollout:
  - Run the DB migration.
  - Deploy backend before the frontend so `sort=last_activity_at` and the new fields are available.
  - No client input changes required; mirrors do not write this field.

<sup>Written for commit 17cbed971247431a674dfd1ebe8aef01228062a7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1660?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/api/crm.yaml
+++ b/backend/api/crm.yaml
@@ -14921,6 +14921,15 @@ components:
         captured_by: { type: string, readOnly: true, description: 'Server-stamped from the authenticated principal (human:<uuid> | agent:<id> | connector:<name>); never client-supplied.' }
         raw: { type: [object, 'null'], additionalProperties: true }
         version: { $ref: '#/components/schemas/RowVersion' }
+        last_activity_at:
+          type: [string, 'null']
+          format: date-time
+          readOnly: true
+          description: >-
+            When something last happened with this person — the newest `occurred_at` of an activity
+            linked to it, maintained on the activity write exactly as `deal.last_activity_at` is
+            (formulas-and-rules §8; a read accelerator, never a second truth — a rebuild must
+            reproduce it). NULL until the first linked activity. Sortable (DM-VOCAB-1).
         created_at: { type: string, format: date-time }
         updated_at: { type: string, format: date-time }
         archived_at: { type: [string, 'null'], format: date-time }
@@ -15161,6 +15170,15 @@ components:
         raw: { type: [object, 'null'], additionalProperties: true }
         partner: { $ref: '#/components/schemas/Partner' }
         version: { $ref: '#/components/schemas/RowVersion' }
+        last_activity_at:
+          type: [string, 'null']
+          format: date-time
+          readOnly: true
+          description: >-
+            When something last happened with this account — the newest `occurred_at` of an activity
+            linked to it, maintained on the activity write exactly as `deal.last_activity_at` is
+            (formulas-and-rules §8; a read accelerator, never a second truth — a rebuild must
+            reproduce it). NULL until the first linked activity. Sortable (DM-VOCAB-2).
         created_at: { type: string, format: date-time }
         updated_at: { type: string, format: date-time }
         archived_at: { type: [string, 'null'], format: date-time }
--- a/backend/internal/compose/integration/lastactivity_integration_test.go
+++ b/backend/internal/compose/integration/lastactivity_integration_test.go
@@ -1,0 +1,105 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+//go:build integration
+
+package integration
+
+// last_activity_at on person and organization (PO-DDL-1/-4 as amended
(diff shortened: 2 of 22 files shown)


LarsGradion:
Superseded: the review found the one-call-site writer stale on capture/relink/employment/archive paths; the replacement keeps the clocks with triggers on the writes.

PR #1664, opened 2026-08-18
feat(records): a person and an account remember when something last happened, and the lists sort by it

### LINKED ISSUE #1131: last_touch_at on the companies list needs the activity link-walk gate and live-employment semantics

LarsGradion:
PO-DDL-1/-4 as amended (foundation#1349): last_activity_at is STORED on
person and organization and sortable on both lists (DM-VOCAB-1/2).

It is kept on the writes themselves. Review of the first cut showed a clock
advanced at one Go call site (activities.insertActivityLinks) is stale on
every other path — capture files activity_link rows without it, relink moves
them, and the reach set changes with no link written at all when an
employment starts or ends, a deal moves account, or an activity is archived
or re-dated. So migration 1787028748 keeps all three clocks (deal too, which
capture had never advanced) with triggers on activity_link, activity,
relationship and deal.organization_id, each a recompute from the live
timeline (max over linked activities; an account through the same three
arms OrgReachSet walks), never an increment — idempotent, and it moves BACK
when the newest activity is archived. A clock move is not an edit:
set_updated_at_bump_version skips it under a transaction-local flag, so an
arriving mail neither rewrites updated_at nor breaks an editor's If-Match.
The Go deal writer is removed; the backfill covers all three columns.

Proven against Postgres: a note on an employed contact moves the contact and
the employer; a note on a deal moves its account; a back-dated capture never
moves a clock backwards; an employment that starts later brings the history
to the new account; archiving the newest note moves the clocks back; the
person's version is still 1 after two notes; an untouched account sorts last.
Full integration lane green (41 packages).

Spec: gradionhq/margince-foundation#1349 (PO-DDL-1/-4, DM-VOCAB-1/2). Closes #1131. Supersedes #1660.

---
## Summary by cubic
Adds a read-only last_activity_at clock to people and organizations and makes both lists sortable by it. Previously only deals tried to advance their clock in Go on log, which missed capture, relinks, and reach changes; now database triggers recompute clocks from the live timeline so they stay correct and can move back when the newest activity is archived.

- Triggers and functions: migration `1787030814` adds `last_activity_at` to `person` and `organization`, recompute helpers, and AFTER triggers on `activity_link`, `activity` (`occurred_at`, `archived_at`), `relationship` (employment), and `deal.organization_id`. All three clocks (`person`, `deal`, `organization`) are derived as a max over unarchived linked activities; idempotent, convergent, and include reach via deals and current employments.
- Version semantics: updates that only change `last_activity_at` no longer bump `updated_at` or `version` (`set_updated_at_bump_version` skips under a transaction-local flag).
- Go path changes: removed the `deal.last_activity_at` advance from `insertActivityLinks`; `LogActivity` no longer passes `occurredAt` to link insert.
- Backfill and indexes: migration backfills `person`, `deal`, and `organization` clocks and adds DESC NULLS LAST keyset indexes to support server-side sort.
- API and UI: adds read-only `last_activity_at` to Person and Organization schemas; People and Companies screens show a “Last activity” column and request `sort=last_activity_at` for server-side ordering; i18n strings and tests included.
- Resets: clock-maintenance triggers on `activity_link` and `relationship` do not block deletes; data reset notes updated.

- Rollout: apply migration `1787030814` (includes backfill and new write triggers). No client write changes required; to use the sort, pass `sort=last_activity_at` on People/Organizations list APIs.

<sup>Written for commit 2c8793c6909ba39bdd451f37a3c4754511725de2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1664?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/api/crm.yaml
+++ b/backend/api/crm.yaml
@@ -14921,6 +14921,15 @@ components:
         captured_by: { type: string, readOnly: true, description: 'Server-stamped from the authenticated principal (human:<uuid> | agent:<id> | connector:<name>); never client-supplied.' }
         raw: { type: [object, 'null'], additionalProperties: true }
         version: { $ref: '#/components/schemas/RowVersion' }
+        last_activity_at:
+          type: [string, 'null']
+          format: date-time
+          readOnly: true
+          description: >-
+            When something last happened with this person — the newest `occurred_at` of an activity
+            linked to it, maintained on the activity write exactly as `deal.last_activity_at` is
+            (formulas-and-rules §8; a read accelerator, never a second truth — a rebuild must
+            reproduce it). NULL until the first linked activity. Sortable (DM-VOCAB-1).
         created_at: { type: string, format: date-time }
         updated_at: { type: string, format: date-time }
         archived_at: { type: [string, 'null'], format: date-time }
@@ -15161,6 +15170,15 @@ components:
         raw: { type: [object, 'null'], additionalProperties: true }
         partner: { $ref: '#/components/schemas/Partner' }
         version: { $ref: '#/components/schemas/RowVersion' }
+        last_activity_at:
+          type: [string, 'null']
+          format: date-time
+          readOnly: true
+          description: >-
+            When something last happened with this account — the newest `occurred_at` of an activity
+            linked to it, maintained on the activity write exactly as `deal.last_activity_at` is
+            (formulas-and-rules §8; a read accelerator, never a second truth — a rebuild must
+            reproduce it). NULL until the first linked activity. Sortable (DM-VOCAB-2).
         created_at: { type: string, format: date-time }
         updated_at: { type: string, format: date-time }
         archived_at: { type: [string, 'null'], format: date-time }
--- a/backend/internal/compose/datareset_integration_test.go
+++ b/backend/internal/compose/datareset_integration_test.go
@@ -309,6 +309,13 @@ var deleteGuardedSweepTargets = gatekit.Waive(map[string]string{
 	// deleting, or the sweep aborts on the first held activity. This entry is
 	// that obligation, not a record of one already met.
 	"activity": "a row-conditional statutory hold, not a protected store: preserving it would leave every activity behind on a reset meant to clear them, and no writer can set the restriction yet (#1557) — when one lands the reset must lift before it sweeps",
+	// Not guards at all: migration 1787030814's last_activity_at maintenance
+	// recomputes the person/deal/organization clocks a deleted link or
+	// employment used to count toward, and refuses nothing. The sweep deletes
+	// those records too, so the recompute's work is discarded with them; it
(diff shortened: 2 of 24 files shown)


LarsGradion:
Superseded by the v3 branch (rebased onto #1666's shared record list, migration re-stamped, and the review's lock-before-derive fix). Same change.

Issue #1639, opened 2026-08-18
A contract change can strand the frontend schema: the backend merge gate enforces one of three regenerations


dungbanguyen:
Regenerating the contract is more than one command, and the backend merge gate only enforces one of them. A contract change that skips the others passes `make check-backend` completely and strands the artifacts nobody regenerated.

This is not hypothetical: it put `main` red for a day ([#1573](https://github.com/gradionhq/margince-poc-v1/issues/1573), fixed by [#1576](https://github.com/gradionhq/margince-poc-v1/pull/1576)). The gap is still open.

## What a contract change actually has to regenerate

Editing `backend/api/crm.yaml` requires all three:

| Command | Regenerates | Enforced by |
|---|---|---|
| `make gen` | `internal/contracts/api_gen.go`, `compose/stubs_gen.go`, agentpolicy, recordshapes, `pkg/extension/crm` | the `drift` gate in `make check-backend` |
| `pnpm gen:api` (in `frontend/`) | `frontend/src/api/schema.d.ts` | **only** `frontend-check`'s `fe-drift` leg |
| `go test ./internal/compose/ -run TestPublishedMCPSurface -update-mcp-info` | the published MCP surface docs | a unit test in `make check-backend` |

`make gen` has zero frontend references, so it does not touch the second. Nothing in the backend lane fails when the second is skipped.

## Why the failure is hard to read

The composed frontend lane typechecks against `build/composition/api/crm.yaml`; the vanilla lane typechecks against the committed `schema.d.ts`. When the two disagree, the error surfaces as a dozen `error TS2322` lines in screens the author never opened, naming properties that exist and parameters that are typed, all cascading from one enum differing by a few values. None of them point at the cause. In [#1573](https://github.com/gradionhq/margince-poc-v1/issues/1573) the trail ran from `customfields.tsx` and `settings.tsx` back to an audit-action contract change two PRs earlier ([#1559](https://github.com/gradionhq/margince-poc-v1/pull/1559), [#1571](https://github.com/gradionhq/margince-poc-v1/pull/1571)).

The Makefile already states the invariant that is not enforced, in `frontend-check`'s own comment:

> The `gen:api` + diff pair is the TS type-drift gate: `src/api/schema.d.ts` is generated from `crm.yaml`, and a contract change that skips regeneration would silently strand the frontend types, so regenerate and commit them together.

Correct — and the lane that would catch it is one a backend-only author has no reason to run.

## Options

Either would close it; the first is the smaller change.

[...]

Issue #1468, opened 2026-08-16
A dynamic list's filter can only be authored through the API — the Filters & views screen does not exist


dungbanguyen:
`AC-filters-and-views-1..8` describe a screen for authoring a filter — pick a
field, pick an operator, give a value, combine the leaves — and that screen does
not exist. There is no nav entry for it, and nothing in `frontend/` builds a
predicate.

The whole filter vocabulary is therefore API-only. #1286 widened it (custom
fields active *and* retired, and a tag leaf on all five record types the
`taggable` CHECK admits), and every one of those additions is reachable only by
a client that hand-writes a `segment_definition` JSON body. A human using the
product cannot build a dynamic list from **anything** — this is not a custom-field
gap, it is the same for `stage`, `owner_id` and every core field in the
vocabulary.

**What exists today.** The only frontend caller of the endpoint is
`frontend/src/screens/companyactions.tsx:266-289`, and it does one thing: find or
create a **static** list by name (`list_type: "static"`) and POST one
organization into it. `GET /lists` is called there solely to dedupe by name.
Nothing reads `segment_definition`, nothing writes one, and no screen renders a
list's membership.

**What the backend already answers**, so this is a frontend-shaped gap rather
than a spine one:

- `POST /v1/lists` validates a `segment_definition` against the live vocabulary
  and returns 422 naming the offending field when it does not typecheck.
- `GET /v1/lists/{id}/members` evaluates a dynamic list live (membership is
  never materialised, so a retired custom field keeps returning its rows).
- Filtered export resolves the same vocabulary through the same method, so what
  a filter may say, what it selects, and what an export of it contains cannot
  differ.

**Why this is filed now.** #693's closing note ticks four boxes and leaves this
one unticked, because it is the gate on the user-facing outcome: the segment
engine is complete and nobody can reach it. Worth deciding before the screen is
designed — the vocabulary is not flat. A custom field arrives as a `cf_*`
column with a per-object catalogue behind it, a tag is a correlated `EXISTS`
over a join rather than a column comparison, and the operator set that typechecks
depends on the field's type (`LVS-PARAM-1`: `eq, neq, gt, lt, gte, lte, in,
contains, exists`). A picker that offers every operator for every field will
produce 422s the user cannot interpret.

**Done when**

- [ ] A human can author a dynamic list's filter in the product: choose a field
      from the live vocabulary for that record type, an operator the field's type
      admits, and a value — with AND/OR nesting.
[...]

PR #1665, opened 2026-08-18
feat(collections): the filter vocabulary can be read instead of restated (LVS-EXT-8)

dungbanguyen:
The backend half of [#1468](https://github.com/gradionhq/margince-poc-v1/issues/1468), now that [margince-foundation#1341](https://github.com/gradionhq/margince-foundation/pull/1341) has landed LVS-EXT-8/9 upstream.

`GET /v1/segments/vocabulary?resource=<t>` answers what a filter may say about one record type. Without it, a filter builder has to keep a second copy of a vocabulary the server owns — which is the failure the closed vocabulary exists to prevent: the copy offers a field the engine refuses, and the human learns the difference as a 422 nobody could have predicted.

LVS-EXT-9 (preview) is deliberately not here. The two are independent and each is independently testable; this one is what unblocks the field picker.

### Parity is structural, not maintained

The field set is not assembled. It is the engine's own map, read through the same `SegmentEngine` call that dynamic-list validation, membership evaluation and filtered export all resolve through. So the equivalence LVS-EXT-8 is defined by — *a field this operation omits is one the engine refuses, and vice versa* — holds because there is no second list to keep in step.

The operator subset comes from a new `storekit.OperatorsFor`, an accessor over the same typed matrix `compileLeaf` refuses against. In one stable order, because the matrix is a map: without that, two identical requests answer different orders and a picker reshuffles between renders.

### Three things deliberately absent

Each because including it would have widened something narrow on purpose.

- **No label.** Labels are admin-facing catalog metadata the `fieldcatalog` seam keeps out of a filtering consumer's reach, and the custom-field catalog already serves them keyed by the same `column_name`. One join, on a surface the screen already calls.
- **No `retired` flag.** I wrote one, then removed it: deriving it needs `ActiveColumns`, and `FilterableReader` deliberately has only `FilterableColumns` because *"collections filters and never writes `cf_*` values."* The catalog's `status` answers it through the same join as `label`. Retired columns **are** listed, because the engine still accepts them so a saved segment keeps evaluating — omitting them would break the equivalence above.
- **No `x-mcp-tool`.** The spec does not ask for one, and a governed tool surface is not something to add as a side effect.

### The gates, and what each can fail on

Every list is derived; nothing restates a vocabulary.

| Gate | Catches |
|---|---|
| `compose/segmentvocabularyenums_test.go` | The contract's operator and type enums vs the engine's matrix, both directions. The handler casts engine strings straight into those enums, so this is what makes the cast safe. |
| `TestEveryReportedOperatorCompilesAndEveryOmittedOneIsRefused` | Per field, per resource — separating an operator refusal from an operand-shape one, so the assertion answers the question it asks. |
| `TestTheFilterVocabularyOverHTTPListsWhatADynamicListAccepts` | That the operation answers **200 rather than the generated 501 stub** — a module handler wins only by embedding depth, which no build can fail on — and that a listed field is one `POST /v1/lists` accepts while an unlisted one is refused 422. |
| `TestEveryResourceTheContractAdmitsHasAnEngine` | Keeps the handler's 404 branch unreachable, which is what makes it honest rather than dead. |

Mutation-verified, four ways: dropping `contains` from the contract enum, adding an operator no type admits, renaming the schema out from under the gate (it reports the rename instead of passing on an empty map), and making `OperatorsFor` over-report. Each fails its intended gate and only that one.

### Two notes for the reviewer

**The enum gate moved during review of my own work.** It started in the root fitness package, where `go-arch-lint` correctly refused it: root may depend on `contracts`, `compose` and `ai`, not on `platform`. Widening that list for a test would have traded an architectural boundary for a convenience, so it moved to compose — the one package already reading this contract and already depending on the engine. It got better on the way: root avoids parser dependencies by policy, so the first version scraped enums with regexes over YAML indentation, while compose already loads the document with a real OpenAPI loader.

**A contract/schema divergence I did not fix here.** `CustomField`'s description says `status=retired` *"hides the field from the API and filtering"*, while the engine deliberately keeps retired columns filterable so saved segments keep evaluating. LVS-EXT-8's equivalence forces this operation to list them, so the description is the half that is wrong. Contract-first, that reconciliation lands upstream rather than as a local edit to a description; the honest reading is "not offered for a new clause" rather than "not evaluable", and the schema should say so.

`make check-backend` and `make check-fe` both green. The frontend schema regeneration is in its own commit — the second of the three a contract change needs ([#1639](https://github.com/gradionhq/margince-poc-v1/issues/1639)), and `fe-drift` caught it exactly as that issue predicts.

---
## Summary by cubic
Adds GET `/v1/filters/vocabulary` (LVS-EXT-8) so builders can read the server-owned filter vocabulary per resource and avoid unpredictable 422s. Replaces the proposed `/v1/segments/vocabulary` and offers only fields a new clause may name.

- Reads the field set from the same engine map used by validation/evaluation/export; sorts by name and returns operators in a stable order via `storekit.OperatorsFor`.
- Excludes retired fields from the offerable set: custom fields use the catalog’s `ActiveColumns`; ADR-retired core fields (e.g., `organization.classification`) are omitted. Saved segments naming retired fields still evaluate.
- Returns `FilterVocabulary` with `fields` of `FilterVocabularyField { name, type (closed enum), operators (closed enum), custom }`. Labels are not included.
- Validates `resource`: invalid values return 422; a contract-admitted resource with no engine returns 404.
- Adds `storekit.OperatorsFor` and pins operator order; gates enum parity between contract and engine; integration tests verify the handler shadows the generated 501 stub and that listed fields are accepted while unlisted ones are refused.
- Updates contract (`crm.yaml`) and generated artifacts (`backend/internal/contracts/api_gen.go`, frontend `schema.d.ts`); collections now uses a catalog reader that supports both filterable and active columns.
- Migration: clients must call GET `/v1/filters/vocabulary` and use the new `FilterVocabulary` types; do not expect retired fields to appear in the picker. LVS-EXT-9 (preview) is not included.

<sup>Written for commit fc63b5cee764a691aa1f6dac16c6663eb1086610. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1665?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


dungbanguyen:
## End-of-work review round

Two reviewers over the seven hand-written files. The security pass found nothing exploitable; the craftsmanship pass found eight things, and one of them was the central design decision, wrong against a ratified AC. All applied except two recorded below.

### The one that mattered

**Retired fields must not be advertised, and I had argued the opposite into two comments.** CUSTOM-FIELDS-AC-13 makes retire "hidden from API + filtering"; AC-14 settles it in words — retired fields stay hidden from record payloads *and the sort/filter vocabulary*, and the admin catalog read is *the one surface that still shows them*. This operation is not that surface. Before this change no surface advertised the vocabulary at all, so AC-13 was satisfiable by silence; this is the first one, and it was advertising the retired set. Worse, both the contract description and the Go header stated the divergence as deliberate design, which is rule 5 — never rationalize a known gap in a comment.

Contract-first, the spec wins. The equivalence I appealed to survives once stated correctly, because it was never an equality: **everything advertised is accepted by the engine, and the gap between advertised and accepted is exactly the retired set.** A picker still cannot compose a clause the engine refuses; a saved segment on a retired field still evaluates; and an admin who retires a field to get it out of everyone's builder now gets that.

Retirement has two sources, so there are two mechanisms rather than one forced union:

- **Custom columns** — per-workspace admin state, so the offerable set is `ActiveColumns`. That needs the catalogue's other reader, and the `fieldcatalog` seam's note that *"a consumer of one has no use for the other"* no longer holds for this consumer: collections now both filters (accepting a retired column) and advertises (offering only active ones). `CatalogReader` names that at the seam instead of leaving it to a type assertion.
- **Core fields** — a decision in the file, taken by an ADR. `organization.classification` (ADR-0079/A124) has no `custom_field` row at all, so no catalogue read and no client-side join could ever discover it is retired. That also made my "both are one join away" claim false as written. `retiredCoreFields` makes it machine-readable, and a test fails if an entry names a field that is not core.

Proved end to end through the real retire operation: create → offered → `POST /custom-fields/{id}/retire` → not offered → the segment built beforehand still evaluates.

### The gate that could not fail

**The operator probe set was circular.** It derived from `OperatorsFor` — the function under test — and so did the compose enum gate's engine side. So `operatorOrder` had **zero** coverage anywhere in the tree: an operator added to `operatorsByType` but not to `operatorOrder` would have left every gate in the diff green while the engine accepted an operator the vocabulary swore it refused. Exactly the failure `OperatorsFor`'s own doc comment claims to prevent.

Fixed where the two are visible together, in storekit, as a fitness function over the matrix rather than a list. Mutation-verified: adding an operator to the matrix alone now fails with both halves of the story.

**And the loop only ever saw three of seven field types** — a nil catalogue leaves core fields, which are only `id`/`text`/`picklist`, so no ordering operator (`gt`/`gte`/`lt`/`lte`) was ever *positively* probed. It now seeds one column per `fieldcatalog.Types()` and fails if a type goes uncovered, so the shortfall cannot come back silently.

### The rest

- **404 was reachable, not unreachable as its comment claimed.** The generated wrapper binds `resource` as a plain string and never calls the `Valid()` it also generates, so `?resource=peron` reached the store and got a bare 404 saying neither what went wrong nor what to do. Now 422 naming the parameter, matching the fourteen other enum-valued query params here — which leaves the 404 for the case its comment actually describes.
- **`/segments/vocabulary` → `/filters/vocabulary`**, schemas renamed to match. A top-level family with one member and no collection behind it, while every consumer lives under `/lists` and `/views` and the spec calls the thing a filter vocabulary. LVS-EXT-9 becomes its sibling rather than a second invented root.
- **The gate's justification was wrong even though the gate was right.** I wrote that a caller who may read lists "can already infer the custom fields by building one" — that appeals to `list:create`, which `read_only` does not hold. The reason that actually holds came from the security pass: every seeded role holding `list:read` also holds `custom_field:read` (core migrations 0064, 0183, 0268), and the catalogue read that grant governs returns strictly more. Swapped.
- A length compare before indexing, so a length regression reports instead of panicking.

### Recorded, not applied

- **Folding `filtervocabulary.go` into `vocab.go`.** Two files whose names both say "vocabulary" is a fair smell, and the ~90 lines would fit under the ceiling. Keeping them separate on purpose: one *is* the engine's vocabulary, the other *reads it out loud* for a client, and they now have visibly different concerns — the read is the only half that knows about retirement, HTTP enums and offer-vs-accept. Judgment call, flagged so the next person can disagree.
- **The `custom_field:read` vs `list:read` divergence** the security pass surfaced → [#1677](https://github.com/gradionhq/margince-poc-v1/issues/1677). An operator who revokes `custom_field:read` from a role while leaving `list:read` still leaks `cf_*` names and types here. Not fixed in this PR because the point fix would be worse: gating the custom half on `custom_field:read` would start omitting fields a filter may legitimately name, breaking the subset guarantee this endpoint exists to provide. It is a decide-once question across the catalogue read, the record projections and this vocabulary, hence `status: needs-decision`. No default install is exposed — every seeded role holds both grants.

### One process note against myself

I mutation-tested uncommitted work and ran `git checkout --` on the file to undo it, which discarded the entire rework rather than just the mutation. I have a standing note to checkpoint-commit before mutating and did not follow it. Rewritten from context, verified identical by re-running the suites, and the mutations were then redone from a commit.

`make check-backend`, `make check-fe`, the collections integration package (18 tests) and `craft static --strict` all green.

Issue #1573, opened 2026-08-17
main is red on fe-typecheck-composed: the committed frontend schema was not regenerated with the audit-action contract change


dungbanguyen:
`make fe-typecheck-composed` fails on a pristine `origin/main`, so every PR inherits it and the `fe-quality` / `frontend` required checks are red for reasons unrelated to the change under review.

## Reproduced on main, not on a branch

```
$ git switch --detach origin/main      # a82b1cc6
$ make fe-typecheck-composed
src/screens/customfields.tsx(820,13): error TS2322
src/screens/settings.tsx(2367,37): error TS2322
MAIN_EXIT=2
```

Both are the same shape: an audit-log row typed by the **committed** `frontend/src/api/schema.d.ts` being passed where the **composed** schema's type is expected, and the two disagree about the `action` enum.

```
Type '"resolve"' is not assignable to type
  '"cancel" | "create" | "update" | … 18 more … | "expire"'
```

and, from the same run on a slightly older base, the count mismatch it comes from — 36 actions on one side, 33 on the other, differing by values including `pin`.

## Why the two schemas diverged

`frontend/src/api/schema.d.ts` is generated from `backend/api/crm.yaml` by `pnpm gen:api`. The composed lane instead generates from `build/composition/api/crm.yaml`, which carries the extension contributions.

Recent audit-action work landed the backend and contract halves without regenerating the committed frontend schema:

- [#1559](https://github.com/gradionhq/margince-poc-v1/pull/1559) — *the audit ledger can say a record was restricted or pinned* (adds `pin`, and the errors name it)
- [#1571](https://github.com/gradionhq/margince-poc-v1/pull/1571) — *an activity can carry the statutory class it earned*

`git show origin/main:frontend/src/api/schema.d.ts | grep -c '"resolve"'` → **0**, while the composed contract mentions `resolve` 117 times.

## Why the local gate does not catch it

`make check-fe` passes on main. The composed typecheck runs inside `fe-quality`, and its own Makefile guard says why the distinction exists:

> `fe-typecheck-composed: … || { echo "pnpm gen:composed-types produced no schema.d.ts — the composed lane would silently typecheck against the committed contract"; exit 1; }`

So the two lanes read two different contracts by design, and only the composed one sees the drift.

## The fix, and the gap behind it

Immediate: run `pnpm gen:api` and commit the regenerated `frontend/src/api/schema.d.ts`.

[...]

PR #1684, opened 2026-08-18
fix(leads): unbreak main — classify the lead dedupe FKs and stop comparing a µs column to a ns clock

dungbanguyen:
The `integration (RLS + erasure + HTTP e2e)` required check is failing on `main`, and every PR opened since inherits it — PR CI tests the merge with the base tip, so a branch that touches no lead code fails on lead tests.

## How it got in

[#1662](https://github.com/gradionhq/margince-poc-v1/pull/1662) *(leads dedupe against leads)* merged at 05:22Z with its own `integration (RLS + erasure + HTTP e2e)` check at **FAILURE** — `integration shard (1/12)` was red at merge time.

`main`'s runs since have been green only because they were docs-only changes whose classifier skips the integration lane, so nothing has re-tested the schema since. The red was never visible on a `main` run.

This is the **third** instance recorded today of a PR merging over a red required check, after [#1559](https://github.com/gradionhq/margince-poc-v1/pull/1559) (`fe-quality` FAILURE → [#1573](https://github.com/gradionhq/margince-poc-v1/issues/1573)) and #1512's migration-version collision. The pattern is now the finding: the gates are doing their job and are being bypassed, which makes every green verdict on this repo unreadable until it stops. Whatever mechanism allows the bypass — admin merge, auto-merge racing a check, or `enforce_admins` being off since 2026-08-09 — is worth closing separately from these two defects.

## What is actually broken

**1. Three foreign keys with no visibility decision.** `TestFK_rowScopedTargetsHaveVisibilityDecision` refuses `dedupe_candidate.left_lead_id`, `dedupe_candidate.right_lead_id`, and `lead.merged_into_id`. The gate is right to ask — a reference to a row-scoped record is a read of it — and the answer is that all three are server-derived, which the fix records with the reasoning verified rather than assumed:

- the pair ids go through the same `recordDedupeCandidate` INSERT as their person/org siblings (`entityLead` only picks a different column pair);
- `lead.merged_into_id` is written by `archiveMergedAway`, which takes the table as a parameter — the same statement that stamps person and organization.

**2. A test comparing a microsecond column to a nanosecond clock.** `TestFirstResponseKeepsTheEarliestReplyWhateverTheDeliveryOrder` writes a `time.Now()` value to a `timestamptz`, reads it back, and asserts `.Equal()`. Postgres stores microseconds, so the nanosecond remainder is lost in the round trip and the assertion fails on a value that stored correctly:

```
first_response_at = 2026-08-18 05:17:12.33026    +0000 UTC
want the earliest reply 2026-08-18 05:17:12.330260974 +0000 UTC
```

**This is why it passed review and failed CI**: macOS hands out a coarser clock than Linux, so the remainder is usually zero locally and almost never zero on a runner. The fix truncates at the source so every derived time round-trips losslessly, rather than loosening the assertion — which would have stopped proving the delivery-ordering the test exists for. Reproduced before fixing by injecting the same 974ns remainder.

The sibling comparison four lines above it is safe and untouched: its expected value comes back *out* of the database, so it is already microsecond-exact.

## Worth a follow-up

The precision trap is not specific to leads — any integration test that writes a Go time and later compares it exactly has it, and it fails only on Linux. A fitness check, or a shared `testtime.Now()` that truncates, would close the class rather than this one instance. Filing that separately if there is appetite; it is not needed to unbreak `main`.

### DIFF
--- a/backend/internal/modules/people/dedupequeue_integration_test.go
+++ b/backend/internal/modules/people/dedupequeue_integration_test.go
@@ -316,3 +316,58 @@ func TestDedupeQueueHidesPairsOutsideTheCallersRowScope(t *testing.T) {
 		t.Fatalf("owner sees %d candidates, want 1", len(rows))
 	}
 }
+
+// The same gate, for leads, and it carries more weight here than the person case
+// above.
+//
+// The schema fitness suite classifies dedupe_candidate's lead pair ids as
+// server-derived rather than requiring auth.EnsureLinkTarget on them
+// (migrations/schema_fitness_integration_test.go). That classification is only
+// safe because THIS gate holds: the detector is deliberately not row-scoped, so a
+// lead pair can legitimately name a lead the reader cannot see, and the read is
+// the only thing standing between that pair and disclosure. An untested security
+// gate looks exactly like a passing one, and until now this one was tested for
+// entity_type='person' only.
+func TestDedupeQueueHidesLeadPairsOutsideTheCallersRowScope(t *testing.T) {
+	e := setupDedupe(t)
+	ctx := e.as()
+	first := e.createLead(ctx, t, "Jonas Petersen", "jonas@nordwind.test", "Nordwind Logistik")
+	e.createLead(ctx, t, "Jonas Peterson", "", "Nordwind Logistik")
+	rows := openCandidates(ctx, t, e, entityLead)
+	if len(rows) != 1 {
+		t.Fatalf("open lead queue holds %d candidates, want 1", len(rows))
+	}
+	c := rows[0]
+
+	// Bind both leads to one rep so the rows are private rather than
+	// workspace-shared, which is the state a row-scope clause can act on.
+	if err := e.store.tx(ctx, func(tx pgx.Tx) error {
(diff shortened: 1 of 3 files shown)


tinnhlt-gra:
Duplicate of #1685 — same two files, same two fixes, opened 35 seconds apart. Keeping #1685 so only one of them lands.

The one place they differ is the `lead.merged_into_id` entry. This PR classifies it *server-derived: stamped by MergeLead through archiveMergedAway*, matching the wording of its `person`/`organization` siblings. But the survivor id is not server-derived: `MergeLead` takes `targetID` from the caller and `mergePair` (`internal/modules/people/mergerelink.go:147`) admits it with `auth.WritableBy` — a row-scope **write** probe — before `archiveMergedAway` stamps the pointer. #1685 says exactly that (*gated then stamped*), which is what the fitness test is asking the author to write down.

Verified on a clean `main` with #1685's patch applied: `TestFK_rowScopedTargetsHaveVisibilityDecision` and `TestFirstResponseKeepsTheEarliestReplyWhateverTheDeliveryOrder` both pass.

Issue #613, opened 2026-08-08
company-context.test.tsx: "names the field each change checkbox selects" is flaky


tinnhlt-gra:
## What happens

`src/screens/company-context.test.tsx > CompanyContextCard refresh review > names the field
each change checkbox selects` fails intermittently with:

```
TestingLibraryElementError: Unable to find role="heading" and name "Review what changed"
```

It passes when the file is run alone, and passes on a re-run of the full suite. In CI it has
taken ~10s before failing, which reads as a timeout waiting for the heading rather than a
wrong assertion.

## Why it is worth a ticket rather than a re-run

Seen on **two unrelated PRs** now — #598 (backend tool descriptions) and #610 (the deal
reopen tier fix) — neither of which touches `frontend/` at all. #610's diff contains zero
frontend files. So it is not a change breaking it; it is the test.

Each occurrence costs a CI cycle and, worse, teaches the next person that a red frontend
gate might be nothing. A flaky gate is one nobody reads.

## Where to look

The heading appears after the refresh-review state resolves, so the likely cause is the test
asserting on it before whatever it awaits has settled — a `findBy*` where a `waitFor` over
the settled condition is wanted, or a query that races the card's own async load. The three
sibling cases in the same describe block do not flake, which should narrow it.

Filed rather than fixed here because #610 is a governance fix on the backend agent surface
and has no business editing a frontend test it did not break.

tinnhlt-gra:
Carried across from #533 and #570, both closed as duplicates. All three report the same symptom in `frontend/src/screens/company-context.test.tsx`: `TestingLibraryElementError: Unable to find role="heading" and name "Review what changed"`.

**The lead, from #570 — the wait is already generous.** `renderReview()` uses `findByRole` with `SETTLE_MS = 10_000`, then a `waitFor` on the checkbox count. So on a loaded runner the review panel is **not rendering within ten seconds** of the Refresh click. That points at the fixtures site-read resolution being starved, not at an assertion written too eagerly — which changes where to look.

**It is not one test (#570).** Two failures on PR #514 on two different heads hit *different* tests in the file: `offers a checkbox only for a change that can be selected` and `names the field each change checkbox selects`. So the flake is the files fixture, not a single case.

**It is not the runner alone (#570).** In one CI job the same suite passed on an earlier vitest invocation and failed on a later one *within that same job*.

[...]

PR #1706, opened 2026-08-18
fix(ci): the per-test ceiling clears the waits a test is allowed to spend

### LINKED ISSUE #1144: Frontend screen tests flake on a 5s timeout under load
## What happens

Six frontend screen-test files fail intermittently with `Test timed out in 5000ms`
— vitest's default `testTimeout`, which nothing in the frontend config overrides.
They pass when the suite runs on its own and fail when the machine is busy, so the
same commit produces a different verdict run to run.

This is a test defect, not a product defect: no assertion is wrong, and each of
these tests passes reliably in isolation.

## How to reproduce

```
make check      # runs check-backend, then check-fe -> fe-unit
```

Repeat it. The failures are load-dependent, so it takes more than one run — and
which tests fail changes each time, which is the tell.

Observed across three runs of the **same commit**:

| Run | Result | Tests that failed |
| --- | --- | --- |
| 1 | 11 failed / 4 files | `onboarding-facts`, `company-act`, `onboarding-conversation`, `onboarding` |
| 2 | 3 failed / 2 files | `logactivity`, `tasks` |
| 3 | green | — |

`make fe-unit` on its own passed every time (197 files, 2447 tests). Two runs
failing **disjoint** sets rules out a regression — a real break fails the same
tests twice.

## The full set seen failing

- `src/screens/onboarding-facts.test.tsx` — "refuses a fact past the ceiling on
  the review card, and says why"; "takes the refusal back on the review card once
  a fact is cleared"; "refuses a fact past the ceiling on the edit form, and says why"
- `src/screens/onboarding-conversation/company-act.test.tsx` — "sorts the blocking
  rows before decisions…"; "folds a still-open clarify decision into the same
  list…"; "blocks a retry until the refetched proposal actually changed…"; "stays
  blocked on the no-data path too…"
- `src/screens/onboarding-conversation/onboarding-conversation.test.tsx` — "folds
  poll-delta narration into an activity group…"; "lets a human fill a missing
  required field right in the thread, enabling Continue"; "confirms with the
  proposal's draft_version and proposal_hash"
- `src/screens/onboarding.test.tsx` — "preserves administrator typing when a newer
  streamed dossier arrives"
- `src/screens/logactivity.test.tsx` — "posts a note linked to the viewed person
  and refetches the timeline"; "keeps the due-date input's space for a note and
  enables it for a task"
- `src/screens/tasks.test.tsx` — "creating a task posts kind=task with the picked
  due/remind instants"

## Why they are slow enough to time out

Two things compound, and neither is a sleep — which is why the craftsmanship gate
does not catch it (`craft static` sweeps the Go trees only, so no frontend test is
in scope for T11 at all today):

1. **`userEvent.setup()` advances on real timers.** Every simulated keystroke and
   click awaits a real macrotask, so a test's wall-clock cost scales with its
   interaction count. The failing tests are the interaction-heaviest ones.
2. **`userEvent.setup()` is called per interaction rather than once per test.**
   e.g. `logactivity.test.tsx:202` and `:214` each construct a fresh instance;
   the same shape recurs in the other files. That multiplies the fixed setup cost
   by the number of interactions.

Under contention the total crosses 5000ms and vitest kills the test. Nothing about
the assertions is timing-dependent — only the driving of the UI is.

## Why it matters

- **CI re-runs read as flaky infrastructure, and train people to re-run red.**
  Once a suite is known to fail randomly, a genuine regression in these files gets
  dismissed as "just the flake" — the failure mode that makes a flaky test worse
  than a missing one.
- **It cost real time here.** A dependency-bump PR (#1139) went red twice on these
  tests, and ruling the change out took a `main` baseline, a third full-gate run,
  and a `pnpm why` trace to prove no causal path existed. That work recurs for
  every unrelated change unlucky enough to hit it.
- **CI hardware is slower than a dev machine**, so the margin there is thinner
  than these local numbers suggest.

## Directions, not a prescription

Whoever owns these screens should pick; each has a different cost:

- Hoist one `userEvent.setup()` per test (cheap, mechanical, no behaviour change).
- Drive interactions on fake timers via `userEvent.setup({ advanceTimers })` so the
  cost stops being wall-clock (correct, more invasive).
- Raise `testTimeout` in the vitest config — honest as a stopgap, but it hides the
  slowness rather than fixing it, and the next slower test re-opens this issue.
- Split the largest files while in there: `onboarding-facts.test.tsx` (1728 lines),
  `company-act.test.tsx` (1327), `onboarding-conversation.test.tsx` (1039). Well
  past the 1000-line ceiling the Go test trees hold, and a suite splits when it
  stops being navigable.

Worth considering separately: the craftsmanship gate has no frontend arm, so T11
(`tests prove behaviour or they are noise` — no real-clock flakiness) is currently
unenforced for every `*.test.tsx` in the tree.

tinnhlt-gra:
## The judgement call, made on measurement

#1144's own ⚠️ says not to raise a timeout before establishing whether the tests
are **slow** or **starved**, because sibling #613 looks identical from outside and
is starvation. Measured under deliberate load (8 spin loops on 10 cores) with the
timeout lifted to 120s:

| test | duration under load |
|---|---|
| `company-act.test.tsx` "re-arms Continue once a skew refetch actually lands a NEW hash" | **609 ms** |
| `onboarding-conversation.test.tsx` "folds poll-delta narration into an activity group" | **910 ms** |
| slowest test in the entire suite (`read-conclusion.test.tsx`) | **3437 ms** |

Both tests I watched fail at a 5000 ms ceiling finish in well under a second.
Nothing is stuck, so it is **not starved**. Nothing is near five seconds of work
either, so "legitimately slow under load" does not describe it — there is no
smooth degradation for a bigger number to absorb.

## It is arithmetic between two budgets nobody compared

Testing Library's `asyncUtilTimeout` and `vi.waitFor`'s timeout both default to
**1000 ms**, and this repo overrides neither (no `configure()` anywhere). So a
test built from N sequential awaited waits may legitimately spend **N seconds**
without any single wait failing. `company-act.test.tsx:845` chains **six**.

Against a 5000 ms per-test ceiling, that test fails while every assertion in it
is passing — and the failure names the test rather than the wait that was slow.
25 tests in the suite carry a chain of ≥5. This is the "total that contradicts
its parts" shape.

## The repo had already found this, in one file

`src/screens/company-context.test.tsx` pins its own `TEST_MS = SETTLE_MS * 4`
and spells out the invariant:

> it must cover EVERY waiter in there: four run in sequence, each bounded by
> SETTLE_MS. A test whose own limit is smaller than the sum lets vitest fire
> while a waiter still has budget, and what surfaces then is an opaque timeout
> rather than the assertion the test was written to make. Vitest's per-test
> default is 5s, which is less than one of these waiters alone.

That is this defect, already diagnosed, fixed for exactly one file. #1144 is the
same arithmetic everywhere nobody applied the override — which is why this is
**one config-level authority** rather than six more per-file constants that drift.

## The value, and why it is not round

```
TEST_TIMEOUT_MS = LONGEST_WAIT_CHAIN × ASYNC_UTIL_TIMEOUT_MS + SLOWEST_MEASURED_TEST_MS
                = 6 × 1000 + 3437
                = 9437
```

Each term is a measurement carried in `frontend/vitest.budget.ts` with how it was
taken, so the ceiling moves when they do instead of being re-guessed. The second
term uses the slowest measured *whole test* as the allowance for the work
*between* the waits, which deliberately over-counts — that figure already contains
its own waits — because a ceiling too high costs a slower red and one too low
costs this bug.

**It cannot mask #613.** 9437 ms is deliberately *below* `company-context.test.tsx`'s
own 10 000 ms-per-waiter budget, and that file keeps its own numbers and is
untouched here, exactly as the issue asks.

## The guard

Nothing in vitest compares the per-test ceiling with the per-wait one — which is
how one drifted under the other and stayed there. `src/test-budget.test.ts` gates
the relationship. Against the 5000 ms default it goes red on all three
assertions:

```
× clears the longest chain of waits a single test may legitimately compose
  → expected 5000 to be greater than 6000
× leaves room for the work between those waits, not only the waiting
  → expected -1000 to be greater than or equal to 3437
× is the sum of its two measurements and nothing else
  → expected 5000 to be 9437
```

Arithmetic on constants rather than a timed run, on purpose: a guard that proves
a timeout by actually waiting six seconds costs six seconds on every run and
still only proves it for the machine it ran on.

**What it does not catch, stated rather than implied:** it gates the *relationship*
between the constants, so someone could still fudge a measurement and land on a
round number. That is a review matter; no test can tell a real measurement from
an invented one.

## Reproduction — the honest version

I **observed** the flake twice today, organically, both in this session's transcript:

- one run: `company-act` + `onboarding-conversation`, both `Test timed out in 5000ms`
- the next run of the same commit: a different file failed
- the third: fully green

Disjoint failure sets on one commit — the #1144 signature, and what rules out a
regression.

But I could **not** reproduce it on demand. Twelve deliberately loaded runs were
all green: 8 spinners, 20 spinners on 10 cores, a continuous `go build`/`go vet`
co-tenant, and two full vitest suites concurrently. **So the before/after
repetition counts cannot carry this argument and I am not claiming they do.**
Post-fix, four full runs under the same 20-spinner load are green (3034/3034
each), which shows no regression — not that the flake is gone.

What carries the argument is the structural proof above, which is deterministic
and does not depend on catching the machine in the right mood.

## Coverage

Test-config change plus one test of the config: **no new product code**, so
new-code coverage does not apply. `frontend/tsconfig.node.json` gains
`vitest.budget.ts` because `vite.config.ts` and the new test both read it and
they must share a program.

## Review

- `/code-review` (Fable) — run before merge.
- **Codex has not reviewed this**: `/codex:*` cannot be invoked by an agent. Two
  reviewers, not three, unless a human types it.

Claims worth attacking:
- *"the longest chain is 6"* — counted with a brace-matching script that
  over-counts on long files; the six-chain cases were hand-verified, the 29/39
  figures it also produced were not, and are not relied on. If a real chain is
  longer than 6, the ceiling is too low and the guard will not say so.
- *"it cannot mask #613"* — rests on that file keeping its own overrides.
- *"not starved"* — established from two tests, not all six.

Closes #1144

---
## Summary by cubic
Prevents flaky “Test timed out in 5000ms” by setting a measured per-test timeout in `vitest` and guarding it. Old: suite used the 5000ms default, which could be lower than the sum of sequential waits; New: `testTimeout` is 9437ms (= 6 × 1000ms wait chain + 3437ms slowest test), with a guard test enforcing this. Per-wait timeouts are unchanged, and `company-context.test.tsx` keeps its overrides so this cannot mask #613.

**Changes and review focus**
- Add `frontend/vitest.budget.ts` exporting `ASYNC_UTIL_TIMEOUT_MS`, `LONGEST_WAIT_CHAIN`, `SLOWEST_MEASURED_TEST_MS`, and `TEST_TIMEOUT_MS`; import in `frontend/vite.config.ts` to set `testTimeout`.
- Add `frontend/src/test-budget.test.ts` to assert the timeout clears the longest wait chain, leaves headroom, and equals the documented sum; it fails under the old 5000ms.
- Update `frontend/tsconfig.node.json` to include `vitest.budget.ts`.
- Reviewer: sanity-check `LONGEST_WAIT_CHAIN = 6` and `SLOWEST_MEASURED_TEST_MS = 3437`; confirm `src/screens/company-context.test.tsx` remains untouched. No migrations required.

<sup>Written for commit 3faf238dfb8349feb6a1934b3db36442441e4b03. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1706?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/test-budget.test.ts
+++ b/frontend/src/test-budget.test.ts
@@ -1,0 +1,45 @@
+import { describe, expect, it } from "vitest";
+import {
+  ASYNC_UTIL_TIMEOUT_MS,
+  LONGEST_WAIT_CHAIN,
+  SLOWEST_MEASURED_TEST_MS,
+  TEST_TIMEOUT_MS,
+} from "../vitest.budget";
+
+// The budget that governs every other suite here, gated the only way a budget
+// can be: against the thing it has to outlast. #1144 was not a slow test — it
+// was a per-test ceiling smaller than the sum of the per-wait ceilings the test
+// was allowed to spend, so tests failed with every assertion in them passing.
+// Nothing in vitest compares those two numbers, which is why one drifted under
+// the other unnoticed for as long as it did.
+//
+// This is arithmetic on constants rather than a timed run on purpose: a guard
+// that proves a timeout by actually waiting six seconds costs six seconds on
+// every run and still only proves it for the machine it ran on.
+describe("the frontend suite's time budget", () => {
+  it("clears the longest chain of waits a single test may legitimately compose", () => {
+    // Below this the failure mode is not a slow test but a wrong verdict: the
+    // waits are each inside their own budget and the test is outside its own.
+    const chainBudget = LONGEST_WAIT_CHAIN * ASYNC_UTIL_TIMEOUT_MS;
+    expect(TEST_TIMEOUT_MS).toBeGreaterThan(chainBudget);
+    // The default this replaced could not, which is the bug in one line.
+    expect(chainBudget).toBeGreaterThan(5_000);
+  });
+
+  it("leaves room for the work between those waits, not only the waiting", () => {
+    // A ceiling that cleared the chain by a millisecond would fail the moment a
+    // render between two waits cost anything at all.
+    const headroom =
+      TEST_TIMEOUT_MS - LONGEST_WAIT_CHAIN * ASYNC_UTIL_TIMEOUT_MS;
+    expect(headroom).toBeGreaterThanOrEqual(SLOWEST_MEASURED_TEST_MS);
+  });
+
+  it("is the sum of its two measurements and nothing else", () => {
+    // The value carries its derivation or it is a number somebody liked. If a
+    // longer chain or a slower test is measured, the ceiling moves with it
+    // rather than being re-guessed.
+    expect(TEST_TIMEOUT_MS).toBe(
+      LONGEST_WAIT_CHAIN * ASYNC_UTIL_TIMEOUT_MS + SLOWEST_MEASURED_TEST_MS,
+    );
+  });
+});
--- a/frontend/tsconfig.node.json
+++ b/frontend/tsconfig.node.json
@@ -29,7 +29,17 @@
       "@composition/copy": ["./src/composition/extlocales.gen"]
     }
   },
-  "include": ["vite.config.ts", "vite.mcp-apps.config.ts", "scripts/**/*.ts", "src"],
+  // vitest.budget.ts sits beside the vite configs because it is config: the
+  // per-test time ceiling, derived from measurements it carries. vite.config.ts
+  // reads it and so does src/test-budget.test.ts, so it has to be in the same
+  // program as both.
+  "include": [
+    "vite.config.ts",
+    "vite.mcp-apps.config.ts",
+    "vitest.budget.ts",
+    "scripts/**/*.ts",
+    "src"
+  ],
   // src/screens/ext/ is the COMPOSED lane's program — its screens call routes
   // that are in `paths` only under the merged contract — so it cannot be typed
   // here any more than it can in the app project, and `@composition/screens`
--- a/frontend/vite.config.ts
+++ b/frontend/vite.config.ts
@@ -5,6 +5,7 @@ import tailwindcss from "@tailwindcss/vite";
 import react from "@vitejs/plugin-react";
 import { defineConfig } from "vitest/config";
 import { serveMcpApps } from "./scripts/vite-inline-views";
+import { TEST_TIMEOUT_MS } from "./vitest.budget";
 
 // The composition alias — the runtime half of the two-lane type story whose
 // compile-time half is tsconfig.app.json / tsconfig.composed.json.
@@ -149,6 +150,11 @@ export default defineConfig({
   },
   test: {
     environment: "node",
+    // Derived in vitest.budget.ts, which carries the measurement and the
+    // arithmetic. The short version: a test of N sequential waits may spend N
+    // seconds without a single wait failing, and the longest chain here is six —
+    // so a five-second ceiling failed tests whose every assertion passed.
+    testTimeout: TEST_TIMEOUT_MS,
     // Rebinds jsdom's Web Storage over the Node ≥23 global stub — see the file.
     setupFiles: ["./vitest.setup.ts"],
     // Playwright owns e2e/ — vitest must not collect its specs
--- a/frontend/vitest.budget.ts
+++ b/frontend/vitest.budget.ts
@@ -1,0 +1,71 @@
+// The frontend suite's time budget, derived rather than picked.
+//
+// Six screen suites failed intermittently with `Test timed out in 5000ms` —
+// vitest's default `testTimeout`, which nothing overrode. They passed in
+// isolation and failed when the machine was busy, so the same commit produced a
+// different verdict run to run (#1144).
+//
+// It is NOT starvation, and it is not vague slowness either. Measured under
+// deliberate load (8 spin loops on 10 cores) with the timeout lifted, the two
+// tests observed failing completed in 609ms and 910ms, and the slowest test in
+// the whole suite took 3437ms. Nothing is stuck, and nothing is close to five
+// seconds of work.
+//
+// What is actually wrong is arithmetic between two budgets that were never
+// compared. Testing Library's `asyncUtilTimeout` and `vi.waitFor`'s timeout both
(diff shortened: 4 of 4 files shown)


tinnhlt-gra:
Heads-up for anyone reading this in `main`: **this PR auto-merged on its first commit while `/code-review` was still running**, so the review's findings landed after the squash and could not go on this PR. They are real and four of them matter — the ceiling here does **not** establish the invariant its comment claims, and the guard beside it cannot fail.

Corrected in **#1710**:

- the 9437ms ceiling still sits *under* the waiter budget of `integrations-provider.test.tsx` (10000), `voice-act.test.tsx:487` (12000) and `company-context.test.tsx:416/:433` (30000) — the chain model assumed every waiter takes the 1000ms default, and those raise their own;
- `src/test-budget.test.ts`'s assertions are algebraic identities over constants defined in terms of each other, so they cannot report the drift they exist to catch;
- nothing binds the guard to `vite.config.ts`'s actual `testTimeout` — deleting the import reintroduces #1144 with the suite still green.

#1710 replaces the hand count with a TypeScript-compiler read of every test's waiter budget against its own ceiling, and it reproduces exactly the cases review found by hand.

PR #1715, opened 2026-08-18
fix(ci): the reader reports what it cannot read, and is tested itself

tinnhlt-gra:
Third round on issue 1144, following #1706 and #1710 — **both of which auto-merged before their `/code-review` finished**, so each round's findings had to land as the next PR. I have not armed auto-merge here; this one waits for review.

Review of #1710 found ten things. Two were HIGH and one of those was a **live violation the guard walked straight past**.

## The reader was assuming the friendliest value

`onboarding.test.tsx:400` waits `{ timeout: READ_POLL_MS * EXPECTED_READS * 5 }` — arithmetic over a constant imported from the hook it drives. The reader could fold neither, so it recorded the *default* 1000 ms where the test spends 8000, and its real 11 000 ms budget read as 4000 and passed under a 10 437 ms ceiling.

That is the same shape as the bug being fixed, one level up: a number nobody could read, quietly replaced by the smallest one that keeps the gate green.

Now the reader folds arithmetic, follows a named import one hop, and — the part that matters — **reports what it still cannot fold**. An unreadable budget fails the guard rather than defaulting:

```
✗ could read every budget it was asked to judge
  → src/screens/integrations-provider.test.tsx:176 states a timeout this reader cannot fold: RENDER_TEST_MS
```

`onboarding.test.tsx`'s dossier case now states its own ceiling, derived from the same terms.

## Four test shapes were invisible

| shape | was | now |
|---|---|---|
| `it.each(cases)(…)`, `it.only`, `it.concurrent` | dropped — callee matched at the immediate name, not the root of the call chain. **17 call sites** under no guard | recorded |
| `it(name, { timeout }, fn)` | dropped entirely, ceiling and body both | recorded |
| `const settle = () => waitFor(cb, { timeout: 9000 })` | counted **0** — the reader descended past the node it was handed, and for a concise arrow that node *is* the waiter | counted 9000 |
| a helper awaited four times | counted **once** — the recursion guard doubled as a de-duplicator | counted four times |

Recovered records: 2919 → 2971.

## The part that was missing rather than wrong

`scripts/test-budget.reader.test.ts` is new. `test-budget.test.ts` holds the **tree** against the reader; nothing held the **reader** against anything — so a regression that blinded it was invisible unless some test in the tree happened to exercise the shape *and* exceed its ceiling. Both reader mutations I tried against #1710's guard passed for exactly that reason.

Eleven cases, each a hole the reader actually had. Mutation-tested:

| Mutant | Result |
|---|---|
| reader stops visiting the node it was handed | **red** — the arrow-helper and repeat-call cases |
| reader matches the immediate callee again | **red** — the `it.each` case |
| the config stops reading the derived value | **red** — `expected 5000 to be 13437` |
| the dossier test drops its stated ceiling | **red** — `expected 11000 to be 10000` |
| reader loses import resolution | **red** — six unfoldable timeouts reported |

## Scope and derivation corrections

- **The exemption is keyed by `file:line`.** Exempting all of `company-context.test.tsx` covered forty tests under a reason describing two, and silently unguarded anything added there later.
- **The default population is "states no ceiling of its own"**, not "ceiling equals the constant". The old spelling made the set's membership depend on the constant measured over it — re-derive one and the other moves, so a developer following the red message could loop.
- **The ceiling is 13 437 ms**, because with the reader seeing what it had been missing the widest budget in that population is 10 000 ms (`company-context`'s two write-posture cases, which state no ceiling and so belong to it). Not 7000; that figure was an artefact of the blind spots.
- **The per-test ceilings added in #1710 used round headroom**, against the rule this change states for the global. They now derive their between-waits allowance from `SLOWEST_MEASURED_TEST_MS`, so local and global rest on one measurement.
- Two stale references fixed: the constant's docstring named a test that does not exist (`grep` finds no "diagnostics card"), and two comments still pointed at `src/test-budget.ts` after the move to `scripts/`.

## Verification

`make check-q` green. `make check-fe` green — **238 files, 3047 tests**.

## Coverage

Test tooling and test config: no new product code, so new-code coverage does not apply. The two mutation batteries above are the substitute proof.

## Review

- `/code-review` (Fable) — three rounds so far, and each one found something real. Running again on this diff; **auto-merge deliberately not armed** until it comes back.
- **Codex has not reviewed this**: `/codex:*` cannot be invoked by an agent.

Claims worth attacking:
- *"reports what it cannot fold"* — one hop of import resolution is arbitrary; a budget two modules away is still reported rather than chased, which is the safe end but not a complete answer.
- *"the widest is 10 000"* — measured by the same reader it is checking. The reader's own tests are what stop that being circular; they are eleven cases, not a proof.
- *"the exemption covers exactly two tests"* — worth confirming nothing else in that file drifts over the ceiling later, since only those two lines are named.

Refs #1144, follows #1706 and #1710

### DIFF
--- a/frontend/scripts/test-budget.reader.test.ts
+++ b/frontend/scripts/test-budget.reader.test.ts
@@ -143,6 +143,88 @@ describe("the waiter-budget reader", () => {
     expect(probe?.unresolved).toEqual([]);
   });
 
+  it("resolves the shorthand `{ timeout }`, which is how a named budget is written", () => {
+    // The idiomatic spelling once the value has a name, and the one a reader
+    // that only looks at property ASSIGNMENTS records as the 1000ms default.
+    const [probe] = read(
+      `const timeout = 4000;
+       it("x", async () => { await waitFor(() => {}, { timeout }); });`,
+      "shorthand.test.tsx",
+    );
+    expect(probe?.waiterBudgetMs).toBe(4_000);
+    expect(probe?.unresolved).toEqual([]);
+  });
+
+  it("reports an options object it cannot see inside, rather than defaulting", () => {
+    // A spread and an options object passed by name each hide a budget. Saying
+    // "no timeout stated" about them is the same mistake as folding one wrong:
+    // both end as the smallest number that keeps the gate green.
+    const spread = read(
+      `const opts = { timeout: 4000 };
+       it("x", async () => { await waitFor(() => {}, { ...opts }); });`,
+      "spread.test.tsx",
+    )[0];
+    const byName = read(
+      `const opts = { timeout: 4000 };
+       it("x", async () => { await waitFor(() => {}, opts); });`,
+      "byname.test.tsx",
+    )[0];
+    expect(spread?.unresolved.length).toBe(1);
+    expect(byName?.unresolved.length).toBe(1);
+  });
+
+  it("reads past the `undefined` placeholder to the options behind it", () => {
+    // `findByText(matcher, undefined, { timeout })` is the ordinary spelling
+    // when the middle argument is skipped, and a reader that treats any
+    // identifier as an opaque options object reports it unreadable instead of
+    // folding the budget that is right there.
+    const [probe] = read(
+      `it("x", async () => {
+         await screen.findByText(/a/, undefined, { timeout: 4000 });
+       });`,
+      "placeholder.test.tsx",
+    );
+    expect(probe?.unresolved).toEqual([]);
+    expect(probe?.waiterBudgetMs).toBe(4_000);
+  });
+
+  it("does not read a non-string title as the ceiling", () => {
+    // Scanning the arguments from the start finds the title first, and then
+    // reports a test's own name as a timeout it cannot fold.
+    const [probe] = read(
+      `const TITLE = "x";
+       it(TITLE, async () => { await screen.findByText("a"); });`,
+      "title.test.tsx",
+    );
+    expect(probe?.unresolved).toEqual([]);
+    expect(probe?.ceilingIsStated).toBe(false);
+  });
+
+  it("ignores a test vitest never runs", () => {
+    // A skipped body spends nothing. Counting it lets a test that cannot fail
+    // set the ceiling for the thousands that do.
+    const found = read(
+      `it.skip("skipped", async () => { await waitFor(() => {}, { timeout: 30000 }); });
+       it.todo("todo");`,
+      "skipped.test.tsx",
+    );
+    expect(found).toEqual([]);
+  });
+
+  it("counts a helper declared inside the test body once per call", () => {
+    // Collected as a helper AND visited where it is defined, it was charged
+    // once for the declaration plus once per call.
+    const [probe] = read(
+      `it("x", async () => {
+         const settle = () => waitFor(() => {}, { timeout: 1000 });
+         await settle();
+         await settle();
+       });`,
+      "inbody.test.tsx",
+    );
+    expect(probe?.waiterBudgetMs).toBe(2_000);
+  });
+
   it("reports a timeout it cannot fold instead of assuming the default", () => {
     // The one that matters: an unfoldable budget quietly recorded as 1000 hid a
     // live 11000ms violation from the guard.
--- a/frontend/scripts/test-budget.reader.test.ts
+++ b/frontend/scripts/test-budget.reader.test.ts
@@ -225,6 +225,109 @@ describe("the waiter-budget reader", () => {
     expect(probe?.waiterBudgetMs).toBe(2_000);
   });
 
+  it("takes the EXPORTED constant, not a local one that shares its name", () => {
+    // A module-level export and a function-local of the same name are both
+    // "the last declaration of X" to a reader that walks a file flat. Taking
+    // the wrong one folded a 10000ms budget to 2ms, silently.
+    writeFileSync(
+      join(dir, "shadowed.ts"),
+      `export const POLL_MS = 5000;\nfunction helper() { const POLL_MS = 1; return POLL_MS; }\n`,
+      "utf8",
+    );
+    const [probe] = read(
+      `import { POLL_MS } from "./shadowed";
+       it("x", async () => { await waitFor(() => {}, { timeout: POLL_MS * 2 }); });`,
+      "shadowed.test.tsx",
+    );
+    expect(probe?.waiterBudgetMs).toBe(10_000);
+  });
+
+  it("reports every options shape it cannot see into, not just the two it knows", () => {
+    // A whitelist of readable shapes sends everything else down the same path
+    // as "states no timeout", and that path ends at the smallest number.
+    for (const [name, argument] of [
+      ["access", "OPTS.slow"],
+      ["call", "makeOpts()"],
+      ["ternary", "cond ? { timeout: 30000 } : undefined"],
+    ] as const) {
+      const [probe] = read(
+        `it("x", async () => { await waitFor(() => {}, ${argument}); });`,
+        `opaque-${name}.test.tsx`,
+      );
+      expect([name, probe?.unresolved.length]).toEqual([name, 1]);
+    }
+  });
+
+  it("records a conditionally-run test, which really does run", () => {
+    // `it.runIf(cond)` runs when cond holds. Treating it as skipped left a
+    // live test under no guard at all — the more expensive of the two mistakes.
+    const found = read(
+      `it.runIf(cond)("conditional", async () => { await waitFor(() => {}, { timeout: 30000 }); });`,
+      "runif.test.tsx",
+    );
+    expect(found.map((probe) => probe.waiterBudgetMs)).toEqual([30_000]);
+  });
+
+  it("ignores every test inside a skipped suite", () => {
+    // isSkipped inspecting only the `it` chain reddened the guard naming a
+    // test vitest never runs — the same failure it was added to prevent, one
+    // level up.
+    const found = read(
+      `describe.skip("suite", () => {
+         it("x", async () => { await waitFor(() => {}, { timeout: 30000 }); });
+       });`,
+      "describeskip.test.tsx",
+    );
+    expect(found).toEqual([]);
+  });
+
+  it("does not treat a local variable as a helper just because the name matches", () => {
+    // Skipping a "helper declaration" that is not a function at all dropped
+    // the waiter inside it.
+    const [probe] = read(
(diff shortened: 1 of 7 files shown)


tinnhlt-gra:
## Seven findings, all seven addressed in 6f3c6f9 — and finding 1 was the same hole again

**1 (HIGH) — confirmed, and it was live in three spellings.** `{ timeout }` shorthand is not a property *assignment*, so it took the 1000ms default; `{ ...opts }` and an options object passed by name did the same. A test spending 4000ms read as 1000 and the gate stayed green. That is exactly what this PR claims to have closed, one spelling over.

The shorthand resolves now. The other two are **reported**: an options argument the reader cannot see inside is a budget nobody checked, and calling that "states no timeout" is the same mistake as folding one wrong — both land on the smallest number that keeps the gate green.

Fixing it broke something, which is worth saying plainly: treating every identifier as opaque options turned `findByText(matcher, **undefined**, { timeout: 4000 })` into four "unreadable" reports. The placeholder is an identifier. Caught by running against the tree, not by a test — so the placeholder now has one.

**2 (MEDIUM) — confirmed.** The ceiling was taken from the first numeric-or-identifier argument anywhere in the call, so `it(TITLE, fn)` reported a test's own name as an unfoldable timeout. The bare number is only ever *after* the callback, and that is where it is read from now.

**3 (MEDIUM) — confirmed.** A helper declared inside a test body was collected as a helper *and* visited at its declaration: 1 + N instead of N. Your probe reproduces exactly (3000 for a 2000ms test).

**4 (MEDIUM) — confirmed.** `it.skip`/`it.todo` were recorded. A body vitest never runs could set `MAX_DEFAULT_WAITER_BUDGET_MS` and push `testTimeout` up for the 2900 tests that do run, or redden the guard naming a test that cannot fail. `.skip`, `.todo`, `.skipIf` and `.runIf` are excluded now, at any link in the chain.

**7 (LOW) — confirmed, both halves.** A reading taken while the recursion guard was suppressing part of the graph is no longer memoized — cached, it handed a later direct call the under-count, so the total depended on which helper was reached first. And `unresolved` is de-duplicated, so one unreadable timeout in a helper awaited four times reads as one problem rather than four.

Six cases join `test-budget.reader.test.ts`, one per hole, each mutation-tested — reverting a fix reddens exactly its own case:

| Mutant | Result |
|---|---|
| stops reading shorthand `{ timeout }` | **red** |
| stops reporting a spread it cannot see into | **red** |
| counts an in-body helper at its declaration too | **red** |
| counts a skipped test | **red** |
| treats the `undefined` placeholder as opaque | **red** (reader test *and* the tree gate) |

## 5 and 6 — the comments that had gone false

**6** is fixed: the exemption docstring still quoted the superseded 10437.

**5 you were right about twice over.** The header said the ceiling is *smaller* than company-context's 10s-per-waiter budget and therefore cannot mask issue 613 — it is now larger, and what actually keeps that file fast-red is the exemption, not the number. Corrected.

The substantive half I have **not** fixed, deliberately, and recorded instead: the suite ceiling is driven by `company-context.test.tsx:335`/`:347`, which state no ceiling of their own. Giving them one — as `integrations-provider.test.tsx` does here — drops `MAX_DEFAULT_WAITER_BUDGET_MS` to 7000 and the ceiling to 10437. But that file carries issue 613 and the instruction for this work is not to touch its timeouts while that investigation is live. **Issue #1717** carries it, with the three seconds every other test pays in the meantime stated in the budget file rather than left for the next reader to discover.

## The stale number

Right — the tree reads **2971**, not 2941. The count moved when the reader stopped being blind; the PR body quoted a figure from before the last fix. Corrected in the description.

## Verification

`make check-q` green. `make check-fe` green — **243 files, 3087 tests**. Census: 2971 records, **0 unreadable**, and the only tests over their ceiling are the two exempted by name.

Auto-merge still not armed.

tinnhlt-gra:
## Eight findings, all eight addressed in 54945d9 — and the two HIGH ones share a cause worth naming

Both reduce to the same thing: **the reader whitelisted the shapes it understood, and everything else fell through to "states no timeout"** — which ends at the 1000 ms default. The smallest number is the one that keeps the gate green, so every gap in the whitelist was a silent pass. That is the fourth time this defect has appeared in this PR under a different spelling, so the default is now inverted rather than the list extended.

**1 (HIGH) — the import hop introduced one of them.** `takeImport` folded the imported module with `declaredConsts`, the same flat last-wins reader used for a test file, so a function-local `const POLL_MS = 1` overwrote the exported `POLL_MS = 5000` and a 10 000 ms budget folded to **2 ms** with nothing reported. Only top-level *exported* declarations cross a module boundary now (`exportedConsts`).

**2 (HIGH) — confirmed, all five shapes.** `OPTS.slow`, `makeOpts()`, `cond ? {...} : undefined`, `opts as WaitOptions`, `opts!` all reached the default. `cannotCarryOptions` replaces `looksLikeOptions`: an argument is assumed to be carrying options unless its shape *definitively cannot* — a matcher, a placeholder, a callback. Everything else that is not an object literal is reported.

**3 (MEDIUM) — you are right and my docstring was wrong.** `runIf`/`skipIf` run on a condition; treating them as skipped left a live test under **no guard at all**, which is the more expensive of the two mistakes. Removed from `NOT_RUN`.

**4 (MEDIUM) — confirmed.** The skip check inspected only the `it` chain, so `describe.skip` suites were recorded and could redden the guard naming a test vitest never runs. Skipping is inherited down the tree now.

**5 (MEDIUM) — confirmed, and it was a regression I introduced.** `declaresHelper` returned true for any variable whose name matched, so `const renderAs = await screen.findByText(…, { timeout: 25000 })` had its whole initializer skipped: **0 ms**, where the pre-diff reader counted it correctly. A declaration is now one whose initializer *is* a function.

**6 (MEDIUM) — confirmed, with a caveat worth reporting.** My first fix reported every name collision as unreadable. Run over the tree it named **68 tests** — `mount` in `App.test.tsx`, `strip` in `company360.test.tsx` — which is a gate nobody would keep. So every declaration is kept and the *disagreement* decides: same cost either way is not an ambiguity that matters; a different cost is genuinely unreadable and is reported. Your 8000/2000 probe reports; two identical 2000 ms helpers do not.

**7 (MEDIUM) — confirmed.** A ceiling written as `SETTLE_MS * 4` was neither folded nor reported, so the test silently rejoined the population it had opted out of. The positional scan takes any expression after the callback and folds it.

**8 (LOW) — confirmed, and it matters more than "low" suggests** given that file is the one under active investigation. The exemption is keyed on the test **name** now:

```
src/screens/company-context.test.tsx::quotes the server when the start itself was refused
src/screens/company-context.test.tsx::keeps a failed status poll to the catalog sentence
```

## Verification

Seven cases join `test-budget.reader.test.ts` (24 there, 29 with the guard). Each mutation-tested — reverting a fix reddens exactly its own case:

| Mutant | Result |
|---|---|
| import folding takes any declaration, not the exported one | **red** |
| unknown options shapes fall through as "no timeout" | **red** (two cases) |
| a conditionally-run test is treated as skipped | **red** |
| `describe.skip` no longer suppresses its tests | **red** |
| any same-named local is treated as a helper definition | **red** |
| disagreeing helpers resolve to one silently | **red** |

Tree: **2971 records, 0 unreadable**, only the two exempted tests over their ceiling. `make check-q` green; `make check-fe` green — 243 files, 3094 tests.

Auto-merge still not armed.

## One thing I want to flag rather than bury

This is round four, and each round has found a real defect of the same class in the reader. That is a fair signal about the approach: a static reader of a dynamic language has an unbounded surface, and each pass finds another shape. The two things that have made it converge are (a) defaulting to *unreadable* rather than to a number, so a gap now fails loudly instead of passing, and (b) the reader having its own tests, which is what turns each finding into a permanent one. If the next pass finds a fifth, I would rather it fail the gate than fold to 1000 ms — and after this change, it will.

tinnhlt-gra:
## Round five — five findings, all five fixed in the pushed head

**1 (MEDIUM) — the same last-wins defect one hop closer, and you are right that it was the one `exportedConsts` had just fixed.** `declaredConsts` walks a file flat, so a module-level `const SETTLE_MS = 9000` shadowed by a function-local `= 1` folded every waiter naming it to **1ms**, silently. Every value a name is given is kept now, and a name given two different values resolves to **nothing** — the waiter is reported rather than folded to whichever came last. Same rule the helpers already follow.

**2 (MEDIUM) — confirmed.** A helper is only ever called by its bare name; matching on the property name charged `userEvent.clear(input)` the whole budget of a file-declared `clear`. Your near-miss at `import.test.tsx:138` is a good catch — it read correctly only because the recursion guard happened to zero a self-collision, which would not have saved two different names.

**3 (LOW) — fixed, and I agree with your reasoning for not rating it higher and fixing it anyway.** A contract that holds only because biome's `useLiteralKeys` runs over `src` is not a contract this reader can rely on. `{ "timeout": 5000 }` and `{ ["timeout"]: 5000 }` both fold now.

**4 (LOW) — the comment was overclaiming and is rewritten.** You are right that the total still depends on reach order: the outermost memoized reading embeds the truncated inner one. Reproduced your 7000/8000 asymmetry. The comment now says what the code does — both directions under-count, which is the safe end — and names the real fix (key the memo on the path) rather than implying the case is handled.

**5 (LOW) — fixed by assertion rather than by re-keying.** `file::name` is indeed not unique here; I checked and found the same three collisions. Rather than reach for a third key, the guard now asserts each `EXEMPT` entry matches **exactly one** budget — so an entry that stops matching, or comes to match two, fails loudly. Mutation-tested: pointing an entry at a name no test has reddens three assertions.

**On the stale PR body** — right, it says `file:line` and the code keys by name. Corrected.

### Verification

Three cases join the reader's tests, one joins the guard's; 33 between them. Each mutation-tested:

| Mutant | Result |
|---|---|
| a shadowed constant resolves to the last value seen | **red** |
| a method call is charged a same-named helper's budget | **red** |
| the timeout key is compared as source text | **red** |
| an exemption entry that matches no test | **red** (three assertions) |

`make check-q` green; `make check-fe` green — 246 files, 3133 tests.

### Where this leaves the PR

Five rounds: 2 HIGH + 5 MEDIUM + 1 LOW, then 2 MEDIUM + 3 LOW. The severity is falling and the last round found nothing live in the tree — every finding was latent, which is the point at which a static reader of a dynamic language stops paying for another pass.

The two things that made it converge are worth stating because they are the transferable part: **defaulting to unreadable rather than to a number**, so a gap fails loudly instead of passing, and **the reader having its own tests**, which turns each finding into a permanent one rather than a fix that the next refactor undoes. I am arming auto-merge on this round.

Issue #1728, opened 2026-08-18
A deploy recipe that builds the role images itself gets no mixed-release guard


lstrojny:
## What is missing

#1705 gave every role a release version and a run-time guard against a torn tag pull. The version arrives as the `MARGINCE_RELEASE_VERSION` build argument, and `docker-bake.hcl` sets it from `VERSION` for all three targets in one place — so the **release workflow** always stamps a real release.

A deploy recipe that builds the targets directly does not. The d13 `dockerBuild` block builds a role with `dockerFile: ./Dockerfile` plus pass-through `arguments: ["--target", "<role>"]` (see the comment on the `role` target in `docker-bake.hcl`), and passes no build argument for the release. The Dockerfile's ARG default is then `dev`, which `internal/shared/buildinfo.Comparable` treats as unknown — and unknown disables every comparison:

- the api records nothing as the installation's release,
- the worker compares nothing and always starts,
- the SPA reports no `release_version` on `/auth/capabilities` and never blocks.

So the guard is silently inert on any installation whose images were built by the deploy pipeline rather than pulled from a release. That is the *correct* fail-safe direction — an unstamped build must never refuse to boot over a release it does not have — but it means the deployments most likely to be exercised by hand are the ones with no protection.

## What would fix it

Have the recipe pass a release version it can honestly name. The pipeline knows the revision it is building, so the value does not have to be a constellation release version — any stable per-build identifier used consistently for **all three roles** makes the comparison meaningful, because the guard compares for equality and never for order.

The change is in the deploy repo, not this one. This issue tracks the gap so it is not mistaken for a guard that is working.

## Why this is not `priority: high`

Nothing is wrong on a stamped installation, and nothing is wrong on an unstamped one that was not already wrong before the guard existed. What is lost is coverage, on deployments that do not go through the release pipeline.

Refs #1705, gradionhq/margince-constellation#149.

dungbanguyen:
## Decision, 2026-09-01: keep the fail-safe; make the inertness visible

Re-triaged against `origin/main` at `d42887998`. Confirmed — `Dockerfile:84`,
`:110` and `:229` all carry `ARG MARGINCE_RELEASE_VERSION=dev`, and
`buildinfo.Comparable` treats `dev` as unknown, which disables every comparison.

### The ruling

**The behaviour stays. The silence does not.**

The fail-safe direction is right and this decision does not touch it: an unstamped
build must never refuse to boot over a release it does not have. That is the correct
[...]

PR #1729, opened 2026-08-18
feat(release): put the release version in each image, and stop a mixed set at start

### LINKED ISSUE #1705: Guard against mixed versions at start: put the release version in each image and stop the services
Out of scope in https://github.com/gradionhq/margince-constellation/pull/229 (entitlement work for https://github.com/gradionhq/margince-constellation/issues/149).

## Problem

The customer pulls each release image by its tag. The registry tag resolver does this. Two tag pulls are two different requests. Thus the customer can get a mixed set. In a mixed set, the service roles come from two different releases. This can happen during a publish operation. The registry cannot prevent this at the pull. The OCI protocol does not permit it. The design accepts this limit.

## Necessary function (from the specification)

- Each release image contains its release version two times: in an OCI label, and in a file in the image.
- At start, each Margince service compares its release version with the versions of the other services in the set.
- If the versions are not the same, the services must not start.
- A mixed `latest` set must not start.

## Notes

- The guard is a function of the Margince application services.
- This repository does only the release pipeline part. It puts the version label and the version file in the images that it builds.

## References

- Specification: `specs/functional/components/02-release-management.md`, `03-registry-auth-realm.md`, `06-dist-service.md` (added by https://github.com/gradionhq/margince-constellation/issues/225)
- Design decision record: https://github.com/gradionhq/margince-constellation/issues/149

lstrojny:
Closes #1705.

The contract change is in the specification repository: gradionhq/margince-foundation#1351. Merge that pull request first.

The design decision record is gradionhq/margince-constellation#149.

## The problem

The customer pulls each role image by its tag. Two tag pulls are two different requests. A publish operation can occur between the two pulls. The customer then gets a mixed set. In a mixed set, the roles come from two different releases.

The registry cannot prevent this at the pull. The OCI distribution protocol has no request for two manifests together. Therefore the services must refuse a mixed set at start.

## Each image carries its release version

The bake file gives one build argument to all three targets. The Dockerfile makes three copies of that value. The three copies cannot disagree, because one build makes all three from one value.

| Location | How to read it | Who reads it |
|---|---|---|
| The OCI label `org.opencontainers.image.version` | `docker inspect`. A pull is not necessary. | An operator who compares a set. |
| The file `/etc/margince/release-version` | `kubectl exec … cat`. | An operator who has a container that restarts. This is also the only run-time source for the web image, because nginx runs no code of ours. |
| A link-time stamp in each Go binary, and a vite define in the bundle | The guard reads it. | The software. |

The release workflow needs no change. It already gives `VERSION` to the bake file as the image tag.

## The api is the authority

The api image contains `cmd/migrate`. The entrypoint applies the schema before the api serves. Therefore the schema of the installation comes from the release of the api.

At start, the api records its own release version. It writes one row in `system_log`. It writes only when the value is different from the last recorded value.

Each other role reads that value at start. A role stops when its own release version is different. The message gives both release versions. It also gives the action for the operator.

### Why one authority, and not a comparison between equals

Two other designs were examined. Both designs fail.

- A comparison against live peers stops each rolling upgrade. A rolling update keeps the old pods until the new pods are ready. The new api finds a live old worker, and refuses. The new worker finds a live old api, and refuses. No new pod can become ready.
- A symmetric design needs a new table. On a mixed set, the role that starts first still starts.

One authority permits an upgrade. The api records the new release version first. The other roles then agree with it.

A rollback works for the same reason. The guard compares for equality only. It does not compare order. Therefore a change to an older release needs no special rule.

## The web tier compares in the browser

The web tier is a static bundle in a browser. It cannot read the database. It can only ask the api.

The web tier must ask before a user signs in. A mixed set stops the login request first. A guard behind authentication runs only on installations that are already correct.

`GET /v1/auth/capabilities` gives the release version of the api. This endpoint is anonymous. The login screen already reads it.

The SPA compares that value with its own release version. If the two values are different, the SPA shows one screen. The screen gives both release versions. It also offers a reload.

The SPA keeps no state for this screen. A rolling deploy therefore clears the screen without an action.

The field is optional. The release version is not a secret, because the anonymous bundle contains the same value.

## An unknown release version disables the guard

An empty value and the value `dev` both mean this: the build has no release version. The guard then makes no comparison. This is true in both directions.

- A role with no release version does not refuse to start.
- An api with no release version does not replace the recorded value.

This keeps `make dev` and `docker build --target api .` in service. It is also the safe direction. A wrong refusal stops a correct installation. A wrong pass leaves the installation in the condition it had before this change.

## Verification

The guard was tested with a real image and a real database, and also in the test lanes.

```
$ docker inspect --format '{{index .Config.Labels "org.opencontainers.image.version"}}' …/worker:1970.9999
1970.9999
$ docker run --rm --entrypoint cat …/worker:1970.9999 /etc/margince/release-version
1970.9999

# An api with release version 1970.9998 starts. It writes one row.
# A restart at the same release version writes no more rows.
release.version_observed|{"release_version": "1970.9998"}

# A worker with release version 1970.9999 starts against that installation.
worker: this role was built from release "1970.9999" but this installation runs release "1970.9998":
a tag pull served images from two different releases, which the registry cannot refuse at the pull.
Re-pull every role image (api, web, worker) at one release and restart; this role will not run half
of one release beside half of another
$ echo $?
1

# A worker with release version 1970.9998 starts.
level=INFO msg="release matches the installation" release_version=1970.9998

$ curl -s …/v1/auth/capabilities
{"oidc_providers":[],"password":true,"password_reset":false,"release_version":"1970.9998"}
```

The test lanes give these results:

- One integration test uses a real Postgres database. It calls the two production functions in this order: no record, a record, an unchanged restart, a match, a mixed set in each direction, no release version in each direction, an upgrade, and a rollback.
- A second integration test uses an installation that has no organization. Both functions return no error, and they write no rows.
- A Go table and a TypeScript table make the same comparisons. The two tables agree.
- Six vitest tests examine the screen in the browser. One test covers the condition in which the probe has no answer yet.

## Two properties are limited

The source file gives these limits. Do not read the guard as more than this.

- The guard makes its comparison at start only. A worker that already runs is not examined again. A restart of the api alone therefore leaves the old worker in service. See #1734.
- The last write wins. During a rolling upgrade, api pods from two releases can be in service together. Each pod can be the last one to write. The recorded release version can therefore change to an older value. See #1735.

One property is true in all conditions: the roles that disagree with the recorded release version refuse to start. Therefore no mixed set gives service.

## Gates

These commands give a pass: `make check-backend`, `make check-fe`, `make craft-static`, and `scripts/check-contract-breaking.sh`. The full integration lane gives a pass with 0 skips.

The function `run()` in `cmd/api` was at the limit of 80 lines. The two boot records are now one phase in `boot.go`. That file holds the phases of the boot sequence.

## Open item

#1728: a deploy pipeline that builds the role images itself gives no release version. The guard is inactive on such an installation. This is the safe direction, but the installation has no protection. The correction is in the deploy repository.

## Summary by CodeRabbit

* **New Features**
  * Added release-version consistency checks across the API, worker, and web application.
  * The web app now detects API version mismatches, displays both versions, and provides a reload action.
  * Authentication capability responses can report the server release version for compatibility checks.
* **Bug Fixes**
  * Prevented mixed-release services from starting or rendering inconsistent application states.
  * Improved release tracking across upgrades, rollbacks, and archived installations.
* **Documentation**
  * Added deployment guidance for coordinated releases and mismatch handling.


lstrojny:
## Review round applied

The craftsmanship and security reviewers found eleven things between them. Three were real defects and are fixed; two are design bounds now stated in the file and tracked as issues; the rest were comment and naming work.

### Fixed — defects

**`/auth/capabilities` is `no-store`.** The best finding of the round, and it is a consequence of this very change: the release version turned an anonymous probe into a kill switch, because the SPA refuses to render when what it reads there differs from its own build. The response carried no `Cache-Control`, no validator and no `Vary` — exactly what an intermediary assigns heuristic freshness to — so **one stale copy in any cache on the shared api/web origin serves the mixed-release screen to every reader behind it, and no reload clears it.** Four siblings in the same module already set the header. Not a leak (the body is not per-principal); a self-inflicted availability failure, and an amplifier for anyone holding a cache-poisoning primitive against the ingress. Asserted by a test now.

**The advisory lock coalesces its GUC, in both call sites.** `pg_advisory_xact_lock` is strict: a NULL argument takes **no lock** and returns NULL rather than raising. An unset `app.workspace_id` made `current_setting(…, true)` empty and the concatenation NULL, which would have left the serialization silently absent — a guard reporting success while holding nothing. Not reachable (`WithWorkspaceTx` refuses an unbound transaction first); the point is that nothing downstream would say so if it became reachable. It is one parameterized constant in `bootscope.go` now, shared with the extension inventory, **which had the same omission** — the invariant, not the call site.

**Two closures assigned the enclosing `err`.** Correct only because they also returned the value, and one wrote a variable the surrounding call was about to overwrite. Both read locally now, matching the sibling.

### Fixed — a claim that was false as written

"this is the role that applies the migrations" — it is not. `cmd/migrate` is its own process role and the only caller of the migrator; what makes the api the authority is that its **image ships and runs it** before serving. The sentence was wrong in `boot.go` and in `docs/deployment.md` and right in `releaseversion.go`, which is a good illustration of the next item.

### Fixed — coverage

The **pre-bootstrap arm now has a test**: both functions return nil and write nothing on an installation with no organization yet, reached through the harness's own reset rather than by deleting rows. That is the arm where a wrong answer in either direction looks exactly like a working guard — an error would refuse to bootstrap the installation the api is about to create, and treating the absent record as a value would refuse over a release nobody recorded.

### Recorded, not fixed — two design bounds

Closing either is a design change, and one of them costs rollback. `releaseversion.go` states both explicitly instead of implying the record only moves forward:

- **#1734** — the check is boot-only, so an api-only restart leaves the old worker running. Needs a periodic re-check that does not exit on an unreadable record and does not churn during a rollout.
- **#1735** — the record is last-writer-wins, so a stale api replica can re-record an older release; and `occurred_at` is transaction-start time, so "latest" is not "last committed" even though the lock serializes the write correctly.

What holds either way is the invariant this exists for: whichever release wins the record, the roles that disagree refuse, so no mixed set serves.

### Judgment calls, with reasons

- **`recordBootFacts` "at the wrong altitude, and the ceiling did not force it."** Measured: inlining the two calls puts `run()` at **83 code lines against the 80 ceiling** and `craft static` blocks it, so the extraction is forced. `boot.go`'s own doc says it holds the phases and five sibling one-caller helpers live there, so a private phase matches the file. Took the rest of the finding: renamed to `recordBootLedger` (concrete, and `bootLedgerScope`/`bootLedgerLock` already establish the noun) and cut the doc down to the ordering invariant instead of re-arguing the design.
- **Extracting the whole observe-on-change frame**, not just the lock. Took the lock (that is where the silent-failure mode was); left the frame, because restructuring a working sibling is a larger diff than this branch should carry. The two now share the statement that could fail quietly.
- **Ordering by `occurred_at`** stays as-is, matching the sibling and its stated rationale. Changing it needs `storekit.LogSystem` to accept an observation time, and it does not close #1735 on its own — see that issue.

### Also

Rationale trimmed from six restatements to one canonical place (which is how the migration claim came to be right in one copy and wrong in another), three paragraphs of deployment reasoning cut out of the client-facing operation description — [foundation#1351](https://github.com/gradionhq/margince-foundation/pull/1351) updated to match — and an unstamped api now logs that it disarmed the guard, symmetric with the assert path.

`make check-backend`, `make check-fe` and `make craft-static` re-run green; the pre-push whole-tree greps pass.

lstrojny:
@coderabbitai Both review rounds are addressed in 0b02eba1. Every finding was verified against the code before it was applied; one of them was better than its framing.

## The workspace scope on `lastObservedReleaseQuery` — confirmed, fixed, and it was worse than "multi-workspace databases"

The suggestion said the read could match rows from other workspaces "if multi-workspace data exists". Verified: it could, and the reachable case is not hypothetical. Row-level security used to supply this predicate on every `workspace_id` table; migration `0217_retire_row_level_security` dropped it schema-wide and its own header names the cost — an unscoped statement now returns another tenant's rows and no test fails by construction. Checked live: `system_log` has `relrowsecurity=f`, zero policies.

Two live organizations are refused by the installation resolver, but an **archived** workspace is skipped by `archived_at IS NULL` while its `system_log` rows remain. A newer `release.version_observed` row of its own would then decide what release this installation runs — and every correctly deployed role would refuse to start against it. The guard would have caused the outage it exists to prevent.

Fixed in both places, because the release query was copied from the extension inventory's and **the sibling had the same missing predicate** (`lastObservedExtensions`). Both now carry the tenant predicate in the spelling `storekit` uses. An integration test pins it through the archived-workspace path, and it was mutation-checked: with the predicate removed, the test fails exactly as the exploit predicts — the role at the *right* release refuses, having read the archived tenant's `1970.99` as the installation's.

## The vacuous SPA-gate assertions — confirmed, fixed

Correct, and the sharpest version of the criticism is that three of the six cases would have kept passing with the comparison deleted outright, because "the app renders" is also true while the probe is still pending. The harness now reports the probe's settlement beside the verdict (`settled:true app renders`), every non-blocking case waits for settlement, and the pending case asserts `settled:false` explicitly rather than by omission. The failed-probe case has to outlast react-query's retry backoff — a timer — so it runs on fake timers per the frontend testing rules. Reload is now clicked and its call observed; jsdom defines `location.reload` non-configurable, so the test swaps the location object and restores it in `afterEach`.

## The `as` cast in `release.test.ts` — confirmed, fixed

Replaced with an explicitly typed case table. A cast in a test silences the one check that would notice the table and the function disagreeing about a type.

## `buildinfo.Unknown` instead of the literal `"dev"` — applied

The identity-module test already used the constant for the same concept; the compose table now does too.

## The `err`-shadowing nitpick — already fixed before this round

The closure-local read landed in fe83361d (the internal review round flagged the same shape); the second CodeRabbit pass reviewed the newer diff and did not re-raise it.

## The four docs findings — all applied, with one nuance

- The table no longer calls `/etc/margince/release-version` "the only run-time reading of the web image's release", which contradicted the SPA section below it. It is an inspection value; the bundle's compiled-in copy is what the web tier compares.
- The worker bullet states the comparison runs **at start only**, linked to #1734 (filed before this round by the internal review).
- The rollback paragraph now carries the last-writer-wins warning: whichever api records last decides the installation's release, older included, so an api rollout must not be left half-done with replicas from two releases alive. Linked to #1735.
- "In both tiers" is replaced with the three roles by name, and the paragraph now states the half that was missing: an unstamped api **leaves an existing record standing** rather than clearing it, so one locally built binary cannot disarm the guard for the roles that boot after it.

All gates re-run green: `make check-backend`, `make check-fe` (248 + 3 files), `make craft-static`, and the touched integration suites against real Postgres.

PR #1742, opened 2026-08-18
fix(ci): the integration lane's connection demand is a number it can state

### LINKED ISSUE #1109: Integration lane's default concurrency exhausts Postgres connections, failing unrelated packages on connect
## Symptom

`make test-integration` fails at its default concurrency with connection errors rather than assertion failures. The failing packages differ on every run, and the errors appear at connect time:

```
failed to connect to `user=margince_owner database=margince_it_pNN_XXXXX`:
	[::1]:15432 (localhost): tls error: EOF
	127.0.0.1:15432 (localhost): tls error: EOF
	[::1]:15432 (localhost): failed to receive message: unexpected EOF
```

The same shape also hits the lane's own migrate step (`migrate: connecting: failed to connect`), and one run left a clone database behind: `FAIL: clone db margince_it_pNN_XXXXX was not dropped — leaked on the test cluster`.

## Evidence it is not the code under test

- Three consecutive runs failed in three different package sets — `overlay`, then `e2e`/`embedding`/`similarity`, then `leadconsent`/`telegram`.
- The affected packages pass in isolation via `make test-it DIR=backend/...`.
- Postgres shows no restart, no OOM kill, and no `too many clients` line; `pg_stat_activity` is near-idle between runs.
- Lowering concurrency makes it green: `INTEGRATION_JOBS=3` and `INTEGRATION_JOBS=4` both complete with `OK: integration passed with 0 skips (38 packages, parallel)`.

## Likely cause

`scripts/test-integration-parallel.sh` defaults to `min(nproc, 8)` concurrent packages, each opening its own pool against one Postgres instance whose `max_connections` is the stock **100**. `infra/docker-compose.dev.yml` deliberately raises `max_locks_per_transaction` to 512 for this lane and documents why, but connection count was never given the same treatment. Each package also clones and drops a database, so a connect racing another package's `DROP DATABASE` is a plausible source of the mid-handshake EOF specifically.

## Why it matters

The failure is indistinguishable from a real one at a glance, and it moves. That is the expensive kind: a contributor bisects a green change, or worse, learns to re-run until green — which is how a genuine flake stops being reported.

## Options, roughly in order of preference

1. Raise `max_connections` on the dev/CI Postgres alongside the existing `max_locks_per_transaction`, with a comment tying it to `INTEGRATION_JOBS` the way the lock comment does.
2. Bound the per-package pool size in the integration harness, so demand is `JOBS × known-max` rather than `JOBS × whatever pgx defaults to`.
3. Lower the default `JOBS`, which trades wall-clock for reliability and is the least satisfying since CI runs 16 and is presumably sized for it.

Retrying on a connect-time EOF is deliberately not on that list: it would hide the condition rather than fix it, and the lane's whole design is that a slicing or isolation problem can only read as red.

## Reproduction

```
make db-up
make test-integration          # fails, differently each time
INTEGRATION_JOBS=3 make test-integration   # green, 0 skips
```

Observed on a machine reporting 8 usable cores.

tinnhlt-gra:
The integration lane's connection demand was a product of two numbers that did not know about each other, against a third nobody had ever set.

| Quantity | Was | Where |
|---|---|---|
| concurrent packages | `min(nproc, 8)` locally, **16 in CI** | `scripts/test-integration-parallel.sh`, `.github/workflows/ci.yml` |
| pool ceiling per pool | **16** | `database.NewPool`'s fallback whenever the DSN omits `pool_max_conns` |
| server ceiling | **stock 100** | `infra/docker-compose.dev.yml` never set `max_connections` |

`MaxConns` is a ceiling and not a reservation, so pgxpool dials lazily and the lane passes most of the time — which is exactly the reported symptom: connect-time failures in a **different package set every run**, green in isolation, green at `INTEGRATION_JOBS=3`. The defect is not "256 > 100 therefore it fails"; it is that **nothing bounded the demand**, so whether a run fits was decided by how the bursts happened to overlap.

## What this does

**1. Bounds the per-package demand.** The lane hands the harness a per-pool ceiling (`MARGINCE_TEST_POOL_MAX_CONNS`, `testdb.PoolMaxConnsEnv`), so the demand is `JOBS × a known number` rather than `JOBS × whatever pgx falls back to`. This is the half that survives someone changing `INTEGRATION_JOBS` later.

It travels as an environment variable and **not in the DSN**, which was the first attempt and is worth recording: `cmd/migrate` and every fixture that dials a bare `pgx.Conn` parse these same DSNs with `pgx.ParseConfig`, which does *not* strip `pool_*` keys — it forwards an unrecognised one to the server as a startup parameter, and the connection dies with `FATAL: unrecognized configuration parameter "pool_max_conns"`. Only `pgxpool.ParseConfig` knows those keys. The lane failed at its template build, immediately and loudly.

**2. Raises `max_connections` to a derived number**, with a comment written to match the `max_locks_per_transaction` comment already in that file — it names the terms, does the arithmetic, and says what the shortfall produces. `408 = 16 jobs × (24 per package + 1 admin connection per slot) + 8 fixed`, where the 24 is itself a sum — two pools at the pinned ceiling of 8, plus 8 for the ad-hoc `pgx` connections a fixture dials that no pool bounds. It also states the consequence the naive version of this change misses: the cluster's lock-slot budget is `max_locks_per_transaction × (max_connections + max_prepared_transactions)`, so this raises the lock budget too — 51,200 → 208,896 slots, measured: the server reports a 230MB shared-memory segment at 408 against 204MB at 280, so the whole widening costs ~26MB. That direction helps (a whole-schema reset needs more slots, not fewer), but it had to be checked rather than assumed.

**3. Makes it permanent.** Four guards, each proven red against the defect it describes before the fix (`WORKFLOW` §4):

| Guard | Lane | Goes red when |
|---|---|---|
| `TestTheLaneFitsInsideTheClusterItRunsAgainst` | `make check` | the committed `INTEGRATION_JOBS × terms` stops fitting the committed `max_connections` — reading all three from where they are declared, never restating them |
| `TestTheBudgetGateRefusesTheConfigurationThatShipped` | `make check` | the arithmetic above becomes vacuous: it re-runs it over the pre-#1109 configuration and requires it *not* to fit |
| `TestTheLanePinsThePoolCeilingItBudgetsFor` | `make check` | either end of the shell→Go seam is renamed or unexported, which would leave every pool back at 16 with the budget still saying 8 |
| `TestTheClusterSeatsTheBudgetTheLaneComputed` + `TestTheSharedPoolTakesTheCeilingTheLaneHandedIt` | integration | the **running container** predates the compose change (Postgres fixes `max_connections` at startup), or the pool ignores the ceiling it was handed |

## Deliberately not done

- **Not lowering `INTEGRATION_JOBS`** — it trades wall clock for reliability and CI runs 16 on purpose.
- **No retry on a connect-time EOF.** The lane's whole design is that a slicing or isolation problem can only read as red.

## What I could NOT reproduce, and it matters

`make test-integration` **did not fail** at the default concurrency on this machine — three runs at 8 jobs and two at 16, all green — so the red half of the reproduction in the issue is not in this PR. What I measured instead, sampling `pg_stat_activity` from **outside** the lane's own transactions (a fresh connection per sample; a sample taken inside a transaction is frozen and under-reports — the trap #970 was written for):

| At `INTEGRATION_JOBS=16` (CI's setting) | |
|---|---|
| peak cluster backends | **87** of 100 — 87%, and 90% of the 97 a non-superuser can reach |
| peak on the lane's clone databases | 79 |
| peak for a single package | 13 |
| the *ceiling* those 13 were drawn from | 256 |

The first version of this PR set `LANE_CONNS_PER_PACKAGE` to 16 — which, re-measured **after** the pool ceiling was applied, is *exactly* the observed per-package peak. A budget equal to its own measurement is not a budget, so the term is 24 and says why. After the widening the peak is unchanged at 16, now against an allowance of 24.

So the lane was running at ~87% of a ceiling it had a ceiling of 256 against. That is the defect this PR fixes, and it is worth fixing whether or not it is the mechanism behind the reported EOF.

**And it is probably not that mechanism.** Exhausting `max_connections` over the exact TCP path the lane uses produces a clean `FATAL: sorry, too many clients already` — measured, 140 concurrent connections against the 100-connection cluster, 40 refusals, all with that message. The reported failure is a *mid-handshake EOF* with no server-side error, which is a different event: the server closed the socket before answering the SSLRequest. A high connect **rate** at low concurrency (600 connections, 40 at a time) produced no failures either. I have filed that residual as its own issue rather than letting this PR imply it is closed.

## Verification

Four consecutive lane runs at the shipped configuration — three at the local default and one at CI's `INTEGRATION_JOBS=16` — each ending `OK: integration passed with 0 skips (41 packages, parallel)`. The pinned ceiling costs about 2% of the long pole's wall clock (`compose/integration`: 178.6s before, 182.1s after, same machine).

Four more runs preceded them at an intermediate ceiling, also 4/4 green.

## Coverage

Config and harness work: **no measured new-code coverage** by construction. The substitute proof is the four guards above (each shown red against its own defect, then green) plus the recorded lane runs.

## Review status

Fable (`/code-review`) and CodeRabbit have run. **Codex has not** — every `/codex:*` command is `disable-model-invocation: true`, so an agent cannot invoke it; a human has to type `/codex:review`. Two reviewers, not three, and this note is here so it does not read as three.

Closes #1109

---
## Summary by cubic
Pins and enforces a predictable Postgres connection budget for integration tests to eliminate flaky connect-time failures. Old: demand was INTEGRATION_JOBS × pgxpool’s implicit 16 against stock max_connections=100. New: lanes set a per‑pool ceiling via `MARGINCE_TEST_POOL_MAX_CONNS`, compute one budget, and the cluster runs with `max_connections=408`, with tests that guard both the committed config and the running cluster.

- Single source of truth: `scripts/lib-testdb.sh` adds `declare_lane_budget` (`LANE_POOL_MAX_CONNS=8`, `LANE_CONNS_PER_PACKAGE=24`, `LANE_FIXED_CONNS=8`) and exports `MARGINCE_TEST_POOL_MAX_CONNS` and `LANE_CONN_BUDGET` for the parallel, one‑package, and serial lanes.
- Harness: `testdb` reads `MARGINCE_TEST_POOL_MAX_CONNS` (env, not DSN) and applies it as `pool_max_conns`; rejects non‑numeric or non‑positive values.
- Infra: `infra/docker-compose.dev.yml` sets `max_connections=408` with the arithmetic documented; clarifies the ceiling pin is via `MARGINCE_TEST_POOL_MAX_CONNS` (not DSN); notes the lock‑budget side effect and ~26MB shared‑memory increase.
- Gates: `make check` verifies the lane terms still fit the committed `max_connections` and continues to refuse the pre‑fix configuration; also asserts the env name is exported and applied. Integration tests fail—not skip—when required envs are absent and verify the live cluster seats the computed budget and pools take the ceiling.
- Tooling: `backend/Makefile` passes the budget in the serial lane; `scripts/test-integration-parallel.sh` and `scripts/test-integration-one.sh` compute and export it. Docs record `MARGINCE_TEST_POOL_MAX_CONNS`.

**Rollout**
- Recreate the Postgres container to pick up `max_connections` before running the lane.
- CI keeps `INTEGRATION_JOBS=16`. If you raise `INTEGRATION_JOBS`, raise `max_connections` and the lane terms together.

<sup>Written for commit fc217f0a17eaca4c73c2290b6a553620027ee985. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1742?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added connection-budget configuration for parallel integration tests.
  - Test pools now support an environment-defined maximum connection limit.
  - Development PostgreSQL is configured with an increased connection capacity.

- **Bug Fixes**
  - Added validation to reject missing, invalid, or non-positive connection limits.
  - Added checks to detect mismatches between test-pool settings and database capacity.

- **Documentation**
  - Documented the new test-pool connection limit setting and its defaults.
  - Added operational guidance for integration-test connection and lock budgets.


tinnhlt-gra:
## UAT — recorded terminal session

Captured with `script(1)`. Six steps: the guard red against the configuration that shipped, red again on the arithmetic once the stock 100 is declared explicitly, green against the fix, the live cluster agreeing with the committed number, the runtime half proving the pool takes the ceiling, and the inverse mutation proving that assertion can still fail.

```
== 1. RED — restore origin/main's compose file (no max_connections at all) ==
--- FAIL: TestTheLaneFitsInsideTheClusterItRunsAgainst (0.00s)
    laneconnbudget_test.go:128: ../infra/docker-compose.dev.yml passes no `-c max_connections=…` to postgres, so the lane runs against the stock 100 — which is the state #1109 reported
FAIL
FAIL	github.com/gradionhq/margince/backend	0.624s
FAIL

== 2. RED, the arithmetic half — declare the stock 100 explicitly ==
--- FAIL: TestTheLaneFitsInsideTheClusterItRunsAgainst (0.00s)
    laneconnbudget_test.go:139: the integration lane can demand 408 connections and ../infra/docker-compose.dev.yml allows 100.
        
            INTEGRATION_JOBS                  16   ../.github/workflows/ci.yml
          x LANE_CONNS_PER_PACKAGE            24   ../scripts/test-integration-parallel.sh
          + one admin connection per slot      1   CREATE/DROP DATABASE at a handover
          + LANE_FIXED_CONNS                   8   ../scripts/test-integration-parallel.sh
          ----------------------------------------
                                             408   demanded
                                             100   max_connections
        
        Raise max_connections in ../infra/docker-compose.dev.yml to at least 408, or lower a term. Do not leave them
        apart: they were unrelated numbers once, and the lane failed at connect time in a
        different package set every run (#1109).
FAIL
FAIL	github.com/gradionhq/margince/backend	0.376s
FAIL

== 3. GREEN — the fix restored ==
--- PASS: TestTheLaneFitsInsideTheClusterItRunsAgainst (0.00s)
--- PASS: TestTheBudgetGateRefusesTheConfigurationThatShipped (0.00s)
--- PASS: TestTheLanePinsThePoolCeilingItBudgetsFor (0.00s)
ok  	github.com/gradionhq/margince/backend	0.368s

== 4. The live cluster agrees with the committed number ==
max_connections=408  lock_slots=208896  shared_memory_size=230MB

== 5. The runtime half: the pool actually takes the ceiling the lane hands it ==
=== RUN   TestTheClusterSeatsTheBudgetTheLaneComputed
--- PASS: TestTheClusterSeatsTheBudgetTheLaneComputed (1.88s)
=== RUN   TestTheSharedPoolTakesTheCeilingTheLaneHandedIt
--- PASS: TestTheSharedPoolTakesTheCeilingTheLaneHandedIt (0.01s)
ok  	github.com/gradionhq/margince/backend/internal/platform/testdb	2.399s

== 6. INVERSE — the harness reads the ceiling and drops it (mutation) ==
    laneconnbudget_integration_test.go:88: the shared pool's MaxConns is 16, want 8 from MARGINCE_TEST_POOL_MAX_CONNS — the lane budgeted for 8 per pool and the pool is free to open 16, so the lane's whole connection arithmetic is out by the difference times INTEGRATION_JOBS
--- FAIL: TestTheSharedPoolTakesTheCeilingTheLaneHandedIt (1.77s)
(mutation reverted)
```

### The lane itself, four consecutive runs at the shipped configuration

```
########## GREEN RUN 1 — default concurrency ##########
  0s  then ./internal/compose/integration waited for a slot
  265s  ./internal/compose/integration occupied a slot, of which 259.6s was its tests
  5.4s  provisioning that slot (clone, compile, process start) — not priced above
OK: integration passed with 0 skips (41 packages, parallel)
########## GREEN RUN 2 — default concurrency ##########
  0s  then ./internal/compose/integration waited for a slot
  215s  ./internal/compose/integration occupied a slot, of which 210.2s was its tests
  4.8s  provisioning that slot (clone, compile, process start) — not priced above
OK: integration passed with 0 skips (41 packages, parallel)
########## GREEN RUN 3 — default concurrency ##########
  0s  then ./internal/compose/integration waited for a slot
  257s  ./internal/compose/integration occupied a slot, of which 252.9s was its tests
  4.1s  provisioning that slot (clone, compile, process start) — not priced above
OK: integration passed with 0 skips (41 packages, parallel)
########## GREEN RUN 4 — INTEGRATION_JOBS=16 (CI's setting) ##########
  0s  then ./internal/compose/integration waited for a slot
  191s  ./internal/compose/integration occupied a slot, of which 183.7s was its tests
  7.3s  provisioning that slot (clone, compile, process start) — not priced above
OK: integration passed with 0 skips (41 packages, parallel)
=== peaks across all four runs ===
peak per-package=16  peak lane-total=78  peak cluster=86  peak per-(package,role)=12  samples=10169
```

Three at the local default and one at CI's `INTEGRATION_JOBS=16`, each ending `OK: integration passed with 0 skips`. The peak per package is 16 against the allowance of 24, and the cluster peaks at 86 against 408 — measured every 100ms from a connection outside the lane's own transactions.

**Not a screen recording:** this change has no UI, and every observable it produces is terminal output. The artifact above is the session capture rather than a description of one.

tinnhlt-gra:
## Review response

Fable's review found nine things. Eight were real and are fixed; the ninth needed a correction to the finding itself. Each was verified before it was acted on.

| Finding | Verdict | What changed |
|---|---|---|
| `t.Skipf` makes `make test-integration-serial` permanently red | **Confirmed** — `backend/Makefile` greps its own log for `--- SKIP` and exits 1, and these were the only `t.Skip` calls in the whole integration lane | The terms and the budget moved to `scripts/lib-testdb.sh`; all three lanes call `declare_lane_budget` (parallel with its concurrency, one-package and serial with 1). Both tests now **fail** on an absent variable rather than skipping, exactly as `sharedAppPool` already fails on an absent DSN |
| The pin does not reach 49 files that open their own pool | **Confirmed** | The term now says so in as many words — a *measured budget*, not a proved ceiling — and **#1744** carries closing it. The PR's claim is narrowed to match |
| The formula is stated twice | **Confirmed** | The Go gate now **runs** `declare_lane_budget` instead of re-implementing `jobs*(perPackage+1)+fixed`. One expression |
| The failure message merges two causes | **Confirmed** — `INTEGRATION_JOBS` is a documented knob, so `INTEGRATION_JOBS=32` against a current container was told to recreate the container, which cannot help | The message branches on `INTEGRATION_JOBS` |
| The mirror does not cover an inverted comparison | **Confirmed** | The comparison is now `fits`, and both directions are exercised by name. Inverting it turns the mirror red (shown below) |
| The compose comment says "one package at 24" | **Confirmed** — 24 is the allowance, 16 is the measurement | Corrected to 16 and 86 |
| `'"'"'` debris in the comment | **Confirmed for my line**, which moved to `lib-testdb.sh` and was rewritten clean. The one remaining occurrence (`test-integration-parallel.sh:570`) is pre-existing and outside this diff — `git diff origin/main` adds none | — |
| `pool.go`'s header still says `max_connections=100` | **Confirmed** | Rewritten: how many connections the server has is the lane's arithmetic, not a fact that file may assume |

### Re-verification

```
$ grep -rn "t\.Skip" --include="*_integration_test.go" backend/ | grep -v "\.Skipped"
(no matches)

$ make -C backend -n test-integration-serial | grep -o "LANE_CONN_BUDGET=[0-9]* MARGINCE_TEST_POOL_MAX_CONNS=[0-9]*"
LANE_CONN_BUDGET=33 MARGINCE_TEST_POOL_MAX_CONNS=8

$ make test-it DIR=backend/internal/platform/testdb RUN='TestTheClusterSeats...|TestTheSharedPool...'
--- PASS: TestTheClusterSeatsTheBudgetTheLaneComputed (1.18s)
--- PASS: TestTheSharedPoolTakesTheCeilingTheLaneHandedIt (0.01s)
```

Inverting `fits` — the defect the mirror could not previously see:

```
$ sed -i '' 's/return demand <= maxConns/return demand >= maxConns/' backend/laneconnbudget_test.go
--- FAIL: TestTheBudgetGateRefusesTheConfigurationThatShipped
    the budget arithmetic makes the pre-#1109 configuration fit (408 <= 100)
```

`make check` exit 0 and `INTEGRATION_JOBS=16 make test-integration` → `OK: integration passed with 0 skips (41 packages, parallel)`, both after these changes.

tinnhlt-gra:
@coderabbitai review

Issue #548, opened 2026-08-07
people contention probes are bounded by probe count, not time, so a loaded runner fails them


tinnhlt-gra:
## Symptom

Two integration shards failed simultaneously on #537 with the same message, in two different tests:

```
ensurechannel_contention_integration_test.go:411: no backend waited on the held row within
20000 probes — the writer never reached the lock, so this run proved nothing
--- FAIL: TestTwoMessagesReportingTheSameRenameAuditItOnce

orgnamelock_contention_integration_test.go:108: no backend waited on the held row within
20000 probes — the writer never reached the lock, so this run proved nothing
--- FAIL: TestAnEvidenceApplyClearingANameWaitsOnTheNameLock
```

Both re-ran green. Neither test has anything to do with that PR's diff (migrations, docs, and the lane script).

## Cause

`waitUntilBlockedBy` (`backend/internal/modules/people/ensurechannel_contention_integration_test.go:309`) busy-reads `pg_stat_activity` waiting for a second backend to block, and gives up after a fixed **iteration** count:

```go
// Generous enough that a loaded machine cannot trip it, small enough that a
// genuine miss reports in seconds rather than minutes.
const maxProbes = 20_000
```

The comment states the intent but the bound does not deliver it. The loop is a race between two quantities that both degrade under load:

- how fast 20,000 probe round-trips to Postgres complete, and
- how long the contending goroutine takes to be scheduled and reach the lock.

On an idle machine the probes are fast and the writer is quick, so the writer wins easily. On a loaded runner both slow down, and there is no reason the ratio holds — the probe loop can burn its whole budget before the writer is scheduled at all. A count is not a duration, so "generous enough that a loaded machine cannot trip it" is not a property this bound has.

This is the T11 real-clock-flakiness trap in a quieter form: replacing a sleep with an iteration count removes the literal `time.Sleep` without removing the timing dependency.

## Worth keeping

The failure mode itself is right, and should survive whatever fix lands: the probe **fails loudly** rather than passing having proved nothing, and the `pid` scoping is correct for a cluster-wide `pg_stat_activity` under the parallel lane. The bug is only in how the give-up point is chosen.

## Options

1. Bound by wall time with a generous ceiling (e.g. 30s) via `context.WithTimeout` on the probe queries — still fails loudly, but the budget means the same thing on every machine.
[...]

Issue #516, opened 2026-08-06
Flaky: TestASecondCreateSharingAPhoneStillRecordsTheReviewPair times out under shard load


dungbanguyen:
`backend/internal/modules/people/creatededupe_contention_integration_test.go` failed on an unrelated PR's CI run with:

```
creatededupe_contention_integration_test.go:158: within 20000 probes the second create neither waited on the held number nor returned
```

It passes locally in 0.12s; in CI it burned 4.55s and then gave up. The test deliberately races two creates sharing a phone number and polls a bounded number of probes waiting for the second create either to block on the held row or to return. That budget is a proxy for "the contention window should be observable by now", and under twelve parallel integration shards on one runner the background create can simply not be scheduled within it.

The behaviour under test is real and worth keeping — a silently-lost review pair is the bug it guards. The flake is in how the wait is bounded: a probe count is a stand-in for scheduling progress, and it does not hold when the machine is saturated.

Options: wait on the lock becoming observable in `pg_locks` rather than counting probes; or drive the racer through an explicit synchronisation point so the window is deterministic instead of scheduled.

Seen on the CI run for PR #513, which does not touch `modules/people`. Re-running the job passed.

dungbanguyen:
Another instance of this shape, recorded here rather than as a new issue since the mechanism looks identical.

`TestTwoMessagesReportingTheSameRenameAuditItOnce` (`internal/modules/people`) failed on **shard 7/12** at **5.95s**, in [run 31145302586](https://github.com/gradionhq/margince-poc-v1/actions/runs/31145302586) on PR #550, and passed on a `--failed` rerun of the same commit with nothing changed. The full local lane passed on that commit both before and after.

Shared characteristics with the test this issue names:

- `internal/modules/people`, on the identity/dedup chokepoint
- a duration in the multi-second range for a test that is sub-second when it passes, which points at a lock or poll timeout being reached rather than at the assertion
- fails only under shard load; green in isolation and green in the unsharded lane

Both tests are proving a concurrency invariant — "a second create sharing a phone still records the review pair", "two messages reporting the same rename audit it once" — so whatever bound they rely on to serialise the racing writers is what shard load is pushing past. If the fix here is to replace a fixed timeout with something the test controls, it is likely the same fix in both places.

No action taken; this is a data point for whoever picks the issue up.

tinnhlt-gra:
Verified against `main` (`bcaeda896`): **this is the same mechanism as #548 and it is fixed.**

[...]

PR #1746, opened 2026-08-18
test(people): the lock ordering is read from the lock manager, not from a clock

### LINKED ISSUE #492: Flaky: the org name-lock ordering test proves its invariant with a 5s lock_timeout
`TestAnEvidenceApplyClearingANameWaitsOnTheNameLock` in `backend/internal/modules/people/orgnamelock_contention_integration_test.go` failed one integration shard on a branch that does not touch the file, and passes 3/3 locally. The failure is a timing threshold, not the invariant the test is about.

## What the test asserts

The apply must take the organization name lock *before* the organization row, so it can never hold that row while waiting for the lock — the cycle a concurrent human rename completes into a deadlock. The test proves *where* the apply is parked by having the lock holder reach for the organization row: if the apply is parked before touching the row, that succeeds at once.

## Why it is timing-dependent

The proof is armed with a fixed wall-clock budget:

```go
if _, err := holder.Exec(ctx, `SET LOCAL lock_timeout = '5s'`); err != nil { ... }
if _, err := holder.Exec(ctx, `SELECT id FROM organization WHERE id = $1 FOR UPDATE`, orgID); err != nil {
    t.Fatalf("the rename could not take the organization row while the apply waited (%v) — ...")
}
```

A correct implementation is supposed to make that `SELECT ... FOR UPDATE` return immediately, so 5 seconds reads as generous. But the shard lane runs 29 packages in parallel against one cluster, and the observed failure took **4.67s** — right at the boundary. Under that load the statement can exceed 5s while the apply is parked exactly where it should be, so the timeout fires and the test reports the deadlock ordering it was written to catch.

The failure mode is the wrong way round: the message names a real and serious defect, so a red run here will be read as "the lock ordering regressed" when it means "the database was busy". That is worse than a plain flake.

Related: the same lane's intermittent `SQLSTATE 53200` is #482. Same root cause class — a shared cluster under 29-way parallelism — different symptom.

## Direction

The timeout is doing two jobs: bounding the test, and *being* the assertion. Splitting those is the fix. Options, roughly in order of preference:

1. Ask Postgres where the waiter actually is instead of inferring it from elapsed time — `pg_locks` for the apply's backend shows whether it holds the organization row while waiting on the advisory lock. That is the invariant stated directly, and it has no clock in it.
2. If the timeout stays, make an exceeded budget distinguishable from a violated ordering: on timeout, check `pg_locks` before failing, and report "the lane was too loaded to tell" separately from "the apply holds the row".

Option 1 also removes the `waitUntilBlockedBy` polling helper from the proof path.

Found while merging `main` into an unrelated branch; `git log` shows the file last changed by #377.

tinnhlt-gra:
`TestAnEvidenceApplyClearingANameWaitsOnTheNameLock` proved **where** the apply is parked by arming a 5s `lock_timeout` on the holder and reaching for the organization row: a quick return meant "parked before the row", a timeout meant "parked while holding it".

```go
if _, err := holder.Exec(ctx, `SET LOCAL lock_timeout = '5s'`); err != nil { … }
if _, err := holder.Exec(ctx,
    `SELECT id FROM organization WHERE id = $1 FOR UPDATE`, orgID); err != nil {
    t.Fatalf("the rename could not take the organization row while the apply waited (%v) — …")
}
```

**The timeout was doing two jobs: bounding the test, and *being* the assertion.** Splitting them is the fix. The observed failure took **4.67s** — right at the boundary — under a lane running 29 packages against one cluster.

And the failure mode is the wrong way round, which is why this is worth more than its label. The message names a real and serious defect, so a red run reads as *"the lock ordering regressed"* when it means *"the database was busy"*. A flake that cries wolf about a deadlock ordering is worse than a plain flake: the next person either burns a day on a live bug that isn't there, or learns to disbelieve the one message that must never be disbelieved.

## What replaces it

Postgres is asked directly. `pg_locks` shows the backend queued on the advisory lock, and every relation lock that backend holds — so *"parked on the name lock, and holding nothing on `organization`"* is **stated** rather than inferred, and **there is no clock in it**.

Two facts are asserted and both are needed: that the waiter is parked on the advisory lock says it reached the name lock at all; that it holds no write lock on `organization` says it hadn't touched the row first. Either alone passes over the defect.

⚠ `pg_locks` is **exempt** from the frozen-snapshot trap that #970 fixed in this package's other probes — it reads the lock manager directly, not the `pg_stat_activity` statistics snapshot, which is materialized once per transaction and cached until it ends. That is *why* this is the honest fix rather than merely a different poll, and the code comment says so, because the next author will otherwise assume the two views behave alike.

## The second half: #970's trap, made into a gate

#970's fix carries a trap held only in one engineer's head, and the memory of it is explicit that the obvious reading is wrong:

> Never conclude "the probe connection is not a transaction" from the probe's own code — it is a property of **every call site**.

`TestEveryContentionProbeClearsTheStatsSnapshot` derives that obligation from the tree: any function asking `pg_blocking_pids` must clear the snapshot first. **It was red on arrival, with two live subjects:**

- `approvals`' `TestTheRowLockProbeSeesABackendThatDialsAfterItsFirstLook` took its first look through `e.owner` — a bare `*pgx.Conn`, which *looks* exempt — while `competingTx` had opened a transaction on that same connection. **The test whose entire subject is a probe that cannot see a late-dialling backend was itself blind in exactly that way.**
- `compose`'s `waitForBlockedOn` probes a pool, where the snapshot is fresh today. It clears anyway: which connection a caller hands it is not a property that file can assert, and that is the whole lesson.

## The proof

| Step | Result |
|---|---|
| **RED**, the call-site guard | red against today's tree, naming both live subjects |
| **GREEN** | both fixed |
| **GREEN**, the ordering assertion | passes |
| **GREEN under load** | 40 busy loops on 10 cores — **0.11s**, where the 5s timeout was the failure boundary |
| **INVERSE** | inverting the production ordering turns it red, naming the mode it found: `RowShareLock` |

**Worth recording:** the *first* inverse attempt mutated `writeCanonicalOrgColumn`, which is not the path this apply takes — and the test **passed**. A vacuous gate looks exactly like a working one from the passing side, which is the entire reason the inverse step exists.

## Which of #492's three parts this closes

The issue accumulated three separate defects. **Only the first was still live**, verified against `main` before starting, and this PR fixes only that one:

| Part | State | This PR |
|---|---|---|
| The 5s `lock_timeout` **is** the assertion | **LIVE** | **fixed** |
| The probe helper's count-based bound (`maxProbes = 20_000`) | Fixed by #901 + #970 — `grep maxProbes` across `backend/`: no matches. #548 and #516 were closed on it | not re-fixed |
| The fixed host port 55432 colliding when two shards share a runner | Fixed — Postgres publishes `15432:5432` and `make check-host-ports` holds the floor | not re-fixed |

## Coverage

Test-only change: **no measured new-code coverage** by construction. The substitute proof is the red → green → inverse sequence above plus the call-site guard's own red run.

## Review status

Fable (`/code-review`) and CodeRabbit have run. **Codex has not** — every `/codex:*` command is `disable-model-invocation: true`, so an agent cannot invoke it; a human has to type `/codex:review`. Two reviewers, not three, and this note is here so it does not read as three.

Closes #492

---
## Summary by cubic
Asserts org-name lock ordering by reading Postgres’s lock manager instead of a 5s timeout, and adds a gate that prevents stale `pg_stat_activity` snapshots from blinding `pg_blocking_pids` probes.

- People tests: replace the timeout-and-select with assertParkedBeforeTheOrganizationRow, which reads `pg_locks`, predicates on the current database, and requires the holder to have exactly one advisory lock to avoid ambiguous joins; treats `pgx.ErrNoRows` as “waited on a different lock,” not an I/O failure.
- Gate: TestEveryContentionProbeClearsTheStatsSnapshot now requires three conditions—clear runs before the probe, on the same receiver, and not through a pool—and walks `backend`, `../extensions`, and `../fixtures`; it scopes like the toolchain does (excludes `testdata`, hidden/underscore dirs, `vendor`, `node_modules`, `build`) and names which condition is missing.
- Fix call sites: approvals and compose tests acquire one connection, call `pg_stat_clear_snapshot()`, then probe, so pool/transaction context cannot blind the probe.

Closes #492.

<sup>Written for commit fe2d73a2fbaceae014321bd1d142b28d099c9ea6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1746?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved lock-contention checks to consistently reflect current database activity.
  * Added validation that waiting operations do not acquire unintended write locks.
  * Enhanced diagnostics when contention checks encounter errors.

* **Tests**
  * Added automated coverage to ensure contention probes refresh database statistics before checking blocking activity.

### DIFF
--- a/backend/contentionprobe_test.go
+++ b/backend/contentionprobe_test.go
@@ -1,0 +1,141 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+package backendarch
+
+// A contention probe that cannot see the backend it is waiting for.
+//
+// pg_stat_activity is not a live view. Its row set is materialized once per
+// TRANSACTION and cached until that transaction ends, so a probe issued inside
+// one is answering from a snapshot taken before the racer it is watching for
+// existed. A racer whose pooled connection is dialled mid-race is then invisible
+// FOREVER — at any budget, on any machine. That is #970, and it cost two issues
+// (#548, #516) that read as timeouts and were not.
+//
+// The trap is that this is a property of the CALL SITE, never of the probe's own
+// code, and the obvious reading of a call site is wrong. In approvals,
+// waitForRowLockWaiter looked exempt because e.owner is a bare *pgx.Conn — but
+// competingTx opens a transaction ON THAT SAME CONNECTION, so pgx ran every
+// probe inside it. Nothing in the probe's file says so.
+//
+// So the obligation is derived rather than remembered: pg_stat_clear_snapshot()
+// drops the cache, and any function that asks pg_blocking_pids must call it. It
+// is cheap, it is a no-op where the snapshot was already fresh, and making it
+// unconditional is what removes the question a reviewer would otherwise have to
+// re-answer at every new call site. A test added next month that passes a
+// transaction-bound connection reintroduces #970 silently; this is what stops it.
+//
+// Scoped to pg_blocking_pids and not to pg_stat_activity generally: counting
+// sessions by role or database (extmigrategate) is a census, not a race, and a
+// stale answer there is not a blind one.
+
+import (
+	"fmt"
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"io/fs"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"testing"
+)
+
+const (
+	// blockingPidsProbe is what makes a query a contention probe: it asks who is
+	// waiting on whom, which is the question the stale snapshot answers wrongly.
+	blockingPidsProbe = "pg_blocking_pids"
+
+	// snapshotClear is the call that makes the next read live.
+	snapshotClear = "pg_stat_clear_snapshot"
+)
+
+// TestEveryContentionProbeClearsTheStatsSnapshot walks every Go tree in this
+// module and fails on a function that asks pg_blocking_pids without first
+// dropping the transaction's cached view of who is connected.
+func TestEveryContentionProbeClearsTheStatsSnapshot(t *testing.T) {
+	var offenders []string
+	fset := token.NewFileSet()
+
+	walkErr := filepath.WalkDir(".", func(path string, entry fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if entry.IsDir() {
+			// Vendored and generated trees are nobody's call site.
+			if name := entry.Name(); name == "vendor" || name == "node_modules" || name == "build" {
+				return filepath.SkipDir
+			}
+			return nil
+		}
+		if !strings.HasSuffix(path, ".go") {
+			return nil
+		}
+		file, parseErr := parser.ParseFile(fset, path, nil, parser.SkipObjectResolution)
+		if parseErr != nil {
+			// Not skipped. A file this gate cannot read is a file it cannot
+			// clear, and a census that quietly drops its unreadable members
+			// reports "no offenders" for a tree it never saw.
+			return fmt.Errorf("parsing %s: %w", path, parseErr)
+		}
+		for _, decl := range file.Decls {
+			fn, isFunc := decl.(*ast.FuncDecl)
+			if !isFunc || fn.Body == nil {
+				continue
+			}
+			probes, clears := scanForProbe(fn)
+			if probes && !clears {
+				offenders = append(offenders, path+": "+fn.Name.Name)
+			}
+		}
+		return nil
+	})
+	if walkErr != nil {
+		t.Fatalf("walking the module for contention probes: %v", walkErr)
+	}
+
+	if len(offenders) > 0 {
+		t.Fatalf(`%d contention probe(s) ask %s without calling %s() first:
+
+  %s
+
+pg_stat_activity's row set is materialized once per transaction and cached until
+it ends, so a probe issued inside one cannot see a backend that dialled after the
+snapshot was taken — at any budget, on any machine (#970). Whether the connection
+is inside a transaction is a property of the CALL SITE and not of the probe: in
+approvals it looked exempt because the field is a bare *pgx.Conn, and competingTx
+opens a transaction on that same connection.
+
+Add `+"`"+`SELECT `+snapshotClear+`()`+"`"+` immediately before the probe. It is cheap, and where
+the snapshot was already fresh it does nothing.`,
+			len(offenders), blockingPidsProbe, snapshotClear, strings.Join(offenders, "\n  "))
+	}
+}
+
+// scanForProbe reports whether a function's string literals contain a
+// contention probe and whether they contain the snapshot clear.
+//
+// Read from the literals rather than from the call graph on purpose: the query
+// text is where both facts actually live, and a probe assembled from a helper
+// this gate could not follow would read as absent — a census that passes by
+// seeing nothing.
+func scanForProbe(fn *ast.FuncDecl) (probes, clears bool) {
+	ast.Inspect(fn.Body, func(node ast.Node) bool {
+		lit, isLit := node.(*ast.BasicLit)
+		if !isLit || lit.Kind != token.STRING {
+			return true
+		}
+		value, err := strconv.Unquote(lit.Value)
+		if err != nil {
+			value = lit.Value
+		}
+		if strings.Contains(value, blockingPidsProbe) {
+			probes = true
+		}
+		if strings.Contains(value, snapshotClear) {
+			clears = true
+		}
+		return true
+	})
+	return probes, clears
+}
--- a/backend/contentionprobe_test.go
+++ b/backend/contentionprobe_test.go
@@ -133,7 +133,7 @@ func scanTree(root string, fset *token.FileSet) ([]string, error) {
 			return err
 		}
 		if entry.IsDir() {
-			if name := entry.Name(); name == "vendor" || name == "node_modules" || name == "build" {
+			if skipDir(entry.Name()) {
 				return filepath.SkipDir
 			}
 			return nil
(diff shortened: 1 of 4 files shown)


tinnhlt-gra:
## UAT — recorded red → green → inverse

Five steps, in order: the call-site guard red against today's tree, green after both call sites are fixed, the clock-free ordering assertion green, green again under the load that used to trip the 5s timeout, and the inverse — the production ordering deliberately broken.

```
== 1. RED — the call-site guard against today's tree (before any fix) ==
--- FAIL: TestEveryContentionProbeClearsTheStatsSnapshot (0.59s)
    contentionprobe_test.go:98: 2 contention probe(s) ask pg_blocking_pids without calling pg_stat_clear_snapshot() first:
        
          internal/compose/sendvoiceoutcome_integration_test.go: waitForBlockedOn
          internal/modules/approvals/lockprobe_integration_test.go: TestTheRowLockProbeSeesABackendThatDialsAfterItsFirstLook
        

== 2. GREEN — both call sites fixed ==
ok  	github.com/gradionhq/margince/backend	0.921s

== 3. GREEN — the clock-free ordering assertion ==
=== RUN   TestAnEvidenceApplyClearingANameWaitsOnTheNameLock
--- PASS: TestAnEvidenceApplyClearingANameWaitsOnTheNameLock (0.11s)
=== RUN   TestAnEvidenceApplyWithoutANameDoesNotWaitOnTheNameLock
--- PASS: TestAnEvidenceApplyWithoutANameDoesNotWaitOnTheNameLock (0.05s)
ok  	github.com/gradionhq/margince/backend/internal/modules/people	0.797s

== 4. GREEN under the SAME load that used to trip the 5s timeout ==
--- PASS: TestAnEvidenceApplyClearingANameWaitsOnTheNameLock (0.11s)
ok  	github.com/gradionhq/margince/backend/internal/modules/people	0.418s

== 5. INVERSE — take the organization row BEFORE the name lock ==
   (production ordering inverted)
    orgnamelock_contention_integration_test.go:116: the apply is parked on the name lock while already holding [RowShareLock] on the organization table — it took the row a concurrent rename would edit BEFORE the lock that rename waits on, which is the cycle that deadlocks
--- FAIL: TestAnEvidenceApplyClearingANameWaitsOnTheNameLock (0.12s)
FAIL
FAIL	github.com/gradionhq/margince/backend/internal/modules/people	0.730s
FAIL
   (production restored)
```

Step 4 is the one that matters for a clock-free claim: **a clock-free assertion that was only ever run on an idle machine has not been shown to be clock-free.** Under 40 busy loops on 10 cores it completes in 0.11s, where the old proof's failure boundary was 5s.

Step 5 is what proves the assertion was replaced rather than deleted. It names the lock mode it actually found (`RowShareLock`) rather than reporting a boolean.

**Not a screen recording:** this change has no UI and every observable it produces is terminal output.

tinnhlt-gra:
## Review response

All five findings were real. Verified each, then fixed each.

**1 & 2 belong together, and the second is the one that matters.** The gate required only that `pg_blocking_pids` and `pg_stat_clear_snapshot` appear somewhere in the same function — co-occurrence, not reachability. Finding 1 is that hole exercised on arrival: `e.Pool.Exec(clear)` followed by `e.Pool.QueryRow(probe)` may touch two different connections, so the clear provably cannot bind, and the comment beside it claimed the opposite. **A gate certifying a call site it does not protect is this file's own defect, one level up.**

Three things now have to hold, and the gate names which one failed, because "no clear at all", "cleared afterwards" and "cleared through a pool" need different fixes:

```
== CONTROL ==                                        ok
== MUTATION: the clear moved AFTER the probe ==
  … asks pg_blocking_pids on `e.owner` with no pg_stat_clear_snapshot() on that receiver before it
== MUTATION: the clear moved to a DIFFERENT receiver ==
  … asks pg_blocking_pids on `e.owner` with no pg_stat_clear_snapshot() on that receiver before it
== MUTATION: the probe moved to a POOL receiver ==
  … probes through `e.pool`, a pool: each call may be handed a different connection, so no clear can bind this probe
```

The tightened gate then flagged the compose call site by itself, before any hand-editing — which is the order that proves it works. `waitForBlockedOn` now acquires one connection and issues both statements on it.

**3 — the advisory join had no database predicate.** Confirmed: `pg_locks` is cluster-wide, and the parallel lane runs a clone per package on one server. The follow-up query resolves `'organization'::regclass`, an OID local to *this* database, so a cross-database match would return no rows and the assertion would pass having examined the wrong backend — a false green on the exact ordering it exists to catch. Fixed with `h.database = w.database` and an explicit `w.database = current_database()`'s OID.

**4 — `pgx.ErrNoRows` was reported as a lock-manager read failure.** Confirmed, and it is the same misattribution class this PR removes from the file. The caller has already established that *something* is waiting, so no ungranted advisory row means the apply reached a different lock first — the ordering defect one statement earlier. It now has its own branch and its own sentence.

**5 — the walk stopped at the backend module.** Confirmed as a future gap (no probe exists outside `backend/` today). Extended to `backend/`, `extensions/` and `fixtures/`, which is what `license_test.go` already enumerates and for the same reason: each is its own module.

### Re-verification

```
--- PASS: TestAnEvidenceApplyClearingANameWaitsOnTheNameLock (0.20s)
--- PASS: TestAnEvidenceApplyWithoutANameDoesNotWaitOnTheNameLock (0.12s)
--- PASS: TestConcurrentSendsSharingADraftReferenceLeaveOneOutcomeAndBothTransmit (2.64s)
--- PASS: TestTheRowLockProbeSeesABackendThatDialsAfterItsFirstLook (0.06s)
```

`make check` exit 0 and `make craft-static` PASS after all of the above.

Thanks for checking the things that turned out sound as well — the parse coverage, the gate not flagging itself, and that the added clear does not defeat `lockprobe`'s deliberate freeze. Those were the three I would have had to re-derive.

tinnhlt-gra:
## CodeRabbit response

Both findings were real and are fixed.

**`contentionprobe_test.go` — `testdata` and tool-ignored directories.** Correct, and the distinction is worth stating because this file argues the opposite elsewhere. A parse failure inside a **compiled** tree still fails the gate: that member belongs to the census, and dropping it silently is the hole this whole file refuses. But `testdata`, `_`-prefixed and `.`-prefixed directories are the **Go toolchain's own** exclusions — it never builds them, so nothing in one can be a call site this gate is responsible for, and a deliberately unparsable fixture there would take the gate red for a file that is not code. That is scoping, not skipping, and the comment now says which.

**`orgnamelock` — report "no waiter" separately, and bind the waiter to the key.** The first half landed in the previous push (`pgx.ErrNoRows` now has its own branch: the caller has already established that *something* is waiting, so no ungranted advisory row means the apply reached a different lock first — the ordering defect, not an infrastructure fault).

The second half is now pinned rather than assumed. The join matches a waiter through **any** advisory lock the holder has been granted, which is unambiguous only while the holder holds exactly one — it does today, since `holdOrgNameLock` takes the name lock and nothing else. So that is asserted, and adding a second advisory lock to the holder turns it red:

```
orgnamelock_contention_integration_test.go:119: the lock holder holds 2 advisory locks,
  want exactly 1 — with more than one, the waiter lookup below cannot tell which key a
  backend is queued on, and could report a waiter on the wrong lock as parked on the name lock
```

Binding to the derived key directly would mean re-deriving `lockOrgNameWrites`'s key in the test, which is a second spelling of the thing under test. Asserting the holder's lock set is the same guarantee without that copy.

`make check` exit 0, `make craft-static` PASS.

Issue #1770, opened 2026-08-19
retention.applied.action is a free string on the wire, so the closed set exists only in the database


LarsGradion:
## What is true today

`PublicEventRetentionApplied.action` in `backend/api/public-events.yaml:1597` is declared as:

```yaml
action:
  type: string
  description: The action that ran (archive | anonymize | erase).
```

The three permitted values appear in the description only. There is no `enum`, so every generated client sees an unconstrained string.

The database does constrain it — `retention_policy.action` carries `CHECK (action IN ('archive','anonymize','erase'))` in `backend/migrations/core/0010_consent_retention.up.sql`, so nothing invalid can be authored or emitted today. The safety argument holds. What does not hold is the contract's half of it.

## Why it matters

A consumer generating types from the public event contract gets `string`. They cannot switch exhaustively over the action, and a compiler cannot tell them when a fourth value arrives. The whole point of a closed set on a public contract is that the consumer's build breaks when it widens; here it would not.

It also splits one rule across two places that can drift. The CHECK and the description have to be kept in agreement by hand, and nothing fails if they diverge — which is the same shape as the hand-maintained lists the review-loop rules already warn about.

## Failure scenario

1. A fourth retention action is added — say `restrict`, which already exists as an audit verb after migration `0287`.
2. The migration widens the CHECK. The contract description is updated or is not; either way the schema still says `type: string`.
3. A downstream consumer that switched on the three known values receives `restrict` and falls through to its default branch. Nothing in their build ever warned them, because the type never promised only three.
4. Their handling of a retention event is silently wrong for the new action.

## The fix

Add the enum to the schema:

```yaml
action:
  type: string
  enum: [archive, anonymize, erase]
  description: The action that ran.
```

Then the widening cannot be silent: adding a value is a contract change that `make contract-breaking-check` sees and a consumer's regenerated types reflect.

Worth checking at the same time whether the enum and the database CHECK can be derived from one source rather than maintained in two. A fitness test asserting they hold the same value set would be the smaller version of that.

## Provenance

Found while migrating the retention-event decision into `docs/adr/`. The decision states the action is a closed set; the implementation closed it in the database and not on the wire. `docs/adr/ADR-0117-the-retention-event-action-is-a-closed-set.md` records the outstanding half rather than describing the decision as fully delivered.

LarsGradion:
[...]

Issue #1768, opened 2026-08-19
An approval on a target with no version pin executes against whatever the row became


LarsGradion:
## What happens

`validateRedemptionTarget` in `backend/internal/modules/approvals/redeem.go` returns `nil` immediately when the staged approval carries no `TargetVersion`:

```go
if a.TargetVersion == nil || a.TargetID == nil || a.TargetType == nil {
    return nil
}
```

A staged action whose target type is absent from `versionTables` stages with a NULL pin, so this short-circuits and the action executes with the approval token check alone. Nothing re-reads the row. Whatever the record became between staging and redemption is what the approved action lands on.

The file's own comment already states this: *"A type missing here stages with a NULL pin, and every one of those three steps short-circuits silently — the human approves a row that anyone may then change before the authorized call lands."*

## Which targets are affected

`versionTables` covers person, organization, deal, lead, activity, offer, product, list, tag, relationship, project, saved view, offer template and webhook subscription.

Two real staged target types are **not** in it:

- `fx_rate` (`targetFxRate`) — currency conversion rates
- `ai_model_rate` (`targetAIModelRate`) — AI model pricing

Both are staged for approval with create and update actions (`backend/internal/modules/approvals/authority.go:138-142`) and both are visibility-probed (`targetvisibility.go:101-102`), so they are live approval targets, not vestigial constants.

## Why this matters

Both carry money. An approver reads a proposed FX rate or model price, approves it, and the value that commits is whatever the row holds at redemption — not what they saw. The window is however long the approval sits in the inbox.

## Failure scenario

1. An operator stages an update to `fx_rate` for EUR→USD at 1.08.
2. The approval lands in the inbox with no version pin, because `fx_rate` is not in `versionTables`.
3. Before it is decided, another actor updates the same rate row to 1.80.
4. The approver reads the staged proposal, sees 1.08, and approves.
5. The action executes. No version check runs. Every conversion downstream uses a rate nobody approved.

## What the decision intended

The retired specification for approval-token binding stated that a target with no version column should **fail closed** — the staged action strands in the inbox rather than executing unchecked. The implementation does the opposite: it executes. This was found while migrating that decision into `docs/adr/`, and the record now describes the actual behaviour with the divergence noted rather than repeating the spec's claim.

## Fixing it

Two options, and the choice is a product call:

[...]

PR #1774, opened 2026-08-19
docs(adr): the guardrail decisions move here, in plain language

LarsGradion:
First batch of the decision-record migration: 17 records covering privacy and retention, agent governance, and security. The remaining 40 follow in three batches.

## What is here

Each record is rewritten from the retired specification in plain language, numbers kept — roughly 2,500 comments in this tree cite them, and renumbering would be a large diff for no gain.

The section that matters is **What it binds in this repository**: real migrations, tables, code paths and test names. Every path and every test name in all 17 was verified to exist. That section is what turns a record from a claim into something the next reader can check.

## ADR-0120 closes a hole

The extension tier was built and shipped but never ratified. The draft sat on an unmerged branch numbered ADR-0069 — a number that already belonged to the embedding-width decision — so it could never land. `CLAUDE.md` and `AGENTS.md` cited that wrong number the whole time.

Both now cite ADR-0120, and both lose a second staleness while I was there: they said the tree ships two first-party extensions. It ships five.

## The index is generated

`scripts/gen-adr-index.sh` (`make adr-index`) builds `docs/adr/README.md` from the records. The system this replaces kept an index and its records in two places and they disagreed about which decisions were live. Here the records are the only input, so the index cannot claim something the tree does not contain.

`backend/adrindex_test.go` holds four properties:

- the index is not stale
- every record states a status
- every "Superseded by" names a record that exists
- no two records claim the same number

The fourth was added after the first three ran and I found a real duplicate — the same decision written twice under two slugs, which the generated index rendered as two ordinary rows with nothing signalling a problem. Mutation-checked in both directions.

The generator takes an output path so the staleness check compares against a scratch file rather than rewriting the file it is checking. A gate that overwrites its own subject corrupts the worktree the moment it fails partway through, and races with anything else doing the same.

## Four divergences recorded rather than repeated

Where the source claimed something the code does not do, the record says what is actually there:

- **Sharing grants** — the row-level-security half was never built, and RLS is now retired entirely by migration `0217`. `platform/auth/rowscope.go` is the only enforcement.
- **The retention event action** — closed by a database CHECK, but still `type: string` on the wire, so a generated client sees a free string. Filed as #1770.
- **The pre-erasure preview** — does not exist. The supplier-correspondence gap it was meant to cover is named as a gap.
- **Restore replay of restrictions** — a documented obligation with no routine behind it.

## Checks

`make check-backend` green. `craft static --strict` on the new test: 0 findings. The public-reference gate passes across the whole tree.

## Note on process

Two factual errors in earlier drafts were caught by review and fixed: ADR-0011 credited the wrong migration for two tables, and ADR-0036 misdescribed why rate-proposal approvals carry no version pin. The second also produced an issue I filed and then closed as incorrect (#1768) — a rate proposal targets the workspace price sheet, not a row, so there is nothing for a version pin to refer to. ADR-0036 now explains that rather than leaving it as an unexplained divergence.

---
## Summary by cubic
Migrates 17 guardrail ADRs to `docs/adr/` with preserved numbers and a generated index, and adds gates to keep the records true. Corrects ADR‑0018: core uses per‑statement tenant predicates while RLS remains only for extensions; the RLS‑claim gate now scans ADR prose by paragraph and ignores fenced code to catch wrapped claims.

- Adds 17 ADRs under `docs/adr/` with “What it binds in this repository”; introduces ADR‑0120 for the extension tier; updates `AGENTS.md` and `CLAUDE.md` to cite ADR‑0120 and list five first‑party extensions.
- Generates `docs/adr/README.md` via `scripts/gen-adr-index.sh`; new `make adr-index` target avoids clobbering during checks.
- Adds `backend/adrindex_test.go` gates: index is not stale; every record has a status; every “Superseded by” resolves; no duplicate numbers.
- Expands `backend/rlsclaims_test.go` to scan `docs/adr/`, match prose forms (e.g., “tenant isolation through row‑level security”), scan by paragraph to catch line‑wrapped claims, and skip fenced code; adds regression tests for both cases.
- Records divergences: no RLS layer for sharing grants; retention event action closed in DB but still open in the contract; no pre‑erasure preview; restore replay documented but not implemented.

**Required actions**
- When adding/changing ADRs, run `make adr-index` and commit the result; include a `**Status:**` line; ensure any “Superseded by” links resolve; take the next number from `origin/main`.
- Update any remaining references to the extension tier to ADR‑0120.
- Do not claim core reads are scoped by RLS; the gate scans code and ADRs.

<sup>Written for commit c80ee040e30d9ea012c601ad98e93ca4a6b3eced. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1774?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Documentation**
  * Added architecture decision records covering compliance, consent, security, approvals, sharing, AI governance, retention, extensions, and web search.
  * Added a centralized, generated index of architecture decisions.
  * Updated extension documentation to reflect the current five first-party extensions and governing decision.

* **Tools**
  * Added a command to regenerate the architecture decision index.

* **Tests**
  * Added validation for decision-record statuses, references, numbering, and index freshness.


LarsGradion:
I resolved six threads on this PR a moment ago and should not have: the fixes are committed on `docs/adr-migration-batch2`, which stacks on this branch, not on this branch itself. This PR still contains the text they flag. Reopening is not possible from the API, so this comment is the correction.

The findings were all valid and all fixed — just one branch up:

- **ADR-0114 read paths** — "leaves every ordinary read path" reads as "remains in", which inverts the guarantee. Now says removed, and names the two deliberate exceptions: the subject's own access export and the controller's restricted-records list. Both verified against `sar.go` and `restrictedlist.go`.
- **ADR-0114 immutability** — the record called the row immutable below every role, then described an admin release four paragraphs later. Migration `0289` refuses ordinary writes via trigger and lets exactly two through, so the record says that now.
- **ADR-0064 provenance** — said no provenance columns while its own bindings named `agent_run_id`. Both true of different things: no provenance about a customer record, operational linkage to the run.
- **ADR-0010 CRA dates** — "December 2027" and "starts earlier" replaced with the exact dates. The reporting duty applied from **11 September 2026**, which is already past. That precision surfaced a real gap: no runbook exists for a live 24-hour obligation, filed as #1785.
- **`adrindex_test.go` status validation** — the test checked that the string `**Status:**` appeared anywhere, accepting an empty value and the marker inside prose. Now anchored, with the value held to the taxonomy. Mutation-checked three ways; my first attempt still passed the empty case because a missing line and an empty value returned the same thing.

If this PR merges first the fixes arrive with batch 2 immediately after. If you would rather they land here, say so and I will move the commit down.

LarsGradion:
Superseded by #1865.

Issue #1769, opened 2026-08-19
retention.applied declares action as an open string where ADR-0117 closed the set


LarsGradion:
## What is wrong

ADR-0117 decided that the `action` field on `retention.applied` carries one of exactly three values: `archive`, `anonymize`, `erase`. The contract does not enforce that.

`backend/api/public-events.yaml`, in `PublicEventRetentionApplied`:

```yaml
action:
  type: string
  description: The action that ran (archive | anonymize | erase).
```

The three values live in the description, which is prose. The schema admits any string, and the generated Go and TypeScript types follow the schema, so nothing refuses a fourth value.

## Why it matters

A subscriber reading the contract sees an open field. That leaves them two bad choices: write a branch for values nobody has defined, or ignore what they do not recognise and lose an event silently. The whole point of closing the set was to let a consumer switch on it exhaustively.

Nothing writes a bad value today. `backend/migrations/core/0010_consent_retention.up.sql` holds a CHECK on `retention_policy.action` with the same three values, and every emit site in `backend/internal/modules/privacy/` passes one of the three. So this is a gap between the decision and its enforcement, not a live defect.

Note that `restrict` is **not** a fourth action — restriction emits `retention.restricted`, a separate event. That part is correct as built.

## Failure scenario

A new retention action is added — say `redact` — and the emit site is written before anyone checks the event contract. The database CHECK refuses it if it reaches `retention_policy.action`, but an emit site that passes the value straight into the event payload bypasses that: the schema validates, the generated types compile, and the event ships. Every subscriber that switches on the three known values now silently drops it, and the drop looks identical to no event having been emitted.

## The fix

Declare the enum in `backend/api/public-events.yaml`:

```yaml
action:
  type: string
  enum: [archive, anonymize, erase]
  description: The action that ran.
```

Then `make gen`, and the generated types carry the constraint. This is an additive tightening within the major version — it refuses values the catalog already called impossible, so no conforming producer or consumer changes.

Worth doing at the same time: check whether the CHECK constraint and the contract enum can be derived from one source rather than maintained as two lists that happen to agree. Two hand-maintained copies of one set is the shape the review-loop rules already warn about.

LarsGradion:
Carried over from #1770, closed as a duplicate: the closed set genuinely exists today, just not on the wire. `retention_policy.action` holds a CHECK with the three values, and every emit site in `backend/internal/modules/privacy/` passes one of them. So there is no live defect — a fourth value cannot reach the event by accident today.

[...]

PR #1783, opened 2026-08-19
docs(adr): the contract, persistence and connector decisions move here

LarsGradion:
Stacked on #1774 — review that one first. This adds thirteen more records, taking the migration to 30 of 56.

Each is rewritten against the code rather than compressed from the old prose, and each names the files, tables, tests and make targets that make it checkable. Every cited path was verified to exist.

## Six divergences between the specification and the code

The value of reading these records against the tree rather than summarising them is that the specification turned out to describe things this repository does not do. Each record says what is actually built.

**The notifications engine (ADR-0056) does not exist.** The specification names a `notification` entity, a preference table, read and mark-read endpoints, and an in-app notification centre. There is no table, no module, and no path in the contract. The mail channel and account recovery beside it are real and fully built, so the record keeps those and drops the rest.

**ADR-0017's upgrade preflight was never built.** The decision rests on a `crm gen upgrade` command, an ownership-partitioned comparator and a three-bucket conflict classifier. None exist. What actually holds the fork seam is `scripts/check-migration-versions.sh`, the two namespace tests and `make contract-breaking-check`, so the record names those.

**The `modules/<name>/custom/` seam is declared and empty.** `CLAUDE.md` and `AGENTS.md` both promise it; `ls backend/internal/modules/*/custom` matches nothing. The record says so rather than implying a populated directory. The migration half is real — `backend/migrations/custom/` holds 48 files.

**The `x_` column prefix is enforced nowhere.** ADR-0002 and ADR-0017 both leaned on it as the fork-ownership signal, and one amendment built a collision guard on it. It survives in exactly one line of overlay provider field mapping, which is a different purpose. Both records bind to the seams that exist instead.

**ADR-0048 was wrong in three ways, and this is the security-critical one.** The specification names a `connector_secret` table owned by the identity module, a key-provider interface with a cloud KMS implementation, and envelope-wrapped data keys. The real thing is `vault_secret` under `backend/internal/platform/keyvault/`, sealed directly with AES-256-GCM under a root key read from the environment, with no KMS provider at all. Worth knowing and now recorded: `vault_secret` deliberately carries no `workspace_id` — the workspace lives inside the reference and inside the GCM additional authenticated data, so a reference from another tenant answers not-found rather than decrypting.

**ADR-0043 describes MFA and single sign-on as delivered.** Neither exists in the identity module. That is why `GET /auth/capabilities` was added: so the login surface asks the server what it can offer instead of advertising a flow the installation cannot complete.

Two smaller ones are recorded in their History sections: ADR-0063's single ordered nightly suite has no dispatcher (the passes are separate jobs on their own cadences, and one says so in its own reason text), and ADR-0060's source claims a row-security policy on `system_log` that migration 0217 retired.

## One triage correction

**ADR-0041 is not migrated.** It was classified as still-binding, but it governs a backlog validator, a 777-story dependency graph and a three-lens specification — `validate_backlog.py` and `backlog-graph.json` are both gone. The one rule from it that still binds, contract drift staying green, is already in ADR-0015. That takes the migrate total from 57 to 56.

## Checks

- Every cited path verified to exist. One was wrong — `config/ai-routing.yaml` is operator-created and gitignored, so the record now points at the example file it is made from.
- `TestPublicTreeCitesNothingPrivate`, `TestNoGoSourceClaimsRLSStillScopesARead`, and the three decision-record gates all pass.
- The index is regenerated: 30 records.

---
## Summary by cubic
Adds 13 rewritten ADRs, regenerates the ADR index, and tightens the ADR status test; records unbuilt decisions as owed work (notifications engine, nightly dispatcher, retention-action contract enum) with no runtime or schema changes.

- Review focus:
  - ADR-0036: moves the version-pin exception into the decision and updates the title; approvals that don’t target a row rely on authority checks rather than a version re-check.
  - ADR-0114: restricted records are removed from ordinary reads; only the subject’s access export and the controller’s restricted list can reach them; ordinary writes are refused by trigger, with only audited release/expiry allowed.
  - ADR-0064: `ai_call` carries no record provenance; `agent_run_id` is operational linkage.
  - ADR-0010: CRA dates clarified — CE marking from 11 Dec 2027; 24-hour exploited-vulnerability reporting duty from 11 Sep 2026.
  - `backend/adrindex_test.go`: requires an anchored `**Status:**` line and enforces the taxonomy (Active, Retired, Superseded by), allowing qualifiers; fails missing/empty/unknown values.

<sup>Written for commit 4caa214a256e14c3a63d1bea477f9609369e1286. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1783?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


LarsGradion:
**Correction on notifications, from Lars's review.**

I had dropped the notifications engine from ADR-0056 on the grounds that nothing implements it. That was wrong. The need is real and still wanted, so deleting it lost the fact that it is *owed* and left the ground unclaimed for something else to grow a competing answer.

It is back in simplified form, marked explicitly not final. The section sketches what a notification is, who it reaches and how a person tunes it — only far enough to hold the question — and says plainly that the entity shape, the preference model, the digest cadence and the read-state semantics all need a design pass before any code. That pass replaces the section in the same change.

[#1784](https://github.com/gradionhq/margince-poc-v1/issues/1784) carries the five product questions it has to answer. The sharpest is whether this replaces the approval inbox or sits beside it — two inboxes is a product decision, not an implementation detail.

**Two more records had the same shape**, and now say so in the status line where a reader sees it without opening the History:

- **ADR-0063** — the single ordered nightly dispatcher is unbuilt, while `backend/api/jobs.yaml` says the cadence "rides this job until the nightly dispatcher lands". Still wanted: a later pass reads what an earlier one wrote, so separate cadences make that ordering incidental rather than guaranteed.
- **ADR-0117** — the action set is held by the database CHECK and every emit site, but the contract still declares an open string. That is [#1769](https://github.com/gradionhq/margince-poc-v1/issues/1769).

The rule these three now follow: **a decision the code has not caught up with stays in the record and says which half is missing.** Deleting it hides the debt.

LarsGradion:
Superseded by #1865.

Issue #1788, opened 2026-08-19
Two modules can emit the same event type and no gate notices


LarsGradion:
## The gap

ADR-0014 says a module owns its tables, its event types and its cache key prefixes, and that continuous integration fails when any of the three has zero owners or more than one.

Only the first is gated. `backend/tableownership_test.go` derives table ownership from the tree and fails on a table with no owner or two. Nothing does the same for event types.

So two modules can write an outbox row carrying the same event type today, and no gate notices.

## Why it matters

The failure is quiet in the direction that hurts. A subscriber filtering on an event type starts receiving events from a second producer it was never designed for, with a payload shape it does not expect. Nothing errors — the consumer either mis-handles the payload or silently drops it, and both look like the first producer misbehaving.

It is also the exact defect the table-ownership test exists to prevent, one axis over. Tables got the gate because a table with two writers is a data-integrity problem. An event type with two producers is the same problem on the wire.

## Failure scenario

1. The `deals` module emits `deal.stage_changed` with a payload carrying the old and new stage.
2. Someone adds a stage-change side effect in another module and, reasonably, emits the same event type from there — with a payload that names the deal but not the stages, because that call site does not have them.
3. Nothing fails. The contract validates both, because the schema is per type, not per producer.
4. A subscriber reading `old_stage` gets it on some events and not others. The bug reads as intermittent data loss in the first module.

## The obvious shape

Extend `backend/tableownership_test.go`, or add a sibling, that reads the event type from each outbox write site and asserts exactly one module writes each type. The write sites are findable: every emit goes through `storekit.Emit`, so the type is an argument at a known call shape.

Cache key prefixes are the third axis ADR-0014 names and are equally ungated, but no defect has been observed there, so they are worth doing in the same change only if it is cheap.

## Not urgent

No known instance of a duplicated event type exists today — this is a missing guard, not a live break. Recorded because ADR-0014 claims the guard exists, and a claimed-but-absent gate is worse than an acknowledged gap: it stops people checking by hand.

LarsGradion:
Approved 2026-08-19. Build it.

Paired with #1802, which is the same species — a gate that reads as complete and is not. ADR-0014 says a module owns its tables, its event types and its cache key prefixes, with CI failing on a bad count. Only tables are gated.

Both derive their subject from the module ownership declarations in `doc.go` that `backend/tableownership_test.go` already reads, so they share most of their machinery and are worth doing as one change.

[...]

PR #1790, opened 2026-08-19
docs(adr): the architecture, layout and AI-runtime decisions move here

LarsGradion:
Stacked on #1783. Fourteen more records, taking the migration to 44 of 56.

## Eight records carry something the code has not caught up with

Each says so in its **status line**, so a reader sees which half is missing without opening the History. That rule came out of Lars's review of batch 2: a decision the code has not implemented stays in the record and names the gap, because deleting it hides the debt.

- **ADR-0005** — the delegated-worker protocol, for handing a scoped task to an external coding agent, has no route, no port, nothing. The same record retires the local stdio surface: only HTTP survives, because a static configured passport cannot honour a revocation mid-session.
- **ADR-0013** — no checked enumeration of the system services that sit below the auth gate. The rule is stated; nothing derives or verifies the exception list.
- **ADR-0020** — there is no in-product screen for a model-provider key. Keys reach the process only through environment variables; the sealed store holds connector credentials alone.
- **ADR-0022** — re-derivation over stored raw payloads does not exist. Backfill reaches further into the mailbox; it does not re-run derivation over what is already stored. Reciprocal rank fusion **is** built and proven by `TestHybridRRFAgreementWins`.
- **ADR-0023** — three of five decision points are unbuilt: no grade classifier, no security backport lane, no licence entitlement check.
- **ADR-0045** — `craft review`, the reviewing arm, is wired into nothing. No Makefile target, workflow job or hook calls it. Only `craft static` and `craft residue` run.
- **ADR-0054** — `modules/<name>/custom/`, the keystone of the fork seam, does not exist. The SQL half does.
- **ADR-0061** — SSO and MFA are absent, and one consequence is live: an admin can mint a set-password link for a member who already has one, with only detective controls, because the named mitigation needs the missing second factor.
- **ADR-0069** — the reindex tooling turned out fully built. What is missing is the retention sweep for vectors left behind by a superseded embedding identity.

## Two corrections to the rulebook itself

Writing these against the tree caught `CLAUDE.md` being wrong about this repository.

**It contradicted itself on the binary count.** Line 201 said three process-role binaries; the layout heading said four. There are three — `api`, `worker`, `migrate` — and the fourth retired with the standalone tool surface.

**It claimed a linter setting that is not set.** The craftsmanship section said the length ceiling agrees with golangci's `funlen`, "configured here `ignore-comments`". There is no `funlen` stanza in `backend/.golangci.yml`. The ceiling lives in `cli/craft`, and the whole-tree check is `scripts/check-go-file-length.sh`, which is what the sentence names now. Both files carry it identically, as the parity test requires.

## One issue filed

[#1788](https://github.com/gradionhq/margince-poc-v1/issues/1788) — ADR-0014 says a module owns its tables, event types and cache key prefixes, with continuous integration failing on a bad count. Only tables are gated. Two modules can emit the same event type today and nothing notices, which is the same defect the table-ownership test exists to prevent, one axis over.

## Checks

Every cited path verified to exist. `TestPublicTreeCitesNothingPrivate`, `TestNoGoSourceClaimsRLSStillScopesARead`, the three decision-record gates, and the shared-rulebook parity test all pass. Index regenerated: 44 records.

---
## Summary by cubic
Migrates 14 architecture and runtime decisions into ADRs and corrects rulebook inconsistencies; no runtime or schema changes. This makes unbuilt gaps explicit and aligns contributor docs with the current layout and craftsmanship checks.

- Adds 14 ADRs (index now 44/56) spanning agent loop and tool surface, web reading behind the connector seam, local/cloud model paths, one governed surface and auth, enforced module boundaries, bring-your-own inference, capture semantics and hybrid retrieval, graded patchability and security lane, craftsmanship gate, backend layout, single-organization installs, embedding width/reindex, and webhook payload contract. Each ADR’s Status calls out any unbuilt work; key items include: delegated-worker protocol, explicit system-service exceptions list, in-product model-key entry, re-derivation over stored payloads, patch-grade classifier/security fast lane/licence gate, unwired reviewing arm of `cli/craft`, missing `modules/<name>/custom/` fork seam, SSO/MFA, and the embedding-retention sweep.
- Fixes contributor docs: both `AGENTS.md` and `CLAUDE.md` now state three `cmd/<role>` binaries (per ADR-0054) and reference the whole-tree file-length check at `scripts/check-go-file-length.sh` instead of a non-existent golangci `funlen` setting.
- Review focus: skim ADR “What is owed” sections to confirm gap statements are accurate; verify corrected paths and terminology (`backend/internal/{shared,platform,modules,compose}`, `cli/craft`, `/mcp`). No rollout or migration actions required.

<sup>Written for commit 3abb078019dcbb1ee958d0590f40ce03803d342a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1790?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


LarsGradion:
Superseded by #1865.

Issue #1792, opened 2026-08-19
Object storage defaults to us-east-1 when the operator sets no region, and the bucket is created there


LarsGradion:
## What happens

`backend/internal/platform/blobstore/s3.go:47-49` falls back to `us-east-1` when `Config.Region` is empty:

```go
region := cfg.Region
if region == "" {
    region = "us-east-1"
}
```

That value is not inert. It reaches `MakeBucket` at line 81, so an operator who configures an endpoint and credentials but leaves the region unset gets a bucket created in a US region.

## Why it matters here

Data residency is a product commitment. ADR-0027 says this company runs no hosting and every installation is operator-run or self-hosted; ADR-0051 makes object storage sovereign by default. The default region contradicts the second one in exactly the case that matters: a real S3 endpoint with an unset region.

For MinIO — the local and self-hosted case — the region is arbitrary and this is harmless. For an operator pointing at genuine AWS S3, or any provider that honours the region on bucket creation, attachment bytes land in the US. Attachments carry whatever a customer emailed, which is the category that made residency a commitment in the first place.

The failure is silent. Nothing warns, and nothing afterwards reports which region the bucket is in.

## Failure scenario

1. An operator in Germany sets `MARGINCE_BLOBSTORE_ENDPOINT` to an S3 provider, plus access key, secret and bucket. They do not set a region, because nothing says they must.
2. The first attachment upload calls `MakeBucket` with `us-east-1`.
3. The bucket is created in a US region. Every attachment from then on is stored there.
4. The installation reports itself as sovereign, because nothing checks.

## Worth deciding, not just fixing

The narrow fix is to refuse to start when an endpoint is set and a region is not. That is the fail-closed shape the rest of the product already uses — an unlicensed boot refuses, a relay without TLS refuses — and it costs an operator one line of configuration.

Defaulting to an EU region instead would be worse: it makes a residency decision on the operator's behalf, and it is wrong for an operator who deliberately wants another region.

Behind it sits a larger question that ADR-0051 records as unbuilt: the blob endpoint is a plain environment variable, so changing where customer bytes live is neither audited nor a governed configuration change. That is the same class of decision as the embedding rebind, which ADR-0069 made a governed deployment action. Whether the blob endpoint deserves the same treatment is a product call.

## Also recorded

[...]

Issue #1791, opened 2026-08-19
The LICENSE says agents are not Seats; the seat count includes them


LarsGradion:
## The contradiction

`LICENSE` defines a Seat and then excludes agents explicitly:

> "Seat" means a single, identified natural person who is provisioned with credentials to access the Licensed Work […] The following do NOT count as Seats: (i) automated agents, AI systems, bots, and service accounts that act under the authority of, and are attributable to, a counted Seat […]

The runtime counts them. `fullSeatsInUseQuery` in `backend/internal/modules/identity/seatusage.go:87`:

```sql
SELECT count(*) FROM app_user
 WHERE seat_type = 'full' AND status NOT IN ('suspended', 'deactivated')
```

Every `full` seat is counted. The agent runner holds one. Nothing in the query or its callers excludes an agent or a service account.

So a customer reading the licence they signed and an operator reading `seats_used` get different numbers for the same installation.

## What it costs

On a fresh install the gap is one seat: the admin plus the agent runner reads as two against a licence that would say one. That is small in absolute terms and large in proportion — an installation entitled to ten seats can provision nine people, and nobody can tell from the product why.

It also runs the wrong way for us commercially. We are over-counting against our own licence text, so the first customer to read both carefully finds we billed or capped them for something the licence says is free.

## Why it is not simply a code bug

The code's position is defensible and the record says so: excluding agents lets an installation act without limit through agents, which is a real avoidance path. The licence's position is also defensible: an agent acting under a named person's authority is that person acting, and charging twice for one human is hard to justify.

Both cannot be true of one number. Which one changes is a commercial and legal decision, not an implementation detail.

## The options

1. **Change the query to match the licence** — exclude seats that are agents or service accounts. Needs a way to tell them apart; `seat_type` currently does not carry that distinction, so this is a schema change plus a migration, not a one-line filter.
2. **Change the licence to match the code** — agents count. Simplest technically, and it is the position the code already takes. Needs counsel, and it is a customer-facing term change.
3. **Count agents but cap them separately** — a person seat and an agent allowance as two numbers. Closest to what both texts are reaching for, most work.

Option 2 is the smallest change and the one I would expect to win, but it is squarely a decision for Lars and counsel.

## Also recorded

[...]

PR #1793, opened 2026-08-19
docs(adr): the residency, jurisdiction and platform decisions complete the migration

LarsGradion:
Stacked on #1790. Thirteen records, and with them **all 56 decisions are migrated**.

## Six carry something the code has not caught up with

Each says so in its status line. The rule is Lars's, set after batch 2: a decision the code has not implemented stays in the record and names the missing half, because deleting it hides the debt.

- **ADR-0042** — the jurisdiction seam declares five contribution interfaces and one shipped. `backend/pkg/extension/jurisdiction` carries `Pack` and `Retention`. No e-invoice formatter, no conformity regime, no trust artifacts, no country export profile exists as a type anywhere.
- **ADR-0049** — almost entirely unbuilt. `extensions/de/` holds four files and no `standards/` directory. None of the three artifacts the decision names is vendored. What exists is the retention floors as typed Go constants with the statute named in a comment.
- **ADR-0051** — "sovereign by default" holds **by omission**. There is no default endpoint at all: with `MARGINCE_BLOBSTORE_ENDPOINT` unset, `blobstore.FromEnv` reports unconfigured and the attachment routes answer 501, so nothing is stored anywhere. That is true and weaker than the decision describes, and the record now says so.
- **ADR-0074** — the total cost-unit mapping is not built.
- **ADR-0090** — the overlay composite setting was never built. `x_sor_mode` and `x_incumbent` are still columns on `workspace`, and `backend/tableownership_test.go` still carries the waiver this decision was meant to remove.
- **ADR-0101 / 0103** — flagged in their own status lines.

**ADR-0113 is fully built** — I checked specifically, and the seat ceiling is enforced inside the writer's transaction under an advisory lock, with integration tests at both module and HTTP levels. Its open question is the one worth carrying, below.

## Two issues filed from reading these against the code

[**#1791**](https://github.com/gradionhq/margince-poc-v1/issues/1791) — **the LICENSE says agents are not Seats; the seat count includes them.** `fullSeatsInUseQuery` counts every `seat_type='full'` row including the agent runner, while the licence text excludes "automated agents, AI systems, bots, and service accounts that act under the authority of […] a counted Seat". One number, two contradictory texts, and it runs against us commercially — we over-count against our own licence. This needs counsel, not code.

[**#1792**](https://github.com/gradionhq/margince-poc-v1/issues/1792) — **object storage falls back to `us-east-1` when the operator sets no region**, and that value reaches `MakeBucket`. Harmless for MinIO; for a real S3 endpoint it puts attachment bytes in a US region silently, in a product whose commitment is EU residency.

## One tooling fix

The index generator took only the **first line** of a status. A record naming which half of it is built usually wraps onto a second line, so the table truncated the qualifier mid-sentence — and then the staleness gate failed against a file the generator had just written, which is the worst kind of gate failure because it looks like the author forgot to run it. It reads the whole metadata block now. Mutation-checked.

## Checks

Every cited path verified. All six record gates plus the shared-rulebook parity test pass. Index regenerated: 43 records on this branch, 56 across the stack.

`STATUS.md` carries the session entry.

---
## Summary by cubic
Removes decision records from this public repository and updates references so ADR numbers are labels only. The rationale now lives outside this repo; code, tests, and `crm.yaml` remain the source of truth. No runtime or schema changes.

- Deletes `docs/adr/` and `scripts/gen-adr-index.sh`; removes `backend/adrindex_test.go`; adjusts `backend/rlsclaims_test.go` to stop scanning ADR prose; public-reference gate remains.
- Normalizes citations: merges ADR‑0114/‑0115/‑0116 into ADR‑0114 across comments in code, `backend/api/crm.yaml`, generated `backend/internal/contracts/api_gen.go`, and `frontend/src/api/schema.d.ts`.
- Updates `AGENTS.md`, `CLAUDE.md`, `README.md`, and `STATUS.md` to document the decision order and that rationale is not stored here.

**Review notes**
- Verify nothing points at `docs/adr/` or any private repo/link.
- Skim comment renumbers to ADR‑0114 and confirm generated artifacts are current.

<sup>Written for commit 8310c15e821fd4408dcc991a946a949a80644a5a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1793?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


LarsGradion:
**Two corrections and one finding worth more than the records.**

I removed the verification worktree while the agent writing the last seven records was still trimming, so ADR-0102 and ADR-0103 committed one pass early — two over-long lines and a History paragraph shaped as an aphorism rather than a statement. Both corrected. The gates that never got to run on this batch have now run: the public-reference sweep, the RLS prose sweep, the three decision-record gates and the rulebook parity test are all green, and every cited path resolves.

**[#1795](https://github.com/gradionhq/margince-poc-v1/issues/1795) — the finance mirror has no audit trail, and its own comment says it does.**

No file in `backend/internal/modules/finance/` calls `storekit.Audit` or `storekit.EmitEvent`. It writes four domain rows: `finance_invoice` inserted at `mirror.go:283` and updated at `:326`, plus `finance_external_customer` and `finance_connection` in `sync.go`.

`sync.go:16-19` states:

> The write shape is the house one: the mirrored row, its audit row and its outbox event in one transaction, `captured_by` from the authenticated principal.

That is false. The file calls `CapturedBy` and `LockRow`, which is why it reads as though it follows the shape, and never reaches the two calls that would make the sentence true.

**Why the gate missed it.** `backend/writeshape_test.go` asserts that every mutation *writing an audit row* commits a paired outbox event. A mutation writing **neither** is invisible to it — there is no audit row to pair, so nothing trips. The gate catches half the obligation, and a store that opts out entirely passes.

The issue names the decision that comes before the fix: whether a mirrored row is a record fact (write shape properly, event type added to the catalog) or operational plumbing (fix the comment, add an explicit ratified waiver). An invoice's paid state looks like a fact to me, but that is a product call. The second half — making the gate catch a module that owns tables and never audits them — closes the class rather than the instance, and is probably the more valuable change.

LarsGradion:
Superseded by #1865.

Issue #1812, opened 2026-08-19
zalo-personal's fleet-wide outage is unclassified AND dies, so it still manufactures dead rows


tinnhlt-gra:
`extensions/zalo-personal` has both halves of the problem the other two connectors no longer have.

**It publishes no class.** `pollFleet` returns a bare `fmt.Errorf("zalo-personal: all %d connection(s) failed this tick", failures)` (`extensions/zalo-personal/poll.go:133`), and the unit declares no `FailureClasses` on its `extension.Extension` at all. So `jobs.FaultForKind` finds nothing to honour and persists the unclassified substitute — `the job failed for a reason it could not classify; the diagnosis is in the process log` — which is exactly the sentence #1758 existed to stop printing. The unit already computes the classification it needs: `failureClass` in `cadence.go:221` returns `provider_unavailable` for a transport failure, and writes it to the member's own row. It dies at the job seam for the same reason zalo-oa's used to.

**And it dies rather than postponing.** With no declared class there is nothing for `extension.Reschedule` to travel under, so a fleet-wide outage spends the child's attempts and discards the row — one piece of dead work per dispatcher tick, for the length of the outage. That is the behaviour #1749 removed from the other two.

This is the sibling copy: #1749 fixed `zalo-oa` and `dispact-connector`, and this unit landed on `main` (#1656) after that branch point with neither half. It was not folded in because the disposition needs the classification first, and what a unit may publish into a fleet-visible unscoped column is its own review — the same argument that kept #1749 out of #1758.

**What it takes, and the one part that is NOT a copy.** The vocabulary is straightforward: five classes mirroring what `failureClass` already returns (`session_withdrawn`, `connection_unusable`, `provider_unavailable`, `poll_failed`), declared as inert data, registered on the Extension, and the fleet failure routed through `extension.Failure` / `extension.Reschedule`.

[...]

Issue #1809, opened 2026-08-19
Neither connector reads Retry-After, so a throttled poll answers on our clock


tinnhlt-gra:
`extensions/zalo-oa` and `extensions/dispact-connector` both classify a 429 as `errTransient` alongside a timeout and a 5xx, and neither reads the `Retry-After` header the provider may send with it. So a provider that names the interval it wants is answered on our own clock instead of on theirs.

Until #1749 this was masked by the disposition: a throttled tick failed, River's ladder retried within seconds, and after three attempts the row was discarded. Now the tick postpones itself to its declared cadence (120s) and keeps doing so for as long as the throttle lasts. That is strictly gentler than what it replaced — 120s is the healthy cadence, so a throttled tick puts no more load on the provider than a successful one — but it is still our number rather than the provider's, and a provider asking for ten minutes gets asked again five times inside that window.

`capture/telegram` already does this properly: `connector.RateLimitedError` carries `RetryAfter`, and `answerPollFailure` returns `river.JobSnooze(limited.RetryAfter)` with the comment that "Telegram named the interval it wants; a ladder of our own on top of it earns a harder limit." The pattern to follow is in-tree.

**What it would take.** Each client would need to distinguish a throttle from the other transient causes and carry the interval out with it — a sentinel or an error type alongside `errTransient`, not a new failure class, since `provider_unavailable` is still the right thing to tell an operator. `extension.Reschedule` already takes the delay as a parameter, so nothing in the seam has to change: the unit would pass the provider's interval instead of its own constant, and `jobs.FaultForKind` would clamp it to the same `[1s, 15m]` bound. Note the ceiling would then be doing real work rather than only catching mistakes, since a provider is free to ask for longer than 15 minutes — worth deciding whether a longer wait should be honoured or whether repeatedly asking at the ceiling is the safer answer for a connector whose poll liveness is a data-integrity concern.

Found while making a transient failure postpone rather than die (#1749); deliberately not fixed there, because reading a header changes what the client understands and that is a different review from what the tick does about a failure.

Related: #1749.

tinnhlt-gra:
Closing as **not worth fixing**, with the reasoning rather than just the label. Decision made alongside #1843, which ships #1817 and #1812.

The gap is real and this issue describes it correctly: a 429 carrying `Retry-After: 600` is answered on our own 120s clock, so a provider naming ten minutes gets asked five more times inside the window it named.

**Why it is not worth the change:**

[...]

Issue #1803, opened 2026-08-19
Job health cannot tell postponed work from idle work


tinnhlt-gra:
A worker that returns `river.JobSnooze` leaves its row in `scheduled`. `healthByKind` counts `scheduled` under `waiting` and its `OldestWaitingAgeSeconds` filter excludes rows whose `scheduled_at` is still in the future — correctly, since nothing is late. The result is that a connector which has been postponing itself for an hour because the provider is unreachable reads on the Maintenance screen exactly like a healthy connector idling between ticks: one waiting row, no age, no failure.

That became reachable behaviour with #1749, which stops a transient provider failure from dying and has it postpone instead. The trade was deliberate and it is the right one — the dead rows were a false alarm about work needing no intervention — but it means the job surface now has nothing to say about an outage that is genuinely running.

It is not invisible. `noteFailure` still writes `last_error_class` and both connector settings screens render it, so an admin who opens the connector sees `provider_unavailable` and its remedy. What is missing is the fleet-level reading: an operator scanning Maintenance for "is anything wrong" gets no signal, and there is no number to alert on.

**Options, and the recommendation.** The cheap one is a `postponed` count on `JobKindHealth` — `count(*) FILTER (WHERE state = 'scheduled' AND scheduled_at > now())` — plus the contract field and the column. That is honest and small, but on its own it renders as a number with no meaning attached: a healthy fan-out child also sits in `scheduled` between ticks, so the count is nonzero in normal operation and only interesting when it *stays* nonzero for the same workspace across many ticks.

Which is the same shape as #1750: the banner is wrong because it has no age window, and a postponed-work count is wrong for the same reason if it has none either. So the recommendation is to settle both together — decide what "needs a hand" means as a function of how long a condition has persisted — rather than adding a count now and an age window to it later.

Not urgent: the per-connector surface covers the diagnosis, and this is about fleet-level legibility.

Related: #1749 (the change that made this reachable), #1750 (the banner's missing age window), #1752 (a failure-class label on the job metrics).

tinnhlt-gra:
A second review pass sharpened this, and the sharper version is worse than what I filed.

[...]

Issue #1750, opened 2026-08-18
The "dead work needs a hand" banner has no age window, so a finished outage keeps it red for a week


tinnhlt-gra:
The Maintenance banner counts every `discarded` and `cancelled` row the caller can see, with no bound on age:

```
count(*) FILTER (WHERE state::text IN (discarded,cancelled))
```

(`backend/internal/platform/jobs/health.go`, `healthByKind`; summed in `frontend/src/screens/jobhealth.tsx` and rendered as `jobs.deadTitle` — "Dead work needs a hand".)

River's job cleaner is left at its defaults, which retain discarded and cancelled rows for **7 days**. So a transient outage that ended an hour ago goes on asserting that hundreds of jobs "will not happen without intervention" for the rest of the week, and an operator who has already read the failures has no way to acknowledge them. Observed with 531 rows from a DNS outage that had been over for an hour; the banner cannot distinguish that from an outage still in progress, which is the one distinction it exists to draw.

The count itself is not wrong — those rows do exist and that work did not happen. What is wrong is presenting a settled history as a live call to action.

**This needs a product decision before it is work,** which is why it is filed rather than fixed:

- **Age-bound the banner** (not the count): "N jobs died in the last hour" is actionable, "N died this week" is a report. Needs a window chosen, and a window is a judgement about how long an operator has to notice.
- **Or acknowledge**: let an admin dismiss a set of dead rows they have read. That is new state and a new mutation, so a much bigger change.
- **Or shorten River's retention** so the rows retire sooner. Cheapest, and the worst of the three: it destroys the failure history precisely when somebody investigating wants it.

Whichever way it goes, `dead` on `JobKindHealth` should probably keep its current meaning — the honest total — and the banner should read something narrower, so an operator can still see the full count without being paged by it.

Found while making a failed job say what happened; out of scope there because that change is about what one failure says, not about which failures shout.

dungbanguyen:
## Decision, 2026-09-01: age-bound the banner; keep the full count as a report figure

Re-triaged against `origin/main` at `d42887998`. Confirmed —
`backend/internal/platform/jobs/health.go:62` still describes Dead as *"discarded
plus cancelled: work that will not happen without intervention"*, counted with no
age bound.

### The ruling

**The banner counts a recent window. The full figure stays, as a number rather than
as an alarm.**

The ticket's diagnosis is the decision: the count is not wrong — those rows exist and
that work did not happen — what is wrong is **presenting a settled history as a live
[...]

PR #1815, opened 2026-08-19
feat(extensions): a transient provider failure postpones the tick instead of dying

### LINKED ISSUE #1749: A transient provider failure spends every attempt and dies instead of snoozing
A connector poll that cannot reach its provider — DNS failure, TLS handshake timeout, a 5xx — returns an ordinary error. River spends all three attempts on it and discards the row. A network outage of any length therefore manufactures dead job rows at the cadence of the poll, and each one raises the Maintenance banner that says work "will not happen without intervention" when in fact no intervention is possible or needed: the cursor never moved, so the next successful tick walks the same region and loses nothing.

`river.JobSnooze` is the right answer and the seam already supports it. `jobs.FaultContext` passes River control returns through untouched, with a comment naming exactly this case:

> A snooze reschedules and a cancel deliberately stops; neither is a failure and neither carries a cause to publish.

`extensions/capture/telegram` already uses `river.JobSnooze` on a provider throttle (`answerPollFailure`), so the pattern is established in-tree.

**What to do.** Where a unit classifies a failure as transient — `errTransient` in `extensions/zalo-oa` and `extensions/dispact-connector`, whose declared class is now `provider_unavailable` — snooze instead of returning the error. Two questions to settle rather than guess:

1. **Snooze how long, and does it interact with the cadence?** A poll on a 60s cadence that snoozes for 60s is indistinguishable from the next scheduled tick. The interesting case is a long outage, where a backoff would be kinder than a fixed delay.
2. **The connection row still needs its `last_error_class`.** The screen must go on saying the provider is unreachable — a snooze that wrote nothing would make a silent outage, which is worse than a noisy one. `noteFailure` writes that row today on the failure path, so the write has to survive the change of return value.

Note that a snooze does not close the gap entirely: an outage longer than the provider retention window still loses conversations, which `extensions/zalo-oa/poll.go` calls out as a data-integrity concern rather than a freshness one. Snoozing removes the false alarm, not the underlying exposure.

Found while making a failed job say what happened (the failure vocabulary change). Deliberately not fixed there: that change is about legibility and this one changes behaviour, so it deserves its own review.

tinnhlt-gra:
## What

A connector poll that could not reach its provider returned an ordinary error, so
River spent the child's three attempts and discarded the row. At a 120s cadence an
outage of any length therefore manufactured **one piece of dead work every two
minutes**, each raising a banner saying the work "will not happen without
intervention" — about work that needs none, because the cursor never moved and the
next reachable tick walks the same region.

This is what produced the 531 dead rows #1758 made legible. Fixing the
classification without fixing the disposition left the product telling an operator
two contradictory things about one row: the screen now says `provider_unavailable`
and "nothing to do: the poll catches up by itself", while the dead row itself says
somebody has to act.

## Why this shape

A composed unit **cannot** return `river.JobSnooze`: it is a separate Go module
that may import only the allowlisted `pkg/extension` surface. So it asks, with the
same declared class it would have failed under:

```go
// zalo-oa: an unreachable provider needs nobody, so the tick runs again instead of dying.
return extension.Reschedule(classProviderUnavailable, pollRetryDelay, cause)
```

`jobs.FaultForKind` honours the request **only for a class this installation
registered for the failing kind** — the same rule the sentence is held to, so
declaring a class buys both halves rather than each half having its own rule. An
undeclared postponement would let a unit silence its own dead work by asking
nicely, so an unregistered class gets the disposition it had before this existed.

Three things are worth stating because the design turns on them.

**A name is not a disposition.** Every failure in these units is classified; only
one of them needs nobody. A refused credential, a lapsed service package, an
unregistered API group, an answer the connector cannot read, and dispact's mixed
fleet class all still become dead work — because each of them needs somebody, and
postponing any of them would be a unit quietly deciding not to tell them. The
asymmetry is held over the **whole declared set** in both units, so a class added
later has to make the decision deliberately instead of inheriting whichever branch
it falls into.

**The delay is the dispatcher's own cadence, and a gate holds the equality.** A
postponed child sits in `scheduled`, one of the states the fan-out's uniqueness
window covers, so while it waits the dispatcher's next insert for that workspace
collapses into it — the postponement *replaces* the tick it would have raced.
Widen a cadence to 600s and leave a delay at 120s and that inverts silently: the
connector polls a refusing provider **harder** during an outage than it does in
health, with nothing about either file looking wrong. So the equality is derived
from the tree in `backend/pollcadenceparity_test.go` rather than asserted in a
comment beside a literal.

**The row write is untouched.** A postponement shows nobody anything on the job
surface, by design — it is not dead work. That makes the connection row the whole
trail, and a postponement that skipped `last_error_class` would have traded a noisy
outage for a silent one. Both connector settings screens already render it.

### Deliberately not a backoff

A decision about **loss**, not about politeness, and not a limitation: a ladder is
buildable — River keeps a snooze count in the job's own metadata, so the earlier
claim that "there is no streak to count" was simply wrong and is not the argument.
For these connectors poll liveness is a *data-integrity* concern rather than a
freshness one. Zalo drops messages from its API after roughly nine days with no
webhook and no depth to page back to, so polling less during an outage widens the
window a connector can **permanently** fall behind by, in exchange for saving one
request every two minutes.

The **throttle** arm gets its own argument, because "the provider is refusing
anyway" is not it there: `errTransient` covers a 429, and a 429 is a reachable
provider asking for less traffic. What makes the same delay right is that it is the
*healthy* cadence — a throttled tick postponing to 120s puts no more load on the
provider than a successful one does, and it is strictly gentler than what it
replaces, where River's ladder retried within seconds and then discarded the row.
Neither client reads `Retry-After`, which is #1809.

### The bounds, and why both do real work

`[1s, 15m]`, clamped at the seam.

- **The floor.** `river.JobSnooze` **panics** on a negative duration, so a unit that
  computed one from a clock — a deadline already past, a subtraction the wrong way
  — would take the worker process down rather than fail a tick. That is caught once,
  for everybody, rather than at each call site.
- **The ceiling** is about the **age reading**, not about visibility, and the
  imprecise version of this was checkably false in an earlier revision of the
  comment. A `scheduled` row **is** counted as waiting — by `health.go`'s
  `waitingStates` and by the queue-depth gauge. What a far-future `scheduled_at`
  escapes is every "how long has this waited" measurement, all of which filter
  `scheduled_at <= now()`, correctly, since nothing scheduled for the future is
  late. So an unbounded delay produces a count nobody can age, forever, beside
  counts a healthy idle tick produces too.
- A **clamped** request logs what it asked for next to what it got. Without that,
  72h and 15m logged identically and the one mistake the bound exists to catch left
  no trace anywhere.

## How verified

- [x] `make check` green on the merged tree (backend + frontend), EXIT=0
- [x] `make test-extensions` green (all six units, including `zalo-personal` from the merge)
- [x] `make test-integration` — `OK: integration passed with 0 skips (41 packages)`
- [x] `make craft-static` — `PASS — 0 blocker, 0 major, 0 minor` across `backend/`, `extensions/`, `fixtures/`
- [x] No secrets, no customer data, no local machine paths, no private specification paths

**Two claims are proved against real River + Postgres rather than reasoned about**
(`backend/internal/compose/extjobsnooze_integration_test.go`), because both are
assumptions about a library that a reader cannot check from our source and that
would each turn this change into a no-op:

| Claim | Why it had to be run |
|---|---|
| A postponement is honoured on the **last** attempt | `MaxAttempts: 1`, so River treating it as an ordinary failure has nowhere to go but `discarded`. During a real outage **every** attempt is the last one, so getting this wrong would work for a blip and fail for exactly the long outage the change is for. Confirmed in River 0.43's `reportResult`: the snooze branch is checked *before* any exhaustion path, sets `attempt-1`, records no attempt error. |
| A postponed row **holds its workspace's slot** | If `scheduled` were outside the fan-out's uniqueness window, an hour's outage would leave thirty live rows for one workspace instead of one — trading dead work for an unbounded queue. |

The `pollcadenceparity_test.go` gate was checked against a **negative control**: set
zalo-oa's delay to 60s against its declared 120s cadence and it goes red naming the
inversion.

Coverage of new code: `rescheduleFor` 100%, `faultFor` 100%, `Reschedule` /
`RescheduleAfter` 100%, both units' `dispositionFor` 100%, `fleetFailure` 100%. Unit
totals `zalo-oa` 87.3%, `dispact-connector` 86.1%.

**Manual flow** is in the PR comments (recording), driven through the real write
path: `dispact_connect` deposits its token without calling the provider, so a
connection to a host that never resolves produces a genuine `provider_unavailable`
through the real `noteFailure`, `dispositionFor`, `jobs.FaultForKind` and a real
`river_job` row. Nothing is simulated and no row is hand-inserted.

## Review findings this PR absorbed

Recorded because the findings are more interesting than the fixes, and two of them
were errors in my own reasoning rather than in the code.

- **The ceiling's stated rationale was checkably false.** The comment claimed a
  far-future delay would take the row "off every screen that shows live jobs —
  waiting, running, retrying, dead". `scheduled` is counted as waiting by both
  readers in this tree, and `STATUS.md` filed the opposite fact as open work in the
  same diff. Corrected to name the age reading, which is what is actually lost.
- **The clamp was silently lossy** in exactly the way the comment above it argued
  against. Now logs `requested=`, with the test asserting a clamped request is
  distinguishable and an unclamped one is not.
- **"River gives no streak to back off from" was wrong.** River keeps
  `metadata.snoozes`. The no-backoff decision stands on the data-integrity argument
  alone; the false half is removed from the code, the docs and STATUS.
- **The split was applied to one connector and not its twin.** `dispact-connector`
  crossed the file ceiling and got a `pollfailure.go`; `zalo-oa` held the identical
  concept at 494 of 500 lines. Mirrored, so the two read alike about the one thing
  they share.
- **The throttle arm was undiscussed** while being the arm the zalo fixture
  actually exercises (`fake.errorCode = codeRateLimited`). Named, argued, and the
  missing `Retry-After` filed.

## Follow-ups filed, deliberately not in this PR

- **#1812 — `zalo-personal` has BOTH halves of the problem, and this change did not
  reach it.** It landed from #1656 *after* this branch point: it returns a bare
  error from a fleet-wide failure and declares no `FailureClasses` at all, so an
  outage there is unclassified *and* dead work. Rule 1 says fix the invariant, not
  the call site — the reason it is not folded in is that the disposition needs the
  classification first, and what a unit may publish into a fleet-visible unscoped
  column is its own review, the same argument that kept #1749 out of #1758. Its
  delay is also **not** a copy of these two: its cadence is a per-member adaptive
  backoff sized against a measured retention floor, so "postpone by the cadence"
  has no single answer there.
- #1803 — job health cannot tell postponed work from idle work. A snoozing
  connector counts as `waiting`, which is the same conversation as #1750's missing
  age window, so it gets settled with that rather than answered sideways here.
- #1809 — neither client reads `Retry-After`, so a throttled poll is answered on our
  clock. `capture/telegram` already honours the interval Telegram names.

## AI involvement

AI-assisted throughout, under my accountability.

- **Hand-written:** the disposition seam (`Reschedule` / `RescheduleAfter`, the
  bounds and their rationale), the `dispositionFor` decision in both units, the
  cadence-parity fitness gate, and the two integration cases. These carry the
  argument, so I wrote them.
- **Independently reviewed** afterwards by an adversarial pass that produced the six
  findings above, all fixed before this was pushed. The two factual errors it caught
  were mine, and the commit and this description say so rather than quietly
  correcting them.
- **Verified by reading the library, not by trusting the summary:** River 0.43's
  snooze semantics (attempt decrement, the pre-exhaustion ordering, the panic on a
  negative duration, the `metadata.snoozes` counter) were read out of the module
  cache, and the two that the design rests on were then proved against a real run.

## Accountability

By opening this PR I confirm I am **accountable** for this change and can **explain
every line** in it — human-written or AI-assisted. Commits are signed off
(`git commit -s`, DCO).

Closes #1749.

---
## Summary by cubic
Postpones connector ticks on transient provider outages instead of failing them; the tick is rescheduled at the dispatcher cadence and no dead work is created while the connection row records the failure class. Ticks that hit their own timeout/cancellation now fail (not treated as outages).

- Details
  - `backend/internal/platform/jobs/fault.go` honors `extension.RescheduleAfter` only for a registered `FailureClass` and returns a `river.JobSnooze`; the delay is clamped to [1s, 15m] and logs requested vs applied; postponement is refused when the tick’s context is done.
  - New API: `extension.Reschedule`/`RescheduleAfter` let units request postponement with a class and cause; units do not import River.
  - Connectors: `extensions/zalo-oa` and `extensions/dispact-connector` postpone on `provider_unavailable` (and throttle 429) and still fail other classes; zalo-oa fixes its write path to record the failure on the current connection row.
  - Guardrail: `backend/pollcadenceparity_test.go` enforces postpone delay equals the dispatcher cadence and respects the seam’s ceiling to prevent over‑polling.
  - Tests/docs: unit + integration cover snooze on last attempt and slot occupancy; docs explain the disposition split.

- Rollout
  - No DB or config changes; behavior for non-transient failures is unchanged.
  - Extensions must call `extension.Reschedule` with a registered `FailureClass` to opt in.
  - Known gaps: `extensions/zalo-personal` not updated; neither connector reads `Retry-After`.

<sup>Written for commit 7ee7738113e68bd8d08384855f900399618c90a7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1815?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Transient provider outages are now postponed and retried automatically without consuming job attempts.
  - Postponements use connector-specific polling intervals and prevent duplicate dispatches.
  - Failure handling now distinguishes temporary provider issues from failures requiring intervention.

- **Bug Fixes**
  - Expired or cancelled polling windows continue to fail safely rather than being postponed.
  - Connection status and reauthentication details are preserved during polling failures.

- **Documentation**
  - Updated job retry guidance and fleet behavior documentation to reflect the new postponement handling.


tinnhlt-gra:
## UAT — driven against the real write path

Recorded on a live `make dev` stack. **No `river_job` row was hand-inserted, and no
failure was simulated:** a row the real writer never wrote proves nothing about the
writer, and the writer is what changed.

`dispact_connect` deposits its token and writes its connection row *without*
calling the provider, so connecting to `https://dispact.invalid` — RFC 2606, never
resolves, passes every `parseBaseURL` check — produces a genuine
`provider_unavailable` on the first poll, through the real `noteFailure` →
`dispositionFor` → `jobs.FaultForKind` → River path.

```
PUT /v1/ext/dispact-connector/connect  {"base_url":"https://dispact.invalid", ...}
→ 200  {"status":"connected","version":1}      # no provider call, as designed
```

### What the worker said

```
worker | WARN jobs: a worker postponed its own tick rather than failing
          kind=ext_dispact_connector_poll_inbox_ws
          class=provider_unavailable
          workspace_id=01a018ce-d31f-7e72-960c-ada0d9ee0988
          correlation_id=01a018db-b4c0-71fc-857d-3554c09cba0b
          err="provider_unavailable: dispact: all 1 connection(s) failed this tick"
          retry_in=2m0s
```

WARN and not ERROR, the class named, the cause present for the diagnosis, and the
delay stated — which is the whole trail a postponement leaves, since River records
no attempt error for one.

### The row, after ~12 minutes of continuous outage

```
 state     | attempt | attempt_errors | snoozes | wakes_in_s
-----------+---------+----------------+---------+------------
 scheduled |       0 |              0 |       6 |          9
(1 row)

 live_rows | dead
-----------+------
         1 |    0
```

Six postponements, and the log timestamps are **exactly 120s apart** — 14:16:49,
14:18:50, 14:20:50 — so the delay is the declared cadence in practice and not only
in the constant.

Every number here is a claim the change makes:

| Observed | Before this change |
|---|---|
| `state = scheduled` | `discarded` |
| `attempt = 0` after 6 failures | 3, then the row is thrown away |
| `attempt_errors = 0` | 3 recorded attempt errors per row |
| `live_rows = 1` | 1 live + **6 dead** and counting |
| `dead = 0` | 6, one per 120s, for the length of the outage |

### The connection row still names the outage

```
 status    | last_error_class     | version
-----------+----------------------+---------
 connected | provider_unavailable |       2
```

Which is where the design says an ongoing outage is named, and both connector
settings screens render it. Not parked — an unreachable provider is not a
credential problem.

### The Maintenance surface

```json
{"kind": "ext_dispact_connector_poll_inbox_ws", "queue": "default",
 "waiting": 1, "running": 0, "retrying": 0, "dead": 0,
 "oldest_waiting_age_seconds": null}

recent_failures naming dispact: []
total dead across all kinds: 0
```

**No dead work, and no entry in the failure list** — the false alarm is gone.

And this output is also the honest cost, visible here rather than argued about:
`waiting: 1` with a null age is *exactly* what a healthy idle tick looks like, so
the fleet surface cannot tell a running outage from a quiet one. That is #1803, and
this is the measurement it was filed on.

### Not covered by this run

**No screen recording.** The Chrome extension was not connected on this machine, so
I could not capture the UI. For this change the observable is a `river_job` row, a
connection row and a log line rather than a UI interaction — all three are quoted
above from the live stack — but the connector settings screen rendering
`provider_unavailable` is asserted only by its existing frontend suite here, not by
a video. Flagging it rather than implying coverage I do not have.

tinnhlt-gra:
@coderabbitai review

Issue #1787, opened 2026-08-19
A refused plan can lose the sentence that says what to fix, because the caller's token is bounded together with the remedy


tinnhlt-gra:
A refusal message that quotes the caller's token puts the token FIRST and the remedy after
it, and `boundFaultText` cuts the whole string at `httperr.MaxFaultText` (300). So a long
enough invented name eats the sentence that says what to do about it — the caller keeps
their own mistake and loses the answer.

`backend/internal/modules/search/queryvalidate.go:248` is the clearest instance:

```
the query plan cannot name "<caller token>" on "person"; read margince://schema/query for
the fields available to you
```

At roughly 270 characters of token the tail is gone and the message ends in `…`. The target
refusal at `:101` and the operator refusal have the same shape.

This is not new and it is not surface-specific: `Classify` bounds the value inside
`moduleDeclaredFault` (`backend/internal/platform/httperr/httperr.go:185`), so the REST
`details.errors` body has always truncated identically. #1782 only made it visible by
carrying the message to MCP as well.

Two candidate fixes, and the choice is the interesting part:

- **Order the sentence remedy-first**, so what survives the cut is the actionable half. Cheap,
  and it changes only wording — but the token is what tells the caller WHICH of several
  predicates was refused, so leading with the remedy costs that when the plan has more
  than one fault.
- **Bound the token separately in the producer**, before it is composed into the sentence,
  so the remedy is never the part that is cut. More code, keeps both halves, and puts the
  bound where the caller-authored text actually enters — which is where the rest of this
  tree puts it (`agents.echoSafe`, `UnknownToolError`'s name bound).

The second is the recommendation: the bound belongs on the caller's contribution, not on
the sentence the server wrote around it.

`BadArgsError.Guidance` already carries the precedent in a comment
(`backend/internal/modules/agents/badargs.go:105`): bounding server-authored guidance
together with a caller echo "made the accepted-field list truncate mid-word on a long
unknown key — cutting away the list the message exists to teach, exactly when the caller
most needed it". That is this defect, in a different file, already diagnosed once.

Found while reviewing #1782 (adversarial review of the MCP renderer fix); not fixed there
because the renderer cannot fix it — the cut has already happened by the time it sees the
message.

tinnhlt-gra:
A second instance of the same defect, one layer further out, found by the same review.

[...]

PR #1831, opened 2026-08-19
fix(records): a person names their employer, and an activity names what it is about

### LINKED ISSUE #1776: The query grammar cannot see a join table, so a person has no employer and an activity is about nothing
Probed live over MCP: every `traverse` from `person` is refused.

```
query_workspace { target: person, traverse: { relation: "employment" } }
→ (validation_error traverse.relation=unknown_relation …)
```

`organization`, `organizations`, `deals` and `activities` answer the same way.
`person` has **zero** traversable relations, and so does `activity`.

## Root cause

The vocabulary is derived, not listed — which is right — but it derives a
relation only from a **scalar `<record type>_id` member**:

```go
// modules/search/queryfields.go:192
// relationSuffix is the contract's own spelling of a reference: a scalar id
// member named `<record type>_id`.
const relationSuffix = "_id"
```

`deriveRelations` skips any field that is not `KindID` with that suffix, and
`deriveReverse` builds the reverse edge from the same scalar. So **every
relationship carried by a join table is invisible by construction**:

- person↔organization employment lives in `relationship`, so `person` can name
  no employer and `organization` can name no staff.
- activity↔person/organization/deal linking lives in `activity_link`, so
  `activity` can name nothing it is about.

This is one gap, not two, and it is not a missing entry in a vocabulary table —
the derivation has no concept of a join table to add an entry for.

## Why it matters more than it looks

It is the reason two separately-reported problems cannot be answered:

- A person's employer cannot be read or counted over MCP, so a duplicate
  employment (#1775) is invisible on the surface that created it.
- `activity` → `person` was never noticed as missing at all, which is what a
  derived vocabulary with a blind spot does: it produces a confident, complete-
  looking answer with a hole in it.

## Done looks like

- `person` traverses to its employer and `organization` to its people.
- `activity` traverses to the records it links.
- The derivation learns about join tables — it does not grow a hand-maintained
  list of exceptions beside a derived vocabulary, because the next join table
  would silently miss the list. `backend/arch_test.go`'s house style (derive the
  obligation from the tree) is the bar.
- A refusal for a genuinely unknown relation still names what IS available, so
  the vocabulary stays discoverable without guessing.

## Related

- #1775 — the duplicate-employment write defect, found in the same session.

tinnhlt-gra:
Closes #1776.

## What was wrong

The query grammar derived a relation from **one** spelling: a scalar
`<record type>_id` member on a contract type. Both derivations put the
reference *on* one of the two records (`contractRelations` reads the record's
own, `inverseRelations` reads another record's), so an edge carried by a table
between them was invisible **by construction** rather than by omission.

`person` and `activity` had zero traversable relations. A derived vocabulary
with a blind spot produces a confident, complete-looking answer with a hole in
it — which is why `activity` → `person` was never noticed as missing at all.

## What makes one rule enough

Grounding the two join tables against the DDL turned up the fact the issue does
not state, and it decides the design: **both carry typed `<record>_id`
columns.**

| Table | Columns onto searchable records |
|---|---|
| `activity_link` (core `0008`, widened by `0038` and `0131`) | `activity_id`, `person_id`, `organization_id`, `deal_id`, `lead_id`, `project_id` |
| `relationship` (core `0007`, widened by `0131`) | `person_id`, `organization_id`, `deal_id`, `project_id` |

`activity_link` also has `entity_type`, but it is a CHECK-enforced discriminator
over columns that are all typed — the contract's `ActivityLink{entity_id,
entity_type}` is the *wire* shape of a link, not its storage. So the rule reads
columns off the live catalog, exactly as the field vocabulary already does, and
never has to know what a link means.

## Hub and arm, not every pair

A join table's references are **not interchangeable**. No `relationship` row
fills both `organization_id` and `deal_id` — they belong to different kinds
(`rel_employment_shape`, `rel_stakeholder_shape`, `rel_project_stakeholder_shape`
each name exactly one pair). Pairing every column with every other would publish
`organization → deals` through that table: a hop no row can satisfy, which a
caller reads as *"no data"* rather than *"not a thing"*.

So one column is declared per table and the rest are arms, and the edges are
hub↔arm. Every legal `relationship` row pairs a person with one arm; every
`activity_link` row pairs an activity with one arm.

## What is declared, and what is derived

**Declared:** two table names and one hub column each.
**Derived:** every relation, both directions, from the live columns. An arm
added to a join table becomes traversable with no edit here — which is exactly
what `0038` and `0131` each did.

`TestEveryJoinTableInTheSchemaIsDeclared` reads the core migrations and requires
a **verdict** for every table carrying two or more record references: either
declared in `joinTables`, or recorded in `notAnEdge` **with the reason it is not
one**. A new table gets neither and fails. A bare exclusion list would read as
"handled"; this reads as "decided".

The census found six other candidates, each now carrying its reason —
`activity_participant` (what capture *matched*, where `activity_link` is what it
*asserted*), `person_consent` / `consent_event` (polymorphic parent, never both),
`activity_retention_evidence`, `person_signature_enrich_state`, `contract`.

## Two things stated rather than papered over

- **`relationship.counterparty_org_id`** strips to `counterparty_org`, which is
  not a record type, so the same test that keeps `deal.owner_id` out of the
  vocabulary keeps it out here. **Org↔org partner edges stay untraversable.**
  Gated by a test on a schema where the role-named column is the *only*
  organization reference, so the derivation cannot pass by finding the properly
  named column beside it. The honest fix is a contract rename.
- **Relations stay named by target type, not by relationship kind.** The issue's
  live probe tried `relation: "employment"`; `relationship` carries three kinds
  already distinguished by which columns they fill, so a kind-named relation
  would be a second vocabulary to maintain.

## Also

A refused hop now **names the hops the caller does have** — their own
RBAC-filtered vocabulary, so the list says the same thing whether the relation
never existed or lands somewhere they cannot read, and no discovery channel
opens. Bounded at 12 with the overflow named, because a hop count is a product
of the schema and not a number the refusal controls (the shape #1787 describes).

## What the reviews changed

Two independent reviews (Codex and a `/code-review high` pass) found six things
between them; **only one was found by both**, which is the argument for running
two. Four were real defects, now fixed in `7539e51`:

- **A person who LEFT still answered as staff.** The hop filtered `archived_at`
  and not `ended_at`. Archived is *deleted*; ended is *left*, and it leaves the
  row in place — core `0007` says so on the column, and every other reader of it
  in the tree filters it. A future `ended_at` is a notice period and is still
  current. The integration fixture had seeded only the archived case, which is
  the state that does not occur.
- **The surface published `activitys`.** No relation had exercised the plural
  rule on `activity` before, because no record declares a scalar `activity_id`.
  The rule also had a **second copy** in the validator's narrowing pass, so
  fixing one alone would have published `activities` and then refused it.
- **`activity_link` has a project arm** (core `0131`) and the fixture denied it,
  so the headline test asserted the absence of a hop production correctly
  publishes. (Both reviewers found this one.)
- **The census missed this repository's own DDL** — multi-line `ADD COLUMN`,
  statements terminated by a semicolon *inside a `--` comment*, `migrations/custom`,
  and case-sensitive `uuid`. `attachment` now surfaces and carries a verdict.

One finding is filed rather than fixed: **#1833**, org↔org partner edges stay
untraversable. And the reviewers disagreed about the tenant boundary — one held
the FORCE RLS comment accurate from core `0014`, the other found `0217` retiring
every policy. The migration settles it; the comment now names row scope and
object RBAC, which is what actually bounds the read.

## Verification

- `make check-backend` green, `make check-fe` green, `make craft-static` green
  (0 blocker / 0 major / 0 minor).
- Full `make test-integration` lane green, exit 0.
- **`TestEveryPublishedRelationExecutes`** — the pre-existing fitness test that
  traverses every published hop against real Postgres — now covers all the new
  edges and passes.
- New-code coverage **92–100%** per function.
- **Live UAT over MCP** against a running stack: the `margince://schema/query`
  resource publishes the new hops, `query_workspace` traversing
  `person → organizations` returns the employed person with the organization as
  evidence, and the exact call from the bug report (`relation: "employment"`)
  now refuses *by name and lists the hops that exist*.
- **Every new gate was mutation-verified** (eleven mutations in all). Two passed under mutation on the
  first attempt and were rewritten: one was an absence assertion that could
  never fire, the other asserted a vocabulary-level collision that no longer
  exists under the hub rule. The row-scope integration test carries a positive
  control on the same principal and edge, because an empty answer most often
  means the fixture never seeded.

---
## Summary by cubic
Enables join-table traversal so a person names their employers and an activity names what it is about. Previously only scalar <record>_id edges existed; now hub↔arm edges are derived from join tables, ended employments are excluded, and RBAC on the `relationship` edge is enforced.

- Derives edges from live columns on `relationship` (hub: `person_id`) and `activity_link` (hub: `activity_id`); skips archived rows and ended employments (`ended_at` ≤ current_date), while admitting future `ended_at`.
- Hides `relationship`-backed hops from the vocabulary when the caller lacks read on the `relationship` object; `activity_link` hops are unaffected. Hops are omitted rather than refused at execution.
- Executes join hops as an IN subquery on the join table; row scope and object RBAC at endpoints bound the read; no tenant predicate or RLS is used. `Via` stays prose; the schema document does not publish join execution details.
- Adds partial indexes on `relationship` reference columns (`*_id` WHERE `archived_at IS NULL`) to avoid per-row sequential scans; keeps kind-specific partials for module reads.
- Unifies pluralization with one `pluralRelationName(...)`; publishes “activities.” When a scalar and a join edge collide on a name, the scalar wins.
- Validator refuses unknown relations by name and lists this caller’s available hops (bounded, with overflow named).
- Schema reads memoize by table and add a `holds(column)` check so join edges publish only when both columns exist; unwired deployments do not publish join edges.
- Migration census scans `core` and `custom`, handles multi-line ALTERs, comment-embedded semicolons, and case variants; records verdicts for non-edges (e.g., `activity_participant`, `person_consent`, `consent_event`, `activity_retention_evidence`, `person_signature_enrich_state`, `attachment`, `contract`). Org↔org partner edges remain untraversable.
- Tests cover derivation, SQL guards (`archived_at`, `ended_at`), integration (row scope and RBAC gating), and ensure derivation tests run with a principal holding the `relationship` grant; a gate also cross-checks identity’s object list against its function signature to prevent drift.

**Required actions**
- When a new table carries two or more `<record>_id` columns, add it to `joinTables` (with hub) or to `notAnEdge` with a reason. The test suite enforces a verdict for every candidate table.

<sup>Written for commit e658cbb83f6a2b36b4e4aed779fbcc89647596c1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1831?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Search traversal now supports relationships through employer, activity, and other join-table connections.
  * Results respect archived, ended, and row-level access restrictions.
  * Search supports irregular plural relationship names and improved relation discovery.

* **Bug Fixes**
  * Corrected valid relationship resolution and inverse traversal behavior.
  * Improved unknown-relationship messages with concise, relevant available options.


tinnhlt-gra:
## UAT — driven live over MCP against a running stack

The defect was reported by driving the live server from a chat client, so it is proved back the same way: a passport against `POST /mcp`, not a unit test and not the web app. Recorded in the MCP Inspector; the verbatim JSON below came from `curl` alongside it, because the Inspector renders results as scrolled text and the `via` strings clip horizontally in its Resources pane.

uat-join-hop

**Fixture**, seeded through the real REST write path: organization *Stuttgart Werke GmbH* (`address.city = "Stuttgart"`), person *Ronny Schmidt*, and an `employment` row linking them. The employment lives in the `relationship` join table — neither record carries a column pointing at the other, which is the whole point.

### 1 · The vocabulary now publishes the hops

`resources/read` on `margince://schema/query`:

| target | relations |
|---|---|
| `person` | `activities` `activity_link(person_id → activity_id)` · `deals` `relationship(person_id → deal_id)` · `organizations` `relationship(person_id → organization_id)` · `projects` `relationship(person_id → project_id)` |
| `activity` | `deals` · `leads` · `organizations` · `persons` · `projects` — all `activity_link(activity_id → …)` |
| `organization` | `persons` `relationship(organization_id → person_id)`, and `deals`/`projects` resolving to the **direct** scalar edges `deal.organization_id` / `project.organization_id` |

Two things worth reading off that table. `activities`, not `activitys` — the irregular plural, which no relation had ever exercised before a join edge gave `activity` its first inverse hop. And `organization → deals` staying the record's own reference rather than the weaker path through `relationship`: that is `mergeRelations` running live, not just in a unit test.

### 2 · The traversal that was impossible

```json
{"version":"v1","target":"person",
 "traverse":{"relation":"organizations",
             "where":[{"field":"address.city","op":"eq","value":"Stuttgart"}]}}
```

```json
"full_name": "Ronny Schmidt",
"evidence": [{"relation":"organizations","record_type":"organization",
              "title":"Stuttgart Werke GmbH"}],
"coverage": "complete_exact",
"executed_plan": "person records, linked to an organization record where address.city is \"Stuttgart\"; at most 50, newest first."
```

### 3 · The inverse

target `organization`, relation `persons`, `full_name eq "Ronny Schmidt"` → `"display_name": "Stuttgart Werke GmbH"`, evidence `{"relation":"persons", … "title":"Ronny Schmidt"}`, `complete_exact`.

### 4 · The refusal from the original report

The bug report tried `relation: "employment"` — a relationship *kind*, not a relation name. It is still refused, because a kind is not a relation. What changed is that the refusal now answers instead of redirecting:

> `"person" has no relationship named "employment"; it has "activities", "deals", "organizations", "projects" (margince://schema/query carries what each one lands on)`

That list is the caller's own RBAC-filtered vocabulary, so it reads the same whether the relation never existed or lands on a record type they cannot see — no discovery channel, and no second round trip to a document.

---

Also verified in the lane rather than by eye: `TestEveryPublishedRelationExecutes` — the pre-existing fitness test that traverses **every** published hop against real Postgres — covers all the new edges and passes. And the row-scope case carries a positive control on the same principal and the same edge, because an empty answer most often means the fixture never seeded.

tinnhlt-gra:
@coderabbitai review

tinnhlt-gra:
@coderabbitai review

tinnhlt-gra:
@coderabbitai review

tinnhlt-gra:
@coderabbitai review

tinnhlt-gra:
@coderabbitai review

tinnhlt-gra:
@coderabbitai review

tinnhlt-gra:
@coderabbitai review

tinnhlt-gra:
@coderabbitai review

tinnhlt-gra:
Both nitpicks addressed in `316c9ebf2`, and the second one was not a nitpick — thank you.

**`querysql_test.go`'s `relationship` fixture omitted `ended_at`.** That is the real finding. `joinRelations` reads `JoinEdge.Ends` off the stored columns, so every plan compiled from that fixture was built with the ended-employment guard **switched off**, and the SQL-level test therefore could not assert it. I verified the consequence by mutation: with the old fixture, deleting the `ended_at` clause from `joinEdgeCondition` left this file green — only the unit derivation and the integration lane would have caught it. Fixture aligned to the DDL, and the compile test now asserts both guards; the same mutation now fails here.

Worth naming because three earlier reviewers walked past it: the fixture *called itself* the DDL's columns, which is exactly the kind of comment that stops a reader checking.

**The role-named test asserted an absence only.** Correct, and it is the failure mode I had already been bitten by twice on this branch. It now derives a control hop through `activity_link` and fails if that one is missing, so the absence it reports is about `counterparty_org_id` rather than about the derivation having stopped working. Mutation-verified the same way — breaking the hub check now fails the control rather than passing silently.

I took your `activity_link` control suggestion as written but scoped the loop's error to `objectRelationship` rather than erroring on every relation, since the control hop is itself a relation and would otherwise fail its own assertion.

tinnhlt-gra:
uat-join-hop

tinnhlt-gra:
@coderabbitai review

Since your last review (`f0b707edd`) this branch gained two commits you have not seen, and one of them is the riskiest in the PR:

- `7edae336f` — an **authz gate**: a traversal through the `relationship` join table was reading the employment/stakeholder edge while gating only the two endpoint records, never the edge's own RBAC object. Plus a migration adding four partial indexes.
- `316c9ebf2` — your own two findings, applied.

Please review those specifically. The gate sits in vocabulary resolution rather than at execution (deliberately — dropping the hop keeps existence-hiding), so the question I most want a second opinion on is whether every path that can execute a hop really goes through that resolution.

Issue #721, opened 2026-08-09
Let deals be filtered by the customer's industry


LarsGradion:
## What is missing

You cannot filter or group deals by the industry of the company they belong to. The deal filter vocabulary contains only deal-owned fields (pipeline, stage, owner, organization, partner, project, status, forecast category) — no attribute of the related organization. There is also no industry filter control anywhere in the UI, and the organization list endpoint itself does not accept an industry parameter (only the segment engine does).

## Why it matters

"Show me the pipeline for manufacturing" is a normal question and a repeatedly requested view. Industry data is already there and already populated — we simply cannot slice by it.

## What already exists — do not rebuild this

- Industry, size band and lifecycle are real columns on the organization: `backend/migrations/core/0005_organizations.up.sql`.
- Industry is populated automatically by enrichment, with evidence, and survives a merge — see `backend/internal/modules/people/coldstartprofile.go` and `organizationfact.go`.
- Organizations can already be filtered by industry through the segments engine: `backend/internal/modules/collections/vocab.go`.
- The deals screen already has a working filter bar to extend: `frontend/src/screens/deals.tsx`.

## What to build

1. Allow deal filters to reach organization attributes (industry at minimum, probably size band and lifecycle too) — a join in the segment vocabulary for deals.
2. Accept industry and size band on the organization list endpoint, so the plain list agrees with the segment engine.
3. Filter controls in the UI for both screens.
4. Watch the retired-vocabulary trap: there are two overlapping fields (the current `lifecycle` and a retired `classification` kept for one release). Do not add new filters on the retired one.

## Done when

- [ ] A user can filter the pipeline board and list by industry.
- [ ] The same filter works on organizations, in saved views and in export.

**Size:** S.
**Spec:** `lists-views-segmentation.md` and the organization classification story S-E02.7.

dungbanguyen:
Progress, plus a correction to this issue's premise.

**Landed:** https://github.com/gradionhq/margince-poc-v1/pull/1837 — a deal filter reaches the customer's attributes (`organization_industry`, `organization_size_band`, `organization_lifecycle`), through the correlated-EXISTS the tag leaf already used. Because the vocabulary drives the surfaces, dynamic lists, saved views, the Filters & views screen and CSV/JSON export all gained it at once. `classification` deliberately not offered, per this issue's fourth point.

[...]

PR #1842, opened 2026-08-19
feat(frontend): the companies list gains a size dial

dungbanguyen:
Part of [#721](https://github.com/gradionhq/margince-poc-v1/issues/721)'s UI point, and small: `GET /organizations` has accepted `size_band` for a while — with enum validation in the handler — but no screen offered it, so the capability was only reachable by hand-writing a URL.

## One thing worth explaining

The dial sits in `dataChips`, not `chips`, which the prop name argues against. A `chips` entry runs **every** label through `t()`, so its options must be message keys — and a size band is a numeral range that reads identically in English, German and Vietnamese. Seven keys whose three translations all coincide would be noise standing in for meaning. The dial's own name and its "any size" entry *do* need translating, and those are the two strings this passes through `t()` itself.

If that reads as a misuse of the prop, the alternative is a `chips` entry plus 21 lines of `"1-10": "1-10"`. I took the comment over the boilerplate; happy to flip it if you'd rather the prop name stayed literal.

## The test

It asserts the **wire spelling**, not that a request went out. A chip whose key does not match the server's parameter name is ignored, and an ignored filter answers the *whole* list with `200 OK` — a filter that reads as working and is not. `listquery.tsx` already carries a comment about that failure mode, having shipped it once (`owner=unassigned:true` instead of `unassigned=true`).

Mutation-verified: renaming the key to `sizeBand` fails the test.

## What this does not do

`industry` is the other half of #721's UI point and is **not** here, because it needs a decision rather than code:

- The column is free text, so a chip has no bounded option list to offer.
- `ListChip.search` is the designed mechanism for an unbounded set — but it needs a source of distinct industry values, and no endpoint returns one. `q` searches names, not industries.
- `translateChip` also drops `search`, so today only a `dataChips` entry could carry it.

So the options are: add a distinct-values endpoint (a contract change #721 does not ask for), make `industry` an enum (a product decision), or add a text dial to `ListSurface` (a shared surface every list uses). I have not picked one.

Worth noting the gap is narrower than #721 states: a user **can** filter companies by industry today — on Filters & views, with saved views and export, via the segment engine. What is missing is an inline dial on the companies list.

`make check-fe` green (3365 tests).

---
## Summary by cubic
Adds a Company size dial to the Companies list, wired to `size_band` on `GET /organizations`, so users can filter by size inline instead of editing the URL. Supports the UI requirement in #721.

- Sends `size_band` with literal band values (e.g., `51-200`) to filter server-side.
- Places the dial in `dataChips` (not `chips`) to avoid auto-translation of numeric range labels; only the dial name (`org.sizeBand`) and “Any size” (`org.filterSizeBandAll`) are translated in `en`, `de`, and `vi`.
- Test now reads the parsed `size_band` query param and expects `51-200` rather than matching a URL substring, preventing false positives and guarding the parameter name.

<sup>Written for commit a4bace0fd15a9ba576c88ce5f61b82f6962eb566. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1842?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added a company-size filter to the organizations list.
  * Selected size ranges are displayed alongside existing filters.
  * Added English, German, and Vietnamese labels for company-size filtering and fields.

* **Tests**
  * Added coverage confirming that selecting the 51–200 company-size range applies the correct filter.

### DIFF
--- a/frontend/src/i18n/de.ts
+++ b/frontend/src/i18n/de.ts
@@ -556,6 +556,7 @@ export const de = {
   "list.viewProspects": "Interessenten",
   "org.filterLifecycleAll": "Alle Phasen",
   "org.filterRelTypeAll": "Alle Typen",
+  "org.filterSizeBandAll": "Alle Größen",
   "person.consent": "Einwilligung",
   "consent.grant": "Erteilen",
   "consent.withdraw": "Widerrufen",
@@ -593,6 +594,7 @@ export const de = {
   // die die abgelöste Einstufung mit einem Wert beantworten wollte.
   "org.lifecycle": "Account-Status",
   "org.relationshipTypes": "Beziehung zu uns",
+  "org.sizeBand": "Unternehmensgröße",
   "org.lifecycle.unknown": "Nicht eingeschätzt",
   "org.lifecycle.target": "Zielkunde",
   "org.lifecycle.prospect": "Interessent",
--- a/frontend/src/i18n/en.ts
+++ b/frontend/src/i18n/en.ts
@@ -572,6 +572,7 @@ export const en = {
   "list.viewProspects": "Prospects",
   "org.filterLifecycleAll": "Any stage",
   "org.filterRelTypeAll": "Any type",
+  "org.filterSizeBandAll": "Any size",
   "person.consent": "Consent",
   "consent.grant": "Grant",
   "consent.withdraw": "Withdraw",
@@ -608,6 +609,7 @@ export const en = {
   // questions the retired classification answered with one value.
   "org.lifecycle": "Account lifecycle",
   "org.relationshipTypes": "Relationship to us",
+  "org.sizeBand": "Company size",
   "org.lifecycle.unknown": "Not assessed",
   "org.lifecycle.target": "Target",
   "org.lifecycle.prospect": "Prospect",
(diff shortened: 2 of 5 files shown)


dungbanguyen:
Applied — the assertion now reads the parsed parameter (a4bace0f).

The point is right and I had it backwards in my own commit message: `includes("size_band=51-200")` asserts less than it looks like it does. The current band set happens to contain no value that starts that way, so the weakness is invisible rather than absent — it would surface the day somebody adds one. Reading `searchParams.get("size_band")` says what the test means and stops depending on the enum staying non-prefixing.

I have not swept the sibling assertions in this file (the lifecycle dial's test has the same shape). Those are outside this change and a sweep is its own diff — happy to file it if you want the pattern gone repo-wide.

Issue #1848, opened 2026-08-19
The domain filter leaf needs a normalizing operand, not a third map entry


dungbanguyen:
Split out of [#1657](https://github.com/gradionhq/margince-poc-v1/issues/1657), which listed `domain` alongside `owner_team_id` and `relationship_type` as "additive: three entries in the resource field maps". The other two were. This one is not, and shipping it as a third map entry would be worse than leaving the gap.

## Why it is not a simple entry

The organization list already filters by domain, and it **folds the caller's value first**:

```go
// organization_list.go
func organizationDomainClause(domain *string, includeArchived bool, arg func(any) int) string {
	…
	arg(foldDomainQuery(*domain))   // values.ParseDomain → the host the column holds
}
```

The comment there is explicit about why: a caller who pasted `https://www.acme.example/careers` out of an email signature is asking about the host the column actually stores.

A predicate leaf's operand does not get that treatment. `storekit`'s `scalarOperand` validates by field type and binds; it does not normalize. So adding `domain` as a `FieldText` leaf would give:

| Surface | `domain = "https://www.acme.example/careers"` |
|---|---|
| `GET /organizations?domain=…` | folds to `acme.example`, finds the account |
| a segment / saved view / export | compares the raw string, **matches nothing** |

Same field name, same stored column, two different answers — and the wrong one is silent. That is precisely the divergence the one-engine rule exists to prevent, and it would be *introduced* by closing this gap carelessly.

## The options

1. **A normalizing field type** (`storekit.FieldDomain`) that folds its operand through `values.ParseDomain`. `values` is Tier-0 (`shared/kernel/values`), so `platform/database/storekit` may import it — the DAG allows this. **But** `FilterVocabularyField.type` is a closed enum in `backend/api/crm.yaml`, so a seventh type is a **contract change** visible to every client, plus the three regenerations, and `TestTheVocabularysTypeEnumIsExactlyWhatIsFilterable` fails until the enum carries it. A new public type for one field is a real cost.
2. **Fold at the compile boundary** — a per-field operand transform on `storekit.Field`. No contract change and no new type, but it adds a general mechanism (arbitrary operand rewriting) to an engine whose current guarantee is that every value travels as a bind parameter unaltered. That guarantee is worth something.
3. **Leave it out and say so.** The gap stays, but nothing lies: a filter naming `domain` is refused by name (`filter_field_not_allowed`) rather than answering an empty page. Costs the promise in the DM-VOCAB table.

[...]

Issue #1657, opened 2026-08-18
The segment engine lacks three filter leaves the spec's DM-VOCAB table promises (owner_team_id, relationship_type, domain)


dungbanguyen:
Three fields the spec's per-resource filter vocabulary promises as predicate leaves are absent from the engine that implements it, so a filter naming any of them is refused today.

## The divergence

`backend/internal/modules/collections/vocab.go` holds the closed core vocabulary — the one engine that dynamic-list validation, membership evaluation and filtered export all resolve through. Measured against the spec's DM-VOCAB table:

| Field | Resource | In DM-VOCAB | In `vocab.go` |
|---|---|---|---|
| `owner_team_id` | people, organizations, leads, deals, projects | yes (the ownership dial) | **no** |
| `relationship_type` | organizations | yes | **no** |
| `domain` | organizations | yes | **no** |

Each names one column with one access path, so each is a leaf the table promises rather than a list parameter — the spec's own leaf-versus-parameter rule turns on exactly that, and none of the three carries the `†` marker that would make it a parameter.

`owner_team_id` is the widest of the three: the spec binds it as one third of a single ownership dial carried by **every** owner-scoped list resource with identical semantics, and the engine has `owner_id` but not the team half. So a segment can say "owned by this person" and cannot say "owned by this team", which is the form a manager's saved view actually takes.

## What this is not

Not a spec error, and worth stating because the resolution direction is easy to get backwards. Contract-first (P3) means the table is the promise and the engine owes it. The alternative — marking these fields as list-parameters-only in the spec so the table matches the code — would relabel a build gap as a semantic limit and close it by fiat. That reading is now the stated rule rather than a judgement call, so the next reader hits the guidance rather than the trap.

## Scope

Additive: three entries in the resource field maps, with `owner_team_id` resolving through the same team edge the list operations already use. The vocabulary is closed and gated, so each addition needs its parity test alongside — the same shape as the `tag` leaf's.

Whether `owner_team_id` should instead be derived once in shared machinery for all five resources, rather than added per resource, is the one design question here; the spec binds it as ONE vocabulary bound once and never re-implemented per resource, which argues for the shared form.

Found while reconciling the DM-VOCAB table against the engine; the decision half of that reconciliation is being recorded with the team separately.

dungbanguyen:
Status, because the Scope section above now describes a state that no longer exists and a stale issue misrepresents the tree as confidently as a wrong one.

[...]

PR #1850, opened 2026-08-19
feat(collections): the segment vocabulary gains the team and relationship leaves

dungbanguyen:
Two of the three leaves [#1657](https://github.com/gradionhq/margince-poc-v1/issues/1657) found missing against the spec's DM-VOCAB table. Contract-first: the table is the promise and the engine owes it, so these are additions rather than a spec correction. No contract change, no new engine capability — `storekit.Field.Link` already does correlated-EXISTS.

## `owner_team_id`, once, for every owner-scoped engine

That was the issue's one open design question, and the spec answers it: the ownership dial is **one** vocabulary bound once, not five per-resource implementations. So this is a single `storekit.Field` shared by all five engines rather than a template string copied into each.

`TestEveryEngineWithAnOwnerAlsoFiltersByTeam` derives the obligation from the engines themselves — every engine carrying `owner_id` must carry the team half, and vice versa, since a team leaf without an owner leaf is equally half a dial. It fails if it ever sweeps nothing, and asserts each engine's leaf is **definitionally identical** to the shared one. Mutation-verified in both directions.

It selects the same rows the `owner_team_id` list parameter answers, said as the EXISTS a link leaf can express. It is a **filter, not a scope**: the executor ANDs it with the caller's row-scope clause, so naming a team the caller cannot see answers their own visible rows filtered to nothing rather than reaching that team's.

## `relationship_type` on organizations

Multi-valued, so the fact lives in its own table and the leaf selects accounts that are **at least** the named type. Archived rows are excluded inside the wrapper — a withdrawn relationship is one the account no longer has, and a segment naming it would otherwise keep selecting on a fact somebody deliberately took back.

The engine does **not** check the value against the contract's enum, where the list parameter 422s a typo. That is the call every picklist leaf here already makes (custom picklists included) and the outcome is honest: a value no row carries selects nothing. It is **gated by a test** rather than explained in prose, and the systemic version is [#1851](https://github.com/gradionhq/margince-poc-v1/issues/1851).

## What only Postgres can prove

Both are correlated subqueries, so what they compile to is readable in the unit lane and what they **select** is not. Four integration scenarios cover that half, using the harness's own seeding (Rep1 and Rep2 share Team1, Rep3 sits in Team2):

- a team filter reaches every member's records and no other team's;
- `exists: false` selects an unowned record as covered by no team (the owner-in-no-team arm needs a fixture the harness lacks — [#1852](https://github.com/gradionhq/margince-poc-v1/issues/1852));
- an account that is a customer **and** a supplier is selected as a supplier;
- withdrawing that relationship stops it matching.

The fixture uses `supplier` rather than `partner` deliberately, and the comment says why: `partner` carries a product invariant of its own — an account is a partner only while a partner programme row backs it — so borrowing it would make these tests fail for a reason unrelated to the leaf under test. I found that out by having the invariant refuse the write, which is the invariant doing its job.

They live in `linkedleaves_integration_test.go` — named for what unites them, leaves that reach a row other than the one being selected.

## `domain` is deliberately not here

It is the third leaf #1657 listed, and it is **not** a third map entry. The list path folds the caller's value through `values.ParseDomain` before comparing, so a leaf that skipped that would answer nothing for a pasted URL — silently, and under the same field name. Same stored column, two different answers, and the wrong one invisible.

Closing it properly needs either a normalizing field type (a new member in the contract's vocabulary type enum, which #1657 did not anticipate) or an operand transform at the compile boundary. Filed as [#1848](https://github.com/gradionhq/margince-poc-v1/issues/1848) with both options and a recommendation, rather than shipped half-right here.

`make check-backend` green, `make craft-static` green (0 findings), and the four integration scenarios pass against real Postgres.

---

## Review round (craft-reviewer + security-redteam)

Both ran on this diff. The security pass found **no exploitable finding** — row scope holds (`query.go:predicateWhere` is the only execution point and ANDs `auth.ScopeClauseFor`), identifiers come only from Go constants, the `relationship_type` value is a bind parameter, and `neq`/`exists:false` negate the wrapper so a withdrawn row never leaks.

It also corrected a premise **I had wrong in shipped code**: I wrote in [#1837](https://github.com/gradionhq/margince-poc-v1/pull/1837) that "the RLS GUC contract binds it exactly as it binds the base table". Core 0217 (ADR-0091 phase A) retired every policy, and `database.go` says so outright — *"the GUC binds nothing on its own now"*. Verified before acting. Both tenancy comments in this file now name the real model (one organization per installation, ADR-0061) instead of a policy that no longer exists.

The craftsmanship pass found one factual error and several honesty problems, all applied:

| Finding | Applied |
|---|---|
| The "identity, not string" claim was **impossible** — `storekit.Field` is a comparable struct of three strings, so `!=` *is* the string comparison the comment disclaimed | Restated as *definitionally identical*, which is the enforceable invariant; the commit message said it wrong too and is corrected |
| The ownership gate could pass vacuously and checked one direction | Fails on an empty sweep; asserts both directions; mutation-verified both ways |
| The enum non-validation was a gap **rationalized in prose** (rule 5) | Gated by a test, prose cut to one line, systemic gap filed as [#1851](https://github.com/gradionhq/margince-poc-v1/issues/1851) |
| The header narrated the 1000-line test ceiling (rule 4: build-process residue) | Deleted |
| The comment **transcribed** the sibling's SQL, inexactly, ungated | Names the behaviour instead |
| `TestNoTeamSelectsTheRecordsNoTeamCovers` promised an arm its body did not prove | Renamed to `TestAnUnownedRecordIsCoveredByNoTeam`; the missing arm needs a teamless seat the shared harness lacks — [#1852](https://github.com/gradionhq/margince-poc-v1/issues/1852), with the blast radius spelled out |
| The 18-line note sat on the const while describing the var; listed a three-member dial with two leaves | Moved onto the leaf; `unassigned` explained as `owner_id exists:false` |
| Filename described half its contents | `linkedleaves_integration_test.go` |
| Needless local `desired := types` | Dropped |

The security pass's `%s`-verb-count nit is applied as a **sweep over every link field in every engine** rather than a per-field assertion, so a new link leaf is covered the day it lands. Mutation-verified.

One judgment call I did **not** take: adding a teamless seat to `integration.Setup` to close that test arm here. Four suites count `app_user` rows, so it belongs in its own change with the full lane behind it.

### DIFF
--- a/backend/internal/compose/integration/collections/linkedleaves_integration_test.go
+++ b/backend/internal/compose/integration/collections/linkedleaves_integration_test.go
@@ -1,0 +1,200 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+//go:build integration
+
+package collections
+
+// The two vocabulary leaves that reach a row OTHER than the one being selected:
+// the ownership dial's team half, which walks team_membership, and an account's
+// relationship to us, which is multi-valued and lives in its own table.
+//
+// Both are correlated subqueries, so what they compile to can be read in the unit
+// lane but what they SELECT cannot. These are the scenarios only real Postgres
+// answers — and the harness already seeds the shape they need: Rep1 and Rep2
+// share Team1, Rep3 sits in Team2.
+
+import (
+	"testing"
+
+	collectionsmod "github.com/gradionhq/margince/backend/internal/modules/collections"
+	peoplemod "github.com/gradionhq/margince/backend/internal/modules/people"
+	"github.com/gradionhq/margince/backend/internal/shared/kernel/ids"
+)
+
+// ownedOrg seeds one account owned by one seat, through the real writer.
+func (f fixture) ownedOrg(t *testing.T, name string, owner ids.UUID) ids.UUID {
+	t.Helper()
+	seat := ids.From[ids.UserKind](owner)
+	org, err := f.people.CreateOrganization(f.ctx, peoplemod.CreateOrganizationInput{
+		DisplayName: name, OwnerID: &seat, Source: "manual",
+	})
+	if err != nil {
+		t.Fatalf("create organization %q: %v", name, err)
+	}
+	return ids.UUID(org.Id)
+}
+
+// "Owned by my team" is the form a manager's saved view actually takes, and the
+// engine had only the person half of the dial until now. The team is walked
+// through team_membership, so this is the whole membership edge under test and
+// not just a column comparison.
+func TestATeamFilterSelectsEveryMembersRecords(t *testing.T) {
+	f := setupFixture(t)
+	firstMember := f.ownedOrg(t, "Held by Rep1", f.e.Rep1)
+	secondMember := f.ownedOrg(t, "Held by Rep2", f.e.Rep2)
+	otherTeam := f.ownedOrg(t, "Held by Rep3", f.e.Rep3)
+
+	list, err := f.lists.CreateList(f.ctx, collectionsmod.CreateListInput{
+		Name: "my team's accounts", EntityType: "organization", ListType: "dynamic",
+		Definition: map[string]any{
+			"field": "owner_team_id", "op": "eq", "value": f.e.Team1.String(),
+		},
+	})
+	if err != nil {
+		t.Fatalf("create list: %v", err)
+	}
+
+	got := memberIDs(t, f, list.ID)
+	// Both members of the team, not just the one who happens to be the caller:
+	// the dial names a team, and a filter that answered only the viewer's own
+	// rows would be `owner_id` wearing a different label.
+	if !got[firstMember] || !got[secondMember] {
+		t.Errorf("the team filter missed a member's records: %v", got)
+	}
+	if got[otherTeam] {
+		t.Error("the team filter reached another team's records")
+	}
+}
+
+// `exists: false` on the team leaf finds the records no team's review covers, and
+// an unowned record is one of them.
+//
+// It does NOT cover the other arm — a record whose owner exists and belongs to no
+// team — and the name says so, because that arm is what distinguishes this leaf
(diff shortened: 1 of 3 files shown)


dungbanguyen:
Applied — 612e1c42.

The finding is right and it names a real inconsistency rather than a style preference: the sibling helper in this suite (`dealForCustomer`, from #1837) explains that going through `CreateDeal` is what makes the joined column the one production stamps, while `ownedOrg` only restated its own signature.

Kept the why rather than deleting the comment, because it is load-bearing for CLAUDE.md's rule 6: a hand-inserted row could carry an owner that no membership edge reaches, and the team join would then be proven against a shape the product cannot produce.

`craft static --strict` still green (0 findings).

dungbanguyen:
**The four red checks here are main's, not this diff's.** Recording the diagnosis so nobody re-investigates it.

All four trace to one failure in `backend/migrations`:

```
--- FAIL: TestEveryMigrationTakingABlockingLockBoundsHowLongItWaits/core
    core/1787133348_relationship_traversal_indexes.up.sql takes a lock that blocks
    writers on a table it did not create, and never bounds the wait.
```

That migration is not in this branch — this commit touches three files, none of them migrations (`git show --name-only HEAD`) — and it **is** on `origin/main`. I checked the file: four `CREATE INDEX` on `relationship` with no `SET LOCAL lock_timeout`. Main's own push run (32247790984) is red for the same reason. PR checks build `refs/pull/N/merge`, so this branch inherited it.

The cascade: `deterministic-gates`, `extension-reference` and `integration unit coverage` each run that package; `integration` then gates on the unit pass. Worth noting the integration **shards** reported `SHARDS_RESULT: success`, so the four new scenarios in this PR passed — the failing lane is the unit one.

Already being fixed by https://github.com/gradionhq/margince-poc-v1/pull/1860, so I am not duplicating it. Once that lands I will re-trigger this PR's checks rather than rebasing, since nothing here needs to change.

Issue #1928, opened 2026-08-20
The provider-run worker writes audit rows that say actor_type='system' with a connector actor_id


tinnhlt-gra:
`providerJobActor` in `backend/internal/compose/jobs_providerruns.go` binds a
principal whose two halves disagree:

```go
principal.WithActor(ctx, principal.Principal{
    Type: principal.PrincipalSystem, ID: "connector:surfe",
})
```

`storekit.Audit` stamps `actor_type` from the **type** and `actor_id` from the
**id**, so every audit row this path writes says `actor_type='system'` beside
`actor_id='connector:surfe'`. `principal.PrincipalConnector` exists and
`audit_log_actor_type_check` has admitted `'connector'` since migration `0016`,
so the honest value is available and unused.

**This writes rows today.** `people:WriteProviderClaims` is on this path and is
audit-only-ratified, so it audits on every polled run. `audit_log` is append-only
— the rows already written cannot be corrected, and every further run adds one.

**The fix is one word:** `PrincipalSystem` → `PrincipalConnector`. There is no
granting human on this path (the run is scheduled, not requested), so it stays a
bare connector with no `OnBehalfOf`; `storekit.OwnerOrActor` already reads that
as the row nobody owns.

**Check what else moves before flipping it.** `auth.AuthzRule` short-circuits to
the literal `"system"` for `PrincipalSystem` and falls through to
`p.Permissions.Rule(...)` for anything else, so the `authorization_rule` column
on this path changes shape. That is a better answer, not a worse one, but it is a
behaviour change and wants its own assertion.

Filed from #1795, which fixed the identical contradiction in
`jobs_finance.go` and deliberately did **not** touch this one so the two changes
stay separately reviewable. Related but distinct from #1127, which is about
workers that bind *no* actor rather than a self-contradictory one.

tinnhlt-gra:
Found while implementing #1861: this defect has an **agent-flavoured sibling**
in the same family, which belongs here rather than as its own issue.

`compose/captureverdict.go:112` and its siblings write `PrincipalSystem` with
an `agent:`-prefixed id:

```go
const verdictActor = "agent:" + verdictReason
...
principal.WithActor(ctx, principal.Principal{
    Type: principal.PrincipalSystem, ID: verdictActor,
})
```

So the row says `actor_type='system'` beside `actor_id='agent:<kind>'` — the
same self-contradiction this issue reports for the connector case, one actor
kind over. The comment there explains the `agent:` prefix as a contract
[...]

Issue #1927, opened 2026-08-20
Decide whether the contract carries finance event types, so the mirror can stop being audit-only


tinnhlt-gra:
The finance mirror now commits an `audit_log` row beside every row it writes, and
deliberately publishes **nothing**. Two entries in `auditOnlyWrites`
(`backend/writeshape_test.go`) ratify that, and this issue is the decision they
are waiting on.

## Why there is no event today

The event catalog is closed: a type exists because the contract declares one, and
a build may not mint a type to satisfy the write-shape rule. The catalog carries
no finance verb at all — `grep -icE "invoice|finance|payment"` over
`backend/internal/shared/kernel/events/catalog.go` and
`backend/api/public-events.yaml` returns 0 and 0, and
`backend/internal/shared/kernel/events/events_test.go` pins that `invoice.created`
is not a known type.

Neither adjacent type fits:

- `mirror.*` is the overlay write-back stream. Staging an accounting fact under
  it routes the envelope to subscribers watching for something else entirely.
- `organization.updated` would tell every subscriber that a company record
  changed when none did — the same argument that already makes
  `people:WriteProviderClaims` audit-only.

A wrong envelope is acted on; an absent one is not. So the mirror is audit-only
until this is decided.

## The decision

Should the contract carry finance event types, and which? The three proposed in
the original report:

| Type | Announces |
|---|---|
| `finance.connection.changed` | the accounting source connected, failed, or recovered |
| `finance.sync.completed` | a pass finished, with what it wrote |
| `finance.invoice.changed` | a mirrored invoice appeared or was restated by the source |

**Recommendation:** ratify `finance.invoice.changed` and
`finance.connection.changed`, and decline `finance.sync.completed`. The first two
name a fact a consumer could act on — money moved, or the figures stopped being
trustworthy. The third announces that a sweep ran, which is operational telemetry
with no subject; the pipeline-event class exists for exactly that shape if it is
wanted at all.

## What ratifying costs

Per type: a `StreamFor` entry, a payload schema in the contract, `make gen`, and
the six gates in `backend/publicevents_test.go`. Then the two
`internal/modules/finance:auditFinance*` entries come out of `auditOnlyWrites`
and the writers call `storekit.EmitEvent` — the call sites are already
single-spelled in `backend/internal/modules/finance/audit.go`, so nothing else
moves.

## What is NOT blocked on this

Nothing that bites today. No real provider exists — the only reader is the
offline generator — and the audit half, which is the half that cannot be
[...]

Issue #1802, opened 2026-08-19
The write-shape gate catches audit-without-emit but not never-audits-at-all, so a module can own tables and audit none of them


LarsGradion:
Approved 2026-08-19 alongside #1788, its sibling. Both are the same species: a gate that reads as complete and is not.

`backend/writeshape_test.go` asserts that every mutation writing an `audit_log` row also commits a paired `event_outbox` event. A mutation writing **neither** has no audit row to pair, so nothing trips. The gate holds half the obligation and reads as holding all of it.

**The worked example is #1795.** No file in `backend/internal/modules/finance/` calls `storekit.Audit` or `storekit.EmitEvent`, while `sync.go:16-19` states the house write shape as though it does. The file calls `CapturedBy` and `LockRow`, which is why it reads as compliant. That comment stood for months and no gate could contradict it.

The finance outcome turned out to be that the module is *right* not to audit mirrored ERP rows — the leading system holds that trail (see #1795). Which is precisely the point: **the gate would have forced that reasoning to be written down at the time, instead of leaving a false comment as the only statement of intent.**

## What to build

A test that fails when a module owns tables (declared in its `doc.go`, per ADR-0014 and already read by `backend/tableownership_test.go`) and no write path in it audits, unless the module states why.

The exception carries its reasoning in-source, the way the existing waiver map in `writeshape_test.go` already does — those entries are prose explaining a decision, not suppressions, and that is the form to keep. Finance becomes the first entry: mirrored rows are a read-only projection of a leading system that holds the audit trail for their content.

## Note on scope

Derive the module list from the tree rather than maintaining it — same rule as `backend/arch_test.go`, so a new module is covered the day it appears rather than the day somebody remembers to add it.

Worth pairing with #1788 in one change: both derive their subject from module ownership declarations and would share most of their machinery.

tinnhlt-gra:
Shipped in #1951.

A module that owns tables (per `tableOwners`) and calls no audit writer now fails `make check-backend` and names itself.

**Two things in this issue turned out to be wrong, and the PR says so rather than working around them:**

**The `doc.go` subject source.** This issue proposed deriving ownership from each module's `doc.go`, "already read by `tableownership_test.go`". Nothing parses `doc.go` — `tableOwners` is a hand-maintained map and the neighbouring "kept in sync" is prose. A doc.go-derived census would have exempted seven modules outright, including the one that held the original defect.

[...]

PR #1936, opened 2026-08-20
fix(finance): every mirrored row commits its audit row in the same transaction

### LINKED ISSUE #783: The finance mirror writes invoices and payments without an audit row or an outbox event (FIN-AC-10)
Found by Codex reviewing PR #780; deferred deliberately and merged, because the
mirror is only reachable through the offline demo provider today — no real
accounting data flows through it.

## What is wrong

Every invoice and payment the finance mirror writes commits with raw SQL and
**no `audit_log` row and no `event_outbox` row**. The connection status update
in `RecordSyncFailure` has the same gap.

- `backend/internal/modules/finance/mirror.go` — `insertInvoice`, `updateInvoice`
- `backend/internal/modules/finance/mirrorpayment.go` — `mirrorPayment`
- `backend/internal/modules/finance/sync.go` — `RecordSyncFailure`, `markSynced`

`grep -n "storekit.Audit\|storekit.Emit" backend/internal/modules/finance/` returns nothing.

## Why it matters

The spec is explicit. `finance-ingestion.md` FIN-AC-10:

> Given one materially changed invoice during incremental sync, then exactly one
> domain row update, one audit entry and one outboxed event exist for it, in one
> transaction; a rolled-back pass produces none.

And FIN-AC-9 requires the mirror image of it: an unchanged pass writes zero rows,
zero audit entries, zero events. The hash discipline already delivers the FIN-AC-9
half correctly — it is only the write side that is missing.

This is also the repo-wide write shape (CLAUDE.md, "The write shape
(non-negotiable)"): domain row + audit_log + event_outbox in ONE transaction.

Consequences while it is missing:
- No audit trail for mirrored money. "Which records did this sync touch?" is
  answered from `audit_log`, so the answer is silently empty.
- No events, so nothing downstream can react to an invoice going overdue.

## What the fix needs

1. Declare three events in `backend/api/public-events.yaml` — none exist today:
   - `finance.connection.changed` (FIN-EVT-1)
   - `finance.sync.completed` (FIN-EVT-2) — one per run, the only event a backfill emits
   - `finance.invoice.changed` (FIN-EVT-3) — one per materially changed invoice,
     never from a backfill and never from a no-op
2. `make gen` to regenerate the payload structs.
3. Thread `storekit.Audit` + `storekit.EmitEvent` through the mirror writes,
   inside the existing transaction.
4. Satisfy the census gates: `publicevents_test.go`, and check
   `backend/internal/platform/events` registration.
5. An integration test for each of FIN-AC-9 and FIN-AC-10 — the row/audit/event
   counts before and after a pass, on both a changed and an unchanged ledger.

Note ADR-0084's distinction: a backfill is ONE operation with one audit entry and
one `finance.sync.completed`; incremental sync is one audit + one event per
changed row. The current code does neither.

## Also from the same review, smaller

- `providerFor` errors return before `RecordSyncFailure`, so a configured but
  unsupported provider leaves the connection reporting `connected`/`syncing`
  rather than `error` (`backend/internal/compose/jobs_finance.go`).
- The sync principal is `PrincipalSystem` with id `connector:finance`; once audit
  rows exist that persists contradictory provenance. Should be a connector type.
- A large first sync runs in one transaction with no cursor, so a five-minute
  timeout rolls the whole pass back and every retry restarts from the first
  customer — the job's own `reason` claims it resumes.

### LINKED ISSUE #1795: The finance mirror writes invoices with no audit row and no event, and its own comment says it does
## What happens

No file in `backend/internal/modules/finance/` calls `storekit.Audit` or `storekit.EmitEvent`. Grep the whole module: zero hits.

It does write domain rows. `mirror.go:283` inserts into `finance_invoice` and `:326` updates it; `sync.go` inserts into `finance_external_customer` and updates `finance_connection`.

So a mirrored invoice appearing, changing state, or being re-synced leaves no audit row and puts nothing on the outbox.

## The comment says the opposite

`backend/internal/modules/finance/sync.go:16-19`:

> The write shape is the house one: the mirrored row, its audit row and its outbox event in one transaction, `captured_by` from the authenticated principal. A sync runs as the system principal, which is what makes those rows say a connector wrote them rather than a person.

That paragraph is false. The file calls `storekit.CapturedBy` and `storekit.LockRow` — which is why it reads as though it follows the shape — and never reaches the two calls that would make the sentence true.

This is the failure mode the review-loop rules already name: a comment claiming a guarantee is a claim to check, and here it stopped anyone checking.

## Why the write-shape gate did not catch it

`backend/writeshape_test.go` asserts that **every mutation writing an audit row commits a paired outbox event**. Its own header says so. A mutation that writes *neither* is invisible to it — there is no audit row to pair, so nothing trips.

The gate catches half the obligation. A store that opts out of the write shape entirely passes.

## Failure scenario

1. A finance connector syncs. An invoice's state moves from open to paid.
2. `finance_invoice` is updated. No audit row is written. Nothing reaches the outbox.
3. An operator asks who changed that invoice and when. `audit_log` has no answer, and the general "which records did X touch" query — which is derived from `audit_log` — cannot see it.
4. Any consumer waiting on a finance event never fires, because none was emitted.

The gap also reaches privacy: erasure and retention reason about what was touched via the audit trail, so mirrored finance rows are invisible to that reasoning.

## Deciding before fixing

Two questions, and the first is genuinely open:

**Is a mirrored row a record fact or operational plumbing?** The write shape exists for facts about customer records. A mirror of an external system's state might reasonably be treated as a projection rather than a mutation — in which case the fix is the comment, plus an explicit ratified waiver in `writeshape_test.go` saying why. Several such waivers already exist and each carries its reasoning.

If it *is* a record fact — and an invoice's paid state looks like one — then the fix is the write shape properly: audit row plus outbox event in the same transaction, with the event type added to the catalog.

**Second, the gate should catch the other half.** A module that owns tables and writes them without ever calling `Audit` should have to say why in the same allow-list, rather than passing by writing nothing. That is a separate change and probably the more valuable one, since it closes the class rather than the instance.

Found while migrating ADR-0084 into `docs/adr/`; the record states the divergence rather than repeating the comment's claim.

tinnhlt-gra:
`grep -rn "storekit.Audit\|storekit.Emit" backend/internal/modules/finance/` returned
**nothing**. Six write statements on four tables — invoices, payments, the source's
customer directory, the connection's status — and not one `audit_log` row between them.

Money arrived in the mirror and nothing recorded that it had. **Audit rows cannot be
backfilled**, so every invoice mirrored before this stays permanently unaccounted for,
and the erasure and retention reasoning that reads `audit_log` is blind to it.

Closes #1795 and #783. (#783 is the fuller statement of the same defect, not a
duplicate to be dropped.)

## Four comments asserted the guarantee that was absent

This is the defect class this repo names outright: grep the stale CLAIM to find the gap.
Each is now true or gone — quoted as it stands today:

| Where | Now reads |
|---|---|
| `finance/sync.go` | *"The write shape here is the mirrored row and its audit row in ONE transaction, `captured_by` stamped from the connector principal the sweep runs as… There is no outbox event, and audit.go carries the whole reason"* |
| `finance/mirror.go` | *"Rewriting the row would bump its version and write an audit row for a change that did not happen — history minted four times a day, forever, by a sweep that read an unchanged ledger"* |
| `migrations/core/0202` | *"an unchanged content hash writes no row, no audit entry and no version bump. It says nothing about events, because the mirror publishes none at all"* |
| `rbacgate_test.go` | *"under the worker's connector principal… Every row it writes commits its audit_log row in the same transaction, so what a grant would have decided is instead RECORDED"* |

## Audit-only, and the waiver states the rule rather than citing it

Two entries in `auditOnlyWrites`, one rationale covering both. The event catalog is
**closed**: a type exists because the contract declares one, and a build may not mint
one to satisfy the write-shape rule. The catalog carries no finance verb at all —
`grep -icE "invoice|finance|payment"` over `catalog.go` and `public-events.yaml` is
`0` and `0`, and `events_test.go` pins that `invoice.created` is not a known type.

Neither adjacent type fits. `mirror.*` is the overlay write-back stream, so staging an
accounting fact under it routes the envelope to subscribers watching for something
else. `organization.updated` would report a company record change that did not
happen — the same argument that already makes `people:WriteProviderClaims`
audit-only. **A wrong envelope is acted on; an absent one is not.**

Which finance types the contract should carry is a product decision, filed as **#1927**
(`status: needs-decision`) with the three candidates, a recommendation, and what
ratifying each costs. Ratifying them removes exactly these two waiver entries — the
call sites are already single-spelled in `finance/audit.go`.

## Three things the audit row forced into the open

**The connector principal contradicted itself.** `PrincipalSystem` beside
`ID: "connector:finance"` writes `actor_type='system'` with a connector `actor_id`.
Invisible precisely because nothing audited; the moment one row is written the first
rows are wrong, and `audit_log` is append-only. Now `PrincipalConnector`, with no
`OnBehalfOf` — finance has no connect flow and so no granting human, which
`storekit.OwnerOrActor` already reads as the row nobody owns. The harness literal moved
with it.

> The identical contradiction in `compose/jobs_providerruns.go` is **#1928**. Deliberately
> **not** fixed here so the two stay separately reviewable — and because flipping it
> changes that path's `authorization_rule` shape, which wants its own assertion. I am
> naming the divergence rather than leaving it to read as inconsistency.

**`RecordSyncFailure` had no id predicate** and marked *every* live connection failed.
Harmless only while an installation holds one. It now resolves the connection the pass
ran against, the same way the pass resolved it — and it had to, because an audit row per
affected connection needs the ids.

**`source` was hardcoded** to `OfflineProviderName` at three sites rather than
`provider.Name()`, so a real provider's invoices would have been stamped as the offline
generator's. Same "not honest about who wrote this" family; fixed here.

## A connection audits its transitions, not its heartbeat

The sweep rewrites `last_attempt_at` every six hours whatever happened. An audit row for
that would file four rows a day per connection saying nothing changed, and the two
transitions anybody reads this history for — the source went down, the source came
back — would be buried in them. `connectionImage` is compared **by value** to decide,
which is why `last_error_code` is a plain string and not a pointer: two pointers holding
the same code are not equal, so a pointer field would report a transition on every failed
pass.

## Proof

**RED** — the new integration test on unmodified `origin/main`:
```
--- FAIL: TestEveryMirroredRowWritesItsAuditRowAndASecondPassWritesNone
    0 audit_log rows for finance_invoice, want 40 — mirrored money nobody can account for
    0 audit_log rows for finance_payment, want 37
    0 audit_log rows for finance_external_customer, want 1
```

**CONTROL** — `make check-backend` on unmodified `origin/main` ran to completion and
passed over this defect. That is the hole #1802's gate exists to close, and it is why
this lands first.

**GREEN** — the test asserts a **delta, not a total**: a second pass over an unchanged
source moved neither the audit counts nor the outbox count.

**MUTATION** — removing the `storekit.Audit` call from `auditFinanceCreate`: the new
test fails, nothing else does.

**INVERSE MUTATION** — restoring `RecordSyncFailure`'s unpredicated `UPDATE`:
```
the bystander connection reads "error" after another source failed, want "active"
0 connection audit rows for one failed source, want 1
```

**ANTI-VACUITY** — the outbox assertion was first written against
`envelope->>'workspace_id'`, a member the envelope does not have, so it would have
passed over a mirror that published on every row. It matches the entity type instead,
and a planted outbox row moves it `0 → 1 → 0`.

Local: `make check` full pass · `make test-integration` — *OK: integration passed with
0 skips (41 packages)* · `make craft-static` — *PASS, 0 blocker, 0 major, 0 minor*.

## Scope truth

**This is hygiene today, and nobody should drop other work for it.** No real provider
exists — `compose/jobs_finance.go` refuses every provider but the offline generator, and
the contract exposes one finance operation, a GET. No customer's money moves through
this path. `priority: high` is a claim about the class, not today's blast radius.

The other half is why it still could not wait: **the day a real provider lands, this
becomes retroactively impossible to fix.**

## Housekeeping

- **No UAT video is owed** — there is no UI on this path. The proof is the red/green
  above.
- New-code coverage is measured normally; this is product code, not a gate.
- `make check` was red on `origin/main` when this branch was cut, for an unrelated
  reason (`release.yml`). Filed as #1932 and fixed in #1933, which is merged; this
  branch is rebased onto it.

Closes #1795
Closes #783

---
## Summary by cubic
Mirrored finance writes now commit an `audit_log` row in the same transaction; previously invoices, payments, external customers, and connection status wrote with no audit history. The mirror remains audit-only (no outbox) until finance event types are ratified.

- Commits audit rows for `finance_invoice`, `finance_payment`, `finance_external_customer`, and connection state transitions; skips both write and audit on unchanged hashes; reads the before image from the locked row and records update deltas. Invoice/payment images now include every column the sync writes (plus `fx_rate_to_base`) so the trail explains the change; connection images record clearing the error code explicitly.
- Replaces upserts on external customers with read‑modify‑write under `FOR UPDATE`; uniqueness collisions roll back so the verb (`create` vs `update`) stays correct and auditable.
- Uses `FOR UPDATE` and `storekit.LockRow` to prevent races; adds typed audit images and transition logic in `backend/internal/modules/finance` (`audit.go`, `mirror.go`, `mirrorpayment.go`, `sync.go`). Ratified as audit‑only at the concrete call sites in `auditOnlyWrites`.
- Runs under the connector principal; stamps `source` from `provider.Name()`. Audit rows now render `authorization_rule` as `"system"` for a bare connector (no human), keep `actor_type: "connector"`, and render a rule when a connector acts on behalf of a user.
- Fixes `RecordSyncFailure` to target only the connection the pass resolved, audit the transition, and leave `last_success_at` unchanged; connection audits only transitions, not the heartbeat.

**Rollout**
- No schema or config changes; no backfill is possible — existing mirrored rows remain unaudited.
- Ensure workers execute with the connector principal; expect no finance `event_outbox` rows until the contract adds finance event types.

<sup>Written for commit 9a898a18a80737ba82432b62b646b89d30bdd7dd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1936?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tinnhlt-gra:
### Coverage, and a second commit closing a real gap in it

Measuring the changed files after the first commit found two new functions at **0.0% of statements** — `updatePayment` and `updateExternalCustomer`. The reason is structural, not an oversight in the test: **the offline generator is deterministic by design**, a source that never changes its mind, so every path that records a *change* is unreachable through it. The first commit's tests exercised the create arms thoroughly and the update arms not at all.

`244d4e0` adds a double at the **provider seam and nowhere else** — the store, the SQL, the images and the audit writer under test are all the real ones — and restates one invoice, one payment and one customer name between two passes.

It does not just count rows. It reads the two images back off the `update` audit row and asserts `11900 → 13090` on the invoice and `5000 → 6000` on the payment, which is what proves the claim the code comment makes: **the before image is read off the ROW, not rebuilt from the source.**

That assertion carries its own defect test. Making `updateInvoice` pass the after image as both arguments:

```
--- FAIL: TestARestatedSourceAuditsTheUpdateWithBothImages
    the invoice's update reads 13090 → 13090, want 11900 → 13090
```

A count-only assertion passes that mutation happily — which is the shape of a guard that has never been run against the bug it describes.

After it: **83.0% of statements** for the package, and no function in the changed files at zero. The sub-80% remainder is error-return branches (`if err != nil { return err }`).

### Reviewers so far

CodeRabbit (pass) and Fable (`/code-review`, in flight). **Codex has not read this** — it is `disable-model-invocation` and needs a human to type `/codex:review`. I have asked for it. If it merges without, the count is two, not three, and this comment is the record of which.

tinnhlt-gra:
## Review response — 5 of 6 fixed, 1 refuted with evidence

`/code-review` returned six findings. Each fix below carries a **mutation** showing the test that now covers it goes red against the defect — this repo has had four fitness functions pass against the bug they described, so a fix without one is not finished.

### 1. `authorization_rule` degraded to a policy the sweep never held — **fixed, and it was the serious one**

`AuthzRule` returns the literal `"system"` only for `PrincipalSystem`. Flipping to `PrincipalConnector` sent it through `Permissions.Rule`, and the worker principal holds no permissions, so every finance row this PR starts writing would have recorded:

```
role[] finance_invoice.create row_scope=
```

which reads years later as a principal that **held** a role and a row scope and had neither. `audit_log` is append-only — the exact stake this PR argues from.

`AuthzRule` now answers `"system"` for a connector with **no human behind it**. That is not a new concept: `storekit.OwnerOrActor` already applies the same test when it leaves such a row ownerless. A connector acting **for** somebody — capture's, which carries the granting human's own RBAC — still renders its rule, because there a grant really did admit the call.

The reviewer's sharpest point was that **the fixture hid it**: `setupFinance` seeded `admin`/`RowScopeAll`, so harness rows read `role[admin] … row_scope=all` while production's read empty. The mutation shows both faces —

```
unit  (production shape): renders "role[] finance_invoice.create row_scope="
integ (fixture shape):    records [role[admin] finance_invoice.create row_scope=all …]
```

— and the new unit test uses the production shape. `TestOnlyAConnectorWithNoHumanBehindItRecordsNoRule` asserts **both directions**, so a fix that blanket-exempted every connector fails it.

`authorization_rule` is written only by storekit and exposed read-only in the contract; nothing parses it.

### 2. `invoiceImage` was narrower than the change key — **fixed**

`invoiceHash` covers net, tax, dates and the flags, so any of them can be the sole reason a row is rewritten — and the audit row would then differ only in `sync_hash`. The image now carries every column the sync writes, plus `fx_rate_to_base`.

The update test is rewritten to the reviewer's own scenario: net `10000 → 11000`, tax `1900 → 900`, **gross unchanged at 11900**. Mutation (drop net/tax from the image):

```
the invoice's update reads net <nil> → <nil>, want 10000 → 11000
```

### 3. fx-repair path wrote `before == after` — **fixed by 2**

That path rewrites a row on an *unchanged* hash precisely to fill in `fx_rate_to_base`. With the rate in the image the two sides now differ, which is why it is in there despite the hash not covering it.

### 4. Customer upsert lost `ON CONFLICT` — **refuted, with evidence**

The premise is that two passes can overlap. They cannot: `workspaceSweepOpts` sets `UniqueOpts{ByArgs: true, ByState: activeSweepStates}`, and `dispatch.go:91` states that **retryable is one of those states**, so a backing-off job suppresses the tick's own re-enqueue of that workspace. There is no manual trigger — the contract exposes one finance operation, a GET.

The read-then-branch is also what the invoice and payment writers already do, and it is what lets the pass know **which verb it is writing**. A rollback is recoverable; an append-only verb is not. The rationale is now in the source rather than only here.

### 5. `omitempty` erased the cleared error code — **fixed**

Dropped. On recovery the after image's job is to say the code was **cleared**, and a projection walking its keys cannot record a change to a key that is not there.

### 6. Waivers keyed on two generic helpers — **fixed**

Moved onto the **seven write sites**. Keyed on a helper, an eighth finance write would have inherited a ratification somebody wrote about four tables and `AssertAllMatched` would have kept reading green. Both waiver mutations verified: deleting an entry names the module, and mis-keying one trips the staleness sweep —

```
waiver internal/modules/finance:auditFinanceCreateXX matched no subject
```

### Gates

`make check` full pass · `make test-integration` — *OK: integration passed with 0 skips (41 packages)* · `make craft-static` — *PASS, 0 blocker, 0 major, 0 minor*.

Reviewers: CodeRabbit, Fable. **Codex still has not read this** — it needs a human to type `/codex:review`.

Issue #1839, opened 2026-08-19
The uat lane can hang indefinitely: no job timeout, and its install step runs apt


dungbanguyen:
## What happened

On [#1830](https://github.com/gradionhq/margince-poc-v1/pull/1830) the `uat (AC screens + axe WCAG 2.2 AA + 390px, mocked)` job hung for **40+ minutes** on a lane that normally finishes in **4m39s** (its own timing on [#1799](https://github.com/gradionhq/margince-poc-v1/pull/1799) an hour earlier, same code path).

Every other check on the PR was green. The one hung job held `mergeStateStatus` at `UNSTABLE`, so the PR was unmergeable with nothing actually wrong with it.

The API says exactly where it stopped:

```
$ gh api repos/gradionhq/margince-poc-v1/actions/jobs/96005123818 \
    --jq '"status=\(.status) steps_running=\([.steps[]?|select(.status=="in_progress")|.name]|join("|"))"'
status=in_progress steps_running=Install deps and the Playwright Chromium browser
```

It never reached a single test. The step is:

```yaml
- name: Install deps and the Playwright Chromium browser
  working-directory: frontend
  run: |
    pnpm install --frozen-lockfile --ignore-scripts
    pnpm exec playwright install --with-deps chromium
```

`--with-deps` shells out to `apt-get` inside the runner. A slow or unresponsive mirror there hangs with no output and no bound.

**The change under review was not the cause** — `make frontend-e2e` runs clean on that branch locally in 27s (91 passed, 14 skipped, 0 failed).

## The actual gap

`.github/workflows/ci.yml` gives the `uat` job **no `timeout-minutes`**, so it inherits GitHub's 360-minute default. A stalled apt therefore burns up to six hours and blocks the PR for all of it, and the only signal a human gets is a check that says `pending` — indistinguishable from "still working".

I recovered by cancelling the run and re-triggering, which is the same manual dance a stuck run needed earlier the same day (that one sat `queued` with **zero jobs allocated** for 2h19m on #1799). Two incidents in one morning, both recovered by hand, both invisible until somebody read the job API.

## Suggested fix

1. **Give `uat` a `timeout-minutes`** — its p99 is under 6 minutes, so something like 20 is generous and still fails 18× faster than the default. Worth doing for every job whose install step reaches the network; `uat` is simply the one that bit.
[...]

PR #1939, opened 2026-08-20
feat(contract): the filter vocabulary says what an id field points at

dungbanguyen:
A filter clause on an id field is a reference to a record, and the vocabulary said only `type: id`. A client with that much can render a text box and hope the reader pastes a correct uuid — which is exactly what the Filters & views builder does today for `tag`, `owner_id`, `owner_team_id`, `stage_id`, `pipeline_id`, `organization_id`, `partner_org_id` and `project_id` alike. `ScalarValue` handles `date`, `boolean` and the numerics and falls back to text.

`FilterVocabularyField.references` closes that: present for every id field, absent for every other type, naming one of seven record types. A builder can then offer the tags, seats or stages that exist.

This is the information half. The picker that consumes it is the next change — filed as one PR each so the contract addition is reviewable on its own terms.

## Why not a client-side map

That was the alternative, and it is rejected for the reason this vocabulary exists at all. A map from field NAME to target is a second list of the engine's fields, and the day the engine gains an id leaf the map does not know about it — the picker silently falls back to a uuid box and nothing fails.

That is not hypothetical: two id leaves landed **yesterday** — `owner_team_id` ([#1850](https://github.com/gradionhq/margince-poc-v1/pull/1850)) and the deal's reach into its customer ([#1837](https://github.com/gradionhq/margince-poc-v1/pull/1837)). A client-side map would already be stale twice over. It is the same drift `operators` is here to prevent, one type down.

## Where the declaration lives

On `storekit.Field`, beside `Type`, not in the vocabulary read. The compiler has no use for it — an id compares as an id whatever it refers to — but a field is *declared* there, and keeping it there is what lets a gate hold every id leaf to naming a target. In the vocabulary read it would be a lookup table for a new leaf to fall out of.

## Three gates, each derived rather than listed

| Gate | What it refuses |
|---|---|
| `TestEveryIDFieldDeclaresWhatItReferences` | An id field naming no target — and any *other* type naming one, since a target on a text column promises a picker for a column holding no reference. Sweeps every engine's every field and **fails when it sweeps nothing**, so it cannot pass vacuously. |
| `TestTheVocabularysReferenceEnumIsExactlyWhatTheEngineDeclares` | The engine and the contract disagreeing, in either direction. It reads the two halves from **independently written sources** — taking both from the contract would only prove it agrees with itself. |
| The existing operator/type enum gates | Unchanged, and now joined by a third of the same shape. |

Both new gates mutation-verified: dropping `stage_id`'s declaration, and adding an engine target the contract lacks, each fail by name.

No custom field is ever id-typed (the catalog's six types have no id), so this only ever describes a core field, and the enumeration is closed for the same reason `operators` is.

## Regenerations

All three ran, per the house rule that a contract change is three: `make gen`, `pnpm gen:api`, and the MCP surface record — which is **unchanged**, since `/filters/vocabulary` carries no `x-mcp-tool`. Running it was the check, not a no-op I assumed.

The frontend drift gate initially failed here and was right to: it diffs the working tree, so a regenerated artifact has to be committed, which its own message says. Not a defect.

## Rebase note

`TestEveryWorkflowJobCarriesATimeoutCeiling` — the gate [#1849](https://github.com/gradionhq/margince-poc-v1/pull/1849) added, acting on [#1839](https://github.com/gradionhq/margince-poc-v1/issues/1839) — was failing on `release.yml`'s `github-release` job when I started. That was main's, not this diff's (this touches no workflow), and [#1933](https://github.com/gradionhq/margince-poc-v1/pull/1933) had already fixed it. Rebased onto that rather than working around it.

`make check-backend` green, `make craft-static` green (0 findings), `make check-fe` green (3529 tests).

---
## Summary by cubic
Filter vocabulary now tells clients what each id field points at, so builders can render record pickers instead of UUID inputs. Previously it returned only type: id; now id fields include references, and non-id fields omit it.

- Contract: Add optional references to FilterVocabularyField (enum: tag | app_user | team | organization | pipeline | stage | project) in `backend/api/crm.yaml` and `frontend/src/api/schema.d.ts`.
- Engine: Add `References` to `storekit.Field` with closed targets (`RefTag`, `RefAppUser`, `RefTeam`, `RefOrganization`, `RefPipeline`, `RefStage`, `RefProject`) and `ReferenceTargets()`. Annotate all id fields across segment engines (e.g., `owner_id`, `owner_team_id`, `pipeline_id`, `stage_id`, `organization_id`, `partner_org_id`, `project_id`, and the tag link).
- Wiring: `FilterVocabulary` includes references only for id fields and omits the key for others (never sends an empty string). Filter semantics are unchanged.
- Gates: 
  - Every id field must declare a valid target; no non-id field may declare one.
  - Keep the contract enum and engine targets in lockstep via an independent set comparison.
- Rollout: No migration. Existing clients continue to work; clients may use references to offer record pickers.

<sup>Written for commit d0ada4205517469f9728352bdc258b4d4401a331. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1939?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added metadata identifying what type of record each ID field references.
  * Supported references include tags, users, teams, organizations, pipelines, stages, and projects.
  * Reference information is included only for applicable ID fields and omitted for other fields.

* **Bug Fixes**
  * Improved consistency and accuracy of reference metadata across vocabulary fields and API responses.

### DIFF
--- a/backend/api/crm.yaml
+++ b/backend/api/crm.yaml
@@ -20630,6 +20630,19 @@ components:
             may set. A retired one is absent here and still accepted by the
             engine, so no `status` lookup is needed to know whether to offer a
             field: if this operation listed it, it is offerable.
+        references:
+          type: string
+          enum: [tag, app_user, team, organization, pipeline, stage, project]
+          description: |
+            The record type this field's ids point at — present for every `id`
+            field, absent for every other type. Offer the record rather than a
+            box for its uuid.
+
+            The values are this contract's own record-type words, so no
+            translation table is needed to key a picker on them.
+
+            Never present on a custom field: the catalog's six types include no
+            `id`, so only a core field can carry a reference.
 
     Tag:
       type: object
--- a/backend/internal/compose/filtervocabularyenums_test.go
+++ b/backend/internal/compose/filtervocabularyenums_test.go
@@ -94,6 +94,17 @@ func TestTheVocabularysTypeEnumIsExactlyWhatIsFilterable(t *testing.T) {
 		"the contract advertises it and nothing filterable carries it")
 }
 
+func TestTheVocabularysReferenceEnumIsExactlyWhatTheEngineDeclares(t *testing.T) {
+	engine := map[string]bool{}
+	for _, target := range storekit.ReferenceTargets() {
+		engine[string(target)] = true
+	}
+	compareVocabularySets(t, "reference target", engine,
+		vocabularyFieldEnum(t, "references", false),
+		"the engine points id fields at it and the contract cannot spell it, so the vocabulary would report a value no client can read",
+		"the contract advertises it and no id field points at it, so no field can ever report it")
+}
+
 // everyFilterableFieldType is the six custom-field types plus id, which only a
 // core field carries. Derived from fieldcatalog.Types() so a seventh custom type
 // joins these gates by existing rather than by somebody remembering them.
--- a/backend/internal/compose/integration/collections/filtervocabulary_http_integration_test.go
+++ b/backend/internal/compose/integration/collections/filtervocabulary_http_integration_test.go
@@ -133,6 +133,10 @@ func TestTheFilterVocabularyOverHTTPOffersWhatADynamicListAccepts(t *testing.T) 
 			Type      string   `json:"type"`
 			Operators []string `json:"operators"`
 			Custom    bool     `json:"custom"`
+			// A POINTER, so this test can tell an absent key from an empty
+			// string. A plain string would decode both as "" and the
+			// omitted-for-a-non-id-field guarantee would be unassertable.
+			References *string `json:"references,omitempty"`
 		} `json:"fields"`
 	}
 	status := e.Call(t, "GET", "/v1/filters/vocabulary?resource=person", nil, nil, &vocab)
@@ -161,6 +165,21 @@ func TestTheFilterVocabularyOverHTTPOffersWhatADynamicListAccepts(t *testing.T) 
 			if f.Custom {
 				t.Error("owner_id is a core field and is reported custom")
 			}
+			// The wire value, which no unit test can reach: an id field names
+			// the record type its values point at, in the contract's own
+			// record-type word.
+			if f.References == nil {
+				t.Error("owner_id is an id field and the response carries no references key, so a builder can only ask for a uuid")
+			} else if *f.References != "app_user" {
+				t.Errorf("owner_id references %q, want app_user", *f.References)
+			}
+		}
+		// And the other arm, for every non-id field in the same response: the key
+		// is ABSENT rather than blank. "" is not a member of the contract's enum,
+		// so a strict client rejects a response that sends it.
+		if f.Type != "id" && f.References != nil {
+			t.Errorf("%s is typed %s and the response carries references=%q; the key belongs only to an id field",
+				f.Name, f.Type, *f.References)
 		}
 		if len(f.Operators) == 0 {
 			t.Errorf("%s reports no operators, so a builder could offer no clause on it", f.Name)
--- a/backend/internal/contracts/api_gen.go
+++ b/backend/internal/contracts/api_gen.go
@@ -4221,6 +4221,39 @@ func (e FilterVocabularyFieldOperators) Valid() bool {
 	}
 }
 
+// Defines values for FilterVocabularyFieldReferences.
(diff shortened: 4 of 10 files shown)


dungbanguyen:
## Review round applied (craft-reviewer + security-redteam)

Both ran on this diff. The security pass found **no exploitable finding** and confirmed several invariants explicitly: `References` cannot reach query text (every non-test read is one of three sites in `collections`; the compile path reads only `Expr`/`Type`/`Link`), no tenant-authored string can reach the enum cast (the only writers are compile-time constants, and `customFieldTypes` has no `id` member), no pointer aliasing, authz unchanged and correctly ordered, and every declared target matches its actual FK. It also confirmed this is not an information disclosure even incrementally — every value is a compile-time constant already published in this repo's `crm.yaml`.

Both passes independently found the same two real defects, and they were right.

### 1. The feature was asserted nowhere

Deleting `References: field.References` from the mapping — the whole feature; a client receives nothing — left **both test packages green**, and the HTTP integration test decoded into a struct with no `references` field at all. My two gates proved the declarations were internally consistent and nothing proved a caller is ever told anything. A test suite that passes on a feature that does not exist.

Now covered three ways: the store reports each target for `deal`, `wireVocabularyField` is asserted for both arms (present for an id field, **nil** for a picklist, and two fields in one response not sharing a value), and the HTTP test decodes `references` as a **pointer** so an absent key is distinguishable from a blank one. Mutation-verified — the deletion now fails by name on five fields.

### 2. My "two independently written sets" claim was self-serving

`everyReferenceTarget()` was a third transcription of the constant block and the collections gate's `targets` map a fourth. The realistic drift — add a constant, point a field at it, update the map next to the failure — left the compose parity gate **passing**, because it compared a stale hand list against a stale contract. So the gate was blind to exactly the drift its own comment advertised, and my earlier mutation only proved the "list vs contract" direction.

Fixed the way the rest of this change argues for: `storekit.ReferenceTargets()` beside the constants (the shape `fieldcatalog.Types()` already uses), both tests read it, both hand lists deleted. It proved itself immediately — after the rename below it failed in **both** directions until the contract caught up.

### 3. My stated invariant was false

The doc comment said `References` is "REQUIRED of every FieldID field". Three sibling id fields in `automation/automations_preview.go` declare none, and the gate sweeps only `collections.segmentEngines`. Those preview vocabularies are never published, so the code is fine and the comment was wrong. Scoped to what actually holds, naming the gate and the set it covers.

### 4. A third spelling for one record type

The enum said `user`; this contract's existing record-type vocabulary says `app_user` and the schema is `Seat`. A client keying a picker on record type would have had to special-case it. Renamed to `app_user`, with the Go constant following.

### Also applied
- `References` is now `type Reference string`, so an unlisted target cannot be assigned — `FieldType` is a named type ten lines up, and `Field{References: "tagz"}` used to compile.
- The contract description dropped from 14 lines to 6: the argument against the client-side map is a decision record, not something an API consumer can act on. It stays in the commit message and the code comment.
- The omission comment gave the weaker reason. The real one is that `""` is not a member of the enum, so sending it would put a value the contract forbids on the wire.
- `TestNoCustomFieldTypeMapsToAnIDColumn` gates "no custom field is ever id-typed", which the commit message asserted and nothing enforced.

### One judgment call recorded, not taken
Widening the id-declares-a-target sweep to automation's preview vocabularies. Those answer a count and publish no field list, so a picker target would be dead weight; scoping the claim was the honest fix rather than declaring targets nothing reads.

### A note on the parallel round
The security pass reported uncommitted `RefInvoice` residue in four files. That was the *other* reviewer's live mutation probe, observed because both ran concurrently over the same files; it reverted its own and the tree is clean. Worth knowing for future rounds — a reviewer can see a sibling's experiment and report it as the author's.

`make check-backend`, `make craft-static` (0 findings) and `make check-fe` (3529 tests) all green; the HTTP integration test passes against real Postgres.

PR #1942, opened 2026-08-20
fix(ci): the ownership gate can see versioned writes again, and stays able to

### LINKED ISSUE #1937: The ownership gate's versioned-patch branch matches a method that does not exist, hiding 38 UPDATE call sites
`backend/tableownership_test.go` carries a branch so that
`TestEveryPackageOnlyWritesTablesItOwns` can see writes issued through
`storekit.Patch` rather than through a SQL string literal:

```go
case *ast.CallExpr:
    // storekit's versioned patch: Patch.Apply(ctx, tx, table,
    // id, ifVersion) issues the UPDATE — the table rides as
    // the third argument.
    sel, ok := node.Fun.(*ast.SelectorExpr)
    if !ok || sel.Sel.Name != "Apply" || len(node.Args) < 4 {
        return true
    }
```

**`Patch.Apply` does not exist.** `grep '^func (p \*Patch)' backend/internal/platform/database/storekit/patch.go`
gives `ApplyWithVersion`, `ApplyGuarded`, `ApplyGuardedIn` and `ApplyLocked`. The
comment names a phantom, and the branch it guards matches nothing in the tree.

## Blast radius

**38 versioned-UPDATE call sites are invisible to the ownership gate** — counted
on `main`:

| Method | Call sites |
|---|---|
| `ApplyGuarded` | 23 |
| `ApplyLocked` | 10 |
| `ApplyGuardedIn` | 4 |
| `ApplyWithVersion` | 1 |

Across 9 modules and 15 core tables: `attachment`, `contract`, `deal`, `lead`,
`lead_disqualify_reason`, `lead_source`, `offer`, `offer_template`,
`organization`, `person`, `product`, `quota`, `saved_view`, `signal`,
`webhook_subscription`.

So `p.ApplyGuarded(ctx, tx, "deal", ...)` issued from the `people` module passes
the gate silently today. That is precisely the cross-module write the gate exists
to refuse.

## It is dead, not wrong

The branch also requires `Args[2]` to be a string literal, so the handful of
unrelated `.Apply(` call sites in the tree record nothing rather than a bogus
table. **No corrupt entries exist** — nobody needs to go hunting for bad data.

Matching the four names naively still misses `ApplyLocked`, whose signature is
`ApplyLocked(ctx, tx, lock RowLock)` — **no table string at `Args[2]` at all**.
The table rides inside the lock, minted earlier by
`LockRow(ctx, tx, table, id, …)`. A four-name match skips those 10 sites while
looking like it covers them, which is the same defect one layer down.

## Suggested shape

Derive the method set from `storekit/patch.go` itself rather than hand-listing
it — an exported `*Patch` method whose first parameter is a `context.Context`
issues the UPDATE, and whether it names the table is visible from whether it has
a `table string` parameter. That census breaks when a fifth method is added,
which a hand-maintained list demonstrably does not: this one rotted silently.

`ApplyLocked` then resolves its table from the `LockRow`/`LockPair` call in the
same function body, and a call the gate cannot resolve should **fail loudly**
rather than be skipped.

## Expect it to go red

Fixing the branch makes 38 writes visible at once. Each is then either already
covered by `tableOwners` or a genuine cross-module write needing ratification.
They should be classified individually — a bulk waiver to reach green would
reproduce the hole in a different shape.

Found while analysing Wave 3. Same defect class as #1802 and #1788: a comment
asserts a guarantee the code does not hold, and the gate reads as complete. This
one is in the gate's own body.

dungbanguyen:
Closes #1937.

## The defect

`TestEveryPackageOnlyWritesTablesItOwns` carries an arm for writes issued
through storekit's versioned patch rather than a SQL string literal. It matched
the method name **`Apply`** — which no `Patch` has. The real ones are
`ApplyWithVersion`, `ApplyGuarded`, `ApplyGuardedIn` and `ApplyLocked`.

So the arm matched nothing, and **38 versioned UPDATE call sites** across 9
modules and 15 core tables passed the gate silently. `p.ApplyGuarded(ctx, tx,
"deal", …)` issued from `people` was a pass — precisely the cross-module write
the gate exists to refuse.

## The fix

- The arm names the four appliers plus **`LockRow`/`LockPair`**. `ApplyLocked`
  takes its table inside a `RowLock`'s *unexported* field, so the lock site is
  the only place that table is legible; without it every `ApplyLocked` write
  stays invisible.
- Table arguments spelled as **package constants** (`entityLead`,
  `projectObject` — the tree's normal style) resolve through the declaring
  package's own consts.

## Keeping it from going quiet again

Rule 2 from the review loop — derive the obligation, don't maintain a list:

1. **An unreadable table argument is a finding, not a skip.** A skipped write
   reads exactly like a module that writes nothing. The one site whose table
   arrives through a struct field (`people`'s two-sidecar evidence writer) is
   ratified through `gatekit.Waive`, with the reason naming both tables it can
   hold so the exception stays re-checkable against the construction sites.
2. **A floor asserts the arm still attributes writes at all.** It attributes
   **70** today; `0` is what a matcher that has stopped matching looks like,
   and that is the state this gate shipped in.

## Result

**No ownership violation surfaced** among the newly visible sites — the gate
was quiet, not wrong about the tree.

## Verification

- `make craft-static` — green
- floor verified by raising it: reports `attributed only 70 storekit table arguments`

---
## Summary by cubic
Makes the ownership gate attribute `storekit` versioned writes again and keep doing so. Previously it matched a non-existent `Apply` method and skipped 38 UPDATEs; now it matches the real appliers and fails when the table arg is unreadable.

- Matches `ApplyWithVersion`, `ApplyGuarded`, `ApplyGuardedIn`, plus `LockRow`/`LockPair` to surface `ApplyLocked` tables.
- Reads table args from string literals or package-level string constants in the declaring package; unreadable args are findings, not skips.
- Adds a targeted waiver for the `internal/modules/people` evidence writer field with its two owned tables.
- Introduces a floor (≥25) on attributed `storekit` table args to detect the matcher going quiet; currently attributes 70.
- No ownership violations surfaced among newly visible sites.

<sup>Written for commit 3e9a8cf835bf5fc798db762e6a5a18b055c5e6b6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1942?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tableownership_test.go
+++ b/backend/tableownership_test.go
@@ -8,7 +8,7 @@ package backendarch
 // against a table it does not own. This test closes that gap — it walks the
 // hand-written Go under internal/modules and internal/compose, extracts every
 // INSERT/UPDATE/DELETE target from SQL string literals (plus the storekit
-// Patch.Apply table argument), and asserts each module only writes its own
+// applier and row-lock table arguments), and asserts each module only writes its own
 // tables. Cross-store writes exist by design (merge relinks, GDPR erasure,
 // ingest materialization); each one is ratified below with a self-contained
 // rationale — an entry without a rationale is a finding, not a pass, and a
@@ -22,9 +22,11 @@ import (
 	"go/parser"
 	"go/token"
 	"io/fs"
+	"maps"
 	"os"
 	"path/filepath"
 	"regexp"
+	"slices"
 	"strconv"
 	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


tinnhlt-gra:
I filed #1937 and then built the same fix in parallel as #1943 — yours was pushed first, so **I am closing mine and deferring to this one.** Our diffs agree on nearly everything: constants resolved, the indirect `w.table` site ratified rather than skipped, and a floor so a matcher that matches nothing cannot read as a clean tree. Sorry for the duplicate; I checked `gh pr list` before branching and yours did not exist yet.

One substantive difference, offered as a review note rather than a competing patch.

### `storekitTableArg` is a hand-maintained list of method names — the exact shape that rotted

The bug being fixed here *was* a hardcoded method name that drifted from storekit. `storekitTableArg` reinstates a hardcoded list of five. Add `ApplyBatched(ctx, tx, table, …)` to `patch.go` tomorrow and this gate goes quiet about it in precisely the way it just went quiet about the other four — no test fails, nothing turns red, and the next person finds it in a wave analysis.

There is an honest derivation. The predicate is structural in storekit's own source:

- an exported method on `*Patch` whose **first parameter is a `context.Context`** issues the statement (`Set`, `Empty`, `Before`, `After` take none);
- whether it names its table is visible from whether it has a **`table string` parameter**, and its position is the argument index.

Parsing `internal/platform/database/storekit/patch.go` for that yields today:

```
map[ApplyGuarded:2 ApplyGuardedIn:2 ApplyLocked:-1 ApplyWithVersion:2]
```

`-1` being "takes a `RowLock`, resolve the table from the lock". A fifth method arrives **read**, and a derivation that matches nothing `t.Fatal`s rather than emptying the census — I verified that by pointing the receiver match at a type that does not exist:

```
derived no *Patch apply methods from internal/platform/database/storekit/patch.go;
the gate would read every versioned write as absent, which is the hole it closes
```

The implementation is ~45 lines and is on `fix/ownership-gate-sees-the-versioned-patch` if you want to lift it. Entirely your call — the list works today and is legible, and I would not block on this.

### Two smaller notes

**Recording `LockRow`/`LockPair` as writes widens what the gate means.** A row lock is a `SELECT … FOR UPDATE` — a read. A module that locks another module's row without writing it now fails an *ownership* gate. Nothing in the tree does that today, so it costs nothing now; it is worth a sentence in the comment so the next person hitting a false positive knows it was deliberate. (I resolved `ApplyLocked`'s table from the lock site within the same file instead, which keeps "write" meaning write — but yours is simpler and I am not sure simpler is worse here.)

**Worth stating the census outcome explicitly in the PR body:** with the branch working, **zero** new ownership violations appear — all 38 newly-visible writes were already covered by `tableOwners` or an existing `crossStoreWrites` entry. That is a real and reassuring finding, and readers otherwise assume a gate repair this size must have papered over something.

### One trap, in case you plant a probe

Do not use `people` → `deal`. `internal/modules/people:deal` is **already** a ratified `crossStoreWrites` entry, so that plant passes against the *fixed* gate and proves nothing — it cost me a control run that read as a working gate. `people` → `contract` is unratified and fails cleanly, naming the module, the table and the owner.

tinnhlt-gra:
Follow-up. I had a review run finish against my now-closed duplicate (#1943), and it found a **blocker in my approach that yours does not have**. Posting it because it is an argument *for* your design, and because the next person to "improve" this file could easily refactor it into my broken shape.

### Attributing at the LOCK site is not merely simpler — it is the correct one

I resolved `ApplyLocked`'s table by looking at the locks minted **in the same file**. The reviewer showed live that this hides exactly the write the gate exists to refuse:

- `deals/offer_lifecycle.go:373` spends a lock minted by `visibleOfferLocked` in **`deals/offer.go:270`** — a different file — on table `offer`.
- The only `LockRow` in `offer_lifecycle.go` is at `:250`, on `deal`, and its result is **discarded** (`_, err :=`).
- So my gate recorded a phantom `deal` write and never recorded the `offer` one.

Then, repointing `offer.go:270` to `LockRow(ctx, tx, "contract", …)` — `deals` writing the contracts-owned table through `ApplyLocked` — **both gates passed green.**

My PR body claimed file scope "can only attribute *more* tables, never fewer, so it cannot hide the cross-module write." That is false whenever the mint is in a sibling file. I withdraw the hedge in my earlier comment: your mint-site attribution is **immune to this by construction**, because the table is read where it is legible instead of being chased across files. That is worth a sentence in `storekitTableArg`'s comment so it is defended rather than incidental.

### Your vacuity floor also measures the right thing, and mine did not

Mine was `if len(writes) == 0`, which counts **all** writes — dominated by SQL literals, so it can never be zero and cannot detect blindness in the patch branch at all. It stayed green through both mutations above. Your `storekitWrites` counter measures what the `CallExpr` arm actually attributed, which is the quantity that goes to zero when the matcher breaks. Same failure mode this whole PR is fixing, and yours is the version that catches it.

### Two refinements that do apply to this PR

**If you take the derived census** from my earlier comment, derive across the whole `storekit` **package directory**, not just `patch.go`. That file is 323 lines and already holds `Patch`, `RowLock`, `LockRow` and `LockPair`; splitting it would drop the moved methods from the derived set silently. Verified on mine: removing two methods from the derived map made ~24 versioned writes invisible and both gates still reported `ok`. Whether you derive or keep the list, **pin the resulting set** so a shrink is a failure rather than a silence — that is the property neither of our versions currently has.

**`indirectTableArg` keys on `internal/modules/people:w.table`**, and the rationale reasons specifically about `organization_fact` and `organization_profile_field`. A second unreadable expression spelled `w.table` elsewhere in `people` — writing any table, including another module's — inherits that ratification without anyone deciding. Keying on `file:line`, or on file plus the resolved table set, keeps it a decision. (Mine keyed on the bare file and was worse.)

Nothing here blocks; the first two points are yours already, and I would rather they be recorded as deliberate than rediscovered.

PR #1943, opened 2026-08-20
fix(ci-tests): the ownership gate reads the versioned patches it claimed to read

### LINKED ISSUE #1937: The ownership gate's versioned-patch branch matches a method that does not exist, hiding 38 UPDATE call sites

tinnhlt-gra:
`TestEveryPackageOnlyWritesTablesItOwns` carries a branch so it can see writes
issued through `storekit.Patch` rather than through SQL text. It matched a method
called **`Apply`**. No such method exists — the real four are `ApplyWithVersion`,
`ApplyGuarded`, `ApplyGuardedIn` and `ApplyLocked`.

So the branch matched nothing, and **38 versioned `UPDATE` call sites** across 9
modules and 15 core tables sat outside the gate's universe while it reported
itself complete.

Closes #1937.

## Derived, not listed

A list is exactly what rotted here, so the method set now comes from storekit's
own source. The predicate is structural: an exported `*Patch` method whose first
parameter is a `context.Context` issues the statement (`Set`/`Empty`/`Before`/
`After` take none), and whether it names its table is visible from whether it has
a `table string` parameter. Derived today:

```
map[ApplyGuarded:2 ApplyGuardedIn:2 ApplyLocked:-1 ApplyWithVersion:2]
```

A fifth method arrives **read**, not skipped. A derivation that matches nothing
fails the gate rather than emptying it.

## The two ways a naive fix stays broken

**`ApplyLocked` has no table at `Args[2]`** — it takes a `RowLock`. A four-name
match skips those 10 sites while looking like it covers them, which is the same
defect one layer down. Its table resolves from the `LockRow`/`LockPair` in the
same **file**. File scope and not function scope on purpose: a lock is routinely
minted at the read and spent at the write 120 lines later
(`deals/offer_lifecycle.go` mints at `:250`, applies at `:373`). Widening to the
file can only attribute *more* tables to the owner, never fewer, so it cannot
hide the cross-module write the gate refuses.

**Tables named by a constant** — `entityLead`, `entityPerson`, `projectObject` —
would be skipped by a literal-only reader in exactly the same silence. They now
resolve.

**And a patch whose table still cannot be read is a failure, not a pass.**
`TestEveryVersionedPatchNamesATableTheGateCanSee` reports it; one ratified entry
covers the single site in the tree, a shared writer whose table is a struct
field. Closing the hole for calls that name their table while passing quietly
over calls that do not would reproduce it in a new shape.

## Census: zero new violations, and that is the finding

With the branch working, **no new ownership violation appeared.** All 38
newly-visible writes were already covered by `tableOwners` or an existing
`crossStoreWrites` ratification. Stating it explicitly because it is the point:
the gate was **blind, not permissive** — nobody had been exploiting the hole, and
nothing needed bulk-waiving to reach green.

## Proof — five mutations, because a gate that has never failed has not been shown to work

| | Result |
|---|---|
| **CONTROL** | On unmodified `origin/main`, `people` writing the contracts-owned `contract` table via `ApplyGuarded` **passes, exit 0** |
| **GREEN** | The same plant here **fails**, naming `people`, `contract` and `internal/modules/contracts` |
| **LOCK LEG** | A `people` lock renamed to `contract` fails **at the `ApplyLocked` site** (`promote.go:450`) — the lock-provenance path is wired, not decorative |
| **WAIVER** | Deleting the ratified entry fails; mis-keying it trips `AssertAllMatched` with *"matched no subject"* |
| **VACUITY** | A derivation matching no method **fails** rather than passing over an empty set |

### One thing worth reading, because it nearly shipped

The first plant chosen was **`deal`** — and it passed against the *fixed* gate.
Not a bug: `internal/modules/people:deal` is an **already-ratified** cross-store
write, so the probe could never have failed. It would have read as "the fix does
not work" (or, had I planted it only once, as "the control proves the hole").
Both control and green were re-run against a table nobody had ratified.

That is the same species as the defect being fixed — a check that cannot fail,
reading exactly like one that passes.

## Housekeeping

- Test-only: **no measured new-code coverage.** The five mutations above are the
- `make check` full pass · `make craft-static` PASS, 0 blocker / 0 major / 0 minor.
- **Ordering:** #1788 (W3) mirrors this same file. If both are in flight, one
  rebases on the other; they do not conflict logically.

Closes #1937

---
## Summary by cubic
Fixes the CI table-ownership gate so it actually reads versioned `UPDATE`s issued via `storekit.Patch`. Old behavior: the gate matched a non-existent `Apply` method and silently ignored 38 versioned writes; new behavior: it derives the real `*Patch` apply methods, resolves their target tables, and fails when a table cannot be determined. No new ownership violations were found; one unreadable site is explicitly waived.

**Reviewer notes**
- Derives apply methods from `storekit` source: exported `*Patch` methods whose first parameter is `context.Context`; identifies the table parameter or marks `ApplyLocked` as lock-based.
- Resolves `ApplyLocked` tables from `LockRow`/`LockPair` in the same file (file scope by design).
- Reads tables named by package string constants in addition to literals.
- Adds `TestEveryVersionedPatchNamesATableTheGateCanSee`; unreadable versioned patches now fail instead of passing silently; one waiver added in `patchTablesTheGateCannotSee`.
- Anti-vacuity: if no methods are derived, the gate fails rather than passing over an empty set.

Closes #1937.

<sup>Written for commit 429811bc429806ceb24173772fd7af84eb81c5c3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1943?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/tableownership_test.go
+++ b/backend/tableownership_test.go
@@ -7,9 +7,9 @@ package backendarch
 // ways, but nothing in the import graph stops a package from writing SQL
 // against a table it does not own. This test closes that gap — it walks the
 // hand-written Go under internal/modules and internal/compose, extracts every
-// INSERT/UPDATE/DELETE target from SQL string literals (plus the storekit
-// Patch.Apply table argument), and asserts each module only writes its own
-// tables. Cross-store writes exist by design (merge relinks, GDPR erasure,
+// INSERT/UPDATE/DELETE target from SQL string literals, adds the writes
+// storekit's versioned patch issues without spelling SQL here, and asserts each
+// module only writes its own tables. Cross-store writes exist by design (merge relinks, GDPR erasure,
 // ingest materialization); each one is ratified below with a self-contained
 // rationale — an entry without a rationale is a finding, not a pass, and a
 // waiver that matches no remaining write is stale and fails too. SELECTs are
@@ -22,9 +22,11 @@ import (
 	"go/parser"
 	"go/token"
 	"io/fs"
+	"maps"
 	"os"
 	"path/filepath"
 	"regexp"
+	"slices"
 	"strconv"
 	"strings"
 	"testing"
@@ -548,11 +550,20 @@ type tableWrite struct {
 }
 
 // collectTableWrites walks every non-test module/compose source file and
-// records each SQL write target (string literals plus storekit's
-// Patch.Apply table argument) under its owning directory.
-func collectTableWrites(t *testing.T) map[string][]tableWrite {
+// records each SQL write target under its owning directory: the tables named
+// in SQL string literals, and the tables storekit's versioned patch writes
+// without ever spelling SQL here.
+//
+// It answers a second map beside them — the patch calls whose table it could
+// NOT resolve, keyed by file. A write the gate cannot see is the failure this
+// whole branch exists to prevent, so it is reported rather than skipped;
+// TestEveryVersionedPatchNamesATableTheGateCanSee is what refuses a new one.
+func collectTableWrites(t *testing.T) (map[string][]tableWrite, map[string][]string) {
 	t.Helper()
-	writes := map[string][]tableWrite{} // owning dir → writes
+	applies := patchApplyMethods(t)
+	writes := map[string][]tableWrite{}      // owning dir → writes
+	unresolved := map[string][]string{}      // file → positions of unreadable patch calls
+	consts := map[string]map[string]string{} // package dir → its string constants
 	fset := token.NewFileSet()
 	for _, root := range []string{"internal/modules", "internal/compose", settingsStoreDir, extSecretsStoreDir} {
 		err := filepath.WalkDir(root, func(path string, d fs.DirEntry, err error) error {
@@ -566,7 +577,12 @@ func collectTableWrites(t *testing.T) map[string][]tableWrite {
 			if err != nil {
 				return err
 			}
-			owner := owningDir(filepath.ToSlash(filepath.Dir(path)))
+			dir := filepath.ToSlash(filepath.Dir(path))
+			if _, seen := consts[dir]; !seen {
+				consts[dir] = stringConstsIn(t, fset, dir)
+			}
+			locks := lockedTablesIn(file, consts[dir])
+			owner := owningDir(dir)
 			record := func(pos token.Pos, tables []string) {
 				for _, table := range tables {
 					writes[owner] = append(writes[owner], tableWrite{pos: fset.Position(pos).String(), table: table})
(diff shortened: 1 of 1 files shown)


tinnhlt-gra:
Closing as a duplicate of #1942, which was pushed two minutes before this one and closes the same issue (#1937, which I filed — another session picked it up immediately).

The two diffs agree on nearly everything. The one substantive difference — that #1942's method list is hand-maintained, which is the exact shape that rotted and produced this bug — is offered there as a review note with the derived alternative, along with the plant that gives a false green (`people` → `deal` is already a ratified cross-store write).

The branch `fix/ownership-gate-sees-the-versioned-patch` is left in place in case #1942 stalls.

Issue #1947, opened 2026-08-20
conversation_claim.changed is documented as published and nothing emits it — there is no correction path to emit from


tinnhlt-gra:
`conversation_claim.changed` is in the catalog
(`backend/internal/shared/kernel/events/catalog.go:162`) and has a payload schema
in `backend/api/public-events.yaml:964`. **Nothing emits it.**

That alone would be unremarkable — four other types are deliberately unemitted.
What makes this one different is that the contract does **not** say so. Compare:

```
Payload for deal.restored. Never emitted today (no restore path …)
Payload for person.restored. Never emitted today (no restore path …)
Payload for pipeline.archived. Never emitted today (no archive …)
Payload for mirror.write_rejected. Never emitted today — reserved …
```

against this one:

```
Payload for conversation_claim.changed — a human typed over the machine, or
dismissed a claim outright. Published because a correction is SHARED truth,
unlike a moment dismissal: a downstream reader holding …
```

It describes a **live, published** event and argues for why it is published. A
subscriber reading the contract has every reason to wait for one.

## There is no path that could emit it

`backend/internal/modules/people/conversationclaim.go` exposes
`RecordConversationClaim` and `ClaimsForPerson`, and nothing else. The sibling
type `conversation_claim.captured` **is** emitted, at `:124`. There is no
correction path, no dismissal path, and therefore nothing that could publish a
change.

So this is a capability gap wearing a contract promise, not a wiring defect.

## Two ways to resolve it, and it needs a ruling rather than a patch

1. **Build the correction path** — a human typing over or dismissing a machine
   claim — and emit the event the contract already describes. This is the option
   the contract's own text assumes.
2. **Mark it unemitted**, in the four siblings' wording, and leave the type
   reserved for whenever the correction path lands.

**Recommendation: (2) now, (1) when the correction path is actually wanted.**
Option 2 costs one description edit and makes the contract honest today; option 1
is a product feature nobody has asked for in this issue.

## Where it is recorded meanwhile

`TestEveryUnemittedEventTypeSaysWhyNothingEmitsIt` (added in the #1788 gate)
ratifies it explicitly, and its waiver says outright that it is waived so the
gate is green over a real state of the tree, **not because the state is right**.
Whichever option is chosen, that entry changes with it.

## Separately, and NOT this issue

The `capture.*` family (`capture.received`, `capture.normalized`,
`capture.failed`, `capture.skipped`) is orphaned differently: those types have no
[...]

Issue #1946, opened 2026-08-20
The approvals module hand-rolls the outbox and audit writers, drops CausationID, and its two waivers state a reason the code contradicts


tinnhlt-gra:
There are **two** outbox writers in this tree, not one:

- `internal/platform/database/storekit/storekit.go:249` — `storekit.Emit`
- `internal/modules/approvals/service.go:268` — a hand-rolled
  `INSERT INTO event_outbox (stream, envelope) VALUES ($1, $2)`

and approvals writes `audit_log` by hand too (`service.go:218`).

## The waiver that ratifies this rests on a claim the code contradicts

`backend/tableownership_test.go:448`:

> `"internal/modules/approvals:audit_log"`: *approval evidence stamps
> `passport_id`/`on_behalf_of`, **columns storekit's writer does not carry**; same
> append-only table, this module's own writer*

`storekit.Audit`'s INSERT at `storekit.go:121` is:

```sql
INSERT INTO audit_log (id, workspace_id, actor_type, actor_id, passport_id, on_behalf_of,
                       action, entity_type, entity_id, before, after, evidence, authorization_rule)
```

and it passes `UUIDOrNil(p.PassportID), UUIDOrNil(p.OnBehalfOf)`. `storekit.Emit`
sets both on the envelope actor at `:214-215`. **The stated reason for the
exception has been false for as long as the exception has existed.**

## What actually diverges — and one of them is a live defect

| | `storekit` | `approvals` |
|---|---|---|
| `Trace.CausationID` | set from `principal.CausationEvent(ctx)` (`:225-227`) | **never set** |
| `audit_log.before` / `after` | written | omitted |
| `audit_log.authorization_rule` | `auth.AuthzRule(...)` | omitted |
| clock | `time.Now()` | `s.now()` (injected — approvals' is *better*) |

**The `CausationID` omission is the one that bites.** Every `approval.*` and
`coldstart.*` envelope leaves the causation chain broken, so a bus-derived
approval cannot be traced back to the event that caused it. That is exactly what
the trace fields exist for, and nothing fails when they are absent.

## Why it matters beyond approvals

Any gate that reasons about emission by scanning `storekit.Emit*` is blind to
this writer and reports `approval.requested`, `approval.decided`,
`coldstart.accepted` and `coldstart.rejected` as never emitted. The event-type
ownership gate in #1788 works around it by walking payload struct literals
instead — which is sound, but the second writer is the reason it had to.

## Suggested direction

[...]

PR #1948, opened 2026-08-20
feat(platform): one module owns an event type, and a gate says so

### LINKED ISSUE #1788: Two modules can emit the same event type and no gate notices

tinnhlt-gra:
The tables have a gate refusing a module that writes another module's rows. The
**event types had none**, so two modules could publish under one type and nothing
noticed — which is how a subscriber ends up receiving one name for two different
facts, with no way to tell them apart, because **the type IS the discriminator**.

Closes #1788.

## This ships RED, and the findings are correct behaviour

**Eleven types already have two or more emitting modules, and every one is
deliberate.** The deliverable is a gate plus ratified waivers, not eleven fixes.
Written expecting green, this would have deleted working design.

The census, run before the waiver set was populated — **77 event types examined**:

| Event type | Emitting modules |
|---|---|
| `person.updated`, `person.archived`, `organization.updated`, `organization.archived`, `deal.archived`, `lead.updated`, `activity.updated` | native module + `overlay` |
| `deal.updated` | `deals` + `overlay` + `people` |
| `project.updated` | `deals` + `people` |
| `lead.created` | `people` + `capture` |
| `activity.captured` | `activities` + `capture` |

**Structure 1 — the overlay write-back announces the NATIVE module's event.** A
subscriber to `person.updated` is subscribed to *a person changing*, and must
hear about one whether the write arrived natively or through the write-back. An
`overlay.*` type would make every consumer subscribe twice to learn one fact, and
would leak which path a write took into a contract that deliberately does not say.

**Structure 2 — a relationship emits its ANCHOR's event.** An employment edge
changing *is* a change to the person and to the company it joins. There is no
`relationship.*` type, and inventing one would make every consumer of the anchor
subscribe to a second name to learn their own record moved.

**Structure 3 — capture publishes the RECORD's type**, with `source_system` set,
because the capture path creates real leads and real activities. A `capture.*`
type would announce that a pipeline ran, which no consumer of the record wants.

### The goal's cross-check

The wave analysis predicted **12** and noted a literal-walk census would find
**11**, the twelfth (`coldstart.read_back_proposed`) being invisible because it is
the dynamic `Announce` case. That is exactly what happened, which is a useful
signal the walk is behaving rather than a discrepancy.

## The waivers are keyed on the PAIR, and that was a hole before it was a decision

`"<event type> <- <module>"`. Keyed on the type alone, ratifying a sharing once
admits **any number of further modules** under the same entry.

That is not hypothetical — I built it type-keyed first. Planting an emit of
`person.updated` inside `deals` **passed silently**: the exact failure this gate
exists to refuse, reproduced inside its own exception list. Re-keyed, the same
plant fails:

```
event type "person.updated" is emitted by internal/modules/deals, and by 3 modules in all
(deals, overlay, people) — … ratify this emitter in
sharedEventTypes["person.updated <- internal/modules/deals"] …
```

## Two traps this had to route around

**The issue's core assumption is false.** It says "every emit goes through
`storekit.Emit`". There are **two** outbox writers — storekit's, and a hand-rolled
`INSERT INTO event_outbox` at `approvals/service.go:268`. A walk keyed on
`storekit.Emit*` would miss the whole approval family and report four live types
as never emitted. This walks **payload struct literals** instead, which is sound
over both writers and over the six helpers that return one of several payloads by
branch. Filed the approvals writer as **#1946** — including that both waivers
ratifying it state a reason the code contradicts (`storekit.Audit` demonstrably
*does* carry `passport_id`/`on_behalf_of`), and that its envelope drops
`CausationID`, breaking the trace chain on every `approval.*` event.

**`PublicEvent*` is not a reliable name test.** The generated file declares **94**
such structs; only **83** carry `EventType()`. The rest are nested field structs,
and a name-prefix census reports `PublicEventActivityChangedFields` as a twelfth
shared type that does not exist. Filtering is on the **method**.

## A second gate, the other direction

Six payload types have no emit site anywhere. Five are documented in the contract
as deliberately unemitted; **`conversation_claim.changed` is not** — the contract
describes it as *published* ("a correction is SHARED truth") and there is no
correction path in `people/conversationclaim.go` to publish from. Filed as
**#1947**. Its waiver says outright that it is waived so the gate is green over a
real state of the tree, **not because the state is right**.

That gate is also this one's **floor**. The vacuity check only catches a walk that
finds *nothing*; a walk that quietly found half would still report every remaining
type as singly-owned and pass. Every type the walk stops seeing lands in the
orphan set instead.

## Proof — five mutations

| | Result |
|---|---|
| **FOREIGN EMIT** | `person.updated` emitted from `deals` **fails**, naming the module and the key to add |
| **WAIVER** | Deleting a ratified pair **fails**; mis-keying one trips `AssertAllMatched` — *"matched no subject"* |
| **VACUITY** | Deriving no event types **fails** rather than passing over an empty tree |
| **COLLAPSE** | Making the walk blind to **one** payload struct **fails** — the check the vacuity guard alone does not catch |
| **CENSUS** | Reported as a number the gate logs: *"examined 77 event types"* |

## Two legs this does not close — stated, not faked

**The cache-key-prefix leg is vacuous.** ADR-0014's rule covers tables, event
types *and* cache key prefixes. **No module owns a prefix**, because no package
under `internal/modules/` imports a Redis client at all. The two files that grep
for "redis" are an error-message comment and the word *"rediscovered"*. All four
key namespaces are platform-tier constants (`ovb:`, `msr:`, `gw:dedupe:`,
`gw:events:crm:`). So #1788's fix covers one leg of two; the honest guard for the
other is a different invariant, and it is not this one.

**Three emit sites are irreducibly dynamic** — the extension ledger's
`namespace + "." + verb` (structurally confined to `ext_*`, disjoint from the
catalog) and the approvals `Announce` seam's `events.Payload` interface field. The
gate cannot attribute those. That is its limit, named here rather than papered
over.

## Housekeeping

- Test-only: **no measured new-code coverage.** The five mutations are the
- No ADR is cited as though a reader could open it — the rule is written out in
  the gate's doc comment instead.
- `make check` full pass · `make craft-static` PASS, 0 blocker / 0 major / 0 minor.

Closes #1788

---
## Summary by cubic
Enforces one-module ownership of each event type and adds a second gate requiring reasons for payloads with no emit site. Previously, duplicate emitters and orphaned payloads were silent; now the gate fails unless the "<type> <- <module>" pair is waived, and it fails if a payload has no emitter without a documented reason.

- Detects emitters by walking payload composite literals keyed on `EventType()` methods, not call sites; now scans `internal/`, `cmd/`, `pkg/`, and `../extensions`, resolves the `internal/contracts` import qualifier per file, handles elided slice/map elements (including pointer element forms) and `new(T)` constructions, ignores decode-only `var T` declarations, and reports file:line.
- Waivers are pair-keyed (`"<type> <- <module>"`) to prevent type-level blanket admits; ratified shares cover overlay write-backs announcing the native event, relationship edges emitting the anchor’s event, and capture publishing the record’s own type.
- Floor and cross-check: derived payload types are set-compared with `crmcontracts.PublicEventVersions` to fail on partial generator shape changes; avoids vacuous passes.
- Limits and follow-ups: dynamic emits (extensions namespace+verb; approvals `Announce`) are unattributed; cache-key ownership is out of scope. Filed #1946 (approvals writer/envelope) and #1947 (`conversation_claim.changed` missing emit).

- Required actions:
  - If you add a second emitter, add `"<type> <- <module>"` to `sharedEventTypes` with a rationale.
  - If a payload should remain unemitted, add its type to `unemittedEventTypes` with the reason.

<sup>Written for commit 03cc33611005929bd333d897b4dba9cfdc1b09fb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1948?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/eventtypeownership_test.go
+++ b/backend/eventtypeownership_test.go
@@ -120,11 +120,11 @@ func collectEmitSites(t *testing.T) map[string][]emitSite {
 	types := payloadEventTypes(t)
 	sites := map[string][]emitSite{} // event type → where it is built
 	fset := token.NewFileSet()
-	// Every hand-written tree that could hold a payload literal, not just the
-	// modules. The rule is that exactly one module decides what a type means, and
-	// a literal added under platform or an extension defeats that as thoroughly
-	// as a second module does — while being invisible to a modules-only walk.
-	for _, root := range []string{"internal", "../extensions"} {
+	// Every hand-written tree that could hold a payload literal, and the list is
+	// the whole list rather than the obvious half: cmd and pkg can import
+	// internal/contracts as legally as internal can, so an emitter added under
+	// cmd/api defeats the rule while being invisible to a modules-only walk.
+	for _, root := range []string{"internal", "cmd", "pkg", "../extensions"} {
 		err := filepath.WalkDir(root, func(path string, d fs.DirEntry, err error) error {
 			if err != nil || d.IsDir() || !strings.HasSuffix(path, ".go") ||
 				strings.HasSuffix(path, "_test.go") || strings.HasSuffix(path, "_gen.go") ||
@@ -136,13 +136,21 @@ func collectEmitSites(t *testing.T) map[string][]emitSite {
 			if err != nil {
 				return err
 			}
+			// The qualifier the contracts package is reachable under IN THIS FILE.
+			// Matching only a selector's terminal name would count an unrelated
+			// package's PublicEventPersonUpdated as the CRM payload, and a file
+			// that does not import contracts at all can hold no emit.
+			qualifier, imports := contractsQualifier(file)
+			if !imports {
+				return nil
+			}
 			module := owningDir(filepath.ToSlash(filepath.Dir(path)))
 			ast.Inspect(file, func(n ast.Node) bool {
 				// A SelectorExpr covers `pkg.Type{…}` and, via the type of a
 				// slice or map literal, the ELIDED inner `{…}` elements whose own
 				// Type node is nil. Without the second, a module batch-building
 				// payloads from a slice literal is invisible here.
-				name, ok := payloadTypeName(n)
+				name, ok := payloadTypeName(n, qualifier)
 				if !ok {
 					return true
 				}
@@ -180,9 +188,9 @@ func modulesEmitting(sites []emitSite) []string {
 // silently, which is the failure this whole gate exists to refuse, reproduced
 // inside its own exception list.
 //
-// Two structures account for every entry, and in both a second module announces
-// a fact that IS the first module's fact. Neither is a second meaning for one
-// name, which is what the rule protects.
+// Three structures account for every entry, and in all of them the second
+// module announces a fact that IS the first module's fact. None is a second
+// meaning for one name, which is what the rule protects.
 var sharedEventTypes = gatekit.Waive(map[string]string{
 	// Structure 1 — the overlay write-back announces the NATIVE module's event.
 	// overlay/writeaudit.go switches on datasource.EntityRef and emits the
@@ -329,6 +337,11 @@ func positionsFor(sites []emitSite, module string) []string {
 // the name lives on the enclosing slice or map literal, so a module
 // batch-building payloads that way would emit without this gate seeing it.
 //
+// What this still cannot see, named so the comment does not overclaim a second
+// time: a payload built by ASSIGNMENT into a zero value, or through reflection
+// or a generic builder. Those need dataflow, nothing in the tree does them, and
+// a walk that guessed at them would misread the decode targets below.
+//
 // A bare `var p crmcontracts.PublicEventX` declaration is deliberately NOT one
 // of them, and the tree says why: both such declarations here are DECODE
 // targets, immediately json.Unmarshal-ed from an inbound envelope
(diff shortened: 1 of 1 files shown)


tinnhlt-gra:
## Review response — 5 fixed, 1 pushed back with evidence

### 1. The floor I claimed was false for half the derivation — fixed, and this was the real find

I argued the orphan gate was this gate's floor: *"a walk that quietly found half … every type it stops seeing lands here instead."* That holds only for a collapse in `collectEmitSites`. **Both** gates derive their universe from `payloadEventTypes`, whose only guard was `len(out) == 0` — so a type lost *there* vanished from the emit census **and** the orphan sweep at the same instant, and nothing was left to notice.

Every realistic trigger is a generator shape change this gate is meant to survive: a pointer receiver, a two-statement body, `return string(someConst)` instead of a literal.

`PublicEventVersions` is generated from the same contract and keyed on the event type — the right thing to compare against. Dropping one payload now:

```
derived 82 event types from …/publicevents_gen.go but PublicEventVersions carries 83;
[activity.archived] have a contract entry and no EventType() this walk could read.
The walk has stopped seeing payloads, so both gates would pass over every type it lost
```

Before the fix, that same mutation left **both gates green** at `examined 76`.

### 2. Roots excluded `internal/platform` and `extensions/` — fixed

The stated rule is that exactly one module decides what a type means, and a literal added under platform defeats that as thoroughly as a second module does. Now walks `internal` and `../extensions`.

### 3. Elided composite literals were invisible — fixed

`[]crmcontracts.PublicEventX{{…}}` — the inner element's `Type` is nil and the name lives on the enclosing slice/map literal.

### 4. `emitSite.pos` written and never read — fixed

Findings now name the file and line:

```
event type "person.updated" is built at internal/modules/deals/claim.go:57:9, and by 3 modules in all …
```

### 5. Both sweeps deferred above a fatal — fixed

Armed after. You were right that it buries the one true line under an entry per waiver, each telling the reader to delete a correct one.

### 6. The failure accused the rightful owner — reworded

The gate genuinely has no notion of an owner, and I did not want to invent a first-emitter heuristic that would be wrong the first time a file is renamed. The message now says which it is: *"If X is not that module, move the emit into the one that is; if it IS, the new emitter is one of the others."*

---

### Pushed back: counting `var p crmcontracts.PublicEventX`

I implemented this, and **it reported two consumers as emitters**:

```
lead.sla_breached is emitted by 2 modules (internal/compose, internal/modules/people)
person.merged     is emitted by 2 modules (internal/compose, internal/modules/people)
```

Both `internal/compose` sites are **decode targets**, immediately `json.Unmarshal`-ed from an inbound envelope — `compose/leadsla.go:69` reading `lead.sla_breached`, `compose/personautoenrich.go:123` reading `person.merged`. Ratifying them would have had this gate certify two sharings that do not exist, and would have put two false entries in the waiver map permanently.

A declaration cannot be told from a construction without dataflow. Here the false positives are **real and present**; the construction case (`var p X` then field assignment) appears nowhere in the tree. Reverted, with the reason recorded in the source so the next reader does not re-add it.

The suggestion was still worth making — chasing it is what surfaced finding 3's elided-literal case as a genuine hole rather than a theoretical one.

---

`make check` full pass · `make craft-static` PASS on all four trees. Census unchanged at 77 examined, 11 shared types, 23 ratified pairs, 6 orphans.

tinnhlt-gra:
## Codex review — four findings, all fixed; the pair keying and the floor confirmed sound

This PR has now had **Fable, CodeRabbit and Codex**.

### 1. My own comment asserted a guarantee the code did not hold — fixed

The comment read *"Every hand-written tree that could hold a payload literal, not just the modules."* The roots were `internal` and `../extensions`. **`cmd` and `pkg` import `internal/contracts` as legally as `internal` does**, so an emitter added under `cmd/api` defeated the rule while being invisible to this gate.

A comment claiming a guarantee the code does not hold — in the file whose entire job is refusing that shape. Roots are now `internal`, `cmd`, `pkg`, `../extensions`, and the comment names the list rather than gesturing at it.

### 2. The walk matched a terminal name without checking the package — fixed

`sel.Sel.Name` alone would count an unrelated package's `PublicEventPersonUpdated` as the CRM payload. It now reads the qualifier `internal/contracts` is reachable under from each file's own imports, and skips a file that does not import it at all.

### 3. Two construction forms were invisible — fixed

- `new(crmcontracts.PublicEventX)` — yields the same `events.Payload` that `&T{}` does.
- The elided element of a slice or map of **pointers**: `[]*crmcontracts.PublicEventX{{…}}`, where the enclosing type is a `StarExpr` around the selector rather than the selector itself. This is the same family as the elided-value case from the previous round; I had unwrapped one indirection and not the other.

### 4. "Two structures" — fixed

The waiver block has carried **three** since capture was added as its own case.

### What remains unreadable, now named in the file

A payload built by **assignment into a zero value**, or through reflection or a generic builder. Those need dataflow; nothing in the tree does them; and a walk that guessed would misread the two documented decode targets. Named rather than left for the next reviewer to find — which is the mistake finding 1 was.

### Confirmed sound by Codex

- **The pair keying is tight.** A third module produces a genuinely absent key and reaches the error; `AssertAllMatched` catches obsolete pairs. This is the hole the previous round found and closed, independently re-checked.
- **The generated-map floor holds** across every derivation-loss path it traced.
- **The `ValueSpec` revert is correct and complete** — both known consumers (`compose/leadsla.go:69`, `compose/personautoenrich.go:123`) remain uncounted.
- **No factual error in any of the eleven rationales**, checked against the construction sites.

`make check` full pass · `make craft-static` PASS on all four trees. Census unchanged: 77 examined, 11 shared, 23 ratified pairs, 6 orphans.

PR #1951, opened 2026-08-20
feat(ci-tests): a module that owns tables and audits none of them fails the gate

### LINKED ISSUE #1802: The write-shape gate catches audit-without-emit but not never-audits-at-all, so a module can own tables and audit none of them

tinnhlt-gra:
`TestEveryAuditedMutationEmitsAnEvent` holds **half** of the write-shape
obligation — audit implies event — and it only starts counting once it has seen a
call named `Audit`:

```go
case "Audit", "AuditWithEvidence":   // <- the entry condition
```

A module that calls **neither** has no audit row to pair, so it is not merely
unwaived — it is **outside that gate's universe**. It could own five tables,
write every one of them, and record nothing, and both halves of the write shape
would report green over it.

That was true of the finance mirror until #1795: six write statements on four
tables, no `audit_log` row anywhere, and a comment in the module claiming the
write shape was *"the house one"*.

Closes #1802.

## Subjects come from `tableOwners`, not `doc.go` — and that matters

#1802 proposed deriving ownership from each module's `doc.go`, *"already read by
`tableownership_test.go`"*. **Nothing parses `doc.go`.** `tableOwners` is a
hand-maintained map, and its neighbours' "kept in sync" is prose.

A `doc.go`-derived census would exempt **seven** modules outright: three have no
`doc.go` at all, two have one with no `Tables owned` line, and two declare
`Tables owned: none` while `tableOwners` assigns them tables — including the
module that held the original defect.

## The goal's predicted census was wrong, and this reports it rather than trimming to fit

The wave analysis predicted **five** members. The gate finds **eleven of
thirty-three modules**. Four were predicted; finance is absent because #1795
fixed it; **seven were not named at all** — five compose subpackages and two
platform packages that `tableOwners` assigns tables to directly, which a
module-root granularity argument does not remove because the map's *values* are
already those paths.

Every one classified against the **code**, not against a comment sitting next to
it:

| Module | Verdict |
|---|---|
| `platform/database/storekit` | **Is** the audit writer — it owns `audit_log`, `event_outbox`, `system_log`, `field_provenance`. Auditing its own writes is circular. |
| `compose/org360`, `person360`, `orgbrief`, `personbrief`, `orgdossier` | Per-**reader** derived state. Verified in the DDL: every table carries a `user_id`, and `0199` says outright that an assembly generated for one reader is never served to another. An audit trail over these records *reading*, not changing. |
| `modules/agents` | Runner bookkeeping. The domain writes a run performs happen inside the **tools** it calls, which carry the full shape — auditing the run row too files one change twice under two entity types. |
| `modules/comms` | Delivery machinery, not the message. The user-visible fact is the **activity** row, which activities owns and audits, and `StageTx` runs inside that transaction. |
| `modules/search` | Projections, thrown away and rebuilt as the corruption remedy. And `embed_store_binding` is stronger than judgement: it is a **singleton with no `workspace_id` column**, and `storekit.Audit` derives its workspace from the GUC — it could not write a valid row for that table if it tried. |
| `platform/extsecrets` | Writes `system_log` on purpose, and says so in-source: a secret changing hands moves no domain row, so there is no `audit_log` entry to attach it to. |
| `modules/approvals` | **Not a waiver of the obligation.** It writes `audit_log` by **hand** (`service.go:218`), bypassing storekit — so the module has history and this gate cannot see it. Filed as **#1946**, including that its writer omits `before`, `after` and `authorization_rule`, its envelope drops `CausationID`, and both existing waivers ratifying it state a reason the code contradicts. |

## `LogSystem` deliberately does not count — and the reason is stated

It writes `system_log`, the ledger for an operation that mutates no record. The
obligation here is **record** history. The existing gate already excludes it, and
the two have to agree about what "audits" means or the same code satisfies one
and not the other. `extsecrets` appears in the census *because* of this, with a
waiver saying it is recorded — in the ledger that fits it.

## Proof — the mutation is mandatory here

#1795 landed first, so this gate is **green on arrival**, and green on arrival is
exactly how a guard ships broken.

| | Result |
|---|---|
| **RED** | Every `storekit.Audit` stripped from finance → the gate **fails naming `internal/modules/finance`** |
| **CONTROL** | Same tree, only `origin/main`'s gates, with the now-stale waivers removed as a developer deleting those calls would remove them: `make check-backend` **EXIT=0** — a complete pass over a module that audits nothing |
| **INVERSE** | Deleting the `search` waiver **fails**; mis-keying `comms` trips `AssertAllMatched` — *"matched no subject"* |
| **CENSUS** | Logged as a number: *"examined 33 modules that own at least one table"* |

The `AssertAllMatched` defer is armed **after** the vacuity fatal, deliberately.
Deferred above it, the sweep runs on the way out and buries the one true line
under an entry per waiver, each telling the reader to delete a correct one.

## Housekeeping

- Test-only: **no measured new-code coverage.** The mutations above are the
- No ADR cited as though a reader could open it — the rule is written out in the
  gate's doc comment.
- `make check` full pass · `make craft-static` PASS on all four trees.

Closes #1802

---
## Summary by cubic
Fails CI when a module in `tableOwners` owns tables but never writes an audit row. Previously, modules with no `storekit.Audit`/`AuditWithEvidence` calls sat outside the audit→event gate and could pass without history; `storekit.LogSystem` remains excluded.

- Subjects come from `tableOwners`; logs the census size and prevents a vacuous pass.
- AST scan requires a real call to `storekit.Audit`/`AuditWithEvidence`, resolves the `storekit` import alias, and skips test/integration files and subdirectories that are themselves subjects so nested packages do not mask parents.
- Error handling: surface walk errors without `FailNow`; run `AssertAllMatched` only after a non-empty census to avoid noisy stale-waiver output.
- Waivers with rationales: audit writer `internal/platform/database/storekit`; per-reader compose subpackages; operational `internal/modules/agents` and `internal/modules/comms`; projections in `internal/modules/search` (bare-pool write leaves the workspace GUC unset); `internal/platform/extsecrets`. Updates: `internal/modules/approvals` now states `approval` audits by hand (invisible here) while `signing_key` has no history.
- On failure: wire `storekit.Audit` into mutations or add a waiver that states where the module’s history lives. Test-only; CI may now fail when new owners are added without audit writes.

Closes #1802.

<sup>Written for commit b2cd657dd90754f503bb98196839bbbc826e1899. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1951?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/moduleaudits_test.go
+++ b/backend/moduleaudits_test.go
@@ -75,7 +75,7 @@ var modulesThatWriteNoHistory = gatekit.Waive(map[string]string{
 	"internal/platform/extsecrets": "extension_secret is written with storekit.LogSystem rather than storekit.Audit, and the package says why in-source: a secret changing hands moves no domain row, so there is no audit_log entry to attach it to. It belongs in system_log, the non-entity operational ledger, which is the same posture the boot's extension inventory takes. This gate deliberately does not count LogSystem, so the module appears here — it is recorded, in the ledger that fits it",
 
 	// NOT a waiver of the obligation — a different defect, filed.
-	"internal/modules/approvals": "TRUE OF ONE OF ITS TWO TABLES, and the entry says so rather than rounding up. `approval` has history: approvals writes audit_log by HAND at service.go:214, bypassing storekit.Audit, so this gate cannot see it — filed as #1946 with what that writer omits. `signing_key` has NONE: the INSERT at token_jws.go:172 mints an Ed25519 private key with no audit row, no hand-rolled row and no system_log row, and the hand-rolled writer could not describe it anyway because it hardcodes entity_type to the literal 'approval'. That is a real gap this waiver does not excuse; it is recorded here so the next reader finds it instead of trusting the module-granular verdict. Which brings out this gate's own limit: it is module-granular, so `owns five tables, audits one` passes it, and approvals is the live instance",
+	"internal/modules/approvals": "TRUE OF ONE OF ITS TWO TABLES, and the entry says so rather than rounding up. `approval` has history: approvals writes audit_log by HAND at service.go:218, bypassing storekit.Audit, so this gate cannot see it — filed as #1946 with what that writer omits. `signing_key` has NONE: the INSERT at token_jws.go:172 mints an Ed25519 private key with no audit row, no hand-rolled row and no system_log row, and the hand-rolled writer could not describe it anyway because it hardcodes entity_type to the literal 'approval'. That is a real gap this waiver does not excuse; it is recorded here so the next reader finds it instead of trusting the module-granular verdict. Which brings out this gate's own limit: it is module-granular, so `owns five tables, audits one` passes it, and approvals is the live instance",
 })
 
 // auditWriters are the storekit calls that put a row in audit_log.
@@ -133,12 +133,26 @@ func moduleWritesAuditRow(module string, subjects map[string]bool) (bool, error)
 		if err != nil {
 			return err
 		}
+		qualifier, imports := storekitQualifier(file)
+		if !imports {
+			return nil
+		}
 		ast.Inspect(file, func(n ast.Node) bool {
-			sel, ok := n.(*ast.SelectorExpr)
+			// A CALL, not a mention. Matching the selector alone would let
+			// `var auditWriter = storekit.Audit` — a reference that never runs —
+			// answer for a module that audits nothing, which is the false green
+			// this gate exists to refuse. And the qualifier has to resolve to the
+			// IMPORTED package: a local value named storekit with an Audit method
+			// would otherwise do the same.
+			call, ok := n.(*ast.CallExpr)
 			if !ok {
 				return true
 			}
-			if pkg, ok := sel.X.(*ast.Ident); ok && pkg.Name == "storekit" && auditWriters[sel.Sel.Name] {
+			sel, ok := call.Fun.(*ast.SelectorExpr)
+			if !ok {
+				return true
+			}
+			if pkg, ok := sel.X.(*ast.Ident); ok && pkg.Name == qualifier && auditWriters[sel.Sel.Name] {
 				found = true
 			}
 			return !found
@@ -188,4 +202,21 @@ func TestEveryTableOwningModuleWritesAnAuditRow(t *testing.T) {
 			"modulesThatWriteNoHistory with a rationale saying where its history lives instead",
 			module)
 	}
+}
+
+// storekitQualifier answers the name storekit is reachable under in one file,
+// and whether the file imports it at all. A file that does not import it can
+// hold no audit call, whatever it happens to name a local variable.
+func storekitQualifier(file *ast.File) (string, bool) {
+	const storekitPath = `"github.com/gradionhq/margince/backend/internal/platform/database/storekit"`
+	for _, imp := range file.Imports {
+		if imp.Path.Value != storekitPath {
+			continue
+		}
+		if imp.Name != nil {
+			return imp.Name.Name, true
+		}
+		return "storekit", true
+	}
+	return "", false
 }
--- a/backend/moduleaudits_test.go
+++ b/backend/moduleaudits_test.go
@@ -1,0 +1,168 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+package backendarch
+
+// A module that owns tables writes their history.
+//
+// TestEveryAuditedMutationEmitsAnEvent holds half of the write-shape
+// obligation — audit implies event — and it only starts counting once it has
+// seen a call named Audit. A module that calls NEITHER has no audit row to pair,
+// so it is not merely unwaived: it is outside that gate's universe entirely. It
+// could own five tables, write every one of them, and record nothing, and both
+// halves of the write shape would report green over it.
+//
+// That is not a hypothetical. It was true of the finance mirror until #1795: six
+// write statements on four tables, no audit_log row anywhere, and a comment in
+// the module saying the write shape was "the house one".
+//
+// The rule, stated rather than cited: every mutation commits its domain row and
+// its audit_log row in ONE transaction. An audit row cannot be written after the
+// fact, so a module that never writes one has no history that can be recovered —
+// and the erasure and retention reasoning that reads audit_log is blind to
+// everything it owns.
+
+import (
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"io/fs"
+	"maps"
+	"path/filepath"
+	"slices"
+	"strings"
+	"testing"
+
+	"github.com/gradionhq/margince/backend/internal/shared/gatekit"
+)
+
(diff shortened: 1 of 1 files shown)


tinnhlt-gra:
## Review response — all three fixed, one of them a real gap in the tree

### 1. A nested subject was answering for its parent — fixed

`filepath.WalkDir(module, …)` descends, and six census subjects sit inside another one. Measured directly rather than argued: a naive walk of `internal/compose` returns

```
internal/compose/briefs/briefstore.go    <- census subject internal/compose/briefs
internal/compose/briefs/briefstore.go    <- census subject internal/compose/briefs
internal/compose/briefs/briefstore.go    <- census subject internal/compose/briefs
internal/compose/datareset.go
internal/compose/export.go
internal/compose/extledger.go
```

So `compose` could own `signal_thread_scan`, the reset path's tables and four more, audit none of them, and read green because **a different subject** audits. The walk now skips any subdirectory that is itself a subject. You were right that the sibling gate already disagreed — `owningDir` normalizes only `internal/modules/*`.

### 2. The approvals waiver was true of one of its two tables — and the other half is a real finding

Verified, and it is worse than "the gate cannot see it":

- `approval` **has** history — hand-written at `service.go:214`, invisible here (#1946).
- `signing_key` has **none**. `token_jws.go:172` mints an **Ed25519 private key** and inserts it with no `storekit.Audit`, no hand-rolled row, no `storekit.LogSystem`. A grep of `storekit.` across approvals' non-test files returns only paging and lock helpers.
- And the hand-rolled writer **could not** describe it if it fired: it hardcodes `entity_type` to the literal `'approval'` (`VALUES (…, 'approval', …)`).

The waiver now says which half is which instead of rounding up, and names the limit this exposes: **the gate is module-granular**, so *"owns five tables, audits one"* passes it, and approvals is the live instance.

**I have not filed this publicly.** `SECURITY.md` puts *"a mutation that skips the audit or outbox row"* in a private advisory rather than a public issue, and this is a signing key. Flagged to the repo owner to route — I would rather over-apply that rule than under-apply it. (#1946 as filed would **not** surface it: routing the existing writer through storekit leaves `signing_key` untouched.)

### 3. `t.Fatalf` inside the walk fired the armed sweep — fixed

You caught the comment claiming to prevent exactly what the code below it did: `FailNow` → `Goexit` → deferred sweep runs, so a module renamed without updating `tableOwners` printed one true line and then ten *"delete this correct waiver"* errors. It returns the error now; the caller reports it without failing the test out.

### On the `search` waiver

You were right that the stated reason was wrong even though the conclusion holds, and I had claimed to fix the wording before I actually had — corrected now. It reads: the audit row takes its workspace from the **GUC**, not from the audited table, so a missing column would not by itself stop one. What stops it is that `binding.go` writes on the **bare pool**, deliberately outside any per-workspace transaction (marked `rls-exempt` in-source, deployment metadata), so the GUC is unset and `audit_log`'s `NOT NULL workspace_id` would take a NULL.

---

`make check` full pass · `make craft-static` PASS on all four trees. Census unchanged: 33 modules examined, 11 waived.

tinnhlt-gra:
## Codex review — three findings, two fixed, one refuted with the reason

This PR has now had **Fable, CodeRabbit and Codex**. Codex found a false green the other two missed.

### 1. The walk matched a *mention* of the audit writer, not a *call* — fixed, and this was the real one

`moduleWritesAuditRow` matched a bare `SelectorExpr`. So `storekit.Audit` appearing anywhere in a module satisfied a gate about whether the module **calls** an audit writer — including as a value that is never invoked. It also never resolved the qualifier, so a local named `storekit` with an `Audit` method would have done the same.

Proven both directions, with `comms`' waiver removed so the module must actually be judged, and `var plantedAuditRef = storekit.Audit` planted in it:

```
with the fix (CallExpr required):   FAIL — internal/modules/comms owns tables and calls no audit writer anywhere
pre-fix (selector-only match):      ok   github.com/gradionhq/margince/backend
```

That is the false green, reproduced and closed. It now requires a `CallExpr` whose `Fun` is the selector, and reads the name `storekit` is reachable under from the file's own imports.

### 2. The approvals rationale pointed one statement short — fixed

It said `service.go:214`. Line 214 is `return ids.Nil, err`; the `INSERT INTO audit_log` begins at **:218**. A locator that sends a reviewer to error handling is worse than none, and `gatekit` checks neither the pointer nor the reason — only that the key still matches a subject.

### 3. The waiver key is module-granular — **refuted as a fixable defect, kept as a stated limit**

Correct that a module gaining a *second* unaudited table is admitted by the entry that ratified the first. But table-granularity is **not available to this gate**: `storekit.Audit` names an **entity type**, not a table, and the two are not a reliable mapping (`person` the entity vs `person` the table happens to coincide; `activity` audited against `attachment` does not).

So the choice is a module-granular gate that says so, or no gate. The file says so, in the waiver text itself — and the one live instance, approvals' unaudited `signing_key` write, is recorded there and tracked out-of-band rather than left to a check that structurally cannot see it.

Codex also **confirmed all three prior fixes** (nested subjects, the approvals half-truth, the `t.Fatalf` inside the walk) and found no defect in hunt targets 4–6, having checked all eleven rationales against their tables, migrations and write sites.

`make check` full pass · `make craft-static` PASS on all four trees.

Issue #1925, opened 2026-08-20
core/0084's NOT VALID comment claims a lock benefit the single-transaction migration runner cannot deliver


tinnhlt-gra:
`backend/migrations/core/0084_profile_field_display_name.up.sql:8-10` says:

> The CHECK is added NOT VALID and validated separately: the immediate form scans the table under a write-blocking lock, while VALIDATE takes only SHARE UPDATE EXCLUSIVE — existing rows all satisfy the widened set anyway.

The first half is true of Postgres in general and false of this runner. `dbmigrate.Up` executes a whole migration file inside ONE transaction (`backend/internal/platform/dbmigrate/dbmigrate.go:142`), so the ACCESS EXCLUSIVE lock taken by `ALTER TABLE … DROP CONSTRAINT` is held until commit and the `VALIDATE` scan runs entirely underneath it. `NOT VALID` + `VALIDATE` is therefore behaviourally identical here to a plain `ADD CONSTRAINT`, just longer.

Nothing is broken — 0084 is applied everywhere and the constraint is correct. The defect is that the comment reads as a **pattern to copy**, and it is: it was copied. While widening the same `attachment.category` CHECK in #TBD the same three lines were written again, on the same false premise, and only review caught it. The next person widening a vocabulary will find 0084 by grep and inherit it a third time.

`0084` has shipped, so it must not be edited (additive migrations only). What would close this:

- state the honest version of the rule somewhere a person looking for it will land — the runner's own doc comment in `dbmigrate.go` is the natural home, since it is the fact that makes the difference; and/or
- a gate: a migration that spells `NOT VALID` followed by `VALIDATE CONSTRAINT` in the same file is doing nothing under this runner, and that is mechanically detectable in the same shape as `TestEveryMigrationTakingABlockingLockBoundsHowLongItWaits`.

Found while widening the attachment category vocabulary; the new migration there was rewritten to the immediate form with the reasoning stated.

PR #1955, opened 2026-08-20
feat(records): a channel file is not an email attachment

tinnhlt-gra:
**A file that arrives on a messaging channel is not an email attachment.** The captured-file writer
stamped `attachment.category` with a hardcoded `'email_attachment'` literal, and the vocabulary
admitted only mail — so the first channel file to land would have been recorded durably mislabeled, in
the row, in the audit image, and in every category filter and document-library query. This is PR 2 of
the channel-attachments arc (PR 1 was #1853), and it stands on its own merits: a mail-labelled channel
file is wrong independently of the feature that will produce one.

**It prevents a mislabel; it does not repair live data.** `connector.Part` is produced in exactly one
place today (`mailmap`), so nothing in production currently lands under the new value — nobody should
go hunting for rows to fix. What this removes is the hardcoded literal at the seam the first
channel-file producer will hit, and `fileCategoryFor`'s "any named transport" shape means that
producer needs no edit here.

### What changed

- **The vocabulary** (`core/1787192241`, additive): `message_attachment` joins the CHECK. No existing
  row changes. `SET LOCAL lock_timeout = '3s'` because it ALTERs a live table.
- **The derivation** — `capture` reads how a message arrived off the record and hands it across the
  `FileKeeper` seam; the `activities` writer binds it in the INSERT **and** in the audit image, where
  the same literal was repeated. Derived, never declared: a connector-supplied category is a string
  an untrusted producer can get wrong on a column that reaches the document library, and the record
  already knows.
- **Three arms, not two.** `ChannelProvider != ""` → `message_attachment`; `Kind == "email"` →
  `email_attachment`; anything else → `other`. See the review section — "not a channel" does not mean
  "mail".
- **The UI**: the label in all three document libraries, a filter chip so the new category is
  reachable rather than only scrollable, and three locales (`en`/`de`/`vi`).
- **Provenance is not a human's claim** (4th commit). The two `*_attachment` values answer "where did
  this come from". A hand upload came from the person uploading it, so the picker stops offering
  either, **and** the endpoint refuses either on a row whose `source` is `upload`
  (422 `category_not_assertable`). A captured file may still be corrected between the two — a
  mislabeled row is what a correction is for. This also closes the pre-existing half: `email_attachment`
  was assertable the same way before this arc.

### What the review found

Reviewed by Fable (craft), an independent adversarial pass, and a security redteam. Every finding below
was real and is fixed in this branch.

1. **The derivation was wrong for a third case, and the "REFUTED" was earned.** "No channel provider"
   was read as "mail". But `capture/gcal` files a `meeting` with no transport, and the offline demo
   connector files whatever kind its fixture names — so either one, the day it carries files, would
   have had them labelled email attachments. Neither carries `Parts` today, so nothing is currently
   mislabeled; the derivation was simply not *correct*. Now three arms, with `other` as the honest
   answer for a captured file that arrived on neither — the contract already calls `other` "the honest
   default, not a fallback for an unknown value", and this is exactly that case. Refusing the file was
   considered and rejected: it would fail the whole capture, and losing the message to protect the
   precision of a label is the worse trade.
2. **`NOT VALID` + `VALIDATE CONSTRAINT` bought nothing, and the comment said it did.** Two reviewers
   converged on it. `dbmigrate.Up` runs a whole migration file in ONE transaction, so the ACCESS
   EXCLUSIVE lock the `DROP` takes is held to commit and the `VALIDATE` scan runs underneath it
   anyway. Rewritten to the immediate form with the honest reasoning. `core/0084` carries the same
   misleading claim and is where this pattern was copied from — it has shipped, so it is filed as
   #1925 rather than edited.
3. **The invariant was enforced at one call site.** The "a human may not claim provenance" rule lived
   only in the upload picker, which no API or MCP client ever sees — while the contract prose this PR
   added asserted derivation was the only source. The contract contradicted itself. Now enforced at
   the endpoint, gated on the row rather than by narrowing the enum, and the contract says so.
4. **Fix narration in a doc comment** ("this writer used to assume mail…") — replaced with the
   invariant that stands alone.
5. **A named sentinel nothing could match.** `errCapturedFileCategoryMissing` was unexported and its
   only test matched a substring of the message, so the test asserted prose. Exported; the test now
   uses `errors.Is`, and moved to the unit lane where a **nil transaction** proves the refusal precedes
   any query — a claim a live database cannot make, because there a refusal and a missing parent are
   indistinguishable from the outside.

The redteam found no exploitable defect: the category reaches no authorization, retention, erasure or
SAR selector; the only category-filtered reads apply it on top of the existing visibility clauses; and
the PATCH endpoint is `x-agent-access: human-only`, so no passport or MCP tool can set a category at
all.

### Gates

- `make check` exit 0 · `make check-fe` exit 0 · `make craft-static` 0 findings
- `make test-integration` — full lane, **0 skips**
- New code covered: `fileCategoryFor` 100%; the derivation, the guard, the audit image and the
  provenance refusal each have a test **proved to fail** without its fix (the derivation one by
  collapsing it to a constant and reading the real failure message).
- **aicert not re-run, deliberately:** no prompt, no AI task and no routing config is touched, so the
  pin cannot have moved.

---
## Summary by cubic
Channel files are categorized as `message_attachment` instead of being mislabeled `email_attachment`. Category is derived from transport; uploads cannot assert provenance; audit images record the stored value.

- Derivation: channel → `message_attachment`, `kind=="email"` → `email_attachment`, neither → `other`. Threaded via `recordParts(..., fields, ...)`; `activities.RecordCapturedFiles` requires a category and returns `ErrCapturedFileCategoryMissing` when absent.
- API/DB: OpenAPI enums add `message_attachment`. Patching an uploaded file to `email_attachment`/`message_attachment` returns 422 `category_not_assertable`. Migration widens `attachment.category` CHECK with a bounded lock timeout; down migration re-labels `message_attachment` to `other` before narrowing.
- UI: Document libraries label/filter “Message attachment”; the upload dialog no longer offers either `*_attachment`; locales updated.
- Tests: Derivation, audit image, endpoint refusal, and UI behaviors are covered with new unit/integration tests.
- Docs: `STATUS.md` entry updated to reference publicly accessible sources.

**Review focus**
- `capture.fileCategoryFor` logic and `recordParts` signature/threading into activities.
- Guard paths: `activities.RecordCapturedFiles` refusal on missing category; `refuseAssertedProvenance` mapping to 422 `category_not_assertable`.
- Audit image writes `from.Category`; schema/migration and regenerated contracts.
- FE filter chip and labels for “Message attachment”; upload picker option removal.

**Rollout**
- Apply the migration; no backfill is required.
- Expect 422s if any client asserts `email_attachment`/`message_attachment` on uploaded files; update automation accordingly.

<sup>Written for commit 1463a33f1606288ad1b2ca4d46896ef065ec40a4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1955?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Message attachments are now categorized separately from email attachments.
  * Document lists and filters display the localized “Message attachment” label.
  * Attachment categories are automatically derived from the message’s transport, with “Other” used when unavailable.
  * Upload forms no longer offer captured-file categories such as email or message attachments.

* **Bug Fixes**
  * Prevented invalid provenance categories from being assigned to uploaded or captured files.


tinnhlt-gra:
## UAT — recorded against a live stack

Driven through Playwright on an isolated `DEV_SLUG=attcat` stack (api :18157, SPA :8157), signed in as the dev admin. One pass, three claims, no manual steps:

channel file reads as a message attachment

1. **The contact's Documents tab** — `roadmap-deck.png` (arrived on Telegram) reads **Message attachment**, and `signed-order.pdf` beside it still reads **Email attachment**. The negative matters as much as the positive: a missing label falls back to the raw key or the neighbouring one, and "a chat photo reported as mail" is the wrong record this whole change exists to stop.
2. **The account's document library** — the **Message attachment** filter chip exists and works: pressing it leaves `group-photo.jpg` and drops `nda.pdf`. A category a row can hold and no control can filter on is one a reader can only find by scrolling.
3. **The upload dialog** — the category picker offers Contract / Offer / Legal / Other and **neither** provenance value. A file arriving through that dialog arrived on no transport, so there is no arrival for it to claim.

### What the recording cannot show, and where it is proved instead

The fixtures for (1) and (2) are seeded rows, because **there is no live producer of channel files yet** — `connector.Part` is still produced only by `mailmap`, and the Telegram fetch lane is Phase 2. So this pass proves the read path: that the value renders, filters, and is not offerable.

The **write** path — that capture derives the value rather than the writer assuming mail — is proved at the row by the integration lane, not here:

- `TestACapturedFilesCategoryNamesTheTransportThatCarriedIt` captures a Telegram message and a mail message in one test and reads both categories back out of Postgres. Proved to fail: collapsing `fileCategoryFor` to a constant gives `a telegram photo's category = "email_attachment", want message_attachment`.
- `TestTheAuditImageOfACapturedFileNamesTheCategoryTheRowHolds` — the trail agrees with the row.
- `TestAnUploadedFileCannotClaimItArrivedOnAChannel` — the endpoint answers 422 `category_not_assertable`, writes nothing, and still accepts `contract`. That one is the server-side half of (3), which a UI recording structurally cannot cover: a curl never sees a dropdown.

Full lane green on this branch: **2825 tests, 0 failures, 0 skips.**

tinnhlt-gra:
@coderabbitai review

tinnhlt-gra:
All four findings checked out as stated; three are fixed in `c717d649c`, the fourth is filed.

**1. Panic as the pass condition — fixed, and it was the sharpest of the four.** `TestTheWriterAdmitsACapturedFileWhoseCategoryWasDerived` passed on *any* panic, so an unrelated nil dereference anywhere on that path would have been indistinguishable from the guard admitting a file correctly. The guard is now `refuseUnderivedCategory(from CapturedFileSource) error` — a function of its argument alone — and both arms are asserted directly, with no panic semantics. A second test keeps what the old one was reaching for: that the *writer* calls the guard, because a guard nothing calls is the same as no guard. The nil transaction stays only in the refusal test, where reaching a query would panic and therefore **fail**.

One correction: the "Affects 2 files" note points at `backend/internal/modules/capture/sinkpartscategory_test.go#L56-L72`, but that file contains no `panic`/`recover` — `grep` finds neither. The finding is real in one file, not two.

**2. Invalid captured-file fixture — fixed.** Right, and the reason is better than it looks: after the server-side rule this PR adds, a `source: "upload"` row carrying `message_attachment` is a payload the API *cannot produce*, so the test asserted a shape no reader ever receives. Now `source: "telegram"` / `captured_by: "connector:telegram"`, with a comment saying why. Swept the sibling — `companydocuments.test.tsx`'s fixture already carried the connector source.

**3. Undocumented `category_not_assertable` on the operation's 422 — fixed.** The code was documented on the request schema and missing from the response the caller actually reads. Both now name it, and the contract is regenerated.

**4. No outbox event for attachment creation — real, but not this PR's to fix: [#1956](https://github.com/gradionhq/margince-poc-v1/issues/1956).** Verified before deciding, and it is broader than the finding says: `grep -n 'storekit.Emit\|EmitEvent'` over `attachmentupload.go`, `capturedfiles.go` and `documents.go` returns **nothing**. No attachment writer has ever emitted — not the human upload, not the captured file, not the metadata patch — and there is no `attachment` event type in `public-events.yaml` at all. So this is a missing capability across the subsystem rather than a call this change dropped, and closing it needs a vocabulary decision first: a captured file already rides in a transaction that emits `activity.captured`, so a second event may be redundant on the capture path and necessary on the upload path, and that asymmetry is the actual question. Inventing a public event type inside a PR about a column is the wrong place to answer it. The issue carries the evidence, the ordering, and a note that the fitness test belongs in `writeshape_test.go` so a fourth writer cannot repeat it.

`make check-backend`, `make check-fe` and `make craft-static` are green on the pushed commit.

tinnhlt-gra:
Fifth finding — the `.tmp/channel-attachments/` reference in `STATUS.md` — also correct, and fixed in `1463a33f1`.

`.tmp/` is gitignored, so a fresh clone has none of it. This repo is public and CLAUDE.md requires that every instruction in it be followable, which makes a pointer at an untracked working directory a dead reference rather than a stylistic nit. Both channel-attachments entries now describe the remaining work where they claim it, so a reader needs nothing they cannot open.

Two things worth adding, since the finding is narrower than the problem.

**The same reference exists elsewhere in the file** — the transcript-ingest entry and `.tmp/ext-core-write-port/PLAN-06-kind-narrowing.md`. I have **not** touched those: they belong to arcs other sessions are working in right now, and the second looks like a plan somebody is actively following. Rewriting another agent's in-flight notes to tidy a doc is worse than leaving them, so it is filed as [#1957](https://github.com/gradionhq/margince-poc-v1/issues/1957) with the two locations named.

**And the real fix there is a gate.** `TestPublicTreeCitesNothingPrivate` already enforces this exact obligation — it catches a private repo name, a `specs/` path and a `foundation#NNNN` reference in tracked prose — but has no pattern for a gitignored working path, which is how four of these accumulated with every check green. The issue proposes deriving the pattern from `.gitignore` rather than listing it, so the next one is caught instead of reviewed.

All five findings are now either fixed here or filed with evidence. `make check-backend` is green on the pushed commit.

PR #1961, opened 2026-08-20
fix(authz): an edge is read under its own grant, and a census keeps it that way

### LINKED ISSUE #1846: Two org360 reads take the relationship edge under the endpoints' grants, without relationship.read
Found by an adversarial review of #1831, in code that PR does not touch.
Routed as a public issue by the repository owner's explicit decision; note
[SECURITY.md](../blob/main/SECURITY.md)'s usual rule that a writable
reproduction goes to a private advisory instead.

## The invariant

`relationship` is a first-class RBAC object (`identity/internal/policy/policy.go`,
`coreObjects`), and this tree already states why, where the people module gates
the same read:

> **`modules/people/relationship.go`** — *"The RBAC gate is the read one… reading
> an edge discloses its endpoints, which is what `relationship` read governs."*

The point is that the endpoints being **separately readable** is precisely the
situation the edge grant exists for. "Who currently works at Acme" is a fact
about the pair, not about either record.

## Where it is not held

| Call site | Gated on |
|---|---|
| `compose/org360/graphreads.go:39` | the endpoints only |
| `compose/org360/graphreads.go:90` | the endpoints only |
| `compose/org360/contacts.go:188` | the endpoints only |

Each reads `FROM relationship` under `person.read` / `deal.read` alone. Two of
the four sibling readers of this table do hold the gate (`people/relationship.go`
`GetRelationship`, and the employment list), so this is a missing copy rather
than a convention nobody adopted — CLAUDE.md's rule 1, *fix the invariant, not
the call site*, applies: grep every read of this table as one change.

## Precondition, which bounds the severity

All six seeded roles grant `relationship.read`, so **a default install is not
affected**. Reaching it takes a deliberate `setRoleObjectGrant` call
(`crm.yaml`, `PATCH /v1/roles/{key}/objects/relationship`) zeroing the grant —
a supported, documented operation whose own description says a system role is
editable.

The reason this is still `priority: high` rather than hygiene: an operator who
restricts that grant gets 403 from `GET /v1/relationships/{id}`, 403 from the
person's employment list, and an empty employment section on the person page —
so the restriction visibly works everywhere they would check, while these three
reads quietly continue to answer.

## What the fix looks like

`auth.Require(ctx, "relationship", principal.ActionRead)` at each entry point,
answering the module's usual existence-hiding shape rather than a bare 403
where the surrounding read already hides absence.

Worth deriving rather than repeating: a fitness function over the reads of
tables that ARE core objects would catch the fourth copy nobody has written yet.

## Related

- #1831 fixed the same class in the query grammar's join-table traversal, where
  the hop is dropped from the published vocabulary rather than refused at
  execution — so it reads as a relation that never existed instead of becoming
  an oracle for which edges the workspace holds. That shape is worth copying if
  any of these three reads can be reached by a caller who should not learn the
  edge exists.

tinnhlt-gra:
Closes #1846.

## What was wrong

`relationship` is a first-class RBAC object because reading an edge discloses
its endpoints **as a pair** — a fact neither endpoint's own grant covers. "Who
works at Acme" is about the pair, not about either record.

The issue names three `compose/org360` reads that answered that fact under the
endpoints' grants alone. The count was wrong in both directions.

**More sites than filed.** Nine compose reads put edge columns, or the pair
itself, into a response ungated — `org360` (3), `network` (2), `meetingbrief`,
`briefs`, `introseams` — plus ten module-internal reads of the same shape.

**And one package already had it right.** `person360` gates all three of its
edge reads through a package-local wrapper, invisible to a grep for
`auth.Require(…"relationship"…)`. That inverts the diagnosis: the defect is not
that nobody knew the rule, it is that **a convention living in one package's
private helper is not a convention** — its correctness is invisible to everyone
who does not already read that file.

## The root cause, and which half of the fix carries it

`platform/auth` owned the edge's ROW scope and says why: *"scope policy has
exactly one spelling … a second copy of a conjunction is a second place for one
of its arms to be forgotten."* The OBJECT gate had no such home, so each site
hand-assembled `Require(<endpoint>) + scopeClause(<endpoint>)` and there was no
function whose absence a reviewer could notice.

`auth.EdgeReadScope` gives it one, fusing the gate with the endpoint
conjunction. **But the helper is not the fix.** Nothing forces a new SQL read
through it — the census is what does, and the tree's own precedent is explicit
that one spelling holds *because* a gate enforces reaching it.

The general form is one level up and stated rather than papered over:
`rbacgate_test.go` judges only module entry points, and
`composerowscope_test.go` says object admission *"deliberately does not count"*
— so **no gate enforces the object half in compose for ANY core object**.
`relationship` is where nine authors already landed. Widening the census to
every object is filed, not built here.

## The three kinds of read

Every read of the table now passes the gate or carries a verdict. The
discriminator matters, because the obvious wording of it is wrong:

- **disclosing** — the edge determines which records, columns or counts appear.
- **predicate** — removing the edge condition could only *widen* what the caller
  already sees.
- **lifecycle** — cascades, merges, privacy sweeps, capture resolution. Gated by
  the write, or `PrincipalSystem`, which `Require` short-circuits.

"The edge's columns are not in the select list" would have filed the people
list's employer filter as a predicate — and filtering by employer answers "who
works at Acme" one page at a time, a **stronger** disclosure than the count on
the account. It carries the gate.

Gating the privacy readers would make a retention sweep **under-delete**;
gating `activities/orgscope.go` would be circular, since it is the machinery
that answers what a caller may see.

## Two sites deliberately NOT gated

- **`relatedCompanies`** — `crm.yaml` states partner/parent/child orgs *"need no
  grant beyond the organization read this whole endpoint already demands"* and
  *"can never be withheld wholesale"*, and `groups_omitted`'s enum has no value
  for them. Gating it is a **contract change and a product ruling**, not a
  defect fix.
- **`readDeparted`** — `DealCoverage` has no withheld channel, so an empty risk
  list renders *"Nothing flagged — this deal passes every coverage check"*.
  Gating it trades a disclosure for a **false all-clear on deal risk**.

Both carry a `deferred` verdict naming their follow-up. That is a decision, not
a quiet exclusion — the distinction #1831 drew when it refused a bare exemption
list.

## The refusal shape

Refuse **before** the statement, always. These queries carry their total as
`count(*) OVER ()` in the same statement as their rows, so filtering after the
read would report a remainder that discloses the size of the withheld
population — and no assertion about the returned rows would catch it.

Omissions go through the channel the contract already has: `groups_omitted`,
`sections_omitted`, a required array present-and-empty. Six affected fields are
`required` in `crm.yaml` with no response validation and unguarded frontend
dereferences, so a dropped field is an SPA crash, not a graceful absence.
`contact_count` goes **absent, never zero** — a zero is a wrong number on screen
where absent is a withheld one, and that field's own description already stated
the rule for `person:read`.

## End users see nothing

- All six seeded roles grant `relationship.read` (slot 9 of `objects(...)`), and
  `core/0181` + `0268` backfill every deployed database. Custom roles cannot be
  created through the API.
- `Require` short-circuits `PrincipalSystem`, so every privacy sweep, the
  auto-enrich pass and the capture resolvers are unaffected by construction.
  Passports and connectors carry the granting human's live merged grants.

## The census

`backend/edgereaders_test.go` — 55 sites: **20 gated, 4 predicate, 17
lifecycle, 5 deferred**. Modelled on `restrictedreaders_test.go`, verdicts as
three `gatekit.Waive` declarations following the `joinTables`/`notAnEdge`
precedent rather than a prefix convention inside a reason string.

Three properties keep it from lying, each closing a way it did:

- **gated AND waived fails** — otherwise a deferred hole stops counting down
  once it is closed;
- **a floor below the gated count** — an extractor that stops recognising SQL
  reads exactly like a clean tree;
- **the subject regex matches the literal's TEXT** — against the quoted token
  the end-of-text alternate can never fire, so a read ending at
  `FROM relationship` was invisible. `restrictedreaders_test.go` carried the
  same expression and the same blind spot; both are fixed, with a boundary unit
  test.

## What the reviews changed

An adversarial Fable pass **ran** the census rather than reading it and found
three holes in it, plus a product bug the census could not see:

- the end-of-literal blind spot above — my own mutation drill missed it because
  the probe ended the *line*, not the *literal*;
- the row half satisfied the census anywhere, so scope-without-gate — the
  inverted form of this very defect — read green;
- a by-name index let `Store.X` vouch for `Handlers.X`, making the gate
  **nondeterministic** (`rbacgate_test.go`'s header warns about that exact pair);
- **`AccountStrengthFor` swallowed the denial into `bucketNone`**, so a
  restricted caller was told an account's strength was *"none"* — a wrong
  verdict, not a withheld one. The same false-signal shape that keeps
  `readDeparted` deferred.

It also caught `edgeread.go` rationalising a gap in prose: it claimed the
conjunction was unreachable without the gate, and this PR's own code reaches it.

## Fixtures that had drifted

Eight fixtures were missing a grant every seeded role holds, and only adding a
gate surfaced it. `harnessperms.go` states the principle — fixtures *"mirror
identity's seeded role documents; a fixture that drifts from that seed still
passes, while proving nothing about production."*

`signalActor` is the one worth naming: the warm/cold verdict **is** the edge, so
without the grant `Warmth` now refuses rather than reporting *cold* — a cold
verdict reached by not being allowed to look for warmth is a false one.

## Verification

- `make check-backend`, `make check-fe`, `make craft-static` — green.
- Full `make test-integration`: **2830 tests, 41 packages, 0 skips**, exit 0.
- **New-code coverage 83.4%**, merging the lane's coverage pods with the unit
  pass the way CI does — the shards instrument only tagged packages, so either
  half alone misreports a DB-backed change.
- Every new gate mutation-verified, including the end-of-literal case the first
  drill missed and the extractor-collapse floor.
- **Six refusal tests pass a nil transaction.** That is the assertion: the gate
  resolves before the statement, so a read-then-filter implementation panics
  instead of passing.
- `make e2e-ai-report` byte-identical before and after — `brief_ranking/rank`
  stays `current/certified` on all three bindings. The cert case builds its
  request from fixture structs and never reaches the gated read, and the diff
  touches no prompt.

## Follow-ups filed

The `relatedCompanies` contract ruling, the `DealCoverage` withheld channel, and
widening the census to every core RBAC object.

---
## Summary by cubic
Gates every read of the `relationship` edge under `relationship:read` and the edge’s endpoint scope, and adds a census that enforces this across compose and modules. Previously several endpoints disclosed edge facts under person/org grants; now they either pass `auth.EdgeReadScope` or are explicitly classified as predicate/lifecycle/deferred. Also pins test time in `connected-agents.test.tsx` to stop date-based flakiness.

- Introduces `auth.EdgeReadScope` (object gate + endpoint conjunction) and applies it across `org360`, `person360`, `network`, `meetingbrief`, `briefs`, people list employer filter, contact counts, and `signals` route-in reads. Refusals happen before SQL to avoid leaking totals.
- Behavior changes:
  - `Organization.contact_count` is absent without both `person:read` and `relationship:read` (docs/typings updated). Old: absent only without `person:read`.
  - People list employer filter now requires `relationship:read`; old behavior answered under endpoint grants.
  - Meeting room withholds `deal_role` when the edge is denied; attendees remain.
  - Account relationship-strength roll-up refuses when the edge is denied; extracted to `strengthaccount.go`. Old: mapped denial to “none”.
- Adds a census (`backend/edgereaders_test.go`) that forces every `relationship` read to gate or declare a verdict; fixes an end-of-literal match bug in both the new gate and `restrictedreaders_test.go`. Parses per build tags, uses a package-constant edge alias, and names the narrows-nothing predicate.
- Defers gating for two surfaces by design: related-companies graph and departed-contact risk; both marked as deferred in code.
- Tests/fixtures: integration/unit tests assert refusals happen pre-statement (nil tx); fixtures now include `relationship:read` to mirror seeded roles; `connected-agents.test.tsx` pins the clock to keep fixtures stable.
- Rollout: no data migration. Seeded roles already include `relationship:read`. For custom roles, add `relationship:read` to preserve contact counts, employer filters, and route-in edges.

<sup>Written for commit 5a3fd522109f94c45f056f777c91ad3f719917f0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1961?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tinnhlt-gra:
## UAT — live stack, grant flipped through the supported route

Driven against a running api built from this branch (`DEV_SLUG=edgegate`, `:18386`).
Fixture: `Demo GmbH` with three live employment edges (Alice/CTO, Bob/Head of Ops, Carol/CFO).
The grant is flipped with `PATCH /v1/roles/{key}/objects/relationship` — the contract route an operator would use — not SQL.

### 1. Baseline, grant held

```
GET /v1/organizations/{id}/360     sections_omitted : []
                                   contact_count    : 3
                                   roster           : ['Alice Müller', 'Bob Schmidt', 'Carol Wagner']
GET /v1/organizations/{id}/graph   groups_omitted   : []
                                   person nodes     : 3
                                   dropped_count    : 0
```

### 2. `PATCH /v1/roles/admin/objects/relationship` → `{"read": false}` → `HTTP 200`

### 3. The same two reads, same session

```
GET /v1/organizations/{id}/360     sections_omitted : ['people', 'strength', 'health']
                                   contact_count    : <ABSENT>          ← not 0
                                   roster           : <ABSENT>          ← and named above
GET /v1/organizations/{id}/graph   groups_omitted   : ['contacts']
                                   person nodes     : 0
                                   dropped_count    : 0                 ← not a remainder
                                   nodes / edges    : present, empty    ← both are required
```

Four things here that a bare "the section is gone" would not show:

- **`contact_count` is ABSENT, not `0`.** A zero is a wrong number on screen where absent is a withheld one — and that field's own contract description already stated exactly this rule for `person:read`.
- **`dropped_count` stays `0`.** The gate refuses *before* the statement, so nothing was read and nothing was dropped. A read-then-filter implementation reports a positive remainder here, and that number discloses the size of the withheld population.
- **`nodes` and `edges` are present and empty.** Both are `required` in `crm.yaml`, there is no response validation, and the generated client types make them non-optional — so a dropped field is an SPA crash, not a graceful absence.
- **`strength` and `health` were withheld with it.** A roll-up over edges the caller cannot read is *unknown*, not dormant. `AccountStrengthFor` previously answered `bucketNone` — a fabricated fact about the account, and the one product bug the code review found.

### 4. The endpoints are still individually readable

```
GET /v1/people/{id}                200
GET /v1/organizations/{id}         200
GET /v1/people/{id}/employments    404   ← existence-hiding, unchanged
```

This is the step that makes the withholding attributable to the **edge** grant and nothing else: both endpoints of every withheld edge stay readable, which is precisely the situation `relationship` exists as a first-class object for.

### 5. Grant restored — the positive control

```
sections_omitted : []      contact_count : 3      groups_omitted : []      person nodes : 3
```

Every absence above is a withholding, not a fixture that never seeded.

---

The same round-trip is asserted in CI rather than left to this transcript:
`TestARoleRefusedTheEdgeSeesTheAccountPageWithoutItsEdgesAndIsToldSo` and
`TestTheContactRosterAndItsDealRolesAreWithheldWithTheEdgeGrant`
(`compose/integration/org360/`), each with its positive control first.

tinnhlt-gra:
@coderabbitai review

The earlier run reported `pass` without reviewing — it hit the OSS review limit. Requesting a real pass now that lint is green.

tinnhlt-gra:
@coderabbitai review

tinnhlt-gra:
## One unrelated fix rode along, deliberately

`fe-unit` and `frontend` went red on a test this branch does not touch:
`frontend/src/screens/connected-agents.test.tsx`. My diff changes **no frontend source** — the only frontend file in it is one generated JSDoc line in `schema.d.ts`, and `frontend/src/screens/` and `frontend/src/design-system/` are byte-identical to `origin/main`.

**It is a time bomb, and today is the day it went off.** The fixture carries `expires_at: "2026-08-20T08:00:00Z"` with `renewable: false`. The card derives its state by comparing that to `now` — `expired && !renewable` → `lapsed` → `ended`, and an ended row renders a badge where a live one renders Disconnect. So the component is correct; past 08:00Z today the fixture genuinely describes an ended connection, and three tests were asserting live-row behaviour against it. Two cases in the same file already stubbed their own clock and kept passing.

**`main` hides this.** The change classifier skips the frontend jobs for commits touching no frontend path, so `main`'s most recent run *skipped* `fe-unit` — it reads green while carrying a red suite. This branch surfaced it only because it touches one generated frontend file. A skipped required check and a passing one are the same colour.

Fixed by pinning the clock once in the file's existing `beforeEach` rather than adding a third per-test stub — the invariant, not another copy of the guard. `make check-fe` green: 3529 + 86 tests.

Diagnosis and three follow-ups (sweep for sibling fixtures, make real-clock dependence a fitness function, and the skipped-check hole) are in #1977. Happy to split this commit into its own PR if you would rather keep this one single-purpose — it is one `beforeEach`, and `main` is red without it.

Issue #1958, opened 2026-08-20
The filter builder types closed-set values freehand, so a typo reads as "0 match"


dungbanguyen:
The Filters & views builder renders **one bare text input** for every field type it has no special case for. `ScalarValue` (`frontend/src/screens/filterbuilder.tsx:451`) handles `date`, `boolean`, `number` and `currency`, and everything else falls through to `<input className="input">`.

So `picklist`, `text` and `id` share a control. For `text` that is right. For the other two it is not, and the picklist case is the one that misleads.

## What a user gets today

Filter deals by `status`. The picker offers the field (correctly — it is in the vocabulary), the operator list is right, and the value is a **free-text box**. Type `Open` instead of `open`:

- the engine compiles the clause — a picklist leaf compares text and validates nothing against the contract's enum ([#1851](https://github.com/gradionhq/margince-poc-v1/issues/1851));
- it matches zero rows;
- the screen reports **"0 deals match"** as a settled answer, with no error anywhere.

That is a wrong answer to a normal question, on a shipped and reachable surface. The two halves compound: either alone is survivable, but a freehand box over an unvalidated closed set produces silence instead of a refusal.

Every closed-set field is affected: `status`, `lifecycle`, `size_band`, `forecast_category`, `phase`, `relationship_type`, and every workspace-defined custom picklist.

`id` fields are the same control but less dangerous, because a UUID box does not look usable: `tag`, `owner_id`, `owner_team_id`, `stage_id`, `pipeline_id`, `organization_id`, `partner_org_id`, `project_id` all require pasting a uuid. That half is what [#1939](https://github.com/gradionhq/margince-poc-v1/pull/1939) added `FilterVocabularyField.references` for, and it needs no further decision.

## Three pieces, only one of which needs a ruling

**1. `id` → relation picker. Unblocked.** The vocabulary now names the target record type, so the control can offer tags, seats, teams, stages, pipelines, organizations or projects. Endpoints all exist (`/tags`, `/users`, `/teams`, `/pipelines`, `/projects`, `/organizations?q=`), as do `useRoster` and `EntityRef` in `screens/entityref.tsx`. Bounded relations fit the catalog's `Select`; organizations and partners are too numerous to enumerate and want async search, which is likely one new catalog control.

[...]

PR #1967, opened 2026-08-20
fix(frontend): an id filter clause names a record instead of a pasted uuid

dungbanguyen:
Part of [#1958](https://github.com/gradionhq/margince-poc-v1/issues/1958). The builder rendered one bare text box for every type it had no case for, so `tag`, `owner_id`, `owner_team_id`, `stage_id`, `pipeline_id` and `project_id` all required **pasting a uuid**. [#1939](https://github.com/gradionhq/margince-poc-v1/pull/1939) made the vocabulary say what each id points at; this spends it.

`owner id is Bruno Sá`, not `owner id is u-2`.

## The lookup is keyed on the target, not the field

That is the whole value of the contract change behind it. A map keyed on `stage_id` would not know about the next id leaf the engine gains — and two landed in one day last week. A map keyed on `stage` covers every field that ever points at a stage, and two fields sharing a target share one cache entry and one request.

## The endpoints said something useful

`/tags`, `/teams`, `/pipelines` and `/stages` accept no `limit` at all — the contract offers them no cursor, because they are configuration a human maintains. That is the same property that makes them safe to put in a dropdown. Only `/users` and `/projects` page, and one page of either is already more than a picker can usefully show. So the `bounded` distinction is not my judgement imposed on the API; it is the API's own shape, and I found it by the types refusing the parameter.

## `organization` deliberately keeps a box

A workspace has as many accounts as it has customers, so it cannot be enumerated — and a half-filled dropdown is **worse** than a box, because a reader cannot tell a missing account from an absent one. `boundedReference` makes that boundary a declaration rather than an accident, and the async picker it needs is its own change.

## What is gated

| Behaviour | Why it matters |
|---|---|
| A seat is chosen by name, and the wire still carries its id | the engine compares ids; the human should not have to |
| An unbounded target keeps its box | the documented exception, so it cannot regress into an empty dropdown |
| `exists` on an id field offers no record at all | the operator already asked the whole question, so a picker there would collect an operand the engine ignores — this pins the ordering inside the control |

Mutation-verified: making nothing bounded fails the first.

The test fixture now carries `references` on its id fields, because the server sends it — a stub that omitted it would exercise a response the API cannot produce. That made the builder read rosters, so the harness owes it a query client and a stub, with a fresh client per mount so one test's options cannot answer another's assertion.

## Story

`filterbuilder.tsx` had no story at all — a pre-existing gap `fe-uat` surfaces once the file is touched. Four now: the record picker, the unbounded exception, the operator-answers-it case, and nested groups.

There is deliberately **no** story for the open listbox. `Select` portals it outside the story root, so the capture was indistinguishable from the closed state — a story whose screenshot does not show what it claims is noise. I removed it after looking at the capture rather than trusting that it worked; the interaction is covered in the unit test, which can reach the portal.

## Still open on #1958

The more misleading half: **closed-set values are also typed freehand.** A mistyped `status` compiles, matches nothing, and reads as a settled "0 match". Custom picklists are solvable today through the catalog's `options`; core ones need a decision about where their allowed values come from. Neither is in this change.

`make check-fe` green (3532 tests), `make fe-uat` green.

---
## Summary by cubic
ID filter clauses now pick a record by name instead of requiring a pasted UUID. Previously `owner_id`, `owner_team_id`, `tag`, `stage_id`, `pipeline_id`, and `project_id` were free-text; now bounded references render a dropdown while the wire still carries the id. Unbounded targets and failed roster reads fall back to a text box; `exists` on an id field shows no record input.

- Keyed on the reference target (e.g., `app_user`, `stage`) rather than the field name; shared cache and request via `@tanstack/react-query` (staleTime 60s). `/tags`, `/teams`, `/pipelines`, `/stages` are unpaged; `/users` and `/projects` use a 200-item limit.
- Fallbacks: `boundedReference` gates dropdowns; when the options read fails, the control renders a textbox instead of an empty dropdown. Placeholders use `filters.pickRecord` and `filters.loadingRecords`.
- Stories document the record picker, the unbounded exception, the operator-answers-it case, and nested groups. Tests stub fetch, wrap with a per-mount `QueryClientProvider`, and add per-target assertions; fixtures include `references` for id fields.

**Migration**
- Ensure the filter vocabulary sends `references` for id fields; without it, those fields remain a text box.
- Render paths using the builder must have a `@tanstack/react-query` provider available.

<sup>Written for commit 505f7c10f2dc1221b0e885c74bb6a77d77db8262. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1967?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/i18n/de.ts
+++ b/frontend/src/i18n/de.ts
@@ -5419,6 +5419,8 @@ export const de = {
     "Die erste Seite der Treffer \u2014 genug, um den Filter zu pr\u00fcfen, nicht die gesamte Auswahl.",
   "filters.noMatches": "Keine Datens\u00e4tze entsprechen diesem Filter.",
   "filters.loadView": "Gespeicherten Filter laden",
+  "filters.pickRecord": "Eintrag wählen",
+  "filters.loadingRecords": "Auswahl wird geladen…",
   "filters.saveList": "Als Liste speichern",
   "filters.saveListTitle": "Diesen Filter als dynamische Liste speichern",
   "filters.listName": "Listenname",
--- a/frontend/src/i18n/en.ts
+++ b/frontend/src/i18n/en.ts
@@ -5426,6 +5426,8 @@ export const en = {
     "The first page of matches — enough to check the filter, not the whole selection.",
   "filters.noMatches": "No records match this filter.",
   "filters.loadView": "Load a saved filter",
+  "filters.pickRecord": "Choose one",
+  "filters.loadingRecords": "Loading choices…",
   "filters.saveList": "Save as list",
   "filters.saveListTitle": "Save this filter as a dynamic list",
   "filters.listName": "List name",
--- a/frontend/src/i18n/vi.ts
+++ b/frontend/src/i18n/vi.ts
@@ -5376,6 +5376,8 @@ export const vi = {
   "filters.noMatches":
     "Kh\u00f4ng c\u00f3 b\u1ea3n ghi n\u00e0o kh\u1edbp b\u1ed9 l\u1ecdc n\u00e0y.",
   "filters.loadView": "T\u1ea3i b\u1ed9 l\u1ecdc \u0111\u00e3 l\u01b0u",
+  "filters.pickRecord": "Ch\u1ecdn m\u1ed9t",
+  "filters.loadingRecords": "\u0110ang t\u1ea3i l\u1ef1a ch\u1ecdn\u2026",
   "filters.saveList": "L\u01b0u th\u00e0nh danh s\u00e1ch",
   "filters.saveListTitle":
     "L\u01b0u b\u1ed9 l\u1ecdc n\u00e0y th\u00e0nh danh s\u00e1ch \u0111\u1ed9ng",
--- a/frontend/src/screens/filterbuilder.stories.tsx
+++ b/frontend/src/screens/filterbuilder.stories.tsx
@@ -1,0 +1,129 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+import type { Meta, StoryObj } from "@storybook/react-vite";
+import { useState } from "react";
+import { FilterBuilder } from "./filterbuilder";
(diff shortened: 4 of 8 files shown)


dungbanguyen:
SonarCloud's quality gate was right, and chasing it rather than re-running it found a real defect.

The first sonar failure was a genuine flake (`SocketException: Connection reset`, 21s, never analysed) and re-running cleared it. The **second** was the quality gate itself, and I measured locally instead of assuming: `filterreference.ts` was at **27% statements**. The uncovered lines were the failed-read arms — and covering them showed the behaviour was wrong.

**On a failed roster read the control rendered an enabled, empty dropdown.** The hook returned `options: []` with `loading: false`, so the Select was live with nothing in it — which tells the reader *this workspace has no tags*. A confident answer to a question the server never answered, and no way to write the clause at all.

Fixed: the hook reports `failed`, and the control falls back to the plain box so a reader can still type an id. Both halves gated — the hook reports failure rather than emptiness, and the builder renders a textbox rather than a combobox.

The coverage gap was also pointing at something real about the happy path. Each target reads a different surface **and a different display column** — a seat is `display_name`, everything else is `name` — and getting one wrong fails quietly: the dropdown shows the right *number* of options, each with a blank label. Those six arms are now asserted one by one, against the endpoint each hits.

Two of my own assertions were imprecise as well: `findByLabelText("Value")` matches the Select's trigger as well as the input, so it resolved before the fallback happened and proved nothing. Both now wait on the textbox by role.

27% → **86% statements, 84% lines, 100% functions**. The remainder is the identical `throwProblem` line in each arm; one failure test proves that shape.

`make check-fe` green (3550 tests), `make fe-uat` green.

Issue #1969, opened 2026-08-20
Field history resolves no display names, so the per-field rail still shows agent:<uuid> and unnamed humans


tinnhlt-gra:
Third audit read surface with the PD-002 defect #1861 fixed on the other two.
Found by review while implementing #1861; filed rather than folded in, because
it is a scope expansion beyond the two surfaces that change agreed to cover.

`FieldHistoryEntry` (`backend/api/crm.yaml`, query in
`backend/internal/modules/privacy/fieldhistory.go`) returns `actor_type` and
`actor_id` with no display-name resolution — no `actor_name`, no
`on_behalf_of_name`. Its sibling reads now resolve both:

- `GET /audit-log` → `actor_name` + `on_behalf_of_name` (#1861)
- `GET /records/{type}/{id}/history` → `actor_name` + `on_behalf_of_name`, and a
  server-composed `summary` whose subject is the person (#1861)
- `GET /records/{type}/{id}/field-history` → **neither**

**What a reader sees.** `frontend/src/screens/history.tsx` renders field-history
rows through the same `ProvenanceTag` as record-history, and
`provenanceOfEntry` maps a machine actor to `{kind: "agent", agent:
entry.actor_id}` — so the rail shows the raw `agent:<passport uuid>`. A human
row resolves no name either, so the chip falls back to "typed by a person"
without claiming which one. On the record-history rows beside it, #1861 now
hands the resolved name in through `renderUser`; the field-history rows on the
same screen cannot, because the field is not on the wire.

The result is one screen where two adjacent rails disagree about whether
attribution names anybody.

**The fix is the one #1861 already built.** `privacy.auditActorNameJoins` is the
shared const both fixed reads use — the same two `LEFT JOIN app_user`s, keyed on
`app_user.id` so neither can duplicate a row. Field history is in the same
package (`backend/internal/modules/privacy/fieldhistory.go`), so this is:

1. Add `actor_name` and `on_behalf_of_name` to `FieldHistoryEntry` (additive,
   nullable — no client breaks).
2. Compose `auditActorNameJoins` into the field-history window query, alias the
   audit row `a`, and select the two `display_name` columns.
3. Pass the resolved name into `ProvenanceTag`'s `renderUser` on the
   field-history rail, the way record-history now does.

**Why `priority: normal` and not high.** #1861 covers the surface an auditor
opens first (the compliance log) and the record timeline. This rail is the
per-field diff feed — the same defect, a narrower audience, and no data is
[...]

Issue #1960, opened 2026-08-20
Two agent audit writers record no human authority, so their rows can name nobody answerable


tinnhlt-gra:
Found while implementing #1861 (the audit trail names the person, not the machine).

PD-002 says a change names the person behind it first. The read path can only
name a person the WRITE recorded. Two callers mint `actor_type='agent'` audit
rows with a NULL `on_behalf_of`, so those rows have no human to name — the
audit screen renders them "No human authority recorded", which is honest but
is a gap, not an outcome.

**1. `compose/extjobsrun.go:292` — an extension job tick.**

```go
return principal.Principal{
    Type:     principal.PrincipalAgent,
    ID:       "agent:" + w.decl.DispatcherKind(),
    UserID:   args.Principal,   // an is_agent app_user, not a human
    ...
}
```

`OnBehalfOf` is never set. `UserID` is the extension's own service identity, so
it is not the answer either. The operator who enabled the unit is plausibly the
human authority here, and the enablement record is where that would come from.
Whether a scheduled unit tick SHOULD carry a human at all is the product
question: an installation-wide tick nobody asked for may legitimately have
nobody behind it, in which case `actor_type='system'` is the honest spelling
and `agent` is the wrong one.

**2. `compose/commsscheduled.go:316` — a send scheduled before migration 0260.**

```go
if !sched.AgentOnBehalfOf.IsZero() {
    actor.OnBehalfOf = sched.AgentOnBehalfOf
}
```

Rows written before 0260 never recorded the authority and cannot be given one
now. This half is unfixable by design and the code already says so; it is
recorded here only so a future reader does not mistake it for the same defect
as (1). It ages out as those rows do.

**Why this was not fixed in #1861.** The strong form — make `storekit.Audit`
refuse an agent principal carrying no `OnBehalfOf`, so a missing attribution
fails at the write instead of being displayed — would break both callers above.
It is the right invariant *once* (1) is settled, and the wrong change to make
before it is.

Note the deliberate NON-finding: a bare CONNECTOR with no `on_behalf_of` is not
this bug. `compose/jobs_finance.go:91` writes one on purpose ("finance has no
connect flow and so no granting human"), and #1861 treats it as a legitimate
shape rather than a gap.

Related: #1928 covers the sibling inconsistency where `actor_type='system'` is
written beside a connector `actor_id`. The same mismatch exists with an
`agent:`-prefixed id (`compose/captureverdict.go` and siblings write
[...]

PR #1971, opened 2026-08-20
fix(privacy): the audit trail names the person, not the machine

### LINKED ISSUE #1861: The audit trail says "A teammate" and "agent:01a01740…" when it knows the person's name — attribution shows the machine, not who is answerable
Founder decision 2026-08-19, recorded as PD-002 in the principles repo: **a change names the person behind it first, and says a machine did the typing second.** A rep working through MCP is the rep — the change reads "Lars, via MCP", not "MCP", not "AI", not an agent slug.

The audit screen currently does the opposite, and it does it in three distinct ways.

**1. A human is never named.** `ActorTag` in `frontend/src/screens/audit.tsx` renders every human actor as either "You" or "A teammate". Never a name. The comment says why — *"A human actor is an opaque uuid on the wire"* — which is true and is the root cause, not a justification.

**2. An agent leads with its own identifier.** The row reads `agent:01a01740-c9c2-736d-a0b6-d3e3dcb13111` with `on behalf of you` as a trailing note. The raw UUID is the prominent half and the person is the footnote. PD-002 inverts that: the person is the label, the tool is the qualifier.

**3. The `on_behalf_of` id is rendered, not resolved.** Same root cause as (1) — the wire carries a UUID and the client has nothing to resolve it against.

**The data is not missing.** `audit_log` carries `actor_type`, `actor_id`, `passport_id` and `on_behalf_of` as four separate columns (`backend/migrations/core/0012_audit_log.up.sql:8-11`), and `storekit.AuditWithEvidence` populates all of them from the principal. The trail knows exactly who acted and for whom. What is missing is the display name on the read path: `AuditLogEntry` in `backend/api/crm.yaml:20543-20547` returns `actor_id` and `on_behalf_of` as bare UUIDs with no accompanying name.

**The fix has two halves:**

- **Contract and handler:** resolve the actor and the on-behalf-of human to display names on the `/audit-log` read path, and add them to `AuditLogEntry`. Both are `app_user` lookups. Names, not identifiers, are what the screen needs and what it cannot currently get.
- **Frontend:** rewrite `ActorTag` so the person is the label and the tool is the qualifier — "Lars, via MCP", "AI, for Lars". Keep "You" for the viewer if that reads better, but never "A teammate" when a name is available.

**Why `priority: high` for a display bug.** Attribution exists so somebody can be asked about a change, and "MCP did it" cannot be asked anything. The current rendering lets everyone disclaim: the rep says the agent did it, and the agent is not a party to anything. It also affects the surface an auditor reads first.

**One thing to settle while building:** what a change with no human shows. PD-002 reserves "system" for changes that genuinely have nobody behind them — a migration, a retention sweep over the whole installation. It must not become the fallback for an attribution the code failed to carry. If a nightly AI pass cannot name whose context it ran in, that is a defect in the caller worth failing on rather than displaying as "system".

tinnhlt-gra:
Closes #1861.

Founder decision PD-002: **a change names the person behind it first, and says a
machine did the typing second.** A rep working through MCP is the rep — the
change reads "Lars, via MCP", not "MCP", not "AI", not an agent slug.

Both audit read surfaces did the opposite, and the issue's scope was widened to
cover both after review, because fixing one would have left the two disagreeing
about how attribution reads.

## What was wrong

**`GET /audit-log`** resolved no display names at all. The wire carried
`actor_id` and `on_behalf_of` as bare uuids, so `ActorTag` could offer only "You"
or "A teammate" — never a name — and an agent row led with
`agent:01a01740-c9c2-736d-a0b6-d3e3dcb13111` with the person as a trailing note.

**`GET /records/{type}/{id}/history`** already resolved *both* names in SQL, but
exposed only `on_behalf_of_name`, and composed *"Agent acting for Lars updated
the record"* — the machine as subject, the person as a prepositional phrase. Same
inversion, on the surface a record page shows.

Attribution exists so somebody can be asked about a change. "MCP did it" cannot
be asked anything, and the old rendering let everyone disclaim: the rep says the
agent did it, and the agent is not a party to anything.

## What changed

- `AuditLogEntry` gains `actor_name` + `on_behalf_of_name`; `AuditHistoryEntry`
  gains `actor_name`. All additive and nullable — no client breaks.
- The two `LEFT JOIN app_user` display-name joins become **one named const**
  (`auditActorNameJoins`) shared by both reads in the package. Two audit
  surfaces resolving attribution differently is how a reader ends up trusting
  one and doubting the other. Both joins match `app_user.id` (a primary key), so
  neither can duplicate or drop an audit row — which would have corrupted the
  keyset pagination silently.
- `composeRecordSummary` inverts: *"Lars, via an agent, archived the record"*.
- `ActorTag` rewritten so the person is the label and the tool the qualifier.
  `audit.teammate` is deleted rather than kept as a fallback.

## The no-human case, settled on the grant rather than the actor word

The issue asks that `system` not become the fallback for an attribution the code
failed to carry. My first pass got this wrong and review caught it: it argued an
agent row with no `on_behalf_of` is legitimate — which is exactly why
`storekit.Audit` was **not** hardened to refuse one — and then had both
renderers call that same row a gap. The frontend test even used
`agent:extension_tick` as its gap example, which is the one writer
(`compose/extjobsrun.go`) whose rows are fine.

The deciding input is whether a **grant was presented**, which the data already
carries in `audit_log.passport_id`:

| Row | Reads as | Why |
|---|---|---|
| `passport_id` set, no `on_behalf_of` | "No human authority recorded" | A passport's granting human is recorded at grant time, so the row failed to carry what existed (the pre-0260 scheduled sends) |
| No `passport_id` | the machine's own id | A background pass nobody's context ran — no human to name, nothing failed |
| `actor_type='system'` | "System" | Genuinely nobody behind it |

So the split is passport-backed vs not, not agent vs connector. A passport-backed
*connector* is now the same gap; a bare connector is not, because some have no
connect flow and so no granting human by design (`compose/jobs_finance.go`).

## Two defects fixed that this change would otherwise have shipped over

1. **"You" had never rendered in production.** `ActorTag` compared
   `entry.actor_id` against a bare `meUserId`, but the wire spells a human actor
   `human:<uuid>` — the strings could never be equal, so every human read as "A
   teammate". `audit.test.tsx` was green on a fixture supplying `actor_id: "u1"`,
   an unprefixed id the real writer never writes (CLAUDE.md review-loop rule 6).
2. **The record-history row would have named the person twice.**
   `history.tsx` appended "on behalf of {name}" after the summary. That completed
   the old machine-first sentence; against the new one it reads *"Ada Authority,
   via an agent, updated the record on behalf of Ada Authority."* The suffix is
   gone, with its i18n key and CSS rule rather than leaving orphans.

## Deliberately out of scope

**The passport label.** `passport.label` holds exactly PD-002's example string
("Claude Desktop", "Cursor"), which would let the qualifier read "via Claude
Desktop". Left out because it raises a separate question this PR should not
answer alone: whether a **revoked** passport's label should keep appearing on
every audit row it ever wrote.

## Filed rather than guessed at

- **#1960** — `compose/extjobsrun.go` and `compose/commsscheduled.go` write
  agent rows carrying no human authority. Whether a job tick *should* carry one
  is a product question; the strong invariant (make `storekit` refuse it) is
  right only after that is settled.
- **#1969** — `FieldHistoryEntry` resolves no names either, so the per-field
  rail still shows `agent:<uuid>` beside record-history rows that now name
  people. Same screen, two rails disagreeing. A genuine scope expansion past
  the two surfaces this PR covers.
- **#1928** (comment) — the `actor_type='system'` beside an `agent:`-prefixed
  `actor_id` inconsistency is the sibling of the connector case that issue
  already tracks, and `compose/jobs_finance.go` resolves the identical trap the
  other way, which suggests one fix covers both halves.

## Verification

| Gate | Result |
|---|---|
| `make check-backend` | green |
| `make check-fe` | green |
| `make craft-static` | 0 blocker / 0 major / 0 minor, all four Go trees |
| Integration (real Postgres) | the 3 affected tests pass |

Reviewed by two independent adversarial passes before push; both findings sets
are folded in above. UAT video to follow in a comment.

---
## Summary by cubic
Names the person behind each audit entry and qualifies the tool (PD‑002). Old behavior led with machine identifiers; new behavior names the human first, reports “No human authority recorded” for passport‑backed rows with no person, shows machine ids only when no grant exists, and keeps “System” unchanged.

**Notes**
- API is additive: `AuditLogEntry` adds `actor_name` and `on_behalf_of_name`; `AuditHistoryEntry` adds `actor_name` (all nullable). Both reads resolve names via one shared `auditActorNameJoins`; predicates are fully qualified; pagination and counts are unchanged.
- Record history sentences flip to person‑first in `composeRecordSummary`, and the “gap” keys on `passport_id` (grant presented) rather than actor kind. Extracted `admitRecordHistoryRead` to reduce complexity and kept docs attached to the right function.
- Frontend attribution leads with the person and qualifies “via an agent/connector.” Fixes the “You” check by matching `human:<uuid>`, adds “Unknown member,” shows machine ids only when no human resolves, drops the on‑behalf‑of suffix in record history, passes `actor_name` to `ProvenanceTag`, updates i18n keys, subordinates machine identifiers in CSS, and adds a Storybook gallery. E2E now asserts filters actually remove non‑matches; seeds and mocks use typed actor ids.
- Tests/fixtures: one clock reading per fixture stabilizes ordering across rows.

**Migration**
- Prefer `actor_name` and `on_behalf_of_name` over parsing `summary`; treat `actor_id` as typed (`human:<uuid>`, `agent:<id>`, `connector:<name>`).
- Replace removed i18n keys with person‑first attribution and qualifiers (“via an agent/connector”).
- Surface the gap case: for passport‑backed rows with no human, show “No human authority recorded” rather than “System.”

<sup>Written for commit ce73a9c2605418da90e3c305b51875049dbb669a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1971?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Audit logs now show human-readable actor names.
  * Agent and connector actions identify the responsible person when available.
  * Entries distinguish unknown members, system activity, machine actors, and missing human authorization.
  * Record history displays clearer attribution and provenance details.

* **Bug Fixes**
  * Unresolved or deactivated actors no longer display misleading names.
  * Audit pagination and filtering retain accurate results across pages.

* **Documentation**
  * API documentation now describes the new audit attribution fields.


tinnhlt-gra:
## UAT — every attribution state, three locales

Audit attribution: the person is the label, the machine the qualifier

Captured by `make fe-uat` (the change-scoped Storybook render lane) from the new
`audit.stories.tsx`. The GIF cycles EN → DE → VI; the artifacts are in
`.tmp/fe-uat/compliance-audit-attribution--*.png`, regenerable with `make fe-uat`.

Every row a reader can actually get, rather than only the happy one:

| State | Renders as | Note |
|---|---|---|
| The viewer's own change | **You** | This branch had **never rendered**. `actor_id` is `human:<uuid>` on the wire and `meUserId` is bare, so the old comparison was always false and every human read "A teammate". |
| Another person | **Lars Vogt** | The headline fix — was "A teammate". |
| A member whose user row is gone | **Unknown member** | Honest. Never an invented name, never a raw uuid. |
| An agent under a human's authority | **Lars Vogt** · via an agent | PD-002: person is the label, tool is the qualifier. The passport uuid — previously the prominent half — is not shown at all. |
| An agent under the viewer's authority | **You** · via an agent | |
| A grant presented, no human resolved | **No human authority recorded** · `agent:scheduled_send` | A gap, and it says so. Deliberately **not** "System". |
| No grant presented | `agent:org_name_promotion` | Not a gap — a background pass nobody's context ran. Claims no missing authority. |
| A connector a person authorised | **Lars Vogt** · via a connector | Same rule as an agent: keyed on the grant, not the actor word. |
| A bare connector | `connector:finance` | No connect flow, so no granting human **by design** (`compose/jobs_finance.go`). Not a gap. |
| Genuinely nobody | **System** | The one case that reads "System". |

The last two rows are the distinction the first review pass got wrong. It argued
an authority-less agent row is legitimate — which is why `storekit.Audit` was
**not** hardened to refuse one — and then had the UI call that same row a gap,
using `agent:extension_tick` as its example, which is exactly the writer
(`compose/extjobsrun.go`) whose rows are fine. The rule now keys on
`audit_log.passport_id`: a grant was presented, or it was not.

DE/VI frames are there because the qualifiers and the gap phrase are localised
strings — a locale falling back to English would be visible. Note the action and
entity chips (`create`, `deal`) stay English by existing design: they are enum
data values passed through `humanizeToken`, not UI copy.

### What is not covered here

No live-stack pass against a real browser session. The Storybook lane renders the
component against wire-shaped fixtures with no server, so it proves the
RENDERING of each state, not the end-to-end path. The end-to-end path is covered
instead by the integration lane, which drives the real writer and asserts the
resolved names on the wire (`TestAuditLogResolvesTheHumanBehindEveryRow`,
`TestRecordHistoryHTTP`).

tinnhlt-gra:
@coderabbitai review

tinnhlt-gra:
## Review coverage on this PR, stated plainly

Merging with 30/32 checks green (`license gate` skipped, `cubic · AI code
reviewer` neutral — neither ran on this diff). One thing worth recording so
nobody later reads a green check as coverage it isn't:

**CodeRabbit's check reported `pass` for a period while it had not reviewed at
all** — its free OSS review quota was exhausted ("Review limit reached"), and an
explicit `@coderabbitai review` was refused a second time. A required reviewer
check reading green while carrying zero coverage is worth knowing about beyond
this PR.

It did subsequently review, and its two findings are fixed in `fe996b2ec`:

1. The integration fixtures read the clock per seeded row, so their ordering
   depended on when each call happened rather than on what the fixture declares.
   One reading per fixture now, every row offset from it. It cannot be a fixed
   literal — the create row above them is stamped by the real writer at real
   "now" and the seeded rows must sort after it, which is why the existing
   convention dates forward.
2. `AC-settings-16` asserted the agent row was visible *after* typing the actor
   filter — but it was already on screen before, so the assertion passed whether
   or not filtering happened. It now also asserts both non-matching rows are
   gone.

SonarCloud's quality gate passed (**97.0% coverage on new code**, 0%
duplication) with one issue: `ListRecordHistory` at cognitive complexity 16
against a limit of 15. Split into `admitRecordHistoryRead` — the object-level
gate reads as one list instead of a preamble to paging, and the row-level half
stays inside the transaction because visibility is a statement. That split
orphaned the doc comment onto the helper (revive), fixed in `ce73a9c26`.

Two independent adversarial review passes ran before the first push and found
five real defects, all fixed before CI ever saw them — including the one that
mattered most: the first implementation argued an authority-less agent row is
legitimate (the stated reason not to harden `storekit.Audit`) and then had both
renderers call that same row a gap, using `agent:extension_tick` — the exact
writer whose rows are fine — as its test example. The rule now keys on
`audit_log.passport_id`: a grant was presented, or it was not.

Issue #1974, opened 2026-08-20
main is red: three connected-agents tests cannot find the per-client Disconnect button


dungbanguyen:
`frontend/src/screens/connected-agents.test.tsx` has three failures on a clean `origin/main` (verified at `b3f24756` in a detached checkout, nothing of mine applied):

```
FAIL  ConnectedAgentsCard > names the client in each row's accessible action, so two connections are told apart
FAIL  ConnectedAgentsCard > disconnects through the connection's own credential, and warns that the whole connection ends
FAIL  ConnectedAgentsCard > leaves focus in the connections list after a disconnect, not on the document
      Tests  3 failed | 8 passed (11)

TestingLibraryElementError: Unable to find an accessible element with the role "button" and name "Disconnect Claude Code"
```

So the per-row action's accessible name no longer carries the client name. All three failures are the same cause: the tests look up the button by `Disconnect <client>`, and whatever renders it now names it differently (or does not render it).

## Why this is critical rather than normal

It is not the severity of the feature — it is that **`make check-fe` is red on main**, so every PR's frontend lane now carries a failure that has nothing to do with its diff. I hit it on my own branch and spent a cycle attributing it before concluding it was inherited; the next person will do the same, and some of them will assume their change caused it. That is the `main`/CI-red bar in CLAUDE.md, and it is also exactly what [#1494](https://github.com/gradionhq/margince-poc-v1/issues/1494) exists to prevent — a red run nobody trusts makes every other verdict unreadable.

## Attribution

Recent main commits touching those files are old (#1225, #1185, #1134), so this is more likely a change to whatever supplies the button's accessible name — a shared control or a message key — than an edit to the card itself. Worth a look at what the three tests have in common: they are the only ones that query by the composed `Disconnect <client>` name; the eight passing ones do not.

I have not fixed it: I am working inside the filters/views arc and this is outside it. Filing so it is visible and attributable rather than rediscovered per-branch.

dungbanguyen:
Fixed on main by [#1978](https://github.com/gradionhq/margince-poc-v1/pull/1978), and the cause is worth recording because my report named the wrong thing.

[...]

Issue #1851, opened 2026-08-19
The segment engine validates no picklist leaf against its contract enum, where the list parameters do


dungbanguyen:
A picklist filter leaf compares text. The engine holds no per-field enum, so a value outside the contract's enum compiles fine and selects nothing — while the **list parameter** for the same fact returns a 422.

## The divergence

| Fact | List parameter | Segment leaf |
|---|---|---|
| `relationship_type` | 422 `not_a_known_value` (`organization_list.go:201-204`) | compiles, selects nothing |
| `size_band` | 422 (`organization_list.go:195-198`) | compiles, selects nothing |
| `lifecycle`, `status`, `phase`, `forecast_category` | validated on their own list paths | same |
| custom picklists | options enforced on write | same |

So a typo in a saved view or dynamic-list definition is silently a filter that matches nothing, where the same typo on a list URL is told to the caller.

## Why it is not obviously a bug

'Selects nothing' is an honest answer for a value no row carries, and it is uniform across every picklist leaf including the custom ones. The engine's operand validation is by TYPE, deliberately — `storekit.Field` has no notion of an allowed-value set. So this is a **consistent** limit rather than a per-field oversight, which is why it is filed rather than patched.

It is currently gated, not just described: `TestAPicklistLeafComparesAnUnrecognisedValueRatherThanRefusingIt` (added in https://github.com/gradionhq/margince-poc-v1/pull/1850) asserts the behaviour and that the operand still travels as a bind parameter, so the day somebody adds validation to one leaf and not the rest, a test says so.

## If we want it closed

A `storekit.Field.Enum []string` (or a validator func) checked in `scalarOperand`, plus each core picklist leaf naming its contract enum. That is a real engine change touching every resource's vocabulary, and it makes the segment surface stricter — worth deciding rather than drifting into. A saved view written before the check would start failing at read time, which is the same class of problem `classification`'s retirement had to handle.

Found by the craftsmanship review round on https://github.com/gradionhq/margince-poc-v1/pull/1850, which flagged the original prose comment as a rationalized gap.

dungbanguyen:
## Decision: refuse on write, stay permissive on read

Ruled today. A POST/PATCH that stores a filter — a dynamic list's `definition` or a saved view's filter state — refuses a picklist value outside the field's set with a 422, exactly as the equivalent list parameter does. **Evaluation of an already-stored definition keeps comparing text and is unchanged.**

[...]

Issue #1677, opened 2026-08-18
Two surfaces disagree about which grant governs custom-field schema metadata


dungbanguyen:
Custom-field schema metadata — the `cf_*` column names and their types — is reachable through more than one surface, and those surfaces do not agree about which object grant governs it. Nothing shipped is exposed; the gap is that an operator who changes a grant does not get the effect the grant's name implies.

## The divergence

| Surface | Gate | Also returns |
|---|---|---|
| `GET /v1/custom-fields` (`modules/customfields/list.go`) | `custom_field:read`, no row scope | `label`, `slug`, `status`, active **and** retired |
| `GET /v1/segments/vocabulary` (`modules/collections/filtervocabulary.go`) | `list:read`, no row scope | operators per field, `custom` flag |
| Record payloads (`people`, `organizations`, …) | that record's `read` | only columns with a value — `ExtractValues` omits NULLs |

Every seeded role that holds `list:read` also holds `custom_field:read` (core migrations `0064`, `0183`, and `0268` for the newer management role), so on a default install the vocabulary read discloses a strict subset of what the catalogue read already gives the same caller. That is why this is not an advisory: there is no configuration the product ships in which a caller obtains something their granted permissions do not already admit.

## Where it does bite

`setRoleObjectGrant` is real and system roles are explicitly editable. An admin who does

```
PATCH /v1/roles/read_only/objects/custom_field {"read": false}
```

intending to hide the field catalogue from read-only members will find those members can still read every `cf_*` column name and type from `GET /v1/segments/vocabulary` — including retired columns and columns with no values on any record, which record payloads would not reveal. The admin's action did not do what its name suggests.

## Why the point fix is the wrong fix

Gating the custom half of the vocabulary on `custom_field:read` would be worse, not better. The engine itself accepts `cf_*` predicates without that grant, so the endpoint would start omitting fields a filter may legitimately name — breaking the equivalence LVS-EXT-8 exists to provide (*a field the endpoint omits is one the engine refuses, and vice versa*), which is the whole reason a filter builder can trust it instead of keeping its own copy.

This is a fix-the-invariant question, not a call-site one: decide **once** whether `cf_*` schema metadata is `custom_field`-gated or ambient to any caller who may read records, then make the catalogue read, the record projections and the filter vocabulary agree with that decision. Whichever way it goes, the answer belongs in one place with a fitness test over the surfaces that expose the metadata, rather than three handlers each choosing.

[...]

PR #1976, opened 2026-08-20
feat(contract): a picklist filter clause picks its value instead of typing it

dungbanguyen:
The misleading half of [#1958](https://github.com/gradionhq/margince-poc-v1/issues/1958). A picklist field offered a **free-text box over a closed set**, so `Open` instead of `open` compiled, matched nothing, and reported *"0 deals match"* as a settled answer — a wrong answer to a normal question, with no error anywhere.

Affected every closed-set field: `status`, `lifecycle`, `size_band`, `forecast_category`, `phase`, `relationship_type`, and every workspace-defined picklist.

The vocabulary now carries a picklist's allowed values and the builder offers them.

## One mechanism, on purpose

Both kinds of field go through `Field.Options`, because a picker for *some* picklists and a box for others would be more confusing than today's uniform-but-wrong control. Where the values come from differs, and that split is the **existing seam** rather than a new one:

| | Source | Why |
|---|---|---|
| custom field | the catalogue | per-workspace admin state — the same place labels come from, for the reason `filtervocabulary.go` already states |
| core field | declared beside the engine's other facts | the contract enumerates them; they are identical in every installation |

## Two things the generator taught

Both would otherwise have surfaced only as a bad dropdown:

- **oapi-codegen emits a `<nil>` member for a nullable enum.** `OrganizationSizeBandLessThannil` is real. A set assembled from the generated constants would have offered `"<nil>"` as something a human could pick — so the declarations are written out, and a test asserts no reported value is empty or `"<nil>"`.
- **A null in a contract enum is the column's nullability, not a value.** `forecast_category` enumerates `[null, commit, …]`; the offered set drops the null, because `exists: false` is how a filter asks for empty.

## Deliberately not included

**Making the engine refuse an out-of-set value — that is [#1851](https://github.com/gradionhq/margince-poc-v1/issues/1851).** It is a live-API behaviour change: a saved segment holding a value since removed from its set would begin failing at read time, which #1851 itself names as the hazard. `Field.Options` is advertisement only today, and it is precisely what makes closing #1851 cheap once somebody rules on the migration.

I split on that line rather than bundling it, because the surface fix carries no migration risk and the engine fix does.

## Gated four ways, each derived rather than listed

- every core picklist offers values, and **no other type does** — a set on a text column would promise a picker for a column with no closed set;
- the retired `classification` field offers **none**, since no surface may offer it at all;
- a custom picklist reports the catalogue's own options, through the real merge;
- the values reach the **caller**, not merely the declaration — the gate the last vocabulary field went without, and the reason its absence was caught in review.

Mutation-verified both ways: dropping the response mapping fails by name, and removing the picker fails the two behaviour tests.

`fieldcatalog.Column` gains `Options` so the catalogue can carry them; the `custom_field.options` jsonb it reads has been there since core 0063. Adding a slice also made `storekit.Field` non-comparable, which turned one test's `!=` into a compile error — the language agreeing with the review that called that check an identity test in name only. It reads `reflect.DeepEqual` now, and says so.

## A note on the frontend lane

`connected-agents.test.tsx` fails on this branch **and on a clean `origin/main`** — I verified in a detached checkout at `b3f24756` with nothing of mine applied. Three tests cannot find the per-client `Disconnect <client>` button. Filed as [#1974](https://github.com/gradionhq/margince-poc-v1/issues/1974) (`priority: critical`, since a red `check-fe` on main taints every PR's verdict). It is outside this arc so I have not fixed it; flagging so this diff is not blamed for it.

`make check-backend` green, `make craft-static` green (0 findings), and my own suites pass (41 tests across the filter screens).

---
## Summary by cubic
Picklist filters now offer a dropdown of allowed values instead of a free‑text box, preventing typos that silently matched nothing. Options come from the contract for core fields and from the catalog for custom fields; when options aren’t available or permitted, the UI falls back to typing.

- Contract/API: adds `options?: string[]` to `FilterVocabularyField` with docs; OpenAPI and generated types updated. The wire omits `options` when empty and clones slices.
- Backend: adds `Options []string` to `storekit.Field`. Declares core picklist sets and gates them against the OpenAPI enums (nulls not offered; no `"<nil>"`). Custom picklist options flow from the catalog via `fieldcatalog.Column.Options`; options decoding is centralized. `FilterVocabulary` always lists fields with `list:read` and includes custom picklist values only with `custom_field:read`; core values are always sent. Privacy fixes a catalog read to scan by name. Fit checks: HTTP integration asserts options travel; a reader census for `fieldcatalog.Column.Options`; and a ban on positional row scans into foreign structs.
- Frontend: the filter builder renders a select with “Choose a value” when `options` exist; otherwise it shows a textbox. i18n adds `filters.pickValue`.

**Rollout**
- Backward compatible: the engine still accepts out‑of‑set values; saved filters continue to work; no data migration.
- Permissions: `list:read` lists fields and operators; `custom_field:read` is required to receive custom picklist values; core picklist values are always included.

<sup>Written for commit 7a13a01169e0caad96d46b87d2e7b5a2fb35aab6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1976?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/.go-arch-lint.yml
+++ b/backend/.go-arch-lint.yml
@@ -245,8 +245,15 @@ deps:
     # generated routing schema, because "what the editor accepts" and "what
     # boots" is one claim and only the root package may hold both halves —
     # the ai module itself must not gain a JSON Schema dependency for it.
+    # corepicklistcontract_test.go asks collections' SegmentEngine what
+    # values each core picklist offers and holds them to the enums in
+    # api/crm.yaml — the same "only root may hold both halves" shape, and
+    # for the same reason: collections must not gain a YAML dependency to
+    # check a document it does not own. It reads the engine through
+    # storekit's own Field type rather than restating "picklist" as a bare
+    # string, which is the platform edge below.
     # Root carries test files only, so this grants no production edge.
-    mayDependOn: [contracts, compose, ai]
+    mayDependOn: [contracts, compose, ai, collections, platform]
     canUse: [yaml, jsonschema]
   cmd:
     mayDependOn: [compose, identity, people, privacy, deals, activities, approvals, agents, automation, ai, search, capture, consent, collections, signals, customfields, quotas, overlay, migration, webhooks, comms, finance, integrations, agreements, commissions, contracts, platform, migrations]
--- a/backend/api/crm.yaml
+++ b/backend/api/crm.yaml
@@ -21237,6 +21237,19 @@ components:
             may set. A retired one is absent here and still accepted by the
             engine, so no `status` lookup is needed to know whether to offer a
             field: if this operation listed it, it is offerable.
+        options:
+          type: array
+          items: { type: string }
+          description: |
+            A picklist field's allowed values, present (non-empty) when
+            `type=picklist` and absent otherwise. Offer these rather than a text
+            box: a value outside the set compiles and matches nothing, so a typo
+            reads as a settled answer instead of a mistake.
+
+            For a workspace-defined field these are the catalog's own `options`
+            for the same column. For a core field they are the values the
+            contract enumerates, minus null — a null in one of those enums is the
+            column's nullability, and `exists` is how a filter asks about that.
         references:
           type: string
           enum: [tag, app_user, team, organization, pipeline, stage, project]
--- a/backend/catalogoptionsreaders_test.go
+++ b/backend/catalogoptionsreaders_test.go
@@ -1,0 +1,141 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+package backendarch
+
+// Who may read a custom field's OPTIONS.
+//
+// fieldcatalog.Column carries three things with two different disclosure rules
+// (the port's own doc states them): Name and Type are schema, ambient to any
+// caller who may read a record carrying the column, while Options is catalogue
+// CONTENT — the values an admin authored — and a consumer handing them to a
+// caller needs `custom_field:read`.
+//
+// A Column cannot enforce that. It holds no context, so the obligation lands on
+// each consumer, and an obligation spread across consumers is one a new consumer
+// meets by not knowing about it. That is how this arrived: the filter vocabulary
+// began reporting options and disclosed them to any principal with `list:read`
+// for two pushes before a review noticed.
+//
+// So this gate does exactly one thing, and it is worth saying what it is NOT.
+// It does NOT verify that a reader checks the grant — a fitness function can see
+// which package reads a field, never whether the grant it checked governs the
+// value it returned. What it does is make a new reader VISIBLE: adding one fails
+// here, and the fix is to add the package below, which is a line away from the
+// rule in the port's doc. A gate that claimed the stronger thing would be a name
+// promising more than its body, which is worse than this.
+//
+// Derived structurally rather than by name, in the shape jobregistrationban_test
+// uses: a file cannot hold a Column without importing the port, so the readers
+// are the files that import it AND select `.Options`. The one blind spot is a
+// file that imports the port and separately selects `.Options` on some other
+// type — it would be reported as a reader it is not. That is a false positive
+// resolved by the same edit as a true one, and the alternative is type-checking
+// the whole module to sharpen a two-entry list.
+
+import (
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"io/fs"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+// The port whose Column type this is about. Only this module can import it —
+// it is internal, and each extension is its own module — so the walk is
+// backend/ alone rather than the license gate's three trees.
+const fieldcatalogPort = "github.com/gradionhq/margince/backend/internal/shared/ports/fieldcatalog"
+
+// gatekit:fixture the reader set this gate asserts, not a waived cost — a
+// package listed here is claimed to gate the read, and an entry no file reaches
+// is reported stale below rather than quietly ratifying nothing.
+//
+// The packages that may read Options, each for a stated reason. Adding one is
+// the point of the gate, not a workaround for it: state why the values reach a
+// caller and which grant that caller had to hold.
+var catalogOptionsReaders = map[string]string{
+	// Owns the column. It writes the jsonb (marshalOptions) and reads it back
+	// (unmarshalOptions), and its own list surface is `custom_field:read`.
+	"internal/modules/customfields": "the catalogue itself",
+	// Reports a picklist's values so a builder can offer them, and gates that
+	// half on `custom_field:read` (customPicklistValuesAreReadable).
+	"internal/modules/collections": "the filter vocabulary",
+}
+
+func TestOnlyTheDeclaredPackagesReadACustomFieldsOptions(t *testing.T) {
+	found := map[string]bool{}
+	err := filepath.WalkDir(".", func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if d.IsDir() {
+			// path != "." because the root's own name is dotted and skipping it
+			// would take the whole tree, reporting nothing for the most
+			// reassuring possible reason.
+			if path != "." && skipOptionsReaderDir(d.Name()) {
+				return filepath.SkipDir
+			}
+			return nil
+		}
+		if !strings.HasSuffix(path, ".go") || strings.HasSuffix(path, "_gen.go") {
+			return nil
+		}
+		if readsCatalogOptions(t, path) {
+			found[filepath.ToSlash(filepath.Dir(path))] = true
+		}
+		return nil
+	})
+	if err != nil {
+		t.Fatalf("walking the module: %v", err)
+	}
+
+	if len(found) == 0 {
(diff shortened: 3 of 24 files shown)


dungbanguyen:
## Review round applied — every finding was real

Two of them mattered more than the rest.

### The comment named a gate that did not exist

My previous push claimed `TestEveryCorePicklistOffersExactlyTheContractsValues` "reads the document and fails when the two part company." **No such test existed.** I verified it myself: the only occurrence of that name in the tree was the comment line asserting it. That is rule 5 in its most load-bearing form — a comment telling the next author the drift is already caught is worse than an ungated list, because it stops them writing the gate.

The gate is written now, as `backend/corepicklistcontract_test.go`, and it lives in the **root fitness tests** rather than in the module — the place this repo already puts gates that read `api/crm.yaml`, and for the stated reason: it walks the authoritative 3.1 document rather than the generated Go, so a contract-only edit cannot slip past it, and `collections` does not take a YAML dependency to check a document it does not own. (`arch-lint` was right to refuse the module version; the root entry in `.go-arch-lint.yml` now declares the two test-only edges with the reason.)

It reads the offered sets through `collections.SegmentEngine` — the production path, not a transcription — reports each direction separately, takes its resource list from the contract's own vocabulary-parameter enum, and **fails** on a picklist with no declared contract location rather than skipping it. It found two fields on its first run that nothing had ever compared: `deal.organization_lifecycle` and `deal.organization_size_band`, the customer link leaves that reuse the account sets.

Mutation-verified (`lost` → `losd` fires both directions).

### A defect CI caught in a module this branch never touched

`fieldcatalog.Column` gained `Options`. `privacy` read the same catalogue positionally:

```go
pgx.CollectRows(rows, pgx.RowToStructByPos[fieldcatalog.Column])
```

over a two-column SELECT. The third field at the **port** turned that into `got 2 and 3`, taking out Art. 17 erasure of cf_* columns and the Art. 15 subject export. Nothing in privacy changed; nothing in privacy mentions the field that broke it. Fixed by scanning into named fields — rule 1, the sibling copy of a read is part of the same change.

It reached CI because a positional mapping is a pgx runtime error, not a compile error: build, vet, the unit suites and every linter were green. So the rule is now derived — a positional row mapping may only target a struct its own package declares (`backend/positionalrowscan_test.go`, read off every call in the tree, failing on a walk that finds none). The other four uses in the tree all target same-package structs, which is what makes editing query and struct together possible.

### The rest

| Finding | Applied |
|---|---|
| `FilterVocabulary` gated only on `list:read` — the custom half is the catalogue's data | second gate on `custom_field:read`, withholding whole **fields** so the subset relation stays true; the engine still accepts a saved filter naming one. Also closes [#1677](https://github.com/gradionhq/margince-poc-v1/issues/1677) |
| the `""`/`"<nil>"` guard covered only `forecast_category` | swept over every core picklist |
| the wire assertion was declaration-side only | the HTTP integration test reads `options` over the real composed server, through the real catalogue write — mutation-verified |
| two spellings of the options jsonb decode | one, `unmarshalOptions`, beside the `marshalOptions` that already existed; the error names the column, not a driver type |
| `optionsOrNil` handed out a pointer into a shared package-level slice | clones, like its neighbour in the same literal |
| the `<nil>` argument over-generalized 2 nullable enums to 7 | says what is true |
| `Field.Options` gestured at "tracked separately" | names the test holding the advertisement-only behaviour |
| history residue in a test comment | replaced by the invariant it was gesturing at |
| the COALESCE comment gave the wrong reason | fixed |

The seed-pairing justification I had written for the authz gap (*"every seeded role holding `list:read` also holds `custom_field:read`"*) was the tell, and I want to name it rather than quietly delete it: that is a fact about a fresh install, not an invariant. Role grants are edited one object at a time.

`make check-backend` green · `make craft-static` green (0 findings) · both previously-failing integration tests pass · `check-fe` green apart from the three pre-existing `connected-agents` failures that also fail on a clean `origin/main` ([#1974](https://github.com/gradionhq/margince-poc-v1/issues/1974)).

dungbanguyen:
Correction to my comment above, on one line: **this does not close [#1677](https://github.com/gradionhq/margince-poc-v1/issues/1677).**

That issue asks for one decision — is `cf_*` schema metadata `custom_field`-gated or ambient? — applied to the catalogue read, the record projections **and** the vocabulary together, with a fitness test over all three. This PR moves one surface, and it moves it in the direction #1677 explicitly argues against, because the values changed the picture: a column name is arguably schema, but a closed set of values an admin typed is content, and after this change the two travel together through one field.

The full reasoning is [on #1677](https://github.com/gradionhq/margince-poc-v1/issues/1677#issuecomment-5354145740), which stays open and still needs a ruling. Also worth being precise about what the gate preserves: the endpoint's guarantee is a **subset** relation (everything advertised is accepted, and the gap is the retired set), and withholding whole fields keeps that exactly true — but not the stronger biconditional #1677 quotes, which retired fields had already broken.

Rebased onto `main` since. That picked up [#1978](https://github.com/gradionhq/margince-poc-v1/pull/1978), which fixed the red frontend gate I had attributed to a missing button: the real cause was a `connected-agents` fixture whose `expires_at` passed at 08:00Z this morning against an unpinned clock — recorded on [#1974](https://github.com/gradionhq/margince-poc-v1/issues/1974), now closed. `check-fe` is fully green on the rebased branch (282 files, 3558 tests), as is `check-backend`.

The rebase also merged this branch's value control with [#1967](https://github.com/gradionhq/margince-poc-v1/pull/1967)'s: an id field gets the record picker, a picklist gets the value picker, and both suites pass together (21 tests in `filterbuilder.test.tsx`). A field carries a reference or a closed set, never both, so the two branches never race.

dungbanguyen:
## Three commits: the #1677 ruling, its gate, and a correction to my own justification

[#1677](https://github.com/gradionhq/margince-poc-v1/issues/1677) has been ruled, and the answer is a **third position** — neither of the two that issue frames, because neither holds both invariants.

| What | Rule |
|---|---|
| a `cf_*` column's **name** and **type** | ambient to any caller with `list:read`, as before |
| a **custom** picklist's **values** | `custom_field:read` |
| a **core** picklist's values | ungated — contract enums, published in `crm.yaml` |

The earlier commit on this branch withheld custom fields **entirely**, which made the operation advertise *less* than the engine accepts — a builder would not offer a field a filter may legitimately name. Withholding only the values keeps the advertise/accept relation exactly true, and the degradation is the **old** behaviour rather than a lost capability: the field is listed, the operators are there, and the value is typed instead of picked, exactly as it was before options travelled at all.

A missing grant is an **answer** here, not a failure, so `customPicklistValuesAreReadable` folds only the denial into `false` and lets every other error reach the caller.

### The gate, and what it deliberately does not claim

`backend/catalogoptionsreaders_test.go` asserts which packages read `Column.Options` — two today, `customfields` (owns the column) and `collections` (gates the read). A third fails, with the message pointing at the rule in the port's doc.

It does **not** verify that a reader checks the grant, and says so in its own header. A fitness function can see which package reads a field, never whether the grant it checked governs the value it returned. Claiming the stronger thing would make it another test whose name promises more than its body — the defect this branch has now had to fix three times, so naming the limit seemed cheaper than making it four.

Derived structurally, in the shape `jobregistrationban_test` uses: a `Column` is unreachable without importing the port, so a reader is a file that imports it and selects `.Options`. The one blind spot is stated at the declaration rather than left to be discovered. Mutation-verified by making `deals` read `Options`.

The reader map needed the `gatekit:fixture` marker — `TestEveryPackageLevelReasonMapIsAWaiverOrADeclaredFixture` refuses a bare subject→reason map, because one "keeps reading as ratification long after the code it named is gone". `wantFixtureAnnotations` moves 18 → 19, which is that census working as designed.

### The correction, which is mine

My own review round caught a justification that is **not true**. Both the operation and the port explained "a name is ambient" with *"the record surfaces already expose them"*. They don't: `ExtractValues` omits a NULL, so a custom column with no value on any record is invisible in every record payload and visible here. #1677 states this in as many words and I wrote past it while citing that issue.

The ruling is unaffected — what was wrong was the reason, and the honest one is stronger. It is the **equivalence**: a builder composes from what this operation lists, so a field withheld here is one the engine accepts and no human can name. That argument needs no support, and its price is now stated rather than glossed — an all-NULL column becomes nameable, which is disclosure the equivalence buys deliberately, since the alternative is a column nobody can filter on until somebody happens to fill it in.

Fixed in both copies, and the port now records that ambient is a **decision, not an observation** — the sentence whose absence made the wrong justification feel safe to write.

### Verification

`make check-backend` green · `make craft-static` green (0 findings) · both new behaviour tests mutation-verified in both directions.

One note on the gates: `lint-modules` failed once mid-run with *"golangci answered from another checkout"*. That is the machine-wide content-keyed cache, not this diff — and the gate refused to report **any** module's verdict, including the ones that had already passed, on the grounds that the same cache answered those too. Cleared with the pinned binary as its own message instructs, and the full gate is green.

Issue #1984, opened 2026-08-20
Widening composerowscope over internal/modules would not have caught #1876 — three obstacles to fix first


tinnhlt-gra:
`backend/composerowscope_test.go` asks exactly the question #1876 was about — "does this read model hand back a row-scoped record id without reaching a row-scope probe?" — but only over `internal/compose`. Its own header (`:25-27`) says extending it over `internal/modules` "is a separate change with its own evidence". This is that issue, filed from #1876 so the evidence is not lost.

**The important part: widening the tier alone would NOT have caught #1876.** Three obstacles, each verified against current source, and each one a hole the widened gate would still have. Fix them, or the widening buys a green check over the same defect.

## 1. The obligation is table-agnostic

`reachesRowScope` (`:287-305`) takes `(fns, name, seen)`. It never consults `site.table`:

```go
func reachesRowScope(fns map[string]*rowScopeFnInfo, name string, seen map[string]bool) bool {
```

So a function that projects `organization_id` satisfies its obligation by reaching a probe over **deal**. `GetDeal` reaching `auth.EnsureVisible(ctx, tx, "deal", ...)` would have passed a gate whose finding was about the *organization* it names. The probe must be matched to the referenced table, not merely counted.

## 2. Any aliased FK is invisible to the extractor

`fkColumn` (`:493-494`):

```go
var fkColumn = regexp.MustCompile(`\b(?:[A-Za-z_][\w]*\.)?([a-z_]+)_id\b`)
```

The `[a-z_]+` capture is greedy, so `partner_org_id` yields table `partner_org`. `referencedTables` then drops it at `:506` on `!tables[table]`, because `partner_org` is not in `ownerScopedTables` — **silently, with no census entry and no waiver**. `partner_org_id` was one of the three references #1876 fixed and the gate could not see it at all.

This generalises: every FK named for its ROLE rather than its table escapes. `superseded_by_id`, `on_behalf_of`, `granted_by` and friends are the same shape.

## 3. The failure would be a shape accident, not a proof

Because of (1), `readDeal` fails only by not reaching *any* probe. Somebody makes it green by adding a deal-scope call, which fixes nothing about the organization it discloses, and the gate now certifies the defect.

## What a correct widening needs

- a per-reference obligation: the probe reached must name the referenced table;
- an explicit alias map for role-named FK columns, and a **census assertion that the extractor's yield is non-empty per module** so a regex that silently matches nothing fails loudly (`scripts/` has this pattern already);
[...]

Issue #1983, opened 2026-08-20
Contract and project reads disclose references the reader cannot open, and the contract makes two of them un-nullable


tinnhlt-gra:
Found while fixing #1876, which closed the same hole on `deal`. The invariant is *fix the invariant, not the call site* — these are the siblings, and they could not be closed in the same PR because the contract forbids the remedy that worked for deals.

## The defect

Two more read paths project a row-scoped record's id with no check on the referenced row:

- `backend/internal/modules/deals/project_read.go` — `projectColumns` projects `organization_id`.
- `backend/internal/modules/contracts/store.go` — `contractColumns` projects `organization_id, deal_id, project_id`.

An organization carries capture privacy: a connector-captured company is the importing user's alone until a human promotes it. A project keeps its own own/team row scope. Handing back the id of either makes the projecting record an existence oracle over a row the reader's own organization/project read would refuse — exactly the reasoning in `capture/tracestore.go:343-350`.

`contracts/visibility.go:47-52` does gate WHICH contracts a caller sees, by deal-scope OR org-scope. That is a different question: a contract admitted through its **deal** still hands back an `organization_id` whose organization the caller may not read. The disjunction is deliberately narrow (`:41-46`) about admission; it says nothing about disclosure.

`Contract.project_id` is not covered by that clause at all, and is nullable.

## Why it is not the same fix

`#1876` withheld the reference: null on the wire, named in `masked_fields`. That works because all three deal references are optional and `type: [string, 'null']` in `backend/api/crm.yaml`.

These are not:

| Field | `required`? | Nullable? | Remedy |
|---|---|---|---|
| `Contract.project_id` | no | yes | withhold, exactly as deals now do |
| `Contract.organization_id` | **yes** | **no** | needs a decision |
| `Project.organization_id` | **yes** | **no** | needs a decision |

Nulling a required field is a breaking contract change and the oasdiff gate will refuse it.

## The decision this needs

For the two required anchors, one of:

1. **Refuse the record instead of the reference.** If you cannot read the anchor company, you do not read the project or the contract — a row-scope clause on the read, not a mask. Narrower surface, but it removes rows that today's readers see, and `Project` has no capture-privacy story of its own.
2. **Make the anchor nullable** (major contract change) and withhold it like a deal reference.
3. **Rule it acceptable**: the anchor id of a record you were admitted to read is not a disclosure, because the record itself already announces the relationship.

[...]

PR #1986, opened 2026-08-20
fix(authz): a deal read does not name a record its reader cannot open

### LINKED ISSUE #1876: Deal reads disclose organization/partner/project ids without a visibility check on the referenced record
Found by the review of #1875 (shared customer identity). Deferred because it reaches code the PR did not touch.

**Where:** `backend/internal/modules/deals/deal_read.go` (GetDeal/ListDeals project `organization_id`, `partner_org_id`, `project_id`).

**What:** With deals readable by every seat, a deal reader learns the UUID (and thereby the existence) of a capture-private organization (`visibility='owner'`) or of a row-scoped project they may not read, through the deal's FK columns. Before the read-class change a deal out of the reader's scope hid these ids with it; the same disclosure already existed for cross-team-visible deals but was narrower.

**What it means if unfixed:** existence-hiding (404) for capture-private companies and row-scoped projects is bypassable by listing deals. No content leaks, only the id and the fact of existence.

**Suggested fix:** null the reference (or mark it withheld) when `auth.VisibleTo` on the referenced row fails, the way `activity.links[]` already drops link targets out of scope (`LinkTargetVisibleClause`). Apply the same rule to the other FK projections on identity records (`contract.organization_id/deal_id`, `project.organization_id`).

tinnhlt-gra:
Closes #1876.

## The claim this PR makes

**A deal read served by the deals module — `GetDeal`, `ListDeals`, and all four mutation responses — no longer hands back the id of an organization or project its reader could not open**, and the list's reference filters no longer confirm what the projection withholds. The reference goes out `null` and `masked_fields` names it, so a reader tells *withheld* from *nobody entered one*.

⚠️ **Scope, stated up front so nobody reads more into this than it delivers.** The same three columns are still disclosed by six read paths OUTSIDE this module — filtered/bundle export, prebuilt reports (`open-deals-per-company` groups by `organization_id` by default), `query_workspace`'s direct field predicate, offer reads (`buyer_org_id`, and `buyer_snapshot` carries the organization's NAME), direct contract reads, and the open question of whether a missing object grant should withhold too. Found by Codex review on this PR and tracked in **#2004**. They live in five modules with five visibility models; a module never imports a sibling, so each belongs in its own change.

## The strongest fact, which the issue omits

`deal_read.go:36` holds the file's only `auth.Ensure*` call:

```go
if err := auth.EnsureVisible(ctx, tx, "deal", id.UUID); err != nil {
```

**It is inert for every human seat.** `deal` is an `identityTable` (`auth/tableclass.go:28-30`) and is not `ownerPrivate` (`rowscope.go:107`), so `UnboundedFor` returns true for any human, `ScopeClauseFor` returns `""`, and `EnsureVisible` returns `nil` at `rowscope.go:376-378` **without issuing a query**. `GetDeal` performed no row check of any kind, on the deal or on anything it names.

## Why nulling is the right answer, and where the rule already lived

The write path has enforced exactly this rule all along — `applyDealLinkPatches` gates all three references with `auth.EnsureLinkTarget` before setting them (`deal.go:138,147,154`). **The system already agreed you may not NAME an organization you cannot see; it just never asked when handing one back.**

Prior art, followed rather than re-invented: `capture/tracestore.go:343-350` ("returning the id would make this surface an existence oracle over rows the timeline itself would refuse") and `compose/pipelinetrace/ladder.go:88-110`.

Contract-legal without a schema change: all three are already optional and `type: [string, 'null']` in `crm.yaml`. Their descriptions now say null can mean withheld.

## Evidence

**RED**, on unmodified `origin/main` with only the new test file added:

```
--- FAIL: TestADealDoesNotNameRecordsItsReaderCannotRead
    organization_id = 01a01e45-8641-…455, want withheld: it names a
      capture-private organization the reader cannot open
    partner_org_id = 01a01e45-8646-…41e, want withheld
    masked_fields is absent, want [organization_id partner_org_id] named
--- FAIL: TestTheDealListWithholdsTheSameReferencesAsTheGet
    the list handed out another team's project: 01a01e45-8697-…8a8
```

The FKs are set **through the real writer** (`UpdateDeal`), not hand-inserted. Capture privacy lands *after* the link write, because the admin's own `EnsureLinkTarget` would otherwise refuse to set a reference it cannot see — which is itself the write-side rule working.

**CONTROL** — `make check` printed PASS over the defect (`ok github.com/gradionhq/margince/backend`, contract drift OK, 165 ok/OK legs). That is the scope-escape evidence: the deterministic gate is structurally blind to this, and #1984 says why.

**GREEN** — both tests pass; `TestARepReadsEveryDealButNotAnotherTeamsAmount` still passes on the same seam; full integration lane `OK: integration passed with 0 skips (41 packages)`; `make check` green; `make craft-static` 0 findings.

**MUTATION** — deleting the three entries from `dealMaskableFields` takes the new test red on all five assertions. `fieldmask.go:26` states a mask naming an unlisted column is inert, so the registration is the load-bearing half, and this proves it rather than asserting it.

## The one real risk, named

A withheld null is indistinguishable from an intentional clear on write-back. The patch path is safe as written — `applyDealLinkPatches` only calls `p.Set` when `in.X != nil` — but a client round-tripping a full read into a PATCH with explicit nulls would clear the FK. `masked_fields` is the existing mitigation, and `amount_minor` has carried the same exposure since it shipped.

## Cost

One statement per referenced table per page, never a probe per row (`auth.VisibleSubset`, the read-only twin of `WritableSubset`). A page naming no organization and no project pays for neither. An admin pays one organization query — capture privacy does not yield to `row_scope=all`, by founder decision — and zero for projects, which short-circuit on `UnboundedFor`.

## Not fixed here, filed instead

- **#2004** — tracker: the six readers outside the deals module that still project these references (Codex review, round 2).

- **#1983** — the same defect on `project_read.go` and `contracts/store.go`. `Contract.organization_id` and `Project.organization_id` are `required` and non-nullable, so the remedy that worked for deals is contract-illegal there. Filed `status: needs-decision` with three options and a recommendation.
- **#1984** — widening `composerowscope_test.go` over `internal/modules`. It would **not** have caught this, for three reasons verified against source: `reachesRowScope` never consults `site.table`; the greedy `fkColumn` regex yields `partner_org` for `partner_org_id` and drops it silently, as it does every role-named FK; and the resulting failure would be a shape accident satisfiable by a deal-scope probe that fixes nothing.

## Reviewer, please look hardest at

- **the reader I may have forgotten** — someone who passes the deal gate and must fail the reference gate in a path other than `GetDeal`/`ListDeals`. `readDeal` is deliberately left unmasked because the update path's before-image depends on it (`deal.go:141,150,156`); is there a third caller?
- **`masked_fields` as two readings of one value** — it now carries role masks and reference withholding in one list. A client cannot tell which is which. Is that the right contract?
- **`VisibleSubset` asks `EnsureVisible`'s question, not `EnsureVisibleLive`'s** — an archived organization is still named. The reasoning is that archived records are readable through the archived filter, so it is not a disclosure. Disagree if you see a path where it is.

## Coverage · UAT

Backend only, no UI: no UAT video owed (the goal file says so explicitly). New code is exercised by the integration lane, which CI reconciles into the Sonar coverage report.

---
## Summary by cubic
Prevents deal reads and mutation responses from leaking `organization_id`, `partner_org_id`, or `project_id` the caller cannot open. Previously these ids were returned unconditionally; now they return null and `masked_fields` names each withheld field. Filters on these ids now honor target visibility to avoid existence oracles and return an empty page when the target is unreadable.

- Adds `auth.VisibleSubset` and extends masking to include unreadable references alongside role-based masks, applied once per row; one query per referenced table per page.
- Routes all caller-facing returns through `readDealForCaller` so GET, LIST, and every mutation response withhold the same fields.
- Narrows `organization_id`, `project_id`, and `partner_org_id` filters with the target’s visibility predicate; `?organization_id=<unreadable>` yields an empty page, not a 404.
- Documents null-as-withheld in `crm.yaml`; regenerates `api_gen.go` and frontend typings; adds unit/integration tests for GET, LIST, all mutations, and filters.
- Client impact: these ids may now be null where they were previously present; rely on `masked_fields` to distinguish withheld from empty. No data migration required.

<sup>Written for commit 39ba4790ffaaea5eea45b8c9e81261dd9a1d9c3a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1986?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Security & Privacy**
  * Deal responses now hide organization, partner organization, and project references that the viewer cannot access.
  * Hidden references are identified in `masked_fields`, while the deal itself remains visible.
  * Filtering by inaccessible references now returns an empty result without revealing whether records exist.

* **Consistency**
  * Visibility masking is applied consistently across deal lists, details, creation, updates, advancement, and archiving.
  * Documentation now clarifies reference-access and masking behavior.

### DIFF
--- a/backend/api/crm.yaml
+++ b/backend/api/crm.yaml
@@ -18592,9 +18592,9 @@ components:
         fx_rate_date: { type: [string, 'null'], format: date }
         pipeline_id: { type: [string, 'null'], format: uuid, description: 'Native mode: always a non-null pipeline FK. Overlay mode: NULL — an overlay-mirror deal has no native Margince pipeline row; the incumbent''s own pipeline id rides `raw` and the code-declared stage→semantic mapping drives tier resolution (overlay-augmentation OVA-MAP-6). A zero/placeholder UUID here is forbidden (dangling FK).' }
         stage_id: { type: [string, 'null'], format: uuid, description: 'Native mode: always a non-null stage FK; must belong to pipeline_id. Overlay mode: NULL (see pipeline_id; the incumbent dealstage id rides `raw`, OVA-MAP-6).' }
-        organization_id: { type: [string, 'null'], format: uuid, description: Primary org; never a raw lead. }
-        partner_org_id: { type: [string, 'null'], format: uuid, description: 'Deal registration/attribution to a partner org (A38/A41/ADR-0032). The org must have a `partner` row.' }
-        project_id: { type: [string, 'null'], format: uuid, description: 'The body of work this deal belongs to. A deal has at most one project; a project carries several deals over time. The deal and the project must name the same company — a cross-company pointer is refused 422.' }
+        organization_id: { type: [string, 'null'], format: uuid, description: 'Primary org; never a raw lead. Null when the caller may not read that organization, in which case `masked_fields` names it.' }
+        partner_org_id: { type: [string, 'null'], format: uuid, description: 'Deal registration/attribution to a partner org (A38/A41/ADR-0032). The org must have a `partner` row. Null when the caller may not read that organization, in which case `masked_fields` names it.' }
+        project_id: { type: [string, 'null'], format: uuid, description: 'The body of work this deal belongs to. A deal has at most one project; a project carries several deals over time. The deal and the project must name the same company — a cross-company pointer is refused 422. Null when the caller may not read that project, in which case `masked_fields` names it.' }
         owner_id: { type: [string, 'null'], format: uuid }
         status: { type: string, enum: [open, won, lost], default: open }
         lost_reason: { type: [string, 'null'], description: Required when status=lost. }
--- a/backend/internal/compose/integration/dealreferences_integration_test.go
+++ b/backend/internal/compose/integration/dealreferences_integration_test.go
@@ -18,6 +18,7 @@ import (
 	crmcontracts "github.com/gradionhq/margince/backend/internal/contracts"
 	"github.com/gradionhq/margince/backend/internal/modules/deals"
 	"github.com/gradionhq/margince/backend/internal/shared/kernel/ids"
+	"github.com/gradionhq/margince/backend/internal/shared/kernel/principal"
 )
 
 // dealReferenceFixture is one deal pointing at a hidden organization pair and
@@ -26,11 +27,13 @@ type dealReferenceFixture struct {
 	hiddenRefs ids.DealID
 	hiddenProj ids.DealID
 	openOrg    ids.UUID
+	privateOrg ids.UUID
+	wonStage   ids.StageID
 }
 
 func seedDealReferenceFixture(t *testing.T, e *Env) dealReferenceFixture {
 	t.Helper()
-	pipeline, open, _ := DealFixture(t, e)
+	pipeline, open, won := DealFixture(t, e)
 	admin := e.Admin()
 
 	// Seeded workspace-visible so the admin's link write passes its own
@@ -63,7 +66,10 @@ func seedDealReferenceFixture(t *testing.T, e *Env) dealReferenceFixture {
 	}); err != nil {
 		t.Fatalf("linking the deal to its project: %v", err)
 	}
-	return dealReferenceFixture{hiddenRefs: hiddenRefs, hiddenProj: hiddenProj, openOrg: openOrg}
+	return dealReferenceFixture{
+		hiddenRefs: hiddenRefs, hiddenProj: hiddenProj,
+		openOrg: openOrg, privateOrg: privateOrg, wonStage: won,
+	}
 }
 
 func TestADealDoesNotNameRecordsItsReaderCannotRead(t *testing.T) {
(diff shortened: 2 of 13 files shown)


tinnhlt-gra:
@coderabbitai review

tinnhlt-gra:
## Review round 1 — all five findings verified against source; three fixed here, two filed

Pushed `39ba479`. Every finding was checked by quoting current source before acting, not accepted on description.

**1 (HIGH) — mutation responses bypassed the mask. CONFIRMED and FIXED.**
Verified: `deal_create.go:204`, `deal_update.go:126` **and the `p.Empty()` early return at `:110`**, `deal_advance.go:157`, `deal.go:352` all returned `readDeal` output straight out. The reviewer is right that the role masks got away with this and reference withholding cannot — write authority over the *deal* says nothing about read authority over the *organization*. This is review-loop rule 3 ("anything that returns a record is a read") and I missed it.

Fixed with one spelling, `readDealForCaller`, used by every caller-facing return. `readDeal` stays unmasked on purpose and now says why: the update path's before-image comes from it, and an audit diff taken against a withheld null would record a change nobody made.

Gated, not just fixed: `TestEveryDealMutationResponseWithholdsTheSameReferences` is table-driven over all four mutations, so a fifth is a compile-time addition to that list.

**RED proof** — reverting `deal_update.go` to the old `readDeal` + `return current, nil`:
```
a patch that changes nothing still echoes the row handed back org 01a01e70-1144-…689 partner 01a01e70-114a-…42e, want both withheld
a patch that changes something handed back org 01a01e70-1144-…689 …
--- FAIL: TestEveryDealMutationResponseWithholdsTheSameReferences
    --- FAIL: …/a_patch_that_changes_nothing_still_echoes_the_row
    --- FAIL: …/a_patch_that_changes_something
```

**2 (MEDIUM) — the filter oracle survived. CONFIRMED and FIXED.**
Verified `appendDealFilters` answered all three columns unnarrowed. Each arm now carries the target's own visibility predicate (`referenceFilterClause`).

I did **not** take the "refuse the filter" reading that `auth/fieldmask.go:7-11` suggests, and the reason is worth stating: refusing outright would break filtering by a company you *can* see, which is the deals screen's main flow. An empty page is the better answer — it is indistinguishable from a visible company that has no deals, which is what existence-hiding wants, where a 404 would itself confirm the organization is real. `TestFilteringByAnUnreadableReferenceAnswersTheEmptyPage` asserts both halves: the hidden org returns nothing, the visible one still returns its deal.

**RED proof** — forcing `referenceFilterClause` to skip the EXISTS: `--- FAIL: TestFilteringByAnUnreadableReferenceAnswersTheEmptyPage`.

`query_workspace` (`search/queryfields.go:200`) is the same shape in a different module — a module never imports a sibling, so it is not fixable here. Noted for #1984's neighbourhood; say the word and I will file it separately.

**4 (LOW) — unspecified evaluation order. CONFIRMED and FIXED.** The clause is now hoisted before `tx.Query`, matching `WritableSubset`. Correct call: `args` is an operand of the same call whose *other* operand appends to it, and the spec fixes left-to-right only across calls.

**3 (MEDIUM) — the client. CONFIRMED, filed as #1989 (`priority: high`, `area: frontend`).**
Not fixed here, and I want to be explicit about why rather than quietly deferring it: it is a user-visible change to a design-system screen, it owes its own vitest coverage and a UAT recording, and this PR is backend-only and reviewed as such. Bolting it on would make the recording owed retroactively for work already reviewed.

One correction to the finding's severity, because it changes how #1989 gets triaged: **there is no silent data loss.** `mapDealUpdate` sends `organization_id: null`, and `idArg` (`deals/mapping.go:39-45`) maps a null body field to a nil pointer, which `applyDealLinkPatches` never sets. A rep editing an unrelated field and saving does not clear the link — the hazard needs them to actively pick a different company. The display gap is real and the blast radius argument (connector-captured orgs are `visibility='owner'`) is right; #1989 carries both, with the correction.

**5 (LOW) — the project→org hop. CONFIRMED, recorded on #1983.** The `deal_project_same_org` CHECK does make the withheld `organization_id` recoverable via a visible `project_id` and `GET /v1/projects/{id}`. Commented on #1983 so it is not triaged as cosmetic — that is precisely what makes it reachable.

## Gates after the fixes
`make check` → `OK: check-q passed` · `make test-integration` → `OK: integration passed with 0 skips (41 packages)` · `make craft-static` → 0 blocker, 0 major, 0 minor across all four trees.

tinnhlt-gra:
@coderabbitai review

tinnhlt-gra:
## Review round 2 — Codex (third reviewer). Six major findings, all confirmed, all filed as #2004

Three reviewers have now read this: **Fable** (5 findings — 3 fixed here, 2 filed), **CodeRabbit** (*"No actionable comments were generated"*, 5/5 pre-merge checks), and **Codex**, which found what neither of the other two did.

Codex's result was worth the wait, and it changes what this PR may honestly claim.

### What it found

Six more readers project the same three columns — or let a client SELECT deals by them — without probing the referenced row:

1. **Filter preview / filtered export / bundle export** (`compose/filterpreview.go`, `filteredexport.go`, `export.go`) — `maskedRowSelects` applies role field masks only; a private organization's UUID reaches `deal.csv`.
2. **Prebuilt reports** (`compose/report.go`) — **verified:** `open-deals-per-company` declares `fieldOrganizationID` in `dimensions` and groups by it *by default*, so the default report hands out private organization UUIDs. `deals-by-stage` accepts the same column as a bare filter — the very oracle `referenceFilterClause` closes on the deals list.
3. **`query_workspace` direct field predicate** (`agents/tools_query.go`, `search/querysql.go`) — hydration masks the returned deal's field but does not un-SELECT the deal the hidden reference chose. The row itself proves the binding.
4. **Offer reads** (`deals/offer.go`) — **verified:** `createOfferTx` probes only the deal, then `resolveBuyerOrg` copies `deal.organization_id` into `buyer_org_id` unprobed. Worse than the deal read ever was: after sending, `buyer_snapshot` carries the organization's display name and possibly its legal name.
5. **Direct contract reads** — already #1983.
6. **Object grants are never consulted** — `VisibleSubset` and `referenceFilterClause` weigh row scope and capture privacy, never `auth.Require(organization, read)`.

I spot-verified 2 and 4 against source myself rather than filing on the reviewer's word; both hold exactly as described.

### On finding 6, where I partly disagree

Codex rates it `major`. It is real, but it is a **policy question, not a defect**, and the system's own precedent points the other way: the write path's `auth.EnsureLinkTarget` is `EnsureVisibleLive` — row scope only, no object grant. Making the read stricter than the write would be a deliberate divergence, and it would change behaviour for existing seats (`integration.RepPerms` carries no `organization` grant on purpose). #2004 sets out three options and recommends changing both halves together rather than tightening one. It needs a ruling, not a patch.

### Why these are filed, not fixed here

They live in five different modules with five different visibility models — compose read models, search, agents, offers, contracts. A module never imports a sibling, so each fix belongs in its own module with its own test, and folding them in would make an already twice-reviewed PR unreviewable. Filed as **#2004** (tracker, `priority: high`, `area: authz`), with a suggested split of one child per area.

### The claim this PR makes, narrowed to what it actually delivers

> **A deal read served by the deals module** — `GetDeal`, `ListDeals`, and all four mutation responses — no longer names a record its reader cannot open, and the list's reference filters no longer confirm what the projection withholds.

That is what the tests here prove. It is **not** yet true of exports, reports, `query_workspace`, offers or contracts, and #2004 is the honest record of that. I have edited the PR body so the claim in the description says the same thing — a PR that overstates its reach is how the next person concludes the invariant already holds.

Nothing in this round changes the diff. Gates re-confirmed green after round 1: `make check` → `OK: check-q passed`; `make test-integration` → `OK: integration passed with 0 skips (41 packages)`; `make craft-static` → 0 blocker, 0 major, 0 minor.

Issue #1985, opened 2026-08-20
The agent surface cannot see what a transport can carry, and the tool listing has no room for it


tinnhlt-gra:
`GET /v1/channel-providers` now publishes an `attachments` object per transport — `carries`, `max_files`, `max_bytes_per_file`, `max_body_with_files` — so a composer can warn a rep BEFORE they press send. The `list_channel_providers` tool does not report it, so an agent staging a channel message with files can only discover the bounds by having the delivery parked.

## Why it was left out rather than added

It was implemented and reverted, which is the useful part of this report. Adding `attachments` to `agents.ChannelProviderEntry` plus one sentence of copy explaining it took the core tool listing from just under the ceiling to ~15040-15067 tokens, and `TestTheToolListingLeavesTheRunRoomInTheWindow` holds it at 15000 of a 24000-token window. That gate is not incidental: the listing rides in every Surface-B prompt and is never elided, so every token it grows comes out of the run's own observations.

So this is not "nobody thought of the agent surface". It is **the listing has no headroom left**, and the next capability that wants to reach an agent hits the same wall.

## What a fix has to decide first

- Does the listing's budget get raised, or does something in the catalog get shortened to pay for this? A per-capability answer will not scale — this is the second arc to meet the ceiling.
- Or does carriage reach an agent some other way than the directory tool: a refusal that names the bounds (the park reason already does, but only after staging), or a field on the send tool's own schema where it is read at the moment it matters.

Recommendation: shorten the catalog rather than raise the ceiling, and treat the listing budget as the thing under review — the ceiling exists to protect the run, and raising it to fit each new field is how it stops protecting anything.

The omission is named in a comment on `channelProviderDirectory` in `backend/internal/compose/handlers_channelproviders.go` so the next reader does not read it as a dropped field.

dungbanguyen:
## Decision, 2026-09-02: re-land it — the budget that blocked it has been raised

Re-triaged against `origin/main` at `d42887998`. **The premise of the revert has
expired.**

This ticket's reason for not shipping was arithmetic: the change *"took the core tool
listing from just under the ceiling to ~15040-15067 tokens, and
`TestTheToolListingLeavesTheRunRoomInTheWindow` holds it at 15000 of a 24000-token
window."*

The budget is no longer 15,000. `agenttooldescriptions_test.go:280-281`:

```go
listingBudgetNumerator   = 17
listingBudgetDenominator = 24
```

[...]

Issue #1977, opened 2026-08-20
connected-agents.test.tsx was a time bomb: it began failing by the calendar, and main hid it


tinnhlt-gra:
Found while landing #1961, whose diff touches `frontend/src/api/schema.d.ts` (one generated JSDoc line) and no frontend source at all.

## What happened

Three tests in `frontend/src/screens/connected-agents.test.tsx` began failing on **2026-08-20**, naming code nobody had touched:

```
Unable to find an accessible element with the role "button" and name "Disconnect Claude Code"
```

The fixture carries an absolute expiry:

```ts
const CONNECTION = { expires_at: "2026-08-20T08:00:00Z",
                     connection: { renewable: false, … } };
```

The card derives its state by comparing that to `now` — `expired && !renewable` → `lapsed` → `ended`, and an ended row renders a badge instead of the Disconnect button. So the component is **correct**: past 08:00Z today the fixture genuinely describes a lapsed connection. The tests asserted live-row behaviour against a fixture that had quietly become an ended row.

Two cases in the same file already stubbed their own clock (`vi.setSystemTime("2026-08-03…")`) and kept passing. The three that did not were on the real clock.

## Why nobody saw it

The change classifier skips the frontend jobs for commits that touch no frontend path. `main`'s most recent run **skipped** `fe-unit` and `frontend` entirely, so `main` reads green while carrying a red suite. It surfaced on #1961 only because that PR touches one generated frontend file and therefore classified into the frontend lane.

That is the more interesting half of this: a required check that is *skipped* is indistinguishable from one that *passed*, so a frontend regression can sit on `main` until an unrelated PR happens to touch a frontend path.

## Fixed in #1961

The clock is pinned once in the file's existing `beforeEach` rather than adding a third per-test stub — the invariant, not another copy of the guard. `afterEach` restores the real clock. 11/11 pass.

CLAUDE.md T11 already forbids this shape: *no `time.Sleep` / real-clock / real-network flakiness … inject a `Clock`*. It is stated for Go; the same rule wants enforcing on the frontend suite.

## Worth following up

1. **Sweep for siblings.** Any other frontend fixture carrying an absolute date near today is the same bomb with a later fuse. `grep -rn "expires_at: \"20" frontend/src` is the starting point.
2. **A gate.** A test that depends on the real clock is mechanically detectable — a fixture with an absolute date in a file that never calls `setSystemTime`. That is a fitness function, not a code review.
[...]

PR #1991, opened 2026-08-20
feat(comms): a channel reply carries files, and the transport says what it can take

tinnhlt-gra:
Phase 1c of channel attachments: a channel reply can carry files from the record library, the one carriage gate checks what the provider can actually take, and the transport directory publishes those bounds so a composer can warn before a rep presses send.

## What changed

**`AttachmentCarrier` answers with a descriptor, not a bool.** `Carriage{Carries, MaxFiles, MaxBytesPerFile, MaxBodyWithFiles}` lives on the published surface, aliased into the connector port like the file types — a unit and a core connector must answer this question with the same type, or the bounds a gate checks are two sets that can disagree. **There is still no default:** a connector that does not implement the seam answers the zero descriptor, so an adapter written before attachments existed cannot be mistaken for capable. A zero *bound* means "no limit beyond the contract's own"; only `Carries` says nothing may go.

`MaxBodyWithFiles` is the bound mail does not have. A channel that carries text-with-files as a caption bounds that text far below a text-only message, and such a message can be neither split into two provider calls — that reintroduces the partial send this seam exists to prevent — nor truncated.

**The gate checks four things, and each refusal says what to do.** One gate for mail and for a channel: `gateAttachmentCarriage` reads the provider name, the staged set and the body, and nothing that says which seam resolved them.

**`attachment_ids` on `SendMessageRequest`, threaded end to end** — request → `SendMessageInput` → `ChannelDeliveryRequest` → `StageChannelInput` → the delivery row's `attachments` column, which is the *same* column mail writes rather than a channel twin of it (`core/0196`'s shape constraint reads the array only, so no migration). A chain that accepted the field and dropped it before the wire would be invisible until a rep's file did not arrive.

**`GET /v1/channel-providers` publishes `attachments` per transport.** The parking gate's own doc asserts the composer already warned; nothing published the capability, so it could not.

**Two invariant pins**, both proved to fail with the property removed: a downloaded attachment is saved and never rendered (an inline `Content-Disposition` on a sniffed `text/html` file from a stranger is stored XSS), and retention treats a `message` activity exactly like a mail one (a kind filter on that selector would strand every captured channel file forever while mail was purged on schedule).

## Fixed under review, before this was pushed

Two adversarial passes ran over the branch. Sixteen fixes from twelve findings; the ones worth reading:

- **A send's `attachment_ids` had no runtime bound**, and this pre-dates the arc on the mail path. `maxItems` in the contract is documentation — nothing in this stack validates a body against its schema — so a long list cost a transaction per element, a jsonb array of that size on the delivery row, and a re-read of every element on every retry attempt. Bounded and de-duplicated in one place both transports resolve through, by a function of its argument alone; a fitness test holds the Go bound and gmail's declared carriage to the contract's own `maxItems`, because four statements of one number otherwise agree by luck.
- **The channel seam read the bytes after committing its at-most-once marker.** That seam cannot detect a prior send, so a file read failing between marker and wire parked a message that never left under *"the provider never confirmed whether this message was delivered… send again if it did not arrive"* — false, and it discourages the resend. The dispatcher now resolves the files once, before the marker, for both seams.
- **Two `MessageSender` adapters ignored `ChannelMessage.Files`** — the exact silent strip the field's own invariant forbids, in the diff that wrote the invariant. Both refuse with `connector.ErrFilesNotCarried`: the gate reads a capability the *connector* declares, so only the connector can refuse its own mistake.
- **An uploaded filename skipped `SafeFilename`**, which the capture path runs every sender-supplied name through — and a channel reply's park reason is a new place that name is read back. Sanitized where it enters the row, so one call covers every producer of an attachment row.
- The size refusal integer-divided by a MiB, so a 900 KiB bound read as "0 MiB" and a 10,000,000-byte one as "9 MiB". Both are instructions nobody can follow.

- **No connector declares channel carriage yet**, so a channel reply with files parks today. That is the no-default rule holding, and Telegram's adoption is the next phase.
- **The agent surface does not publish the carriage object.** It was implemented, measured, and reverted: it takes the core tool listing past the ceiling that protects a run's own observations. Named at the seam and carried by #1985, whose real content is that measurement.

## Gates

`make check` green (backend and frontend halves), `make craft-static` 0 findings whole-tree, full `make test-integration` green with 0 skips. `main` is red for an unrelated calendar-dependent frontend test (#1974 / #1977) — confirmed on a pristine `origin/main` worktree, not this branch.

---
## Summary by cubic
Channel replies can carry files from the record library, and the transport directory now publishes each provider’s attachment carriage so composers warn before send. Old: channels had no file carriage and composers lacked capability info. New: a `connector.Carriage` descriptor replaces the bool; the carriage gate enforces allowed/max files/max bytes per file/max body-with-files; the dispatcher resolves file bytes before the at-most-once marker; adapters that cannot send files refuse with `connector.ErrFilesNotCarried`. Side effect: no channel connector declares carriage yet, so channel messages with files park until a connector opts in.

- API: `GET /v1/channel-providers` adds a required `attachments` object per entry; `frontend/src/api/schema.d.ts` reflects this.
- Storage: channel replies snapshot attachments into the existing `attachments` column; `attachment_ids` are capped and de-duplicated consistently.
- Connectors: `AttachmentCarrier` now exposes `Carriage()`; Gmail declares bounds; `telegram` and unit transports refuse files via `connector.ErrFilesNotCarried`.
- Safety: uploaded filenames pass `SafeFilename`; downloads serve attachments as save-only; retention treats `message` activities like mail.

**Rollout**

- Update composers to read `attachments` from `GET /v1/channel-providers` and warn before send.
- Connectors that can carry files must implement `AttachmentCarrier.Carriage()` and honor declared limits; until implemented, return `connector.ErrFilesNotCarried` when files are present.
- No database migration; ensure clients accept the new required `attachments` field in the directory response.

<sup>Written for commit 7b186d9d37bf43a5dd084ea289d04857a792291d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/1991?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tinnhlt-gra:
@coderabbitai review

tinnhlt-gra:
@coderabbitai review

tinnhlt-gra:
## UAT — live stack, real HTTP, all four behaviours verified

uat-pr1991-channel-attachment-carriage

The clip logs into the SPA and then loads `GET /v1/channel-providers`, so the new `attachments` object is visible per transport. The rest was driven over HTTP against the same stack; every response below is real.

**1. Every transport publishes what it can carry.** All five channel transports (`dispact`, `telegram`, `whatsapp`, `zalo_oa`, `zalo_personal`) carry the object with all four keys. `telegram` reports `carries: false` with zero bounds, which is the no-default rule holding — no channel connector declares carriage yet. `gmail` is a mail sender and correctly absent from this directory.

**2. A channel reply with files is accepted, and then parks rather than sending the text alone.** 202 with the outbound activity, and the delivery row afterwards:

```
provider | telegram
status   | parked
attempts | 1
inflight | f
files    | 1
reason   | the telegram channel cannot carry files, and this message has "evil.html" attached;
           it was not sent, because sending the text alone would misrepresent what it contains
```

`inflight = f` is the part worth reading: the at-most-once marker was never committed, so nothing reached the provider — which is the ordering fix in this PR, seen from the outside.

**3. `too_many_attachments`, both halves.** Eleven distinct ids → `422 too_many_attachments`, `field: attachment_ids`, detail *"this message names 11 files and a message may carry at most 10; send the rest as a second message"*. The same id eleven times → **202**, and the staged delivery holds exactly one file: duplicates collapse rather than being refused, and the collapse survives into the snapshot.

**4. HTML bytes download as a file, never rendered.** A 72-byte `<script>alert(1)</script>` page uploaded as `text/html` answers `Content-Disposition: attachment; filename="evil.html"` with `X-Content-Type-Options: nosniff`. Never `inline`.

### What was NOT demonstrated, and two things found on the way

- **The send-email 422 was not observed live.** The mailbox pre-flight refuses first (`422 mailbox_not_send_capable`) because this stack has no send-capable mailbox, so the cap is unreachable on that path. It is the same shared `boundAttachmentIDs` both transports call and send-message exercises it — but the email 422 is not claimed as seen.
- **Making a Telegram send possible required bypassing the API.** `POST /v1/channel-connections` validates the bot token against a hardcoded `https://api.telegram.org` with no override, so no channel connection is creatable on a dev box. The connection row and channel identity were inserted directly (the token sealed through the *real* keyvault); everything downstream of those two rows — the refusals, the 202, the park — is genuine product behaviour.
- **`make seed-dev` writes its SQL half to the wrong database on a slugged stack** (`-d margince` while the stack runs on `margince_dev_attcarr`). Unrelated to this PR, filed separately.

tinnhlt-gra:
uat-pr1991-channel-attachment-carriage

Issue #1780, opened 2026-08-19
Configuration has three homes: boot file, environment, and runtime rows


dungbanguyen:
## What this is

#1252 unified the configuration *files*. This is the next split: which of the
three homes each setting belongs in, and moving the runtime-changeable ones out
of the file layer entirely.

| | Home | Changing it | Whose decision |
|---|---|---|---|
| **1. Boot configuration** | `margince.yaml` | needs a restart | the operator's |
| **2. Environment** | `.env.example` documents it | per environment | the operator's |
| **3. Runtime configuration** | `setting` rows + key vault | takes effect live | the **admin's**, through the API |

## The allocation rule has two axes, not one

"Can it change without a restart" is necessary but **not sufficient** for
category 3. The second axis is authority, and the repo already enforces it —
`settingscatalog_test.go` fails when a key exists in both the settings registry
and `margince.yaml`'s runtime surface, which is ADR-0061 §2's "no key exists in
both surfaces".

Two switches are runtime-changeable in principle and must NOT be category 3:

- **`operations.allow_data_reset`** — an admin who could arm it through the API
  could arm the purge of their own tenant's data. This was just made explicit
  (#1721): the arming is stated by the deployment, never inferred.
- **`capture.trace_payloads`** — the shipped comment says why: it retains
  correspondence the CRM otherwise refuses to store, so it is "a decision for
  whoever runs the installation, not for the people whose colleagues' mail it
  is."

So: **category 3 requires runtime-changeable AND the admin's to make.** Anything
that is the operator's call stays in the file even when it could technically
move.

## Proposed allocation

**Stays in `margince.yaml`** — `version`, `bootstrap_admin`, `seeds`, `auth`,
`mcp.connector_enabled` (routes are mounted at boot), `operations.*`,
`capture.trace_payloads`, `ai.capture_payloads`, `uploads.*`.

**Stays in the environment** — DSNs, Redis, ports, public/API base URLs,
connector OAuth client ids and secrets, the connector-state HMAC key, the
webhook sealing key, the `/metrics` bearer, the key-vault root key. The vault
cannot hold the key that opens it.

**Moves to `setting` rows** — `ai.routing` (the whole tier→model document),
`rates.*`, `email.smtp` (host/port/username), `company_context.rollout`,
`overlay_budget.*`, `capture.freemail_extra`.

**Moves to the key vault** — the BYOK provider keys (`GEMINI_API_KEY`,
[...]

PR #2000, opened 2026-08-20
feat(config): a deployment declares its model binding where it reviews it

dungbanguyen:
Part of increment 6 on [#1780](https://github.com/gradionhq/margince-poc-v1/issues/1780), and it exists because scoping that increment found the ticket's wording load-bearing in a way it did not intend.

## The problem with "retire ai-routing.yaml"

Routing now lives in the database and is editable through `PUT /v1/ai/routing`. Removing the file outright would leave the API as the **only** way to give an installation its first binding — so an infrastructure-as-code deployment could no longer declare which vendor it uses in the file it already reviews. It would need a post-boot API call, which is a real regression the ticket did not weigh.

## The fix

`margince.yaml` gains `seeds.ai_routing`. Consumed exactly once, like every other seed; the database is authoritative afterwards and an admin re-points a lane through the API without a restart.

```yaml
seeds:
  ai_routing:
    profile: eu_hosted
    tiers:
      premium: { provider: gemini, model: gemini-3.5-flash }
      ...
    embeddings: { provider: gemini, model: gemini-embedding-001, dimensions: 1536 }
```

Credentials are not there and never will be — a binding names providers, and each one's BYOK key is read from its conventional environment variable, so the section is safe to commit.

## Two decisions worth review

**Carried as a raw YAML node, not a typed struct.** The type it becomes lives in `modules/ai`, and `platform` may not import `modules` (the arch DAG). Mirroring the tier/provider/embeddings shape in `deployconfig` would be a second copy of a structure that already has a validator, free to drift from the one that enforces it. Compose decodes it through the ai module's own parser, so a seed is refused for exactly the reasons a stored binding is — and it fails the **bootstrap**, which is the one moment somebody is watching.

**Seeded inside the bootstrap transaction**, beside the rest. An installation that declared a binding must not be reachable with the organization present and the binding absent, even briefly, or the first request through an AI surface answers as unconfigured.

## What this does NOT do, and why

`--ai-routing` survives. It is one of three mutually exclusive ways the **no-database** debug lanes name a model — `compose.TaskProbeBrain`, `worker siteread`, `make ai-probe` — and those have no database to read a binding from. The FILE stops being where an installation's config lives; the FLAG stays a debug input.

## Tests

Integration, against a real database: a declared binding is served from first boot, and **outranks a routing file that says something different** — the file's only remaining job is seeding an installation that has none.

`make -C backend check` exit 0 · `make craft-static` PASS.

Refs [#1780](https://github.com/gradionhq/margince-poc-v1/issues/1780)

---
## Summary by cubic
Bootstraps a fresh installation’s model routing from `seeds.ai_routing` in `margince.yaml`, applied once at bootstrap/claim. Previously the first binding required `PUT /v1/ai/routing` or a routing file; now the database is authoritative and a file never overrides stored bindings.

- Seeds inside the bootstrap transaction; claim uses the same seed. Resets re-run seeds but do not overwrite `ai_routing`; they log `kept_keys` when the stored binding survives by design.
- Decodes/validates via the `modules/ai` parser; rejects unknown tiers/profiles, sovereign+cloud mixes, and out-of-range embeddings. Resolves YAML anchors, so aliases work.
- Reports discards: bootstrap/claim include `ai_routing` in `discarded_keys` when a stored binding exists; reset logs `kept_keys`. Stored binding always wins.
- Endpoint validation no longer echoes `base_url`; parse/scheme/host errors omit or redact the input to avoid logging credentials.
- `deployconfig.Seeds.AIRouting` is a `yaml.Node` value, so the documented block parses and round-trips; tests enforce the example decodes, aliases resolve, and bad configs fail at bootstrap. New integration coverage proves the declared binding reaches the database through the real `EnsureInstallation` path.
- Integration perms add `ai_routing` read/update. `--ai-routing` remains for no-database debug lanes.

**Operator actions**
- If you use `ai-routing.yaml`, move the binding to `seeds.ai_routing` in `margince.yaml`; the file now only seeds empty installs and never overrides the database.
- No action for existing installs unless changing the binding; use `PUT /v1/ai/routing`. Data resets preserve the stored binding and will log that it was kept.

<sup>Written for commit 01beb0bbd37df11e6a9cc6f516c8ba31dde41e70. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/2000?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added optional AI routing configuration during installation bootstrap.
  * Configured routing bindings are saved automatically and take precedence over conflicting routing files.
  * Added support for selecting AI profiles, providers, models, and embedding dimensions.

* **Bug Fixes**
  * Invalid routing configurations are now rejected with clear validation errors.

* **Documentation**
  * Added an example configuration showing how to define AI routing seeds.


dungbanguyen:
Addressed the review finding about the bootstrap path rather than marking it outdated — it was right, and the direct-call tests it flagged could not have caught what it named.

`backend/internal/compose/integration/routingbootstrapwiring_integration_test.go` starts from a deployment document, runs the real `EnsureInstallation` against an empty installation, and asks the database what it holds afterwards. `EnsureInstallation` → `configuredSeed` → `seedRoutingBinding` is a chain of three and an unwired seed **fails open**: the bootstrap succeeds, the installation comes up, and the declared binding is silently absent while every existing test still passes. Removing the `seedRoutingBinding` call from `configuredSeed` now fails the new case with the sentence that names the cause.

The fixture is parsed through `deployconfig.Parse` rather than hand-built, for the reason this branch already learned once: `Seeds.AIRouting` is a `yaml.Node`, and a node assembled in a test is a shape the deployment file cannot produce — which is exactly how the pointer-typed spelling of that field passed its tests while refusing every real binding.

One thing found while writing it, unrelated and left alone: `BootstrapAdmin.validate` refuses the newer `password: \${file:...}` reference form and demands the legacy `password_file`, even though `ResolvePassword` handles both and the field's own doc says to prefer `password`. Pre-existing on `main`; the test uses the accepted spelling and I will file it separately rather than widen this diff.

PR #2027, opened 2026-08-20
fix(ci-tests): the repair suite rolls back to a migration, not a count (#2024)

### LINKED ISSUE #2024: main is red: the RBAC repair suite steps back ONE core migration and assumes that step is the phase D drop
`make test-integration` fails on unmodified `origin/main` (`628421dac`). The
`./migrations` package is red, so every branch that runs the integration lane
inherits a failure it did not cause.

## Reproduction

```
$ git worktree add --detach <tmp> origin/main && cd <tmp>
$ make test-it DIR=backend/migrations RUN='TestTheRepairMigrationsHealADatabaseThatAlreadyRecordedTheLostBackfill'

=== RUN   TestTheRepairMigrationsHealADatabaseThatAlreadyRecordedTheLostBackfill/core/0192
    rbac_repair_integration_test.go:66: one core step back did not restore role.workspace_id — the newest core migration is no longer the phase D drop these repairs need rolled back, so this suite is replaying them in the wrong era again
=== RUN   TestTheRepairMigrationsHealADatabaseThatAlreadyRecordedTheLostBackfill/custom/20260806120000
    rbac_repair_integration_test.go:66: (same)
--- FAIL: TestTheRepairMigrationsHealADatabaseThatAlreadyRecordedTheLostBackfill (2.51s)
FAIL	github.com/gradionhq/margince/backend/migrations	3.077s
```

## Cause

`rollBackThePhaseDDrop` (`backend/migrations/rbac_repair_integration_test.go:271`)
rolls core back by a **count**:

```go
if _, err := dbmigrate.Down(context.Background(), conn, core, 1); err != nil {
```

That is correct only while the phase D drop is the newest core migration. It no
longer is:

```
1787216237_role_drops_the_tenant_column.up.sql     <- the phase D drop the repairs need
1787225001_deal_lost_reason_only_when_lost.up.sql  <- newest, landed by #2013
```

One step back now reverts the lost-reason migration and leaves `role.workspace_id`
dropped, so the repairs replay in exactly the era the helper exists to avoid.

The suite's own failure message predicted this — it ends "**again**", so this is at
least the second time a migration landing on top has broken it. The assertion is
doing its job; the rollback is the part that is wrong.

## Why it was not caught

Every recent `main` run is `cancelled`, not `success`:

```
$ gh run list --branch main --limit 4 --json headSha,status,conclusion
d51b03d6c completed/cancelled
d51b03d6c completed/cancelled
d08553163 completed/cancelled
d08553163 completed/cancelled
```

A cancelled run is that commit's only verdict and it reads like a pass in every
summary view.

## Suggested fix

**Derive the target instead of counting.** The helper knows the property it wants —
`role.workspace_id` present. Either roll back until that column exists (bounded, and
failing loudly if it never does), or name the phase D version and roll back **to**
it. Both survive the next migration landing on top; `Down(…, 1)` cannot, and will
break again the next time anyone adds a core migration.

This is the "a bound must be derived, not chosen" rule — a positional assumption
standing in for the condition actually wanted.

Found while rebasing #1990, which touches no migration and no RBAC file.
Related: #1494.

tinnhlt-gra:
Closes #2024. **`main` is red without this** — the `./migrations` package fails on
unmodified `origin/main`, so every branch running the integration lane inherits a
failure it did not cause.

## The defect

`rollBackThePhaseDDrop` reverted core by a **count**:

```go
if _, err := dbmigrate.Down(context.Background(), conn, core, 1); err != nil {
```

Correct only while the phase D drop is the newest core migration. It no longer is:

```
1787216237_role_drops_the_tenant_column.up.sql     <- the drop the repairs need reverted
1787225001_deal_lost_reason_only_when_lost.up.sql  <- newest, landed by #2013
```

One step back reverted the lost-reason migration and left `role.workspace_id`
dropped, so every repair replayed in the era the helper exists to avoid.

The suite's own assertion caught it and said so — ending *"in the wrong era
**again**"*. The assertion was doing its job; the rollback was the broken part,
and it had broken this way at least once before.

## The fix

The distance is derived from where the drop sits in the namespace, so a migration
landing on top moves it. A count is a guess about which migration is newest; the
name is the thing actually meant, and `coreStepsBackToPhaseD` fails loudly when
it is gone rather than silently reverting something else.

This is the repo's "a bound must be derived, not chosen" rule — a positional
assumption standing in for the condition actually wanted.

## Proof

Red first, on unmodified `origin/main` in a pristine detached worktree:

```
rbac_repair_integration_test.go:66: one core step back did not restore role.workspace_id — the newest core migration is no longer the phase D drop these repairs need rolled back, so this suite is replaying them in the wrong era again
--- FAIL: TestTheRepairMigrationsHealADatabaseThatAlreadyRecordedTheLostBackfill (2.51s)
```

Green after the fix. Then the test that matters — **does it survive the next
migration landing on top?** I planted a newer core migration
(`1787999999_a_planted_migration_on_top`) and ran both arms:

| arm | result |
|---|---|
| the derived rollback, plant on top | **PASS** |
| the counted rollback (`steps := 1`), same plant on top | **FAIL** — `rolling core back 1 step(s) to role_drops_the_tenant_column did not restore role.workspace_id` |

That second arm is the regression this PR prevents, reproduced deliberately. The
plant and the mutation were both removed.

## Why nothing caught it

Every recent `main` run is `cancelled`, not `success`:

```
d51b03d6c completed/cancelled
d51b03d6c completed/cancelled
d08553163 completed/cancelled
d08553163 completed/cancelled
```

A cancelled run is that commit's only verdict and reads like a pass in every
summary view. Worth a separate look; not fixed here.

## Coverage

**Test-only change** (`backend/migrations/rbac_repair_integration_test.go`), so it
produces no measured new-code coverage. The proof is the two-arm mutation above
rather than a coverage number.

## Gates

`PASS — 0 blocker, 0 major, 0 minor` on all four trees · `make test-integration` →
`OK: integration passed with 0 skips (41 packages, parallel)`.

---
## Summary by cubic
Fixes the repair integration suite by rolling `core` back to the phase‑D drop migration by version, not by a fixed count, so repairs replay in the correct era. Previously it always stepped back one migration; once newer migrations landed, it reverted the wrong change and left `role.workspace_id` dropped.

- Reuses `stepsDownTo` to derive the rollback distance to `phaseDDropVersion` from `core.Migrations` (version-based, no name matching).
- Proves the landing: verifies `dbmigrate.Down` reverted the derived count and asserts the newest applied core version is immediately before the drop.
- Migrates forward in the narrowed-arm test so assertions run at head; fails fast if the target version is missing. Test-only change in `backend/migrations/rbac_repair_integration_test.go`.

<sup>Written for commit 2560869678f77b77d137a0eeaed248d07ac2fedf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/2027?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Tests**
  * Improved migration rollback integration coverage.
  * Added validation that rollbacks return to the exact intended migration version.
  * Added checks for migration metadata and schema consistency after rollback and restoration.

### DIFF
--- a/backend/migrations/rbac_repair_integration_test.go
+++ b/backend/migrations/rbac_repair_integration_test.go
@@ -55,14 +55,14 @@ func TestTheRepairMigrationsHealADatabaseThatAlreadyRecordedTheLostBackfill(t *t
 				resetSchema(t, admin)
 				migrator := asMigrator(t, admin)
 				migrateAll(t, migrator)
-				// Then back ONE core step, because the newest core migration is
-				// ADR-0091 §8 phase D's drop of role.workspace_id and these
-				// repairs name it. The migration is correct in its own position
-				// — core runs in order, so core/0192 executes while the column
-				// is still there — and replaying it at head asks it to run in an
-				// era it never belonged to. One step back IS that era, and the
-				// suite migrates forward again below so every assertion is read
-				// at head.
+				// Then back to just below ADR-0091 §8 phase D's drop of
+				// role.workspace_id, which these repairs name. The migration is
+				// correct in its own position — core runs in order, so core/0192
+				// executes while the column is still there — and replaying it at
+				// head asks it to run in an era it never belonged to. How many
+				// steps that is depends on what has landed since; the helper
+				// derives it. The suite migrates forward again below so every
+				// assertion is read at head.
 				rollBackThePhaseDDrop(t, admin)
 
 				want := grantsWrittenBy(t, repair)
@@ -151,7 +151,7 @@ func TestTheRepairMigrationsHealADatabaseThatAlreadyRecordedTheLostBackfill(t *t
 				resetSchema(t, admin)
 				narrowMigrator := repointed(t, admin, migrator)
 				migrateAll(t, narrowMigrator)
-				// Same one step back as above, for the same reason: the replay
+				// Back to the same era as above, for the same reason: the replay
 				// below runs a migration from before role.workspace_id was
 				// dropped, so the fixture is seeded in that era too.
 				rollBackThePhaseDDrop(t, admin)
@@ -163,6 +163,12 @@ func TestTheRepairMigrationsHealADatabaseThatAlreadyRecordedTheLostBackfill(t *t
 					t.Fatalf("applying the %s/%s repair to the narrowed installation: %v",
 						namespace.Name, repair.Version, err)
 				}
+				// Forward again, as the shape above does: the assertions read the
+				// schema an installation actually runs on, and the suite does not
+				// exit sitting however many migrations back the rollback needed —
+				// a depth that grows with every core migration that lands.
+				migrateAll(t, narrowMigrator)
+
 				for _, object := range objects {
 					for _, role := range systemRoleKeys {
 						if held, present := readGrant(ctx, t, admin, narrowed, role, object); !present ||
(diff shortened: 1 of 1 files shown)


tinnhlt-gra:
## Review round 1 (Fable) — 5 findings, all 5 fixed in 256086967

Each was verified in the tree before being acted on.

| # | Finding | Verified | Fix |
|---|---|---|---|
| 1 | The call-site comment still taught the refuted invariant — *"the newest core migration is … the phase D drop"*, *"One step back IS that era"* | read at `:58`; both clauses false on this branch | Rewritten |
| 2 | Second copy: *"Same one step back as above"* | read at `:154` | Rewritten |
| 3 | `coreStepsBackToPhaseD` re-implemented `stepsDownTo`, already in this package with four callers — and keyed on `Name` | `comms_outbound_inflight_rollback_integration_test.go:131`; `dbmigrate.go:108` errors on a duplicate **version** and says nothing about names | Helper deleted; `stepsDownTo(t, core, phaseDDropVersion)` |
| 4 | `reverted` discarded and the surviving guard one-sided | `Down` spends its budget only on APPLIED migrations (`dbmigrate.go:170-179`) while `steps` derives from the embedded list | Two assertions added |
| 5 | The narrowed shape never migrated forward | the shape above it does, and says why | `migrateAll` added |

**Finding 3 is the one I should have caught myself.** The derivation already existed in the same package, in the same build-tagged file set, called by three other suites. I wrote a sixth spelling of it while fixing a bug caused by copies of one idea drifting apart.

**Finding 4 is the sharpest.** The old check — "is `role.workspace_id` back?" — is true at *every* depth older than the drop, so a rollback that went too far read identically to one that landed. Proven by mutating `steps` to `stepsDownTo(...) + 1`:

```
rbac_repair_integration_test.go:66: core sits at 1787203045 after the rollback, want 1787213222 — the step immediately below 1787216237; the repairs replay in the wrong era from here and the failure will name them, not this
--- FAIL: TestTheRepairMigrationsHealADatabaseThatAlreadyRecordedTheLostBackfill (2.49s)
```

The column check passed throughout that run. That is the same defect class this PR exists to close — a guard that proves *far enough* rather than *exactly there* — and it was still in my fix for it.

The original regression arm was re-run after all five: with a newer core migration planted on top, **PASS**.

Gates: `make check-q` → `OK: check-q passed` · `make craft-static` → `PASS — 0 blocker, 0 major, 0 minor` on all four trees · `make test-integration` → `OK: integration passed with 0 skips (41 packages, parallel)`.

tinnhlt-gra:
@coderabbitai review

Issue #1989, opened 2026-08-20
The deals screen reads a withheld company as 'no company' — the client honours masked_fields for amount_minor only


tinnhlt-gra:
Lands the moment #1986 merges. The contract now says a deal's `organization_id` / `partner_org_id` / `project_id` can be **null because it is withheld**, and names the field in `masked_fields`. The shipped client cannot tell that from empty.

## What the client does today

`frontend/src/screens/deals.tsx:582` is the only place `masked_fields` is consulted, and only for `amount_minor`:

```tsx
deal.masked_fields?.includes("amount_minor") ? <FieldGuard mode="masked" /> : …
```

So:

- **`deals.tsx:314`** — `toBoardDeal` resolves the company from `orgs?.get(deal.organization_id)`; a withheld null means the card renders with **no company at all**, indistinguishable from a deal nobody has linked.
- **`deals.tsx:423`** — the table's `organization_id` column has the same gap.
- **`deals.tsx:1612`** — the edit form seeds `organization_id: deal.organization_id ?? ""`, so the company picker opens **blank** on a deal that does have a company, inviting the user to pick one and silently re-point the deal away from the organization a colleague linked.

## What it is NOT

Worth stating so this is not triaged as data loss. `mapDealUpdate` (`deals.tsx:366`) sends `organization_id: str(values.organization_id) || null`, and an explicit null is **inert** server-side: `idArg` (`deals/mapping.go:39-45`) maps a null body field to a nil pointer, and `applyDealLinkPatches` (`deals/deal.go:138`) only calls `p.Set` when the pointer is non-nil. A rep who edits an unrelated field and saves does **not** clear the link. The risk needs the rep to actively choose a different company.

## Why it matters anyway, and why `high`

Connector auto-create mints organizations `visibility='owner'` (ADR-0063 §7, `auth/rowscope.go:94-107`), so an unpromoted captured company is private to whoever captured it. Every deal on such a company therefore reads as company-less to every colleague from the moment #1986 ships — which is exactly the state a connector-heavy installation is mostly in. The product's own UI becomes the counter-example to the contract's `masked_fields` claim.

## What is owed

1. A masked-aware company cell in both the board (`toBoardDeal`) and the table column — `FieldGuard mode="masked"`, the control `settings.tsx:1259` already uses.
2. The edit form must render a withheld reference as withheld rather than as an empty picker, and must not include it in the patch.
3. The same for `partner_org_id`, and `project_id` wherever it is shown.
[...]

Issue #1889, opened 2026-08-19
Bulk disqualify from the leads list sends no reason


LarsGradion:
Deferred from #1883.

`LeadBulkBar` (frontend/src/screens/leadbulk.tsx) still DELETEs `/leads/{id}` without a body for every selected row, so bulk-closed leads carry no disqualification reason while the single-lead dialog requires one.

What to do: a reason Select (active reasons from `useLeadDisqualifyReasons`) in the bulk bar, sent as `{reason_id}` with each DELETE.

luit-nfq:
Already done on `main` (#2036), so closing.

`screens/leadbulk.tsx` reads the disqualify-reason catalog with `useLeadDisqualifyReasons()` and sends `{ reason_id }` on the bulk DELETE, and the verb is refused until a reason is picked — so a bulk disqualify cannot go out reasonless, rather than merely being able to carry one.

Issue #1887, opened 2026-08-19
Lead page: reword the manual-signal form as three plain questions


LarsGradion:
Deferred from #1883.

`frontend/src/screens/leadsignals.tsx` still asks for Factor / Value / Evidence quality / Confidence / Why — a data-entry form no rep fills. The planned shape: three plain questions ("Website traffic?", "Company size?", "Budget?"), one Select each, an optional "How do you know?" line, and the Evidence-quality and Confidence selects behind a "More…" disclosure (the spec's provenance fields stay on the wire).

Why deferred: copy and control work only, no backend change; it did not fit the work-surface PR.

Issue #1669, opened 2026-08-18
A screen sheet is a second author for a shape the design system owns


tinnhlt-gra:
Tracker for the places where the visual authority for one shape lives in two files.

The design-system README already states the failure this causes: *"a copy drifts silently … every one of them looked fine in review on its own."* It has happened here at least once — #1148's value clamp and label wrap were tuned on the company side **after** `StatStrip` took its numbers from them, so a change to either row's legibility now has to be made twice, by someone who knows the other copy exists.

The set splits by *why* the second author exists, and the split is what tells you the fix:

- **The primitive lacks the variant, so a screen grew one.** #1526 — `PanelRow` carries a hover fill, so a panel of ruled blocks that are read rather than clicked grew `.co-worth-block + .co-worth-block` instead; hover on inert content tells the reader it is pressable. #1527 — `Meter` ships one bar height and two screen sheets reach into `.meterbar` to set another. A second caller of the same override is the signal that the primitive should own the size.
- **The shape was never moved in.** #1550 — the drop zone's styles are split across two *screen* sheets, and a component needs both; `contractform.tsx` looked correct only by accident of bundle load order, and drew the browser's raw file control when rendered alone. #1148 — the company page still draws `.co-strip` plus two scoped overrides where `StatStrip` exists and the contact page already uses it.
- **Two primitives answer one question.** #1014 — `Card` and `Panel` both mean "a card that carries a header", arrived in parallel from two authors, and nothing says which to pick.
- **Nothing notices the residue.** #805 — fourteen orphaned `.ob-*` rules style elements that no longer exist, and no gate stops a fifteenth. #807 — `tokens.css` declares the dark palette under `[data-theme="dark"]` alone, which is correct for the SPA that always stamps the attribute and wrong for an MCP App view whose host may state nothing, so that surface compensates in JS for what the stylesheet should own.

Filed as one tracker because the fix is a single rule applied seven times — **one authority per shape, and it lives in `design-system/`** — and because each migration is cheap only while someone is holding that rule in mind.

Related, and deliberately not adopted: #1502 (no layout primitives, so spacing and type cannot be gated the way colour is) is the same subject but stays under #27, where the review that raised it is tracked.

Children are listed below.

dungbanguyen:
Status pass on the seven children. Six are closed:

| child | why it existed | state |
|---|---|---|
| #1526 — `PanelRow` conflates a hairline row with a pressable one | the primitive lacked the variant | **closed** |
[...]

PR #2036, opened 2026-08-20
fix(frontend): a sweep through the open frontend issues

luit-nfq:
A sweep through the open `area: frontend` issues, taking the ones that could be
closed properly rather than the ones that were quickest. Seven issues closed,
one confirmed already fixed and locked with the coverage it was missing, and
one refused with its reason.

## What is in here

| Issue | What changed |
|---|---|
| #1533 `capability-gap` `high` | **"Send later" stops being a one-way door.** A new `#/scheduled` queue: see what is waiting, move a moment, withdraw a message before it goes. |
| #1369 `capability-gap` | **An access level can be changed without revoking first.** The share picker offers an already-granted subject, states what they hold, and confirms a downgrade. |
| #1444 `capability-gap` | **A unit's own records carry a name.** `GET /channel-providers` publishes `capture_sources`, so `ext:dispact-connector:dispact` reads as "Dispact" like the same unit's channel messages already did. |
| #1592 `bug` | **The contact page's lead verb names the transport it will open** — and is disabled, with a reason, when there is no way to reach the contact at all. |
| #1549 `fast-track-debt` | **An agreement's paper says when it is showing part of the list.** Both surfaces read `has_more` and report the remainder honestly. |
| #1887 `fast-track-debt` | **The manual-signal form is three plain questions**, with the provenance the wire needs behind "More…". |
| #1889 `fast-track-debt` | **Bulk disqualify asks why**, exactly as the single-lead dialog does. |
| #1526 · #1527 · #805 (under #1669) | **One authority per shape.** `PanelRow` is inert unless it says it is pressable, `Meter` owns its own dense geometry, and 63 orphaned onboarding rules are gone behind a gate that derives both sides from the tree. |
| #1250 `bug` `critical` | **Already fixed** by `26aecb4ff`; this adds the App-level regression test that can actually see the loop. |
| #1989 `bug` `high` | **Not touched — blocked on #1986**, which has not merged. See below. |

## Two things worth reading before the code

**#1250 was not a `<Shell>` defect.** It read as one because the throw surfaces
there: `useRoute` is a `useSyncExternalStore` over `location.hash`, so a
`navigate()` inside a passive effect makes React force a synchronous re-render
which runs the effect again, with no macrotask in between. The actual cause was
the *pair* of the app's only two automatic navigators — the onboarding gate
saying "go to onboarding" while the conversation's restore verdict said "you are
done, go home", each undoing the other forever on a fresh org where the wizard
row and the company profile disagreed. That guard landed in #1714; what was
missing is coverage, because the only test was a pure-function check of the
verdict and the loop is the pair. Removing the guard fails the new test with the
issue's verbatim error.

**#1989 is premature, not skipped.** It opens with "Lands the moment #1986
merges", and #1986 is still an open PR. In this tree `dealMaskableFields` holds
`amount_minor` and `currency` only, the migration seeds one mask row, and the
`Deal` schema's reference properties say nothing about withholding — so the
client fix would be masked-aware branches for a mask the server cannot emit, and
any test would have to fabricate a `masked_fields` value no backend produces.
The issue also undercounts the call sites (`partner_org_id` in the table column
and the edit form, two further org pickers, and `mapDealUpdate` sending an
explicit `null` for both references from a blank picker). Worth attaching to
#1986 and doing in the PR that merges the backend half.

## Contract change

One, additive: `ChannelProviderDirectory.capture_sources` plus
`CaptureSourceEntry`. Published *beside* the transports rather than among them
because these are not `ProviderRef` values — nothing sends on one, no
`activity.channel_provider` names one, and the id fails
`channel_provider.provider`'s own grammar, so registering them as transports
would mint a send anchor for a transport nobody supplies. Absent, not empty, on
an installation whose composed units declare no ingress. Generated artifacts are
regenerated in the same commit.

## Gates

`make check` exits 0 on the rebased tree: 296 frontend test files / 3659 tests,
86 extension tests, `tsc -b` clean, the four frontend script gates green,
`make craft-static` reporting 0 findings across all four Go trees.

## What is NOT verified, and what I would want looked at

- **No browser hit-test of the onboarding wizard.** The dead-CSS sweep deletes
  ~900 lines from `onboarding.css`, and a swallowed rule body still parses and
  still paints, so a passing suite is not proof a live rule survived. I checked
  it statically two ways — every removed selector against a whole-tree token
  search, and every class token named in markup against every stylesheet that
  declares one — and found nothing live deleted. That is the limit of what static
  analysis sees; the wizard wants one pass in a browser against a fresh org.
- **`make fe-uat` did not run.** It diffs against `origin/main` by commit and
  wants a stack; the new and changed stories all compile and register (the
  Storybook build is part of the gate above), but nobody has looked at them in
  both themes on this branch.
- **`make test-integration` did not run** — no database in this session. The one
  backend change here is a pure shaping function over the composed declaration
  set, covered by unit tests, and touches no store or SQL.

Draft on purpose: I would rather the two items above got a human's eyes before
the review pipeline spends a round on them.

---
## Summary by cubic
Adds a scheduled-send queue and clarifies several UX seams: scheduled messages are now manageable, extension-capture provenance renders as names, contract paper shows partial lists honestly, and the person page’s Write button names the transport or disables when none. Share updates can modify existing grants, manual lead signals ask three questions, copy drops em/en dashes, and share dialogs return focus to the subject field.

- Scheduled sends: new `#/scheduled` queue groups held/scheduled/settled; reschedule inline; withdraw with confirm; Tasks links only when items wait; composer confirm reflects “send later”; rows name the picked zone when it differs; picker seeds via the inverse of `scheduleFields` (`localDateTimeValue`).
- Directory: `GET /channel-providers` publishes `capture_sources` so `ext:<unit>:<system>` provenance renders as names (falls back to raw id).
- Contracts: form and row read paginated attachments via a shared hook; chips show files reached and state when more exist instead of truncating.
- Person page: the primary action names the transport (Email or “Message on <provider>”), falls back to “Write” when multiple, and disables with a reason when unreachable; transport derivation is shared with the composer.
- Share: picker allows selecting an already-granted subject, opens on their current level, confirms downgrades, no-ops report “Nothing changed.”, and seat-ceiling refusal names the subject’s seat; access pills use `Badge`; share dialogs return focus to the subject field.
- Leads: manual-signal form asks factor/band/source with optional “How do you know?”; provenance sends `assumption` with null confidence unless chosen. Bulk disqualify requires an administered reason and refuses the verb until one is picked.
- Design system and polish: `PanelRow` is inert by default; pass `interactive` for row-wide hover. `Meter` adds `dense`. Onboarding dead-CSS sweep is gated; a new app-level test asserts the onboarding gate and restore verdict cannot loop. Copy keys replace em/en dashes with plain sentences.

**Review/rollout**
- API: adds `ChannelProviderDirectory.capture_sources: CaptureSourceEntry[]` (additive, safe to cache; not transports). No migration required.
- Routing: scheduled-send screen lives at `#/scheduled`; the router key stays a literal to avoid pulling the lazy screen into startup.
- Error handling: new `isVersionSkewOf` helper de-duplicates version-skew checks; scheduled-send writes surface a single “reload” path on 409s.
- Stories/tests: new suites cover scheduleds, lead signals, bulk disqualify, contract-paper pagination, panel/meter geometry, and the onboarding guard.

<sup>Written for commit dbefcfde6d60b90f429958cbc0eea0b3485514dc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/2036?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added scheduled-message management, including timezone-aware scheduling, rescheduling, cancellation, statuses, and conflict handling.
  - Added scheduled-message indicators and navigation from Tasks.
  - Improved lead signals, bulk disqualification, sharing workflows, and person contact actions.
  - Added capture-source labels for clearer provider provenance.
  - Added release-version compatibility checks.

- **Bug Fixes**
  - Improved contract document pagination and partial-result messaging.
  - Fixed onboarding redirects and prevented redirect loops.
  - Refined panel hover behavior, meter sizing, and refusal-message wrapping.

- **Documentation**
  - Updated design-system guidance and translations for new workflows.


luit-nfq:
## End-of-work review round: applied

Two reviewers on the backend diff. Security found no exploitable finding and
confirmed the authorization argument holds by tracing both fields to compile-time
declarations. Craftsmanship found seven. Applied in `4791be476`, `make check`
still exit 0.

**Real defects, fixed:**

- **The label skipped the compiled-in names.** `titleCasedID` where
  `providerLabel` belonged, so a unit capturing from `whatsapp` published
  "Whatsapp" in the same document where `data` carries "WhatsApp" — two spellings
  of one transport in one list, which is the defect the change exists to remove.
- **`titleCasedID` had started collapsing empty segments.** Widening it from `_`
  to both separators swapped `Split` for `FieldsFunc`, so `deal__room` became
  "Deal  Room" → "Deal Room" while migration 0247 seeds the former. Both spellings
  are legal under the column's CHECK, so a fresh install and a booted one would
  have disagreed about that provider's name — the one thing the seeded-label
  fitness test promises cannot happen. Regression I introduced; `-` is folded onto
  `_` now.
- **Nothing asserted the handler serves the field.** Every test shaped a slice the
  caller handed in, leaving `publishedCaptureSources(ComposedExtensions())`
  covered by nothing. Now tested through the real composed set and the real
  handler, reading response bytes rather than the struct — and verified failing
  with that line removed.
- **The contract claimed a completeness it does not have.** A core connector's own
  source system (`gmail`, `imap`, `graph`, `gcal`) is in neither array; a client
  treating a directory miss as an anomaly would report the most common trace row
  in the product.
- **The authorization argument defended only half of itself** — label provenance,
  not the widened set. Both halves stated, including the one genuinely new thing:
  a capture-only unit becomes visible to a seat holding no grant on it.
- Pattern check derived from the grammars boot admits rather than sampled from
  three hand-written units; tests moved to the files their subjects live in; three
  comments that narrated the change now state the invariant.

**One finding I did not act on, with the reasoning:** both reviewers flagged that
`unitChannelFacts` refuses a unit declaring a reserved core transport while the
ingress path has no analogue. The analogy breaks on what the two declarations
claim. `Channels` says *"I transmit on this"*, and a unit taking `telegram` there
re-points the core row at the unit's own credential — which is why that one dies
at boot. `IngressSource.System` says *"I bring records in from this"*, and a unit
that ingests Telegram exports naming it `telegram` is telling the truth. A boot
refusal would break the honest case to stop a reviewed first-party unit from lying
about what it captures. The shared label is then correct rather than a collision —
the record really did travel on that transport — which is the whole point of the
change, and it is now written down where the next reviewer will ask.

**Recorded, not fixed:** the agent surface (`channelProviderDirectory`) does not
carry `capture_sources`, so a passport reading REST resolves a unit's `ext:`
provenance and the same passport's tool listing does not. Nothing is governed
differently — both are the one read-scope operation and an unresolved id falls
back to itself — so an agent loses a display string, not an answer. It is named in
that type's doc comment beside the `attachments` divergence it already documents,
because an unnamed divergence is how two surfaces come to disagree about
something that matters.

luit-nfq:
Review round applied. Nine of the ten findings are fixed in `ca892eaaf`; one is declined.

**Fixed, and three of them were real defects rather than hygiene:**

- **`share.tsx` — held grants keyed by `subject_id` alone.** A user and a team that share an id collided on one entry, so the picker could show the wrong access level and read a fresh grant as a reassertion. One helper now spells the composite `subject_type:subject_id` key and every `get`/`has`/`set` goes through it. Two neighbouring sites carried the same identity bug and are fixed with it: the picker row's React key (duplicate keys across kinds) and the `aria-pressed` comparison, which marked the wrong-kind row as pressed.
- **`leadsignals.tsx` — a retry re-wrote factors that had already landed.** The loop stops at the first refusal, `onSuccess` never runs, and the pending answers still held the bands the server took. Pressing save again re-stamped `set_by`/`set_at` on factors nobody touched and appended a history row recording no change. Each answer is now retired as its own write lands. A new test proves it, and it fails against the old code.
- **`leadbulk.tsx` — the mutation read its reason and owner out of render state.** The `BulkAction` is now the mutation's variable, so both the disqualify and the assign path are fixed by the same change.

- **`App.tsx`** — the route key is the `scheduled` literal again, typed against `Screen`. Importing `SCHEDULED_SCREEN` pulled the module into the startup graph and made the `lazy()` beside it load a chunk the browser already had. Nothing else in the startup graph reaches the module now; `tasks.tsx`, which does import it, is itself lazy.
- **`companycontracts.tsx` / `companydocuments.tsx`** — `rec-main` and `rec-end` are `div`s. `SurfaceState` can render a paragraph in the first, and `OverflowMenu`'s root is a `div` in the second, so both were flow content inside a `span`. Both classes are already `display: flex` in the CSS, so nothing moves.
- **`channelproviders.test.tsx`** — the `as never` casts are gone, and they were hiding a real gap: every fixture omitted the `attachments` block `ChannelProviderEntry` requires. The mock is now a `fetch` stub (the house pattern in this suite) against a fixture typed as `ChannelProviderDirectory`.
- **`contractformpaper.test.tsx`** — `CONTRACT` is typed as the generated `Contract` at all four call sites, which also pins `status`, `value_basis` and `source` to their literal unions.
- **`leadbulk.test.tsx`** — the fixture is annotated against the generated `Lead`, so the status literals survive without an assertion.
- **`readings.test.tsx`** — the element is required before its class is read. The `toBe(false)` case previously passed on a `Meter` that rendered no bar at all, which under T11 is a test that can only fail by panicking. Swept the whole file; the two `Sparkline` point assertions had the same shape.
- **`share.test.tsx`** — the DOM cast is replaced by attribute assertions, and the button now passes `pending={grant.isPending}` with `disabled={!subject}`. The design-system `Button` deliberately does not set `disabled` for `pending` — it renders a busy mark, sets `aria-busy`/`aria-disabled` and swallows the click — precisely so focus stays on the control.

**Declined: the em-dash objection.** Nothing in this repository forbids an em dash in user-visible text — there is no gate, no style rule in `CLAUDE.md`, `AGENTS.md`, the design-system README or `.coderabbit.yaml`. The English catalog already carries several hundred, and the house voice uses them deliberately for exactly this construction. Rewriting four new strings would make them the inconsistent ones.

luit-nfq:
Two more, plus a correction to my last comment.

**I was wrong to decline the em-dash finding.** VOICE-RULE-5 is real and it is written down in this tree — the English catalog states it beside `auth.forgotSub` ("VOICE-RULE-5 forbids an em or en dash anywhere in user-facing copy") and cites it again at `app.errorBody`, `auth.resetServerFailed` and `share.linkCopied`. I searched for a gate and a style file, found neither, and concluded there was no rule; the rule lives in the catalog it governs. The three keys this branch adds now read as the sentences they were, in all three locales:

- `share.unchanged` — "Nothing changed. {name} already had {access} access to this record."
- `lead.signalReasonUnstated` — "No source given. Entered by hand."
- `sched.group.held` — "Stopped, waiting on you"

Left alone deliberately: the pre-existing strings the thread also lists (`vi.ts#L1708`, `en.ts#L1716-L1725`), which this branch does not touch, and `lead.signal.employees.*`, where the en dash IS the numeric range rather than a joined clause.

**Focus return on the share dialogs** — fixed, and it applied to both, not just the downgrade. Each destroys its own trigger on success: a revoked grants row leaves the roster, and a reduction clears the picker that opened it. Focus fell to the document body, so a keyboard reader restarted the surface. Both now pass `returnFocusTo` pointing at the subject search field — the one control on the page that is always present — returning the nullable lookup directly, no assertion. A new test asserts `document.activeElement` after a confirmed reduction; I verified it fails when the prop comes off.

Local gates green: `make check` exit 0, `make fe-uat` 165 stories clean.

Issue #1964, opened 2026-08-20
No gate enforces the object half of RBAC in the compose tier, for any core object


tinnhlt-gra:
Found while building #1846 (PR #1961). `relationship` was the instance; this is the class.

## The gap, in the tree's own words

Two gates cover the compose tier and both say explicitly that they do not cover this:

- `backend/rbacgate_test.go` judges only exported methods on a module's `*Store` / `*Service`, and its `entryPointsOutsideModules` set ratifies the compose read services as **subjects it has deliberately not judged**.
- `backend/composerowscope_test.go` covers compose, but for ROW scope only, and refuses to count object admission as satisfaction: *"Object admission (auth.Require) deliberately does not count: both halves of #632 passed auth.Require(ctx, \"deal\", ActionRead) and served a deal the caller could not open, so a gate that accepted it would have read green over the defect it was written for."*

Between them: **the object half of RBAC has no fitness function anywhere in compose**, for `person` and `deal` and `organization` exactly as it had none for `relationship`. `tableownership_test.go` exempts SELECTs with a rationale naming the workspace predicate and the row-scope clauses — a sentence that silently omits the object half.

That is why nine compose reads of one table drifted while a correct implementation sat one directory away.

## What #1846 built that generalises

`backend/edgereaders_test.go` is the shape: derive the SITES from SQL literals in the tree, resolve the OBLIGATION per function transitively across the package, and require a VERDICT (gate, or one of three declared non-disclosing kinds) for every site. It carries three properties worth keeping — a site that is both gated and waived fails, a numeric floor catches extractor collapse, and the subject regex matches the literal's text rather than its quoted token.

Widening it means, per core object: a table→object mapping (`identity/internal/policy.go`'s `coreObjects` is the vocabulary, and `IsCoreObject` already exists for exactly this question), and a discriminator per object for what counts as a predicate read. `relationship`'s discriminator — *the edge determines which records, columns or counts appear* vs *removing the condition could only widen what the caller already sees* — should generalise, but each object needs its own pass to find the circular cases (`activities/orgscope.go` is one: gating the row-scope walk on a grant is circular).

## Also worth doing here

`edgereaders_test.go` and `restrictedreaders_test.go` are now structural near-twins (regex-over-SQL-literals subject, marker list, waiver sets, `gatekit.Scope`). Factoring the shared walk into `gatekit` belongs in this change, not before it — two copies is a shape, three is a rule.

[...]

PR #2037, opened 2026-08-20
fix(authz): a withheld coverage view says so instead of reading as a clean deal

### LINKED ISSUE #1962: Do partner/parent/child org edges need relationship.read? The contract says no
Split out of #1846 (PR #1961), which gated every other disclosing read of the `relationship` table and left this one alone deliberately.

## The conflict

`org360/graphreads.go` `readRelatedOrganizations` reads the org↔org partner edges (`partner_of`, `referred_by`, `co_sell_with`) and returns `r.kind`, under `organization.read` alone.

By #1846's own rule that is a disclosing read and should carry `relationship.read`. **But the contract says the opposite.** `backend/api/crm.yaml`, on `OrganizationGraph.groups_omitted`:

> The parent, child and partner organizations are not a group here: they need no grant beyond the organization read this whole endpoint already demands, so they are row-scope pruned like every other node and **can never be withheld wholesale**.

And `groups_omitted` is a closed enum — `[contacts, deals, intro_path, our_side]` — with no value that could name them. So there is no way to withhold the group honestly today: dropping it silently renders "Nothing linked to this account yet", which is a lie by the page's own standard.

## Why this is a ruling, not a fix

CLAUDE.md says to name a conflict with a guardrail rather than silently pick a side. The contract is a deliberate statement about what a partner edge discloses, and #1846's rule is a deliberate statement about what an edge discloses. They disagree.

It is also #1833's neighbourhood — org↔org partner edges are already known-untraversable in the query grammar, for a different reason (the role-named `counterparty_org_id` column).

## The options

1. **Keep the contract's posture.** A partner edge is a fact about two organizations the caller may already read, and unlike an employment edge it names no person. Cheapest, and defensible: the sensitivity argument for `relationship` is mostly about disclosing PEOPLE. Cost: one read of the table stays outside the rule, permanently, and the census carries it as a standing exception.
2. **Gate it, and pay the contract change.** Add a `related` value to `groups_omitted`, rewrite the description, add i18n in en/de/vi, and guard `connections.tsx:606,663`. Cost: a contract addition and a frontend change for a case no default install hits.

**Recommendation: option 1**, with the census entry rewritten from `deferred` to a permanent `predicate`-style verdict stating the ruling. The edge grant exists because an edge discloses its endpoints as a pair; a partner edge's endpoints are two organizations this endpoint already required the grant for, and no person is named. But this is a product call, not mine.

Tracked in the census as `internal/compose/org360/graphreads.go:readRelatedOrganizations` in `deferredEdgeReads` (`backend/edgereaders_test.go`), so it cannot be forgotten either way.

### LINKED ISSUE #1963: DealCoverage and deal health need a withheld channel before their edge reads can be gated
Split out of #1846 (PR #1961). Three disclosing reads of the `relationship` table stay ungated, and the reason is the same for all three: **gating them today would replace a disclosure with a wrong number.**

## The sites

- `compose/network/coveragefacts.go:readDeparted` — who LEFT the account
- `modules/deals/engagement.go:EngagedStakeholders` and `:Stakeholders`
- `modules/deals/health.go:healthActivityEvidence`

Each carries a `deferred` verdict in `backend/edgereaders_test.go`.

## Why gating them is worse than the defect

Departed stakeholders surface as `champion_left` / `stakeholder_left` risks. `DealCoverage` has **no withheld channel** — `required: [deal_id, stakeholders, our_side, risks]` — and an empty risk list renders as:

> Nothing flagged — this deal passes every coverage check

So a restricted operator would get a confident **wrong verdict** instead of an omission. The health composite is the same shape one layer down: its engagement factor is computed from seat edges, so a caller refused the edge gets a silently **lower score**, not an absent one. A number that is wrong is worse than a number that is missing.

MCP makes it wider: `account_coverage`, `whats_slipping_this_week` and `catch_me_up_on` all read this, and `compose/riskretriever.go:70-73` swallows `ErrPermissionDenied` with no marker at all.

## What is needed first

1. A withheld channel on `DealCoverage` — the `groups_omitted`/`sections_omitted` shape the 360 payloads already use, so a caller can be told "risks were withheld" rather than shown none.
2. The same for the health payload's factor breakdown, or a rule that a factor computed from a withheld input makes the composite absent rather than lower.
3. `riskretriever.go`'s unmarked swallow named, not extended.

Then the four reads take `auth.EdgeReadScope` like their neighbours, and the census entries move from `deferred` to gated. The census fails a site that is both gated and waived, so the verdicts cannot be left stale once this lands.

## Severity

`priority: high` and not `normal`, deliberately. #1846's own argument applies more strongly here now: an operator who restricts the grant sees it honoured on the account page, the roster, the contact count and the coworker graph — and these four surfaces keep answering. Priority is a claim about severity, not about schedule.

### LINKED ISSUE #1977: connected-agents.test.tsx was a time bomb: it began failing by the calendar, and main hid it

tinnhlt-gra:
Closes #1963. Closes #1962. Closes #1977. Part of #1964.

The four follow-ups #1961 filed, in one branch — they touch the same census file and would have conflicted as separate PRs.

## #1963 — a withheld coverage view says so instead of reading as a clean deal

Four reads of the `relationship` table carried a `deferred` verdict for one reason: gating them replaced a disclosure with a **wrong number**. An empty risk list renders *"Nothing flagged — this deal passes every coverage check"*, and a false all-clear on deal risk is worse than the pair it stopped disclosing. The contract work had to come first.

`DealCoverage` gains `sections_omitted`, `required` and present-and-empty. **All three sections together** — `our_side` is derived from the seats, every risk rule but `going_cold` reads them, so there is no partial answer to give. **A named section is empty, never partial**: `going_cold` needs no edge and could have survived, but a `risks` array holding one finding under a name that says it was withheld leaves a client unable to say whether the list is complete.

`CoverageFor` takes the edge admission **first, before any statement**, and that is the one place a denial becomes an omission. Three catch points would be three chances to convert one into a wrong number instead of a named absence.

### The filed shape changed on one point

`Store.DealHealth` has **no production caller** — `grep` finds only its unit and integration tests. So #1963's item 2 (a withheld channel for the health payload) is not needed: there is no health payload on the wire. Gating the read outright gives exactly what the issue asks for — *a factor computed from a withheld input makes the composite absent rather than lower* — because the engagement factor is a count of edges over a norm, and a refused caller now gets no score rather than a quietly smaller one. It also stopped the health engine carrying its own copy of the engagement query; it calls `deals.EngagedStakeholders`, which is where that definition lives.

### Every consumer carries the marker

Or the withholding becomes a false all-clear one layer out:

| Consumer | After |
|---|---|
| `GET /deals/{id}/coverage` | `sections_omitted`, and the card renders "coverage was withheld" instead of the clean state |
| `account_coverage` | a `section_withheld` warning **in words** — a model handed three empty arrays will otherwise write "the deal looks well covered" |
| the at-risk sweep (`whats_slipping_this_week`) | `coverage_withheld`, the shape `Truncated` already had: the deal is absent from the findings either way, and only the flag stops that absence reading as a clean deal |
| the risk retriever (`catch_me_up_on`) | the withholding is an **item**, not a silently empty section |

`riskretriever.go`'s `ErrPermissionDenied` swallow stays for the deal gate it was written for, and now says so: the edge denial no longer takes that path.

## #1962 — ruled, not deferred

Partner/parent/child org edges **keep the contract's posture**. `crm.yaml` states those organizations need no grant beyond the organization read the endpoint already demands and can never be withheld wholesale, and `groups_omitted`'s enum has no value that could name them. The edge grant exists because an edge discloses its endpoints *as a pair* — and both endpoints here are organizations this endpoint already required the grant for, with no person named.

The census gains a **fourth verdict**, `ruledEdgeReads`. Not `predicate` (removing the condition would not merely widen), not `lifecycle`, and not `deferred` — a deferral says the work is pending, and recording a ruling as one leaves a hole nobody can ever close. `deferredEdgeReads` is now **empty**, which is the state the census was built to reach; it stays declared so the next honest deferral has somewhere to go.

Census after: **23 gated, 4 predicate, 17 lifecycle, 1 ruled, 0 deferred.**

## #1977 — the suite's verdict must not depend on the calendar

The sweep decided the gate's shape. *"An absolute date in a file that never pins the clock"* matches **129 files** here, nearly all harmless: a fixture date only becomes a bomb when the **component compares it to `now`** to decide a state, and nothing static separates those from the dates a component merely formats.

So the gate is a second **run**, not a pattern. `make fe-clock-drift` runs the suite at +200 days and requires the same verdict. That is also the sweep's answer: **3602 tests, all green 200 days out** — no live bomb in the tree today. A drill confirms it catches the original defect: restoring the real clock with an expiry four months out reds the same three tests while a bare `make fe-unit` stays green.

It runs daily on `main` from `scheduled.yml`, not in `ci.yml`, because what breaks these tests is **time passing rather than a diff** — and the classifier's frontend skip is why #1977 sat on `main` for a month. A failing run files an issue like every other check there.

## #1964 — the enabling half only

Several censuses derived their sites the same way, and the copying had already cost something: two carried one two-part flaw (a required trailing space, and matching against the quoted token so the end-of-text alternate could never fire), fixed by hand in #1961. `gatekit` now owns `TableReadPattern`, `LiteralText`, `TableReads`, `DeclReads`, `FileReadsTable`, `CallsAny` and `FirstLineOf`, with the boundary test beside them. Both censuses report **exactly the verdicts they reported before**, and a mutation drill confirms the activity census still reds an unguarded reader.

**The widening stays open.** 460 read sites of row-scoped core tables in the compose tier, and #1964 is explicit that each of the 37 core objects needs its own pass for the circular cases. Measured here, not attempted.

## Two findings worth carrying

- **`auth.EdgeReadScope` was missing from `composerowscope_test.go`'s row-scope spellings.** The first compose read whose only row bound came through it reported as unscoped — and the fix that message invites is a *second* scope call over a column the endpoint conjunction already covers. Adding it also made `briefs:briefEvidenceRows`'s waiver stale, correctly, and the unbundling stays one-directional: this census accepts the pair because the pair includes the row half, while `edgereaders_test.go` still refuses the row half alone.
- **`tableownership_test.go`'s SELECT exemption** named the workspace predicate and the row-scope clauses and stopped, which reads as the list of guards a read *needs*. It now names the object half and where it is unenforced.

## Verification

- `make check-backend`, `make check-fe`, `make craft-static`, `make fe-unit` (3602), `make fe-clock-drift` (3602 at +200 days) — green.
- Targeted integration: the withheld coverage view, the at-risk sweep, the risk retriever, and the pre-existing coverage/health suites — green against a real Postgres.
- **Six refusal tests pass a nil transaction.** That is the assertion: the gate resolves before the statement, so a read-then-filter implementation panics instead of passing.
- **Mutation drills**, not just tests: reversing `dealAtRisk`'s withheld/empty ordering reds the sweep test (that ordering is what would silently reclassify every unassessable deal as healthy); removing `EdgeReadScope` from the row-scope spellings fails exactly two sites; swapping `readDeparted`'s `EdgeReadScope` for `EdgeReadAdmitted` fails the edge census.
- An adversarial security pass found no exploitable path and verified both census claims by mutation. An adversarial craft pass found eight things; two were defects and are fixed in `afc9c5ba2` — the clock-drift lane could pass having shifted no clock (`Number("typo")` is `NaN`, so it silently ran the ordinary suite), and the two producers of the withheld signal were untested while stubs handed the flag in ready-made.

## What end users see

Nothing. All six seeded roles grant `relationship.read`, and custom roles cannot be created through the API. `Require` short-circuits `PrincipalSystem`, so every sweep and capture path is unaffected by construction.

## Filed, not fixed here

`compose/briefs`' `seatEvidenceBound` swallows the same denial and **floors** a brief's warmth factor rather than absenting it — the same wrong-number shape, pre-existing on `main`, and the next instance of this invariant.

---
## Summary by cubic
Marks withheld deal coverage instead of rendering a clean all‑clear when `relationship` edge reads are denied. Adds a frontend clock‑drift lane so date‑dependent tests fail when the calendar breaks them, not weeks later.

- API/UI: `DealCoverage.sections_omitted` is required (present and empty on normal reads). `GET /deals/{id}/coverage` includes it; the card shows “Coverage was withheld.” `DealCoverageSectionsOmitted` is generated and used end‑to‑end.
- Authz: coverage checks `EdgeReadAdmitted`/`EdgeReadScope` before any SQL; partner/parent/child org edges are ruled, not deferred; the deferred set is now empty. Refusal tests pass a nil tx to prove admission resolves pre‑statement.
- Compose/agents: propagate the marker. Agents add `DealCoverageAnswer.sections_omitted` and required `AtRiskReport.coverage_withheld`; the retriever emits a human‑readable “coverage withheld” item. Deal health calls `deals.EngagedStakeholders`; on denial the engagement factor is absent, not lower.
- Tooling/tests: share SQL read detection in `gatekit`. Integration tests prove withheld vs. finding for principals that differ only by the edge grant. Frontend adds a withheld story and test. `vitest.setup.ts` supports `make fe-clock-drift` (+200 days), rejects bad skew, and asserts the shift took; scheduled in `scheduled.yml` with reporting via `scripts/scheduled-report.sh`.

- Migration: regenerate API/types (`frontend/src/api/schema.d.ts`). Treat non‑empty `sections_omitted` as “withheld,” not “clean.” Agent integrations must handle `DealCoverageAnswer.sections_omitted` and required `AtRiskReport.coverage_withheld`. When touching date logic, run `make fe-clock-drift`. End‑user impact: default roles grant `relationship:read`; no change unless that grant is removed.

<sup>Written for commit af4a619a98fd561616d454fa4b0326ac71805ebe. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/2037?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tinnhlt-gra:
The `integration shard (4/12)` failure on the previous run is **#2046, not this branch**: `TestFK_rowScopedTargetsHaveVisibilityDecision` fails on `commission_entry`'s FKs to `deal` and `organization`, a table this PR does not touch. It reproduces on a clean `main`, and #2046 already carries it as `priority: critical`.

Rebased onto `70cf56054` and re-pushed; the shard will stay red until #2046 lands. Everything else in that run was green or still queued.

tinnhlt-gra:
The remaining red check on this PR is **#2067, not this branch**: `uat` fails on `AC-pipeline-7` with a strict-mode violation on `getByText("Fleet retrofit")`, in a spec and a screen this PR does not touch.

I bisected it and added the result to #2067 — it enters at `140eab5fc` (#2032, bulk assign from the deals table), whose new row-select checkbox carries an sr-only `"<deal> auswählen"` label, and it reproduces on a pristine `origin/main` build. Everything else here is green.

Issue #2046, opened 2026-08-20
main is red: commission_entry's FKs to deal and organization have no visibility decision


LarsGradion:
## What is failing

`TestFK_rowScopedTargetsHaveVisibilityDecision` in `backend/migrations/schema_fitness_integration_test.go` fails on `main`:

```
FK commission_entry.deal_id -> deal has no visibility decision: a reference to a
row-scoped record is a read of it — gate it (auth.EnsureLinkTarget) or classify it here
FK commission_entry.partner_org_id -> organization has no visibility decision: ...
```

Reproduced on a clean checkout of `main` with no local changes, so it is not a stale worktree or a shared-template artifact.

## Where it comes from

The `commission_entry` table arrives with the partner-commission work (PR #2019). The gate's rule is that a foreign key to a row-scoped record is a read of that record, so each such FK must either be gated with `auth.EnsureLinkTarget` at its write sites or be classified in the test's own table with a reason.

## Why critical

`main` is red, which makes every other run's verdict unreadable — the honest test for `priority: critical` in CLAUDE.md. It also blocks the integration lane for anybody working near migrations.

## Fix direction

Decide which each FK is: a gated link (add `auth.EnsureLinkTarget` where commission entries are written) or a deliberate exception (classify it in the test with the reason). The test names both options in its own failure message.

Found while running the migrations lane during unrelated deals work.

tinnhlt-gra:
**#2019 broke a second gate, in a different package, and this issue names only the first.** Flagging so whoever fixes it does not go green on one and leave `main` red on the other.

`TestSweepTargetsCarryNoDeleteBlockingTrigger` in `backend/internal/compose/datareset_integration_test.go:350`, reproduced on a clean detached checkout of `origin/main` at `70cf56054`:

```
sweep target "organization" carries DELETE-firing trigger "organization_delete_clears_deal_partner" — an append-only or otherwise protected table belongs in preservedResetTables; a row-conditional guard belongs in deleteGuardedSweepTargets, with what the reset owes it
```

Same origin: `git log -S "organization_delete_clears_deal_partner"` → `648371973` (#2019).

So the partner work added a table and a trigger, and each tripped a different fitness function — one in `./migrations`, one in `./internal/compose`. They are the same omission seen from two directions, and both want the same thing: a decision recorded where the gate can read it, with the reason.

[...]

PR #2049, opened 2026-08-20
fix(tests): the RBAC repair suite finds its era instead of counting to one

LarsGradion:
## What was red

`TestTheRepairMigrationsHealADatabaseThatAlreadyRecordedTheLostBackfill` failed on `main` after the deal lost-reason migration landed:

```
one core step back did not restore role.workspace_id — the newest core migration is no longer
the phase D drop these repairs need rolled back, so this suite is replaying them in the wrong era again
```

## Why

The suite needs the schema in the era *before* ADR-0091 §8 phase D dropped `role.workspace_id`, and it reached that era by rolling core back exactly **one** step. That was true when the suite was written — the drop was the newest core migration — and it stops being true the moment any migration lands on top of it. The failure message predicted this precisely.

## The fix

It steps back until the column reappears, rather than a fixed number of times. The bound (200) exists to fail a runaway loop, not to track how far the drop has drifted.

Verified against a real database: the two subtests that were failing (`core/0192`, `custom/20260806120000`) now pass.

## Also found

`TestFK_rowScopedTargetsHaveVisibilityDecision` is **separately red on clean main** — `commission_entry.deal_id` and `commission_entry.partner_org_id` have no visibility decision. That arrives with #2019 and is unrelated to this fix; filed as #2046.

---
## Summary by cubic
Makes the RBAC repair integration test find its correct migration era by checking for `role.workspace_id` instead of rolling core back exactly one step. This removes brittleness when new core migrations land and fixes the failures seen on main.

- Replaces fixed rollback with a loop that steps back until `role.workspace_id` exists, capped by `maxStepsBackToPhaseD = 200`.
- Adds `roleCarriesWorkspaceID` helper; updates `rollBackThePhaseDDrop` accordingly.
- Test-only change; no production schema or runtime behavior.
- Verified against a real database: previously failing subtests now pass.

<sup>Written for commit 423a309e33c6209e112972ce750d69d73ef1b565. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/2049?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/migrations/rbac_repair_integration_test.go
+++ b/backend/migrations/rbac_repair_integration_test.go
@@ -268,20 +268,45 @@ func repointed(t *testing.T, admin, conn *pgx.Conn) *pgx.Conn {
 // becoming a no-op if the newest core migration is ever something else, which
 // would leave the replay below failing for a reason nobody could read off the
 // error.
+// rollBackThePhaseDDrop returns the schema to the era these repairs belong to:
+// the one where role.workspace_id still exists, because ADR-0091 §8 phase D has
+// not dropped it yet.
+//
+// It steps back until the column reappears rather than a fixed number of times.
+// The drop was the newest core migration when this suite was written, so one
+// step reached it — but every later core migration pushes it further back, and
+// hardcoding the distance made an unrelated migration fail this test with a
+// message about an era it had nothing to do with.
 func rollBackThePhaseDDrop(t *testing.T, conn *pgx.Conn) {
 	t.Helper()
 	core, _ := namespaces(t)
-	if _, err := dbmigrate.Down(context.Background(), conn, core, 1); err != nil {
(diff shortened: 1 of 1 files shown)


LarsGradion:
Superseded by #2027, which landed the same fix in a better shape: it targets the phase D migration by VERSION and fails loudly when that version is gone, where this branch stepped back until the column reappeared. Verified #2027's version passes the two subtests on a real database. Closing unmerged.

PR #2055, opened 2026-08-20
fix(privacy): a sweep waiver ratifies one trigger, not a table (#2046)

### LINKED ISSUE #2046: main is red: commission_entry's FKs to deal and organization have no visibility decision

tinnhlt-gra:
Closes #2046's second half.

**Scope changed after #2052 landed.** A parallel session shipped the
`commission_entry` FK classification while this was open, so that half is gone
from here and main's wording is kept. What remains is the gate #2046 did not
name and #2052 did not touch — plus a hole in that gate found while fixing it.

## The failure

```
sweep target "organization" carries DELETE-firing trigger "organization_delete_clears_deal_partner"
    — backend/internal/compose/datareset_integration_test.go:350
```

Still red on `origin/main`. Same origin as the FK half: `648371973` (#2019) added
the table *and* the trigger, and each tripped a different fitness function, in a
different package. #2046 named only the first.

## The decision

`organization_delete_clears_deal_partner` is `BEFORE DELETE ON organization FOR
EACH ROW` and nulls `deal.partner_org_id` / `partner_attribution`. It refuses no
delete — the same shape as `activity_link` and `relationship`, already classified
as maintenance rather than guards.

But "the sweep deletes those deals too, so the clearing is discarded" is **not**
why it is safe, and the entry does not say that. `resetTargetTables` orders
alphabetically and the sweep retries per pass; `offer.deal_id` is
`ON DELETE RESTRICT` (core/0218:330) and `offer` sorts after `deal`, so an
installation holding one offer defers `deal` to a later pass — and `organization`
is deleted in that same pass with `deal` still populated. The trigger then runs
its `UPDATE` once per organization row, and `idx_deal_partner` is partial on
`archived_at IS NULL` (core/1787021916:53) while the trigger's predicate carries
no such term, so the planner cannot use it: a sequential scan of `deal` per
organization, repeated on every deferred pass.

It is safe because the work is **discarded**, not because it never happens. The
entry records the cost, which is what the gate exists to force someone to write.

## The hole in the gate itself

The gate reports a **(table, trigger) pair** and looked its waiver up by **table**:

```go
if targetSet[table] && !deleteGuardedSweepTargets.Waived(t, table) {
```

So an entry ratifies the *category*: it clears the trigger it was written for
**and every DELETE trigger added to that table afterwards**. Adding `organization`
under the old key would have exempted that table from this gate permanently —
fixing the reported bug by widening the hole.

Proven, not argued. I planted a second trigger on `organization` that raises on
every delete — a real blocking guard, exactly what this gate exists to catch:

| waiver keyed on | plant |
|---|---|
| the **pair** (this PR) | **REFUSED** — `carries DELETE-firing trigger "organization_planted_blocks_delete"` |
| the **table** (main today) | **PASS** — silently admitted |

All four entries are re-keyed to the pair. `gatekit`'s ratchet caught three
trigger names I guessed wrong while doing it, which is that mechanism working.

An independent enumeration during review confirmed the re-key loses nothing:
seven non-internal DELETE-firing triggers exist across `migrations/core`,
`migrations/custom` and `extensions/`, and every sweep-target table among them
has a key.

## Coverage

**Test-only change** — one classification table and one waiver key — so it
produces no measured new-code coverage. The proof is the planted-guard matrix.

## Reviews

Fable reviewed twice (5 findings on the first shape, all fixed; the enumeration
above is from the second). CodeRabbit reviewed with no findings.
**Codex has not reviewed this PR** — `/codex:*` cannot be invoked by an agent.

## Gates

`PASS — 0 blocker, 0 major, 0 minor` on all four trees ·
`TestSweepTargetsCarryNoDeleteBlockingTrigger` **PASS**,
`TestFK_rowScopedTargetsHaveVisibilityDecision` **PASS** on the rebased head.

### DIFF
--- a/backend/internal/compose/datareset_integration_test.go
+++ b/backend/internal/compose/datareset_integration_test.go
@@ -296,6 +296,11 @@ func TestDropResetCustomFieldColumns(t *testing.T) {
 // A row-conditional hold refuses deletes for SOME rows and is not a protected
 // store — preserving it would exempt a table the reset exists to clear. So the
 // second shape is classified here, with what the reset owes it.
+//
+// Keyed "<table> <trigger>", on the PAIR the gate reports, because a table-keyed
+// entry ratifies the category rather than the instance: it clears the trigger it
+// was written for AND every DELETE trigger added to that table afterwards,
+// including one that really does block. Each trigger earns its own line.
 var deleteGuardedSweepTargets = gatekit.Waive(map[string]string{
 	// Not a protected table: the guard refuses a DELETE only while a row carries
 	// `restricted_at`, which is a statutory hold on that one record. Preserving
@@ -308,10 +313,16 @@ var deleteGuardedSweepTargets = gatekit.Waive(map[string]string{
 	// lands, the reset must LIFT the restrictions it is entitled to clear before
 	// deleting, or the sweep aborts on the first held activity. This entry is
 	// that obligation, not a record of one already met.
-	"activity": "a row-conditional statutory hold, not a protected store: preserving it would leave every activity behind on a reset meant to clear them, and no writer can set the restriction yet (#1557) — when one lands the reset must lift before it sweeps",
+	"activity activity_refuse_restricted_mutation": "a row-conditional statutory hold, not a protected store: preserving it would leave every activity behind on a reset meant to clear them, and no writer can set the restriction yet (#1557) — when one lands the reset must lift before it sweeps",
 	// Not guards at all (migration 1787032690).
-	"activity_link": "a clock-maintenance trigger, not a guard: it recomputes the last_activity_at of the records the deleted link reached and refuses no delete; the sweep deletes those records too, so the recompute is discarded with them",
-	"relationship":  "a clock-maintenance trigger, not a guard: it recomputes the employer's last_activity_at and refuses no delete",
+	"activity_link activity_link_last_activity": "a clock-maintenance trigger, not a guard: it recomputes the last_activity_at of the records the deleted link reached and refuses no delete; the sweep deletes those records too, so the recompute is discarded with them",
+	"relationship relationship_last_activity":   "a clock-maintenance trigger, not a guard: it recomputes the employer's last_activity_at and refuses no delete",
+	// Also not a guard (migration 1787226902): BEFORE DELETE ON organization,
+	// it sets deal.partner_org_id and deal.partner_attribution to NULL so a
+	// deleted partner leaves no dangling attribution behind. It refuses no
+	// delete. The sweep clears `deal` too, so the rows it just nulled are
+	// deleted immediately after and the clearing is discarded with them.
+	"organization organization_delete_clears_deal_partner": "a clear-the-reference trigger, not a guard: deleting an organization nulls the partner attribution on deals that named it and refuses no delete; the sweep deletes those deals too, so the clearing is discarded with them",
 })
 
 func TestSweepTargetsCarryNoDeleteBlockingTrigger(t *testing.T) {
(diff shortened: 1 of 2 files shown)


tinnhlt-gra:
## Review round 1 (Fable) — 5 findings, all 5 fixed in e66089fae

Every one was verified against the code before being acted on, and two of them corrected claims my own waiver text made.

| # | Finding | Verified | Fix |
|---|---|---|---|
| 1 | The `organization` reason gave the wrong grounds for safety — "the sweep deletes those deals too" implies the trigger fires against an empty `deal` | `resetTargetTables` sorts alphabetically; `offer.deal_id` is `ON DELETE RESTRICT` (core/0218:330) and `offer` sorts after `deal`, so one offer defers `deal` and `organization` deletes with it populated. `idx_deal_partner` is partial on `archived_at IS NULL` (core/1787021916:53); the trigger's predicate has no such term | Entry states the real cost: a scan of `deal` per organization, discarded rather than skipped |
| 2 | `partner_org_id` classified on the write side only; the read decision missing | `VisibleClause` requires `deal:read` and composes only the deal's scope | Entries now say reads are scoped by the deal, and say plainly what they do **not** settle |
| 3 | The tripwire was phrased around request **bodies**, but a shipped op already names both ids | `ListCommissionEntries` takes both as query params (`handlers.go:38-49`), ANDed after the visibility clause (`list.go:100-105`) | Tripwire removed in favour of naming the gate that actually holds |
| 4 | **The reasons understated the code** — `accrueTx` already calls `auth.EnsureLinkTarget` on both | `accrue.go:91-97`, with its own comment saying why | Re-classified with the map's `"gated: … (H1)"` vocabulary |
| 5 | The block split the A165 comment from the entry it explains | `activity_retention_evidence.deal_id` sat under a comment about commissions | Moved after the finance block |

**Finding 4 is the one that mattered.** I wrote "server-derived, so there is no request for a target probe to check" — into a table whose entire purpose is to tell the next person whether a probe exists. A probe *does* exist, on the shared insert path, so a future writer would have been told to build its own or skip it. A waiver that understates the code is as misleading as one that overstates it, and it is harder to notice because it reads cautious.

Both gates re-verified after all five: `TestFK_rowScopedTargetsHaveVisibilityDecision` **PASS**, `TestSweepTargetsCarryNoDeleteBlockingTrigger` **PASS**.

Gates: `make check-q` → `OK: check-q passed` · `make craft-static` → `PASS — 0 blocker, 0 major, 0 minor` on all four trees.

Finding 2 also surfaced something that is **not** in this PR and is being handled separately and privately, per SECURITY.md. This PR neither introduces nor describes it.

Issue #2058, opened 2026-08-20
main is red: the data reset's sweep gate refuses organization's new partner-clearing trigger


tinnhlt-gra:
## What is failing

`TestSweepTargetsCarryNoDeleteBlockingTrigger` in `backend/internal/compose/datareset_integration_test.go` fails on `main`:

```
sweep target "organization" carries DELETE-firing trigger
"organization_delete_clears_deal_partner" — an append-only or otherwise
protected table belongs in preservedResetTables; a row-conditional guard
belongs in deleteGuardedSweepTargets, with what the reset owes it
```

Reproduced against a branch that touches neither the trigger nor the gate: `git diff origin/main` is empty for `backend/migrations/core/` and `backend/internal/compose/datareset_integration_test.go`.

## Where it comes from

`organization_delete_clears_deal_partner` arrives with the partner-commission work (PR #2019, `git log -S`). The reset's sweep gate requires every DELETE-firing trigger on a sweep target to be classified — either the table is preserved, or the trigger is a row-conditional guard listed in `deleteGuardedSweepTargets` with what the reset owes it. The new trigger is neither.

## Why critical

`main` is red, so every other run's verdict is unreadable — CLAUDE.md's own test for `priority: critical`. This is the second gate #2019 left failing; #2046 is the first (`commission_entry`'s FKs have no visibility decision). They are separate gates in separate packages and need separate decisions, so this is a sibling rather than a duplicate.

## Fix direction

Decide what the trigger is to the reset. It clears `deal.partner_*` when an organization is deleted, which is a row-conditional guard rather than a protection — so `deleteGuardedSweepTargets` with a note on what the reset owes it looks right, but that is the call the gate is asking somebody to make, not one to assume.

Found while running the full integration lane during unrelated test-harness work.

tinnhlt-gra:
Verified fixed at `origin/main` `0a475573f` — closing.

`deleteGuardedSweepTargets` in `backend/internal/compose/datareset_integration_test.go:340` now ratifies the exact trigger this failure named, with its cost written out rather than waved through:

[...]

Issue #1996, opened 2026-08-20
The overlay integration suites make real HTTPS calls to api.hubapi.com


tinnhlt-gra:
`backend/internal/compose/integration/overlay` reaches the public internet on every connect. Its own header comment says it does not.

## What happens

`compose/serveroptions_integration.go:35` substitutes the incumbent resolver on `s.sorDispatch` **only**:

```go
func WithOverlayIncumbentResolver(resolve func(context.Context) (overlay.Incumbent, error)) Option {
	return func(s *Server, _ *pgxpool.Pool) { s.sorDispatch.SetOverlayIncumbentResolver(resolve) }
}
```

while `compose/overlay.go:94,99` hands the **Service** the live factory:

```go
incumbent := overlayIncumbentFactory(backfillLimit)
svc := overlay.NewService(...).WithIncumbentFactory(incumbent)
```

`overlay/connection.go:274` (`fetchPortalID`) and `:307` (`seedUserMapOnConnect`) each dial it during connect. Both are best-effort, so the 401s are swallowed as warnings and the suite stays green while talking to HubSpot.

## Evidence

One `make test-it DIR=backend/internal/compose/integration/overlay` run emits **64** of these:

```
WARN overlay connect: fetching the incumbent portal id failed; webhook binding deferred incumbent=hubspot err="permission denied"
WARN overlay connect: fetching the owners directory to seed mirror_user_map failed  incumbent=hubspot err="permission denied"
```

32 connects, two real round trips each, **~718 ms per connect** against a `curl` baseline of 450-480 ms to the same endpoint.

## Why this is `priority: high` and not just slow

1. **A merge gate depends on the public internet.** On a runner without egress to `api.hubapi.com` these tests stall against `hubspot/client.go:33`'s 30 s timeout rather than failing fast — a nondeterministic required check.
2. **The suite asserts the opposite of what it does.** `overlay_write_shadow_integration_test.go:19-25` states "no mocked provider, no real HubSpot account, **no network call (T11)**". A comment naming a guarantee the code does not hold is worse than no comment: it is what the next reader checks instead of the code.
3. It is 23 s of a 34 s package. Removing the live factory took the package **33.8 s → 9.6 s** in a measured A/B, with the same three pre-existing failures.

`netguard` cannot catch this — it refuses *private* addresses, and a public vendor host is exactly what it permits.

## Suggested fix

[...]

Issue #1744, opened 2026-08-18
49 integration suites open their own pool, outside the lane's connection ceiling


tinnhlt-gra:
## What is true after #1109

The integration lane now declares a connection budget (`scripts/lib-testdb.sh`, `declare_lane_budget`) and hands the harness a per-pool ceiling as `MARGINCE_TEST_POOL_MAX_CONNS`. `testdb.Pool` applies it, so the two pools it keys by DSN — owner and app — are bounded.

## What is not

`MARGINCE_TEST_POOL_MAX_CONNS` reaches `testdb.Pool` and nothing else. **49 integration test files build their own pool** from the same DSNs, through `database.NewPool` or `pgxpool.New` directly, for example:

- `backend/internal/compose/integration/jobfanout/runner_integration_test.go`
- `backend/internal/modules/ai/ratestore_integration_test.go`
- `backend/cmd/worker/boot_integration_test.go`

Those keep `database.NewPool`'s fallback `MaxConns = 16` — **and `MinConns = 2`, dialled eagerly**. So a package that opens one of them can hold `8 (owner) + 8 (app) + 16 (its own)` = 32 against a declared per-package allowance of 24, while `TestTheLaneFitsInsideTheClusterItRunsAgainst` reads green.

## Why it is worth fixing rather than absorbing

The allowance is currently a **measured budget** and not a proved ceiling, and the term in `lib-testdb.sh` says so. That is honest but it is the weaker of the two things it could be: the measured per-package high-water mark is 16, so nothing is over budget today, and nothing would notice the day a new suite pushed it over — the gate compares declared numbers, not observed ones.

Closing this makes the lane's central claim — the demand is `jobs × a known number` — true rather than approximately true.

## Options

1. **A harness helper.** Give `testdb` an exported constructor the suites use instead of reaching for `database.NewPool` themselves, so the ceiling applies by construction. Then a fitness function can assert no `*_integration_test.go` calls the raw constructors — the same shape as `check-rls-store-path`, and the direction with teeth: a new suite is enrolled the day it is written.
2. **Widen the term** to `2 × ceiling + 16 + ad-hoc` and provision for it. Correct but wasteful: it sizes the cluster for a worst case only one suite per package can reach, and it does not stop the second such pool.

Option 1 is the one that turns a budget into a ceiling; option 2 only makes the budget bigger.

[...]

PR #2066, opened 2026-08-20
perf(ci-tests): the lane's harness stops paying for itself

tinnhlt-gra:
Four of the six remaining items from the integration-lane cost audit, one commit
each so every claim can be read alone. #1994 was the first half; this is the
harness half.

| | before | after | how measured |
|---|---|---|---|
| `compose/integration/overlay` | 47.78s | **14.23s** | back-to-back, origin/main worktree vs this branch |
| `platform/keyvault` | 6.49s | **1.72s** | package solo |
| `cmd/worker` | 2.86s | **1.51s** | package solo |
| `migrations` | 48.98s | **44.11s** | package solo, interleaved A/B |
| lane dispatch on a fresh clone | discovery order | longest-first | — |

## What each commit claims

**`a fresh clone dispatches the lane's long pole first`** — the runner orders
packages longest-first from `.tmp/integration-lane-timing.txt`, which no clone
has until it has already run the lane once; until then the long pole queues
behind short packages (21s of a full run, measured). A baseline from a real run
is committed beside the runner, and the machine's own measurements still win
whenever they exist. The committed file is never written by a run — a shard
measures a twelfth of each package, and a machine's durations belong to that
machine.

The ordering moved to `scripts/lib-laneorder.sh` with `make test-laneorder`
beside it, and writing that test found the defect worth having it for: **the
sort was not stable**, so every unnamed package shared one key and `sort -r`
broke those ties in *reverse* input order. A stale baseline was inventing an
order rather than degrading to the one it would have had with no hint at all.

**`two waits stop being the way a test states its claim`** — both of these
passed by letting a clock run out, so both paid the full wait on every green
run. keyvault's control asserted "Get takes a connection of its own" by
deadlocking it against a single-connection pool and watching a 2s deadline
expire — 63% of the package, spent proving a negative. The pool already counts
acquires, so the claim is now made directly and there is no clock left in the
assertion to be slow or contended. cmd/worker's redis arm dialled a dead port
and paid go-redis's retry ladder to re-refuse a refused connection three more
times.

**`the head-catalog tests stop rebuilding the schema to read it`** — ten tests
in `backend/migrations` only read the head catalog and were each paying a
`DROP SCHEMA CASCADE` plus a full replay to reach a schema the lane had already
handed them. `headSchema` builds head once per process and verifies it before
each later use.

It compares a **digest of the catalog, not the ledger's version set**. The
ledger records a version, not a checksum, so it answers "which migrations were
applied" and cannot answer "is the schema still what they build" — a rewinding
suite that drops an index leaves it reading head. The digest covers columns,
types, nullability, constraints and index definitions, which is the object class
these tests assert on.

The tagged test files move to `package migrations_test` to reach
`platform/testdb` for the data reset: testdb imports migrations, so the internal
test package would be a cycle. `.go-arch-lint.yml` gains `migrations` in its own
`mayDependOn` — a dependency no production file can express, since Go forbids a
package importing itself.

**`the integration lane stops calling api.hubapi.com`** — closes **#1996**.
Every overlay suite connects an overlay, and Connect resolves the incumbent's
portal id and owners directory through the Service's own factory, which
`WithKeyvault` wires to a live HubSpot adapter. Both calls are best-effort, so
the vendor's 401 came back, was logged as a warning, and the suite went green
while a merge gate depended on a third party being up. 32 connects, two round
trips each, ~718ms per connect.

`WithOverlayIncumbentResolver` could not reach it — it substitutes the read
dispatch's resolver, not the Service's factory — and widening it would not have
been enough either: options run in order and `WithKeyvault` *builds* the
handlers when it runs, so an option applied after it sets a field nobody reads.
So the concrete incumbent is split by build tag, as compose already splits
`WithOverlayIncumbentResolver` itself.

## Review found four defects, all of them in the gates

A Codex pass over the branch before it was pushed. Every finding was real and
every one was in added machinery, not in the change — the same shape as #1994,
where thirteen out of thirteen were.

1. **`lib-laneorder.sh` read an unrecorded duration as ZERO**, which sorts that
   package LAST — the exact schedule the ordering exists to avoid. Reachable:
   the lane's own report has a branch for a package that "recorded no test
   duration", and such a line reaches the hint as `package|`. A duration that is
   not a positive number is now no duration at all, so it sorts first with the
   other unknowns. Mutation-tested.
2. **`headSchema`'s digest was blind to three of the four object classes its
   readers assert on** — view options (`security_invoker`), generated-column
   expressions and defaults, and triggers. And its comment claimed the coverage
   it lacked, which is worse than no digest: it *reads* as verification. The
   digest now covers all of them, and there is a subtest per class, each
   deriving its own target from the live catalog. Mutation-tested by narrowing
   the digest back: three of four subtests fail.
3. **The refusing incumbent's error read as one object class's data defect.**
   `jobs_overlay.go`'s `isConnectionLevelIncumbentError` decides whether a
   reconcile sweep aborts or carries on, so a sweep where EVERY incumbent call
   refused would have logged each refusal and then recorded itself a success — a
   green sweep that did nothing, the precise false green this build exists to
   remove. The refusal now wraps `hubspot.ErrUnreachable`, which is also simply
   what happened, and all eleven methods are pinned.
4. **`WithHTTPClient` walked straight around the egress gate**, and the first
   version of the comment *rationalized* that as by-design. It is applied after
   every Option now, and it guards a nil transport too (net/http falls back to
   `DefaultTransport`, which dials). A caller's own RoundTripper is still left
   alone — it reaches no socket — and that exemption has its own test. The
   mutation is stark: the pre-review shape takes **105 seconds** of real dial
   attempts where the fixed one takes 0.2.

## The gates, which is where to look hardest

Every defect on #1994 — thirteen across three review passes — was in machinery
added to make a change safe, never in the change. This branch adds four such
pieces, so each is called out:

- **`make test-laneorder`** — mutation-tested against a non-stable sort (fails 3
  assertions) and against accepting a malformed duration. A speculative awk
  change was *deleted* because its mutation **passed**: an unrecognised hint key
  is simply never looked up, so comment-skipping was unobservable.
- **`headSchema`'s digest** — mutation-tested in both directions: a helper that
  always rebuilds, and a digest narrowed to what it covered before review.
- **the egress dialer** — `Proxy` on that transport is deliberately nil. With
  `ProxyFromEnvironment`, a machine carrying `HTTPS_PROXY` would dial the proxy,
  the gate would see the proxy's address, admit it if it were loopback, and the
  proxy would fetch the vendor on the test's behalf. A gate a stray environment
  variable routes around is not one. Every transport shape a client can arrive
  with has its own subtest.
- **the refusing incumbent** — two claims that were resting on the accidental
  401 now stand on their own: `overlay_write_shadow`'s "no network call" said so
  and was false, and the connect path's tolerance of an unreachable incumbent
  had no test at all. Mutation-tested against a refusing incumbent that answers.

## What is NOT here, and why

**D6 (module suites take the shared pool) needs D5 (`EnsureSchema` probes the
ledger) first, and the audit did not say so.** `testdb.Pool` returns
`ErrSchemaNotReady` until `EnsureSchema` has run, and the module suites
deliberately never migrate — they ride the pre-migrated clone. Forcing
`EnsureSchema` on ~14 packages adds roughly what D6 saves. Doing D6 the cheap
way instead gives up both things that make it worth having: the
`AssertPoolsQuiesced` pairing (a shared pool converts per-test blast radius into
per-package) and bringing those pools under the lane's connection pin, which
`lib-testdb.sh` names as what would turn its *measured* budget into a proved
ceiling (#1744). They ship together, next.

## main was red on arrival — three times, none from here

`git diff origin/main` was empty for every file involved. All three are fixed
now and this branch is rebased onto all of them, so CI here is green.

| | found how | fixed by |
|---|---|---|
| `TestFK_rowScopedTargetsHaveVisibilityDecision` | already filed as #2046 | #2052 |
| `TestSweepTargetsCarryNoDeleteBlockingTrigger` | filed from this branch's lane run as **#2058** | #2055 |
| `AC-pipeline-7: board↔table` | filed as **#2067** | **#2068**, from this session |

The third is the one worth keeping. It had been red since #2032 and nobody could
see it: `ci.yml:183` puts the root `Makefile` in the FRONTEND path filter, so the
AC lane opens for a backend PR that happens to edit it and skips for one that
does not. That is why a backend/harness PR is what surfaced a frontend
regression, and why "the uat lane is green" on a typical PR means "the uat lane
did not run".

## Verification

- `make check-backend` green
- `make craft-static` — 0 blocker / 0 major / 0 minor across backend, extensions,
  fixtures and desktop
- `make arch-lint` clean
- full `make test-integration`, and every touched package re-run individually
  after each rebase
- every timing figure is a per-package A/B on this machine. The lane's absolute
  wall clock is **not** cited: two full runs this session came in at 200s and
  274s on the same tree, and the first lost seven packages to `tls error: EOF`
  when a parallel session's lane shared the container's 408 connections.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ SEED_DSN ?= postgres://margince_owner:dev@localhost:15432/margince
 # every company renders as a placeholder initial.
 MINIO_PORT ?= 29000
 
-.PHONY: help install ai-routing-local dev-fresh check check-backend check-q check-go check-gates check-fe build test test-v test-cover test-integration e2e-siteread e2e-ai e2e-ai-report ai-probe test-db-up test-it test-integration-serial bench-perf bench-perf-check bench-record bench-capture perfdoc lint arch-lint vet gen gen-workflow mcp-apps-vocab gen-types gen-types-check drift composition check-composition test-extensions db-up db-init db-wait migrate migrate-up migrate-down migrate-create run psql redis-cli tidy dev dev-stop dev-logs clean vuln tools tools-go infra-up infra-down infra-logs infra-reset seed-dev seed-dev-db seed-demo verify-demo seed-reset verify-boot frontend-check frontend-e2e bench-mobile perfdoc e2e-company fe-install fe-typecheck fe-typecheck-composed fe-lint fe-build fe-preview fe-format fe-test fe-test-ext fe-ds-gates fe-drift fe-unit fe-quality fe-bundle fe-storybook ds-purity font-lock icon-lint ds-spacing space-tokens native-controls ext-imports fitness-jurisdiction storybook fe-uat craft-static craft-test craft-residue check-craft-doc test-golangci-guard test-scheduled-report test-ci-verdict secret-scan test-secret-scan test-dev-dsn test-api-entrypoint check-image-pins check-host-ports ci-doc-parity make-target-parity check-ext-migrations contract-breaking-check contract-frontend-drift test-contract-frontend-drift migration-versions test-lanes env-reads gofmt lint-modules go-file-length rls-store-path no-jurisdiction pkg-freeze hooks sbom sbom-normalize sbom-supplement sbom-parity sbom-validate sbom-sign sbom-check
+.PHONY: help install ai-routing-local dev-fresh check check-backend check-q check-go check-gates check-fe build test test-v test-cover test-integration e2e-siteread e2e-ai e2e-ai-report ai-probe test-db-up test-it test-integration-serial bench-perf bench-perf-check bench-record bench-capture perfdoc lint arch-lint vet gen gen-workflow mcp-apps-vocab gen-types gen-types-check drift composition check-composition test-extensions db-up db-init db-wait migrate migrate-up migrate-down migrate-create run psql redis-cli tidy dev dev-stop dev-logs clean vuln tools tools-go infra-up infra-down infra-logs infra-reset seed-dev seed-dev-db seed-demo verify-demo seed-reset verify-boot frontend-check frontend-e2e bench-mobile perfdoc e2e-company fe-install fe-typecheck fe-typecheck-composed fe-lint fe-build fe-preview fe-format fe-test fe-test-ext fe-ds-gates fe-drift fe-unit fe-quality fe-bundle fe-storybook ds-purity font-lock icon-lint ds-spacing space-tokens native-controls ext-imports fitness-jurisdiction storybook fe-uat craft-static craft-test craft-residue check-craft-doc test-golangci-guard test-scheduled-report test-ci-verdict test-laneorder secret-scan test-secret-scan test-dev-dsn test-api-entrypoint check-image-pins check-host-ports ci-doc-parity make-target-parity check-ext-migrations contract-breaking-check contract-frontend-drift test-contract-frontend-drift migration-versions test-lanes env-reads gofmt lint-modules go-file-length rls-store-path no-jurisdiction pkg-freeze hooks sbom sbom-normalize sbom-supplement sbom-parity sbom-validate sbom-sign sbom-check
 
 # Bare `make` lists every command instead of running the first target.
 .DEFAULT_GOAL := help
@@ -74,7 +74,7 @@ ai-routing-local:
 ## #1639 is about: a backend-only author never runs the lane that would
 ## otherwise catch a stranded frontend schema. On a pull request CI covers the
 ## same ground from the other side, through fe-quality's fe-drift.
-check-backend: check-craft-doc craft-test test-golangci-guard test-scheduled-report test-ci-verdict check-image-pins check-host-ports ci-doc-parity make-target-parity contract-breaking-check contract-frontend-drift test-contract-frontend-drift migration-versions test-lanes env-reads gofmt lint-modules go-file-length rls-store-path no-jurisdiction pkg-freeze
+check-backend: check-craft-doc craft-test test-golangci-guard test-scheduled-report test-ci-verdict test-laneorder check-image-pins check-host-ports ci-doc-parity make-target-parity contract-breaking-check contract-frontend-drift test-contract-frontend-drift migration-versions test-lanes env-reads gofmt lint-modules go-file-length rls-store-path no-jurisdiction pkg-freeze
 	$(MAKE) -C backend check
 
 ## check — the full merge gate: backend + frontend
@@ -557,6 +557,14 @@ test-scheduled-report:
 ## the real thing on every dashboard.
 test-ci-verdict:
 	@./scripts/test-ci-verdict.sh
+
+## test-laneorder — prove the integration lane still dispatches its long pole
+## first. The order is only a scheduling hint, so a regression here never turns
+## a lane red: it makes the run longer and says nothing, which is why it needs a
+## test of its own rather than the lane's own verdict. Also pins the committed
+## baseline a fresh clone orders by, which has no measurements to fall back on.
+test-laneorder:
+	@./scripts/test-laneorder.sh
 
 ## check-image-pins — every `uses:` in .github/workflows/ AND every container
 ## `image:` (workflow service containers + infra/docker-compose.dev.yml) is
--- a/backend/.go-arch-lint.yml
+++ b/backend/.go-arch-lint.yml
@@ -218,8 +218,13 @@ deps:
     # emersion: the in-memory IMAP server the standing-connect transport
     # test drives; production compose reaches IMAP only through capture.
     canUse: [pgx, chi, oapi, river, xtext, yaml, emersion, kinopenapi]
+  # migrations depends on itself, which no production file can express: Go
+  # forbids a package importing itself. It is reachable only from the external
+  # test package this directory's tagged suites live in (package
+  # migrations_test), which they need in order to reach platform/testdb — and
+  # testdb imports migrations, so the internal test package would be a cycle.
   migrations:
-    mayDependOn: [platform]
+    mayDependOn: [platform, migrations]
     canUse: [pgx]
   # The root fitness tests walk the authoritative contract document
   # (api/crm.yaml) directly — the RD-AC-7 negative-scope gate reads the
--- a/backend/cmd/worker/observe_integration_test.go
+++ b/backend/cmd/worker/observe_integration_test.go
@@ -146,7 +146,13 @@ func TestEachDeclaredReadinessCheckCanActuallyFail(t *testing.T) {
 		// unlike closing the shared test client it leaves no other test's
 		// dependency broken. Without this arm the redis check could be deleted
 		// outright and every other test here would stay green.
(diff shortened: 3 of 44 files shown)


tinnhlt-gra:
@coderabbitai review

tinnhlt-gra:
## UAT — every claim run live, before against after

Run on one machine against one Postgres, with **`origin/main` checked out into a
second worktree** so each before/after pair is the same database minutes apart.
That worktree is the point: the numbers below are not "what it costs now" versus
"what a document said it used to cost".

### 1. D2 — the overlay suite stops calling api.hubapi.com (#1996)

| | package wall | connect-path warn lines |
|---|---|---|
| `origin/main` | **47.78s** | 64 |
| this branch | **14.23s** | 66 |

What the connect path was told, before:

```
level=WARN msg="overlay connect: fetching the incumbent portal id failed; webhook binding deferred" incumbent=hubspot err="permission denied"
```

That `permission denied` is HubSpot's own 401, coming back over the public
internet, 32 connects × 2 round trips per run. After:

```
err="compose: the integration build binds no live incumbent — this suite reached one;
     pass your own through WithOverlayIncumbentResolver (applied AFTER WithKeyvault)
     if it needs incumbent behaviour: hubspot: could not reach HubSpot"
```

**The warn lines do not go away, and an earlier version of this PR said they
did.** Both call sites still log a best-effort failure — 64 before, 66 after,
the two extra from the connect test this branch adds. I had only ever run that
grep against the AFTER state, with a pattern (`WARN overlay connect`) that
cannot match slog's `level=WARN msg="overlay connect:` either way. An after-only
zero proves nothing. Corrected in the commit message and the PR body.

What actually changes is what a warn costs and what it says.

### 2. The egress gate, over every transport shape a client can arrive with

```
--- PASS: TestTheIntegrationBuildRefusesToDialOffThisHost
    --- PASS: .../the_client_NewClient_builds_itself
    --- PASS: .../a_caller's_client_with_no_transport_of_its_own
    --- PASS: .../a_caller's_client_carrying_a_real_transport
    --- PASS: .../a_caller's_transport_with_its_own_TLS_dialer
--- PASS: TestTheEgressGateLeavesACallersOwnRoundTripperAlone
--- PASS: TestTheEgressGateStillAdmitsALoopbackTestServer
```

The last three arms are review findings. `WithHTTPClient` used to walk straight
around the gate, and so did a transport carrying `DialTLSContext` — which
net/http calls **instead of** `DialContext` for https, the scheme the vendor is
reached over. Mutation-tested, and the mutation is the clearest number here:
the pre-review shape takes **105 seconds of real dial attempts** where the fixed
one takes **0.2s**.

### 3. D8 — two tests stop passing by waiting out a clock

```
--- PASS: TestLocalGetOnReadsThroughTheCallersTransaction (1.70s)     was 6.01s
--- PASS: TestEachDeclaredReadinessCheckCanActuallyFail/redis (0.42s) was 1.77s
```

### 4. D7 — headSchema rebuilds for every class its digest claims to watch

```
--- PASS: TestHeadSchemaDoesNotRebuildADatabaseAlreadyAtHead
--- PASS: TestHeadSchemaRebuildsWhateverTheDigestClaimsToWatch
    --- PASS: .../an_index_is_dropped
    --- PASS: .../a_trigger_is_dropped
    --- PASS: .../a_view_stops_being_security_invoker
    --- PASS: .../a_trigger_function's_body_is_replaced
    --- PASS: .../a_grant_is_revoked
    --- PASS: .../a_column_loses_its_default
```

Six classes because review found the digest blind to four of them while its own
comment claimed the coverage. Narrowing the digest back to the pre-review shape
fails three of six; dropping just the two newest arms fails exactly those two.

### 5. D9 — the dispatch order and its gate

```
test-laneorder: OK
./internal/compose/integration|255.831
```

### Known red

`uat (AC screens …)` fails here on `AC-pipeline-7`, which is **`main`'s**
regression since #2032, not this branch's — reproduced on an unmodified
`origin/main` checkout (1 failed / 93 passed). Filed as #2067 and fixed in
**#2068**, whose own `uat` lane passes. This PR rebases once that lands.

tinnhlt-gra:
## UAT recording — 2m16s, every section run live

https://github.com/user-attachments/assets/380e97b7-49b2-4ba7-99c9-a7201e2498ac

The same script as the comment above, recorded end to end. Nothing is replayed
from a log: each command runs against a real Postgres, and the "before" half
runs in a second worktree checked out at `origin/main`.

Three things worth watching for:

1. **The before/after pair at the top.** `origin/main` is the vendor round
   trips; this branch is the same suite answering locally. The absolute seconds
   differ from the PR body's table — 38.9s → 11.9s in the recording, 47.78s →
   14.23s there — because the machine was under different load. That is exactly
   why the claim rests on a back-to-back pair rather than on a wall-clock
   number.
2. **`connect-path warn lines: before 64 after 66`.** The warns do not go away,
   and the narration in the recording says so out loud. An earlier version of
   this PR claimed they dropped to zero; I had only ever run that grep against
   the AFTER state, with a pattern that cannot match slog's format either way.
   An after-only zero proves nothing.
3. **Six subtests under `TestHeadSchemaRebuildsWhateverTheDigestClaimsToWatch`**,
   one per object class the digest claims to watch. Four of the six exist
   because review found the digest blind to them while its own comment claimed
   the coverage.

tinnhlt-gra:
@coderabbitai review

tinnhlt-gra:
@coderabbitai review

tinnhlt-gra:
@coderabbitai review

tinnhlt-gra:
## CodeRabbit never reviewed this PR, and its check still says `pass`

Recording this rather than letting the green check speak for itself. The
`CodeRabbit` status on this PR reads **pass** with the description **"Review
rate limited"** — there are zero reviews and zero inline comments on it. The
account's review quota was spent on #2068 earlier in this session; I
re-requested with `@coderabbitai review` three times across two full quota
cycles (~1h15m) and got "Action not completed / Review rate limited" each time.

A green check over a review that never ran is exactly the failure mode this PR
is about, so it should not be discovered later from the check list.

What stands in its place:

- **Two independent adversarial reviews before the first push** — Codex, then
  Fable — with **10 findings, every one real and every one in machinery added to
  make a change safe** rather than in the four changes themselves. All fixed;
  details in the PR body.
- **Every gate mutation-tested in the shape that would actually slip through**,
  not the shape that is easy to plant: a non-stable sort (fails 3 assertions);
  the digest narrowed to its pre-review coverage (fails 3 of 6 subtests) and
  then to its pre-Fable coverage (fails exactly the 2 newest); a `Get` that
  answers without touching the pool; a bare refusal sentinel (all 11 methods
  flagged); a method dropped from the refusal census; and the egress gate
  applied before the Options instead of after — that last one takes **105
  seconds of real dial attempts** where the fixed shape takes 0.2s.
- **Full CI green**, including all six integration shards and the `uat` lane.
- **A UAT recording** in which the "before" half runs against a second worktree
  checked out at `origin/main`.

Happy to hold for CodeRabbit if a reviewer would rather wait for it — say so and
I will re-request rather than merge.

Issue #2069, opened 2026-08-20
What should a re-bootstrap do with a margince.yaml that disagrees with the stored identity?


tinnhlt-gra:
#863 had two halves. The visibility half shipped in #2071: a bootstrap that
discards the identity it was given now says so, in the log, naming the keys. This
is the half that needs a ruling.

## The situation

`setting` is not tenant-scoped. A bootstrap over a database that still holds a
previous installation's rows creates a new `workspace` beside them, and
`ON CONFLICT DO NOTHING` keeps the OLD `installation.name`, `installation.timezone`
and `installation.base_currency`. The operator's `margince.yaml` is read,
validated, and dropped.

Since the workspace columns that used to carry these values were removed, there
is no second copy to diverge from — the only evidence is now the warning #2071
adds.

## The options

| | Behaviour | Cost |
|---|---|---|
| **A. Keep discarding, keep warning** | today's behaviour, now visible | An operator who edits the file and restarts still gets nothing, and has to read a log to find out |
| **B. Overwrite** (`DO UPDATE`) | the file wins | Breaks the rule the clause exists for: a restart would silently revert a value a human changed in the UI. Distinguishing "a restart" from "a re-bootstrap" is the whole difficulty |
| **C. Refuse** | boot fails when the file disagrees with stored identity | Safest, and the most disruptive — an installation that was fine yesterday stops booting because someone tidied a config file |
| **D. Explicit reset flag** | discard only when asked | Needs a new operator-facing switch and a story for what it resets |

**I recommend D**, with A as the interim — the warning is already there, and a
flag makes the destructive reading opt-in rather than inferred from a file edit.
But this is a product call about first-run behaviour and I have not made it.

## Why it is not just this

It is entangled with **#1268**'s restructure of the first-run flow. Whatever is
decided here should be decided as part of that, not bolted onto the seed path.

Filed from #2071, which deliberately scoped to visibility only.

dungbanguyen:
## Decision, 2026-09-02: option D — an explicit reset flag — with A as the interim, designed inside the first-run restructure

Re-triaged during the `status: needs-decision` sweep. Taking the ticket's own
recommendation.

### The ruling

**D: discard the stored identity only when an operator explicitly asks. A stays as the
honest interim.**

The decisive argument is against **B**, and the ticket states it exactly: overwriting
*"breaks the rule the clause exists for — a restart would silently revert a value a
human changed in the UI."* Editing a config file must never be a way to overwrite what
[...]

Issue #1268, opened 2026-08-14
Restructure first-run onboarding: setup-token claim flow, then cold start


dungbanguyen:
## Problem

An installation's first admin is created from a credential the operator writes into `margince.yaml`, consumed once at boot. That works for unattended provisioning and has no claim race, but it has three costs:

- The operator chooses the admin's password, so the human who ends up using the account never set their own credential, and nothing forces them to change it (#1261).
- The plaintext credential is materialized on the container filesystem, and the deploy entrypoint re-writes it on every start rather than only the first (#1252).
- There is no guided first-run. The admin lands in an installation whose organization identity was decided by whoever edited the YAML, and the company profile step is disconnected from account creation.
- **The account with the highest permission is the only one whose email nobody confirmed.** Every user invited afterwards proves their mailbox by following the link that sets their password. The first admin proves nothing: the operator asserted an address, and password reset delivers there from then on.

Two things are **not** wrong today and are not being changed by this work, stated here so they are not re-opened during implementation:

- The first admin **is** a real `app_user` — password hash, `status='active'`, `is_agent=false`, and a real `role_assignment` to `admin`. It is not a synthetic principal. The non-authority identity in the tree is the Agent Runner seat, which deliberately holds no password and no role.
- Bootstrap's writes are attributed to a **system** principal rather than a human, which is correct and ratified as B37 (`system:seed`). No human had signed in, so naming one would make the audit record assert something false.

## Proposed flow

**1. Infrastructure.** An operator brings up api, database and Redis. They are not an app user. On first boot against an empty database the api mints a **one-time setup token**, writes it to the server log and to a file, and seeds nothing else.

**2. Claim.** The first visitor gets a form requiring the setup token, plus:

- installation identity — organization name, timezone, base currency
- the admin — email, display name, password

The form states plainly that this is the root account with the highest permission in the installation.

On submit, **one transaction** creates: the workspace, installation identity settings, the admin `app_user`, the system roles, the admin role assignment, the Agent Runner seat, and the configured bootstrap seeds. It runs under the same advisory lock that protects boot today, so two simultaneous submissions cannot both succeed. The token is consumed.

[...]

PR #2070, opened 2026-08-20
fix(platform): a discarded bootstrap seed is reported, not swallowed (#863)

### LINKED ISSUE #863: Re-bootstrap silently inherits the previous installation's identity
Found during review of #862 (the `workspace.{name,base_currency,timezone}` drop), not fixed there.

`identity.seedInstallationIdentity` uses `settings.Seed`, which is `ON CONFLICT (key) DO NOTHING` — correct for bootstrap's consumed-once semantics. But `setting` is non-tenant, so those rows outlive workspace archival.

An operator who archives the installation's workspace and re-bootstraps from a **changed** `margince.yaml` gets a new workspace that silently inherits the **old** installation's name, timezone and base currency. The configured values are discarded without a log line.

Before the column drop, the workspace columns at least held the configured values, so the divergence was visible in the row. There is now no second copy to notice it from.

Suggested fix: log — or refuse — when a bootstrap seed no-ops against a row whose stored value differs from the configured one. A refusal is arguably right for `base_currency`, since deals may already have frozen rates against the stored one.

Refs ADR-0090/A135, ADR-0091 phase 4, #521.

tinnhlt-gra:
Closes #863's visibility half. The product question it also poses is filed as
**#2069** (`status: needs-decision`), cross-referencing #1268 — see the end.

## The defect

`Seed`'s `ON CONFLICT DO NOTHING` is deliberate and stays: a restart must not
overwrite a setting a human has since changed. What the comment was silent about
is the other way to reach that clause.

`setting` is **not tenant-scoped**. `seedInstallationIdentity` runs inside the
same transaction as `INSERT INTO workspace … RETURNING id`, so a bootstrap over a
database that still holds a previous installation's rows creates a **new
workspace beside the old settings** and keeps the OLD identity. Every value the
operator put in `margince.yaml` is read, validated, and dropped. The workspace
columns that used to carry a second copy were removed, so nothing downstream can
notice the divergence.

## The fix, and which shape

`tx.Exec` already returns a command tag saying whether the insert fired. The
information existed and was thrown away.

The goal offered two shapes. **`Seed` reports it** — chosen — rather than having
only the bootstrap path collect the identity trio, because the second covers
three keys and the first makes the invariant hold at **every seed call site**.
That is the repo's own "fix the invariant, not the call site" rule; the trio is
where the defect was noticed, not where it lives.

`platform/settings` acquires **no logger**. It reports up, and the two
composition roots that already hold one warn: `compose/installation.go` for the
configured path and `handlers_setup.go` for the claim path — the claim refuses a
*provisioned* installation, but an archived workspace leaves the settings rows
behind, so the human who just typed a name into the claim form can lose it too.

Two other call sites came with the change:

- **`seedRouting`** gained the same answer, and the adoption line no longer
  claims a binding it did not write: when a row appears between the
  `Unconfigured()` read and the seed — another replica booting alongside — it
  warns instead of announcing an adoption that did not happen.
- **`seedRetentionPosture`** is the one seed that drops the answer, and says why:
  its value is the constant `true`, so a discard can only have preserved a row
  that already says it. Nothing an operator supplied is lost, which is exactly
  what separates it from the identity trio.

## The test

The **second** attempt is the defect; the first insert proves nothing. Both
directions are asserted **separately**, because either alone passes for the wrong
reason — "the value did not change" passes against code that reports nothing, and
"a discard was reported" passes against code that also overwrote.

Each was mutation-proven independently:

| mutation | which assertion fired |
|---|---|
| `Seed` returns `true` always (the `origin/main` behaviour) | **the reporting one**: "reported that it stored a value it did not store" |
| `DO NOTHING` → `DO UPDATE` | **the value one**: "overwrote the stored value … DO NOTHING is deliberate and must stay" |

It lives in the integration lane over a real database, deliberately: the whole
question is what Postgres puts in the command tag when the conflict clause fires,
and a fake `Exec` would be supplying the answer under test.

## Scope, stated

**`ON CONFLICT DO NOTHING` is unchanged.** The goal was explicit that changing it
means leaving the goal, and it is right — which of overwrite / refuse / explicit
flag is correct is a product call, entangled with #1268's first-run restructure.
A silent discard becoming a loud one is progress under every possible ruling and
needs no decision. That decision is **#2069**.

## Ripple

`Seed` and `SeedValue` changed signature, so six seed call sites and — because
the answer is threaded out through `createInstallation` — the
`BootstrapInstallation` and `ClaimInstallation` call sites moved too. Most are
tests taking `_`.

## Coverage · Gates

`PASS — 0 blocker, 0 major, 0 minor` on all four trees.

UAT is the issue's own scenario and is **not** attached: it needs a dev stack, and
`make dev` sweeps every stack on this machine while several agent worktrees are
active. The mutation pair above is the behavioural proof; the boot-level
scenario is what a reviewer should ask for before merge if they want it.

**Codex has not reviewed this PR** — `/codex:*` cannot be invoked by an agent.

---
## Summary by cubic
Make bootstrap seed discards visible. Previously ON CONFLICT DO NOTHING hid whether a seed wrote or was ignored; now callers get a stored/discarded signal and composition roots warn when identity or routing values are dropped.

- `settings.Seed(ctx, tx, def, raw) (bool, error)` and `settings.SeedValue(...) (bool, error)` return whether the insert stored a row; tests cover entry validation failures and insert errors so callers do not misread failures as discards.
- `identity.Service.BootstrapInstallation(...) (wsID, created, discardedKeys []string, err)` and `ClaimInstallation(...) (wsID, discardedKeys []string, err)` surface identity keys the bootstrap/claim could not store; `EnsureInstallation` and `setupClaim` log a warning when discards occur. `setupClaim` now takes a `*slog.Logger`, and routes wire it.
- Routing adoption returns whether the file binding was stored; `ResolveRouting` warns if another replica or a preexisting unconfigured row prevented adoption, and re-reads the stored routing after seeding. An integration test asserts no false “adopted” announcement.
- The retention posture seed ignores the boolean by design (value is always true).
- Adds an integration test proving both directions: the stored value is preserved on a second seed and the discard is reported.

**Migration**
- Update call sites to handle the boolean from `settings.Seed`/`SeedValue`.
- Adjust to `identity.Service.BootstrapInstallation` and `ClaimInstallation` returning `discardedKeys`; pass a logger to `setupClaim` where warnings are emitted. No schema or config changes.

<sup>Written for commit 842d98159d933c3523d9ea0c4d6f96dbdfee6356. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/2070?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/internal/compose/extjobseat_integration_test.go
+++ b/backend/internal/compose/extjobseat_integration_test.go
@@ -36,7 +36,7 @@ func TestTheDispatcherResolvesTheSeatBootstrapMinted(t *testing.T) {
 		t.Fatalf("archiving the harness workspace: %v", err)
 	}
 
-	wsID, created, err := identity.NewService(e.Pool).BootstrapInstallation(ctx,
+	wsID, created, _, err := identity.NewService(e.Pool).BootstrapInstallation(ctx,
 		func() (identity.InstallationBootstrap, error) {
 			return identity.InstallationBootstrap{
 				OrganizationName: "seatjoin",
--- a/backend/internal/compose/handlers_setup.go
+++ b/backend/internal/compose/handlers_setup.go
@@ -6,7 +6,9 @@ package compose
 import (
 	"errors"
 	"fmt"
+	"log/slog"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/jackc/pgx/v5/pgxpool"
@@ -83,7 +85,7 @@ func setupStatus(svc *identity.Service, limit *ratelimit.Limiter) http.HandlerFu
 }
 
 // setupClaim creates the organization and its first admin from a claim.
-func setupClaim(svc *identity.Service, pool *pgxpool.Pool, seeds deployconfig.Seeds, limit *ratelimit.Limiter) http.HandlerFunc {
+func setupClaim(svc *identity.Service, pool *pgxpool.Pool, seeds deployconfig.Seeds, limit *ratelimit.Limiter, log *slog.Logger) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		if !limit.Allow(httpserver.ClientIP(r)) {
 			httperr.Write(w, r, apperrors.ErrBudgetExceeded)
@@ -106,7 +108,7 @@ func setupClaim(svc *identity.Service, pool *pgxpool.Pool, seeds deployconfig.Se
 			return
 		}
 
-		wsID, err := svc.ClaimInstallation(r.Context(), in.SetupToken, identity.InstallationBootstrap{
+		wsID, discarded, err := svc.ClaimInstallation(r.Context(), in.SetupToken, identity.InstallationBootstrap{
 			OrganizationName: in.OrganizationName,
 			BaseCurrency:     in.BaseCurrency,
 			Timezone:         in.Timezone,
@@ -130,6 +132,15 @@ func setupClaim(svc *identity.Service, pool *pgxpool.Pool, seeds deployconfig.Se
 		case err != nil:
 			httperr.Write(w, r, err)
 			return
+		}
+		// The claim refuses a provisioned installation, but `setting` is not
+		// tenant-scoped: a claim over a database whose previous workspace was
+		// archived creates a new one beside settings rows that survived, and the
+		// identity the human just typed into the claim form is discarded. They
+		// see the old name in the UI and have no way to tell why (#863).
+		if len(discarded) > 0 {
+			log.Warn("the claim kept the identity already stored and discarded what was submitted; a previous installation's settings survived because they are not scoped to a workspace",
+				"discarded_keys", strings.Join(discarded, ", "), "workspace_id", wsID.String())
 		}
 		httperr.WriteJSON(w, http.StatusCreated, setupClaimResponse{WorkspaceID: wsID.String()})
 	}
--- a/backend/internal/compose/installation.go
+++ b/backend/internal/compose/installation.go
@@ -18,6 +18,7 @@ import (
 	"log/slog"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/jackc/pgx/v5"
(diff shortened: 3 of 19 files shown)


tinnhlt-gra:
## A premise correction on #863, found while writing its own scenario as a test

I wrote the bootstrap-level test the goal asked for — bootstrap, archive the workspace, bootstrap again with a different name/timezone/currency — and it does not reach the discard at all:

```
second bootstrap: created=false err=ERROR: duplicate key value violates unique
constraint "role_key_unique" (SQLSTATE 23505)
```

Migration `0225_collapse_composite_keys` dropped `workspace_id` from that constraint (`role_key_unique UNIQUE (key)`), so role keys are unique **installation-wide**. `seedSystemRoles` runs before the identity settings are read back, and a second bootstrap over a database that still holds the first installation's rows now **fails loudly on the roles** rather than silently inheriting an identity.

So #863's headline scenario — an installation quietly wearing a previous one's identity — is **not currently reachable through the configured bootstrap path**. What is there instead is a raw constraint error, which is its own poor outcome but is not silence.

**This does not make the change unnecessary, and I have not widened it to chase the new finding.** The discard mechanism is real and still silent at every seed call site; `seedRouting` is one that a re-bootstrap is not required to reach. What changes is the claim: this PR fixes the invariant (`Seed` reports), not a scenario that now errors earlier. The PR body says so.

I removed the bootstrap-level test rather than keeping one that passes for the wrong reason. The settings-level test remains, and both its assertions are mutation-proven independently.

Worth someone deciding whether the role collision is the behaviour wanted — a raw `SQLSTATE 23505` at boot is not a message an operator can act on. It is not this PR's to fix, and I have noted it on #863.

## Also in this push

`createInstallation` takes the discards as an out-parameter rather than a third return value. The function has ten error returns, and making it a result would restate every one to add a `nil` — turning ten untouched error paths into changed lines, which is both a larger diff than the change deserves and reads as new code to anything comparing revisions. The comment at the signature says so.

Issue #2072, opened 2026-08-20
The scheduled cache-warm job fails: it lints without installing golangci-lint


tinnhlt-gra:
`main` shows `failure` on its most recent runs, and the failing workflow is not
CI — it is the scheduled **cache-warm** job:

```
$ gh api .../actions/runs/<id> -q '.name + " | " + .event'
cache-warm | schedule
```

Its `warm plain` job ends:

```
FAIL: golangci-lint not found — run 'make tools'
FAIL — golangci-lint findings in: backend/tools cli/craft composition desktop/launcher
       extensions/de extensions/dispact-connector extensions/notes extensions/yogi
       extensions/zalo-oa extensions/zalo-personal
make: *** [Makefile:643: lint-modules] Error 1
```

It reaches `lint-modules` without having run `make tools`, so the linter is
absent and every module reports "findings" that are really a missing binary.

## Why it matters more than a warming job failing

**It makes `main` read as red.** `gh run list --branch main` shows `failure` at the
top, and the natural reading is that main's merge gate broke — which sends anyone
checking main's health (an agent following "if main is red, fix it first", or a
person before branching) chasing a failure that does not exist in CI. Two separate
main-red incidents happened in this repo today; a third signal that is noise makes
the real ones harder to see.

It also means the cache the job exists to warm is **not being warmed**, so every
subsequent run pays the cold-start it was meant to avoid.

## Suggested fix

Run `make tools` (or the same install step CI's lint job uses) before
`lint-modules` in the cache-warm workflow — or, if warming does not need the
linter at all, warm without invoking a target that requires it.

Found while checking whether `main` was red before merging #1990.

dungbanguyen:
Fixed on `main`. The `cache-warm` workflow now installs the gate binaries before it runs anything that needs them:

```yaml
      # `make check-backend` runs golangci over every module, so the gate
      # binaries have to be here or lint-modules fails before it compiles
      # anything — which is what this job exists to do. Same cache and same key
      # as the ci lane's own step, so the two share one entry rather than each
      # paying the ~75s compile.
      - name: Cache the gate binaries
        id: gate-binaries
        uses: actions/cache@… # v6.1.0
        with:
          path: ~/go/bin
          key: gate-binaries-${{ runner.os }}-…
      - name: Install the pinned gate binaries
        if: steps.gate-binaries.outputs.cache-hit != 'true'
        run: make tools
      - name: Gate binaries onto PATH
[...]

Issue #1580, opened 2026-08-17
The setup token is logged in plaintext on every unprovisioned boot, not only on write failure


dungbanguyen:
The one-time setup token is written to the server log in plaintext on every unprovisioned boot, including when the file write succeeded. The log's reader set is far larger than the file's.

Found during an adversarial review of the Windows portability change in PR #1120. **Pre-existing on `main`; not introduced by that PR.**

## What happens

`compose.announceSetupToken` ends with:

```go
log.Warn("installation is unprovisioned: claim it with this one-time setup token",
    "setup_token", raw, "token_file", path, "write_error", writeErr)
```

`raw` is the plaintext credential that claims the installation (ADR-0105). Presenting it makes the bearer the first admin — organization, admin email, admin password all theirs.

The two-channel design is deliberate and the reasoning is sound: the log is where an operator watching a first boot is already looking, and it survives a failed file write. The defect is that the token goes to the log **unconditionally**, including on the path where the file write worked.

## Why the log is the wider channel

On a server this is stdout → container logs → whatever ships them, with retention. The readers include anyone with `pods/log` RBAC, anyone with a viewer role on the log store, and any CI job that scrapes container output. That set is strictly larger than "can read a `0600` file as the server user", and unlike the file it persists in a searchable index after the installation is claimed.

The effect is that the careful `O_EXCL` handling on the file protects the narrower channel while the wider one is unconditional.

## Shape of a fix

Log the path plus a short fingerprint — first 6 characters, or `sha256(raw)[:8]` — on every boot, and the full token **only** when `writeErr != nil`. That is exactly the justification the existing comment gives for having two channels: the log is the fallback for a failed write, not a duplicate of a successful one.

Worth checking at the same time whether the token should be redacted from the log at the slog handler level, so a future call site cannot reintroduce it.

## Related

The companion issue covers the file itself: the parent directory is unguarded on every platform and Windows sets no DACL, so a writable working directory before first boot can capture the same credential. Note the two interact — `desktop/launcher/process.go` writes the api's stdout to `data/logs/api.log`, so on the desktop path the log channel becomes a file channel with the weaker of the two protections.

tinnhlt-gra:
[...]

PR #2073, opened 2026-08-20
docs(status): record wave 5 and the two red mains it ran into

tinnhlt-gra:
Session record for wave 5. Two entries at the top of the index:

- **The four wave-5 goals** — the setup token in the log (#1580), the token
  file's parent directory and the Windows DACL (#1579), the object store's
  region (#1792), and the discarded bootstrap seed (#863).
- **The two red mains hit on the way** — the RBAC repair suite's counted
  rollback (#2024) and the partner work's two unclassified fitness functions
  (#2046), plus #2072 for the cache-warm job that makes `main` read red for a
  third, unrelated reason.

Each entry carries the findings rather than the narrative. The ones most likely
to save someone else time:

- **A gate's exception is where it reads green over its own class.** Every plant
  I wrote attacked the matcher; both holes reviewers found were in the escape
  hatch.
- **`git add` with a suppressed stderr can produce a commit that is a bare file
  rename**, while the local gates pass over a working tree that still holds the
  work.
- **`0225_collapse_composite_keys` made role keys unique installation-wide**, so
  #863's scenario now dies on `SQLSTATE 23505` before the defect it describes can
  happen.
- **The sweep gate keyed a waiver on the table while reporting a pair** — fixing
  the bug it reported would have exempted the table from it for good.

Docs only; no code paths touched.

---
## Summary by cubic
Updates `STATUS.md` to record wave 5 outcomes (credential ownership hardening) and two CI‑missed red mains, and aligns ship dates with UTC merge dates while fixing an MD038 lint (rephrases the git‑status marker). This matters because it captures the behavior changes and incident learnings and keeps the index consistent with git history. Docs only; no code paths change.

- Wave 5 records changed behavior: setup token logs only on persistence failure; parent directory is created/verified and Windows DACL applied; implicit object store region is refused; discarded bootstrap seed is reported.
- Red mains entry records fixes and CI gaps: RBAC repair derives migration distance; sweep gate waiver key corrected; partner work broke two fitness functions; cancelled runs masked failures; cache‑warm job issue linked.

<sup>Written for commit 62d4f1a486f7da584a05be9fdfe7c981e51a7071. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/2073?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Documentation**
  - Updated the project status with Wave 5 deliverables and incident findings from August 21, 2026.
  - Documented improved credential-log redaction, secure token-file handling, and Windows access-control protection.
  - Recorded stricter object-store region configuration and clearer reporting for discarded bootstrap seeds.
  - Added notes on RBAC migration-distance calculations, sweep-gate corrections, cancelled-main visibility, and scheduled cache-warm tooling gaps.


tinnhlt-gra:
Both applied in 62d4f1a48.

The date one was right and I had the wrong reading: the merges are stamped `2026-08-20` in UTC, which is what `git log` will show, so the entries now say that rather than the local date my session had rolled over to.

MD038 fixed by describing the marker in words instead of a code span — which also removes the ambiguity you flagged, since the sentence now says explicitly that it is the WORKTREE column (unstaged) being mistaken for the index column (staged).

PR #2103, opened 2026-08-21
fix(deals): a deal can be marked won from the UI

### LINKED ISSUE #2088: A deal cannot be marked won from the UI at all: nothing links a contract to a deal, or moves it out of draft
## The claim, tested

**No deal can be closed as won through the UI.** Not a company-less deal, not any deal. Verified end to end on a seeded stack.

Marking a deal won requires evidence (`backend/internal/modules/deals/win_evidence.go`, `evidenceQuery`): a contract **on that deal** (`c.deal_id = $1`), past `draft`, with `signed_on` set, carrying an attachment categorised `contract`/`legal` in `doc_state` `current`/`final`.

The UI cannot produce such a contract:

1. **Nothing sets `deal_id`.** The only contract form is "Record an agreement" (Company → Documents → **Add a contract**). Its fields are Title, Contract number, Value, "This value is", Starts, Ends, Renews, Notice period, Signed, and a Signed-document dropzone. **There is no deal field.** `grep -rn "deal_id" frontend/src/screens/contractform.tsx frontend/src/screens/companycontracts.tsx` returns nothing.
2. **Nothing leaves `draft`.** A contract is created in `draft` and no frontend surface calls `POST /contracts/{id}/status`.

Filled the form in a browser (title + signed date + no file), saved, then read the row:

```
          title           | status | signed_on  | deal_unset
--------------------------+--------+------------+------------
 Fleet retrofit agreement | draft  | 2026-08-01 | t
```

Two of the four conditions fail immediately, and the first cannot be satisfied from any screen.

## What the user sees

Press **Won** on the stepper (or drop on the Won column) → the ordinary confirm dialog appears and offers **Confirm** → the deal stays **open** and a message appears naming raw enum members:

> a won deal needs a signed contract with its paper attached, or a reason why there is none (imported, purchase_order, verbal, renewal_by_email, other)

None of those words exist anywhere in the interface, and the one thing that would satisfy the rule cannot be created.

## Relationship to the other issues

- #2084 filed this as a *surfacing* problem — a bad error message with a working-but-hidden path. **That was wrong, and this issue supersedes that reading.** There is no working path.
- #2085 (500 on a company-less deal) is a separate, narrower bug on the same journey.

## Why critical

Closing a won deal is the terminal act of the pipeline and the source of every revenue figure — forecast attainment, quotas, win-loss, `deals-by-stage`. Today a rep cannot record one. This is the product's core loop being unusable on a default install, which is the `critical` bar in CLAUDE.md.

## Fix directions

Any one unblocks it; the first two are the smallest.

1. **Collect the reason at the point of closing.** The backend already validates `won_without_contract_reason` (+ detail for `other`). Add it to the terminal-advance dialog when the target stage is won — the same dialog already collects a lost reason.
2. **Let the contract form name a deal**, and offer a status control so an agreement can leave draft.
3. **Say it before the click.** The dialog knows the target semantic; when it is won and no qualifying contract exists, say what is missing rather than offering a Confirm that cannot work.

Whatever is chosen, the raw enum list should not reach a reader.

Found while verifying the end-user deals guide (#2086); the guide has been corrected to state that the path does not exist rather than instructing readers to walk it.

LarsGradion:
## What was broken

**No deal could be closed as won from any screen.** Not a company-less deal — any deal. This was #2088, filed as `priority: critical`.

The win-evidence rule (`backend/internal/modules/deals/win_evidence.go`) admits a win two ways:

1. a signed contract **on the deal**, past draft, with its paper attached; or
2. a stated reason why there is none, from a closed vocabulary.

Neither was reachable. Path 1 has no UI at all — nothing sets `contract.deal_id` and no screen moves a contract out of `draft`. Path 2 had no field. So pressing **Won** always produced a 422, the deal stayed open, and the message named raw enum members (`imported, purchase_order, verbal, renewal_by_email, other`) that appear nowhere in the interface.

## The fix

The confirm dialog offers the reason — **but only once the server has asked for it.**

The first **Confirm** sends the win exactly as before, so a deal that *has* a contract closes in one click. Only when the server answers `win_evidence_required` does the dialog stay open and grow the question, rather than dropping the reader back on a board that has already snapped the card home.

This ordering is deliberate. Asking every rep to justify a win the paperwork already explains is how a required field becomes a field everyone fills with the same lie — and it would also make "won with paper" and "won without it" indistinguishable in reports, which is the whole reason the server keeps them apart.

Two details that follow the server's own rules:

- **"Something else" holds Confirm until the detail says what it was** (`WonReasonDetailRequiredError`), so the reader is not refused twice for the same omission.
- **The dialog keys on the field code, not the 422.** An advance is refused for several reasons — a version conflict among them — and only this one has an answer the reader can give in this dialog.

The reason vocabulary's TS type is derived from the generated contract, so adding a member to `crm.yaml` without offering it here stops compiling instead of silently leaving a choice the server accepts and no screen shows.

Both surfaces are covered by one change: the board's drag and the deal page's stepper share this dialog.

## Scope

**Frontend only.** Every field already existed in `crm.yaml`, was already mapped in `advanceDealInput`, and was already validated and persisted. Only the UI was missing. No contract change, no migration, no `make gen`.

## Verification

Driven in a browser against a real stack, not only against the test stub:

- Pressing **Won** on a deal with no contract → dialog stays open, reason panel appears, **Confirm** disabled until a reason is picked.
- Picking *On a purchase order* and confirming → deal reads **won**, stepper goes inert, `won_without_contract_reason = purchase_order` in the row.
- The write shape held: domain row + `audit_log` (`advance_stage`) + `event_outbox` (`deal.stage_changed`, `to_status: won`) in one transaction, linked by `audit_log_id` and `correlation_id`.
- The real server's refusal body was confirmed to be `details.errors[].code = "win_evidence_required"` — the exact shape the dialog matches on.

Both new tests were mutation-checked twice: once with the reason panel suppressed, once with the request-body wiring removed. Each half fails independently, so neither passes for the wrong reason.

`make check-fe` green. `make craft-static` and `make rls-store-path` green.

## Not in this change

The stored reason is never displayed back on the won deal — filed separately rather than widening this fix.

Closes #2088

---
## Summary by cubic
Allow deals to be closed as Won from the UI. Old behavior: clicking Won returned 422 and left the deal open. New behavior: a deal with a signed contract closes in one click; otherwise the dialog asks how it was won and gates Confirm accordingly.

- The first Confirm is unchanged; on `win_evidence_required` the dialog stays open, renders a reason selector, and disables Confirm until answered. “Something else” additionally requires a detail; zero‑width input is rejected to match the server’s rule.
- Detection keys on the problem field code, not HTTP status. The refusal is pinned to the specific deal; opening Won on another deal does not inherit the reason UI.
- The dialog closes on success and for non‑evidence errors; it stays open only for a win‑evidence refusal. Confirm is disabled while submitting. State clears on dismiss.
- The reason detail is cleared when changing away from “Something else” and is only sent with that reason.
- Won‑reason type is derived from the generated contract; adding a reason in `crm.yaml` without wiring it here fails compilation.
- Shared dialog for board drag and the deal stepper; i18n strings added in `en`, `de`, and `vi`. Docs updated to describe both win paths and current contract‑linking limits.
- Frontend only; no API or contract changes; no migration. Tests cover the server‑demanded evidence path, dialog lifecycle, “other + detail” validation (including zero‑width), state isolation between deals, and non‑regression of ordinary advances.

<sup>Written for commit e193ec50f34131dedb3759775b577521b8da298f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/2103?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/how-to/work-your-pipeline.md
+++ b/docs/how-to/work-your-pipeline.md
@@ -69,33 +69,37 @@ Moving to a **Won** or **Lost** stage never happens on the drop. A dialog asks f
 lights up. If you cancel — or press Escape, or click outside — anything you typed is cleared,
 so the next deal you close never inherits the last one's reason.
 
-**Won needs evidence, and this is the part that surprises people.** A deal can only be marked
-won when the deal has a **contract** attached that is:
+**Won asks what is behind the win.** Margince accepts a won deal two ways, and both are
+legitimate — it just refuses to record a win that says nothing at all.
 
-- past **draft**,
-- with a **Signed** date filled in,
-- carrying an attached file categorised **contract** or **legal**,
-- in a **current** or **final** document state.
+The first way is a **signed contract on the deal**. Press **Confirm** and, if one is there, the
+deal closes immediately with no further questions.
 
-**Today you cannot satisfy that from the UI, so a deal cannot be marked won here at all.**
-This is a defect, not something you are missing —
-[#2088](https://github.com/gradionhq/margince-poc-v1/issues/2088) — and until it is fixed,
-pressing **Won** leaves the deal open and prints a message naming values that appear nowhere in
-the interface.
+The second way is **saying there is no contract**. When there is none, the dialog does not give
+up — it stays open and asks:
 
-Two links in the chain are missing. The contract form (Company → **Documents** →
-**Add a contract**) has no field that names a **deal**, so a contract recorded there belongs to
-the company but not to the opportunity. And a contract is created in **draft** with no control
-anywhere to move it out, which the evidence rule also rejects. Recording the agreement and
-attaching the signed PDF both work; the two links that would make it count do not exist yet.
+> **How was it won?**
+> This deal has no signed contract attached, so tell us how it was won. The answer is kept on
+> the deal and counted in reports.
 
-The API can express all of it — including a short reason why there is no contract at all (an
-import, a purchase order, a verbal agreement, a renewal by email) — so anyone driving Margince
-through the API or an agent can close a won deal today. It is the screens that are missing.
+Pick one of five: **On a purchase order**, **Verbally, in person or by phone**, **Renewed by
+email**, **Imported from another system**, or **Something else**. Choosing *Something else*
+opens a **What was it?** box that must say something before **Confirm** lights up — the other
+four explain themselves, that one does not.
 
-Until then, a deal you have actually won stays open in Margince, and the honest workaround is to
-say so where a human will read it: put the outcome in a note on the deal, so the record does not
-silently claim the deal is still live.
+You are only asked when there is nothing to point at, so a deal with paper behind it is still
+one click. The answer is stored on the deal, which is what lets a report later count how many
+won deals had no agreement and why.
+
+Two things to know about the contract path, because you cannot complete it from the UI yet: the
+contract form (Company → **Documents** → **Add a contract**) has no field naming a **deal**, and
+a contract is created in **draft** with no control to move it out. Both links are missing, so in
+practice today every win goes through the reason. Recording the agreement and attaching the
+signed PDF still work and are still worth doing — they just do not yet count as the evidence.
+
+**The reason is not shown back on the deal afterwards**
+([#2105](https://github.com/gradionhq/margince-poc-v1/issues/2105)). It is recorded and reports
+can read it; the record page does not display it yet.
 
 **Closing is one deal at a time, on purpose.** The bulk bar offers open stages only: a lost
 reason is specific to one deal, and one reason standing for a dozen would be a lie in the
@@ -224,9 +228,11 @@ Stated plainly, so you don't hunt for them:
   ([#2034](https://github.com/gradionhq/margince-poc-v1/issues/2034))
 - **The board cannot be dragged on touch devices.** Use the deal page's stage stepper.
 - **There is no text search within the deals list.** Use global search.
-- **A deal cannot be marked won from the UI at all.** The evidence rule needs a contract linked
-  to the deal and out of draft, and no screen does either. See *Close a deal* above.
+- **A contract cannot be tied to a deal, or moved out of draft.** So a win always goes through
+  the reason rather than the paperwork. See *Close a deal* above.
   ([#2088](https://github.com/gradionhq/margince-poc-v1/issues/2088))
+- **A won deal does not show why it had no contract.** The answer is recorded, just not
+  displayed. ([#2105](https://github.com/gradionhq/margince-poc-v1/issues/2105))
 
 ## Where deals meet the rest of Margince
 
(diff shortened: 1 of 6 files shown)


LarsGradion:
## Review round: four defects found and fixed

An adversarial review of the first commit found four real defects, three of which could write a **false record**. Each now has a test that fails without its fix (mutation-checked individually).

**1. A refusal outlived the deal that earned it.** The reason panel keyed on the shared mutation's `error`, which is global. Cancel on a refused deal, open **Won** on a deal that *does* have a contract, and the picker greeted it — and `ensureWinEvidence` takes a stated reason at its word *before* looking for a contract, so that deal would be stored as won-without-paper when it was not. That falsifies the exact count the vocabulary exists to make truthful.

Fixed by pinning the refusal to the `dealId` it names, and clearing it on every dismissal. Verified in a browser on the real path: deal A refused → Cancel → drop deal B on Won → clean dialog, no picker.

**2. A successful first-confirm win never closed the dialog.** `needsWonReason` was false on that path, so the close guard didn't fire and there was no `onSuccess` close — leaving **Confirm** live over a version the write had just replaced. A 403 or 409 also held it open, with the error text rendering *behind* the modal overlay.

The dialog now acts on the outcome of the attempt it made (`mutateAsync`, resolved to the error rather than thrown). Only the evidence refusal keeps it open, because only that has an answer the reader can give here. Confirm is disabled while the write is in flight.

**3. The detail survived a change of reason.** Type a detail under *Something else*, switch to *On a purchase order*, and the field disappears — but the text was still sent, and the server writes both columns as given. Text the reader believed discarded would sit on the deal and in its audit trail. Now cleared when the reason leaves that member, and serialized only with it.

**4. The detail's client check disagreed with the server's.** `trim()` admits a zero-width space; `saysSomething` in `win_evidence.go` does not. Confirm would light up, then the server would refuse — the reader refused twice for one omission. The client rule now mirrors the server's visible-character test, checked against the same cases including `​`.

Also addressed: the vocabulary is now a `Record<WonReason, MessageKey>`, so a new contract member stops this file compiling until it has a label (the previous array shape only proved the listed values were valid, not that the list was complete), and the Select's string output is narrowed rather than cast.

The review separately confirmed no defect in the two-step version pin (both requests use `pending.version`; the evidence refusal returns before `ApplyGuarded`, so it does not consume a version) and that `mutationFn` closes over no render state.

`make check-fe` green — 3699 tests.

Issue #2126, opened 2026-08-21
CI red on main: the voice-profile lifecycle suite expects a provenance the source-rename migration changed


dungbanguyen:
`make test-integration` fails on `origin/main` (verified by checking out
`origin/main` detached and running the suite alone, so this is not a branch
artifact):

```
--- FAIL: TestVoiceProfileLifecycleUpgradePreservesBuiltProfile
    voice_profile_lifecycle_integration_test.go:71: corpus provenance = ("manual", "system"), want (ui, system)
    voice_profile_lifecycle_integration_test.go:72: profile backfill = source "manual", captured_by "system", built_at true; want ui, system, true
    voice_profile_lifecycle_integration_test.go:73: immutable version provenance = ("manual", "system"), want (ui, system)
```

The suite seeds a pre-migration installation and asserts what the upgrade
leaves behind. It expects `source = 'ui'`; the rows come out `'manual'`.

Almost certainly `core/1787315000_source_names_the_origin` (#2102), which
renames source values — the fixture seeds the old vocabulary and the assertion
still names the new one, or the reverse. Whoever wrote that migration will
recognise which half is wrong faster than a bisect would find it.

`priority: critical` per the rubric: a red integration lane makes every other
verdict unreadable, and this one fails for everybody on every branch.

## Second finding, same area, no separate issue

That migration is stamped **`1787315000`** — a round number, so hand-picked
rather than `date +%s`. At the time of filing that is roughly **ten hours ahead
of real UTC**, which means every migration authored between now and then has to
be stamped into the future to clear `check-migration-versions`. I hit it on
#2012 and worked around it by stamping `1787315001`.

It clears `TestCoreMigrationVersionsAreUnixSecondsAfterTheClosedSequence`
because that check has slack, so nothing is red — but the slack is what is
absorbing it, and a stamp far enough ahead would trip it and block the lane
outright. Worth a rule that `make migrate-create` is the only way to stamp a
core migration.

PR #2132, opened 2026-08-21
fix(ci): the voice lifecycle suite expects the vocabulary head actually holds

dungbanguyen:
**Main's integration lane is red**, and has been since #2102 landed. This is the
fix. Closes the first half of #2126.

## What broke

`TestVoiceProfileLifecycleUpgradePreservesBuiltProfile` asserts `source = 'ui'`
on all three voice tables — which is what `core/0107`'s own `DEFAULT` writes.

But the suite seeds a **pre-0107** installation and migrates it to **head**, and
`core/1787315000_source_names_the_origin` (#2102) collapses `'ui'` and `'mcp'`
into `'manual'` across `voice_profile`, `voice_corpus_source` and
`voice_profile_version`. So `'ui'` is a value no migrated database ends up
holding, and the three assertions have failed for everybody on every branch
since.

```
voice_profile_lifecycle_integration_test.go:71: corpus provenance = ("manual", "system"), want (ui, system)
voice_profile_lifecycle_integration_test.go:72: profile backfill = source "manual", captured_by "system", built_at true; want ui, system, true
voice_profile_lifecycle_integration_test.go:73: immutable version provenance = ("manual", "system"), want (ui, system)
```

## The fix, and why it is not a weakened assertion

The expectation moves to `'manual'` **and says why in the code**. A bare value
change here reads exactly like somebody loosening a test to make it pass — so
the note names the migration that does the collapse and states the principle:
the suite exists to pin what an installation that took this upgrade path
actually holds, and asserting `0107`'s default would assert a value no migrated
database has.

## Still open on #2126

That migration is stamped **`1787315000`** — a round number, hand-picked rather
than `date +%s`, and roughly **ten hours ahead of real UTC** at the time of
writing. Every migration authored before then has to be stamped into the future
to clear `check-migration-versions`. It clears the clock check only because that
check has slack. Left on the issue as its own decision.

## Verification

---
## Summary by cubic
Restores green CI by aligning the voice lifecycle integration tests with the current migration behavior. Tests now expect source = "manual" (was "ui") across voice tables because `core/1787315000_source_names_the_origin` collapses "ui" and "mcp" into "manual" for upgraded installs.

- Updates assertions and messages in `backend/migrations/voice_profile_lifecycle_integration_test.go`; adds an inline note explaining the migration-driven expectation.
- No runtime changes; test-only fix. Partially addresses #2126 (timestamp follow-up remains).

<sup>Written for commit 5393e77471ea20eb407a5a0f82e18e98e8057df1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/2132?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/migrations/voice_profile_lifecycle_integration_test.go
+++ b/backend/migrations/voice_profile_lifecycle_integration_test.go
@@ -98,6 +98,12 @@ func migrateToBeforeVoiceLifecycle(t *testing.T, conn *pgx.Conn) dbmigrate.Names
 	return core
 }
 
+// The three asserts below expect source = "manual", not the 'ui' that 0107's
+// own DEFAULT writes. This suite migrates a pre-0107 installation all the way to
+// HEAD, and core/1787315000_source_names_the_origin collapses 'ui' and 'mcp'
+// into 'manual' across all three voice tables — so 'manual' is what an
+// installation that took this upgrade path actually holds. Asserting 0107's
+// default would be asserting a value no migrated database ends up with.
 func assertMigratedVoiceCorpus(t *testing.T, conn *pgx.Conn, profileID, wantContentHash string) {
 	t.Helper()
 	var contentHash, kind, register, source, capturedBy string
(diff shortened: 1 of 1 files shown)


dungbanguyen:
Superseded by #2128, which fixed the same failure on main while this sat unrebased — and fixed it better: it names the expectation `wantMigratedSource` with the reason on the constant, rather than repeating the literal at three call sites the way this branch did.

#2126 is closed by that PR. Closing this unmerged; nothing here is worth carrying forward.

Issue #2133, opened 2026-08-21
main is red: the data reset's sweep gate refuses activity_link's new project-clock trigger


tinnhlt-gra:
`TestSweepTargetsCarryNoDeleteBlockingTrigger` (`backend/internal/compose/datareset_integration_test.go:342`) fails on a clean `origin/main` worktree at `7496aeb9b`:

```
datareset_integration_test.go:375: sweep target "activity_link" carries DELETE-firing
  trigger "activity_link_project_last_activity" — an append-only or otherwise protected
  table belongs in preservedResetTables; a row-conditional guard belongs in
  deleteGuardedSweepTargets, with what the reset owes it
```

Reproduced in isolation:

```
make test-it DIR=backend/internal/compose RUN=TestSweepTargetsCarryNoDeleteBlockingTrigger
--- FAIL: TestSweepTargetsCarryNoDeleteBlockingTrigger (5.69s)
```

## Cause

#2120 (`69329499c`) added migration `1787320000_project_last_activity`, whose
`activity_link_project_last_activity` trigger fires `AFTER INSERT OR UPDATE OR DELETE
ON activity_link` to recompute `project.last_activity_at`. `activity_link` is a sweep
target, and the gate refuses any DELETE-firing trigger on one until it has been
classified — which is exactly what the gate is for. The trigger was not classified.

This is the same gate and the same shape as #2058, one table over.

## The classification, for whoever takes it

It is not a guard. `trg_activity_link_project_last_activity` never `RAISE`s; it calls
`move_project_last_activity(OLD.project_id)` and returns NULL. `project` is not in
`preservedResetTables`, so the sweep deletes those rows too and the recompute is
discarded with them — which is word for word the reasoning already recorded for the
sibling entry `activity_link activity_link_last_activity`.

So the fix looks like one more line in `deleteGuardedSweepTargets`, keyed on the
`<table> <trigger>` pair as that map requires.

## Why critical

`main` is red, so every open PR's integration lane fails on something none of them
changed, and a red run nobody trusts makes every other verdict unreadable.

PR #2142, opened 2026-08-21
fix(agents): an approval is spent on what it approved (#2014, #2015, #2016)

### LINKED ISSUE #2014: A confirm-first archive stages on readability but executes on writability, spending the approval
Found by a Codex review during PR #2013 (archive_record serving activity). Filed rather than fixed: it affects every archivable record type, not the one that PR touches.

## What is wrong

`archiveResolver.Guards` (`backend/internal/modules/agents/command.go:225`) refuses only an archive whose target the caller cannot **read**. The store then also requires the caller to be able to **write** it — `EnsureActivityWritable` (`backend/internal/modules/activities/lifecycle.go:150`), `EnsureWritable` (`backend/internal/modules/people/organization.go:211`), and the equivalents on the other types.

Those are not the same permission. A team-scoped rep can read another team's activity that is linked to a workspace-readable contact, and cannot modify it — there is already a test showing exactly that at `backend/internal/compose/integration/activity_scope_integration_test.go:94`.

Approval visibility uses content visibility rather than writability (`backend/internal/modules/approvals/targetvisibility.go:372`), so the rep can stage the archive AND approve it. Redemption then consumes the one-shot approval (`backend/internal/modules/approvals/redeem.go:82`) and the retry fails with `permission_denied`.

## Why it matters

The authority object is spent on a call that could never run, and the human who granted it sees a refusal after saying yes. It is the same defect class PR #2013 fixed for `activity` at the *type* level — a call admitted at staging and refused at execution — one layer down, in permissions rather than in the served set.

Nothing is archived that should not be: the store still refuses. The cost is the wasted approval and the misleading sequence.

## Done looks like

`Guards` asks whether the caller may WRITE the target, not merely read it, so the refusal arrives before anything is staged. Stage-time truth equals execute-time truth — the principle `archivableRecordTypes` now states for the type check in `backend/internal/modules/agents/tools_confirm.go`.

Worth checking the other confirm-first verbs for the same asymmetry while in there.

### LINKED ISSUE #2015: An approved archive can land on a version the human never approved
Found by a Codex review during PR #2013. Filed rather than fixed: it is on both doors and predates that PR.

## What is wrong

The approval pins the target's version, and redemption checks that pin and consumes the approval in one transaction (`backend/internal/modules/approvals/redeem.go:82`). The archive then runs **afterwards**, in a separate transaction, with an unconditional `UPDATE ... SET archived_at` (`backend/internal/modules/activities/lifecycle.go:153`) that never compares the approved version.

So between redemption and the write, a concurrent update or relink can change the record, and the archive lands on a version nobody approved.

Not MCP-specific: the REST handler calls the same unversioned store method (`backend/internal/modules/activities/handlers_lifecycle.go:32`). PR #2013 only makes the MCP path reach it for activities as well.

## Why it matters

An approval is a judgment about the record as the human saw it — that is why StageInfo pins the version at all. Losing the pin before the write makes the pin decorative in the race.

The window is small and needs a concurrent writer, which is why this is filed rather than blocking.

## Done looks like

The archive enforces the approved version inside the transaction that performs it, on both doors, so a changed record refuses instead of archiving.

### LINKED ISSUE #2016: archive_record stages types overlay mode cannot archive, if the mode changes after staging
Found by a Codex review during PR #2013, which added `activity` to the archivable set.

## What is wrong

`archivableRecordTypes` (`backend/internal/modules/agents/tools_confirm.go`) is context-free and lists what the NATIVE provider archives: person, organization, deal, project, relationship, activity.

Overlay archives only three (`backend/internal/modules/overlay/provider_writes.go:328`):

```go
var archivableTypes = map[datasource.EntityType]bool{
	datasource.EntityPerson: true, datasource.EntityOrganization: true, datasource.EntityDeal: true,
}
```

Its comment claims this is "the same set the native provider archives (person/organization/deal)" — that is stale. Native has archived `project` and `relationship` for some time, so the divergence predates PR #2013; `activity` makes it three types wide.

The dispatcher picks native or overlay from request context (`backend/internal/compose/dispatcher.go:383`).

## The window

A call made while already in overlay mode is refused before staging by the non-authoritative-record guard (`backend/internal/modules/agents/approvals.go:216`), so the common case is safe.

The gap is a **pending native-mode approval that outlives a connection change**. On retry the dispatcher freshly selects overlay, redemption has already consumed the approval (`backend/internal/modules/agents/registry.go:253-277`), and overlay then refuses the type — the spent-approval bug PR #2013 fixed, recreated for project, relationship and activity whenever the system-of-record mode changes between staging and retry.

## Done looks like

The served-set question becomes context-aware — an optional capability like `Archivable(ctx, type)` on a V2 interface, so `StageInfo` asks the provider that will actually execute rather than a hardcoded native list. `SystemOfRecordProvider` itself is frozen (`backend/internal/shared/ports/datasource/datasource.go:104-110`, enforced by `freeze_test.go`), so it cannot simply gain the method.

Also worth correcting `archivableTypes`' comment, which asserts a parity that no longer holds.

### LINKED ISSUE #2159: main is red: organization_geocode_state's FK to organization has no visibility decision
`TestFK_rowScopedTargetsHaveVisibilityDecision` (`backend/migrations/schema_fitness_integration_test.go:385`) fails on a clean `origin/main` worktree at `9d9820b6e`:

```
FK organization_geocode_state.organization_id -> organization has no visibility decision:
  a reference to a row-scoped record is a read of it — gate it (auth.EnsureLinkTarget)
  or classify it here
```

Reproduced in isolation after `make test-db-up`:

```
make test-it DIR=backend/migrations RUN=TestFK_rowScopedTargetsHaveVisibilityDecision
--- FAIL: TestFK_rowScopedTargetsHaveVisibilityDecision (1.27s)
```

## Cause

#2119 (`a55ae20ad`, migration `1787320001_organization_geocode`) added
`organization_geocode_state`, whose `organization_id` is a foreign key to a row-scoped
record. The gate requires every such FK to be either gated at the writer or classified in
`rowScopedFKDecisions`, and this one is neither. That is the gate doing its job — it exists
so a new reference cannot arrive without somebody deciding what it discloses.

This is the same gate as #2046 (fixed by #2052), one table over.

## What I found, for whoever owns the decision

I did not add the entry myself, because unlike #2133 this is an authz judgement about a new
feature rather than a mechanical copy of an existing sibling. What the code shows:

- The only writer is `internal/modules/people/geocode.go` (lines 311 and 327).
- `AddressForGeocode` opens with `auth.Require(ctx, "organization", principal.ActionRead)`
  and its query is workspace-bound through the `app.workspace_id` GUC.
- The organization id is **not client-supplied**: the sweep selects organizations itself,
  and the file's own comment at line 88 says the path "runs under the SYSTEM principal from
  a job".

So the shape looks like the map's existing `child row:` / system-derived category rather
than the `gated: auth.EnsureLinkTarget` one — a sidecar keyed on an organization the sweep
chose, not one a caller named. **The part that needs the author's eye is what the SYSTEM
principal means here**: it bypasses row scope by design, so the entry must say why a
per-caller visibility gate is not owed, not merely that no caller is involved.

Whoever writes it: `rowScopedFKDecisions` is keyed `"<table>.<column>"` and the reason has
to state what the exception costs, or gatekit refuses it.

## Why critical

`main` is red, so every open PR's integration lane fails on something none of them changed.
This is the second such break today (#2133 was the first, fixed in #2134).

### LINKED ISSUE #2163: main is red: the integration harness crossed the 500-LOC cap
## What fails

`make check-backend` on a clean checkout of `origin/main` (`273047148`):

```
FAIL: backend/internal/compose/integration/harness.go is 505 lines (> 500) — split into one-concept-per-file packages
```

Reproduced in a pristine worktree at `origin/main` with no local changes:
`bash scripts/check-go-file-length.sh` fails on that file alone.

## Why it matters more than its size

`go-file-length` runs over the **whole tree**, not the diff. So this fails every
open PR regardless of what that PR touches — the author sees a red gate naming a
file they never opened, in a lane whose other legs are green.

## Where it comes from

`git log -- backend/internal/compose/integration/harness.go` names **PR #2154**
(`fix(deals): a deal may only name a company that is a partner`) as the commit
that carried it over. Nothing in that change is wrong; the file was at the edge
and one fixture pushed it past.

## Not the waiver

`scripts/go-file-length-waivers.txt` is for files that were **already** over the
cap when the gate landed, frozen at their count — its own header says shrinking
is allowed and growing is not. A newly-over file belongs on the other side of
that line: it gets split.

## Fix

Split it one-concept-per-file. It is a test harness, so the seams are natural
(env construction vs. seeding vs. principal builders).

tinnhlt-gra:
## Why

`archive_record` stages a human confirmation and archives on the approved retry. The module states the rule its own shape depends on, in `agents/tools_confirm.go`:

> *"Stage-time truth has to equal execute-time truth here […] a type admitted at staging and refused at execution **spends a human's approval on a call that could never run**."*

That paragraph exists because it already happened once, to `activity`. Three more instances were open, and they are one defect wearing three hats: **the gate asked a question the seam could not answer, so it answered from an assumption about the executor.**

| | Question | What answered it | Issue |
|---|---|---|---|
| WHAT | does the routed executor archive? | a hard-coded list of what **native** archives | #2016 |
| MAY | this caller archive **this row**? | only whether they could **read** it | #2014 |
| WHICH | version was approved? | nothing — the write carried no version clause | #2015 |

**#2016 in one line:** overlay archives `person`/`organization`/`deal` and refuses `project`/`relationship`/`activity`, while its own comment claimed the set was *"the same set the native provider archives"*. An overlay installation staged three types its writer always refused.

**#2014 in one line:** every archive store requires write authority **and** the object `delete` grant; staging asked for neither. A rep who may read a colleague's record staged an archive, a human released it, the store refused the retry.

**#2015 in one line:** redemption verifies the pinned version and commits, then the archive runs in a **later** transaction with an unconditional `UPDATE … SET archived_at`. A concurrent update in that window lands the archive on a version nobody approved — and nothing reports that it did.

## What changes

`SystemOfRecordProvider.Archive` takes a ref and nothing else, and the v1 method set is **frozen**. The freeze's own escape hatch is named by the test that enforces it:

> *"A post-v1 verb is added as a NEW SystemOfRecordProviderV2 interface plus a runtime capability probe […] an adapter that cannot serve a v2 verb returns ErrUnsupportedBySoR."*

So this adds `datasource.RecordArchiverV2` — `ArchivableTypes` / `RefuseArchive` / `ArchiveAt` — implemented on the compose `Provider` and `Dispatcher`, `people`, `deals`, `activities` and `overlay`. **The v1 interface is untouched** and its freeze test still passes.

A provider that does not answer it keeps v1 behaviour, and every caller states what it assumes in that case rather than assuming silently. **The one case that refuses** is a released pin against a provider that cannot honour it — falling back there would spend the approval on precisely the write it was granted against something else, quietly, and only on the installations whose adapter happens not to answer the newer seam.

The five archive stores also move off bare `UPDATE … SET archived_at` and onto storekit's guarded patch, whose own doc says *"an unguarded update is not expressible"* — which the archive verb was quietly the exception to.

## ⚠️ A gate had already ratified this, and its reason is the interesting part

`TestEveryByIDUpdateCarriesAConcurrencyGuard` waived four of these archives:

> *"absolute idempotent archive transition; concurrent archives converge, the visibility pre-read only feeds the response"*

That answers **"do two archives race each other"**, and it is still true. It does not answer **"did this archive land on the record the decider judged"** — a concurrent `UPDATE` in the window changes the record *without racing the archive at all*. The waiver was correct about a different question.

`gatekit` verifies a waiver's key still matches a subject; nothing checks the stated rationale still answers the right question. The four are retired here, and the block now says which question its reason covers and why the remaining entries stand (types no approval ever pins).

`ArchiveActivity` carried **no** waiver — the gate accepted its `RowsAffected() == 0 → ErrNotFound` as a guard. It is a guard against the row *disappearing* and not against the row *changing*, and for a confirm-first archive it is change that matters.

## A test was staging an archive that could never run

`TestBothDoorsStageOneRowForOneOperation` went red on the executor probe: its agent carries `RepPerms`, which grants person `create/read/update` and **no `delete`**. Its archive was always going to be refused after a human answered — #2014, reproduced by a test that had been asserting both doors did it identically.

Its sibling states the requirement in its own words: the hiding must be done by the row scope, *"not a missing grant"*. Both now run under a fixture that carries the grant, so the row-scope refusal is provably row-scope.

## Proof

Every fix is **mutation-verified** — reverted one at a time, each turning its own test red:

| Fix | Mutation | Result |
|---|---|---|
| #2016 | `archivableHere` ignores the executor | `TestStagingRefusesATypeTheRoutedExecutorDoesNotArchive` FAIL |
| #2014 | `Guards` stops asking the executor | `TestGuardsRefuseWhatTheExecutorsOwnProbesWouldRefuse` FAIL |
| #2015 | `archiveAt` drops the pin | `TestAnApprovedArchiveCarriesTheApprovedVersionIntoTheWrite` FAIL |
| #2015 (SQL) | `archivePersonRows` passes `nil` | `…AtAStaleVersion…/person` FAIL |

**The unit fake cannot prove #2015** — a fake that records the input it was handed agrees with a store that reads the field and ignores it, which is exactly the shape this defect had. So `archivepin_integration_test.go` runs both directions against real Postgres for person, organization and deal: a stale pin must refuse **and leave the record live**, and the current pin must archive.

That test caught itself passing for person and organization: the concurrent patch named `job_title` and `website`, fields those update requests do not carry. Neither errored — both carry an `AdditionalProperties` map, so the unknown field was absorbed, the patch set nothing, the version never moved, and a "stale" pin that was not stale admitted the archive. `bumpVersion` now reads the version either side and asserts the move.

## Gates

- `make check-backend` — **exit 0**
- `craft static --strict` over the diff — **0 blocker, 0 major, 0 minor**
- `make test-integration` (42 packages) — one failure, `TestSweepTargetsCarryNoDeleteBlockingTrigger`, which **fails identically on a pristine `origin/main` worktree**. It is a fresh red on `main` from the `activity_link_project_last_activity` trigger; filed separately, not this branch's.

## Not in scope

- **REST `If-Match` on the archive routes.** Only `archiveProject` declares the parameter; the other five pass `nil`, which is today's behaviour. Declaring it on the rest is a contract change and its own PR.
- **`archivableRecordTypes` stays a stated list.** It is the fallback for a provider that answers no capability probe, and the native set is the only honest guess about a fork's own adapter.
- **The overlay pin.** `overlay_mirror` carries an `UpdatedAtBaseline` and no `version` column, so an approved-at-a-version archive is refused there rather than silently unpinned. Whether the mirror should grow one is a product question, not this fix.

Closes #2014
Closes #2015
Closes #2016
Closes #2159
Closes #2163

---
## Summary by cubic
Prevents agent approvals from being spent on archives that could never run and guarantees the archive lands at the exact version a human approved. Old behavior: staging used a native-type list and read-only checks; execution issued unguarded updates; REST archives dropped If-Match; overlay ignored preconditions. New behavior: staging asks the routed executor what it archives and what it would refuse; execution archives with CAS at the approved version; REST archive routes accept If-Match and return 409 on version skew; overlay forwards If-Match and refuses pinned archives it cannot honor. Also classifies the geocode sidecar FK to satisfy the schema fitness gate (unblocks CI).

- Adds `datasource.RecordArchiverV2` (`ArchivableTypes`, `RefuseArchive`, `ArchiveAt`) and implements it in `compose.Provider`/`Dispatcher`, `people`, `deals`, `activities`, and `overlay`; centralizes archive routing in `compose.Provider.archiverFor`.
- Agents run the executor’s `RefuseArchive` at staging and call `ArchiveAt` with the approved version at execution; the organization-anchor refusal moved to staging.
- REST/API: all archive routes now declare If-Match and surface 409; handlers parse and pass the version; overlay REST forwards If-Match and refuses pinned requests it cannot satisfy.
- Storage/tests/frontend: archives use guarded patches (CAS); integration verifies version pins against Postgres; routing is asserted DB-free; frontend `schema.d.ts` and `if-match-coverage` reflect the new precondition on six archive screens.
- Rollout
  - REST clients archiving records should send If-Match with the last-seen version; omitting it remains last-write-wins.
  - Forks with v1-only SoR adapters must implement `RecordArchiverV2` to allow agent-approved, pinned archives; otherwise such archives are refused at staging.
  - Overlay installations: pinned archive requests are refused until the mirror supports versioning.

<sup>Written for commit e5b420e2bf4e6a6b21bf22271b9068b32d9061bb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/2142?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Archive operations now support version-aware requests using the `If-Match` header.
  * Archive capabilities are detected by record type and storage mode, improving routing and validation.
  * Archive requests now perform permission and writability checks before execution.

* **Bug Fixes**
  * Prevented stale archive requests from overwriting newer changes.
  * Stale versions now return `409 Conflict`, leaving records unchanged.
  * Added guarded archiving for people, organizations, deals, activities, and relationships while preserving existing success and not-found behavior.


tinnhlt-gra:
@coderabbitai review

tinnhlt-gra:
## Update — review findings addressed (3 commits since the first review)

A review pass found six things; four were defects this branch introduced or left half-done. All are fixed, and the two that changed behaviour are called out below because they are decisions, not cleanups.

**The regression I introduced.** `refuseArchiveHere` stood down for a provider that does not answer `RecordArchiverV2`, while `archiveAt` refused that same provider whenever a pin exists — and `archive_record` is statically `TierConfirmationRequired`, so every redemption carries one. On a fork whose adapter implements only the frozen v1 seam that is **100% of archives failing after the human answered** — precisely the defect this branch exists to close, one layer along. The two halves now agree, and they agree at **staging**.

⚠️ **That is a behaviour choice for such a fork, and worth your ruling.** The alternative is to archive unpinned, which means a human approving "archive this at version 4" and the write landing on whatever the record became. I chose the loud refusal; `TestStagingRefusesAProviderThatCannotCarryAnApprovedVersion` states the alternative in its own doc so overruling it is a one-line change, not an archaeology exercise.

**#2015 was closed on one door only.** The agent gate forwards a released approval's version as the request's own `If-Match` (`compose/agentgatestaging.go`), but **none of the five archive routes declared the parameter** — so the generated wrapper dropped the header, each handler passed `nil`, and a passport archiving over REST after a human release still ran the unconditioned `UPDATE`. The routes now declare `If-Match` and `409`; the handlers parse it; two new integration tests drive the real handler and are mutation-verified (reverting the handler to `nil` turns the 409 into a 200).

**The anchor refusal belongs at staging.** `RefuseArchiveOrganization` omitted `refuseIfAnchor` on the grounds that domain state can change while a human decides. `is_anchor` cannot: stamped at bootstrap (`0083`), no verb moves it, and `0193` adds a CHECK keeping an anchor from being archived at all. Leaving it out spent an approval on the one target that can never succeed — the same defect class as #2014, for the one case where it is certain.

Also: `archivableHere` no longer swallows the `ArchivableTypes` error and falls back to the **wider** native set (a failed mode read is not a reason to admit the three types the check exists to refuse); three doc comments went back onto the functions they describe; and one of my own test assertions was vacuous in the direction it guarded.

### Two gates caught the second-order cost, and both were right

- **`check-contract-frontend-drift`** — the contract change stranded `schema.d.ts`. Regenerated.
- **`if-match-coverage`** — declaring the precondition put **six SPA screens inside the gate's subject set for the first time**, and it failed them. They join `UNPINNED_WRITES` with the reason: the *endpoints* learned the precondition, the screens did not change, and those calls were previously outside the gate rather than inside it and passing. Threading a version through six screens is its own change with its own proof.

### Gates at `bc87a6e33`

- `make check-backend` — exit 0 · `make check-fe` — exit 0
- `craft static --strict` over the diff — 0 blocker, 0 major, 0 minor
- `make test-integration` (42 packages) — one failure, `TestSweepTargetsCarryNoDeleteBlockingTrigger`, **identical on a pristine `origin/main` worktree** and already filed as **#2133**

tinnhlt-gra:
## Rebased onto `273047148`, and two reds on `main` found on the way

**One conflict, resolved by keeping both sides.** Upstream added a geocode waiver to `unguardedByIDUpdates`; this branch added the paragraph saying which question the archive rationale answers. Different subjects, both kept.

### ⚠️ `main` is red twice right now, and neither is this branch's

Both were found by running the gates against a **pristine `origin/main` worktree**, not inferred from a red PR:

| | |
|---|---|
| **#2163** *(filed by me)* | `go-file-length`: `internal/compose/integration/harness.go` is 505 lines. **Carried over by PR #2154.** |
| **#2159** *(already filed)* | `TestFK_rowScopedTargetsHaveVisibilityDecision`: `organization_geocode_state.organization_id` has no visibility decision. |

**#2163 is fixed in this PR, and that is scope creep on purpose.** `go-file-length` runs over the **whole tree**, so a 505-line file nobody's PR touched fails *every* open PR — including this one, with a message naming a file its author never opened. It is split along the grain (row fixtures → `harnessseed.go`; Env construction, principal builders and exec helpers stay), not at whatever line the cap fell on. The waiver file was not the answer: its own header says it holds files that were *already* over when the gate landed and that growing is not allowed.

**#2159 is left alone** — it needs a decision about how that FK is classified, and guessing it inside an unrelated PR is how a wrong classification ships quietly.

`deals/deal.go` also crossed the cap, and that one **is** mine: the rebase brought it to 481 and `RefuseArchiveDeal` pushed it to 503. Split into `dealarchive.go`, refusal and write together — what the store refuses at execution is exactly what a confirm-first staging must be able to ask beforehand, and the two drifting apart is the defect the pair exists to prevent.

### Gates at `3d47ef112`

- `make check-backend` — exit 0 · `make check-fe` — exit 0
- `craft static --strict` over the diff — 0 blocker, 0 major, 0 minor
- `make test-integration` (42 packages) — one failure, **#2159**, reproduced identically on pristine `origin/main`

tinnhlt-gra:
## This PR and #2160 are deadlocked — neither can go green alone

Flagging with the evidence, because both PRs currently *look* like they are waiting on CI when neither will ever move.

**#2142 → red on `integration / integration shard (6/6)`:**

```
schema_fitness_integration_test.go:385: FK organization_geocode_state.organization_id ->
  organization has no visibility decision
--- FAIL: TestFK_rowScopedTargetsHaveVisibilityDecision (5.57s)
```

That is **#2159**, which **#2160** fixes.

**#2160 → red on `deterministic-gates`:**

```
FAIL: backend/internal/compose/integration/harness.go is 505 lines (> 500)
```

That is **#2163**, which **this PR** fixes.

Each PR carries the other's inherited failure. `ci` is the only required status context and it is red on both, so neither auto-merges.

### Why it will not resolve itself

The `main-required-status-checks` ruleset has `strict_required_status_checks_policy: false`, so branches are not required to be up to date. Merging either PR would **not** re-run the other — the red check keeps its recorded value and armed auto-merge never fires. A red check against a base that has since been fixed is indistinguishable from a red check against a real defect.

### The resolution

One PR has to carry both changes, and the **one-line change should travel, not the file split**: adding the seven-line `rowScopedFKDecisions` entry from #2160 to this PR makes it green on both counts. It goes next to `"activity_link activity_link_last_activity"`, and the commit is `380d91ba2` on `fix/geocode-state-fk-visibility`.

The reverse (pulling the 153-line `harnessseed.go` split into #2160) moves far more code through a PR whose subject is a one-line waiver, and would strand this one with a conflict.

Whichever way it goes, the merged PR should carry **both** `Closes` lines — #2163 **and** #2159 — or one issue is left open behind a fix that landed.

I have not touched this branch and will not without its author's say-so. Raising it here so it is visible to whoever gets there first.

tinnhlt-gra:
## This PR now carries three fixes it did not start with, and the reason is a deadlock

**#2142 was red on #2159. #2160 (which fixes #2159) was red on #2163. #2142 fixes #2163.**

`main`'s ruleset is `strict_required_status_checks_policy: false`, so a branch is never required to be up to date — which means **merging either would have re-run nothing on the other**. Both would have sat at a red `ci` against a base that had since been fixed, armed and never firing, indistinguishable from a PR waiting on a slow lane. One PR had to carry both, and the one-line waiver is the change that should travel rather than a 153-line file split.

So this PR now also closes **#2163** (the harness file-length cap) and **#2159** (the geocode FK classification). Both `Closes` lines are in the body so neither issue is stranded behind a fix that landed — the thing we flagged on #2150/#2153.

## ⚠️ The carried waiver's reason was rewritten, and that is the finding

The entry came from the parallel session's #2160. Reading the path before carrying it showed the original asserted more than the code does. It said *"every query on this path carries an explicit `app.workspace_id` predicate rather than trusting the job args."* Two of four do:

| `people/geocode.go` | predicate |
|---|---|
| `:108-115` `AddressForGeocode` read | ✅ |
| `:297-303` `UPDATE organization` | ✅ |
| `:310-315` `INSERT … organization_geocode_state` | ❌ |
| `:325-334` `INSERT … organization_geocode_state` | ❌ |

Three facts turn that from pedantry into cost: the table has **no workspace column of its own** (`1787320001:66` — the FK is its only tenancy); the `UPDATE`'s result is **discarded** (`if _, err := tx.Exec(…)`), so a zero-row update does not stop the transaction; and control therefore reaches the sidecar `INSERT` regardless. A job carrying a foreign organization id updates nothing and still writes that organization's attempt ledger.

**A write, never a read**, and latent — `enqueueGeocode` only mints the id its own workspace-checked transaction was addressing. Filed as **#2167**. The entry that landed states what is true and names it, so the FK is ratified on its real cost.

Worth stating plainly, because it is the same defect class this PR is about: **a waiver was one step from ratifying an authz gate on a guarantee the code does not provide** — and the pressure that made it likely was the deadlock, which was itself produced by the two red mains being fixed. What caught it was re-deriving the claim before carrying it, rather than trusting a teammate who had just correctly caught *me* over-claiming an hour earlier.

## Gates at `e5b420e2b`

- `make check-backend` exit 0 · `make check-fe` exit 0 · `craft static --strict` 0/0/0
- `TestFK_rowScopedTargetsHaveVisibilityDecision` now passes locally against real Postgres

tinnhlt-gra:
@coderabbitai review

Two commits landed since your review of `3d47ef112`, and both are substantive rather than cleanup — flagging so the re-review targets them:

- `61ce00371` — your own finding. `overlayArchive` now parses `If-Match` and forwards it as `ArchiveInput.IfVersion`. The overlay provider **refuses** a pin it cannot honour (the mirror has no version column), rather than archiving unconditioned. The native arm returns before the parse, so nothing reads the header twice.
- `e5b420e2b` — a gate waiver carried in from another PR to break a merge deadlock, with its reason rewritten because the original asserted a predicate two of four statements do not carry. Residue filed as #2167.

Issue #1737, opened 2026-08-18
A PR that fixes a filed issue should have to reference it — four in one eight-issue batch did not


tinnhlt-gra:
## What happened

A batch of eight issues was proposed, ordered and approved for work. **Four of the eight described defects that had already been fixed** — by PRs that never referenced them. The batch was planned partly on work that was already done, and the sweep that caught it cost a sitting.

| Issue | Actually fixed by | Why it stayed open |
|---|---|---|
| #1341 | #1385 | The fix was filed against the duplicate (#1372); nobody linked the sibling report. |
| #1064 | #1067 | Same shape — fixed as #1015. |
| #548 | #901 + #970 | **#970's body says "Closes the residual half of #548" and its metadata references nothing.** |
| #516 | #901 + #970 | Same PR, same omission. |

#970 → #548 is the worked example, and it is the interesting one because the author *did* write it down: the PR body states in as many words that it closes the issue. GitHub believes the metadata, not the prose, so the issue stayed open for six days after its fix merged and was then re-queued as new work.

## Why this is one defect and not four clerical ones

An open issue that is already fixed is the same lie a green gate over a real bug tells, pointed the other way. Everything downstream of it is competent work aimed at nothing, and it reads as a normal iteration right up to the point where there is no defect to fix. The cost is not the clerical fix — it is planning, sequencing, and sometimes re-doing work.

It is also silent: nothing in the current process notices. An unreferenced fix leaves no trace on the issue, so the only way to find one is the sweep this issue came out of.

## Proposal

A PR that fixes a filed issue must **say so in metadata**, one of two ways:

1. `Closes #N` / `Fixes #N` in the body — so `closingIssuesReferences` is non-empty and GitHub closes it on merge; or
2. an explicit **"does not close #N, because …"** line, when the PR deliberately fixes only part of it (`#970` was in fact half of two issues, which is exactly the case worth stating).

Mechanically, the cheapest useful version is a check that runs on PRs whose body mentions a bare `#N`: if the reference is not in `closingIssuesReferences` and no "does not close" sentence accompanies it, the check asks the author to pick one. It is advisory-by-nature — a PR may legitimately mention an issue it does not close — so the useful bar is **it must be answered**, not **it must close something**.

A weaker but zero-infrastructure alternative: add the question to the PR template's checklist ("does this fix a filed issue? if so, is it referenced in the body so GitHub closes it?").

## Not in scope

[...]

PR #2150, opened 2026-08-21
fix(ci-tests): the reset gate knows about the project clock, and main is green again

dungbanguyen:
`main` is red. https://github.com/gradionhq/margince-poc-v1/pull/2120 added `activity_link_project_last_activity` — a second clock-maintenance trigger on `activity_link`, this one for `project.last_activity_at`. `TestSweepTargetsCarryNoDeleteBlockingTrigger` scans every data-reset sweep target for DELETE-firing triggers and fails on any it has not been told about, so the gate went red on the merge and has stayed red since.

Reproduced against `origin/main` in a clean worktree:

```
--- FAIL: TestSweepTargetsCarryNoDeleteBlockingTrigger (2.65s)
    datareset_integration_test.go:375: sweep target "activity_link" carries DELETE-firing trigger
    "activity_link_project_last_activity" — an append-only or otherwise protected table belongs in
    preservedResetTables; a row-conditional guard belongs in deleteGuardedSweepTargets, with what
    the reset owes it
```

## The fix

It is the same shape as `activity_link_last_activity` sitting immediately above it in the map: `AFTER INSERT OR UPDATE OR DELETE`, returns `NULL`, refuses no delete, and only recomputes a clock. So it is registered the same way, with why it is safe rather than merely that it is — and the claim is checked rather than assumed: `project` is a sweep target and is **not** in `preservedResetTables`, so the row the recompute writes to is deleted in the same transaction.

No product code changes. The gate is the thing that was wrong about the schema, not the schema.

---
## Summary by cubic
Restores green on main by teaching the data-reset gate about the new activity_link_project_last_activity trigger. Previously the gate treated it as a DELETE-blocking guard; now it is registered as clock maintenance that refuses no delete, so TestSweepTargetsCarryNoDeleteBlockingTrigger passes.

- Registers "activity_link activity_link_project_last_activity" in deleteGuardedSweepTargets as a non-guard clock-maintenance trigger, mirroring the existing last-activity clock on the same table.
- Confirms project is a sweep target and not in preservedResetTables; recomputed last_activity_at is deleted in the same transaction, so the trigger cannot block deletes.

<sup>Written for commit 497514a60618deb2d97316f2e3454bbd8887c8a5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/2150?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/internal/compose/datareset_integration_test.go
+++ b/backend/internal/compose/datareset_integration_test.go
@@ -316,7 +316,11 @@ var deleteGuardedSweepTargets = gatekit.Waive(map[string]string{
 	"activity activity_refuse_restricted_mutation": "a row-conditional statutory hold, not a protected store: preserving it would leave every activity behind on a reset meant to clear them, and no writer can set the restriction yet (#1557) — when one lands the reset must lift before it sweeps",
 	// Not guards at all (migration 1787032690).
 	"activity_link activity_link_last_activity": "a clock-maintenance trigger, not a guard: it recomputes the last_activity_at of the records the deleted link reached and refuses no delete; the sweep deletes those records too, so the recompute is discarded with them",
-	"relationship relationship_last_activity":   "a clock-maintenance trigger, not a guard: it recomputes the employer's last_activity_at and refuses no delete",
+	// The project arm of the same clock, on the same table (migration
+	// 1787320000). A second trigger rather than a fourth id inside the one
+	// above, so it needs its own entry — and it earns it the same way.
+	"activity_link activity_link_project_last_activity": "a clock-maintenance trigger, not a guard: it recomputes the project's last_activity_at when a link is written, moved or removed, and refuses no delete; `project` is swept too, so the recompute is discarded with it",
+	"relationship relationship_last_activity":           "a clock-maintenance trigger, not a guard: it recomputes the employer's last_activity_at and refuses no delete",
 	// Also not a guard (migration 1787226902): BEFORE DELETE ON organization, it
 	// sets deal.partner_org_id and deal.partner_attribution to NULL so a deleted
 	// partner leaves no dangling attribution. It refuses no delete.


tinnhlt-gra:
Heads-up: this is now a no-op — the same classification merged as **#2134** about two hours ago (from **#2133**).

`origin/main` already carries the entry:

```
"activity_link activity_link_project_last_activity": "a clock-maintenance trigger, not a guard: it recomputes
  project.last_activity_at for the project the deleted link filed the activity against and refuses no delete;
  `project` is not preserved, so the sweep deletes those rows too and the recompute is discarded with them",
```

So `TestSweepTargetsCarryNoDeleteBlockingTrigger` is green on main, and this branch would land an empty change while holding a full integration lane — which is worth reclaiming right now, with the runner pool as backed up as it is.

Two other PRs converged on this same fix independently (#2150 and #2153), which is worth noticing on its own: a red `main` pulls several people onto the same one-line fix at once. Filing the issue first is what would have made that visible — #2133 was open for about twenty minutes before #2134 closed it.

One thing to carry forward whichever way you go: this PR carries no closing reference, so even if it had been the fix that landed, **#2133 would still be open**. `Closes #2133` in the body is what keeps the issue and the fix from drifting apart.

No action needed from me — flagging it so you can decide rather than find out from a conflict.

dungbanguyen:
Superseded by https://github.com/gradionhq/margince-poc-v1/pull/2134, which landed while this was in CI and registers the same trigger with the same reasoning. `main` is green on `TestSweepTargetsCarryNoDeleteBlockingTrigger` again — verified against `origin/main` — so there is nothing left here to merge.

Their version is the better of the two on one point worth keeping: it says explicitly why the entry is not folded into the `activity_link` line above it — the map is keyed on the PAIR so that a table-keyed entry cannot ratify the next DELETE trigger on that table sight unseen, including one that blocks.

tinnhlt-gra:
Closing as superseded — this change is already on `main`.

**#2134** (`9d9820b6e`, *"fix(ci-tests): the sweep gate classifies the project clock it was handed (#2133)"*) landed the same waiver key, `"activity_link activity_link_project_last_activity"`, with an equivalent reason. Nothing here adds coverage `main` does not have, so merging it would mean resolving a conflict in order to land an empty change — and spending a CI lane to do it, on a runner pool currently ~18 deep.

That is also why this reads `DIRTY`/`BLOCKED`: the conflict is with the very lines that already merged.

## Worth recording, because this is the fourth copy

#2146, #2150 and #2153 are all the same one-line fix, and #2134 is the one that landed. Four sessions independently hit red `main`, read the same gate failure, and wrote the same waiver.

The interesting part is the timing: **#2133 was open for about twenty minutes before #2134 closed it.** So this was not somebody ignoring an existing issue — the duplicates were in flight before the issue could be triaged. That is what a red `main` does on a shared runner pool, and the cost is not the wasted writing but the lanes each duplicate holds while everything else queues behind it.

**None of the four carried a `Closes #2133` trailer.** So even the copy that was correct would have left the issue open for the next person to re-discover — the pattern tracked in #1737, which is that a fix which names its issue only in prose is indistinguishable from one that never looked.

No action needed on the classification itself; it is on `main` and the gate is green over it.

tinnhlt-gra:
**Correction to my close comment above.** The close itself stands — this PR is superseded by #2134 and its change is on `main` — but two facts I stated about the surrounding episode are wrong, and one of them is unfair to #2134.

- I wrote that **none of the four PRs carried a `Closes` trailer**. False. **#2134 carried `Closes #2133`**, and that trailer is what closed the issue automatically on merge. The three duplicates (this one, #2150, #2153) carried none; #2134 did it correctly.
- I wrote that **#2133 was open for about twenty minutes**. It was open for **69** (03:51:20 → 05:00:48 UTC).
- I implied the duplicates were **in flight before triage could happen**. Also false: #2134 was opened **112 seconds** after the issue, and this PR was opened **40 minutes after that fixing PR already existed**.

The corrected reading, with the timeline, is on **#1737**. The short version: the trailer mechanism worked here. What failed is that an issue with an *open* fixing PR looks identical to an unclaimed one — a `Closes` trailer creates a cross-reference, not a claim, so it sets no assignee and no label until the PR merges. That window was 69 minutes today because CI was ~18 deep, and three sessions walked into it.

Sorry for the noise on a closed PR; a wrong reason recorded next to a correct close is the kind of thing that gets quoted later.

PR #2153, opened 2026-08-21
fix(ci-tests): classify the project-clock trigger, which guards nothing

dungbanguyen:
**Main's integration lane is red** on
`TestSweepTargetsCarryNoDeleteBlockingTrigger`, and has been since
`core/1787320000` added `activity_link_project_last_activity` (#2120).

```
datareset_integration_test.go:375: sweep target "activity_link" carries
DELETE-firing trigger "activity_link_project_last_activity" — an append-only or
otherwise protected table belongs in preservedResetTables; a row-conditional
guard belongs in deleteGuardedSweepTargets, with what the reset owes it
```

## Neither of the gate's two homes fits, and that is the finding

The gate offers `preservedResetTables` (append-only table) or
`deleteGuardedSweepTargets` (row-conditional hold). This trigger is **a third
shape**:

```sql
AFTER INSERT OR UPDATE OR DELETE ON activity_link
FOR EACH ROW EXECUTE FUNCTION trg_activity_link_project_last_activity();
-- ... RETURN NULL;
```

`AFTER … RETURN NULL` — it refuses nothing and **cannot fail a DELETE**. It
recomputes `project.last_activity_at` from the live timeline when a link is
written, moved or removed. The gate reads only `tgtype` bit `0x08` and cannot
tell a recompute from a guard, which is precisely what the waiver map exists to
record.

## Why `deleteGuardedSweepTargets` and not `preservedResetTables`

Preserving `activity_link` would leave every activity's filing behind on a reset
whose whole purpose is to clear them.

**The sweep owes this trigger nothing**, and the entry says so: `project` is a
sweep target too, so a reset clears both tables and the clock it maintains has
no row left to be wrong on. That is stated explicitly because the neighbouring
waiver (`activity` / `restricted_at`) documents a *real* obligation the reset
will have to honour once #1557 lands — a reader has to be able to tell the two
apart at a glance.

## Verification

---
## Summary by cubic
Fixes the red integration gate by classifying `activity_link_project_last_activity` as a non-guard trigger. Previously the gate flagged a DELETE-firing trigger on `activity_link`; now it is waived as an AFTER trigger that returns NULL and cannot block deletes, restoring green CI without runtime changes.

- Records the waiver under `deleteGuardedSweepTargets` instead of `preservedResetTables` to avoid preserving `activity_link` on reset.
- The sweep owes this trigger nothing: `project` is swept alongside `activity_link`, so the recomputed clock has no rows left to be wrong on.
- No migrations, no behavior changes, test-only configuration.

<sup>Written for commit 2d07a45aa533e2fcd2775299fd112a280b16d7cf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/2153?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/internal/compose/datareset_integration_test.go
+++ b/backend/internal/compose/datareset_integration_test.go
@@ -302,6 +302,17 @@ func TestDropResetCustomFieldColumns(t *testing.T) {
 // was written for AND every DELETE trigger added to that table afterwards,
 // including one that really does block. Each trigger earns its own line.
 var deleteGuardedSweepTargets = gatekit.Waive(map[string]string{
+	// Not a guard at all, which is a THIRD shape the catalog cannot show: this
+	// one is AFTER ... RETURN NULL, so it refuses nothing and cannot fail a
+	// DELETE. It recomputes project.last_activity_at from the live timeline when
+	// a link is written, moved or removed (core 1787320000).
+	//
+	// The sweep owes it nothing. project is a sweep target too, so a reset
+	// clears both tables and the clock it maintains has no row left to be wrong
+	// on. It is classified here rather than preserved because preserving
+	// activity_link would leave every activity's filing behind on a reset whose
+	// whole purpose is to clear them.
+	"activity_link activity_link_project_last_activity": "an AFTER trigger that returns NULL: it recomputes project.last_activity_at and refuses no delete, and project is swept alongside activity_link so the clock has nothing left to be wrong on",
 	// Not a protected table: the guard refuses a DELETE only while a row carries
 	// `restricted_at`, which is a statutory hold on that one record. Preserving
 	// `activity` would leave every conversation behind on a reset whose whole


tinnhlt-gra:
Heads-up: this is now a no-op — the same classification merged as **#2134** about two hours ago (from **#2133**).

`origin/main` already carries the entry:

```
"activity_link activity_link_project_last_activity": "a clock-maintenance trigger, not a guard: it recomputes
  project.last_activity_at for the project the deleted link filed the activity against and refuses no delete;
  `project` is not preserved, so the sweep deletes those rows too and the recompute is discarded with them",
```

So `TestSweepTargetsCarryNoDeleteBlockingTrigger` is green on main, and this branch would land an empty change while holding a full integration lane — which is worth reclaiming right now, with the runner pool as backed up as it is.

Two other PRs converged on this same fix independently (#2150 and #2153), which is worth noticing on its own: a red `main` pulls several people onto the same one-line fix at once. Filing the issue first is what would have made that visible — #2133 was open for about twenty minutes before #2134 closed it.

One thing to carry forward whichever way you go: this PR carries no closing reference, so even if it had been the fix that landed, **#2133 would still be open**. `Closes #2133` in the body is what keeps the issue and the fix from drifting apart.

No action needed from me — flagging it so you can decide rather than find out from a conflict.

tinnhlt-gra:
Closing as superseded — this change is already on `main`.

**#2134** (`9d9820b6e`, *"fix(ci-tests): the sweep gate classifies the project clock it was handed (#2133)"*) landed the same waiver key, `"activity_link activity_link_project_last_activity"`, with an equivalent reason. Nothing here adds coverage `main` does not have, so merging it would mean resolving a conflict in order to land an empty change — and spending a CI lane to do it, on a runner pool currently ~18 deep.

That is also why this reads `DIRTY`/`BLOCKED`: the conflict is with the very lines that already merged.

## Worth recording, because this is the fourth copy

#2146, #2150 and #2153 are all the same one-line fix, and #2134 is the one that landed. Four sessions independently hit red `main`, read the same gate failure, and wrote the same waiver.

The interesting part is the timing: **#2133 was open for about twenty minutes before #2134 closed it.** So this was not somebody ignoring an existing issue — the duplicates were in flight before the issue could be triaged. That is what a red `main` does on a shared runner pool, and the cost is not the wasted writing but the lanes each duplicate holds while everything else queues behind it.

**None of the four carried a `Closes #2133` trailer.** So even the copy that was correct would have left the issue open for the next person to re-discover — the pattern tracked in #1737, which is that a fix which names its issue only in prose is indistinguishable from one that never looked.

No action needed on the classification itself; it is on `main` and the gate is green over it.

tinnhlt-gra:
**Correction to my close comment above.** The close itself stands — this PR is superseded by #2134 and its change is on `main` — but two facts I stated about the surrounding episode are wrong, and one of them is unfair to #2134.

- I wrote that **none of the four PRs carried a `Closes` trailer**. False. **#2134 carried `Closes #2133`**, and that trailer is what closed the issue automatically on merge. The three duplicates (this one, #2150, #2153) carried none; #2134 did it correctly.
- I wrote that **#2133 was open for about twenty minutes**. It was open for **69** (03:51:20 → 05:00:48 UTC).
- I implied the duplicates were **in flight before triage could happen**. Also false: #2134 was opened **112 seconds** after the issue, and this PR was opened **40 minutes after that fixing PR already existed**.

The corrected reading, with the timeline, is on **#1737**. The short version: the trailer mechanism worked here. What failed is that an issue with an *open* fixing PR looks identical to an unclaimed one — a `Closes` trailer creates a cross-reference, not a claim, so it sets no assignee and no label until the PR merges. That window was 69 minutes today because CI was ~18 deep, and three sessions walked into it.

Sorry for the noise on a closed PR; a wrong reason recorded next to a correct close is the kind of thing that gets quoted later.

PR #2160, opened 2026-08-21
fix(authz): the geocode sidecar's FK to organization is classified

### LINKED ISSUE #2159: main is red: organization_geocode_state's FK to organization has no visibility decision

tinnhlt-gra:
Closes #2159. **`main` is red without this** — the second such break today, after #2134.

## What is broken

`TestFK_rowScopedTargetsHaveVisibilityDecision` fails on a clean `origin/main` worktree at `9d9820b6e`:

```
FK organization_geocode_state.organization_id -> organization has no visibility decision:
  a reference to a row-scoped record is a read of it — gate it (auth.EnsureLinkTarget)
  or classify it here
```

#2119 added the table; the FK was never classified. The gate exists so a new reference to a row-scoped record cannot arrive without somebody deciding what it discloses, and it did its job.

## The classification

It is a **child row**, and the entry records the reasoning `people/geocode.go` already states rather than inventing one:

- **No caller ever names the organization it keys on.** `enqueueGeocode` runs inside `UpdateOrganization`'s own transaction, for the id that write was already addressing (`people/organization.go:253`), and only when the patch actually moved the address.
- The two later touches, `AddressForGeocode` and `recordGeocodeAfter`, open with `auth.Require` on organization `read` and `update`.

**And the part those gates do not cover, which is the reason the entry is long.** The worker runs under the SYSTEM principal, which passes row scope — so the tenant authority on this path is the explicit `app.workspace_id` predicate every query carries, *not* the job args. `geocode.go` names the attack that closes: a job carrying workspace A with an organization id from workspace B would read and write B.

That predicate is the exception's whole cost, and the entry says so: drop it from one query on this path and the FK becomes a cross-tenant read with nothing else in the way. `gatekit` refuses a reason that does not state what the exception costs, which is the right bar here.

## Verified

```
make test-it DIR=backend/migrations RUN=TestFK_rowScopedTargetsHaveVisibilityDecision
--- PASS (1.27s)

make test-it DIR=backend/migrations RUN='TestFK_|TestSchema_'
3 passing, 0 failing
```

Before the change, on the same worktree at `origin/main`, the first of those fails.

## Note for the geocode author

I classified rather than gated, on the reading above. If the intent was for this path to take a per-caller visibility gate as well, say so on #2159 and this entry should be replaced with that gate — a waiver is a decision on the record, not a way to quiet a gate.

---
## Summary by cubic
Classifies the `organization_geocode_state.organization_id` FK as a child-row exception so `gatekit`’s visibility gate passes. Main was failing `TestFK_rowScopedTargetsHaveVisibilityDecision`; this restores green without changing runtime behavior.

- Adds a `rowScopedFKDecisions` entry in `backend/migrations/schema_fitness_integration_test.go` describing why the FK is a child row.
- Old: FK unclassified, test failed. New: FK classified, `TestFK_rowScopedTargetsHaveVisibilityDecision` and `TestSchema_*` pass.
- Rationale: the row is created within `UpdateOrganization`’s transaction; later touches (`AddressForGeocode`, `recordGeocodeAfter`) require organization read/update via `auth.Require`. The worker runs as SYSTEM, so tenant isolation relies on an explicit `app.workspace_id` predicate in every query on this path; removing that predicate would permit a cross-tenant read/write.

<sup>Written for commit 380d91ba22a203df94571666f3228b9abd057629. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/2160?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/migrations/schema_fitness_integration_test.go
+++ b/backend/migrations/schema_fitness_integration_test.go
@@ -156,6 +156,13 @@ var rowScopedFKDecisions = gatekit.Waive(map[string]string{
 	"conversation_claim.task_activity_id":   "PENDING WRITER: no code writes this table yet. The task an extracted commitment creates is written through the tasks substrate's own gated path, and this entry is replaced with that gate when the routing edge lands",
 	// Owned child rows: the row is an attribute of its visible parent,
 	// written only through the parent's own gated paths.
+	"organization_geocode_state.organization_id": "child row: the sidecar for an organization's own coordinate lookup, and no caller ever names the organization it keys on. " +
+		"The row is created by enqueueGeocode inside UpdateOrganization's transaction, for the id that write was already addressing (people/organization.go), " +
+		"and touched afterwards only by AddressForGeocode and recordGeocodeAfter, which open with auth.Require on organization read and update. " +
+		"What decides the TENANT is not those gates: the worker runs under the SYSTEM principal, which passes row scope, so every query on this path " +
+		"carries an explicit app.workspace_id predicate rather than trusting the job args — geocode.go states the attack that closes, a job carrying " +
+		"workspace A with an organization id from workspace B. That predicate is the exception's whole cost: drop it from one query here and the FK " +
+		"becomes a cross-tenant read with nothing else in the way.",
 	"activity_link.activity_id":  "child row: written only inside LogActivity for the new activity",
 	"lead_score_history.lead_id": "child row: one point in a lead's own score series, appended only from inside the lead's gated write paths — CreateLead/CreateLeadTx (lead:create), UpdateLead (lead:update) and RecomputeLeadScore (lead:update), each of which has already admitted the caller before the append runs in its transaction",
 	// comms_outbound is delivery machinery for one activity, not a


tinnhlt-gra:
## Deadlocked against #2142 — recorded so this red is not misread

This PR's only failures are `deterministic-gates` and the `ci` rollup that aggregates it. `deterministic-gates` fails on:

```
FAIL: backend/internal/compose/integration/harness.go is 505 lines (> 500)
```

That is **#2163**, not this change. This PR adds one entry to `rowScopedFKDecisions` and touches nothing else.

**#2142 fixes #2163 — and it is red on #2159, which this PR fixes.** Each carries the other's inherited failure, and `ci` (the only required context) is red on both, so neither auto-merges. Full evidence on #2142.

`strict_required_status_checks_policy` is `false` on this repo, so merging either would not re-run the other. This will not clear itself.

Resolution proposed on #2142: the seven-line entry here travels into #2142 rather than #2142's 153-line file split coming here. If that happens I will close this PR as superseded — but **#2159 must then be closed by whichever PR actually lands the entry**, or the issue is left open behind a shipped fix.

Everything else on this PR is green: 16 passing, including the full integration lane apart from the harness gate.

tinnhlt-gra:
Superseded by #2142, which merged as `eb8f5edd2` carrying this entry.

Verified on `origin/main` rather than assumed:

```
main                                          eb8f5edd2
schema_fitness_integration_test.go            contains organization_geocode_state  ✅
#2159                                         CLOSED
```

The entry that landed is **not** the one this PR proposed, and the difference matters. Mine claimed *"every query on this path carries an explicit `app.workspace_id` predicate"* — false: two of the four statements in `recordGeocodeAfter` carry none, against a table whose FK is its only tenancy, and the guarded `UPDATE`'s zero-row result is discarded so the unguarded ledger write proceeds regardless. The version on main states what is actually true, and the residue is filed as **#2167** (latent cross-tenant write), which I am taking.

Closing rather than merging: with the entry already on `main` this branch is now an empty change.

PR #2172, opened 2026-08-21
fix(ci-tests): the integration harness splits at its own seam, and main is green again

dungbanguyen:
`main` is red. https://github.com/gradionhq/margince-poc-v1/pull/2154 grew `backend/internal/compose/integration/harness.go` to **505 lines**, five past the 500-LOC cap, so `scripts/check-go-file-length.sh` fails on `main`'s tip — and `deterministic-gates` fails on every open PR with it, whatever the PR changed.

Reproduced against `origin/main` in a clean worktree:

```
FAIL: backend/internal/compose/integration/harness.go is 505 lines (> 500) — split into one-concept-per-file packages
```

(This is the second same-shape breakage on `main` today — see https://github.com/gradionhq/margince-poc-v1/pull/2134 — and the second one found by an unrelated PR going red rather than by its own. https://github.com/gradionhq/margince-poc-v1/pull/2156's two-hourly `main-health` lane is what should catch the next one.)

## The split

Not a new seam — one the package had already drawn and written down. `seed.go` opens by naming it: raw-SQL seeders there, store-mediated fixtures on `Env` in `harness.go`, *"which is the line the two files are split on"*. That second half now has a file of its own.

- **`harness.go`** (327) — what a suite needs before it can act at all: the environment, the pools, the principals.
- **`harnessfixtures.go`** (208) — what it reaches for once it has those: the id wideners and the store-mediated record fixtures.

Naming follows the package's existing convention (`harnessdb.go`, `harnessperms.go`). `seed.go`'s cross-reference is updated in the same commit, since it named the file the fixtures just left.

## One line is not byte-identical

Moving lines makes them **new** to the strict lint set — `.golangci.strict.yml` uses `new-from-merge-base` — so a literal `"person"` that had sat in `MakeCapturePrivate` unremarked became a `goconst` finding on arrival. It now uses `objPerson` and `objOrg`, the constants `harnessperms.go` already declares for exactly this. Everything else moved verbatim.

No behaviour changes. `make check` exits 0, `craft-static` is 4×PASS, and `check-go-file-length.sh` reports `no hand-written Go file exceeds 500 LOC (waivers: 0)`.

---
## Summary by cubic
Splits the integration test harness along its existing seam to get under the 500‑LOC cap and unblock CI on `main`. Previously `harness.go` was 505 lines, tripping `go-file-length` and failing `deterministic-gates`; now the split keeps both files under the cap and CI is green.

- Store‑mediated fixtures and widener helpers moved to `backend/internal/compose/integration/harnessfixtures.go`; environment, pools, and principals remain in `harness.go`.
- Replaces table name literals in MakeCapturePrivate with constants `objPerson` and `objOrg` to satisfy strict lint; otherwise code is moved verbatim.
- Updates `seed.go` comments to reference `harnessfixtures.go`.
- No behavior changes; `scripts/check-go-file-length.sh` and the strict lint lane pass.

<sup>Written for commit a478e35752f40951c9080c0fd05d30f2372b9662. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/2172?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


dungbanguyen:
Superseded by https://github.com/gradionhq/margince-poc-v1/pull/2142, which split the same file while this was in CI — into `harnessseed.go` and `harnessinstallation.go`, taking `harness.go` to 373 lines. Verified against `origin/main`: `check-go-file-length.sh` reports `no hand-written Go file exceeds 500 LOC (waivers: 0)`.

Two of these in one session, so the pattern is worth naming rather than shrugging at. `main` goes red, several people notice it the same way — through an unrelated PR failing — and more than one fix gets written before any of them lands. https://github.com/gradionhq/margince-poc-v1/pull/2156's two-hourly `main-health` lane addresses the detection half, but not the duplication half: nothing tells someone about to write the fix that a fix already exists. Worth considering whether a red `main-health` run should open (or comment on) a tracking issue, so the second person finds it before they start.

Issue #2173, opened 2026-08-21
Every geocode query names organization.workspace_id, a column dropped three days before the feature shipped


tinnhlt-gra:
## The defect

All three SQL statements in `backend/internal/modules/people/geocode.go` filter on
`organization.workspace_id`. **That column does not exist.**

| Line | Statement | |
|---|---|---|
| `:116` | `AddressForGeocode` — reads the address to look up | `AND o.workspace_id = NULLIF(current_setting('app.workspace_id', true), '')::uuid` |
| `:227` | `addressHashInTx` — re-reads the address before recording | same clause |
| `:302` | `recordGeocodeAfter` — `UPDATE organization SET geocode_lat…` | same clause |

```
$ psql -tAc "SELECT column_name FROM information_schema.columns
             WHERE table_name='organization' AND column_name='workspace_id'"
(0 rows)
```

Dropped by core migration **`1787047322_organizations_drop_workspace.up.sql`** —
*"the organization cluster drops the tenant column (ADR-0091 §8 phase D)"*, merged
**2026-08-18** in #1701.

The geocode feature merged **2026-08-21** in #2119, three days later, writing
queries against a column that had already been gone.

## Reproduction, through the real store

```go
e := integration.Setup(t)
err := e.People.RecordGeocode(e.Admin(), someOrgID, "ok", nil, nil, "test", "hash")
```

```
recording the geocode: re-reading the address before recording its point:
ERROR: column "workspace_id" does not exist (SQLSTATE 42703)
```

Every call fails the same way. `AddressForGeocode` fails identically, so the job
cannot even read the address it was enqueued to resolve.

## Impact

**Geocoding never works and never records.** `enqueueGeocode` fires inside
`UpdateOrganization`'s transaction on every address change, so each address edit
enqueues a job that errors at its first statement. No coordinate is ever
resolved, no `geocode_status` is ever written, and the attempt ledger in
`organization_geocode_state` is never touched.

Operators see failing jobs; users see a feature that silently does nothing.

## Why nothing caught it

There is no integration test that drives `AddressForGeocode` or `RecordGeocode`
against a real database. A SQL string naming a dropped column compiles fine, and
the unit lane never executes it. `TestFK_rowScopedTargetsHaveVisibilityDecision`
examines the FK's *classification*, not whether its queries run.

## The second casualty: a gate waiver ratified on a predicate that cannot execute

`rowScopedFKDecisions` ratifies `organization_geocode_state.organization_id`
[...]

Issue #2167, opened 2026-08-21
The geocode sidecar's attempt ledger is written without a workspace predicate


tinnhlt-gra:
## What is unguarded

`recordGeocodeAfter` (`backend/internal/modules/people/geocode.go:297-334`) runs four
statements in one transaction under the **SYSTEM principal**, which passes row
scope. Two carry an explicit tenant predicate and two do not:

| Statement | `app.workspace_id` predicate |
|---|---|
| `AddressForGeocode`'s read (`:108-115`) — joins the sidecar through `organization` | ✅ `o.workspace_id = …` |
| `UPDATE organization SET geocode_lat…` (`:297-303`) | ✅ `AND workspace_id = …` |
| `INSERT INTO organization_geocode_state … ON CONFLICT` — success reset (`:310-315`) | ❌ none |
| `INSERT INTO organization_geocode_state … ON CONFLICT` — failure backoff (`:325-334`) | ❌ none |

`organization_geocode_state` has **no workspace column of its own** — its only
tenancy is the FK to `organization` (`migrations/core/1787320001_organization_geocode.up.sql:66`,
`organization_id uuid PRIMARY KEY REFERENCES organization(id)`).

And the `UPDATE organization` result is **discarded** (`if _, err := tx.Exec(…)`),
so a zero-row update does not stop the transaction: control falls straight
through to the sidecar INSERT.

## What that means

A geocode job carrying an `organization_id` from another workspace would:

1. update **zero** `organization` rows — the predicate holds, nothing disclosed;
2. then **write that organization's attempt ledger anyway** — creating the
   sidecar row if absent, or resetting `attempts`/`next_attempt_at` if present.

**This is a write, never a read.** Nothing leaks. The effect is on the other
tenant's geocode scheduling: a success-path insert clears their backoff, a
failure-path insert bumps their attempt count toward the cap that stops
re-geocoding.

## Reachability — latent, not live

`enqueueGeocode` only ever mints the id that `UpdateOrganization`'s own
workspace-checked transaction was already addressing
(`people/organization.go`), so no production path produces a foreign id today.
Exploiting it needs a forged or corrupted job payload — which is precisely the
attack the geocode code names in its own comments as the reason the read
predicate exists.

Filed public rather than as an advisory on that basis: no reproduction exists
against the shipped enqueue path, so this is the exception's cost written down,
not a live breach.

## Why it is filed now

Found while carrying #2159's FK classification into #2142. The waiver's first
[...]

PR #2174, opened 2026-08-21
fix(agents): four review findings on the merged archive change

tinnhlt-gra:
## Why

A third review pass over #2142's range, run **after it merged** because my own review turned out to be six commits stale — it had seen commit 1 of 7, and the 1047 insertions since included a public contract change. That gap was found by applying to my own work a check I had been carefully applying to CodeRabbit's.

Four findings. Two are text corrections; two are real.

## 1 · The overlay REST shadow answered one header two ways *(medium — a regression #2142 introduced)*

Forwarding `If-Match` into `archiveInMode` reached `overlay.Provider.ArchiveAt`, which **refuses** a version it cannot honour. Correct on the seam, wrong through that door:

| Same client, same record, same overlay installation | before #2142 | after #2142 |
|---|---|---|
| `PATCH /v1/people/{id}` with `If-Match: 7` | 200, header ignored | 200, header ignored |
| `DELETE /v1/people/{id}` with `If-Match: 7` | archived | **422** `unsupported_by_sor` |

And 422 is not among the responses `archivePerson` declares. `overlayUpdate`'s own doc twenty lines above already states the house answer — *"the header is accepted and discarded… State the consequence plainly rather than only the cause"* — so the archive now matches its sibling. The gap both verbs share is **one** question, not two divergent behaviours.

## 2 · The doc justifying a removed backstop rested on something unguaranteed *(low)*

#2142 dropped `refuseUngovernedAgentEgress` from `RefuseArchive`, arguing that `refuseStagingElsewhere` refuses an overlay staging first. That holds only while the seam read and the refusal resolve the **same mode** — and they do not: `Dispatcher.Read` uses the **cached** mode (5s TTL), `RefuseArchive` uses `isOverlayUncached`. Within 5s of a flip, on a replica the local `Invalidate` never reached, the read routes native and passes while the refusal routes overlay. The doc now claims only what is actually guaranteed.

## 3 · A comment named a mechanism that does not exist *(low)*

It said an undeclared contract parameter is never parsed, so the handler passes nil. **The handlers read the header directly** with `httperr.IfMatchVersion` and ignore their `Params` struct entirely — an undeclared parameter would still be parsed. What left the archive unpinned was the handler passing a literal `nil`. Declaring `If-Match` is about the contract telling the truth, and about `409` being a declared response.

This is the comment-asserts-a-false-invariant class this repo polices, committed inside a PR about that class.

## 4 · The geocode waiver overstated its cost — and measuring it found worse

Both entry points funnel through `recordGeocodeAfter`, whose **first** statement is a workspace-filtered read that aborts before either unguarded `INSERT`. The cross-tenant write I described in #2167 is unreachable; **#2167 closed as not-a-defect.**

That sentence has now been wrong twice in opposite directions in one afternoon, by two sessions reading carefully. So I went to *measure* it rather than read it again — and the fixture found **#2173**: `organization` has had no `workspace_id` column since core `1787047322` dropped it (2026-08-18), and `geocode.go` shipped **three days later** naming it in all three of its statements. Every one fails with `SQLSTATE 42703`. **The predicate this waiver rests on has never executed**, and geocoding is dead on `main`.

⚠️ **The fixture is parked, not shipped.** While the column is missing it cannot distinguish *"the workspace guard refused"* from *"the query cannot run"* — so it would read **green exactly when the subject is most broken**. That is an inverted test, not a weak one. The file says so, and it goes in when #2173 lands.

My first attempt at that fixture was itself a false pass — it drove a nonexistent org id, so the sidecar `INSERT` died on its own foreign key and the test agreed with a deliberately broken guard. Caught only by mutating.

## Also: the coverage SonarCloud was right about

Sonar reported new-code coverage at **70.8%** against an 80% bar on #2142. Not a required check, which is exactly why it was worth reading rather than merging past: **`ArchivableTypes` and `RefuseArchive` on the Dispatcher had ZERO coverage** — the two questions #2016 and #2014 turn on. An untested route there is how the answer comes from the native provider while the write goes to the mirror.

Three tests over the existing `cachedModeDispatcher` fixture, which leaves the native provider nil so a wrongly-routed question panics rather than quietly passing. Mutation-verified: routing `ArchivableTypes` to the other mode panics.

## Gates

`make check-backend` exit 0 · `craft static --strict` 0/0/0 · archive integration suites green against real Postgres.

---
## Summary by cubic
Aligns overlay archive semantics and closes a pinning gap: DELETE now ignores a caller’s If-Match like PATCH (200), while a redeemed approval’s version pin is still forwarded so the store can refuse stale approvals. Adds dispatcher tests to prove archive questions route by the current mode, not a stale cached one, and extracts `archivePrecondition` to separate caller headers from approval pins.

- Split header handling in `overlaywriteshadow.go`: ignore a caller `If-Match`; when `agents.ApprovalRedeemed` is set, forward the pin via `IfVersion` to the store. New `archivePrecondition` helper documents the untestable redeemed branch.
- Tests: `overlay_write_shadow_integration_test.go` adds DELETE-vs-If-Match gate; new `dispatcherarchive_test.go` pins that `ArchivableTypes` answers overlay’s three, `RefuseArchive` re-reads mode (not cache), and refusing `project` returns `apperrors.ErrUnsupportedBySoR` (native provider left nil to panic on misroutes).
- Docs and test hardening: clarify cache vs uncached mode in `dispatcherarchive.go` (write backstop stays); agents guard tests now assert distinct sentinels and ordering so “held elsewhere” beats the executor probe; update archive pin comments to reflect `httperr.IfMatchVersion`.
- Geocode waiver: note `organization.workspace_id` is missing and those statements do not execute (#2173); fixture remains parked until fixed.

**Rollout**
- No migrations or config. Overlay DELETE with caller `If-Match` now returns 200; approvals remain pinned and are refused when stale.

<sup>Written for commit aeda8a25e677a63675602e999afaf166b54af3df. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/2174?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/internal/compose/archivepin_integration_test.go
+++ b/backend/internal/compose/archivepin_integration_test.go
@@ -290,20 +290,7 @@ func archiveOverREST(as context.Context, t *testing.T, e *integration.Env,
 	return rec
 }
 
-// The geocode sidecar's cost is NOT measured here, and #2173 is why.
-//
-// This is where a fixture would go proving what `rowScopedFKDecisions`'s
-// waiver on organization_geocode_state.organization_id claims — that the
-// app.workspace_id predicate is what keeps the FK from being a cross-tenant
-// read. Writing it is how #2173 was found: `organization` has had no
-// workspace_id column since core 1787047322 dropped it (2026-08-18, ADR-0091
-// phase D), and geocode.go shipped three days later naming it in all three of
-// its statements. Every one fails with SQLSTATE 42703.
-//
-// So the fixture cannot exist yet, and shipping it anyway would be worse than
-// having none: it could not tell "the workspace guard refused" from "the query
-// cannot run", which means it would read GREEN exactly when the subject is most
-// broken. An inverted test, not a weak one.
-//
-// It goes in when #2173 lands. The waiver's own text needs revisiting then too
-// — its stated cost is currently unmeasurable rather than mis-stated.
+// No geocode fixture lives here. #2173 records why, and carries what a future
+// one has to prove: while organization.workspace_id is missing, a fixture
+// cannot tell "the workspace guard refused" from "the query cannot run", so it
+// would read GREEN exactly when the subject is most broken.
--- a/backend/internal/compose/archivepin_integration_test.go
+++ b/backend/internal/compose/archivepin_integration_test.go
@@ -221,11 +221,16 @@ func archivedAt(t *testing.T, e *integration.Env, table string, id ids.UUID) boo
 // closed on one door only.
 //
 // The agent gate forwards a released approval's version as the request's own
-// If-Match (compose/agentgatestaging.go), so an archive route that does not
-// DECLARE the parameter never parses it: the generated wrapper drops the
-// header, the handler passes nil, and the released archive runs unconditioned
-// exactly as before. Declaring it is what makes the forward mean something,
-// and only a request through the real handler proves the chain.
+// If-Match (compose/agentgatestaging.go). What left the archive unpinned was
+// not the contract: these handlers read the header straight off the request
+// with httperr.IfMatchVersion and ignore their Params struct entirely, so an
+// undeclared parameter would still have been parsed. It was the handler
+// passing a literal nil to the store.
+//
+// Declaring If-Match on the route is therefore about the CONTRACT telling the
+// truth — a client cannot be expected to send a precondition the API does not
+// advertise, and the 409 it can now provoke has to be a declared response.
+// Only a request through the real handler proves the chain either way.
 func TestTheRESTArchiveHonoursAStaleIfMatch(t *testing.T) {
 	e := integration.Setup(t)
 	native := NewProvider(e.Pool)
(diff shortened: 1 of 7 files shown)


tinnhlt-gra:
@coderabbitai review

No review has landed on this PR yet — the check reads `pass` with the description *"Review rate limited"*, which is green with nobody having looked. Requesting one against `a773fd40f`.

Where a reviewer's attention is best spent, since four of the five files are tests and only one is production code:

- **`overlaywriteshadow.go`** is the only behaviour change, and it is a REVERT of something you flagged on #2142. Forwarding `If-Match` there reached `overlay.Provider.ArchiveAt`, which refuses a pin it cannot honour — so an overlay client got 200-and-ignored from `PATCH /people/{id}` and **422** from `DELETE /people/{id}`, same record, for a request that archived fine before. It now matches `overlayUpdate`'s documented accept-and-discard twenty lines above. Worth checking I picked the right side of that consistency, since the alternative is making the update shadow refuse too.
- **`schema_fitness_integration_test.go`** carries a waiver whose rationale has now been rewritten twice. The current text says the predicate it rests on is present on the two statements that could disclose and absent on the two that write — and #2173 records that the column those statements name does not exist at all.
- **`archiveauthority_test.go`** and **`dispatcherarchive_test.go`** tighten assertions that were passing for the wrong reason: one satisfied by any error, one by a sentinel with three producers on the same path.

This PR is itself the product of a review pass, so the useful question is less "is it correct" than "does it claim more than it proves".

PR #2182, opened 2026-08-21
fix(ci-tests): the vocabulary test asserts the operator answers, not that it cannot

### LINKED ISSUE #2179: main is red: within_radius is published without its unavailability marker
## What fails

`TestTheComposedMCPMountPublishesTheQueryVocabulary` fails on a clean checkout of
`origin/main`:

```
queryvocabulary_integration_test.go:93: within_radius is not published as
unavailable, so a caller cannot tell it apart from one that works
```

Reproduced in a pristine detached worktree at `origin/main` (`1a1562a26`) with no
local changes:

```
make -C backend test-it \
  DIR=backend/internal/compose/integration/agentaccess \
  RUN='TestTheComposedMCPMountPublishesTheQueryVocabulary'
--- FAIL: TestTheComposedMCPMountPublishesTheQueryVocabulary (3.18s)
```

## Why it matters

The test states the obligation in its own failure message, and the obligation is
the interesting part: an operator with no geocoder configured has a query
grammar that can *parse* `within_radius` and cannot *answer* it. Publishing the
operator without marking it unavailable means an agent cannot tell a filter that
will work from one that will silently return nothing — which is the exact
failure the geocode work set out to prevent (`platform/geocode/geocode.go` opens
by naming it: "query surface could parse and not answer: within_radius was in
the grammar").

So this is not a cosmetic vocabulary detail. Either the publication path lost the
unavailability marker, or the test's expectation no longer matches how
availability is expressed — and until it is settled, nobody can tell which.

## Where it comes from

`within_radius` and its availability posture arrive with the geocode work
(`platform/geocode/`, `compose/geocodejobs.go`, `modules/search/queryoperand.go`,
`modules/agents/results.go`). The most likely candidates are the geocode merge
(#2119) or the vocabulary-publication change that followed it, but I have not
bisected it — the reproduction above is on the tip, and that is what makes it
actionable.

It fails identically on every branch cut from this tip, so it reds unrelated
PRs; the integration lane is a required check.

## Not this, for the avoidance of doubt

Found while running the integration lane for an unrelated branch
(`feat/agent-activity-is-live`, a personal agent-activity read). That branch
touches nothing in the query vocabulary, geocode, or search paths — verified with
`git diff origin/main --name-only` — and the failure reproduces on `origin/main`
with that branch nowhere in sight.

dungbanguyen:
**Main's integration lane is red** on
`TestTheComposedMCPMountPublishesTheQueryVocabulary`, and has been since #2171.
Closes #2179.

```
queryvocabulary_integration_test.go:93: within_radius is not published as
unavailable, so a caller cannot tell it apart from one that works
```

## The production code is right; the assertion went stale

#2171 emptied the published `unavailable` list **on purpose**, and says why at
the field:

> within_radius used to be declared permanently unavailable here because no
> record carried coordinates; companies do now, so the operator answers for them.

The test still asserts the old SEARCH-AC-17 shape. Declaring an operator
unavailable *when it works* sends a caller to a text match on a city name —
which is precisely the wrong answer that declaration existed to prevent. So the
test now pins the inverse.

## Both halves, because either alone passes for the wrong reason

- an **empty unavailable list** is also what dropping the operator altogether
  would produce;
- a **published operator** would also appear if the deployment still refused it.

So it asserts `within_radius` is absent from `unavailable` **and** that the
`organization` target offers a geo field admitting it.

`dealVocabulary`'s lookup becomes `targetVocabulary`, keyed by name — the new
assertion needs the organization target, and a second copy of a four-line
`IndexFunc` is how a file grows two spellings of one thing.

## Verification

This is the last thing standing between #2136 (removing the 19 MB binary from
`main`) and a green lane.

---
## Summary by cubic
Updates the query vocabulary integration test to reflect that `within_radius` now answers due to geocoded companies, fixing the red integration lane. Previously the test expected `within_radius` in the published `unavailable` list (SEARCH-AC-17); it now asserts the operator is absent from `unavailable` and that the `organization` vocabulary exposes a `geo` field admitting `within_radius`.

- Production behavior is unchanged; this corrects stale assertions only.
- Adds `targetVocabulary(name)` and has `dealVocabulary` delegate to it to fetch targets by name.
- Asserts both absence from `unavailable` and presence on the `organization` `geo` field to prevent false positives.

<sup>Written for commit 433417af8f4fde25559a5e9ee22a0c68a4ed4974. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/2182?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/internal/compose/integration/agentaccess/queryvocabulary_integration_test.go
+++ b/backend/internal/compose/integration/agentaccess/queryvocabulary_integration_test.go
@@ -86,11 +86,26 @@ func TestTheComposedMCPMountPublishesTheQueryVocabulary(t *testing.T) {
 		t.Error("the published deal vocabulary has no derived organization hop")
 	}
 
-	// The declared-but-unanswerable operator arrives declared (SEARCH-AC-17).
-	if !slices.ContainsFunc(doc.Unavailable, func(u vocabularyUnavailable) bool {
-		return u.Op == "within_radius" && u.Answers == search.CodeDistanceRankingUnavailable
+	// SEARCH-AC-17 inverted, which is what #2171 earned. within_radius was
+	// declared permanently unavailable while no record carried coordinates;
+	// companies are geocoded on ingestion now, so the operator ANSWERS — and
+	// declaring it unavailable would send a caller to a text match on a city
+	// name, the exact wrong answer the declaration existed to prevent.
+	//
+	// Both halves, because either alone passes for the wrong reason: an empty
+	// unavailable list would also be produced by dropping the operator
+	// altogether, and a published operator would also appear if the deployment
+	// still refused it.
+	if slices.ContainsFunc(doc.Unavailable, func(u vocabularyUnavailable) bool {
+		return u.Op == "within_radius"
 	}) {
-		t.Error("within_radius is not published as unavailable, so a caller cannot tell it apart from one that works")
+		t.Errorf("within_radius is still published as unavailable, so a caller is told to avoid an operator that answers: %+v", doc.Unavailable)
+	}
+	org := targetVocabulary(t, doc, "organization")
+	if !slices.ContainsFunc(org.Fields, func(f vocabularyField) bool {
+		return f.Kind == "geo" && slices.Contains(f.Ops, "within_radius")
+	}) {
+		t.Errorf("the organization vocabulary offers no geo field admitting within_radius, so nothing tells a caller the operator is theirs to use: %+v", org.Fields)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


tinnhlt-gra:
**This may already be merged — `c414f122f` (#2178) is on `main` and appears to make the same change.** Flagging early rather than after you spend more on it.

```
$ git show origin/main:backend/internal/compose/integration/agentaccess/queryvocabulary_integration_test.go
 89:  // within_radius ANSWERS now (#2171: a company carries coordinates), so the
 95:  if slices.ContainsFunc(doc.Unavailable, func(u vocabularyUnavailable) bool {
 98:      t.Error("within_radius is still published as unavailable; …")
102:  org := targetVocabulary(t, doc, "organization")
106:      t.Error("no organization field publishes within_radius, …")
114:      t.Error("the deal vocabulary publishes within_radius; a deal is never anywhere, …")
```

That is the inverse assertion, both halves (absent from `unavailable` **and** the organization target offering the field), plus the `dealVocabulary` → `targetVocabulary` rename — the same three things your description lists, including the deal-never-offers-it case. Worth diffing against `origin/main` before going further; if it is the same change, this closes as already-landed and #2179 with it.

Not wasted either way: I hit the same failure from a different direction on #2135 and can confirm your reading of the cause. #2171 emptied the list deliberately and the assertion went stale.

One thing that may save you time if you are still chasing it — **CI builds `refs/pull/N/merge`, the branch merged with `main`, not your head sha.** A branch whose merge-base predates #2171 fails in CI while passing every local run, because locally you only ever have your own tree, which is internally consistent. I burned eleven local runs on that before checking `git log origin/main <merge-base>..`. If your branch is behind, a rebase may be the whole fix.

dungbanguyen:
Superseded by #2178, which fixed the same failure on main while this was in review — and covered more than this branch did.

Mine asserted two halves: within_radius absent from the deployment-wide `unavailable` list, and published on the organization target. #2178 asserts a third that I missed — that it stays ABSENT from the deal vocabulary, because a deal is never anywhere, so offering it there invites a query that can only be refused.

Closing unmerged. Nothing here is worth carrying forward.

Issue #2184, opened 2026-08-21
vault_secret has no tableOwners entry, so its writers are ungated


dungbanguyen:
`backend/tableownership_test.go`'s `tableOwners` map does not name `vault_secret`, and the gate does not walk `internal/platform/keyvault` — so the two packages that write the table are invisible to it:

- `backend/internal/platform/keyvault/local.go` (the local provider's ciphertext store)
- `backend/internal/compose/datasweep.go`

The gate walks `internal/modules`, `internal/compose`, and exactly two named platform directories (`settingsStoreDir`, `extSecretsStoreDir`). `keyvault` is neither, so its writes are never compared against an owner, and the table having no owner entry means the compose write is not checked either.

**Impact is narrow today** — nothing is writing the wrong table — but the gate's whole purpose is that a *second* writer of a table cannot land unnoticed, and for this table it can.

**Found while** placing every core table into the migration baseline (every table had to resolve to exactly one owning module; this was the only one that resolved to none). Not fixed there: it is a different gate with its own scope decision to make, namely whether `internal/platform/keyvault` joins the walked set or `vault_secret` gets an entry pointing at it.

Two options, either of which closes it:
1. Add `internal/platform/keyvault` to the walked platform directories and give `vault_secret` an owner entry naming it.
2. Derive the walked platform set from "platform packages that write rows" instead of a hand-named list of two, which is the version that does not need this issue filed again for the next one.

dungbanguyen:
Correction to this issue's own text, recorded before it closes: it claims `compose/datasweep.go` writes `vault_secret`. **It does not.** That file names the table in the sweep's *preserve* list — the opposite of a write. The finding this issue reports is still real (the gate never walked `internal/platform/keyvault`, so the local provider's INSERT and DELETE were compared against nothing), but the second writer I named was never there.

Fixed in https://github.com/margince/margince/pull/2546.

Issue #1844, opened 2026-08-19
The phase-D migrations take ACCESS EXCLUSIVE on live tables without a lock_timeout


tinnhlt-gra:
`TestEveryMigrationTakingABlockingLockBoundsHowLongItWaits` (added in #1816)
requires any migration taking a writer-blocking lock to bound how long it waits.
It is armed from a baseline above `main` at the time it landed, so it binds new
work only. The backlog it cannot clear is filed here.

## The hazard, once

The problem is the lock's ACQUISITION, not its duration. A pending strong request
queues behind whatever transaction is already running, and every request arriving
after it queues behind the request. So one idle-in-transaction session turns a
migration into an installation-wide write stall on that table for as long as the
migration is willing to wait — which, without `lock_timeout`, is forever.

`core/0139` writes this out in full and sets `lock_timeout = '3s'`; `0147` and
`0165` follow it. `SET LOCAL lock_timeout = '3s'` turns the stall into a fast,
loud failure: the transaction rolls back whole and the deploy retries.

## What is unbounded today

Roughly a hundred migrations, but they are not equal:

**The ones that matter — phase D (ADR-0091 §8), still landing.** Each is
`ALTER TABLE … DROP COLUMN` on tables that are live and hot:

| Migration | Tables |
|---|---|
| `1787123080_authz_drop_workspace` | `team`, `team_membership`, `record_grant`, `extension_secret`, `onboarding_wizard_state` |
| `1787128082_singletons_drop_workspace` | (see file) |
| earlier phase-D members | `person`, `organization`, `deal`, `activity`, `relationship`, … |

These are worth fixing: they have not all been applied everywhere yet, so the
timeout still changes an outcome. And more are coming — the phase is not done.

**The ones that mostly do not.** Migrations from `0002`–`0200` that created their
tables and have been applied on every installation for months. Adding a timeout
to a version that will never re-run changes nothing for any deployed database.

## Why this was not fixed in #1816

Two reasons, both about blast radius rather than difficulty. A rule introduced in
a PR about employment currency should not begin by demanding edits to other
people's already-merged migrations. And several sessions are writing to this tree
concurrently — a gate that failed on `main`'s newest migration would have blocked
all of them on a file none of them wrote.

## Done looks like

- Every phase-D migration sets `lock_timeout` before its first `ALTER TABLE`,
  ideally as part of the phase rather than as a sweep behind it.
- `lockTimeoutBaseline["core"]` moves down to the first phase-D version once they
[...]

PR #2189, opened 2026-08-21
refactor(platform): core's history is one baseline, and the schema it builds is committed

dungbanguyen:
core/ carried **318 migration pairs** and custom/ 24 — ~17k lines of SQL whose end state nothing in the tree described. 69 core files still declared row-level security policies a later migration had already retired, which is the shape of the problem: the namespace recorded the *path*, and the path had stopped matching the destination.

core/ is now **27 pairs** and custom/ **3**.

| | before | after |
|---|---|---|
| `backend/migrations/core/` | 318 pairs | 27 pairs |
| `backend/migrations/custom/` | 24 pairs | 3 pairs |
| `extensions/*/migrations/` | 12 pairs | **untouched** |

## ⚠️ Your database will refuse to migrate, and that is the design

The baseline reuses version `0001`, whose ledger row on an existing database names `foundation`. `dbmigrate.assertLedgerMatches` refuses and says `make dev-fresh`. That is the outcome such a database needs — its schema was built by migrations that no longer exist and there is no forward repair. A baseline stamped *above* the history would instead be applied ON TOP of a schema it was meant to create, and fail halfway with nothing to say why.

- **Dev:** `make dev-fresh`
- **Staging:** the same against its DSN, then re-bootstrap from `margince.yaml`
- **Integration lane:** `DROP DATABASE margince_test` once. `lib-testdb.sh` says outright that it cannot heal a template that DIVERGED rather than fell behind, and this one diverged.
- **In-flight branches:** no re-stamp needed. A ten-digit unix stamp still sorts above `0027`.

## The proof is a script, not a claim

`scripts/migration-baseline.sh verify origin/main` builds one database from a named ref's migrations and one from this tree's, then diffs normalized `pg_dump`s of both:

```
OK: 30 migration(s) build the schema origin/main's 342 build — byte-identical
```

Column order, defaults, constraint and index names, trigger bodies, grants and reference-row contents all compared exactly. Only two value kinds are masked, both generated per run: a `uuidv7()` key and a `now()` timestamp. **The script is kept** — it is how the next consolidation gets checked.

## What replaces the history as a description of the schema

`backend/migrations/testdata/head_catalog.txt` commits the schema itself and `TestMigrationsBuildTheCommittedSchema` compares the two. Its projection is built from the **same `catalogParts` query** `headSchema`'s fingerprint already used — one definition, so the in-process check and the committed expectation cannot come to cover different object classes. A migration that changes head changes that file in the same commit, and the diff a reviewer reads IS the schema effect. `-update-catalog` regenerates it.

## Layout

`0001` extensions + the `ext` schema → `0002` functions → one file per owning module (from `tableownership_test.go`'s `tableOwners`, so a table's baseline file and the module allowed to write it cannot disagree) → `0026` functions that read a table (a `LANGUAGE sql` body is resolved at creation, so these cannot precede their tables) → `0027` everything needing every table to exist: constraints, indexes, triggers, grants, reference rows.

A table in neither `tableOwners` nor a small explicitly-reasoned map is a **hard error** in the generator, never a default bucket.

## Two defects this surfaced, fixed here

**1. A lost update on `voice_corpus_source`.** The table is version-pinned, but `updateVoiceSource` and `DeleteSource` ran by-id UPDATEs with no compare-and-set: a writer committing between the version read and the write had its change folded away by a `coalesce` reading the new row against a decision made on the old one. `updateguard_test.go` could not see it because the column arrived by `ALTER TABLE` in core/0107 and its derivation only scans inside `CREATE TABLE` blocks — the baseline puts it in the block. Both statements now carry `AND version = $n`, and the gate credits a version predicate as the CAS its own error message already offered.

**2. Two `project` CHECKs with no refusal message.** `project_errors_test.go` read only the migration that CREATED `project`, so CHECKs added later were never held to the rule. Reading head finds them; both are ratified as contract-unreachable **with the reason per entry**, because a bespoke message for a constraint the contract prevents reaching is a branch no caller can execute.

## Fourteen replay tests retired (~2,100 lines)

Backfills, repairs, rollbacks and the archived-tenant gate — every one replays a prefix of a history that no longer exists, against a pre-baseline database that no longer can. What they asserted about *head* is now covered more broadly by the committed catalog.

`migrationrole_integration_test.go` **stayed and earned it**: it caught two baseline defects the equivalence proof could not, because both sides of that proof run as superuser — a `COMMENT ON EXTENSION` needing extension ownership, and `ALTER DEFAULT PRIVILEGES FOR ROLE`, neither of which the deployed non-superuser migration role may execute.

## Why the generated SQL is in this repo's dialect

Nineteen fitness tests across five packages derive schema facts by *parsing the migration text*. They read `CREATE TABLE public.deal` and `CHECK ((status = ANY (ARRAY['open'::text])))` as "the column is not there" — a false **negative**, the direction that passes silently once someone relaxes the assertion. Teaching nineteen parsers a second dialect would leave every future one having to know both. Every rewrite is semantics-preserving, and the equivalence proof is what says so rather than a comment.

## Gates

| gate | result |
|---|---|
| `make check-fe` | exit 0 |
| `make test-integration` | `OK: integration passed with 0 skips (42 packages, parallel)` |
| `make craft-static` | 0 blocker, 0 major, 0 minor across all four trees |
| `migration-baseline.sh verify` | byte-identical |

### The version gate's escape hatch, and why it is safe to commit

`check-migration-versions.sh` cannot tell a consolidation from the collision outage it exists to report — both look like "version 0001 claimed by two different migrations". So the consolidation **declares itself** with `MIGRATION_VERSIONS_BASELINE_RESET=1`, set in `ci.yml`'s `deterministic-gates` env.

**The declaration is checked, not believed.** CI's environment is a committed file, so "set it for one PR" is not something the environment can express — and a forgotten flag leaves this gate reporting findings it never enforces, permanently and looking green. So:

- A consolidation replaces a namespace wholesale, sharing **no** `(version, name)` pair with the base. A broken rebase — the case that would otherwise ride in on the same flag — drops some migrations and leaves the rest at the versions and names they already had, so it shares many. The gate requires zero shared pairs, **per namespace**, before honoring the flag.
- **It self-expires.** Once this merges, the base IS the baseline, every version keeps its name, and the check refuses the declaration on its own. The gate re-arms automatically; there is no cleanup PR to forget.
- **It fails closed.** Admitting is what disarms the gate, so an empty row set or a comparison yielding anything but a number refuses. `wc -l` rather than `grep -c` deliberately: `grep -c` exits 1 when it counts zero, which is the *admitting* case, so a failure and a true consolidation would share an exit status.

Verified both directions: a true consolidation is admitted; planting one surviving base migration refuses it and names the count. That logic has no self-test, which every sibling gate in this repo does have — filed as [#2190](https://github.com/gradionhq/margince-poc-v1/issues/2190) rather than grown into this diff.

## Rulebook

`CLAUDE.md`'s DO NOT TOUCH now records **two** authorized exceptions, each naming the condition that made it safe rather than the fact that it happened. This one is available only while no installation holds data somebody cannot rebuild. Both files' citations of migration `0217`/`0190`/`20260806120000` are gone — those version numbers no longer exist, and a rule that cites a dead migration teaches a reader to go looking for it.

## Follow-ups

- [#2184](https://github.com/gradionhq/margince-poc-v1/issues/2184) — `vault_secret` has no `tableOwners` entry, so `keyvault`'s writes are ungated. Found because every table had to resolve to exactly one owner.
- [#1844](https://github.com/gradionhq/margince-poc-v1/issues/1844) **closed** — the consolidation deleted every migration it was about. The lock-timeout pin's comment no longer justifies itself by that backlog; leaving that rationale in place after deleting its subject would have been a false explanation in the source.


dungbanguyen:
**No AI review coverage on this PR, and it should be said out loud rather than read off a green check.** CodeRabbit posted `pass` with *"Review skipped: 779 files exceed the limit of 300"* — a pass that means "did not look", which is the one kind of green worth distrusting.

The file count is not padding: 342 migration pairs deleted and 30 written accounts for nearly all of it. But a reviewer cannot lean on the bots here, so where the real risk sits:

1. **`scripts/migration-baseline.sh`** and the claim it makes. Everything else rests on the equivalence proof being an honest comparison, so the thing worth reading closely is what `strip_preamble` and `dump_data` normalize AWAY — three preamble classes and exactly two value kinds (a `uuidv7()` key, a `now()` timestamp). If something meaningful is being masked there, the byte-identical result means less than it appears to.
2. **The two product fixes**, which are the only behaviour changes: `voice_source_mutations.go` (a version predicate added to two UPDATEs — this can now return `ErrVersionSkew` where it previously always succeeded) and the ratified-unreachable list in `project_errors_test.go`.
3. **`backend/migrations/core/0027_baseline_constraints.up.sql`** — 2.6k lines, and the only file where an ordering mistake would not have shown up as a failure to apply.

The 27 module files are mechanically derived and individually dull; skimming them is fine. Reading the reversals (`*.down.sql`) is worth more, and the up→down→up round trip in `schema_integration_test.go` exercises them.

Verified locally: `check-backend` and `check-fe` exit 0, `test-integration` reports `0 skips (42 packages)`, `craft-static` clean across all four trees, and `migration-baseline.sh verify origin/main` byte-identical.

Reminder for whoever runs CI: this branch needs `MIGRATION_VERSIONS_BASELINE_RESET=1` for `check-migration-versions.sh`, for the reason in the PR body.

dungbanguyen:
**Update after the reshape to a single baseline file, and after the two review passes.**

The file count moved (725 now, was 779), so the figure in my earlier comment is stale — but the substance is unchanged: **CodeRabbit still skipped**, `Review skipped: 725 files exceed the limit of 300`. A `pass` that means "did not look".

What changed since that comment:

- **Layout**: `core/` is now ONE file (`0001_baseline.{up,down}.sql`, ~5.9k lines) rather than 27 per-module files, and `custom/` likewise one pair. The order inside each file is a dependency order and is documented at the top, since that is the only structure a reader now has.
- **Five defects found and fixed** by the review round, four of which I had shipped. Detailed in the PR body; the one worth a reviewer's attention is that `project_phase_check` turned out to be **reachable** — I had exempted it on the grounds that the contract declares `phase` an enum, which was inference rather than verification. `httperr.Decode` does not validate enums and no request-validator middleware exists, so `{"to_phase":"bogus"}` reached the CHECK and returned the constraint's own name to the caller.

Where the risk still sits, updated for the new shape:

1. **`backend/migrations/core/0001_baseline.up.sql`** — the whole schema in one file. An ordering mistake inside it is the one class of error that would not have shown up as a failure to apply, and the equivalence proof cannot distinguish "correct order" from "an order that happens to work".
2. **`scripts/migration-baseline.sh`** — everything rests on this being an honest comparison. Read what `strip_preamble` and `dump_data` normalize AWAY: three preamble classes and exactly two value kinds (a `uuidv7()` key, a `now()` timestamp). Defect 4 in the PR body is precisely a real difference that hid inside that masking, so the blindness is demonstrated, not hypothetical.
3. **`backend/internal/modules/ai/voice_source_mutations.go`** — the only behaviour change. Two UPDATEs can now answer `ErrVersionSkew` where they previously always succeeded.
4. **`scripts/check-migration-versions.sh`** — `reset_admitted` decides whether a gate enforces. It fails closed and requires both no-shared-versions and a collapse, but it has no self-test ([#2190](https://github.com/gradionhq/margince-poc-v1/issues/2190)).

Verified locally on this head: `check-backend` 0, `check-fe` 0, `test-integration` `0 skips (42 packages)`, `craft-static` clean across all four trees, `migration-baseline.sh verify origin/main` byte-identical, and an up→down→up round trip leaving nothing behind.

The `MIGRATION_VERSIONS_BASELINE_RESET` note in my earlier comment is obsolete: it is committed in `ci.yml` now, and the script checks the declaration rather than believing it — see the PR body for why that is safe to leave in place.

PR #2207, opened 2026-08-21
perf(ci-tests): the harness stops paying full-strength Argon2 per test, and seed-demo's copy of those parameters gets a gate

tinnhlt-gra:
Two changes, one subject: the Argon2id cost parameters, and who is allowed to know them.

**D3 is the last unshipped item in the `integration-lane-cost` design.** D2, D7, D8 and D9 all turned out to be already merged — I checked each against the tree rather than building from the doc, which is the only reason this PR is two commits and not six. Notes at the bottom.

---

## 1 — `f62347d6c` the harness stops paying full-strength Argon2 per test

The fixture runs the real first-boot ceremony on every test — `EnsureInstallation`, login, change-password, login again — which is **five derivations before a test asserts anything**. `argon2.processBlocks` measures **28.5% of all CPU samples** in the overlay package.

The ceremony stays. The harness comment is right that a suite arranged around a state the product cannot reach would prove nothing, and none of that depends on how long a hash takes.

| | per derivation |
|---|---|
| `t=2 m=19MiB p=1` — shipped | **18.861 ms** |
| `t=1 m=64KiB p=1` — integration build | **48 µs** (395×) |

`internal/modules/identity` end to end: **17.830 s → 11.676 s**. Full lane: **254 s → 170 s**.

64 KiB rather than the 8 KiB spec floor (`m >= 8*p`) deliberately — the floor saves a further 31 µs, nothing against ~2200 derivations, and sits exactly on the boundary a future argon2 release is most likely to tighten.

### A build tag, not a config field

`go build ./...` carries no tags, so `params.go` is what compiles into api, worker, migrate and the desktop bundle. The cheap values **do not exist** in a shipped binary rather than being refused by one at runtime.

A config seam would instead put the weak value into a deployment's vocabulary — and a PHC hash carries its own parameters, so one hash written under a wrong config verifies successfully **forever**, because nothing re-hashes on login.

### Three constants, not five

`keyLength` and `saltLength` stay untagged in `password.go`. Neither is a cost knob, and `keyLength` has a correctness floor: `Verify` refuses a key outside 16..64 bytes, so a "faster" shorter key would make every hash this package writes fail its own `Verify` — surfacing as a correct password being rejected inside a fixture, nowhere near the constant that caused it. Keeping them out makes that mistake **unavailable**, not merely unlikely.

### The gate

`TestProductionParametersHoldTheOWASPBaseline`, untagged so it compiles only against the shipped values. Before the split the constants sat beside the code that used them and a reviewer reading `Hash` saw them; now they do not.

Mutation-verified against both ways weak parameters could reach production:

```
memoryKiB lowered to 64  -> FAIL "the shipped Argon2id memory in KiB is 64,
                                  below the OWASP 2024 baseline of 19456"
build tags flipped       -> FAIL on time cost AND memory
```

A third mistake needs no test: two files declaring these constants without disjoint tags is a duplicate-declaration error, so a dropped tag **cannot build**.

---

## 2 — `ac64ac00c` seed-demo's copy gets a gate, not a promise

Found while doing the above. `tools/seed-demo/users.go` restates these parameters — the package is behind a nested `internal` and in another module — and the comment guarding that restatement said:

> *"if they change there, they must change here, and a demo user failing to log in is the signal."*

**There is no such signal.** `Verify` parses the cost out of the PHC hash:

```go
fmt.Sscanf(parts[3], "m=%d,t=%d,p=%d", &m, &t, &p)
```

So a hash written at *any* parameters verifies successfully, and nothing re-hashes on login. Lower seed-demo's constants and every demo user signs in perfectly while their stored hash is weak — permanently, because the weakness lives in rows already written, and invisibly, because the one event that would have to break for anyone to notice is the event that keeps working.

A comment asserting an invariant the code does not hold is its own defect class. This one asserted the absence of a hazard it was the only thing standing between.

The gate reads both files' source and compares, which is all the module boundary permits. It evaluates `19 * 1024` rather than matching the spelling, and reports a **missing** constant as a failure rather than skipping it — a rename is how a parity gate usually stops comparing anything without going red.

```
argonMemory 19*1024 -> 1024   FAIL "seed-demo hashes at argonMemory=1024
                                    where the product ships memoryKiB=19456"
argonTime renamed             FAIL "argonTime is not declared ... any more,
                                    so nothing pins seed-demo's timeCost"
```

---

## Verification

```
make check-backend        CHECK=0
make test-integration     OK: integration passed with 0 skips (42 packages)
                          wall clock 170s  (254s before this change)
craft static --strict     PASS
```

`make check` caught this PR's own new code once and correctly: `argonParityPairs` is waiver-shaped, so the census refused it until it was declared `gatekit:fixture`, and the fixture count pin moved 20 -> 21 in the same commit as the declaration.

## On the four items that were already done

The design doc predates them. D2's egress gate shipped in #2066 and is **better than what I started building** — it clears `DialTLSContext`, not just `DialContext`, and for an https URL net/http calls that one *instead*, so a DialContext-only guard would have been bypassed on exactly the scheme the vendor is reached over. D7 memoizes via `headSchema`/`headFingerprint`; D8 counts `AcquireCount()` instead of waiting out a deadline, which the design had explicitly ruled impossible; D9's baseline is committed at `scripts/integration-lane-timing.txt`.

---
## Summary by cubic
Speeds up CI by moving Argon2id costs behind a build tag: production keeps the OWASP 2024 baseline, while the `integration` build uses cheaper values. Adds gates and tests that pin shipped costs, keep `tools/seed-demo` in parity, and harden `Verify` against malformed hashes.

- Build-tag split: production parameters in `backend/internal/modules/identity/internal/password/params.go`; `integration` parameters in `params_integration.go`. `go build ./...` (no tags) ships production values; cheap values are compiled out.
- Behavior change in tests only: t=2, m=19 MiB, p=1 → t=1, m=64 KiB, p=1. Measured: `internal/modules/identity` 17.830s → 11.676s; full lane 254s → 170s.
- Gates: `TestProductionParametersHoldTheOWASPBaseline` prevents weakening shipped costs; `TestSeedDemoHashesAtTheParametersTheProductShips` compares `params.go`/`password.go` with `tools/seed-demo/users.go` and fails on drift or missing/renamed constants.
- Auth tests: `verifyrejects_test.go` refuses malformed PHC inputs with targeted errors, pins the 16..64 key-length bound, and asserts `Hash` always salts; `hashroundtrip_test.go` pins valid key/salt ranges and that a hash written by this build verifies under it.
- `keyLength` and `saltLength` remain untagged in `password.go`; not cost knobs and have correctness floors.

**Rollout**
- No config or data migration. Shipped binaries unchanged; only the `integration` build used by CI is affected.

<sup>Written for commit 95906a9a81a4f5026e17bee4b96e94987c4fb6a5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/2207?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/gatecensus_test.go
+++ b/backend/gatecensus_test.go
@@ -50,7 +50,7 @@ const (
 	// Permitting the marker without counting it would let it spread quietly and
 	// reopen the class these rules close; pinned, every new one moves a number a
 	// reviewer sees.
-	wantFixtureAnnotations = 20
+	wantFixtureAnnotations = 21
 )
 
 // censusDecl is one package-level declaration this census governs — a map from
--- a/backend/internal/modules/identity/internal/password/hashroundtrip_test.go
+++ b/backend/internal/modules/identity/internal/password/hashroundtrip_test.go
@@ -1,0 +1,43 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+package password
+
+import "testing"
+
+// keyLength and saltLength are NOT cost parameters and are deliberately absent
+// from the tagged files. This pins the property that makes that separation
+// matter: Verify refuses a key outside 16..64 bytes, so a keyLength below that
+// floor would make every hash this package writes fail its own verification.
+//
+// The failure would not look like a bad constant. It would look like a correct
+// password being rejected, somewhere in a fixture, with nothing pointing here.
+func TestTheKeyAndSaltLengthsStayInsideWhatVerifyAccepts(t *testing.T) {
+	if keyLength < 16 || keyLength > 64 {
+		t.Errorf("keyLength is %d, outside the 16..64 Verify accepts — every hash written "+
+			"with it would fail its own Verify", keyLength)
+	}
+	if saltLength < 16 {
+		t.Errorf("saltLength is %d; RFC 9106 wants 16 bytes for password hashing", saltLength)
+	}
+}
+
+// A hash written by THIS build must verify under it. Trivially true today and
+// the point is that it stays true under the integration tag as well: the tagged
+// build compiles this same test with the cheap parameters, so if a future edit
+// puts an invalid combination there — a memory below the 8*threads floor argon2
+// requires, say — the integration lane fails here rather than somewhere a
+// fixture logs in.
+func TestAHashThisBuildWritesVerifiesUnderIt(t *testing.T) {
+	const plaintext = "correct horse battery staple"
+	phc, err := Hash(plaintext)
+	if err != nil {
+		t.Fatalf("hashing with this build's parameters: %v", err)
+	}
+	if err := Verify(plaintext, phc); err != nil {
+		t.Fatalf("this build wrote a hash it cannot verify: %v", err)
+	}
+	if err := Verify("wrong", phc); err == nil {
+		t.Fatal("Verify accepted the wrong password, so the comparison proves nothing")
+	}
+}
--- a/backend/internal/modules/identity/internal/password/params.go
+++ b/backend/internal/modules/identity/internal/password/params.go
@@ -1,0 +1,22 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+//go:build !integration
+
+package password
+
+// The Argon2id cost parameters every build EXCEPT the integration test build
+// uses: the OWASP 2024 interactive-login baseline.
+//
+// These are the values the product ships. `go build ./...` — which is what
+// produces api, worker, migrate and the desktop bundle — has no build tags, so
+// this is the file that compiles into every binary a user ever runs.
+//
+// TestProductionParametersHoldTheOWASPBaseline pins them. That gate is not
+// ceremony: once the values live behind a build tag, nothing else in an
+// ordinary `make check` would notice them being lowered.
+const (
+	timeCost  = 2
+	memoryKiB = 19 * 1024
+	threads   = 1
+)
--- a/backend/internal/modules/identity/internal/password/params_integration.go
+++ b/backend/internal/modules/identity/internal/password/params_integration.go
@@ -1,0 +1,39 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+//go:build integration
+
+package password
+
+// The Argon2id cost parameters the INTEGRATION TEST BUILD uses, and only it.
+//
+// The harness runs the real first-boot ceremony on every test — EnsureInstallation,
+// login, change-password, login again — which is five derivations before a test
+// asserts anything. At the production parameters that is ~94 ms per test, and
+// argon2.processBlocks measures 28.5% of all CPU samples in the overlay package.
+// The fixture needs the FLOW to be real, not the work factor: no test in this
+// tree asserts the cost, and none can, because Verify reads a hash's parameters
+// out of the hash itself.
+//
+// Measured on this tree, per derivation:
+//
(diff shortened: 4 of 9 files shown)


tinnhlt-gra:
## Coverage: the requirement is structurally vacuous here, so I closed the real gap instead

**This PR's production diff has zero coverable lines.** It is two `package` clauses and two `const` blocks; constants are not executable statements, so Go emits no coverage blocks for them — verified against the profile rather than assumed:

```
$ grep -cE "params\.go|params_integration\.go" pw.cov
0
```

Everything else in the diff is `*_test.go`, which `sonar.test.inclusions` classifies as tests rather than as code to be covered. So a "new-code coverage" figure here would be 100% or N/A whatever the tests did, and would mean nothing either way.

Reporting that number and moving on seemed worse than looking at what it was standing in for. In the package this PR touches, `Verify` sat at **77.3%**, and every uncovered branch was a malformed-hash rejection.

### `95906a9a8` — the rejection branches get tests

These decide what happens when `app_user.password_hash` holds something `Hash` never wrote: a truncated column, a hand-edited row, a value from a tool that formats PHC differently. The failure they prevent is not a crash — it is `Verify` reaching `argon2.IDKey` with garbage it half-parsed, comparing against a key length it inferred from that garbage.

| | before | after |
|---|---|---|
| package | 77.8% | **96.3%** |
| `Verify` | 77.3% | **100%** |

**The key-length bound is the one with history.** An oversized length used to be masked with `&0xFFFF` before the `int -> uint32` conversion, which *silently truncated* it — so a hash claiming a 65536+ byte key was compared at a length nobody chose. The explicit 16..64 bound replaced that mask, and nothing held it. A bound with no test is a comment.

Mutation-verified — delete the bound and the refusal disappears rather than the test merely going red:

```
--- FAIL: .../a_key_too_short_to_be_an_Argon2id_key
    the refusal does not say what is wrong: got "password: mismatch",
    want it to mention "implausible key length"
```

It also pins that `Hash` salts: two hashes of one password must differ, or a stolen table reveals which accounts share a password. Every other assertion in the file passes with a constant salt.

`make check-backend` CHECK=0 · `craft static --strict` PASS 0/0/0 on the pushed head.

tinnhlt-gra:
@coderabbitai review

The earlier attempt hit the OSS review limit, so this PR has **no review coverage at all** right now — the CodeRabbit check reads `pass` because the review could not start, not because it found nothing. All 21 CI checks are green and Sonar's gate passed, which is exactly why the missing review is easy to miss here.

Three things I'd most like challenged:

1. **`params_integration.go` puts a security parameter behind a build tag.** The safety argument is that `go build ./...` carries no tags, so the weak Argon2 values are compiled out of every shipped binary rather than refused at runtime. If there is any path by which the `integration` tag reaches a production artifact, that argument fails and this should not land.

2. **Three constants are split; `keyLength` and `saltLength` deliberately are not.** `Verify` refuses a key outside 16..64 bytes, so a "faster" shorter key would make every hash fail its own verification — surfacing as a correct password being rejected inside a fixture, far from the constant that caused it. Push back if that separation reads as arbitrary rather than load-bearing.

3. **The seed-demo parity gate reads source with `go/parser`** rather than importing (the module boundary forbids it), and evaluates `19 * 1024`. If there is a spelling of those constants it would fail to read, it goes **quiet rather than red** — that is the failure mode I want a second pair of eyes on, more than any style question.

Issue #2205, opened 2026-08-21
The retention sweep's activity erase leaves activity.raw behind


dungbanguyen:
Three statements destroy activity content, and they do not agree about what "erased" means:

| Path | Nulls |
|---|---|
| Art. 17, `privacy/erasuretimeline.go:53` | `body`, `raw`, `counterparty_email`, channel `source_id`/`thread_key`, tombstones `subject` |
| Restriction lift, `privacy/retentionrestricted.go:128` | `subject`, `body`, `raw`, `counterparty_email` |
| Retention sweep, `privacy/retentionactions.go:177` | `body` only; tombstones `subject` |

The retention arm leaves `activity.raw` — the column the schema calls the ["re-parseable original"](https://github.com/gradionhq/margince-poc-v1/blob/main/backend/migrations/core/0008_activity.up.sql) . It is thorough about six *derived* stores (embeddings, graph, attachments, approval quotations, transcript readings, delivery log) and misses the column next to the one it clears. `anonymizePersonRecord`, 15 lines below in the same file, does clear `raw`.

### No live effect today — this is a trap, not a leak

Nothing in the tree ever populates `activity.raw`: neither `activities/activity.go:160` nor `capture/sink.go:346` lists it, and every other reference sets it to NULL. The first connector to store the original — and `mailmap.Parse` already has it in hand — makes the retention sweep silently keep full message bodies past their window.

Filed publicly rather than as an advisory for exactly that reason: no privilege boundary is crossed and there is no reproduction that leaks anything.

### Why nothing would catch it

- Migrations `0290`/`0291` install a data-layer guard that refuses an incomplete erasure — but only on the **restriction-lift** path.
- `piicoverage_test.go` has a `retentionErasures` field that pins which assignments constitute a sweep erasure, but `activity` is declared `{erasureWrite: true, sarRead: true}` with no `retentionErasures`, so its sweep arm is unchecked. `voice_learning_signal` is the only table using the field.

`0291` exists because a guard was "written from the smaller of two content lists" — its own comment says two copies is how the two paths come to destroy different things. This is the third copy.

### Scope

1. The retention sweep clears `raw`.
2. Declare `retentionErasures` for `activity` so the assignments are pinned and a future divergence fails the gate.

[...]

Issue #2145, opened 2026-08-21
13 different pieces of code can change a company, and they no longer agree


LarsGradion:
## In one sentence

There are 13 different pieces of code that can change a company, they have quietly drifted apart, and the differences between them have already caused real bugs.

## What is going on

When something in margince changes a company — a rep edits the form, the website read-back fills in a blank, an import runs, two companies get merged — a piece of code writes that change to the database.

There should be roughly one such piece of code. There are **13**, written in **10 different styles**, and several of them can change the *same field*:

| Field on a company | How many different pieces of code can change it |
|---|---|
| Legal name | 7 |
| Industry | 6 |
| Company name | 5 |
| Description | 5 |
| Street address | 4 |

Creating a company is fine — there is exactly one place that does it, and it says so. Only *changing* a company grew this way.

## Why this is a problem and not just untidy

The 13 have stopped agreeing with each other. Three examples, all live today:

**1. Some of them can edit an archived company.** Four of the writers check "is this company archived?" before writing. Four do not. So depending on which path a change comes through, a company someone deliberately archived can still be modified.

**2. They disagree about protecting against two people editing at once.** There are four different approaches in use across the 13 — one proper mechanism, one manual version bump, one relying on a database trigger, and one doing nothing at all. Which protection you get depends on which screen you happened to use.

**3. It has already caused a deadlock.** A note in the code (`renamerecheck.go`) records that three of these paths took a database lock in the wrong order and could freeze against each other. It says they "were only found by review" — meaning nothing automatic caught it, and nobody knows if there is a fourth.

There is a fourth example from this week. While adding geocoding, I needed every address change to mark the company's coordinates as out of date. I wired it into the two writers I could find. A code review found **four more I had missed**. Without the fix, a company that moved would keep showing up in "customers near Stuttgart" at its old address — answering confidently, and wrongly. I had to push that rule down into the database itself, because there is no single place in the code to put it.

That is the real cost: **any new rule about companies has to be added 13 times, and the one that gets missed fails silently.**

## What the code already knows

This is not a surprise to the codebase. It is full of comments defending each writer against the one next to it, and one file carries a hand-written checklist of four obligations a writer owes — a reminder list where a shared code path should be.

One of those comments makes a claim that is no longer true: `companyform.go` says it is "the only writer of that column a human drives." Two other writers also change that field on a human's action.

## Suggested approach

Please do **not** simply add a 14th, better writer on top. The other 13 keep working and nothing forces anyone to use the new one.

[...]

PR #2220, opened 2026-08-21
fix(records): a research claim lands in the columns it names, and reuse becomes a rule

tinnhlt-gra:
## What this changes

Two commits, and the second exists because of the first.

1. **`people/researchclaim.go` bound every column of its insert to the wrong placeholder.** `SaveResearchClaims` errored on every call, so the person-research accept path has been dead on `main`.
2. **The rule against building a capability twice is now in both rulebooks** instead of only in ten source comments.

Found while auditing #2145's "same duplication with Meeting Brief in MCP and Website" comment.

## 1. The broken writer

```
INSERT INTO person_profile_field (person_id, field, value, evidence_snippet, source_ref, source, captured_by)
VALUES ($1, $1, $2, $3, $4, $6, $5)
```

Arguments passed: `MustWorkspace(ctx)`, `personID`, `claim.Field`, `claim.Value`, `claim.Quote`, `claim.SourceURL`, `by`, `researchSource`.

| Column | Was bound to | Should be |
|---|---|---|
| `person_id` | `$1` = workspace id | `personID` |
| `field` | `$1` = workspace id | `claim.Field` |
| `value` | `$2` = personID | `claim.Value` |
| `evidence_snippet` | `$3` = claim.Field | `claim.Quote` |
| `source_ref` | `$4` = claim.Value | `claim.SourceURL` |
| `source` | `$6` = claim.SourceURL | `researchSource` |
| `captured_by` | `$5` = claim.Quote | `by` |

The statement also declared `$1..$6` while eight arguments were passed — `by` and `researchSource` were never referenced.

Postgres refused it before any of that reached a row, because `$1` served a uuid and a text column in one statement:

```
ERROR: inconsistent types deduced for parameter $1 (SQLSTATE 42P08)
```

**Introduced** in #1764 ("close the gaps the phase-D column removals left in SQL literals"), 2026-08-19. The sweep removed the workspace column from this statement and renumbered wrongly.

`MustWorkspace` goes with the fix: `person_profile_field` has no workspace column — phase D removed it — and `uq_person_profile_field` is `UNIQUE (person_id, field)`. The comment claiming the branch had to name the workspace described a column that no longer exists.

## Why no gate caught it

Nothing in the tree executed the statement:

- `compose/personresearch/service_test.go` has two tests; both filter input and never reach the database.
- `rbacgate_test.go` is a static census.
- Only one test file in the whole tree mentions `person_profile_field` at all.

That is CLAUDE.md rule #6 — *"a test that supplies its own version of production proves nothing about production."*

## The tests

Seven integration tests, asserting the stored row **column by column** rather than that a row exists — the defect wrote a row every time, and each value was its neighbour's.

Verified they fail without the fix:

```
$ git stash push -- backend/internal/modules/people/researchclaim.go
$ make test-it DIR=backend/internal/modules/people RUN=TestSavingAResearchClaimWritesEachColumn
    researchclaim_integration_test.go:67: SaveResearchClaims: save the research claim "title":
      ERROR: inconsistent types deduced for parameter $1 (SQLSTATE 42P08)
--- FAIL: TestSavingAResearchClaimWritesEachColumnTheValueItWasGiven (0.30s)
```

Coverage of the changed code:

| Function | Coverage |
|---|---|
| `SaveResearchClaims` | **84.6%** |
| `webSourceURL` | **83.3%** |

Uncovered remainder is the unreachable error returns (`CapturedBy`, `Audit`, `EmitEvent` failures).

## 2. The rulebook change

The rule was already written ten times, in `compose/*seam*.go` comments — `briefseam.go`: *"one queue rather than two readings of it"*; `importseam.go`: *"it delegates rather than reimplementing, and that is the whole design"*. A comment inside the right file only helps someone who already found the right file.

No gate could catch it either:

- `gate.Inputs` (`cli/craft/gate/prompt.go:14-22`) carries the diff, touched files and their **siblings**. Reviewing a new assembler in `modules/agents/`, craft never sees `compose/meetingbrief/`.
- Reuse appears once in the rubric, as `P5 restraint`, in the section `prompt.go:88` labels *"never blocks"*.

So `prep_for_meeting` was written beside a working meeting brief, and the two now answer one question with different grounding: the web surface drops a sentence whose citations do not resolve, the tool keeps everything.

The new section states four rules — search the whole tree; the tool and web surfaces share one engine and you write the seam if none exists; never hand-type a SQL placeholder; a comment may not claim to be the only implementation unless a test holds it. It goes in **both** files and into `sharedSections`, so the parity test holds the copies together, and because `AGENTS.md` is fed whole into the gate prompt the rule reaches the reviewer as well as the author.

## Verification

```
make check                 exit 0  (backend + frontend)
craft static --strict      PASS — 0 blocker, 0 major, 0 minor
make test-it (people)      7/7 pass
```

Rebased on `origin/main` at `ed6e76a1a`.

## Deliberately not in this PR

Found in the same audit, each wants its own change:

- `compose/sitepeople.go:80` uses `strings.ToLower`, which does not fold ß→ss, while `namesim.go:46` says full casefold exists because *"the DACH market meets that pair daily"*. A German site rendering a name two ways mints two leads.
- `privacy/retentionactions.go:211` claims *"the same in-place anonymization the eraser performs"*; it never nulls custom columns and never deletes `field_provenance` or `ai_feedback` for person or lead. Extends #2205, which covers the `activity` arm only.
- `compose/report.go:51` calls itself *"the one spelling of weighted value"* that *"cannot drift apart"*; `orgrollup.go:107` computes it again in Go with no test comparing them.
- Unlike its three siblings, `SaveResearchClaims` never calls `storekit.StampFields`, so a research-accepted field gets no `field_provenance` row.
- `relationship` has four independent INSERT writers bypassing `CreateRelationship`.

Refs #2145.

---
## Summary by cubic
Fixes research claim saves and makes “reuse before you build” an enforced rule. Old: `SaveResearchClaims` reused `$1`, misnumbered placeholders, and failed with 42P08. New: the insert binds `$1..$7` in order, removes workspace, and upserts by `(person_id, field)`, preventing duplicate implementations noted in Linear #2145.

- Backend
  - Old: reused `$1`, declared `$1..$6` but passed 8 args; wrote wrong columns or failed. New: `INSERT ... VALUES ($1..$7)` with `ON CONFLICT (person_id, field)` updating value, evidence, source, and captured_by; trigger sets timestamps/version; drops obsolete workspace reference.
  - Adds Postgres integration tests that assert each stored column and the write shape: 1 `audit_log`, 1 `event_outbox`, audited count matches; acceptance is atomic on unknown fields; denies row-scope-only writers (`ErrPermissionDenied`); rejects archived persons (`ErrNotFound`).
  - Updates schema fitness waiver for `person_profile_field.person_id` to document both writers (enrich vs `SaveResearchClaims`) and gating.

- Rulebook and parity
  - Adds “Reuse before you build” to `AGENTS.md` and `CLAUDE.md`; `backend/agentsdocparity_test.go` now enforces four shared sections and byte-identical copies.
  - Tightens SQL placeholder guidance: derive `$N`; use only column literals or sanitized identifiers via `pgx.Identifier.Sanitize`; never hand-type placeholders; removes brittle counts.

<sup>Written for commit 10ada25f31c1630f562f95b9be394f44b1d77c0a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/gradionhq/margince-poc-v1/pull/2220?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -430,6 +430,12 @@ binding every column to the wrong value. That is not hypothetical: it shipped in
 `people/researchclaim.go`, and the accept path was dead for two days because no
 test executed the statement.
 
+The `%s` in that pattern is the COLUMN, and it carries its own rule: a compile-
+time literal, or a catalog name quoted with `pgx.Identifier.Sanitize` — the one
+spelling this repo uses (`storekit/customcolumns.go`). Never a string off a
+request body. Values are always `$N`; only identifiers are ever formatted, and
+an identifier a caller chose is an injection with a placeholder's manners.
+
 **4. A comment may not claim to be the only implementation unless a test holds
 it.** "the one spelling of X", "the only writer of Y", "the same anonymization
 the eraser performs" — if no test fails when a second one appears, delete the
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -412,7 +412,7 @@ answer one question with different grounding rules.
 
 **2. The tool surface and the web surface share ONE engine.** An MCP tool never
 re-derives what an HTTP handler already computes. The binding is a
-`compose/*seam*.go` file, and the ten that exist each state the rule in their
+`compose/*seam*.go` file, and the seams that exist each state the rule in their
 own words — `briefseam.go`: "one queue rather than two readings of it";
 `importseam.go`: "it delegates rather than reimplementing, and that is the whole
 design". **If no seam exists for the capability you need, write the seam** — do
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -397,6 +397,55 @@ scope clauses in `platform/auth`): object denial →
 `apperrors.ErrPermissionDenied` (403), row-scope miss →
 `apperrors.ErrNotFound` (404, existence-hiding).
 
+## Reuse before you build (non-negotiable)
+
+A second implementation of one capability is not untidy — it is two answers to
+one question, and the two drift until they disagree in front of a user. Four
+rules, each of them here because this tree has already paid for it.
+
+**1. Search the whole tree, not your directory.** Before adding a capability,
+grep its nouns across `backend/`, `frontend/src/` and `extensions/`. The
+duplicate is almost never in the package you are editing — that is precisely why
+it gets missed. The agent tool `prep_for_meeting` was written beside a working
+`compose/meetingbrief/` that a one-word grep would have found, and the two now
+answer one question with different grounding rules.
+
+**2. The tool surface and the web surface share ONE engine.** An MCP tool never
+re-derives what an HTTP handler already computes. The binding is a
+`compose/*seam*.go` file, and the ten that exist each state the rule in their
+own words — `briefseam.go`: "one queue rather than two readings of it";
+`importseam.go`: "it delegates rather than reimplementing, and that is the whole
+design". **If no seam exists for the capability you need, write the seam** — do
+not write a second assembler. A module may not import a sibling or `compose`
+(ADR-0054 §3), so rolling your own will always look like the cheaper path; it is
+the wrong one, and this is the case where saying so out loud is the only thing
+that helps.
+
+**3. Never hand-type a SQL placeholder.** Derive `$N` from the argument slice —
+`args = append(args, v)` then `fmt.Sprintf("%s = $%d", col, len(args))`, as
+`deals/offer_lines.go` does — or use `storekit.InsertFragments`. Nothing in this
+repo checks that a statement's column count, placeholder count and argument
+count agree, so a hand-numbered statement is one careless sweep away from
+binding every column to the wrong value. That is not hypothetical: it shipped in
+`people/researchclaim.go`, and the accept path was dead for two days because no
+test executed the statement.
+
+**4. A comment may not claim to be the only implementation unless a test holds
+it.** "the one spelling of X", "the only writer of Y", "the same anonymization
+the eraser performs" — if no test fails when a second one appears, delete the
+claim or write the test. Every such claim audited in this tree was false. A
+false uniqueness claim is worse than silence: the next author greps, finds it,
+and stops looking.
+
+**Two writers of one invariant either share a helper or say why they do not.**
+If you are adding the second, put the reason in the code beside it, not in the
+pull request where the next reader will not see it.
+
+Catalogs to read before building anything they might already list:
+[docs/reference/modules.md](docs/reference/modules.md) for backend capabilities,
+[frontend/src/design-system/README.md](frontend/src/design-system/README.md) for
+every control that already exists.
+
 ## Craftsmanship
 
 The anti-tell catalog T1–T11, in full below — this list is the rule, not a summary
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -444,7 +444,7 @@ answer one question with different grounding rules.
 
 **2. The tool surface and the web surface share ONE engine.** An MCP tool never
 re-derives what an HTTP handler already computes. The binding is a
-`compose/*seam*.go` file, and the ten that exist each state the rule in their
+`compose/*seam*.go` file, and the seams that exist each state the rule in their
 own words — `briefseam.go`: "one queue rather than two readings of it";
 `importseam.go`: "it delegates rather than reimplementing, and that is the whole
 design". **If no seam exists for the capability you need, write the seam** — do
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -462,6 +462,12 @@ binding every column to the wrong value. That is not hypothetical: it shipped in
 `people/researchclaim.go`, and the accept path was dead for two days because no
 test executed the statement.
 
+The `%s` in that pattern is the COLUMN, and it carries its own rule: a compile-
+time literal, or a catalog name quoted with `pgx.Identifier.Sanitize` — the one
+spelling this repo uses (`storekit/customcolumns.go`). Never a string off a
+request body. Values are always `$N`; only identifiers are ever formatted, and
+an identifier a caller chose is an injection with a placeholder's manners.
+
 **4. A comment may not claim to be the only implementation unless a test holds
 it.** "the one spelling of X", "the only writer of Y", "the same anonymization
 the eraser performs" — if no test fails when a second one appears, delete the
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -4,11 +4,11 @@ This file provides guidance to Claude Code (claude.ai/code) when working in this
 repository. It is the long form: the full operating detail lives here.
 [AGENTS.md](AGENTS.md) is a deliberately shorter digest for other agent
 harnesses that links back here. Most sections differ on purpose, so do not sync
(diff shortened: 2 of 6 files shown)


tinnhlt-gra:
## UAT — accepting a research claim, on a real stack

`make dev-fresh` + `make seed-dev`, api `:18080`, SPA `:8080`, Chrome. Recorded end to end.

uat-2220-research-claim-lands-in-the-right-columns

### Before

Alice Müller's **Research** tab:

- *"No data provider is connected, so nothing has been bought."* — the run half needs a licensed provider, which dev has none of. That is the named `not_connected` state, not a failure.
- *"No enriched field carries evidence yet."*

Overview → Details → Title: empty (`Add title`), and the relationship brief reads *"Alice Müller is recorded here with no title or employer."*

### The call

The accept half takes its claims from the request body, so it is exercisable without a provider — through the app's own session, which is what the drawer's Accept button does:

```
POST /v1/people/01a024af-.../research/save
{"claims":[{"field":"title","value":"Head of Procurement",
            "source_quote":"Alice Müller leads procurement across the Demo GmbH group.",
            "source_url":"https://demo.test/team/alice-mueller"}]}

→ HTTP 200  {"saved":1}
```

On `main` this call cannot succeed at all — the statement fails type deduction on `$1` before it writes anything.

### After

The **Enrichment evidence** panel now renders:

| | |
|---|---|
| Title | **Head of Procurement** |
| Captured by | `typed by a person` |

Those are precisely the two columns the bug transposed: `value` used to receive the person id, and `captured_by` used to receive the evidence quote. The panel showing the value under `Title` and the actor as a person is the visible form of the fix.

### The stored row

```
person_id        | 01a024af-4d11-70c1-80c1-371c13c50020   ← Alice (was: the workspace id)
field            | title                                  ← (was: the workspace id)
value            | Head of Procurement                    ← (was: the person id)
evidence_snippet | Alice Müller leads procurement across the Demo GmbH group.
                                                          ← (was: claim.Field)
source_ref       | https://demo.test/team/alice-mueller    ← (was: claim.Value)
source           | person_research                        ← (was: claim.SourceURL)
captured_by      | human:01a024ae-bce1-79e9-9e51-7781d48d22d3
                                                          ← (was: the evidence quote)
version          | 1
```

Every column holds what the caller supplied for it.

### Two things this UAT surfaced, neither in scope here

1. **Details → Title stays empty.** `SaveResearchClaims` writes only the `person_profile_field` evidence row; it never fills `person.title`. Its sibling `enrichsignature.go` *does* fill the column (fill-if-empty). So an accepted title is visible on the Research tab but not in Details — worth a product ruling rather than a silent difference.
2. **`make seed-dev` cannot seed a fresh install.** `ADMIN_PASSWORD` defaults to `demo-password-123` and `config/margince-admin-password` holds the same value, so `sign_in_as_admin`'s first login returns 200, the function early-returns, and the bootstrap rotation never runs — then every write 403s `password_change_required`. Worked around with `ADMIN_PASSWORD='…' make seed-dev`. Filing separately.

Both are pre-existing and unrelated to the binding fix.

tinnhlt-gra:
## Review round — what was applied, and what was not

Four reviewers on this diff: Codex, Fable, a craftsmanship pass and a security redteam. All four cleared the fix itself; between them they raised nine findings. Seven are applied in `65bc6e1d8`/`747871e66`. The two that are not are recorded below rather than papered over.

### Applied

| Finding | Change |
|---|---|
| Rulebook rule 3 showed `fmt.Sprintf("%s = $%d", col, …)` without saying what may fill `%s` — and that text goes verbatim into the craft gate prompt, so it was one careless reading from teaching identifier injection | rule 3 now says the column is a compile-time literal or a catalog name through `pgx.Identifier.Sanitize`, never request text |
| The schema census waiver for `person_profile_field.person_id` said the column is "never from a request body". `SaveResearchClaims` takes it from the request path — and this branch is what makes that writer live, so the waiver was accidentally true only while the statement was broken | waiver now names both writers and what holds each. The census is keyed by column, so it cannot notice a second writer of one already waived — the text is the only control |
| The suite asserted stored columns but not the write shape | asserts one `audit_log` row, one `event_outbox` row, and the audited count read off the row rather than trusted from the return value |
| Nothing proved the acceptance is all-or-nothing — every refusal case returns before the transaction opens | new case: a field outside the closed set is refused mid-transaction and the reader's first claim does not survive |
| Every case ran unbounded, where the write-authority arm short-circuits — so the gate was untested on a path that was dead until this branch | new case: a rep who may only *read* a person is refused |
| A comment claimed a failing second claim would leave the first saved — the opposite of what the transaction does | restated as the invariant the test holds |
| The source-scheme case duplicated `researchclaim_test.go`'s unit-lane coverage and asserted nothing this layer decides | deleted |

One correction worth stating: my new gate test first asserted `ErrNotFound`, expecting existence-hiding. It answers `ErrPermissionDenied`, and that is **correct** — `auth/writescope.go` runs the visibility probe first and keeps its 404, and this caller passes it, so there is nothing left for a 404 to hide. I verified the read genuinely succeeds for that seat before pinning the sentinel.

### Not applied — needs a decision, not a comment

**1. `SaveResearchClaims` never calls `storekit.StampFields`.** Three of the four reviewers raised this independently. All three sibling writers of `person_profile_field` stamp `field_provenance`; this one does not. The consequence the redteam found is concrete: a site or search pass fills `role` and stamps provenance, a human then accepts a research claim for `role`, `DO UPDATE` replaces the evidence row — and `FieldOrigins` still reports the **bot** as that field's origin. Two provenance layers naming different authors for one value.

I did not fix it and I did not add a "why not" comment, because I could not write one that is true. The new rulebook section says two writers of one invariant either share a helper or say why they differ; this is the case where the honest answer is that nobody decided. Wants a ruling: does a human acceptance become the field's origin (almost certainly yes), and does it clear a prior verdict?

**2. A stale `corrected` verdict masks an accepted value.** `compose/person360/sectionstimeline.go:327` overlays a verdict keyed on `profile_field:<field>` with no recency check against `person_profile_field.updated_at`, and the accept path clears no verdict. So: reader corrects the enrichment's `title`, later accepts a research claim for `title`, and both the 360 and the sidecar keep serving the old corrected value while badging the new one as `corrected`. Harmless while this write was dead; reachable now. The `DO NOTHING` siblings never hit it because they cannot change a field a verdict exists for.

### Also found, unrelated to this diff

- **Accepting a `title` claim does not fill `person.title`** — only the evidence row. So it shows on the Research tab and not in Details, while `enrichsignature.go` fills the column. Visible in the UAT above.
- **`make seed-dev` cannot seed a fresh install.** `ADMIN_PASSWORD` defaults to `demo-password-123` and `config/margince-admin-password` holds the same value, so `sign_in_as_admin`'s first login returns 200, the function early-returns, and the bootstrap rotation never runs — then every write 403s `password_change_required`.

PR #2260, opened 2026-08-22
feat(privacy): correspondence filed under a project is business correspondence

LarsGradion:
An email filed under a project and no deal failed the Handelsbrief shield, so the
erasure path scrubbed it — subject replaced, body and `raw` nulled — exactly like
an internal note. The word "project" did not appear in the privacy module at all.
The attribution ladder has been writing project links since it shipped, so the
volume of correspondence in that state was growing.

## The rule

A project qualifies its correspondence by existing. It is a commercial engagement,
so mail filed under it documents an actual transaction whether or not a deal on it
has closed — which is also what reaches mail from a negotiation that was **lost**,
and from delivery work years after the deal that started it. The deal rule alone
misses both.

**The link is the event.** A deal qualifies at an instant, so its stamp sweeps
every activity filed against it when it concludes. A project qualifies
continuously, so each activity is stamped as it is filed, inside the transaction
that writes the link. There is no window in which an erasure sees the record
unclassified.

## ⚠️ The backfill is not reversible

It stamps every project link the ladder already wrote. Of the two ways to be wrong
only one destroys data — the principle `deals/retentionstamp.go` already settled —
and an unstamped Handelsbrief sitting in the backlog is the destructive error.
Every link in that backlog came from a matched thread, an inherited deal link, or
a typed `[KEY]`: statements, not guesses.

Once merged, six-year marks exist that need a named person and a written reason to
remove. The statement is idempotent (write-once class plus `ON CONFLICT DO
NOTHING`), so a partial failure can be re-run —
`TestTheBackfillStampsPreExistingLinksAndIsIdempotent` proves it by executing the
statement twice.

## Four writers, all stamping

`insertActivityLinks` (create/send), `RelinkActivity` (human relink), the capture
ladder, and `copyEchoLinks` (echo absorb). Capture reaches `activities` through a
seam, and the matcher and the stamp arrive as one `ProjectAttribution` value rather
than two parameters — the wiring that could attribute without classifying does not
exist rather than being checked for.

## An agent could have written six-year floors unattended

`relink_activity` was `auto_execute` and its enum admits `project`. After this
change, filing under a project writes a mark that is write-once in the database and
is **not** lifted by relinking away — so an agent with an ordinary write passport
could have marked a whole mailbox with nothing in the product to undo it. That is a
denial of the subject's Art. 17 right the controller cannot reverse.

The operation is now `dynamic`: confirm-first for a **project** destination, still
auto-executing for every other. The resolver only ever raises, and an unrecognised
destination fails toward the gate. `relink_activity` also gains the decision-grant
mapping it never had, so a staged card can actually be released — without it the
raise would have been a dead end, not a gate.

## Schema

`are_derived_names_its_deal` demanded a `deal_name` on every non-pin row, which a
project row satisfies in neither branch. Restructured to what it was actually
asserting, and the name is now bound to the **basis** rather than merely required
to be present — an `OR` would admit a `deal_won` row naming only a project, which
is evidence about the wrong record.

The freeze trigger names every column it protects, so the new pair was unguarded by
default; it now covers `project_name`/`project_id` on the same terms as the deal
pair. The uniqueness index widens so two projects over time do not collide.

The controller's restricted list gains a `projects` array beside `deals`
(additive). A held record whose stated reason resolves to nothing is evidence of
nothing.

## Verification

Every guard is mutation-checked — reverting the shield arm, the relink stamp, the
capture stamp, the note exclusion, or the `ON CONFLICT` each fail their own test.

Two cases worth naming because they are counterintuitive:

- **A note filed under a project is stamped and still erased.** The stamp is
  unfiltered by kind, matching the deal writer; `handelsbriefShielded` excludes
  notes, which is what keeps that class inert. If those two rules drift apart,
  every internal jotting on a two-year engagement gets a six-year floor.
- **Deleting a project takes its links but leaves the evidence.** The shield tests
  the link; the erasure's stamp arm joins the project behind it. They agree only
  because the link FK is `ON DELETE CASCADE` — a test asserts that rather than a
  comment claiming it.

`make check-backend`, `make craft-static`, `make check-fe`, and the compose,
activities, capture and migrations integration lanes are all green.

## One commit is not this branch's work

`chore(migrations): record the deal-room audit verbs in the head catalog` — the
migration from #2247 shipped without regenerating the catalog, so
`TestMigrationsBuildTheCommittedSchema` has been red on main since. Regenerated
with the flag, not hand-edited. Carried here only because CI runs that test and the
catalog is a derived artifact with one right answer.

It slipped because `make check-backend` does not run it: the check is
integration-tagged.

---
## Summary by cubic
Classifies correspondence filed under a project as business correspondence and stamps it when the link is written; previously, project‑only mail failed the Handelsbrief shield and was erased like a note. The stamp is write‑once, survives relinks, and shield/erasure now hold these records; restricted lists now show qualifying projects.

- Writers and stamping: create/send, `RelinkActivity`, capture via a single `ProjectAttribution`, and echo absorb all stamp; capture now stamps even if the project link already exists, and absorb stamps the copied project link.
- Auth/governance: filing under a project requires activity.UPDATE on both create and capture paths; `relink_activity` is `dynamic` with a destination‑type resolver and asks confirmation when the destination is a project (auto‑exec for others). The MCP door now stages correctly and mints an approval card.
- Shield/erasure: adds a project arm and a legacy backfill so pre‑stamp rows are held; the floor’s fallback includes project links.
- Restricted list/API: adds `projects` alongside `deals`; frontend types and docs updated.
- Schema: evidence adds `project_id`/`project_name`; constraints/freeze cover them; uniqueness widens to avoid cross‑project collisions.

Rollout
- Run DB migration 1787370000. It backfills stamps for existing project links; the backfill is idempotent and not reversible.
- Update any automation using `relink_activity` to handle dynamic confirmation for project destinations; ensure actors linking under a project have activity.UPDATE.

<sup>Written for commit e2328ef7cb57b22a802af25f9ab1daabb3232a2e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/2260?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added deal-room participant roster management, including invitations, updates, resends, revocation, capability controls, one-time credentials, and delivery tracking.
  - Project-linked correspondence now receives retention classification and appears in restricted-record details.
  - Restricted records now distinguish qualifying deals from projects.

- **Bug Fixes**
  - Activity relinking requests confirmation for projects, unknown destinations, and invalid requests.
  - Project attribution and retention evidence remain consistent through relinking, duplicate capture, and record deletion.

- **Documentation**
  - Updated privacy guidance to explain project-based retention classification and release requirements.


LarsGradion:
Superseded by #2284 (same content, rebased onto main with the migration re-stamped above main's highest). Merged there.

PR #2262, opened 2026-08-22
fix(records): the committed schema records the deal-room audit verbs

tinnhlt-gra:
## What main was red on

`main` has been failing since the Deal Room migration landed. Three separate reds, and this PR fixes the one still open:

| Red | Where | Status |
|---|---|---|
| `arch-lint`: dealrooms → `gopkg.in/yaml.v3` | backend gate | already fixed on main by `038aef0b0` |
| `sonarcloud quality gate` | scheduled | #2256 repoints the scan |
| **`TestMigrationsBuildTheCommittedSchema`** | integration shard 3/6 → `./migrations` | **this PR** |

## The defect

`core/1787357528_audit_admits_deal_room_lifecycle_verbs` widened `audit_log`'s
action CHECK with `publish`, `pause`, `resume`, `close` — but
`migrations/testdata/head_catalog.txt` was not regenerated in the same commit.
The golden exists precisely so that "the migrations changed the schema" is a
decision somebody committed rather than something that happened, and it has been
out of date with head ever since.

The consequence is not only the one red test: `./migrations` fails early in
integration shard 3/6, so the shard reports failure and the whole
`integration (RLS + erasure + HTTP e2e)` rollup goes red — every later package in
that shard reports under a lane nobody can read a verdict from.

## The fix

Regenerated the golden the way the failure message instructs:

```
go test -tags integration ./migrations -run TestMigrationsBuildTheCommittedSchema -update-catalog
```

The diff is **one line**, and it is exactly the four verbs the migration added —
nothing else in head moved. That is the assurance the golden is there to give,
and it is worth reading the diff for: had the migration changed something it did
not mean to, this is where it would show.

## Verification

- Reproduced red on `origin/main` first, then green after — `make test-it DIR=backend/migrations RUN=TestMigrationsBuildTheCommittedSchema`
- `make check-backend` — exit 0, 0 failures, all legs ran (build, vet, lint, arch-lint, tests, gates, drift, composition)

---
## Summary by cubic
Records the deal-room lifecycle verbs in the committed schema so migrations and the golden catalog match. Previously, the migration admitted publish/pause/resume/close but the committed `head_catalog` did not, causing TestMigrationsBuildTheCommittedSchema to fail and the integration lane to stop early.

- Updates `backend/migrations/testdata/head_catalog.txt` to add publish, pause, resume, and close to the `audit_log.action` CHECK.
- No new migration or runtime change; this only aligns the committed catalog with the already-applied migration and restores green CI.

<sup>Written for commit b7976fe3399b15b45a7e0775a97f15a9902c845a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/2262?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Audit logs now support recording additional actions, including publishing, pausing, resuming, and closing.


tinnhlt-gra:
Closing as superseded — **main is green, and #2258 got there first.**

#2258 (`feat(deals): a named buyer gets let into a Deal Room`) landed at `6f858fb4f` while this was in CI, and its last commit — `fix(migrations): the head catalog admits every Deal Room audit verb` — regenerates the same golden line. Its version is a strict superset: it carries `invite` and `revoke` as well as the four lifecycle verbs this PR covered, because that branch added two more audit verbs of its own. Keeping this open would only re-conflict on one line.

#2258 also removed the `dealrooms` → `yaml` arch-lint grant rather than keeping it — `lifecycle_test.go` now derives its state vocabulary from the generated `DealRoomState` constants instead of parsing `crm.yaml`, so the vendor dependency is gone rather than permitted. That is the better resolution of the arch-lint red than `038aef0b0`'s grant, and it settles the inconsistency with the `root` component's "must not gain a YAML dependency" rule that the grant had opened up.

Verified against `origin/main` @ `6f858fb4f` before closing:
- `make test-it DIR=backend/migrations` — whole package PASS, including `TestMigrationsBuildTheCommittedSchema`
- `make arch-lint` — `OK - No warnings found`; `dealrooms` no longer imports `gopkg.in/yaml.v3`

The remaining known red on main is the SonarCloud quality gate, which #2256 owns.

**One thing from this worth keeping**, and #2258's own commit message says it too: `make check-backend` does *not* catch head-catalog drift. The catalog test is integration-tagged, so it only runs under `make test-integration` or a directory-scoped `./migrations` run — which is exactly how two migrations in a row shipped without regenerating the golden. A migration that changes the schema needs that lane before it is called done.

Issue #2040, opened 2026-08-20
The agent tool listing is near its token ceiling: 633 bytes of headroom


LarsGradion:
## What is wrong

`TestTheToolListingLeavesTheRunRoomInTheWindow` caps the Surface-B tool listing at 15000 tokens of a 24000-token window. The listing currently renders **59982 bytes — 18 bytes under the 60000-byte limit**.

That is not headroom. It means the next person who adds a filter, a report dimension, a record type, or a sentence to any tool description will fail a gate that has nothing to do with their change, and will have to go shave someone else's wording to get their own feature in.

## How we got here

Three separate times in one session, adding a legitimate field pushed the listing over and had to be paid for by trimming unrelated text:

- `partner_attribution` on the deal filter vocabulary (40 bytes) — paid for by abbreviating the non-string type annotations (`(boolean)` → `(b)`) in `agents/tools_list.go`.
- `partner_org_id` as a report dimension — paid for by shortening the report catalog's repeated `aggregate fields:` label to `aggregates:` in `agents/tools_report.go`.
- A `partner-performance` prebuilt report (~101 tokens) — **abandoned entirely**; `deals-by-stage` can group by partner instead, so the convenience report was dropped rather than fight the budget.

Each of those trims was defensible on its own. Collectively they are a feature budget being paid out of prose quality, and the next one has nowhere left to take it from.

## What this blocks right now

Making `partner` a real `record_type` for the generic tools (`search_records`, `read_record`, `update_record`, `list_records`) needs roughly 45 bytes across the enums, before any provider work. It does not fit. That work is planned and currently cannot land.

## Options

1. **Raise `listingBudgetNumerator`/`Denominator`** in `backend/internal/compose/agenttooldescriptions_test.go` (currently 5/8 of a 24000-token window). The number was chosen to leave room for the run's own observations; whether 5/8 is still right is a judgement about how much context a run needs, and somebody should make it deliberately rather than by attrition.
2. **Elide the listing** — the test's own message says "the listing is never elided", which is why it must fit. If tools could be offered in a narrowed set per run, the ceiling stops being global.
3. **Compress the rendering** — the filter vocabulary repeats each record type's full name and the catalog repeats per-report labels. There is real redundancy left, but it is one-off savings, not a structural fix.

Recommend (1) or (2) with a deliberate decision, not another round of (3).

## Where

- The gate: `backend/internal/compose/agenttooldescriptions_test.go`
[...]

PR #2354, opened 2026-08-23
feat(agents): an agent can read a partner by name, not only by inference

### LINKED ISSUE #2041: Make partner a real record_type for the generic agent tools
## What is missing

`listPartners` and `upsertPartner` carry `x-mcp-tool` annotations mapping them to `search_records` and `update_record` with `record_type: partner`. That is **admission metadata only**. The generic tools' `record_type` enums in `backend/internal/modules/agents/tools.go` (and `tools_update.go`) do not include `partner`, so `search_records(record_type="partner")` and `update_record(record_type="partner")` are not callable. `getPartner` carries no annotation at all.

The practical effect: an agent can find a partner company and see that it carries the `partner` relationship type, but cannot read or write the partner extension's own fields — tier, certification status, relationship stage.

`docs/how-to/set-up-a-partner-program.md` claimed otherwise; that has been corrected to describe what actually works.

## What it takes

Real data routing, not an enum edit. An enum value the provider refuses is worse than a missing one, which is why the annotation on `getPartner` was deliberately NOT added:

- `partner` as an entity in `backend/internal/shared/ports/datasource/datasource.go`
- the search/read/update switches in `backend/internal/compose/provider.go`
- `backend/internal/modules/people/provider.go` (people owns the partner table)
- `backend/internal/modules/agents/recordfields.go`
- the enums in `tools.go` (search, read) and `tools_update.go` (update)
- overlay-mode behaviour

Note `create_record` should NOT gain `partner`: there is no create path — a partner row is made by upserting onto an existing organization.

## Blocked on

**#2040** — the agent tool listing has 18 bytes of headroom against its token ceiling. The enum additions need roughly 45 bytes before any provider work. This cannot land until that ceiling is resolved.

## Context

Filed while shipping the partner program (#2019 attribution + commission ledger, #2039 partner reporting). The MCP half was in the plan and is the one piece that could not be built.

LarsGradion:
## What this does

A company could be marked a partner, and an agent could see the company but never its terms. `partner` was not a record type the generic tools accepted, so tier, certification and relationship stage were unreachable. `docs/how-to/set-up-a-partner-program.md` told operators as much.

`partner` is now a record type for the two READ verbs:

- `read_record(record_type="partner", id=<organization id>)` returns one partner's terms.
- `search_records(record_type="partner")` lists them, narrowed by `partner_role` or `cert_status`.

This is Phase 4 of the partner-program plan, the one phase left unbuilt when the rest shipped on 2026-08-20.

## Three shape decisions, each ruling something out

**A partner is addressed by its ORGANIZATION's id.** The row is that company's 1:1 extension and has no id of its own. `GetPartner` gates on both the `partner` and `organization` objects and checks the organization is visible, so a caller who cannot open the company cannot read its commercial terms through the other name.

**Partner stays out of `RecordType` and out of the untyped sweep.** Nothing points AT a partner — you tag, list, link and grant against the organization — so a `RecordPartner` would widen five polymorphic columns to reach a target they already reach. A sweep matches on text, and every word worth searching lives on the organization, so including it would return the same company twice under two names.

**Read-only.** No create or update verb, and deliberately no partner in the `create_record` enum: a partner row is made by upserting on an existing organization, and an enum value the provider refuses is worse than a missing one.

## The refused text query

`ListPartners` has no text index. A text query is REFUSED rather than dropped, because a dropped term answers an unfiltered page — which reads as "these all matched".

## Why there is a migration

Declaring `EntityPartner` obliges the four `EntityType`-bound schema CHECKs (attachment, embedding, field_provenance, custom_field) to carry it. `TestEveryDomainEnumMatchesItsSchemaCheck` derives the Go set from the package's constants, so a half-widening cannot get past the gate. Migration `1787444866` is that reconciliation.

`custom_field` widens with the other three but stays closed to authoring: `customfields.FieldObjects` is the acceptance set and partner is absent from it. The catalog CHECK staying wider than that list is the posture the other three already have.

## Tests

Six integration tests through the real writer, in `partnerseam_integration_test.go`:

1. The admit case first — a refusal test against an authority that refuses everyone proves nothing.
2. A seat with no `partner` grant is refused (the harness's `AdminPerms` deliberately holds none).
3. Not-found for a company that never joined the programme.
4. The text query is refused.
5. Role narrowing works, including the empty page for a role nobody holds.
6. An unknown filter is refused.

## On the token ceiling

The tool listing grew 200 bytes and now sits **750 bytes under** its budget. #2040 is untouched by this — it remains the standing decision about that ceiling, and its numbers had already moved before this change (PR #2300 raised the budget to 17/24; the issue still describes 18 bytes against 5/8).

## Gates

- `make check-backend` — exit 0, 134 packages ok
- `make test-integration` — all six new tests pass
- `make craft-static` — 0 blocker, 0 major, 0 minor across all four roots

Closes #2041.

---
## Summary by cubic
Agents can read partner terms directly by name. Previously, generic tools refused `partner`, so an agent could see a company but not its tier, certification, or stage.

- read_record
  - `read_record(record_type="partner", id=<organization id>)` returns a partner’s terms. `GET /v1/organizations/{id}/partner` is mapped with tier `auto_execute` and scope `read`; access requires both `partner` and `organization` read.
  - A partner is addressed by its organization ID and remains read-only in generic tools. `PUT /v1/organizations/{id}/partner` still permits writes when the passport has write scope and the seat has `partner:update` and `organization:update`.
- search_records
  - `search_records(record_type="partner")` lists partners; it refuses `q` (no text index), and the tool exposes no filters. The tool schema now says `q` is not accepted with `record_type=partner`.
  - Partner stays out of untyped sweeps. Compose separates what is “searchable” (sweep) from what is “nameable” (allowed to request); `partner` is nameable but not searchable. Cursors are validated and resumed against the nameable set so mixed walks don’t restart.

- Migration
  - Apply migration `1787444866` to add `datasource.EntityPartner` and widen the EntityType CHECKs on `attachment`, `embedding`, `field_provenance`, and `custom_field` using ADD NOT VALID/VALIDATE/DROP/RENAME. `custom_field` remains closed to authoring.
  - Rollback may require removing partner-linked attachments referenced from deal rooms before narrowing, due to ON DELETE RESTRICT.

<sup>Written for commit fff079fd159a6b778914c9f095de15b05be3a587. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/2354?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Agents can read and search partner records using an organization ID.
  * Partner searches support filtering by role and certification status.
  * Partner records are available through generic record tools and MCP schemas.
  * Partner data is supported across attachments, embeddings, custom fields, and provenance.
  * Broad searches exclude partners unless explicitly requested.
  * Text queries are not supported for partner searches.

* **Documentation**
  * Updated partner program guidance and MCP reference documentation with lookup behavior, permissions, and search limitations.

### DIFF
--- a/backend/api/crm.yaml
+++ b/backend/api/crm.yaml
@@ -2492,6 +2492,7 @@ paths:
       tags: [Organizations, Partners]
       operationId: getPartner
       summary: Read the partner extension on an org (404 if the org is not a partner).
+      x-mcp-tool: { verb: read_record, record_type: partner, tier: auto_execute, scope: read }
       responses:
         '200':
           description: The partner record.
--- a/backend/internal/compose/agentpolicy_gen.go
+++ b/backend/internal/compose/agentpolicy_gen.go
@@ -233,6 +233,7 @@ var agentPolicies = map[string]agentPolicy{
 	"GET /v1/organizations/{id}/evidence/{entityType}/{entityId}":        {Op: "getClaimEvidence", Access: "human-only", Tool: "", RecordType: "", Tier: "", Scope: ""},
 	"GET /v1/organizations/{id}/growth-fit":                              {Op: "getOrganizationGrowthFit", Access: "human-only", Tool: "", RecordType: "", Tier: "", Scope: ""},
 	"GET /v1/organizations/{id}/logo":                                    {Op: "getOrganizationLogo", Access: "human-only", Tool: "", RecordType: "", Tier: "", Scope: ""},
+	"GET /v1/organizations/{id}/partner":                                 {Op: "getPartner", Access: "tool", Tool: "read_record", RecordType: "partner", Tier: "auto_execute", Scope: "read"},
 	"GET /v1/organizations/{id}/site-reads/latest":                       {Op: "getLatestSiteRead", Access: "tool", Tool: "read_record", RecordType: "organization", Tier: "auto_execute", Scope: "read"},
 	"GET /v1/organizations/{id}/site-reads/{readId}":                     {Op: "getSiteRead", Access: "tool", Tool: "read_record", RecordType: "organization", Tier: "auto_execute", Scope: "read"},
 	"GET /v1/overlay/export":                                             {Op: "downloadOverlayExport", Access: "human-only", Tool: "", RecordType: "", Tier: "", Scope: ""},
--- a/backend/internal/compose/integration/partnerseam_integration_test.go
+++ b/backend/internal/compose/integration/partnerseam_integration_test.go
@@ -1,0 +1,156 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+//go:build integration
+
+package integration
+
+// The partner extension through the datasource seam — the surface an MCP
+// agent reads partners by.
+//
+// A partner is the 1:1 extension of an organization, so the seam addresses it
+// by the ORGANIZATION's id. That is the detail worth holding: a reader who
+// cannot open the company must not learn its commercial terms through the
+// other name, and the seam must not offer a second way in that skips the gate
+// the HTTP handler applies.
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/gradionhq/margince/backend/internal/modules/people"
+	"github.com/gradionhq/margince/backend/internal/shared/apperrors"
+	"github.com/gradionhq/margince/backend/internal/shared/kernel/ids"
+	"github.com/gradionhq/margince/backend/internal/shared/kernel/principal"
+	"github.com/gradionhq/margince/backend/internal/shared/ports/datasource"
+)
+
+func partnerSeamProvider(e *Env) *people.Provider { return people.NewProvider(e.DB()) }
+
+// partnerReader is a seat holding the partner and organization grants the
+// read needs. AdminPerms carries no `partner` object on purpose (see
+// SeedPartnerOrg), so a suite that wants the ADMIT case must say so.
+func partnerReader() principal.Permissions {
+	return principal.Permissions{
+		RoleKeys: []string{"admin"},
+		Objects: map[string]principal.ObjectGrant{
+			"partner":      {Read: true},
+			"organization": {Read: true},
+		},
+		RowScope: principal.RowScopeAll,
+	}
+}
+
+// The admit case first: without it, a refusal test proves only that the
+// authority refuses everyone.
+func TestPartnerReadsThroughTheSeamByItsOrganizationID(t *testing.T) {
+	e := Setup(t)
+	p := partnerSeamProvider(e)
+	tier := "tier2_20"
+	org := e.SeedPartnerOrg(t, "Seam Partner GmbH", &tier, nil)
+	ctx := e.As(e.AdminUser, nil, partnerReader())
+
+	rec, err := p.Read(ctx, datasource.EntityRef{Type: datasource.EntityPartner, ID: org})
+	if err != nil {
+		t.Fatalf("reading a partner through the seam: %v", err)
+	}
+	if rec.Ref.Type != datasource.EntityPartner {
+		t.Fatalf("record type = %s, want partner", rec.Ref.Type)
+	}
+	if rec.Ref.ID != org {
+		t.Fatalf("record id = %s, want the organization id %s", rec.Ref.ID, org)
+	}
+}
+
+// The seam is not a weaker copy of the HTTP path: a seat with no partner
+// grant is refused here too.
+func TestPartnerSeamRefusesASeatWithoutThePartnerGrant(t *testing.T) {
+	e := Setup(t)
+	p := partnerSeamProvider(e)
+	tier := "tier1_15"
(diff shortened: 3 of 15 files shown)


LarsGradion:
Thanks — one of these was a genuine bug that would have shipped a broken feature. All seven findings addressed in e220c9e.

## Fixed

**`search_records(record_type="partner")` was refused** (cubic, `tools.go:125`). Correct and serious: `sweepOrder` used `searchable` for two different questions — where an unguided sweep looks, and what a caller may name. Keeping partner out of the sweep therefore made it unnameable, so the PR's headline feature did not work. The diagnosis was exactly right, including that my integration tests missed it by calling `people.Provider.SearchEntity` directly instead of crossing the compose path.

`nameable` is now a separate list, a superset of `searchable`. `resumeStream` judges cursors against it too — a partner page mints a cursor in its own stream, and the old check would have refused this seam's own token on page two. Six unit tests in `sweeporder_test.go`; three fail when the fix is reverted.

**CHECK swaps hold ACCESS EXCLUSIVE during validation** (CodeRabbit, both migration halves). Valid. This repo already established the staged pattern in `1787352862`, whose comment explains that `lock_timeout` bounds only the wait for the lock, never how long it is held. Both directions now do `ADD ... NOT VALID` / `VALIDATE` / `DROP` / `RENAME`. Verified by rolling the old version back and re-applying: all four constraints keep their original names, all four report `convalidated`, and `head_catalog.txt` is byte-identical.

**The down migration assumes partner attachments are unreferenced** (cubic, `down.sql:7`). Correct — deal-room document, thread and decision references are `ON DELETE RESTRICT`. I did not add a cascade: stopping the rollback is the right outcome, because the alternative is deleting negotiation evidence to make a downgrade tidy. The migration now says so, and says what to do if it fires.

**"read-only to an agent" was false** (cubic, `set-up-a-partner-program.md:127`). Right, and worth stating precisely: `partner` is absent from `update_record`'s enum, but `PUT /organizations/{id}/partner` carries a write annotation and a passport is a REST credential as well as an MCP one, so an agent with the `partner` object grant can reach it. The doc now says the object grant decides this, not the tool vocabulary. That annotation predates this PR.

**The role and certification filters are advertised but unreachable** (both bots, `mcp-info.json`). Correct. `list_records` carries filters and does not serve `partner`; `search_records` takes no filters at all. Rather than half-wire it, the doc now says an agent gets the whole list and narrows it itself. Filed as #2361 — the store half already exists, the contract half needs `listPartners` re-annotated, and it costs listing tokens against the budget tracked in #2355.

## Gates

`make check-backend` exit 0 · `make test-integration` — OK, 0 skips, 43 packages · `make craft-static` — 0 blocker, 0 major, 0 minor

LarsGradion:
Round 2 — one more real defect found, fixed in 2626588.

**`resumeIndex` restarted a mixed walk** (cubic, `provider.go:278`). Correct, and it is the third place the one-list conflation reached. `slices.Index` answers -1 for a stream missing from the list it ranks against, and -1 compares below every real position, so a walk of `[person, partner]` resumed at a partner cursor returned index 0 and re-served every person the caller already had. Now ranked against `nameable`. Two tests — the advance case and the ordinary first-stream re-entry — and the first fails when reverted.

**The REST write route needs three things, not one** (cubic, doc:139). Right, and my wording would have caused exactly the mistake described: `UpsertPartner` requires `partner:update` **and** `organization:update` (becoming a partner stamps the company's relationship types), plus write scope on the passport. The doc now lists all three as the places to withhold.

**Tool name inconsistency** (CodeRabbit, doc:126-129). Fair — the passage named `list_records` while the surrounding text used `search_records`, and the distinction was load-bearing rather than a typo. Reworded: `search_records` serves partner and takes no filters at all; `list_records` carries filters and does not serve partner. Both halves now stated.

## Not changed: the migration transaction finding

cubic is technically right that `ADD CONSTRAINT ... NOT VALID` holds ACCESS EXCLUSIVE until commit, so the `VALIDATE` scan runs under it rather than under the weaker lock it would take alone. `dbmigrate.inTx` wraps every migration file in one transaction, so that is unavoidable here.

I did not restructure it, for two reasons. The fix proposed — separately committed phases — is a change to how *every* migration in this tree runs, not something one migration may decide. And the staged shape is the pattern `1787352862` deliberately established here, whose own comment explains the `lock_timeout` trap.

What I did do is stop the migration overclaiming. It now states plainly what the staging does not buy, why getting it would require a migrator change, and that the exposure is bounded today because these are the small tables — embedding is the largest at ~1.5k rows. Better to record the limit than to imply a safety property that is not there.

## Gates

`make check-backend` exit 0 · `make test-integration` — OK, 0 skips, 43 packages · `make craft-static` — 0 blocker, 0 major, 0 minor · migration rolled back and re-applied clean, constraints keep their names and all report `convalidated`

PR #2427, opened 2026-08-23
fix(frontend): the native-control gate derives its pre-filter from its own list

### LINKED ISSUE #2425: The native-control gate's prefilter is a second copy of its own element list
`frontend/src/design-system/native-controls.test.ts` states the rule once:

```ts
84:  const nativeControls = new Set(["select", "option", "optgroup"]);
```

and then restates it, by hand, in the prefilter that decides which files are
parsed at all:

```ts
119:  if (!/select|option|optgroup|\\/.test(text)) return [];
```

**Add a fourth element to the rule and the gate does not detect it**, because
the prefilter skips every file that mentions only the new name. Reproduced on
`origin/main` by doing exactly what a future author would do — extend the set
and add a probe case:

```
const nativeControls = new Set(["select", "option", "optgroup", "datalist"]);
{ name: "a datalist", fires: true, src: "const a = <datalist />;" },

× reports a datalist
Tests  1 failed | 33 passed (34)
```

The probe suite catching it is the *lucky* half. The dangerous half is the
census over the real tree: a file mentioning only the new element is never
parsed, produces no finding and no error, and a skipped file is indistinguishable
from a clean one. An author who extends the set without adding a probe gets a
green census that scanned nothing for their element.

**Fix, verified** (34/34 pass, census still ~1.3s):

```ts
if (!new RegExp(`${[...nativeControls].join("|")}|\\\\`).test(text)) return [];
```

The `\\` arm stays and keeps its own reason: the string scan reads COOKED text,
so an escape can spell a name that is not in the raw source.

**Owed alongside the fix:** a case asserting the prefilter is DERIVED, not merely
correct today — the shape of the bug is that both lists are individually right
while nothing binds them. Something like: for every member of `nativeControls`,
`findNativeControls` reports a file containing only that element.

## Provenance

The prefilter is mine, from #2386. Found by generalizing a finding
`margince-next-9b` hit twice in #2414/#2419 and named for me:

> The prefilter is where an AST census goes blind, and it must be derived FROM
> the arms. A dropped file produces no finding and no error, which is
> indistinguishable from a clean file.

Filed rather than fixed because that session owns the frontend-gate backlog now.
`frontend/scripts/ext-imports.test.ts` was checked against the same class and is
clean — its allowed set is read from `frontend/package.json`'s `exports` map
rather than restated, and it carries no exemptions at all.

tinnhlt-gra:
Closes #2425.

## The defect

`native-controls.test.ts` states its rule once:

```ts
const nativeControls = new Set(["select", "option", "optgroup"]);
```

and then **restates those three names by hand** in the pre-filter that decides
which files get parsed at all:

```ts
if (!/select|option|optgroup|\\/.test(text)) return [];
```

Both lists are individually correct, so nothing failed. Add a fourth element to
the declared set and the gate goes **silently blind** to it — a file spelling
only `<datalist>` is skipped before it is ever parsed, and a skipped file
produces no finding and no error. That is indistinguishable from a clean file.

Measured both directions, with each mutation verified **by diff** rather than
by a string match that might not have applied:

| | `nativeControls` + `"datalist"`, plus a probe case for it |
|---|---|
| before | **1 failed \| 33 passed** — `× reports a datalist` |
| after | **35 passed** |

## Why it survived review

This is the gate's own subject wearing different clothes: the pre-filter is a
second copy of the element list, **inside the census whose job is finding second
copies**. It sat on the axis nobody was watching — that file has no exemptions,
and exemptions were where we had been looking after #2419.

## The fix, and the test that is actually owed

The alternation is derived: `` new RegExp(`${[...nativeControls].join("|")}|\\\\`) ``.

The **backslash arm stays hand-written**, with the reason beside it: it is a
property of ESCAPES, not of any element name. `'<select>'` holds no
verbatim `select` in its raw text but cooks to one, and every escape begins with
a backslash. Deriving that from the set would be wrong.

The assertion is **not** "the pre-filter is right today" — two lists that agree
while nothing binds them pass every test anybody would think to write. It
asserts the **derivation**:

- every name in the declared set is admitted by the pre-filter;
- a name outside it is *not* (otherwise "derived" would be satisfied by a regex
  matching everything);
- the escape arm still holds.

Written against the set rather than against a literal, so it cannot itself
become a third copy.

**Honest about its reach**, since that is this file's own standard: with the
pre-filter hand-restated but the set unchanged, the new test does **not** fire —
the two lists agree and there is nothing yet to catch. It fires the moment the
set grows, naming the element:

```
AssertionError: the pre-filter does not admit a file spelling <datalist>
```

## Provenance and scope

Found by a sibling session from a note about the same defect class in
`consumermailonelist_test.go` (#2419), where a gate carried a hand-written
*sample* of the set it judged. `ext-imports.test.ts` was checked against this
class and is **clean** — its allowed set is read from `package.json`'s `exports`
rather than restated.

Test-file only; no product code changes. `make check-fe` green, **4396 tests**.

---
## Summary by cubic
Derives the native-control pre-filter from the declared set and verifies reachability through the real detector. Previously the pre-filter hardcoded tag names, so expanding the set could skip files silently; now the alternation tracks the set while preserving the explicit escape arm.

- Builds the pre-filter regex from `nativeControls`; keeps the backslash escape as a hand-written arm with rationale.
- Asserts through `findNativeControls` that every declared tag is reachable, an unrelated tag is not, and escaped spellings still match.
- Test-only change; no product code impacted.

<sup>Written for commit 8dcd1df67b4771afd2bbb7009532ccbc57ae3f44. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/2427?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Tests**
  * Expanded regression coverage for native control detection across all declared controls.
  * Added validation that escaped control names are recognized correctly.
  * Added checks to ensure unrelated elements are not classified as native controls.

### DIFF
--- a/frontend/src/design-system/native-controls.test.ts
+++ b/frontend/src/design-system/native-controls.test.ts
@@ -83,6 +83,24 @@ const SUBST = "$" + "{id}";
 // The elements a native dropdown is made of.
 const nativeControls = new Set(["select", "option", "optgroup"]);
 
+// The pre-filter's alternation, DERIVED from the set above rather than restated
+// beside it.
+//
+// It was restated, and that is the defect this file's own subject describes
+// wearing different clothes: the prefilter was a second copy of the element
+// list, inside the gate whose job is finding second copies. Both lists were
+// individually correct, so nothing failed — until a fourth element was added to
+// the set and the gate went silently blind to it, because a file spelling only
+// `<datalist>` was skipped before it was ever parsed.
+//
+// The backslash arm is NOT part of the derivation and stays hand-written: it is
+// a property of ESCAPES, not of any element name. `'<sel\u0065ct>'` holds no
+// verbatim `select` in its raw text but cooks to one, and every escape begins
+// with a backslash — so a file with neither a name nor a backslash cannot spell
+// one. Deriving it from the set would be wrong, which is why it is stated once
+// with its reason instead.
+const prefilter = new RegExp(`${[...nativeControls].join("|")}|\\\\`);
+
 // Tests and stories are scanned too: a test that drives a native control is a
 // test of the wrong control, and a story catalogues what we ship. Which files
 // those are is ../../scripts/lib/source-tree.ts's answer, shared with the
@@ -116,7 +134,7 @@ function findNativeControls(path: string, text: string): string[] {
   // character reaches cooked text only by appearing raw or by coming from an
   // escape, and every escape begins with a backslash — so a file with neither a
   // name nor a backslash cannot spell one, and skipping it costs no coverage.
-  if (!/select|option|optgroup|\\/.test(text)) return [];
+  if (!prefilter.test(text)) return [];
   const source = ts.createSourceFile(
     path,
     text,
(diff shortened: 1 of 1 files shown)


tinnhlt-gra:
@coderabbitai review

The head moved to `8dcd1df67` after your last pass — the derivation test was reshaped to assert through `findNativeControls` (the census's real entry point) rather than against `prefilter.test`, so it now fires on a newly added element with nothing else changed. Test-file only; no product code. Please review the current head.

PR #2451, opened 2026-08-23
docs: a gate that reads a smaller tree still says PASS

tinnhlt-gra:
## Why

The duplication sweep closed slice A — fourteen merged PRs, each a second implementation collapsed into one owner with a gate behind it. What the sweep learned while shipping is recorded nowhere in the tree, and two of the lessons are about the gates themselves:

- **Every gate the sweep produced was found, after shipping and by a reviewer, to have hard-coded part of its own subject** — a sampled domain list, a restated element list, a prefilter narrower than the detector behind it. A consumer-mail census shipped with its own 25-domain sample *inside the test that forbids second consumer-mail lists*; two reviewers named it independently.
- **Every hand-written census was short**, and never by a little (3 → 8 disclosure sites; "8 sites" → 25 literals). Each miss had the same shape: the thing that hid a copy from a reader also hid it from the detector written to find copies.

Neither failure has an assertion. A gate that reads a smaller tree reports the same word for it — PASS. The rulebooks carried no rule against either, so the next sweep repeats them.

## What changes

Docs only. No code, no gate behaviour.

**Two binding rules** added to *Rules learned from the review loop*, byte-identical in `CLAUDE.md` and `AGENTS.md` (`TestSharedRulebookSectionsAreIdenticalInBothDocs` holds that), with the full rationale in `README.md` where rules 1–6 already have theirs:

7. **One invariant broken in two languages is ONE item, not two.** Two errors on opposite sides of a wire can cancel, so fixing one half is a regression rather than half a fix — the money scale case, where the screen agreed with itself and only the record was wrong, and making the server currency-aware alone would have printed a hundred times the price on an outbound offer.
8. **A census that can fail short has already failed.** No prefilter or skip-list without measuring what it buys; statements, not lines, with the join bounded; and once the gate is green, plant the shape it cannot see.

**A fifth reuse rule** in *Reuse before you build*: a gate that hard-codes any slice of what it guards has become a second copy of it, and a copy inside a test is invisible to the author extending the owner.

**Method, with the failure that bought each clause**, added to the two principle pages that already hold these methods rather than to the rulebooks, which stay short:

- `docs/principles/derive-the-obligation.md` — the gate is part of its own subject; plant the shape the detector cannot see; measure a shortcut before defending it (one survived six rounds of being found narrower, and deleting it was *faster* than keeping it); statements not lines, and bound the join; a self-test that reads only an exit status proves nothing about a waiver; let a parser own the grammar (six defect classes in one PR, every one of them in the matcher rather than the rule). Plus a new section on finding the other side of a wire before fixing this one.
- `docs/principles/one-source-of-truth.md` — grep the other language, because a topic that reaches a screen is spelled twice; porting half a rule puts one rule in two homes (the TS/CSS gate split, still an open decision); and the second gate of a kind is when to extract the shared traversal — two gates here each owned a copy of the extension walk and disagreed four extensions against eight, so a unit was gated by one and invisible to the other.

## Verification

- `make check-backend` — green, exit 0 (includes the rulebook parity gate, `TestPublicTreeCitesNothingPrivate`, `check-craft-doc`, and the doc-parity gates).
- Stale counts fixed: the principles index said "the six rules"; `derive-the-obligation.md` said the same.
- No private path, repository or decision-record citation added; every artifact named (`freemail.Domains()`, `frontend/src/format/minorunits.ts`, `backend/frontendminorunits_test.go`, `frontend/scripts/lib/source-tree.ts`) exists at `origin/main` and was checked there rather than in this checkout.

---
## Summary by cubic
Documents two new binding rules and a reuse rule to stop gates that silently pass while reading a smaller tree or that duplicate their subjects; clarifies cross-language invariant guidance so reviewers do not overclaim protections.

- Rulebooks (`AGENTS.md`, `CLAUDE.md`): add rule 7 (one invariant broken on both sides of a wire is one item) and rule 8 (a census that can fail short has already failed), byte‑identical; add reuse rule 5 (a gate that hard‑codes any part of its subject is a second copy).
- Principles: `docs/principles/derive-the-obligation.md` expands the method for gates that hold (derive corpus from the owner, plant blind‑spot shapes, measure shortcuts, match statements and bound joins, assert on gate output, prefer parsers). Adds “find the other side before you fix this one” with clarified scope: only when both sides actually decide the rule; the minor‑unit example gates a shared TABLE (`values.MinorUnitExceptions()` vs `frontend/src/format/minorunits.ts`) while suites keep their own cases, and it reads TS with regexes (called out as a caveat, not a parser to copy).
- Principles: `docs/principles/one-source-of-truth.md` reinforces deriving the gate’s corpus, planting unseen shapes, and when to extract a shared traversal once a second gate appears.
- `README.md` adds rationale for rules 7–8; `docs/principles/README.md` updates “six rules” to “eight”. No code or gate behavior changes.

<sup>Written for commit 62ba19554f1ac6df3441695418fdfb08ebae2424. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/2451?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Documentation**
  * Expanded engineering guidance from six to eight review-loop rules.
  * Added standards for coordinating cross-language invariants, shared tables, mirror checks, and coverage.
  * Documented robust census practices, including complete scans, statement-aware matching, parser-aware validation, blind-spot tests, and waiver self-tests.
  * Clarified source-of-truth expectations and deriving guarded content from its owner.
  * Added guidance to prevent duplicated gate logic, traversal rules, and unnoticed defects.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -573,16 +573,19 @@ the short form:
    (integration tests live directly in `package compose` so unexported
    adapters are in scope). An unexpectedly uncovered new file usually
    means a test double stands where the real thing should.
-7. **One invariant broken in two languages is ONE item, not two.** When
-   the same rule is wrong in Go and in TypeScript, fixing one side alone
-   can be a REGRESSION rather than half a fix — a money scale wrong on
-   both sides of the wire cancels, the screen agrees with itself, and
-   making the server correct by itself prints a hundred times the price
-   on the offer a buyer signs. Find the other side before you fix this
-   one, land both in one change, and give the two suites one corpus to
-   read (`frontend/src/format/minorunits.ts` +
-   `backend/frontendminorunits_test.go`) rather than two tables that
-   agree today.
+7. **One invariant spelled on both sides of a wire is ONE item, not
+   two.** Most topics are implemented once — but where Go and TypeScript
+   each carry a spelling of the same rule, fixing one side alone can be
+   a REGRESSION rather than half a fix: a money scale wrong in both
+   directions cancels, the screen agrees with itself, and making the
+   server correct by itself prints a hundred times the price on the
+   offer a buyer signs. So check for the other side before you fix this
+   one, and land both in one change. Then make one side a DECLARED
+   mirror of the other, held by a gate that fails in BOTH directions —
+   `values.MinorUnitExceptions()` against
+   `frontend/src/format/minorunits.ts`, in
+   `backend/frontendminorunits_test.go` — rather than two tables that
+   only happen to agree today.
 8. **A census that can fail short has already failed.**
    Under-recognition is the one way a gate must not break: it reads a
    smaller tree and reports the same word for it, PASS, and there is no
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -416,7 +416,7 @@ scope clauses in `platform/auth`): object denial →
 ## Reuse before you build (non-negotiable)
 
 A second implementation of one capability is not untidy — it is two answers to
-one question, and the two drift until they disagree in front of a user. Four
+one question, and the two drift until they disagree in front of a user. Five
 rules, each of them here because this tree has already paid for it.
 
 **1. Search the whole tree, not your directory.** Before adding a capability,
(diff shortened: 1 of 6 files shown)


tinnhlt-gra:
All four threads accepted and fixed in `62ba19554` — every one of them was the text claiming a protection that is not there.

**Cross-language rule overclaimed its scope** (CodeRabbit, cubic P3). "A topic that reaches a screen is spelled twice" is false: most topics are decided once in Go and merely rendered by TypeScript, so as written it sent a reader looking for a counterpart that does not exist. Now conditional on both sides actually *deciding* the rule — which is the only case the cancellation argument covers.

**The cited gate is not "one corpus both suites read"** (cubic P2). Verified against `backend/frontendminorunits_test.go`: it holds one TABLE the two sides exchange, bidirectionally — a code present on one side only fails, and so does a differing digit count — while the two suites keep their own cases. A shared case corpus is a further step this tree does not take. The text now says what the gate does and explicitly warns against reading more protection into it.

**"Neither parser is hand-written" was false about the example under my own rule** (cubic P2, confidence 10 — and correct). That gate reads the TypeScript table with hand-written regexes (`tsMinorUnitEntry`, `tsComment`, marker slicing), cited two paragraphs below *let a parser own the grammar*. The gate's own comments already enumerate the holes each regex was added to close — a quoted `"MGA": 0` parsing as nothing, a comment mentioning a code keeping it green after the real entry was deleted — and every one of those is a hole `ts.createSourceFile` does not have. It is now cited for the direction it gates and named as the caveat rather than the parser to copy.

That last one is the finding I am most glad to have taken: a page teaching "a gate must not overclaim" shipped with an overclaim about its own worked example.

Re-ran green: `TestSharedRulebookSectionsAreIdenticalInBothDocs`, `TestPublicTreeCitesNothingPrivate`, `TestTheFrontendMinorUnitTableMatchesTheGoOne`.

PR #2454, opened 2026-08-23
fix(frontend): one locale for every rendered value, and one reading of a zone

tinnhlt-gra:
## What this is

A rendered number or date is written in SOME locale, and the only question is
whose. `format/format.ts` holds this product's answer — `INTL_LOCALE` maps the
three codes the app knows onto BCP-47 tags, with A100 pinning unconfigured
English to **en-GB**. Thirty-four call sites were answering it themselves.

Duplication audit item **B15**.

## The census, and why the audit's own was half of it

| class | count | what it renders in |
|---|---|---|
| 1. `toLocale*` with no locale | 23 sites, 11 screens | the **browser's guess** — the reader's OS, not the reader's choice |
| 2. `new Intl.*` handed the app's own **code** | 11 sites, 4 files | `"en"` resolves to **en-US** inside a product pinned to en-GB |
| 3. `new Intl.NumberFormat(undefined)` | 1 site | deliberate; the reason was not written at the subject |
| 4. a second reading of "a zone this renders" | 1 site | **a live bug** — below |

PLAN-v2 counted 13, a later hand recount got 25. Both missed class 2 **entirely**,
because a hand census greps for `toLocale` and class 2 contains none.

Neither class shows up on the machine that writes it: an English developer in
en-GB sees the right rendering from all three spellings.

## The live bug class 4 turned up

`scheduledsends.tsx` carried a second reading of `format.ts`'s `assertIanaZone`.
It asked Intl whether the zone name **resolved**; `format.ts` also refuses fixed
offsets, because an offset freezes the DST rules of the day it was picked. Every
fixed offset resolves. Measured:

| wire `scheduled_tz` | old probe accepts | `formatDateTime` then throws |
|---|---|---|
| `Etc/GMT-1` | yes | **yes** |
| `Etc/GMT+5` | yes | **yes** |
| `GMT` | yes | **yes** |
| `+01:00` | yes | **yes** |
| `Europe/Berlin` | yes | no |
| `Not/AZone` | no (falls back) | — |

So four of six take the **whole scheduled-sends list** down over one row —
through the exact hole the fallback's own docblock says it exists to close, and
reachable from wire data. `isRenderableZone` is now the formatter's own
predicate, derived from the assertion rather than restated beside it. Four
parameterized cases; **all four fail against the old spelling** (verified by
reverting the fix, below).

## Four copies collapsed, and two surfaces that disagreed

- Four byte-identical private `shortDate`/`dayMonth` functions → `formatDayMonth`.
  `personrail.tsx` **claimed in a comment** that its copy could not disagree with
  `personmemory.tsx`'s. Nothing held that, and both read the browser's locale.
- `personstrip.tsx` rendered `next_meeting.starts_at` in the browser's zone while
  `persontabs.tsx` renders **the same field** in `RECORD_ZONE` — so the strip and
  the tab beside it could name different days for one meeting.

Zone rulings follow the house convention rather than inventing one
(`company360.tsx:1460` states it in prose): a record's own occurrence and a
date-only wire value take `RECORD_ZONE`, a deadline the reader acts on takes
`viewerZone()`. Every site here is the first kind.

## The gate: `format/one-locale.test.ts`

Sibling of `format/zone-by-purpose.test.ts`, same shape — AST comment-stripping,
waivers that must carry a reason, and an arm that fails on a waiver which has
stopped covering a real site.

**Its subject is DERIVED, and that is the load-bearing part.** Every gate this
audit has produced was later found — by a reviewer, after shipping — to have
hard-coded some fragment of its own subject, so the thing it was written to find
kept getting through in the shape it could not see. Here:

- `intlFormatters()` asks the **runtime** which members of `Intl` carry
  `supportedLocalesOf` — that property *is* "takes a locales argument". Nine,
  including `DurationFormat`, which this TypeScript version does not declare.
- `localeMethods()` walks **TypeScript's own lib declarations** for `toLocale*`
  and `localeCompare`. No method list is maintained here.

**The derivation was written wrong first, in this audit's signature shape.** An
earlier `intlFormatters()` read the lib declarations and understood only the
inline `var X: { new (...) }` form, so it returned **five of nine** and missed
every named-constructor formatter — including `NumberFormat` and `DateTimeFormat`,
the two the whole gate is about. It would have passed a tree full of the defect.
That is why arms assert the derived sets reached their own subject: a derived
subject still has to be shown to have arrived.

### Mutation-tested, each mutant verified applied by diff before its run

| mutant | gate went red on |
|---|---|
| M1 restore a bare `toLocaleString` in a screen | leaves no surface rendering in a locale of its own |
| M2 restore `new Intl.NumberFormat(appLocale)` | leaves no surface rendering in a locale of its own |
| M3 blind the derivation to the named-constructor shape | derives the Intl formatters it gates + planted-defect arm |
| M4 make `namesTheMapping` wave everything through | sees each shape of the defect |
| M5 widen `isZoneLookup` to swallow every construction | sees each shape of the defect |
| M6 add an exemption covering nothing | keeps no exemption that has stopped covering a site |
| M7 drop the rendering-method filter | derives the locale-sensitive methods + two more |

A harness that edits and does not change the file reports `MUTANT DID NOT APPLY`
rather than a green run, because a `str.replace` whose pattern misses applies
nothing and the suite passes.

## Deliberately out of scope, named rather than dropped

Collation (`localeCompare`, 21 sites) and locale-sensitive casing
(`toLocaleLowerCase`, 2). They are **derived into the same set and then
excluded**, so the exclusion is a decision about members that were found rather
than a shorter search. The right answer is per site and opposite in each
direction — a machine key must sort identically for every reader, so a pinned
`"en"` is *correct* there, while a human-readable label must not be. That is a
ruling per call site, not a sweep. Filed separately.

## One deliberate exemption

`mcp-apps/bridge.ts` keeps `undefined`. An MCP app document is inlined whole into
a page this product does not own: no `LocaleProvider` above it, no i18n bundle
inside it, and importing one to learn a single tag would ship the whole catalog
to a third-party host. The reason is now written **at the subject** as well as in
the waiver.

## One visible change beyond the locale

An en reader's cost disclosure now reads **`US$0.0042`** rather than `$0.0042` —
what en-GB does with a dollar whose country it will not assume. The test that
asserted the old string is updated with that reason.

## Verification

- `make check-fe` — **green**, 4417 + 36 tests (re-run after rebasing onto `f155e0027`).
- `make check-backend` — **green** (diff is frontend-only).
- New-code coverage over the changed executable lines: **96.77% (30/31)**.
- The zone regression test reverted against the pre-fix `renderZone`: **4 failed | 11 passed**, all four the new cases.
- `format/zone-by-purpose.test.ts` gains one pinned entry, because this gate must spell the browser-zone lookup as a fixture to prove it does *not* report it — the carve-out that keeps the two gates from each claiming one rule.

---
## Summary by cubic
Unifies all number/date rendering under `INTL_LOCALE` and standardizes zone use. Old code mixed the browser’s locale and app codes like "en"; new code always renders in the reader’s locale, rejects fixed-offset zones with `isRenderableZone`, and en‑GB readers now see “US$0.0042”.

- Changes
  - Adds `formatNumber`, `formatDate`, `formatDateAbbrev`, `formatDayMonth`, `formatTimeOfDay`, `formatDateTime`; props move from a BCP‑47 string to `Locale`.
  - Replaces private date renderers and aligns zone purpose: record facts and date‑only fields render in `RECORD_ZONE`; reader actions use `viewerZone()`. Person and voice screens updated.
  - Scheduled Sends uses `isRenderableZone` with a logged fallback to the reader zone; parameterized tests cover fixed offsets with literal-text assertions.
  - Locale gate (`format/one-locale.test.ts`) derives Intl consumers from the runtime and locale‑taking methods (and argument positions) from TypeScript libs; recognizes element access and destructured/assigned aliases; sweeps extension frontends; excludes only `format.ts` and the gate itself, with named compute‑only exemptions; asserts `src/` coverage and the `INTL_LOCALE` mapping; independent of Node version; faster by scanning imports before the shadowing walk.
  - Numbers interpolated via `t()` now go through `formatNumber` where they represent quantities (voice insights, onboarding, backfill, voice screens). Workbench spend/tokens and micro‑USD costs use the mapped locale.

- Migration
  - Do not call `toLocale*` or construct `Intl` formatters directly; use the helpers or pass `INTL_LOCALE[locale]`.
  - Change props from a BCP‑47 tag to `Locale`.
  - Use `isRenderableZone` before rendering lists; fall back to the reader zone on invalid/fixed‑offset names.
  - Apply zone purpose consistently: `RECORD_ZONE` for record timestamps and date‑only fields; `viewerZone()` for reader actions.
  - Localize displayed numbers via `formatNumber`; pin a fixed locale only with a named exemption in the gate.

<sup>Written for commit beaa00dad8a77ace8de701685e7658501e9f7769. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/2454?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added localized day/month, time-of-day, number, count, cost, and statistics formatting.
  - Displays record-related dates in the record timezone for consistent results.
  - Validates unsupported time zones and safely falls back to the reader’s local timezone.
  - Standardized locale handling across onboarding, workbench, voice, and person views.

- **Bug Fixes**
  - Prevented browser-local formatting from causing inconsistent dates and numbers.

- **Tests**
  - Expanded coverage for locale formatting, timezone behavior, invalid zones, and localized currency.


tinnhlt-gra:
## Review round 1 — five gate holes, two false comments, one live defect the gate could not see

Three reviewers (CodeRabbit, cubic, and two adversarial passes). Every finding was reproduced before it was fixed, and every fix carries the planted case that proves it. Head is `3bd21ceed`, `origin/main` merged.

### The gate was blind in five directions

| finding | reproduced | fix |
|---|---|---|
| The exception was the whole `format/` **directory**, so a new formatter dropped beside `format.ts` could render in the browser's locale and pass | narrowing it surfaced 5 real sites | narrowed to two named FILES; the four legitimate ones are named exemptions with reasons |
| `INTL_LOCALE["en"]` passed — the predicate read the identifier, never the key | planted, confirmed | a literal key is a finding |
| `Intl["NumberFormat"]`, `const { NumberFormat } = Intl`, `const NF = Intl.NumberFormat` were all invisible — and the clean fixture **asserted the destructured shape was fine**, so the blind spot was vouched for | planted, confirmed | all three judged; the fixture line corrected |
| `localeMethods()` selected by NAME (`/^toLocale/`) — the hand-maintained fragment of its own subject the file's own header warns about | — | reads the SIGNATURE: a parameter called `locales`, which also yields the POSITION, so nothing in the file knows `localeCompare` takes it second |
| extension frontends were outside the sweep, though a unit's screen ships in the same bundle | — | uses `scripts/lib/source-tree.ts`, the walk two other gates already share, rather than a third copy |

### The zone regression test asserted nothing for half its cases

`new RegExp("in +01:00")` reads `+` as a quantifier, so it could not match the string it was built from and `queryByText` returned null either way. Two of four parameters proved nothing. Literal leaf-text matching now — **all four fail against the pre-fix `renderZone`**, verified by reverting it.

### A planted fixture refused by an earlier guard

Caught by mutation-testing rather than by reading: the planted-defect fixture did not import `INTL_LOCALE`, so every line was refused by the import check *before the argument was looked at* — and a mutation deleting the argument check **passed the suite**. The harness reported it as a green run over a mutant, which is the failure mode a mutation sweep exists to make visible. Fixture imports the mapping now; that mutant fails.

Fourteen mutants, each verified applied by diff before its run. A no-op edit reports `MUTANT DID NOT APPLY` rather than a green.

### Two comments claiming things that are not true

- `personrail.tsx` called an employment boundary "a date-only wire value". `Person360Employment.started_at` is `format: date-time` (`crm.yaml:19247`). The ruling stands; the reason is restated truthfully — it is written from a date picker, so the stored midnight carries no information and a reader west of UTC would name the previous day. (`personcards.tsx`'s identical comment about `close_date` **is** correct: `format: date`.)
- `mcp-apps/bridge.ts` said importing the mapping would ship the i18n catalog. Importing `INTL_LOCALE` alone would cost almost nothing — what is missing is the locale to look it up **with**, which is the stronger reason.

`COLLATION_ISSUE` no longer carries a site count: nothing derived it and nothing checked it, which is the same stale claim this rule exists to stop.

### A third class of the defect, live, that no gate here can see

The second pass found it and it contains **no `Intl` call and no `toLocale`**, so neither this gate nor any grep reaches it: a number handed straight to `t()` is interpolated with `String`, which groups nothing and writes a decimal *point* where a German reader writes a comma. The `locale` is in `translate`'s own signature and never reaches the number.

`voice-insights` had all three figures of ONE sentence on two rules — `words` localized, `sources` and `meanSentence` not — so a German reader read `1.234 Wörter · 87 Quellen · 12.5`. **Fixed here**, because two spellings inside one sentence is the drift this change is about.

The rest is **not** swept here and that is deliberate: the right answer differs per site and both directions are wrong somewhere — a quantity belongs in `formatNumber`, an identifier or ordinal (a profile version, a step number) must not be grouped at all. Measured rather than estimated by narrowing the parameter type and reading what the compiler refused: **568 sites across ~60 files**. Filed as #2463 with the fix that makes it stay fixed.

### Also filed

- #2463 — the interpolation class above.
- #2465 — the colour gate cannot see a JSX comment, so `#2463` in one read as a `#rgba` literal and failed the build.

`make check-fe` green on the merged tree: 4418 + 36.

tinnhlt-gra:
## Review round 2 — four more holes, and the one my own mutation sweep caught

Head `af51117d7`. Every thread above is resolved with the change that closes it; each fix carries the mutant that found it, and each mutant is verified applied by diff before its run — a no-op edit reports `MUTANT DID NOT APPLY` rather than a green.

### The gate depended on which Node ran it

The planted fixture hard-coded `new Intl.DurationFormat(locale)`. That only fires where the runtime has it — **Node 22 does not**, CI pins 24 — so on Node 22 the line yields no finding and the whole arm goes red for a reason that is not a defect. It is wrong in the other direction too: one name proves one formatter rather than the set. The fixture now appends **one line per formatter the runtime reports**, so the subject and its proof come from the same place. Thanks — this is the kind of thing that gets a gate turned off rather than fixed.

### A parallel index list, inside the gate that exists to find those

The expectation was line numbers kept beside the fixture, and it was wrong within the hour. Each fixture line now says for itself whether it is a finding.

### A floor that the regression it guards would clear

`expect(files.length).toBeGreaterThan(200)` over a ~900-file sweep: a walk that stopped descending into `screens/` — two thirds of the subject — passes it, and every assertion below then vets a silently smaller tree. It now requires **every immediate subdirectory of `src/` that holds a module file** to be represented, derived from the directory listing. Mutant M15 (drop `screens/`) fails it.

### A guard with no test, because an earlier guard refused every fixture

This one my own mutation sweep found rather than a reviewer, and it is the sharper lesson. `bindsTheNameLocally` closes the case where a file imports the mapping and *then* shadows the name with a local or a parameter — the exact hole both bots raised. Disabling it entirely left the suite **green**: the only fixture that reached it had no import, so `importsTheMapping` refused it at the first branch and the shadowing check was never the decider. Two fixtures now import the mapping AND shadow it, one local and one parameter, and that mutant fails.

On the checker question specifically — both bots asked for full symbol resolution. What landed is the cheap half: the import must exist, and nothing in the file may declare the name. It fails **closed** (an ambiguous file's call sites are not proven) and costs one walk instead of a whole `ts.Program`. The file says exactly that at the subject rather than claiming to be a resolver.

### Two more spellings judged

A method reached by element access (`x["toLocaleString"]()`) and a formatter bound by assignment rather than declaration (`LF = Intl.NumberFormat`).

### One comment corrected

`"en"` **is** valid BCP-47 — a language-only tag. The defect is the unspecified *region*, which is exactly the gap A100 closes. Saying "not a BCP-47 tag" rested a true finding on a false premise, which is the thing this PR is about.

---

**18 mutants, all caught.** `make check-fe` green on the merged tree: 4418 + 36.

tinnhlt-gra:
## UAT — live stack, real formatters, real wire data

https://github.com/user-attachments/assets/d970e0cf-207e-496d-8dd1-51da09bbb2a7

Run against an isolated dev stack (`DEV_SLUG=b15loc`, api `:18941`) seeded through the real API, driving the **shipped** `format/` functions — not a reimplementation of them. The machine running it sits in `Asia/Saigon`, which is what makes the zone half visible.

### The headline: the same instant, a different day

`last_inbound_at` on the wire is `2026-06-05T04:30:00+07:00` — that is 21:30Z on **4 June**.

| | renders |
|---|---|
| **before** `new Date(x).toLocaleDateString()` | `6/5/2026` |
| **after** `formatDate(x, "de", RECORD_ZONE)` | `04.06.2026` |
| | `04/06/2026` (en → en-GB) |
| | `04/06/2026` (vi) |

Two things wrong at once and both are visible in that line: the **day** (a reader east of UTC saw 5 June for a message that arrived on the 4th, so two colleagues quoting one relationship disagreed) and the **format** (`6/5/2026` is US month-first, in a product pinned to en-GB).

### The four private copies

`started_at` is `2024-06-04T00:00:00Z` — and the write side takes it as `format: date`, which is the live confirmation of the corrected `personrail` comment: the stored midnight carries no information.

| | renders |
|---|---|
| **before** four private `toLocaleDateString(undefined, …)` | `Jun 4` — for every reader |
| **after** `formatDayMonth(x, locale, RECORD_ZONE)` | `4. Juni` (de) · `4 Jun` (en) · `4 thg 6` (vi) |

A German reader and a Vietnamese reader were both being shown `Jun 4` on four separate surfaces of one person record.

### The zone bug, reproduced against the real formatter

| `scheduled_tz` | Intl resolves it | `formatDate` throws |
|---|---|---|
| `Etc/GMT-1` | ✅ | ✅ |
| `Etc/GMT+5` | ✅ | ✅ |
| `GMT` | ✅ | ✅ |
| `+01:00` | ✅ | ✅ |
| `Europe/Berlin` | ✅ | — |

All four **resolve**, which is the only question the old probe asked — so it accepted them and the formatter threw one line later, taking the whole scheduled-sends list down over one row. `isRenderableZone` refuses all four and the row falls back to the reader's zone.

### Note on the recording

A browser UAT was the first attempt; Chrome here refuses every loopback address (`ERR_CONNECTION_REFUSED` on `localhost`, `127.0.0.1` and `[::1]`) while `curl` reaches all three, so it is a browser-side block rather than the stack. Rather than keep pushing at it, the harness drives the shipped modules directly against the running API — which for a rendering change is the stronger evidence anyway: it shows the actual before/after strings rather than a screenshot of one of them.

The harness itself is deliberately **not** committed; the assertions it makes that belong in the tree are in `format.test.ts` and `scheduledsends.test.tsx`.

tinnhlt-gra:
## Round 3 — the cheap question first

`importsTheMapping` ran the whole-file shadowing walk **before** checking whether the file mentions the mapping at all. The overwhelming majority of the ~900 swept files never do, so that spent a full AST descent to learn nothing about them. The import scan reads top-level statements only, so it goes first now and the walk runs only for files that pass it.

Same answer either order — both conditions must hold and neither has a side effect — and the mutant that removes the shadowing half **still fails**, verified after the reorder rather than assumed from it.

`origin/main` merged; `make check-fe` green (4418 + 36).

---

### On the one red check

`uat` is red, and it is **not this PR**: it is #2469, which I filed from this branch after confirming it fails on a **docs-only** diff (#2445) that no frontend change can affect. It has since been diagnosed and fixed in #2470 — the e2e mock answered `/v1/me` with locale `"de-DE"`, a value the contract does not allow (`en | de | vi`), so `catalogs["de-DE"]` was undefined and every lookup threw *including the error boundary's own* `t("app.errorTitle")`. The boundary sits inside the locale provider, so it crashed rendering its own message and the page went blank — hence zero elements rather than a wrong count.

`uat` is not part of the required `ci` rollup, and `ci` is green here.

PR #2466, opened 2026-08-23
feat(import): a spreadsheet can correct the companies already held

LarsGradion:
`on_duplicate` took `create` or `skip`. Neither applies a corrected export —
`create` mints a twin of every company already held, `skip` leaves the estate
stale — so the honest answer to *"I have a cleaned-up file of our accounts"* was
to retype it in the app, or to have an assistant loop row by row with no
preview, no transaction and no resume.

`update` writes the row's mapped values onto the incumbent, reported as an
update. It reuses `reconcile`, the same path a row this importer wrote before
takes: the incumbent arriving some other way changes *who* to write onto, not
*how*.

## What it will overwrite

Only a record whose name the row matches **outright — legal form included** — or
one matched by domain. That bar is deliberately higher than the dedupe ladder's.

The ladder strips the trailing legal suffix before scoring, so `Acme Inc` and
`Acme GmbH` reach it as one string and score a perfect match. It routes them to
a human anyway, and `fuzzyOrganization` says why in as many words: *different
legal entities are a human's call, not a merge.* A rule keyed on that score
performs the merge the ladder refused, silently, over the wrong company's data.

The asymmetry that sets the bar: `undo` reverses the rows a run **created**, so a
bad `create` is repairable. An `update` writes over a company the run did not
create, and nothing restores its previous values.

Two earlier drafts were wrong in opposite directions, both now pinned by tests:
requiring `exact_collision` alone was safe and useless (that tier needs a domain,
which a spreadsheet of names has none of); accepting confidence 1.0 was useful
and unsafe.

## Two defects found on the way, neither introduced by the mode

**The preview disclosed matches the caller cannot see.** Every mode's report
counts the row in `duplicates` and names it "already in the CRM", so a preview
answering on an invisible match made the importer an existence oracle for a
colleague's owner-private capture, one CSV row at a time. The preview now always
filters; the commit still decides without the filter, which is correct.

**The report dropped skipped rows.** It deduplicated by the *line* derived from
the external id, and `lineOf` answers `0` for every id not shaped `line N` — every
id in a file with its own `source_key`. Two refused rows were reported as one
skip and one unchanged, so the four counts summed to less than `rows_read`.
Pre-existing; `update` makes it reachable in bulk, since refusing a near miss is
its ordinary behaviour.

## Documentation

Two pages still said consequential verbs stage for approval, which stopped being
true at ADR-0055. Corrected, with the two exceptions that survive named.

New: [`how-to/import-a-company-spreadsheet.md`](docs/how-to/import-a-company-spreadsheet.md)
— the guide this session kept needing. The three-step shape, the fields a company
can receive, a table of what `update` will and will not overwrite, and how to read
a disposition.

## Verification

Full `make check-backend`, `make check-fe` and `make craft-static` green.
Integration lanes green: `compose`, `compose/integration`, `modules/people`.
Each new safety test was verified by breaking the thing it checks.

---
## Summary by cubic
Adds `on_duplicate: "update"` to CSV organization imports so a row can overwrite the matching incumbent instead of minting a twin or being skipped. This enables correcting existing companies from a cleaned spreadsheet while fixing preview privacy leaks and inaccurate skipped-row counts.

- Review notes
  - API: `ImportOnDuplicate` now includes `update`; regenerated contracts and `frontend/src/api/schema.d.ts`; MCP `preview_import` input schema accepts `update`.
  - Duplicate policy: new `csvduplicatepolicy.go` writes only on domain match or `SameOrganizationName` (exact name, legal suffix preserved, same name axis). Fuzzy or ambiguous matches are skipped. Previews filter by visibility; commits decide unfiltered; skip reasons no longer disclose owner-private records.
  - Update path: reuse `reconcile`; previews label updated vs unchanged. A non-locking pre-read can race, so a previewed create may commit as an update; the overwrite rule still guards correctness.
  - Reporting: created+updated+unchanged+skipped now equals rows_read; `duplicates` reported separately. Skipped-row dedupe keys by external id, fixing undercounts when `source_key` is present.
  - Tests: certainty gate (legal-form differences refused; cross-axis name matches refused), unchanged accounting, privacy-safe reasons, external-id dedupe, and end-to-end update integration. The invalid `on_duplicate` error lists accepted values from the contract with a guard test.
  - Docs: new “Import a company spreadsheet” how-to; agent surface and MCP reference updated; tool budget snapshots refreshed.

- Rollout
  - No migration. Regenerate SDKs/clients to accept `on_duplicate: "update"`. Installations may floor `commit_import` to confirm-first. Undo cannot restore fields overwritten by `update`.

<sup>Written for commit aaa81bf38855904399243f6375fd0460942f22e3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/2466?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added `update` as a duplicate-handling option for company CSV imports.
  - Exact-name matches can update existing records without creating duplicates.
  - Similar names, legal-form variants, ambiguous matches, and inaccessible records are skipped safely.
  - Import previews and reports distinguish created, updated, unchanged, skipped, and unresolved rows.
  - Duplicate counts and skipped-row reporting are more accurate and preserve individual rows.
- **Documentation**
  - Added a company spreadsheet import guide covering mapping, previews, duplicates, permissions, and reporting.
  - Updated import and assistant documentation to describe the new duplicate-handling behavior.

### DIFF
--- a/backend/api/crm.yaml
+++ b/backend/api/crm.yaml
@@ -16803,7 +16803,7 @@ components:
 
     ImportOnDuplicate:
       type: string
-      enum: [create, skip]
+      enum: [create, skip, update]
       default: create
       description: |
         What to do with a row naming a record the estate ALREADY holds — found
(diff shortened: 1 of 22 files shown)


LarsGradion:
Closing in favour of an id-based update, which is correct by construction rather than by enumerating the ways a name match can hit the wrong company.

Three reviews found three distinct permanent-overwrite paths in the name-matching rule: differing legal forms scoring identically, two companies legitimately sharing a name with a uuid tie-break, and a cross-axis pairing hiding a second equally-good match. Each fix was right for the case in front of it and blind to the next.

The cause is structural. The dedupe ladder answers "should a human look at these two?" and deliberately blurs — it strips legal suffixes, scores across name axes, breaks ties arbitrarily. For proposing a review that blurring is free; for choosing a record to overwrite irreversibly every blur is a way to hit the wrong one.

An id column removes the question. The disclosure and accounting fixes found along the way are real and predate the feature; they ship separately.

PR #2471, opened 2026-08-23
feat(ci-tests): a uniqueness claim names the gate that holds it, or it is debt

tinnhlt-gra:
## What this is

A comment saying a declaration is the ONLY one of its kind is not decoration. It
is what stops the next author looking — they grep, they find the claim, they
stop. **So a false one is worse than silence**: silence at least leaves the
search on.

The rulebook has said so since *Reuse before you build* rule 4:
> a comment may not claim to be the only implementation unless a test holds it.

Instruction was not enough. A sweep audited ten such claims exhaustively and
**nine were false**:

| the claim | the truth |
|---|---|
| "the only definition of a current employment in this product" | eight compose sites spelled the predicate by hand — the exact notice-period bug the comment said it existed to prevent |
| "the one spelling every table uses" for a credential handle | there were three; the two nobody counted carried ciphertext past a workspace wipe |
| "the ONE way a caught failure becomes words on a screen" | five screens bypassed it, one putting an RBAC object and verb in front of a user |

**None was found by a gate.** Every one was found by a human re-reading a file
for some other reason, months later.

Duplication audit item **D1**.

## What it does, and what it deliberately does not

It does **not** audit the claims. It cannot: there are **649**, and whether one
is true is a question about the whole tree. It makes the class **stop growing**,
which is the half a test can hold.

- **A claim that names its gate is held.** `Held by: TestName (path)` in the same
  doc comment — and the named test must *exist*, in the file named, with a
  signature `go test` will actually run. A bare test name is refused.
- **Everything already here is in `backend/uniquenessclaims.txt`** — **646
  entries**, a debt register closed to new claims. One that is neither
  registered nor bound fails, so a claim written tomorrow must be held the day
  it is written.
- **The count is pinned** (`registeredDebt`) and checked in both directions, so
  the register cannot quietly grow. Lowering it is the point; raising it is what
  a widened detector shape requires.
- **An entry that stops describing an unheld claim fails**, so the register
  tracks the tree rather than rotting beside it.

**646 entries, no per-entry reasons, deliberately.** 646 rationalisations
written by somebody who has not audited the claims would be worse than an honest
count of what is unaudited. The number is the point.

## What the detector hard-codes, said plainly

The shape list **cannot be derived**. There is no grammar for "a sentence
asserting uniqueness" the way there is a runtime answer for "which `Intl`
members take a locale" — prose is prose. Pretending otherwise is how a gate ends
up hard-coding a fragment of its own subject and going quietly blind.

So the list carries two obligations, both enforced:

1. **Every shape must be attributed a claim by the real sweep**, and the arm
   logs which one. A shape matching nothing fails — a detector matching nothing
   reports a clean tree. *This is derived from the tree, so there is nothing to
   fabricate.*
2. **A near-miss corpus, one case per shape**, must not match. A shape widened
   until it reads ordinary English fails.

**Deliberately out of scope, measured rather than guessed:** "the one place",
"the only way", "the one path" and "rather than a second" are prose here far
more often than claims — **540 hits**, mostly *"the only way to ask the
DATABASE"*. And `never-twice`'s bare `two` was removed at a measured cost of
~20 real sites, because it read *"the grace window is not two hours but three"*
as a claim. Both gaps are named in the file rather than left to be discovered.

### One shape IS derived, and it had to be

`is EVERY` was case-sensitive to avoid matching *"every bit as guarded as the
last"*. That made the gate blind to `readProfileFields is every read of
person_profile_field` — **the worked example the whole rule is argued from**.

The discriminator is now Go's own doc convention: a doc comment opens with the
identifier it documents, so the claim form is `<the declaration's name> is
every …`. **That found 83 more claims.**

## Scope: doc comments on declarations

| swept | count |
|---|---|
| every comment in the tree | 4308 — overwhelmingly ordinary English |
| declaration doc comments, loose shapes | 994 |
| **declaration doc comments, sharp shapes** | **649** |

Struct fields and interface methods are included — a field's doc comment *is* a
doc comment on a declaration.

## Three claims are HELD, which proves the mechanism end to end

| claim | held by |
|---|---|
| `EmploymentIsCurrentSQL` — "the ONE spelling of 'this job is still theirs'" | `TestEveryEmploymentCurrencyTestUsesTheOneDefinition` |
| `CurrentPrimaryEmploymentSQL` — "Spelled once so a new reader cannot trust the flag alone" | the same census |
| `readProfileFields` — "is every read of `person_profile_field` that RENDERS it" | `TestEveryReaderServingProfileFieldValuesConsultsTheVerdictLedger` |

All three tests already existed. **Nothing pointed at them.**

## Files

- `backend/uniquenessclaimsdetector_test.go` — which comments are claims, which
  declaration each belongs to, and the key that tells two of them apart.
- `backend/uniquenessclaims_test.go` — the rules and their fixtures.
- `backend/uniquenessclaims.txt` — the register.

Split because the file reached 1014 lines against a 1000 cap and the craft gate
refused the push. Both sources are skipped by the sweep, since both spell out
the phrases they hunt for.

## Reviewed in five rounds; what they found

Every finding reproduced before it was fixed. The pattern is worth naming: **almost
every one was a guard the tree happens not to exercise**, so deleting it left
everything green.

| finding | why it mattered |
|---|---|
| the register key collided (`func Close` for every `Close` in a file; `const block` for every block) | a collision is a **standing permission for a new claim** — demonstrated end-to-end across ~80 files |
| `internal/modules/contracts/` was skipped by a bare directory-NAME match | a whole product module needed no gate |
| struct fields and interface methods escaped entirely | a field doc is a declaration doc |
| a claim could name a gate `go test` will never run (`Testhelperghost`) | it compiles, exists, and cannot execute |
| the binding's file check matched inside a FILENAME | `(claims_test.go)` bound against a file that does not exist |
| **the corpus said "taken VERBATIM from this tree" and most was invented** | the defect this file refuses, in its own docblock |
| the near-miss corpus covered four shapes of nine | the missing ones could be widened freely |
| the ceiling did not exist — "closed to new entries" was counted by nothing | a line added for a claim written this morning passed every arm |
| the embedded-field key was asserted *beside* the walk, not on it | the branch never ran; losing it stayed green |
| a trailing `.` / `$` / quote in the table name | different relations reported as the target |

And **fixing the key silently broke the derived shape** for every method — 24
claims including the one the docblock argues from — while every arm stayed
green, because they ask whether a shape matches ANYTHING. There is an arm for
that now.

Comment residue was corrected throughout: the rulebook forbids review and fix
narration in source, and this had drifted well past it.

## Verification

- `make check-backend` — **green**, with a private `GOLANGCI_LINT_CACHE`.
- Re-run after each rebase onto `main`, since a claim landing meanwhile needs registering.
- ~20 mutants across the rounds, each **verified applied by diff** before its run, and each re-run until it **compiled** where the first attempt did not.
- Roots swept: `backend/`, `extensions/`, `desktop/`, `cli/`, `fixtures/`. Only `backend/` is required non-empty, and the reason each other root may be empty is stated.


tinnhlt-gra:
## Review round 1 — eight findings, five of them demonstrated as live escapes

Every one reproduced before it was fixed, and each fix re-verified by re-running the reviewer's own probe. Head `52ecac091`.

### The register key was not unique — and a collision is a standing permission for a NEW claim

`func Close` was the key for **every** method called `Close` in a file, and every multi-spec block keyed as the bare word `const block`. Demonstrated: adding a second `const` block to a file that already had a registered one let a **brand-new unheld claim** through the gate whose entire promise is that it cannot. ~80 files carried such a key, and **seven real claims were already aliasing into three register lines** — so the count was wrong *and* the idle-entry arm was blind exactly there.

Methods carry their receiver now; blocks are named by their first declared name.

### A whole product module was invisible

The skip list matched the directory **name** `contracts`, catching the generated `internal/contracts` **and** the hand-written `internal/modules/contracts/`. A claim written anywhere in the contracts CRM module needed no gate at all. Matched by path now.

### Struct fields and interface methods escaped entirely

A field's doc comment *is* a doc comment on a declaration. `Addresses is EVERY address this record names` sits in one, and no arm ever saw it.

### A claim could name a gate `go test` will never run

`Test[A-Za-z0-9_]*` admits `Testhelperghost()` — legal Go, never executed. The name now needs the uppercase continuation `go test` requires, and the signature must take `*testing.T`.

### The binding's file check matched inside a FILENAME

`Held by: TestX (claims_test.go)` bound against `uniquenessclaims_test.go` — **a file that does not exist**. Path-segment suffix now.

### The corpus said "taken VERBATIM from this tree" and most of it was invented

This is the one that matters most, because it is the defect this file exists to refuse, committed in its own docblock. `companyform.go` does not exist. `signalTone` does not exist. The `is EVERY` case was a real sentence **with its `every` upper-cased** so the case-sensitive pattern would match it.

The positive proof is **derived** now: every shape must be attributed a claim by the real sweep, and the arm logs which claim. There is nothing left to type, so nothing left to fabricate.

### The near-miss corpus covered four shapes of nine

…and the derived shape was in neither loop. Every shape has a case both ways now — and the very first one written for `never-twice` found the shape too loose: a bare `two` read *"the grace window is not two hours but three"* as a claim. Bare `two` is gone, at a **measured** cost of ~20 real sites, stated in the file rather than left to be discovered.

The derived shape also had a live false positive its own docblock had warned about: `Flush is every bit as ordered as Write`. `every bit` and `every inch` are English intensifiers, named as such — a closed register that swallows an innocent sentence teaches people to write worse comments, not fewer false claims.

### And fixing the key broke the derived shape, silently — the best finding of the round is one the fixes produced

Receiver-qualified labels meant the derived shape looked for `Service.readProfileFields` in a comment that says `readProfileFields`. It stopped matching **every method in the tree** — 24 claims, including `readProfileFields`, *the example the whole docblock argues the rule from* — and **every arm stayed green**, because they ask whether a shape matches ANYTHING and its plain-function claims still did.

There is an arm for that now: the derived shape must be attributed a claim on a `func` **and** on a `method`. Mutation-verified by putting the receiver qualifier back.

---

Register: **646** entries, up from 638. The gate sees more than it did, which is the direction that number is allowed to move for a reason.

`make check-backend` green.

tinnhlt-gra:
## Review round 3 — two more key collisions, and the comment rule

Head `d1386b117`, `origin/main` merged.

### The gate skipped every file named like itself

It skips its own file because it spells out the phrases it hunts for — but `filepath.Base(path) == gateFile` skipped **any** `uniquenessclaims_test.go` in any swept tree, so a nested one could carry an unbound claim. Matched by path now; a planted nested file fails.

### Embedded fields collided, the same way methods did

`genericReceiverName` returned `"?"` for a selector expression, so `io.Reader` and `fmt.Stringer` embedded in one struct shared the key `type X.<embedded ?>` — a register entry for one authorising a later claim on the other. That is exactly the collision the receiver rule removed for methods, one declaration kind over, and I missed it because I fixed the kind the reviewer had named rather than the *class*. Package-qualified now.

### And the comments were stating history where they should state rules

This is the one worth calling out, because the rulebook says so plainly and I broke it repeatedly. AGENTS.md, *Rules learned from the review loop* #4:

> **No build-process residue in comments** — no review-ticket numbers or fix narration; state the invariant so it stands alone. History belongs to git, not the source.

Mine had drifted well past that: *"it did, once"*, *"a reviewer grepped them"*, *"the first version covered four of nine"*, a ticket number in a doc comment. **Fourteen passages rewritten** to say what holds and what it costs, keeping every measured number and dropping the story of how each was found.

The test is whether the invariant reads the same to somebody who was not here. It does now, and it did not before — which is the more useful way to state the rule than "no narration", because narration is easy to write without noticing you are the only reader it serves.

---

All five threads resolved. `make check-backend` green on the merged tree.

tinnhlt-gra:
## Review round 4 — the closure is counted now, and the key is proven through the walk

Head `80e0a65db`.

### "CLOSED to new entries" was a claim nothing counted

The sharpest finding of the round, and it is this gate's own subject turned on itself. The other arms check that each register line describes a **live unheld claim** — which a line added for a claim written this morning satisfies perfectly. Only a count catches that, and there was none.

`registeredDebt` is that count. It fails in **both** directions: too many entries, and a ceiling left standing above a register that shrank. Raising it is legal and is exactly what a widened detector shape needs — as the ONE line a reviewer agrees with, rather than growth spread across a diff nobody reads to the end.

### The embedded-field key was asserted beside the code, not on it

The probe's embedded fields carried no doc comment, so `recordFields`' embedded branch **never ran during the walk**. Losing the owner scope or the `<embedded …>` shape would have kept the suite green — the collision returning under a test named for holding it. Both fields carry a doc now and their labels come from the walk's own output.

Three mutants: drop the owner scope, stop visiting embedded fields, stop qualifying the selector. Each caught, each **compiling** — two earlier attempts failed the build, and a mutant that does not compile tests nothing.

### The file split, because the craft gate refused the push

1014 lines against a 1000 cap, and it was the right refusal: the file had grown to answer two questions. `uniquenessclaimsdetector_test.go` holds which comments are claims and how a declaration is keyed; `uniquenessclaims_test.go` keeps the rules and their fixtures. Both are skipped by the sweep — the skip is a **set** now, and removing the second entry fails.

### Comment residue, again

The register header and the docblock were explaining a commit procedure. How to sequence a change is PR guidance; source states the invariant. Reduced to it.

---

All threads resolved. `make check-backend` green.

PR #2481, opened 2026-08-23
fix(records): one writer moves a visit baseline, and person360 stops copying it

tinnhlt-gra:
## What this is

`user_record_view` carries one fact per (user, record): the moment that human
last said *"I have seen this"*. Every 360 surface measures its since-last-visit
counts against it.

**Two byte-identical statements wrote it.** `org360`'s — which **owns** the
table, declared in its `doc.go` and named in `tableownership_test.go` — and
`person360`'s, which acknowledges the person view under an ownership waiver.
Nothing differed between them, which is precisely why nothing signalled that
they had to stay the same.

Duplication audit item **B6**.

## The correctness is in one word

```sql
DO UPDATE SET last_viewed_at = GREATEST(user_record_view.last_viewed_at, EXCLUDED.last_viewed_at)
```

`GREATEST` is what stops a slow tab's late-arriving acknowledgement rewinding a
newer one — two tabs open on the same record converge on the later visit instead
of racing the baseline backwards.

A copy that lost it (written from memory, simplified, a merge resolving the
wrong way) would hand a reader back an unread marker **they had already
consumed**, on a surface where "3 new" and "0 new" are both plausible and
neither looks like a bug. No error, no log line, and no way for the reader to
know.

## What moved, and what deliberately did not

`org360.RecordVisit` is the statement now, exported by the package that owns the
table. It takes the caller's **transaction** rather than opening one, so the
visibility gate and the write cannot land in different transactions — a record
that stopped being visible between them would otherwise still get a baseline.

What each caller **keeps** is the part that legitimately differs:

| caller | gate | why |
|---|---|---|
| `org360` | `EnsureVisible` | an account the caller cannot read must not be confirmed to exist |
| `person360` | `EnsureVisible**Live**` | Art. 17 anonymizes a person in place and stamps `archived_at` while **leaving `owner_id` alone**, so the plain probe still admits their owner |

That difference is a ruling per record type. The upsert is not.

## The ownership waiver is deleted, not reworded

`person360` no longer writes the table, so the exception has nothing left to
ratify. The register of exceptions in `tableownership_test.go` is one entry
shorter — the direction it should move, and a reminder that a waiver is worth
checking for whether its *subject* still exists, not only whether its reason
still reads well.

## The gate

`TestUserRecordViewHasOneWriter` censuses every string literal in the tree.

**Per LITERAL, not per line.** This upsert spans six source lines; a line-scoped
sweep would see `INSERT INTO user_record_view` and `DO UPDATE SET …` as
unrelated fragments — matching the first and never asking whether the second
carried `GREATEST`.

**It judges `UPDATE` as well as `INSERT`,** because the defect is not "a second
INSERT" but a second answer to *how the mark moves*. The worst second writer is
a bare `UPDATE … SET last_viewed_at = $1`: no `GREATEST` at all, rewinding
unconditionally.

**And it reads every spelling Postgres accepts of the name** — this took three
rounds to get right, each round a shape the pattern could not see:

| | |
|---|---|
| `public.user_record_view` | one qualifier |
| `tenant1.public.user_record_view` | two — so the qualifier repeats |
| `public . user_record_view` | whitespace around the separator |
| `"public"."user_record_view"` | quoted, in any combination |
| `UPDATE ONLY user_record_view` | inheritance-scoped, and it rewinds just as hard |

The name ends on a **delimiter, not `\b`**: `$`, `.` and `"` are all legal
around a Postgres identifier and none is a word character, so a boundary alone
reported `user_record_view$archive`, `"user_record_view.extra"` and
`user_record_view.extra` — different relations — as this one. The quoted branch
is case-**sensitive** while the rest is not, because that is Postgres: unquoted
`USER_RECORD_VIEW` folds to this table, quoted `"USER_RECORD_VIEW"` is a
different one. Go's regexp has no lookahead, so that distinction is two branches.

`TestTheViewBaselineCensusSeesWhatItClaimsTo` plants **12 must-catch and 11
must-miss**. The must-miss half carries every near-neighbour — `SELECT` (which
every 360 does), `user_record_view_archive` in bare, qualified and `ONLY` forms,
and the three false positives above — because widening a pattern is how a
boundary gets lost, and a gate that starts reporting the archive table is a gate
somebody turns off.

## Mutation-tested

| mutant | result |
|---|---|
| restore person360's copy of the statement | caught |
| a second writer that loses `GREATEST` | caught |
| drop the `ONLY` branch | caught |
| make the qualifier non-repeating | caught |

The first was re-run until it **compiled** — the initial attempt failed the
build, and a mutant that does not compile tests nothing.

## Coverage

`RecordVisit` is the only new executable code, exercised through **both** real
callers rather than by a unit test of itself:

- `TestOrganizationViewAckIsMonotonicAndPerUser` builds a second service whose
  clock **lags by an hour** and asserts the mark keeps the later instant — that
  is `GREATEST` asserted at the database, the exact property the census
  protects.
- `person360_http_integration_test.go` covers the person side through the HTTP
  door.

A unit test of `RecordVisit` in isolation would exercise the function while
proving nothing about either caller reaching it.

## One finding refused, with evidence

CodeRabbit flagged `RecordVisit` for committing a domain row without an audit
and outbox row, quoting the write-shape rule correctly. It does not apply:
`user_record_view` is ratified as audit-free in **two places predating this
change** — `moduleaudits_test.go:62` ("per-reader: one person's last-seen
cursor") and the ownership entry ("view state, not a record fact … the
saved-view ruling"). This change adds no write; it moves one. The load-bearing
part is that **the write-shape gates pass**, which is what makes it a recorded
exception rather than an asserted one.

## Verification

`make check-backend` green, with a private `GOLANGCI_LINT_CACHE`.

### DIFF
--- a/backend/internal/compose/org360/viewbaseline.go
+++ b/backend/internal/compose/org360/viewbaseline.go
@@ -86,11 +86,11 @@ func (s *Service) Acknowledge(ctx context.Context, orgID ids.OrganizationID) (cr
 // statement in this product that writes user_record_view.
 //
 // Exported because a SECOND record type acknowledges visits — person360 — and
-// it held a byte-identical copy of this upsert. Two writers of one table's
-// write shape drift silently: `GREATEST` is the whole correctness argument, and
-// a copy that lost it would rewind a baseline on a late-arriving ack from a
-// slow tab, consuming an unread marker nobody consumed. Nothing in the two
-// statements differed, so nothing signalled that they had to stay the same.
+// a caller reaching for its own upsert instead is what this export exists to
+// prevent. `GREATEST` is the whole correctness argument: a copy that lost it
+// would rewind a baseline on a late-arriving ack from a slow tab, consuming an
+// unread marker nobody consumed. Two writers of one write shape signal nothing
+// while they agree, and they agree right up until one of them is edited.
 //
 // What the callers keep is the part that legitimately differs: their own
 // visibility gate. org360 asks `EnsureVisible`, person360 asks
--- a/backend/internal/compose/org360/viewbaseline.go
+++ b/backend/internal/compose/org360/viewbaseline.go
@@ -69,13 +69,8 @@ func (s *Service) Acknowledge(ctx context.Context, orgID ids.OrganizationID) (cr
 		if err := auth.EnsureVisible(ctx, tx, "organization", orgID.UUID); err != nil {
 			return err
 		}
-		return tx.QueryRow(ctx, `
-			INSERT INTO user_record_view (user_id, entity_type, entity_id, last_viewed_at)
-			VALUES ($1, $2, $3, $4)
-			ON CONFLICT (user_id, entity_type, entity_id)
-			DO UPDATE SET last_viewed_at = GREATEST(user_record_view.last_viewed_at, EXCLUDED.last_viewed_at)
-			RETURNING last_viewed_at`,
-			userID, entityTypeOrganization, orgID, now).Scan(&stored)
+		stored, err = RecordVisit(ctx, tx, userID, entityTypeOrganization, orgID.UUID, now)
+		return err
 	})
 	if err != nil {
 		return crmcontracts.RecordViewAck{}, err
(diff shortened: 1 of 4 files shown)


tinnhlt-gra:
## Review round 1 — two taken, one refused with evidence

Head `cadd23a39`. `make check-backend` green.

### The census could not see a schema-qualified or quoted name

`UPDATE public.user_record_view` and `UPDATE "user_record_view"` both passed a pattern anchored on the bare word — **exactly like a clean tree**, which is the one direction this gate must not fail in. That is also this file's own stated rule ("plant the shape it cannot see") applied to itself and failed, which is worth naming rather than quietly fixing.

The optional qualifier is in the pattern now. Three spellings planted as must-catch (`public.`-qualified, quoted, and both), and the **qualified form of the near-neighbour** `user_record_view_archive` planted as must-miss — because widening a pattern is how a word boundary gets lost, and a gate that starts reporting the archive table is a gate somebody turns off.

### The comments narrated the duplication rather than stating the rule

Same rulebook line as the sibling PR: history belongs to git. What survives is the invariant and what it costs — *two writers of one write shape signal nothing while they agree, and they agree right up until one of them is edited*.

### Not taken: the write-shape finding

> `RecordVisit` mutates `user_record_view` but does not call `storekit.Audit` + `Emit`. Every mutation commits domain row + audit_log row + event_outbox row in ONE transaction.

The rule is quoted correctly and it does not apply here. `user_record_view` is **ratified as audit-free in two places that predate this change**:

- `backend/moduleaudits_test.go:62` — `"internal/compose/org360": "user_record_view and suggestion_dismissal are per-reader: one person's last-seen cursor and one person's dismissals"`
- the ownership entry in `tableownership_test.go` — *"view state, not a record fact, so it is written without an audit row — the saved-view ruling"*

and `org360/viewbaseline.go`'s own docblock states the reason: it is written on every visit, no other user may read it, and no consumer can act on it.

**This change adds no write.** It moves an existing statement out of a second copy and into the package that owns the table. If the write shape were owed here it was owed before, on both copies.

The load-bearing part is that **the write-shape gates pass**, which is what makes this a recorded exception rather than an asserted one — the distinction that matters, and the reason I checked rather than replying from the docblock alone.

tinnhlt-gra:
## Coverage of the extracted writer

`org360.RecordVisit` is the only new executable code in this change, and it is exercised through **both** real callers rather than by a direct unit test of itself:

- `internal/compose/integration/org360/org360_visit_integration_test.go` — `TestOrganizationViewAckIsMonotonicAndPerUser`, which builds a **second service whose clock lags by an hour** and asserts the mark keeps the later instant. That is `GREATEST` asserted at the database, and it is precisely the property the new census exists to protect.
- `internal/compose/integration/person360_http_integration_test.go` — the person side, through the HTTP door.

Worth stating because the alternative was tempting and worse: a unit test of `RecordVisit` in isolation would exercise the function while proving nothing about either caller reaching it. Both paths run the real statement in a real transaction after their own visibility gate, which is the part that differs between them.

The rest of the diff is comments, one deleted waiver, and a test file.

PR #2483, opened 2026-08-23
fix(records): one answer when a domain is already taken

tinnhlt-gra:
## What this is

A domain maps to at most one organization (data-model §4.2), so *"is this domain
taken?"* is one question. It had **four answers**.

Duplication audit item **B7**.

| door | answer |
|---|---|
| create an organization | typed `DuplicateDomainError`, id gated by `auth.VisibleTo` |
| edit its domains | **byte-identical copy** of the above, one clause apart |
| save its profile website | **a bare conflict sentinel** — no type, no id |
| the unique-index race | typed error, no id |

## The user-visible half

Saving a company profile with a domain another organization owns reached the
client as a **409 with no `existing_id`**, while the domains endpoint sent
`duplicate_domain` **naming the company that holds it**. Same collision, two
doors, and the profile door dropped the one fact that makes it actionable — the
reader was told "conflict" and left to find the owner themselves.

## The dangerous half

The two probes were identical **including the disclosure rule**:

> `ExistingID` is filled only when `auth.VisibleTo` says the caller could have
> read that row anyway.

That rule exists because answering "taken" already reveals an organization exists
with that domain, and naming *which* reveals a record the caller may not read.

**A second copy of an existence-hiding rule is where a leak appears**, because
nothing fails when one copy stops asking. The copy keeps answering — it just
answers with more than it should, on a path nobody re-read. No error, no failing
test; the client simply receives an id it was not entitled to, in a field that is
supposed to be sometimes-absent.

## And the rule was untested

Every existing case asserted the id **is** carried. Nothing asserted it is
**withheld** — the half a leak lives in. I removed the visibility gate and the
whole suite stayed green.

There is an integration test for it now, and it **fails by naming the leaked
organization id**:

```
409 disclosed an out-of-scope organization 01a02fe6-65fd-7706-9706-5d9d3a685e80
--- FAIL: TestDuplicateDomain409DoesNotDiscloseAnOrgOutOfScope
```

**One test covers all four doors**, which is what the extraction buys. Before,
each could have drifted on its own.

Its fixture needed a local perms set, and both halves of that are load-bearing:
without `organization: Create` the probe never runs and the test reports a
permission denial instead of a disclosure verdict — which is what it did on the
first run — and an unbounded caller can see every org, making the withheld case
unreachable.

## The census that keeps it

`TestEveryDomainClaimAnswersThroughOneProbe` needs **both halves** of its
pattern: `organization_id` in the projection **and** a filter on the domain. The
projection alone matched the bulk read that fetches every domain row for a *set*
of orgs — the opposite direction — and it did so **in the owner's own file**,
where the census trusts what it finds. The owner is required to hold **exactly
one** such statement for the same reason: skipping the owner wholesale would let
a second probe live beside the first, at the one address nothing checks.

**Two readers are ratified rather than adopted.** Dedupe tier 1
(`exactOrgByDomain`) and the signal resolver ask *"which org does this domain
resolve to"*, answered to the system so a record can land or be scored — never
to a caller as a refusal, so no disclosure rule governs them. Ratified **by
name** because there is no syntactic difference between the two questions: any
pattern narrow enough to miss them would also miss a real second probe, and miss
it silently. A reason a reviewer can disagree with beats a regular expression
nobody can see the edge of.

## Mutation-tested

| mutant | result |
|---|---|
| the company door returns a bare sentinel again | caught |
| a second probe inside the **owner** file | caught |
| an idle waiver entry | caught |
| **the disclosure gate removed** | caught, against a real database, by name |

## One gate caught me

`TestEveryMutationOfAShareableRecordProbesForWriteAuthority` reported both
write-authority waivers as stale — they were keyed on the two function names I
had just collapsed into one. Merged into a single waiver carrying both reasons.
That is the gate working exactly as intended, and worth naming: a waiver keyed on
a name does not survive a rename, and nothing else would have said so.

## Verification

- `make check-backend` — **green**.
- `make test-integration` — **green, 0 skips, 43 packages**.
- The new test run against a real database both ways: passing on the fix, failing on the mutant.

---
## Summary by cubic
Unifies duplicate-domain detection behind one probe and aligns all doors to return a typed 409 with a visibility-gated owner id. Previously the profile website path returned a bare conflict; now all paths return `DuplicateDomainError` and keep the response shape stable, including under unique-index races.

- Routes organization create, domain edit, and profile website save through `claimedDomainOwner` in `internal/modules/people`; returns `DuplicateDomainError` with `existing_id` only when `auth.VisibleTo` admits it.
- Checks `Scan` errors before comparing ids; the unique-index race now returns the typed error with no id.
- Adds an integration test that withholds out-of-scope `existing_id` and confirms the visible case includes it.
- Adds a census that enforces a single domain-claim probe; both censuses read SQL via `sqlStatementsIn` (decodes escapes, flattens `+` concatenations).
- Updates write-authority waivers to `claimedDomainOwner`.
- Client impact: the profile website save now returns `duplicate_domain` and, when visible, an `existing_id`. No migration required.

<sup>Written for commit 406a4f9f290a627229965d438cb8a0d7ea1800e3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/2483?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved duplicate company domain handling during creation and editing.
  * Prevented disclosure of company IDs outside the viewer’s permitted scope.
  * Preserved clear conflict responses for concurrent domain claims.
  * Ensured domain checks handle database errors safely.
  * Standardized domain ownership validation across company workflows.

* **Tests**
  * Expanded coverage for duplicate-domain conflicts, visibility protections, concurrent claims, and ownership validation.
  * Added safeguards to keep domain ownership checks consistent across workflows.


tinnhlt-gra:
**Correction on the two comment threads above:** I resolved them before the fix had actually landed. My edit failed to apply (a mismatched anchor), the working tree was clean, and I resolved anyway on the assumption it had worked. Both are genuinely fixed as of `406a4f9f2` — the package comment states why there is one reader rather than how there came to be one, and the function's doc names `sqlStatementsIn` instead of the `sqlLiteralsIn` the rename removed.

Flagging it because a resolved thread is a claim that something was done, and that claim was briefly false. Verified this time by grepping for both stale strings and getting zero.

---

## Round 3 summary

| finding | verdict |
|---|---|
| `ast.BasicLit.Value` is source text, so escapes never decode | **fixed** — and it was already shipped in the census that merged this morning |
| the two censuses read SQL two different ways | **collapsed** into one function |
| the shared reader's comment narrated its own history | fixed |
| its doc named the removed helper | fixed |

The first is the one worth remembering. Measured against the merged census:

| spelling | matched |
|---|---|
| `"INSERT\nINTO user_record_view …"` | **false** |
| `"INSERT INTO\nuser_record_view …"` | **false** |
| the same in backticks | true |

The tree happens to write these in backticks, which is the only reason nothing failed. A second writer added in double quotes would have passed that gate while sending exactly the statement it exists to find.

Collapsing the two readers is what makes that fix stick: they were the same walk under two names in one package and had already drifted, so fixing one would have left the other narrower — and a census whose reader is narrower than its sibling's reports a clean tree over what the sibling sees.

`make check-backend` green.

PR #2490, opened 2026-08-23
feat(ai-models): every prompt says what language to answer in

LarsGradion:
## The defect

A model asked nothing about language answers in whatever language its input happened to be in.

That is right for a reply to a German thread and wrong for a claim filed on a record: the thread has one reader who wrote it, the record has everyone. So a Vietnamese conversation produced a Vietnamese claim a German colleague then had to read, and nothing in the product had ever decided otherwise — **26 prompt sites, roughly two dozen of them saying nothing about language at all**.

## The change

`promptlang.Rule` is where the instruction is written. The gate in `backend/promptlanguage_test.go` finds every `model.Request` literal in `internal/compose` and `internal/modules` and fails when a new one carries no rule and no waiver. A waiver needs a reason in source, the same bar `//craft:ignore` holds.

### Shared writing takes the installation's base language

The deal status card, drafted offer lines, extracted signal summaries, transcript proposals, domain-triage reasons, and an agent run's final summary. Each is filed on a record the whole team reads.

### Onboarding takes the admin's locale, not the base language

It has one reader. Those prompts already tried, with their own `Respond in <locale>.` sentence — a **second spelling of this rule missing the carve-outs**, so a model told to write German could translate a JSON key or a status value into a parse failure. They now compose the same rule as everything else.

### Ten sites are waived because they return data, not prose

Classification enums, verdicts, signature fields, page facts, profile values, document extractions. Not style exemptions: each is checked by a gate requiring the value to appear character-for-character in a foreign-language source page, so a "write in German" instruction would **drop every fact it extracted**. The voice builder is waived for the opposite reason — it describes how one person writes, from their own samples.

## Deferred, and labelled deferred rather than exempt

`orgbrief`, `orgdossier` and growth-fit are cached per input fingerprint, and language has to enter that key **with** the rule. Otherwise an installation that switches language keeps serving the old-language text until some unrelated fact moves, and the setting appears to do nothing. Those fingerprints are being reworked concurrently in #2486, so the two edits belong in one change rather than as a conflict.

`/v1/site-reads/{id}/messages` carries no locale field the way its onboarding sibling does, so it has no language to pass. Adding one is a contract change.

## What the gate can and cannot see

It reads syntax: it proves a rule was **attached**, never that the right language resolved. Two things cover the rest.

- `compose/integration/promptlanguage_integration_test.go` writes a settings row and reads the language back out of a rendered rule — the whole chain, through `compose.BaseLanguageForPrompt` rather than a hand-copied settings read.
- The runner is waived because its rule arrives pre-rendered from compose across a package boundary the gate cannot follow. Two tests in that package hold what it cannot see, including that an absent rule leaves no stray separator behind.

A second spelling of the rule now fails `TestOnlyPromptlangSpellsTheLanguageRule`.

## One check that proved nothing until it was fixed

That uniqueness gate first searched for `promptlang.Heading`'s **runtime** value — `LANGUAGE` plus a real newline. No Go source file contains that: source spells it as the two characters `\` and `n`. It passed against a deliberately planted rival rule and was vouching for nothing. It now searches both source quotings, and was re-checked against a planted rival in escaped and raw-literal form.

## Verification

`make check-backend`, `make craft-static`, `make rls-store-path`, `make no-jurisdiction` — all green.

Mutation-checked, each mutant confirmed to compile before its result was believed:

| mutant | result |
|---|---|
| the runner drops the rule it was handed | **FAIL** |
| the runner writes its separator unconditionally | **FAIL** (catches the empty case) |
| German loses its English name | **FAIL** — two tests, one naming the language |
| every language renders English | **FAIL** — identical rules for two languages |
| the carve-out list is deleted | **FAIL** ×5 |
| a second rule renderer appears (escaped literal) | **FAIL** |
| a second rule renderer appears (raw literal) | **FAIL** |
| onboarding ignores the requested locale | **FAIL** |

Five onboarding tests asserted the literal `Respond in de.` / `Respond in en.`. They now assert through `promptlang.Rule` — a copy of the wording in a test is what let those assertions go stale in the first place.

`enrichextract.go` was at 499 lines and the waiver put it over the 500 ceiling. The request builder moved to its own file rather than the file taking a ratchet entry.

## No aicert re-run is owed by the waived sites; the governed ones do owe one

Prompt text changed at the sites that gained a rule, so their certification records now describe a request the tree no longer builds. The cert cases pass English explicitly and pinned, so a score stays comparable between installations — but the records themselves need the paid lane re-run before they mean anything again.

---
## Summary by cubic
Enforces a single language rule across governed model prompts and gates it. Previously replies echoed the input language; now shared records answer in the installation’s base language, onboarding answers in the admin’s locale, and non‑prose prompts are explicitly exempt. Deal‑status caching now keys by language and does not cache fallback cards when a model lane fails.

- Introduces `backend/internal/compose/promptlang` and a static gate `backend/promptlanguage_test.go` that fails any `model.Request` without a rule or reasoned waiver; the gate also rejects requests built via var+field assignments to keep all builds visible.
- Centralizes base‑language resolution in `identity.BaseLanguageForPrompt`; callers log failures and fall back to English.
- Applies the rule to deal status cards, offer drafts, signal summaries, transcript proposals, site triage reasons, and agent run final summaries; onboarding prompts switch to `promptlang.Rule(locale)` to preserve identifiers.
- Runner passes a pre‑rendered `LanguageRule` in `runner.Job` and inserts it into the system prompt; certification paths pass an empty rule to keep the corpus fixed.
- Updates deal status: fingerprints include language; a failed lane’s floor is not cached, while absent lanes still are; covered by `dealstatus/lanefailure_test.go`.
- Defers adding the rule to `orgbrief`, `orgdossier`, and growth‑fit until their fingerprints include language; `/v1/site-reads/{id}/messages` lacks a locale and needs a contract change.

**Rollout**
- Re‑run certification for governed prompts (prompt text changed); waived sites do not need a re‑run.

<sup>Written for commit 10dbde977cf737c176541306c7ae7b974f0be089. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/2490?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * AI-generated summaries, transcripts, offer drafts, onboarding content, and site explanations now follow the installation’s configured base language.
  * Non-translatable content such as names, URLs, identifiers, email addresses, and quoted text is preserved.
  * English is used automatically when no supported language is configured.

* **Bug Fixes**
  * Deal status fallback results from failed responses are no longer cached, preventing stale fallback cards from being reused.


LarsGradion:
## Certification records: measured, not assumed

I checked rather than guessing, and the PR body's line about this was vaguer than it should have been.

`TestEveryCommittedRecordNamesTheCurrentPromptVersion` is **warn-only** (`promptversion_test.go:263` is a `t.Logf`), so `make check-backend` staying green proves nothing here. Running it verbosely and diffing against `main`:

| | stale records |
|---|---|
| `main` | 14 — `agent_loop`, `capture_counterparty_verdict`, `draft_reply`, `site_fact_extract` |
| this branch | 20 |

**This PR adds exactly two tasks**, and both are genuine consequences of a real prompt edit:

- `offer_draft` — gained a language rule.
- `cold_start` — via `cold_start/acts` and `cold_start/company_message`, the two onboarding prompts that previously carried their own `Respond in <locale>.` sentence.

`cold_start` was worth chasing down because it looked wrong at first: `companyFactsRequest` is a site this PR only *waived*, and a comment cannot move a digest. It turned out `cold_start` also covers the onboarding prompts, which did change. The file split of `enrichextract.go` moved that request builder without altering a byte of it, and it contributes nothing.

**No waived site went stale**, which is the check that matters — a waiver that silently changed a prompt would be the bug.

Filed as #2491. It needs the paid lane, not a code change.

LarsGradion:
## Review findings addressed

Four defects found, all fixed and mutation-checked. Two more filed rather than patched.

### The resolver existed twice

`dealstatus` cannot import `compose`, so the pool-based base-language read was written out twice, byte-identical. That is exactly the duplication `promptlang`'s own docblock warns about — written into the same change that warns about it.

It now lives beside the setting it reads, as `identity.BaseLanguageForPrompt`. Both callers already imported `identity`.

### A failed lane no longer freezes its fallback

A wired model lane that times out returns the deterministic floor, and that floor was saved under the current fingerprint. A stored card whose fingerprint matches is served **without asking again**, so one transient outage stood until an unrelated fact moved.

This predates the PR — and the PR made it reachable a new way. The language is part of the fingerprint now, so an installation switching to German mints a fresh key for every deal, and a lane that is down while it does would freeze the **English** floor as that installation's German card.

`write` now reports whether the lane *failed*, as distinct from being absent or held back by the call floor. Only the first declines to cache: a lane nobody wired answers no differently next time, so its floor is the real answer and belongs in the cache.

### The gate now sees a request built any way it can be built

`var req model.Request` followed by field assignments reaches a model exactly like a literal does, and was invisible to a census that looks for composite literals. Rather than trace assignments — a type-checker's job, and it would still miss a field set through a pointer — the pattern is refused outright. Nothing in the tree uses it.

### A degraded resolve leaves a trace

The resolver returns a string and nothing else, so no caller *could* notice a failed settings read and report it. My comment claiming callers would log it was wrong on its own terms. It logs now, and says why a missing row never reaches that line.

### Mutation checks

| mutant | result |
|---|---|
| a lane failure reported as success | **FAIL** |
| an absent lane reported as a failure | **FAIL** |
| a planted `var`-form request | **FAIL** |

### Not fixed here

- **#2491** — six certification records describe prompts the build no longer sends. Needs the paid lane.
- **#2493** — Go-composed evidence sentences stay English and now wrap base-language model text. ~34 strings and a design question about where server-composed prose gets its language.

### On the census's file-level reach

The review also notes that a `promptlang.` reference anywhere in a file blesses every request literal in it. That is real and it is stated in the gate's own docblock: it proves a rule was *attached*, never that the right language resolved. Following the value means resolving constants across packages. The behavioural half is covered by the integration test, which writes a settings row and reads the language back out. I would rather leave the imprecision documented than build a half-type-checker inside a fitness test.

Local gates green: `check-backend`, `craft-static`, `rls-store-path`, `no-jurisdiction`, plus the full `compose/integration` suite (159s, all passing).

PR #2508, opened 2026-08-24
feat(ai-models): one language for AI text, and the browser stops deciding it

### LINKED ISSUE #2497: A reader-language brief has two sources of truth: Accept-Language and app_user.locale
## The decision that is missing

Some AI writing is read by one person and should follow **their** language; the rest is filed on a record and follows the installation's base language. That split is settled and implemented — the dossier, growth fit, deal status card, offer lines, and extracted summaries all take base language (#2490, #2494).

What is not settled is where a **reader's** language comes from. Two mechanisms exist and can disagree:

1. `compose/meetingbrief/language.go` reads **`Accept-Language`** from the request.
2. `app_user.locale` holds the language a person **chose** in the app (shipped in #2429).

They differ for the same person on a borrowed laptop, on a colleague's machine, or in a browser whose locale nobody ever set. Neither is obviously right:

- The stored choice is what the person actually asked for, and follows them between devices.
- `Accept-Language` is what the browser in front of them is set to, and needs no round-trip.

## What is blocked on it

`orgbrief` — the per-company brief — is cached per reader and carries a `//promptlang:exempt DEFERRED` waiver naming this issue. It is the one prompt in the tree that should take a reader's language and currently takes none, so it answers in whatever language its input happened to be in.

Picking one source inside `orgbrief` would put a **third** answer in the tree, which is the duplication the `promptlang` package exists to prevent.

## Recommendation

**Stored locale first, `Accept-Language` as the fallback, base language as the floor.** An explicit choice should beat a browser default, and a person with no stored choice still gets something better than nothing. That ordering also makes the two existing mechanisms one: `meetingbrief` would consult the seat before the header rather than being replaced.

The counter-argument worth hearing: a stored locale is stale the moment someone changes it elsewhere, and a brief cached under it needs the same fingerprint treatment the base language just got.

## Scope when it is decided

- One resolver, in the place `identity.BaseLanguageForPrompt` already sits.
- `meetingbrief` adopts it instead of reading the header directly.
- `orgbrief` composes `promptlang.Rule` and adds the language to `Fingerprint` — **both halves or neither**, or a language change serves the old brief until an unrelated fact moves.

LarsGradion:
## Every prompt in the product is now governed. Zero deferrals remain.

The product had **two answers** to "what language is AI writing in":

- The installation's base language governed everything filed on a record.
- The **meeting brief** read the browser's `Accept-Language`.
- The **company brief** read nothing at all — it answered in whatever language its input happened to be in.

Both now take the installation's language, so there is one answer.

## Why the per-reader split was the wrong seam

A brief is cached per reader, and that is about **permissions**, not preference: it's assembled from records that reader may see, so two people with different access legitimately get different **facts**.

Language is not a permission. Treating the per-reader cache as evidence that language should follow the reader confused the two.

`compose/meetingbrief/language.go` is deleted with its `Accept-Language` parsing, and the test whose only subject was that parser goes with it.

## Two bugs the old mechanism was hiding

**1. The meeting brief has never followed anyone's language.** Its prompt said *"write in the language named by `language` in the summary"* — and `json.Marshal` emitted that field as `Language`, capital L. The field the prompt named was never present.

The test that should have caught this asserted only that the word `"language"` appeared somewhere in a prompt whose surrounding prose contains it. It now asserts the language the call was **given**, built for German so a builder that ignores its argument fails.

**2. `askSystemFor` took a language and never used it.** Caught by `unparam`, not by me. The Ask surface would have shipped carrying no rule at all while every signature said otherwise.

A third was avoided rather than fixed: `AskRequest`'s inner closure declared its own `lang`, shadowing the outer one. Renamed to `forLang`.

## The key carries the language

`orgbrief.Fingerprint` gains it as its own component, beside the derived prompt version — which is digested at a fixed English for the reason `dealstatus` and `orgdossier` already do it: the digest is a package-level `var` computed at init, where no installation's setting is readable, and what it must capture is the **wording**.

| test | what breaks without it |
|---|---|
| two languages hash differently | a language switch serves stale text |
| the same language hashes identically twice | nothing is ever a cache hit; every read rewrites |
| every shipped language keys distinctly | a fold that lowercased or truncated a code serves one language's brief as another's |

## Verification

`make check-backend`, `make craft-static`, `make rls-store-path`, `make no-jurisdiction` — green.

Integration lane against a real database:

```
--- PASS: TestTheBaseLanguageSetOnTheInstallationReachesThePrompt
--- PASS: TestAnInstallationThatNeverNamedALanguageStillGetsAPrompt
--- PASS: TestTheDossierAndGrowthFitPromptsCarryTheInstallationsLanguage
```

Mutation-checked: dropping the language from the key fails the first fingerprint test; hard-coding English in the builder fails the meeting-brief assertion.

## Behaviour change worth naming

An admin whose installation is set to German but whose browser is English currently sees English meeting briefs. After this they see German ones — consistent with every other AI surface, and consistent with the setting they chose.

Closes #2497.

---
## Summary by cubic
Sets one source of truth for generated text language: all briefs and answers now write in the installation’s base language. Previously, meeting briefs followed the browser’s `Accept-Language` and org briefs followed input language; now both follow the installation setting. Certification prompts are pinned to English. Cache keys include language, and the browser no longer influences language.

- Deletes `meetingbrief/language.go`; removes per-request language from `meetingbrief.Input`; services fetch language via `identity.BaseLanguageForPrompt`.
- Adds `lang` to `meetingbrief.BriefRequest`/`Write`, `orgbrief.BriefRequest`/`Write`, and `orgbrief.AskRequest`/`Answer`; prompts use `promptlang.Rule(lang)`.
- `orgbrief.Fingerprint` includes `lang`; prompt digest stays fixed English to capture wording only. Expect cache churn when the installation language changes.
- Certification paths (`certcase_orgbrief`, `certcase_orgask`) explicitly pass English.
- Frontend invalidates language-dependent caches on settings save using `LANGUAGE_DEPENDENT_QUERY_PREFIXES`; `aicaches.test.ts` enforces list completeness. Readers see the new language immediately.

**Migration**
- Update all callers to pass the installation language via `identity.BaseLanguageForPrompt(ctx, pool)` to `BriefRequest`, `Write`, `AskRequest`, and `Answer`.
- Remove any reliance on `Accept-Language`; the browser no longer affects language.

<sup>Written for commit ba7c7a8421c808c075ad053c4e1dfefc897275dc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/2508?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Meeting briefs and organization briefs now use the installation’s configured language.
  * Organization questions use the configured language, while certification requests remain fixed to English.
  * Language-specific brief caching prevents content from being reused across languages.

* **Bug Fixes**
  * Reader language preferences no longer override the installation language during brief generation.
  * Language-dependent content refreshes automatically after installation language settings are changed.


LarsGradion:
## Review finding addressed — the browser was the missing half

Codex found the backend sound and the **frontend** wrong, which is the half I had not looked at.

The server rewrites every model-written surface the moment the base language changes; its cache keys carry the language now. **The browser never asks.** Its own keys are the record id and nothing else, and `refetchOnWindowFocus` is off product-wide — so an admin changes the language and the company page they are looking at keeps rendering the old one. The setting appears to do nothing, which is the exact complaint base language exists to answer.

Saving installation settings now invalidates the seven surfaces a model writes: account brief, dossier, growth fit, meeting brief, person brief, deal status card, home brief.

Unconditionally, rather than only when `base_language` moved. Deciding that means diffing against the pre-patch value, and the costs are asymmetric: an extra refetch of one open page is cheap, a missed one serves a reader the wrong language.

### The list is derived, not maintained

A list of caches to invalidate is a thing to forget to add to — and forgetting is the exact failure it prevents. `aicaches.test.ts` walks the screens, matches every query key that reads like a model-written surface, and fails when one is not named.

It checks the other direction too: a prefix nobody queries invalidates nothing and reads as coverage that is not there, which is worse than an obviously short list because it *looks* complete.

| mutant | result |
|---|---|
| drop `personBrief` from the list | **FAIL** |
| add a prefix no screen queries | **FAIL** |

### Not fixed here

A **second tab** changing the setting still leaves this one stale, because window-focus refetching is off product-wide (FE-PARAM-3, a deliberate default). Same-tab is the path an admin actually takes, and changing that default is a product decision about every query in the app, not this feature's to make.

### Also on this branch

Rebased onto #2510, which fixed a red `main` I caused in #2490 — the offer-draft test fixture built `offerDrafter` field by field and never got the pool that change added.

Gates re-run after the rebase: `check-backend`, `check-fe`, `craft-static`, `rls-store-path`, `no-jurisdiction` — all green.

PR #2549, opened 2026-08-24
feat(platform): one dev stack per worktree, claimed not hashed

dungbanguyen:
## Purpose

One dev stack per worktree, so several agent sessions can run at once — and
`make dev` stops destroying the other ones.

Isolation is by NAME, not by container: api/worker/vite stay host processes and
Postgres/Redis/MinIO stay one shared compose stack. A slug derived from the
worktree selects the database, the Redis logical database, the port pair, the
object bucket and the test template.

## Five defects, all of them silent

| Defect | Where it was |
|---|---|
| The registry handing out Redis logical databases lived under the **worktree's own** `.tmp/dev/`, so a second worktree read an empty registry and claimed database 64 again. Two stacks then shared one consumer group: whichever worker read an entry first resolved it against its own database, found nothing, and acked it. | `scripts/dev.sh` |
| Ports were **hashed** (`8080 + cksum(slug) % 1000`), not claimed. Two slugs whose hashes differed by a multiple of the block size took different ports and the SAME Redis database — a collision no port check can see. Redis got a claim registry; ports never did. | `scripts/dev.sh` |
| `MARGINCE_BLOBSTORE_BUCKET=margince-dev` was hardcoded at both call sites, so every stack wrote attachment bytes into one bucket. | `scripts/dev.sh` |
| A bare `make dev` **swept the machine** on the way up — every margince process, every recorded stack, every `margince_dev_*` database. The routine command was also the destructive one. | `scripts/dev.sh` |
| The integration lane shared one `margince_test` template per machine. `build_template` recreates it from scratch, so a parallel session on another branch rebuilt it underneath you and your lane failed wholesale with schema-shaped errors that read as your own migration being broken. | `scripts/lib-testdb.sh` |

The first one is the expensive one. `scripts/dev.sh` promised "a running stack's
db is never handed out twice"; across worktrees it did exactly that, and the
comment beside it records that the resulting silent event loss once cost a day
and a wrongly-filed critical bug.

## What changed

- **The registry is machine-global** — `$XDG_STATE_HOME/margince/dev/` — which is
  what makes the existing claim-under-lock logic mean anything across worktrees.
- **The slug comes from the worktree.** A linked worktree derives it from its own
  directory name; the primary worktree keeps `:8080` and the shared `margince`
  database, because `make migrate`, `make seed-dev` and `make verify-boot` target
  that database by name. `DEV_SLUG=` still overrides.
- **Ports are claimed, not hashed**, from 8081–8179 (api at +10000), under the
  same lock and recorded before anything binds — a claim only visible once the
  stack is up is not a claim. A port with a foreign listener is skipped, which a
  hash cannot do.
- **Per-stack object bucket.** `blobstore.New` creates a missing bucket, so no
  infra change.
- **`make dev` touches nobody else's stack**; `make dev-stop` stops this
  worktree's; the machine-wide clear is now `make dev-sweep`, asked for by name.
- **The test template is per worktree.** Named per worktree rather than per
  migration-set hash on purpose: `migrate_template` migrates an EXISTING template
  up incrementally, and content-addressing would rebuild the whole schema on every
  migration edit. The primary worktree and CI keep `margince_test` unchanged.

## Evidence

**Three stacks at once, on this machine, with another session's stack among them.**
`:8080` was a stack from a different session — the kind the old code killed. It
was still answering afterwards:

```
dev-isolation    FE_PORT=8081 REDIS_DB=64
probe-second     FE_PORT=8082 REDIS_DB=65

:8080 /readyz -> 200      (another session — survived my `make dev`)
:8081 /readyz -> 200
:8082 /readyz -> 200
```

Before this change both worktrees would have taken Redis database 64.
`make dev-stop` in one worktree then stopped only that stack and dropped only its
database, leaving the other two running.

**`make test-integration`** — `OK: integration passed with 0 skips (43 packages,
parallel)`, against `margince_test_dev_isolation`.

**`make check-backend`** — exit 0.

**Every new assertion was watched failing.** Five mutations, each red, each green
on restore: a relative state root, an absolute-but-inside-the-checkout state root,
dropping the same-slug reclaim, dropping the foreign-listener check, and
`pick_free_db` always answering the block floor.

Two of those mutations exist because the gate was blind first:

- The registry assertions all point the root at a temp directory through the
  override, so **none of them could see the default move back under the
  worktree** — which is the entire defect. The first version of the added check
  tested only for a path under the checkout, which calls a relative `.tmp/dev`
  clean: that is the spelling the old code actually used, so the assertion passed
  over the one case it existed to catch.
- `rundir` was **left pointing at `.tmp/dev/` in the first draft**, so the
  reservation carried ports and never the pids and the sweep read a directory
  only its own worktree could see. Both halves looked right in isolation and the
  whole was still worktree-blind. The lifted-function checks structurally cannot
  see that — they exercise pure functions and the run directory is assigned at
  the top level — so a gate that greps for a second spelling of the state home
  stands in for what they cannot reach. It was caught by running two stacks, not
  by the unit checks.

## Also fixed, same invariant

`scripts/test-integration-parallel.sh` printed `TEMPLATE_DB` — the database
segment of `MARGINCE_TEST_DSN` — while cloning from `TEMPLATE_NAME`. Identical
strings until the template became per-worktree, and then the banner reported one
template while the run used another. `TEMPLATE_DB` had no other caller, so it is
gone rather than kept in step.

## Known gap, deliberate

Nothing reaps a stale `margince_test_<slug>` template after its worktree is
deleted. `make dev-sweep DROP=1` drops `margince_dev_*` but not `margince_test_*`,
and widening it into the test namespace risks deleting the template a parallel
lane is actively cloning from. Left out rather than half-built.

## Migration note

A stack recorded under the old `<worktree>/.tmp/dev/` is invisible to the new
registry. Its processes are still caught by `sweep_stacks`'s pgrep arms, which
match on the binary name and a margince connection string rather than on a state
file, so `make dev-sweep` still stops it. Run it once after taking this change.

## Summary by CodeRabbit

- **New Features**
  - Development environments now use isolated ports, databases, Redis resources, and object storage for each worktree.
  - Added `make dev-sweep` to clean up all local development stacks.
  - Added `make dev-stop` to stop only the current worktree’s stack.
  - Added optional custom development slugs and worktree-specific service URLs, logs, and test databases.
  - Development startup now preserves other active worktree sessions.

- **Documentation**
  - Updated development workflow and command reference documentation.

- **Tests**
  - Added coverage for isolation, allocation, conflicts, and capacity limits.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -92,14 +92,14 @@ Commands and flags: [docs/reference/make-targets.md](docs/reference/make-targets
 Config and endpoints: [docs/reference/configuration.md](docs/reference/configuration.md).
 CI: [infra/ci-pipeline.md](infra/ci-pipeline.md).
 
-## Exactly one dev stack at a time
+## One dev stack per worktree
 
-`make dev` sweeps first: it kills every margince api, worker and Vite on the
-machine, evicts `:8080`, and force-drops stray `margince_dev_*` databases before
-booting one stack. So you never stop the old one by hand — but it is **not**
-harmless: it reaches other worktrees and other sessions, and drops a `DEV_SLUG`
-stack's data with it. `make dev-stop` is the mirror; `DEV_SLUG=x` gives an
-isolated stack, spared until the next bare `make dev`.
+`make dev` starts a stack for THIS worktree and touches nobody else's: a linked
+worktree claims its own database, Redis logical database, port pair and object
+bucket, with no flag to remember. The primary worktree keeps `:8080` and the
+shared `margince` database, which is what `make migrate` and `make seed-dev`
+target. `make dev-stop` stops this worktree's; `make dev-sweep` clears every
+stack on the machine and is the only thing that does.
 
 **The API does not hot-reload.** Vite does, so the frontend is live as you type;
 the API is a compiled binary and every backend change needs `make dev` again. A
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ SEED_DSN ?= postgres://margince_owner:dev@localhost:15432/margince
 # every company renders as a placeholder initial.
 MINIO_PORT ?= 29000
 
-.PHONY: help install ai-routing-local dev-fresh check check-backend check-q check-go check-gates check-fe build test test-v test-cover test-integration e2e-siteread e2e-ai e2e-ai-report ai-probe test-db-up test-it test-integration-serial bench-perf bench-perf-check bench-record bench-capture perfdoc lint arch-lint vet gen gen-workflow mcp-apps-vocab gen-types gen-types-check drift composition check-composition test-extensions db-up db-init db-wait migrate migrate-up migrate-down migrate-create run psql redis-cli tidy dev dev-stop dev-logs clean vuln tools tools-go infra-up infra-down infra-logs infra-reset seed-dev seed-dev-db seed-demo verify-demo seed-reset verify-boot frontend-check frontend-e2e bench-mobile perfdoc e2e-company fe-install fe-typecheck fe-typecheck-composed fe-lint fe-build fe-preview fe-format fe-test fe-test-ext fe-ds-gates fe-drift fe-unit fe-clock-drift fe-quality fe-bundle fe-storybook ds-purity font-lock icon-lint ds-spacing space-tokens native-controls ext-imports fitness-jurisdiction storybook fe-uat craft-static craft-test craft-residue check-craft-doc test-golangci-guard test-scheduled-report test-ci-verdict test-laneorder secret-scan test-secret-scan test-dev-dsn test-dev-isolation test-api-entrypoint check-image-pins check-host-ports ci-doc-parity make-target-parity check-ext-migrations contract-breaking-check contract-frontend-drift test-contract-frontend-drift migration-versions test-migration-versions test-lanes env-reads gofmt lint-modules go-file-length rls-store-path no-jurisdiction one-spelling test-one-spelling money-scale test-money-scale test-selfdir pkg-freeze hooks sbom sbom-normalize sbom-supplement sbom-parity sbom-validate sbom-sign sbom-check
+.PHONY: help install ai-routing-local dev-fresh check check-backend check-q check-go check-gates check-fe build test test-v test-cover test-integration e2e-siteread e2e-ai e2e-ai-report ai-probe test-db-up test-it test-integration-serial bench-perf bench-perf-check bench-record bench-capture perfdoc lint arch-lint vet gen gen-workflow mcp-apps-vocab gen-types gen-types-check drift composition check-composition test-extensions db-up db-init db-wait migrate migrate-up migrate-down migrate-create run psql redis-cli tidy dev dev-stop dev-sweep dev-logs clean vuln tools tools-go infra-up infra-down infra-logs infra-reset seed-dev seed-dev-db seed-demo verify-demo seed-reset verify-boot frontend-check frontend-e2e bench-mobile perfdoc e2e-company fe-install fe-typecheck fe-typecheck-composed fe-lint fe-build fe-preview fe-format fe-test fe-test-ext fe-ds-gates fe-drift fe-unit fe-clock-drift fe-quality fe-bundle fe-storybook ds-purity font-lock icon-lint ds-spacing space-tokens native-controls ext-imports fitness-jurisdiction storybook fe-uat craft-static craft-test craft-residue check-craft-doc test-golangci-guard test-scheduled-report test-ci-verdict test-laneorder secret-scan test-secret-scan test-dev-dsn test-dev-isolation test-api-entrypoint check-image-pins check-host-ports ci-doc-parity make-target-parity check-ext-migrations contract-breaking-check contract-frontend-drift test-contract-frontend-drift migration-versions test-migration-versions test-lanes env-reads gofmt lint-modules go-file-length rls-store-path no-jurisdiction one-spelling test-one-spelling money-scale test-money-scale test-selfdir pkg-freeze hooks sbom sbom-normalize sbom-supplement sbom-parity sbom-validate sbom-sign sbom-check
 
 # Bare `make` lists every command instead of running the first target.
 .DEFAULT_GOAL := help
@@ -147,6 +147,14 @@ dev-fresh:
 ## just that one. DROP=1 also drops the per-slug databases (never `margince`).
 dev-stop:
 	@bash scripts/dev.sh stop "$(DEV_SLUG)" $(if $(filter 1,$(DROP)),--drop,)
+
+## dev-sweep — clear EVERY margince dev stack on this machine: kill every
+## api/worker/vite (recorded, orphaned, or belonging to another worktree) and
+## forget their claims. `DROP=1` also drops every per-slug margince_dev_*
+## database. This is the old bare-`make dev` behaviour, now explicit: `make dev`
+## starts your worktree's stack and leaves everyone else's alone.
+dev-sweep:
+	@bash scripts/dev.sh sweep "" $(if $(filter 1,$(DROP)),--drop,)
 
 ## dev-logs — follow the dev stack's log, coloured per process (api/worker/fe)
 ## and per severity, with the job-queue heartbeat hidden. ROLE=<api|worker|fe|boot>
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ SEED_DSN ?= postgres://margince_owner:dev@localhost:15432/margince
 # every company renders as a placeholder initial.
 MINIO_PORT ?= 29000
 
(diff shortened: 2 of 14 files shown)


dungbanguyen:
Thanks — nine of these were real, and two were bugs the first round shipped. All
addressed in 22d4715f0 except one, which I have left with a reason.

**Fixed:**

| Finding | Fix |
|---|---|
| Startup continued after a refused claim | `claimed="$(claim_stack …)" \|\| exit 1` — the here-string reported `read`'s status, so an exhausted block printed FAIL and the api came up on `:10000`. My own T2 violation. |
| `bucket_for_slug` not injective (`a_b` / `a-b`) | Underscore is out of the slug alphabet, so the mapping is a prefix and nothing else. Asserted as the *property*, not the spelling. |
| `DEV_SLUG=_base` collides with the primary stack | Refused, with the reason in the message. |
| `make dev-logs` broken by the state move | `scripts/lib-devstate.sh` now owns slug, state root, bucket, database name and app URL; every script sources it. |
| `make seed-dev` / `verify-boot` target `:8080` and `margince` | Both halves resolve this worktree's stack. Found by *running* it: the API half seeded the worktree's database and the SQL half the primary's, surfacing as a NOT NULL violation on `app_user.workspace_id`. |
| Template name over 63 bytes / basename collisions | Bounded to 49, disambiguated by a digest of the full path. Both cases are now planted tests. |
| Reservation leaked on failed startup | Released on any exit before the state write. |
| `dev-sweep` missed a sibling worktree's orphaned vite | Matches the git **common** directory — every worktree resolves vite from its own `node_modules`, so the old filter only ever found our own. |
| Docs named `<slug>` for the primary worktree | It is `_base`, and the docs say so. |

**The gate was too narrow, which is the more useful finding.** The
"state home is spelled once" check greppped `dev.sh` alone, so it could not see
`dev-logs.sh` carrying its own spelling — exactly the shape it was written to
catch. It now scans every script in the tree, derived rather than listed.

**On the test-bypasses-production point:** correct, and it is why the
`dev-logs.sh` bug got through. `claim_stack` is now exercised end to end against a
temp registry — first claim, second claim, same-slug reclaim, reservation-on-disk,
and refusal propagating as a non-zero status — plus `_testdb_worktree_slug` in real
throwaway worktrees.

**Not fixed, deliberately:** `make dev` racing `make dev-sweep` can still have the
sweep delete a claim mid-startup. Closing it means holding the claim lock across
the whole sweep, which makes the machine-wide clear block on every startup. The
window is small, `dev-sweep` is an explicitly destructive command somebody typed,
and the failure mode is a stack that refuses to start — not two stacks sharing
state, which is the class this branch exists to remove.

dungbanguyen:
Seven more addressed in c3dcaeb82, and the important ones are **regressions the
last commit introduced** — worth naming as such.

| Finding | What was wrong |
|---|---|
| Trap deleted the claim but left processes running | It now kills what this run started. It also keeps the **log** and releases only the claim: removing the directory took the log with it, so a failed boot deleted the one file explaining the failure. That cost me a debugging cycle before it could cost anyone else one. |
| `stack_recorded=1` above the FE readiness probe | Moved inside the success branch. An FE that never came ready was leaving a claim holding a port and a Redis database. |
| Re-run on a live stack destroyed its record | A pre-existing record is returned as-is and never released. The flag saying so is computed in the **parent** shell — `claim_stack` runs in a command substitution, so a variable it sets never leaves that subshell. Same trap as the swallowed claim status one line above, and my first version of this fix walked straight into it. |
| Empty slug meant "primary worktree" | A pure-punctuation directory name sanitised to nothing and put a linked worktree on the shared `margince` database and `:8080`. Path digest instead, with the sanitiser's own behaviour asserted first so the planted case is provably the real one. |
| Same basename → same slug | Disambiguated by a path digest, but only when a sibling actually collides, so the common case stays a name a human recognises. |
| `dev-sweep` missed a sibling's vite | Deriving a parent from the git common dir was no better than matching this checkout — a linked worktree can live anywhere. It enumerates `git worktree list`. |
| The seed's two halves could still split | With no recorded stack a linked worktree now **refuses** rather than sending the API half to `:8080` while the SQL half goes to `margince_dev_<slug>`. The primary worktree still answers `:8080`, because that stack's port is fixed rather than guessed. |

One finding I had to fix twice over: pointing `DB_NAME` at the per-stack database
broke `make dev` outright, because it also feeds `db-init` and `db-wait`, which run
before dev.sh has created that database. The seed targets get their own
`DEV_DB_NAME`, and it has no silent fallback to the shared database — a failed
resolution means an invalid slug or an unanswerable worktree identity, and
defaulting to `margince` there would point a linked worktree's psql at the
primary's data.

**Verified live, not just by tests:** boot; a re-run refusing while the live stack
keeps its record *and its pids*; `dev-stop` finding and dropping it; and another
session's `:8080` stack untouched throughout. The unit checks could not have caught
the trap or the subshell bug — both are top-level wiring — which is why the live
run is in the loop.

dungbanguyen:
All eleven addressed in 1867f6e78. Two were design flaws in earlier fixes on this
branch, not nitpicks — worth naming as such.

**The slug could be renamed out from under a running stack.** The digest was
appended only when a same-named twin *currently* existed, so creating a second
`feature` worktree renamed a running stack from `feature` to `feature-<digest>`,
after which `dev-stop` resolved a slug with no record and could not stop it. The
digest is always appended now: noisier than it could be, and reliably its own.
That change made the length finding bite universally, so the name half is bounded
to 24 chars — `margince_dev_<slug>` fits Postgres's 63-byte identifier limit and
`margince-dev-<slug>` fits S3's. Both are planted assertions.

**The EXIT trap named a function defined 200 lines below it.** Any exit between the
two printed `command not found` and left the claim behind. That is the same
ordering mistake as `port_listeners` earlier on this branch — twice now, so the
helpers moved above the trap rather than being patched in place.

| Also fixed | |
|---|---|
| Crashed stack's pids believed | The "already running" check now requires the process to be **alive**. One crash was otherwise disabling cleanup for every later failed boot. |
| Two concurrent `make dev` for one slug | The reservation records `STARTER_PID`; the release only removes its own, so the run that loses the port race cannot delete the winner's claim. |
| Release sent only TERM | Goes through `kill_pids` (TERM then KILL) — a bare TERM could leave a server standing after its claim was gone. |
| `dev-sweep` bare-substring vite match | Matches on a path boundary, so a project at `<worktree>-scratch` is not swept as ours. |
| `make psql` on the shared database | Follows `DEV_DB_NAME`, like the seeds. |
| Apostrophe in the checkout path | The library path is a positional argument, not interpolated into a single-quoted script. |
| Primary-worktree probe ignored an inherited `DEV_SLUG` | Cleared for the child shell, so it tests the path it claims to. |

Verified live on the new slug: boot, both halves of a seed resolving the same
stack, `dev-stop` finding and dropping it, and `make check-backend` green (282
legs).

dungbanguyen:
Addressed in 020f1664d.

**Moving a checkout renamed a running stack.** The slug came from the worktree's
PATH, so `git worktree move` changed it, after which `dev-stop` resolved a slug with
no record and `make dev` started a second stack beside the first. It now derives
from the worktree's **admin name** under `<common>/worktrees/`, which git carries
across a move and guarantees unique among one repository's worktrees; the digest
identifies the repository rather than the checkout, so two clones on one machine
still cannot collide on a shared name like `feature`. Move-stability is a planted
test.

That is twice this slug has been made stable against something — first a sibling
appearing, now the worktree itself moving. Same shape both times: an identity
derived from something that changes.

**A live starter could be overwritten.** A second `make dev` for the same slug
rewrote `STARTER_PID`, so whichever wrote last owned the reservation — and the one
that then failed its port check released the claim the other was booting against. A
reservation held by a live starter is refused now, naming the pid.

**The apostrophe fix did not work.** Passing the library path positionally still
embedded it in the `$(shell …)` string, which make hands to `/bin/sh` — the quoting
breaks before bash is reached. It goes through a one-line script invoked by a
relative path, which has nothing to interpolate. Good catch; my first fix addressed
the inner quoting and left the outer.

Two in the test suite: `cd … && f` compared an empty string when the directory was
missing instead of failing, and the "recognisable prefix" assertion only checked
length — which `wt-12345678` passes while telling a reader nothing. It asserts the
name is still on the front.

`make check-backend` green; verified live on the new slug (boot, both halves of a
seed agreeing, `dev-stop` finding and dropping it).

dungbanguyen:
All three in the latest push, and the first is a regression the move-stability fix
introduced — worth naming as such.

**Two long sibling worktrees folded onto one slug.** Making the slug survive
`git worktree move` moved the digest from the worktree's path to the **clone's**
common directory, which every worktree of that clone shares. With the readable half
truncated to 24 characters so the database and bucket names fit, two siblings
agreeing that far derived the same slug — one database, one Redis logical database
and one bucket between two stacks, i.e. exactly what this PR exists to remove,
reintroduced while fixing something else. The digest now covers the common directory
**and** the worktree's admin name, so truncating the readable half cannot cost
uniqueness. Planted in both halves: that the two names really do share 24 characters
(so the case is real), and that their slugs still differ.

**A recycled pid could wedge a slug permanently.** The "another `make dev` is already
starting this" check used a bare `kill -0`, so once a stale claim's starter pid was
reassigned, every later `make dev` for that slug was refused with no way to tell why.
It goes through `still_ours`, which reads the command line.

**A make-level `DEV_SLUG` override never reached the resolver** — confirmed by
measuring rather than reasoning: `make seed-dev DEV_SLUG=alpha` resolved the
worktree's database, not alpha's. Exporting the variable was not sufficient either;
it is passed as an argument now, and both paths are verified.

`make check-backend` green, and a live boot/stop cycle on the new slug.

dungbanguyen:
All three addressed, and the first made the guard added two commits ago inert.

**`still_ours` rejected the very process it was looking for.** It identifies a
running *server*, by the margince connection string on its command line; a starter
is a shell running `dev.sh`, which carries neither. So "another `make dev` is
already starting this stack" never fired, and a second run overwrote the first's
reservation. It looks for a live `dev.sh` now. (A bare `kill -0` was the version
before that, and is the other wrong answer: pids get recycled.)

**The resolver took `$(DEV_SLUG)` unquoted** — make's shell parsed it before
anything validated it. Verified inert after the fix: `DEV_SLUG='x; touch /tmp/PWNED'`
executes nothing and yields an empty name.

**And it collapsed an omitted argument into an empty one.** Make passes no argument
when `DEV_SLUG` is unset and an empty one when it is set empty; those mean "derive
it" and "the caller named nothing", and merging them let an exported slug from some
parent shell pick a database the command did not name.

PR #2555, opened 2026-08-24
docs: cut the three off-audience pages, and put a measured budget on page length

dungbanguyen:
## Purpose

Two changes to `docs/`, from the audience-cut brief: remove the pages that are not
for the person changing this code, and put a measured budget on how long a page
may be.

## The audience cut is much smaller than it looked

The brief was "keep only what is needed to CHANGE the code". Reading all 28
`explanation/` pages rather than inferring from the directory name: **27 of them
document code internals** — `write-backbone`, `authorization`, `job-fleet`,
`composition-layer`, `contract-first`, `frontend-architecture`,
`rbac-roles-and-teams` — and `backend-onboarding.md` is the orientation hub
`getting-started.md` sends every new engineer to. Exactly one is product
narrative.

Of three tutorials, `getting-started.md` is engineer onboarding (clone → running
instance in five commands) and **stays**; the other two are end-user walkthroughs
that say so in their own words: *"No code, no API. You need a sign-in, a company
in Margince…"*.

So three pages go, not thirty:

| Page | Lines |
|---|---|
| `explanation/margince-conversational-workspace-concept.md` | 733 |
| `tutorials/run-your-first-project.md` | 385 |
| `tutorials/run-a-partner-program.md` | 188 |

Six inbound references relinked to the how-to carrying the same material.

## The rest of the size is length, not audience

So it gets a budget: `backend/docspagelength_test.go`, with the numbers
**measured, not chosen**. Each ceiling is roughly 1.5× its section's 75th
percentile today, which makes the gate an outlier detector rather than a demand
that every page be short.

| Section | Budget | Why |
|---|---|---|
| `how-to/`, `tutorials/` | 350 | a procedure followed with one hand on the keyboard |
| `explanation/` | 450 | carries a mechanism |
| `reference/`, `evidence/`, root | 400 | a table, bounded by what it tabulates |
| `principles/` | 250 | |

**The first draft used 150 and 250, picked before measuring, and put 45 pages over
budget.** That is a waiver file nobody would ever work through — rotted on
arrival, and read by the next author as a list of pages *allowed* to be long. At
measured numbers it is 10, six of them within a quarter of their budget. That is a
ratchet somebody can actually finish.

**No page is compressed here, on purpose.** With ceilings derived from what pages
in this tree actually are, achievability is shown by the 66 already inside budget
rather than by one demonstrative rewrite — and the obvious candidate
(`flip-an-overlay-to-native.md`) is an overlay cutover procedure I would have been
editing without the context to do it safely.

## The gate derives both halves of its subject

- **Corpus**: a walk of `docs/`, so a new page is covered the moment it lands. It
  fails if it finds no page at all, rather than passing by having nothing to check.
- **Exemption**: it reads the `do not edit by hand` marker a generated page carries
  about *itself*. A list of which pages are generated would go quietly short the
  day another generator landed — the failure *Reuse before you build* rule 5 names.
- **Ratchet teeth**: a waived page that has come under budget must leave the
  waiver, and a waiver naming a page that no longer exists fails. Without both, the
  file rots into "pages allowed to be long".

**Four mutations watched failing**, each green on restore: a new over-budget page,
a waived page brought under budget and left listed, a stale waiver line, and the
generated exemption removed (which correctly surfaces `mcp-info.md` at 11,010
lines).

## Found while indexing

`docs/README.md` **never listed `mcp-info.md` at all** — 11,011 lines, the largest
file under `docs/` by an order of magnitude, unreachable from the index. Listed
now, with its `agent-tool-budget` sibling, and the rule that a generated page is
neither hand-edited nor shortened stated once.

## Verification

- `make check-backend` — 55 gate legs OK. `lint-modules` could not complete
  locally: parallel sessions on this machine keep the shared golangci cache
  answering with foreign worktree paths, so the lint ran with an isolated cache
  instead — **0 issues**. CI's cache is clean, so the verdict here is the one that
  counts.
- `backend` package green.

## Merge order

This touches `docs/README.md` and `STATUS.md`, which #2482 also changes, so expect
a small conflict with whichever of the two lands second. #2482 deletes `STATUS.md`
outright; the one-line edit here exists only so the reference does not dangle if
this lands first.

## Summary by CodeRabbit

- **Documentation**
  - Added a user-guide entry point with walkthroughs for first projects and partner programs.
  - Improved navigation between user guides, setup instructions, project workflows, partner-program references, and MCP resources.
  - Clarified the distinction between user-facing and contributor documentation.
  - Updated outdated links and redirected the Projects documentation link to the user-guide walkthrough.
  - Removed the Margince conversational workspace concept from the published documentation index.


dungbanguyen:
All seven addressed in 77c6ecb48. Two were defects in the gate itself and the Major
one is the sharpest kind.

**"CLOSED to new entries" was a claim nothing held.** Adding an over-budget page and
its waiver line in the same change passed both directions — exactly as reported.
That is a comment claiming what no test holds, which is the defect *Reuse before you
build* rule 4 names, and a gate is a bad place to commit it. The list's **size** is
pinned now (`waivedPageBudget`), so growth has to be written into the test where a
reviewer sees it and can ask why; the prose says "only shrinks" because something
checks it. It also fails when the pin drifts more than two entries above the list, so
the ratchet cannot quietly stop noticing. Watched failing on precisely the bypass
described.

**A page could exempt itself by mentioning the rule.** `isGeneratedDoc` matched the
phrase anywhere in the opening lines, so a hand-written page discussing generated
pages was silently excused from the budget. The marker must now sit inside an HTML
comment — which is how every generator in this tree writes it — and
`phrase-in-prose.md` is planted as a case.

**The relinking was lazy, and all three reports are right.** Replacing the deleted
tutorial's link with "this page" left three guides telling a new reader to start with
the page already open, while promising an ERP-rollout walkthrough that page does not
contain. The prompts are gone; the real cross-link to the sibling guide stays. The
missing blank line before the generated-pages note is fixed too — it was rendering
as part of the preceding bullet.

dungbanguyen:
Both addressed in 37cad84fd.

**The marker check was too loose in a second way.** Requiring only the `<!--`
opener let a line that merely starts one and mentions the phrase exempt a
hand-written page; the check now requires a **complete** HTML comment with the
marker inside it, and both refused spellings (`<!-- do not edit by hand` with no
terminator, and `<!-- note --> do not edit by hand`) are planted.

**The pin cannot see a swap, and you are right that it reads as though it can.**
Bring one waived page under budget, delete its line, add a new over-budget page —
count unchanged, every check green. Detecting that needs a baseline of the list,
and the list *is* the baseline; the only place the addition shows is the diff,
which is where a reviewer reads it. That limit is now stated in the gate rather
than left for the pin to be mistaken for a complete guard.

dungbanguyen:
Both are the same point and both are right: I drew the boundary as "docs/ is for
engineers" and then sent end users into `docs/how-to/`.

The boundary is not actually there, so both pages now say what is. `docs/how-to/`
carries product how-tos beside engineering ones, and that is correct — a how-to is
what you come back to for one answer, as true of a field reference for a rep as of
a recipe for an engineer. What `user-guide/` holds is the other shape: a
walkthrough followed once, end to end. The split is by document TYPE, not strictly
by tree, and asserting otherwise in two places was the defect.

PR #2582, opened 2026-08-24
fix(contract-api): a confidence this contract cannot carry is not served

tinnhlt-gra:
## The defect

`PersonProviderPhone.confidence` is declared in the contract as:

```yaml
confidence: { type: [number, 'null'], minimum: 0, maximum: 1 }
```

The value that reaches it comes from a provider claim's **`value_json`**, which carries no CHECK. The `confidence` **column** beside it *is* bounded 0–1 — and is not where this number comes from.

So the whole path was unguarded:

```
Surfe confidenceScore  →  value_json (no CHECK)  →  person360 foldPhones  →  served
```

**A vendor that scores out of 100 — a normal thing for a vendor to do — puts `87` on a field the contract says is at most `1`.** The page renders a confidence as a percentage, so a reader would be shown 8700%.

Nothing in the tree validated it, and nothing failed: the bounded column made the path *look* guarded while carrying none of the traffic.

## Why the guard is at the fold

`foldPhones` is the **one place a bounded contract field is written**, and every claim source reaches it — Surfe today, the offline fixture, whatever is added next. A check at each provider would be one check per provider, and the next provider would not have it.

## An unstatable score costs the confidence, never the number

The phone number is what the row was bought for, and it is unaffected by how a vendor scales its certainty. The contract makes `confidence` optional precisely so it can be absent.

**Saying nothing about how sure we are beats saying something the contract cannot express.**

Zero is carried rather than dropped — it is a real confidence, and its absence means something different.

## Verification

| case | served |
|---|---|
| `0.87` | `0.87` |
| `0` | `0` — a real confidence, not an absent one |
| `1` | `1` |
| **`87`** (vendor scoring out of 100) | **nothing** |
| `1.0000001` | nothing |
| `-0.5` | nothing |
| absent | nothing |

Plus: an out-of-range score never takes the number with it.

Mutation-proven — removing the guard fails on exactly the case that matters:

```
served confidence 87 for a score of 87 — the contract declares this field 0..1,
and a page renders it as a percentage
```

`make check-backend` green.

## Scope note

The audit item that led here also listed `enrichextract.go`'s `(0,1]` against six `[0,1]`, and three float widths. Those are a separate question — whether zero is a legal confidence — and are not touched. This PR is the one where an unbounded value reaches a bounded contract field, which is the half with a reader on the other end.

Branched from `50f57116` rather than current main, which is red on an unrelated migration (#2576). Will rebase before merge.

---
## Summary by cubic
Prevents out‑of‑range confidences from being served in `PersonProviderPhone.confidence`. Previously unbounded values from `value_json` could set 87 and render 8700%; now the fold serves confidences only in [0,1] and omits invalid ones while always keeping the phone number.

- Adds `statableConfidence` in `backend/internal/compose/person360/providerclaimfold.go` and applies it in `foldPhones`; sets `Confidence` only when the score is within [0,1] (NaN, +Inf, and −Inf are rejected by the appropriate bound). Zero remains a valid confidence.
- Adds `backend/internal/compose/person360/providerclaimconfidence_test.go` to cover valid (0, 0.87, 1), invalid (>1, <0), absence, and the column fallback path; asserts the phone value is never dropped.
- Clarifies comments: this is the one place a provider claim’s confidence reaches the contract; explains why each unstatable value is excluded; documents the difference from `compose/enrichextract.go`; and states why the guard runs at read time.

<sup>Written for commit dacd1b08b6018c7b4d99aa2a3ac051ec68d23699. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/2582?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved mobile phone data handling by accepting only valid confidence scores between 0 and 1.
  * Preserved phone values when confidence information is missing or invalid.
  * Added fallback handling for claim-level confidence scores.

### DIFF
--- a/backend/internal/compose/person360/providerclaimconfidence_test.go
+++ b/backend/internal/compose/person360/providerclaimconfidence_test.go
@@ -5,12 +5,12 @@ package person360
 
 // A provider's confidence reaches this contract through `value_json`, which
 // carries no CHECK. The column beside it is bounded 0..1 and is NOT where this
-// number comes from — so between a vendor and a reader there was nothing at
-// all.
+// number comes from, so the bound on it governs nothing here.
 //
-// The contract declares the field `minimum: 0, maximum: 1`. A vendor that
-// scores out of 100, which is a normal thing for a vendor to do, would put 87
-// on it, and the page renders a confidence as a percentage.
+// The contract declares the field `minimum: 0, maximum: 1`, and the page renders
+// a confidence as a percentage — so a vendor scoring out of 100 would read as
+// 8700%. Every Surfe score in this tree is a proportion today; what is guarded
+// is that nothing makes it one.
 
 import (
 	"encoding/json"
@@ -65,6 +65,38 @@ func TestAConfidenceTheContractCannotCarryIsNotServed(t *testing.T) {
 				t.Errorf("dropped a confidence of %v, which the contract can carry", *tc.score)
 			case tc.want != nil && *got != *tc.want:
 				t.Errorf("served %v, want %v", *got, *tc.want)
+			}
+		})
+	}
+}
+
+// The claim's own column is the fallback when an entry carries no score of its
+// own, and it passes through the same guard. The column is CHECK-bounded, so a
+// value that fails here cannot come from a live database — but the guard runs
+// after the fallback merge, and a branch with no test is a branch that stops
+// working quietly.
+func TestTheColumnFallbackPassesThroughTheSameGuard(t *testing.T) {
+	for _, tc := range []struct {
+		name   string
+		column float64
+		want   *float32
+	}{
+		{"a bounded column is used when the entry has no score", 0.42, providerPtr(float32(0.42))},
+		{"an out-of-range column states nothing either", 42, nil},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			claim := phoneClaim(t, nil)
+			claim.confidence = providerPtr(tc.column)
+			var out crmcontracts.PersonProviderProfile
+			if err := foldPhones(claim, &out); err != nil {
+				t.Fatalf("folding: %v", err)
+			}
+			got := out.MobilePhones[0].Confidence
+			switch {
+			case tc.want == nil && got != nil:
+				t.Errorf("served %v from a column of %v", *got, tc.column)
+			case tc.want != nil && (got == nil || *got != *tc.want):
+				t.Errorf("served %v, want %v", got, *tc.want)
 			}
 		})
 	}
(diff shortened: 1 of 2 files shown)


tinnhlt-gra:
## Review round 1 — the guard was right and its explanation was wrong twice

Reviewed by Fable. The defect is confirmed real and reachable end to end, the fix correctly placed, the guard hole-free. **What was wrong was what I wrote about it.**

### A comment asserting a false invariant, inside a guard

> NaN and the infinities fall out of the comparison — neither is `>= 0`.

**`+Inf >= 0` is `true`.** Fable ran it. +Inf is stopped by the *upper* bound alone; only NaN and -Inf fail the lower one. The guard behaves correctly and my account of why it does was false — which is precisely the defect class this tree hunts by name, sitting in the one place whose job is to be trustworthy.

### A uniqueness claim no test held

> this is the one place a bounded contract field is written

False tree-wide — `enrichextract.go` and `orgdossier/evidence.go` write bounded confidences too. Scoped to what is true: the one place a **provider claim's** confidence reaches the contract, which Fable verified (the only other `value_json` readers take job titles and domains, or export the raw map with no contract bound).

### And there was already a sibling guard I didn't name

`enrichextract.go:444` guards the same invariant for extracted fields and answers it **differently**: it drops the whole field, and treats zero as invalid.

Both are right for their own input — an extracted value is model-fabricated and worth nothing without a confidence, where a phone number was paid for and stands on its own. But the rulebook says two writers of one invariant either share a helper or say why they do not, *beside the code*. Neither named the other. Now one does.

Fable's answer to "is the fold the right single home" is better than my framing: **this is the last unguarded member of a class**, not one instance among many. Every `confidence` column in the catalog is CHECK-bounded, the LLM lane is guarded at enrichextract, fill rates and health scores are computed-bounded, and research-claim/moment confidences are enums.

### Overstated as a live defect

Every Surfe score in this tree is a proportion — `0.65` in both the test fixture and the offline demo. Nothing suggests 0-100. This is **hardening of a vendor-controlled input that nothing validated**, which is reason enough; the PR now says that instead of implying a bug in flight.

### On the silent drop

Fable's concern: if a vendor switched scales, every phone confidence would vanish from every page with nothing to explain it — and this adapter's own doctrine argues against silence.

I did not add a log; `foldPhones` has no context and threading one through every fold function costs more than the finding. **The read-time placement is the better answer**: `value_json` keeps what the vendor actually asserted, so a scale corrected later restores every confidence retroactively, the Art. 15 export still shows what was received, and a systemic change is diagnosable from the stored claims. That is now stated rather than incidental.

### Also

The column-fallback branch had no test — the one guarded branch a live database cannot reach, since that column *is* CHECK-bounded. Covered now, because a branch with no test is a branch that stops working quietly.

`make check-backend` green.

PR #2617, opened 2026-08-25
feat(mcp-apps): a card may ask the device where it is

LarsGradion:
## Why

The vision deck's business-card scenario ends with the assistant noticing you are at a conference and offering the venue as a tag. Every route to that fact either **asks the user** — which makes it a form, not a convenience — or **guesses from an IP address**, which on conference wifi names the wrong city. The device already knows.

Whether a view may ask it is the **host's** decision. The MCP Apps extension is explicit that a host *"MAY honor these permissions… but are not required to"*, so this cannot be settled by reading a spec: it has a different answer on claude.ai web, on Android, on iOS and on desktop. This PR makes the question askable.

An earlier probe (19 Aug) tested this through a claude.ai **artifact** and got `"Geolocation has been disabled in this document by permissions policy"` — the iframe `allow` attribute. But an artifact and an MCP App view are different embedding paths, so that was strong evidence and not proof. This settles it.

## What changed

**`sandbox()` declares `Geolocation: true`, on every view rather than one.** A host grants this per iframe, and a view that asks only when it expects to need an answer has already lost the case it was for: the position is wanted when somebody hands over a card, not one card later.

**`navigator.geolocation` came off the admission vocabulary.** It sat under `offOrigin`, among `fetch(`, `WebSocket(` and `sendBeacon` — swept in with the calls it resembles rather than judged on what it does. Reading a sensor sends nothing anywhere.

The promise `offOrigin` keeps is untouched: every network token still stands and the published CSP still names no origin, so **a view may now learn where it is and still has nowhere to send it**. A coordinate leaves only the way every other answer does — through the host, into a tool call the user sees. The reasoning lives in `vite-inline-views.ts`, since JSON cannot carry a comment.

**A `geo-probe` view and a `check_location_support` tool, which exist to be DELETED.** The five product views state that they render and nothing else, and a button that asks for a permission is exactly the authority they refuse — so the question got its own surface instead of weakening a card a customer sees.

## Verified

`_meta.ui` on the wire, from the regenerated `mcp-info.json`:

```
ui://margince/account-brief.html      perms={'geolocation': {}}  csp=all lists empty
ui://margince/relationship-map.html   perms={'geolocation': {}}  csp=all lists empty
ui://margince/commitments.html        perms={'geolocation': {}}  csp=all lists empty
ui://margince/handoff.html            perms={'geolocation': {}}  csp=all lists empty
ui://margince/pipeline-review.html    perms={'geolocation': {}}  csp=all lists empty
ui://margince/geo-probe.html          perms={'geolocation': {}}  csp=all lists empty
```

`{"geolocation": {}}` is the extension's presence-based shape, not a boolean flag.

The built probe document contains the geolocation call and **zero** `http://` or `https://` origins, so it passed both halves of the admission gate.

## Gates

`make check-backend`, `make check-fe`, `make craft-static`, `make rls-store-path`, `make no-jurisdiction` — all green.

Two existing gates were **tightened rather than relaxed**. Both compared the policy against the zero value; both now pin `Geolocation:true` as a whole value, so a *second* permission still fails. Proven by reverting the declaration:

```
hold_test.go:312: the read policy is not the declared one:
  got {Geolocation:false}, want {Geolocation:true}
agentapps_test.go:210: the view ui://margince/account-brief.html asks for
  browser permissions {Geolocation:false}, want {Geolocation:true}
  [and the five other views]
```

Two house rules caught real violations in the second commit: the focus ring spelled its own width instead of reading `--focus-ring`, and `describeEnvironment` put a caught error's own message on screen. Both fixed. The browser's *geolocation* error message is still shown verbatim — that is the finding this view exists to produce.

## What this does NOT do

It does not make the deck's scenario work. It makes it **testable**. The host may still refuse, and refusal stays the expected outcome until a run shows otherwise. The four-host matrix is the next step.

Refs #2616

---
## Summary by cubic
Adds a temporary location probe view and tool to test host geolocation support. Previously all views declared geolocation; now only the probe does, while CSP stays origin-empty so product cards remain permissionless.

- Policy: `sandbox(uri)` now sets permissions per view; `ui://margince/geo-probe.html` declares `geolocation`, all product views declare none. Tests pin the full UI policy on catalog and read, and fail any second permission.
- Admission vocabulary: remove `navigator.geolocation` from the off-origin forbid list; all network denies remain unchanged.
- Probe surface: adds `check_location_support` (registers unconditionally) and a `geo-probe` view with a single button; no data reads or network calls. Shows the browser’s message verbatim and classifies refusals (host-blocked vs user-declined; unknown stays unknown). Unit tests cover the classifier.
- Security clarification: an empty CSP does not restrict `postMessage`; containment is that the probe sends nothing, not that the sandbox makes coordinates unsendable.
- Docs and gates: regenerate `mcp-info` and tool budget; tighten gates to compare full UI policy values; fix focus ring token use and avoid surfacing non-actionable env-query errors.

<sup>Written for commit c7370bfca146f678f31c15e14e95bdd4be5a59a7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/2617?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added a location-support check that explains whether geolocation is available without reading or storing location data.
  - Added an interactive location check view with clear results for successful, denied, unavailable, or host-blocked access.
  - Limited geolocation permission requests to the dedicated location check view; existing views no longer request it.

- **Documentation**
  - Updated published tool and resource catalogs with the location check and revised totals and usage metrics.

### DIFF
--- a/backend/internal/compose/agentapps_test.go
+++ b/backend/internal/compose/agentapps_test.go
@@ -201,14 +201,19 @@ func TestEveryViewIsServedUnderTheAppProfile(t *testing.T) {
 		// added to the seam later would be invisible to a hand-written list, and
 		// silently-unchecked is the one thing a permission must not be.
 		//
-		// Geolocation is the one asked for, on every view, and the CSP assertion
-		// directly above is what makes it safe: a view may learn where it is and
-		// has no origin to send it to. A coordinate leaves only the way every
-		// other answer does — through the host, into a tool call the user sees.
-		// A SECOND permission appearing here is a real widening and fails.
-		if want := (mcp.ResourcePermissions{Geolocation: true}); r.UI.Permissions != want {
-			t.Errorf("the view %s asks for browser permissions %+v, want %+v — a permission beyond geolocation "+
-				"is a widening of the sandbox and this assertion is where it gets argued",
+		// The expected value is per view, and that IS the assertion. A host maps
+		// this declaration onto an iframe `allow` attribute, so a card that
+		// declares a permission it never uses would carry the capability if its
+		// code were ever substituted. Geolocation belongs to the probe, which is
+		// the only view that reads a position; every product card asks for
+		// nothing. A permission spreading to a second view fails here.
+		want := mcp.ResourcePermissions{}
+		if r.URI == apps.GeoProbeURI {
+			want.Geolocation = true
+		}
+		if r.UI.Permissions != want {
+			t.Errorf("the view %s asks for browser permissions %+v, want %+v — a card declaring a permission "+
+				"it does not use is a widening of the sandbox, and this assertion is where it gets argued",
 				r.URI, r.UI.Permissions, want)
 		}
 	}
--- a/backend/internal/compose/agentapps_test.go
+++ b/backend/internal/compose/agentapps_test.go
@@ -197,12 +197,19 @@ func TestEveryViewIsServedUnderTheAppProfile(t *testing.T) {
 				"the sandbox and this assertion is where it gets argued: state why in the diff that adds it",
 				r.URI, r.UI.CSP)
 		}
-		// Compared against the ZERO value rather than field by field: a permission
+		// Compared against a WHOLE value rather than field by field: a permission
 		// added to the seam later would be invisible to a hand-written list, and
 		// silently-unchecked is the one thing a permission must not be.
-		if r.UI.Permissions != (mcp.ResourcePermissions{}) {
-			t.Errorf("the view %s asks for browser permissions %+v; none of these views needs one",
-				r.URI, r.UI.Permissions)
+		//
+		// Geolocation is the one asked for, on every view, and the CSP assertion
+		// directly above is what makes it safe: a view may learn where it is and
+		// has no origin to send it to. A coordinate leaves only the way every
+		// other answer does — through the host, into a tool call the user sees.
+		// A SECOND permission appearing here is a real widening and fails.
+		if want := (mcp.ResourcePermissions{Geolocation: true}); r.UI.Permissions != want {
+			t.Errorf("the view %s asks for browser permissions %+v, want %+v — a permission beyond geolocation "+
+				"is a widening of the sandbox and this assertion is where it gets argued",
+				r.URI, r.UI.Permissions, want)
 		}
 	}
 	if views == 0 {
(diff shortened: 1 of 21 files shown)


LarsGradion:
## Codex review — two real defects found and fixed in `c7370bfca`

**1. My own comment made a false security claim.** It said a coordinate "can only travel back the way every other answer does — through the host, into a tool call the user sees", citing the empty CSP as the guarantee. **That is wrong.** A view has `window.parent.postMessage` (bridge.ts:102 — it is how the bridge talks to its host at all), no origin allowlist governs postMessage, and `postMessage` is not on the admission vocabulary. A substituted view could send a coordinate to the host origin with no tool call.

The comment now states what is actually true: the containment is that the one view holding a position is a probe that displays it and sends nothing — not that the sandbox makes a coordinate unsendable. Filed as **#2619**; the gap predates geolocation and applies to every fact a view holds.

**2. Every card got the permission, and only one uses it.** A host maps the declaration onto an iframe `allow` attribute, so five product cards that never call the API would have carried the capability if their code were substituted. `sandbox()` now takes the URI and grants it to `geo-probe` alone.

The "ask on every view" reasoning — the position is wanted when somebody hands over a card, not one card later — is a real product argument and may come back. It belongs in the diff that makes a product card actually read a position, not ahead of one.

Wire now reads:

```
ui://margince/account-brief.html      perms=None
ui://margince/relationship-map.html   perms=None
ui://margince/commitments.html        perms=None
ui://margince/handoff.html            perms=None
ui://margince/pipeline-review.html    perms=None
ui://margince/geo-probe.html          perms={'geolocation': {}}
```

**3. Also fixed: an unrecognised refusal claimed the user declined.** Code 1 covers a permissions-policy block, an insecure context, a person declining and a persistent OS denial, and the messages are not standardised. The old default told a tester "the permission got through, try again and accept" for a host block — the opposite of the truth, about the one question this probe exists to answer. Two positive matchers now, plus an explicit `refused-unclassified`.

`geo.test.ts` covers the mapping, which had none. Proven by restoring the old default: the three unclassified cases fail.

**Not fixed, deliberately:** the e2e lane still only exercises `account-brief` and `relationship-map`. Extending it to a click-triggered permission request is bigger than this PR and is noted in #2619.

Gates re-run after the changes: `check-backend`, `check-fe`, `craft-static`, `rls-store-path`, `no-jurisdiction` — all green.

Issue #2620, opened 2026-08-25
create_record files a duplicate candidate and never tells the caller, so the agent reports the opposite


LarsGradion:
## What happened

A user asked Claude to create a person who already existed, deliberately, to see the guard work. The second create had the same full name and the same E.164 phone, and no email.

**Margince did the right thing.** A `dedupe_candidate` row was written in the same transaction as the create — identical microsecond timestamp — at confidence 1.000:

```
entity_type person
left        01a0363c-15d9…   (the real record)
right       01a03641-09fb…   (the duplicate)
confidence  1.000
evidence    [{"field":"phone","signal":"collide","score":1,
             "left_value":"+84911066550","right_value":"+84911066550"}]
disposition open
```

**Claude told the user the opposite.** Its summary said *"Same full name and same E.164 phone number wasn't enough to stop it"* and concluded *"the guard is email-only"*. Both false — the phone lane fired and filed the review. It then advised the user that `resolve_entities` is what earns its keep, on the strength of a guard it wrongly believed absent.

## Why it got it wrong

Nothing on the MCP surface can see the queue.

- `resolve_entities` answers a question asked BEFORE a write. Its own tool copy says `/dedupe/candidates` serves the stored review queue and is *"a different question from 'who does this payload name'"*.
- `create_record`'s response says nothing about having filed a candidate.

So the agent wrote the row, could not observe it, and reported from what it could see. The user now believes a guard that works does not exist.

## Suggested fix

The row exists at the moment of the write, in the same transaction. `create_record`'s response can carry it:

```
"duplicate_candidate": {
  "other_record_id": "01a0363c-15d9…",
  "confidence": 1.0,
  "evidence": [{"field":"phone","signal":"collide", …}]
}
```

That needs no new tool and no in-chat decision card. It turns a silent create into a reported one.

The **merge-or-create-anyway choice** is a separate, larger piece and still needs the decision card — that half is unchanged.

## Related

This is the same shape as #2615: an agent narrating a state it has no tool to observe, and getting it backwards. Two instances now on different subsystems, which suggests a surface rule — **a write that had a side effect the caller would want to know about should say so in its own response**, rather than relying on the model to go looking.

Worth considering whether that rule should apply to the approval queue (#2615) and this queue with one mechanism instead of two.

## What it costs if nobody fixes it

Users are told a working safeguard is missing. That is worse than silence: it argues for building something that already exists, and it undermines trust in the guards that do fire.

PR #2623, opened 2026-08-25
feat(people): a second business card for someone already here asks a human

LarsGradion:
## The case

One person, two cards, two addresses. The second card gets typed in and nothing on it matches what is stored except the name — a personal address instead of the work one, or a typo, and often no phone at all.

Until now that created a second live record **in silence**. Nobody was asked.

Found while testing case 2 from the vision deck: a user deliberately created a duplicate to watch the guard work, and the guard was not there.

## Why the existing tiers could not catch it

- **Email lane** cannot fire — the addresses differ, which is the whole premise.
- **Phone lane** cannot fire when the card carries no number.
- **Fuzzy tier** cannot reach it *by arithmetic*: a perfect name scores `0.55 × 1.0` against a `0.72` bar, so it clears only on employer agreement.

And here is the part that is easy to misread as a weighting problem. **`CreatePersonInput` carries no employer at all.** The employment edge is a separate call made *after* the person exists, so at create time `orgMatch` is structurally `0`. The pair was unreachable, not merely unlikely, and no amount of tuning the weights would have changed that — one input the formula needs does not exist yet when the formula runs.

## The lane

Two records written with exactly the same name — after the same folding the rest of the ladder uses — go on the review queue.

A lane of its own rather than a lowered threshold, for two reasons. The weights are shared with organization matching, so moving the bar to admit this pair would drag every company comparison down with it. And an exact name is not a probability — it either is the same string or it is not — so scoring it against a fuzzy bar was the wrong instrument.

**It flags. It never merges and never refuses.** A father and son at one firm are a real pair of records that share a name and an employer; an automatic rule that merged them would destroy data no undo restores. So it asks — the same shape the phone lane already takes for a shared switchboard.

## Opt-in, and this is the part with teeth

`PersonCandidate.QueueNameCollisions` gates the lane, because the answer is only safe for one kind of caller.

- A caller that **creates** wants it. The worst case is a queue row somebody dismisses.
- A caller that **routes** must not have it. Capture and the channel path ask *"which existing person does this message belong to"*, and for them a `PersonID` is a delivery address. An unbound channel identity carrying a common name would name somebody it has no business naming, and **a message would land on a stranger's timeline**.

`TestExactPersonByChannelIdentityMatchesOnProviderAndUserID` caught exactly that when the lane was first written unconditional. That test failing is why the flag exists.

## Implementation note

The best name-identical row is tracked **apart from** the best-scoring row. They are different questions and the winners differ: a near-name at a matching employer outscores an exact name with no employer known, so reading the name lane off the fuzzy winner would lose the very pair the lane is for.

No `dedupe_near_match` ledger line is written for this lane. That ledger records the fuzzy tier's own scoring, and this lane did not score — it compared two normalized strings. A line carrying a confidence would invite someone to tune a threshold that had no part in the decision.

## Tests

Three, each **proven to bite** by reverting the opt-in flag:

1. the second card is queued for a human
2. two real people sharing a name both survive, neither is merged, and the question is still asked once
3. a name nobody shares leaves the queue empty

Test 2 deliberately puts the junior's address on a **different domain**. Both Fischers on `fischerbau.test` scores 0.8 on the org term and carries the pair over the fuzzy bar on its own — which would have proven the old tier still works and said nothing about this one. That was a real defect in the first draft of the test, caught by the sabotage run.

## Gates

`make check-backend`, `make craft-static`, `make rls-store-path`, `make no-jurisdiction` — green.

People integration lane: **451 pass, 0 fail**.

Also verified directly against Postgres that the fuzzy candidate set cannot exclude an exact-name row: identical strings always satisfy the trigram `%` operator regardless of length, and `f_fold_apostrophes(lower(…))` folds accents and ß the same way Go's `normalizeName` does.

## What this does not do

The **merge-or-create-anyway choice in chat** is still not expressible — `create_record` reports nothing about having filed a candidate, so an assistant cannot tell the user. That is #2620 and it is the other half of this problem.

Refs #2620

---
## Summary by cubic
Queues human review on exact-name collisions during person creation and guards it under contention. Previously, a second card with only a matching name created a duplicate silently; now we flag it without merging or refusing, and routing paths are unaffected.

- Adds `DecisionNameCollisionReview`; tracks the exact-name candidate separately from the fuzzy best; no weight/threshold changes and no auto-merges.
- `manualDedupePerson` opts in via `PersonCandidate.QueueNameCollisions = true`; routing callers keep it false to avoid misdelivery.
- Writes only to `dedupe_candidate` with `nameCollisionEvidence`; no `dedupe_near_match` entry because this lane is non-probabilistic.
- Introduces `lockNameLane` (normalized key, taken after the phone lock) to prevent two concurrent identical-name creates from bypassing the queue; both creates may land, and one review pair is enqueued.
- Tests cover: second card is queued; identical-name different people remain unmerged; unrelated names do not queue; contention enqueues exactly once. Related to #2620 (chat still cannot surface “queued” vs “created”).

<sup>Written for commit 4bbf92460f49636a03e8ebac5c824876082aada1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/2623?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added human-review queue support for people with identical names.
  * Matching records are flagged for review without being automatically merged or rejected.
  * Similar records, including additional addresses at the same employer, can be queued for review.
  * Records with different names remain unaffected and are not added to the review queue.
  * Improved handling of simultaneous record creation to ensure matching entries are consistently queued for review.

### DIFF
--- a/backend/internal/modules/people/creatededupe.go
+++ b/backend/internal/modules/people/creatededupe.go
@@ -63,7 +63,13 @@ func manualDedupePerson(ctx context.Context, tx pgx.Tx, in CreatePersonInput) (P
 	if err := lockPhoneLane(ctx, tx, phones); err != nil {
 		return PersonResolution{}, err
 	}
-	return DedupePerson(ctx, tx, PersonCandidate{FullName: in.FullName, Emails: emails, Phones: phones})
+	// QueueNameCollisions, because this caller creates rather than routes. A
+	// pair it names goes on the review queue and nowhere else — no message is
+	// delivered on the strength of it, so the worst case is a row a human
+	// dismisses. See PersonCandidate for why the routing callers must not ask.
+	return DedupePerson(ctx, tx, PersonCandidate{
+		FullName: in.FullName, Emails: emails, Phones: phones, QueueNameCollisions: true,
+	})
 }
 
 // phoneLaneKeys is the request's numbers in the form the phone lane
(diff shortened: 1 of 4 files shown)


LarsGradion:
## Codex review — one real defect fixed in `4bbf92460`, one filed

**Fixed: the lane had the phone lane's race and no guard for it.**

At READ COMMITTED two creates carrying one name both read no committed incumbent, both fall to no-match, and both commit. Neither files a candidate, and nothing writes one afterwards — every row on that queue comes from the path that detected the collision. Two live duplicates, no human asked. That defeats the lane entirely in the case it was built for: two people typing the same business card in at once is exactly when this happens.

`lockNameLane` is `lockPhoneLane` for the name, and its existing comment already makes the argument — the lane has no unique index and must not have one, so nothing structural is left.

The key is the **normalized** name. Locking the raw string would let `Lucy Vo` and `LUCY VO` take different locks and race each other, which is the pair most likely to be entered twice. The contention test uses those two spellings deliberately, so it fails if the key is ever taken on the raw name.

Proven by removing the lock:

```
creatededupe_contention_integration_test.go:249: the review queue holds 0 candidates,
  want exactly 1 — two live people are written the same way and no human will ever be asked
```

The test follows the phone lane's pattern: one create frozen by hand past its lock and its insert, the second released only once a backend is provably waiting on it. No sleep, no clock.

**Filed as #2624: the SQL prefilter and the Go equality test fold names differently.**

The candidate query admits rows by `f_fold_apostrophes(lower(…)) %`; the collision test uses Go's `normalizeName`. SQL does `lower` + `unaccent`; Go does Unicode *full* folding plus combining-mark removal. `namesim.go` says outright they need not agree — which was fine while SQL was a prefilter for a *score*, and is not fine for an equality lane.

I could not reproduce it. Probed the live database and Postgres folded every case the same way Go does — `Straße`/`Strasse`, `Müller`/`MÜLLER`, `İhsan`/`Ihsan`, `ǅon`/`Dzon`, and identical strings at any length always satisfy `%`. So it is structural rather than demonstrated, and deferred on that basis: two independent normalizations with no equality fallback will drift, and nothing currently fails if they do.

**Confirmed safe by the review, worth recording:** the opt-in is sufficient (only `manualDedupePerson` sets it; all five routing callers leave it false); no caller treats the review-only `PersonID` as a routing or merge target; the tie-break is deterministic; the new queue write has identical tenancy scope to the existing one; and skipping the `dedupe_near_match` ledger line breaks nothing — two other producers already write queue rows without one, and no production reader of that action string exists.

Gates re-run: `check-backend`, `craft-static`, `rls-store-path`, `no-jurisdiction` green. People lane **452 pass, 0 fail**.

PR #2625, opened 2026-08-25
feat(company360): the overview leads with the account's work in flight

LarsGradion:
The Company 360 Overview opened with a written account brief: one narrative over
every deal and every project at once. On an account carrying several engagements
that is the wrong shape — correspondence about one project becomes a sentence
about another, and a figure read out of the blend has nowhere to be checked.

The overview now leads with **the account's work in flight**: one line per open
deal, one per live project, each written from that record's own facts and
linking to it. A deal and a project are two stories, and the structure says so —
two groups under their own subheads, never interleaved, and the header states
counts and nothing else.

Each line carries at most ONE reason it wants a person, picked server-side: an
overdue task, or a commitment they made to us that is still open. The sentences
are i18n templates over typed fields, never model-written. A commitment is
QUOTED rather than asserted over — the body is free text an extractor read out
of a conversation, so "they owe us {body}" produces nonsense where the body is a
proposition rather than a noun phrase.

## Four states, not three

| Situation | The slot shows |
|---|---|
| both halves readable, rows exist | the lines |
| both readable, both empty | the growth-fit panel — the question an account with no work is actually asking |
| either half withheld | that group withheld, the other renders, **no growth fit and no count** |
| `attention_withheld` | rows render, with one line saying the statuses are incomplete |

A withheld pipeline must never read as "no pipeline, go read the fit", and
"0 in flight" to a reader who can see the deals but not the projects is a false
statement about the account rather than a partial one.

## Also here

**The facts box.** Open pipeline, in-flight counts and owner beside the account
name, in RecordView's `controls` slot — the same slot the deal record draws its
own facts box in. The owner control MOVES there rather than being copied: one
implementation, still editable, and gone from the identity meta line. Lifecycle
stays in the name badge where it is edited.

The pipeline figure has three states a reader must tell apart: a missing
commercial reading is the READER lacking the deal grant, zero open deals is a
fact about the account, and open deals with no total means nobody has priced
them — which a dash would report as "we do not know".

**The brief endpoint stays, deprecated.** Only the UI goes. `POST …/ask` is
served from the same handlers and the same assembly, so deleting the wiring
would take Ask with it.

**Since-last-visit moved rather than went away.** It sat in the brief's footer
and is about the account, not about the brief, so it is now the work card's
footer. Deleting it silently would have removed the only display of "N new items
since your last visit" and the withheld-sections notice.

## Fixed along the way

- `Organization360Project.quiet` is computed with the module's own
  `ProjectQuietSQL` and made REQUIRED. Optional would put an absent value and a
  false one on the same wire, and a card reading the absence would call a silent
  project active.
- The projects surface fetches cap+1 and reports `has_more` through a new
  `projects_page`. It truncated at 25 silently, so a card counting rows would
  state a number a portfolio account does not have. (`projects` stays an array —
  turning it into an object is a breaking change the contract gate rejects.)
- `projects.go`'s doc comment said the section reads "through the deals store".
- `dealfacts.tsx` claimed the company page already used `controls`. It now does.

## Testing

- 14 frontend unit tests (`companywork.test.tsx`), 5 backend integration tests
  against real Postgres.
- **Mutation-checked**: reverting the account walk fails 4 integration tests
  including the cross-account one; reverting the open-and-undisputed claim filter
  fails the settled/contradicted case; reverting the withheld-count rule, the
  growth-fit fallback and the one-status-clause-per-row rule each fail their own
  test.
- **Seen working in a browser** against a real stack, in German: both sentence
  shapes render, the precedence holds (completing the task promoted the
  commitment), and the growth-fit panel takes the slot on an empty account.
- **Not verified by hand**: the withheld and `attention_withheld` states. The dev
  seed has no reduced-permission seat to log in as. Both rest on the integration
  and unit tests above, both mutation-checked.

The query-count budget rose 39 → 42, one per new set-based read, each flat in
the size of the account.

## Contract

Additive only — `oasdiff` reports no breaking changes. `x-enum-varnames` pins the
new enum's constants, and `CommitmentTheirs` on `ConversationClaimKind` is
unchanged (the known collision hazard).

---
## Summary by cubic
Leads the Company 360 overview with the account’s work in flight and moves account facts beside the name. Replaces the written brief with deterministic, per-record reasons and fixes visibility so commitments don’t leak across missing grants.

- Work-in-flight card: one line per open deal and live project; header shows counts only; each row carries at most one server-picked reason (overdue task or quoted commitment). Supports four states including `attention_withheld` when reasons can’t be shown.
- Facts box beside the name: open pipeline value, in‑flight counts, and an editable owner control. The pipeline figure distinguishes withheld, zero, and unpriced.
- Contract (additive): adds `projects_page.has_more` (cap+1 read), makes project `quiet` required and server-computed, adds per-row `attention` with enum `Organization360WorkAttentionKind`, and adds `attention_withheld`. Marks `GET /organizations/{id}/brief` and `POST /organizations/{id}/brief` deprecated; `POST /organizations/{id}/ask` remains.
- Security and correctness fixes: commitments now require `person:read`; person scope is selected (not filtered) so dropped actors render with `attention_withheld`; `projects_page.has_more` ignores cut closed projects; never‑touched quiet projects render a specific sentence.
- Query budget rises 39 → 42 from three flat, set-based reads. Tests and docs updated; new integration tests cover attention selection; new unit tests for the work card; e2e and screen specs assert the new lead slot.

**Rollout**
- Replace the brief card with the new work-in-flight card in the overview.
- Use the new facts box beside the company name; the owner control moved there and remains editable.
- If your UI consumed the brief’s since-last-visit footer, read it from the work card’s footer now.
- Continue serving `…/brief` only for legacy clients; plan to remove it later.

<sup>Written for commit 212cd8be1b34ab3a336dde6cee07259a021eb96e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/2625?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added company facts showing pipeline value, deal/project counts, and ownership.
  * Added a “What’s in flight” view for active deals and projects, including overdue tasks, commitments, statuses, and due dates.
  * Added pagination indicators and clear messaging when details are restricted.
  * Added English, German, and Vietnamese translations for the new views.

* **Changes**
  * Replaced the company account brief on record pages with facts and active-work summaries.
  * Existing account-brief integrations remain available but are deprecated.


LarsGradion:
## Review pass — four defects found and fixed

A review of the branch found four real defects. All four are fixed in `212cd8be1`, and each fix is mutation-checked (revert the guard, watch the matching test fail).

**Two were disclosures.**

`CommitmentsTheirsForProjects` gated on `activity:read` and never on `person:read`. It then called `auth.ScopeClauseFor(ctx, "person", …)`, which narrows a set an object grant has already opened and returns no predicate at all for an unbounded actor. A reader holding project and activity but not person could read a commitment's body and the name of the person who made it. The gate now lives in the store method rather than at one call site.

A claim made by someone outside the caller's row scope was filtered away in SQL and vanished without a word, so a project silently missing its commitment read as a project with nothing outstanding — the exact failure this card exists to prevent. Person scope is now SELECTED rather than filtered on: the row still wins its project, the caller does not get it, and `attention_withheld` says so.

**Two were honesty defects in what the card states.**

`projects_page.has_more` reported a cut CLOSED project. Closed projects sort last, so an account with one live project and twenty-five finished ones overflowed the cap without dropping anything in flight — and the card read "1+ in flight" off it. The flag now reports *what* was cut, not that something was.

A never-touched quiet project rendered "Nothing has been filed against this project since **never**". The server measures its quiet from the day it was opened and the payload carries no `created_at`, so that case gets its own sentence.

Three new integration tests and two new unit tests cover these; the suite is now 9 integration and 16 unit.

### One test that passed for the wrong reason

My first row-scope test changed a person's owner and expected them to be hidden. They were not — customer identity is workspace-readable here, and capture privacy (`visibility='owner'`) is what actually hides a person from a colleague. It failed against correct code, which is how I caught it; had the production code been broken it would have passed and proved nothing.

PR #2637, opened 2026-08-25
feat(agents): a create that files a duplicate for review says so

### LINKED ISSUE #2620: create_record files a duplicate candidate and never tells the caller, so the agent reports the opposite

### LINKED ISSUE #2624: The dedupe name lane's SQL prefilter and its Go equality test fold names differently
## What is wrong

The person dedupe name-collision lane (PR #2623) decides a pair collides with Go's `normalizeName`, but the candidate rows it can even see are chosen by SQL:

```sql
f_fold_apostrophes(lower(p.full_name)) % f_fold_apostrophes(lower($1))
   OR ($2::uuid IS NOT NULL AND r.organization_id = $2)
```

Those are two different normalizations. SQL does `lower` + `unaccent` + apostrophe removal. Go does Unicode **full** case folding, combining-mark removal and trimming. `namesim.go` says outright that they are not required to agree rune for rune, on the reasoning that SQL only narrows the candidate set.

That reasoning held while the SQL side was a **prefilter for a score**. It does not hold for an equality lane: a pair Go would call identical can fail the SQL `%` test, never enter the candidate set, and never reach the equality check at all.

The org fallback cannot rescue it on the create path — `CreatePersonInput` carries no employer, so `$2` is always NULL there.

## How likely

Unclear, and that is the honest answer. Probed against the live database and Postgres folded every case tried the same way Go does:

| pair | SQL folded | trigram match |
|---|---|---|
| `Straße` / `Strasse` | `strasse` / `strasse` | yes |
| `Müller` / `MÜLLER` | `muller` / `muller` | yes |
| `İhsan` / `Ihsan` | `ihsan` / `ihsan` | yes |
| `ǅon` / `Dzon` | `dzon` / `dzon` | yes |
| identical strings, any length | — | always |

So no reproducing case was found. The defect is **structural rather than demonstrated**: two independent normalizations with no equality fallback will drift, and nothing currently fails if they do.

## Suggested fix

Add an equality arm to the candidate query so the lane's own comparison is what admits its own rows, rather than relying on the trigram operator to be a superset of it. Roughly:

```sql
OR f_fold_apostrophes(lower(p.full_name)) = f_fold_apostrophes(lower($1))
```

That is still SQL folding, so the stronger version is a stored normalized-name column written by the same Go function, indexed, and compared exactly — which also gives the lane an index instead of a trigram scan.

Either way the invariant worth asserting in a test: **any pair `normalizeName` calls equal must be admitted by the candidate query.**

## What it costs if nobody fixes it

The lane silently misses some Unicode-equivalent names — the exact gap it was added to close. It fails quietly, in the direction of doing nothing, which is the failure mode nobody notices.

Found by Codex review on PR #2623.

LarsGradion:
## Why

Asked to create a person who already existed, the ladder **did its job**: it filed the pair on the review queue at confidence 1.000 on a shared phone. `create_record` answered with the record and nothing else, so the assistant driving it could not see what it had just done — and told the user *"the guard is email-only"*, which was false, and argued for building a safeguard that was already working.

That is worse than a missing feature. A user reading the chat is told a working guard does not exist, which undermines trust in the ones that do fire.

Closes **#2620** and **#2624**.

## What the answer carries now

Both channels, because each says what the other cannot:

- a **warning** — what a model reads without being asked
- **`duplicate_candidates`** in the data — the other record's id, the confidence, the evidence. No sentence should make a model parse an id out of prose.

```
data.duplicate_candidates[] = { other_record_id, confidence, evidence[] }
evidence[]                  = { field, left_value, right_value, signal, score }
```

`createdRecord` embeds `wireRecord`, so the shape a caller already parses is unchanged and the member is `omitempty` — additive.

## Design constraints worth stating

**Read back, not returned by the write.** `datasource.SystemOfRecordProvider` is frozen at v1: a fork may ship its own adapter, so widening `Create`'s return would break every one. The candidate is read after the write through a seam compose injects, so RBAC and row scope apply exactly as on the HTTP path.

**It cannot fail the call.** The record is committed by the time this runs. A caller handed an error retries, and a retry creates a *second* duplicate — precisely the harm the report exists to prevent. A failed check gets its own warning code, separate from "a duplicate was filed", because folding them would let a caller report "no duplicate" for a check that never ran.

## Codex review found two real defects, both fixed

**The disclosed records were free.** Each candidate names another record and quotes values off it. This surface's own rule — stated in `noteEvidence` — is that naming a record to an agent is handing it over, and it charges the read bound for exactly that. A create returning five candidates charged for the one record it wrote; repeating it walked the workspace for free. Each disclosed record is now cited and charged.

**The answer claimed a trust tier it had not earned.** The evidence quotes the other record, and this call never read that row's provenance. A contact captured off inbound mail is T2, and carrying its number under the T1 envelope of a record a human just typed raises the trust of the untrusted half. `noteDerivedContent` is the honest label.

Both are asserted in the integration test and proven to bite — removing them fails with *"the disclosed record … was handed over uncharged"*.

Codex **cleared** row visibility, placeholder numbering, the swallowed-error decision, wire compatibility, and the org360 change.

## #2624, and an honest result

The name lane's candidate query gained a third arm comparing folded names for equality, so the lane's decision and the set it decides over use one predicate. The divergence is real: `normalizeName` trims and SQL does not, so `"  Lucy Vo  "` and `"Lucy Vo"` are equal in Go and unequal to a bare SQL comparison.

**But no known pair needs the arm.** Every case that could be constructed — case, accents, ß, a trailing space — is already admitted by the trigram arm. Proven by running the new invariant test with the arm removed: all five pass. So it ships as a **guarantee, not a bug fix**, and the comment says exactly that. The trigram operator is a similarity threshold, and this lane's correctness should not rest on one approximate predicate happening to cover another normalization's output.

`TestEveryNameGoCallsEqualReachesTheNameLane` states the invariant as a test. A pair Go does not call equal *skips* rather than passes, so the table cannot quietly stop asserting anything.

## Also: main was red before this branch

Two fitness tests were failing on `org360/workattention.go` from #2625 — a hand-spelled "still works here" predicate where `identity.LiveMemberSQL` is the definition, and a uniqueness claim naming no gate. Both fixed in place. `LiveMemberSQL("u")` emits the identical SQL, so behaviour is unchanged.

## Gates

`check-backend`, `craft-static`, `rls-store-path`, `no-jurisdiction` — green.
People lane **686 pass**, compose integration lane **1312 pass**, 0 fail.

---
## Summary by cubic
create_record now reports when it filed a duplicate for review. Previously it was silent; now it returns structured duplicate_candidates and a warning, so assistants can see and act on the guard’s outcome.

- Adds a read-after-write seam `compose.openDuplicatesFor(pool)` that uses `people.OpenCandidatesNaming`. It cannot fail the call: on success it includes `duplicate_candidates[] = { other_record_id, confidence, evidence[] }` and warns with `duplicate_filed_for_review`; on error it warns `duplicate_check_unavailable`. Each disclosed record is cited (quota-charged) and the answer is labeled as derived content.
- Extends the person-name lane’s candidate query with a folded-equality arm to ensure any pair `normalizeName` considers equal reaches the lane; ships as a guarantee with an invariant test. Closes #2624.
- Hardens evidence decoding: numeric/null values render as strings; unreadable values surface as `unreadable: …` instead of vanishing.
- Updates the `create_record` output schema to include additive `duplicate_candidates`; docs (`mcp-info.*`) updated. Closes #2620.
- `org360/workattention.go` switches to `identity.LiveMemberSQL("u")`; behavior unchanged.

**Migration**
- Update all `agents.RegisterCoreTools(...)` call sites to pass the new duplicates seam: `agents.RegisterCoreTools(reg, provider, stages, promoter, ownership, consumerMail, duplicates)`. Pass `nil` if no queue is available.
- Optional: Consumers may read `duplicate_candidates` and warnings from `create_record`; the response remains backward compatible. Monitor `duplicate_check_unavailable` as a wiring/availability signal.

<sup>Written for commit f0890504cc1f8d5bba203b16496e34d57254c947. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/2637?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added duplicate detection to record creation, including candidate records, comparison evidence, confidence levels, and warnings.
  - Record creation remains successful when duplicate checks fail, with an appropriate warning.
  - Added support for identifying duplicate candidates by normalized names across people, organizations, and leads.
  - Improved evidence handling for duplicate comparisons, including unavailable or unreadable values.

- **Bug Fixes**
  - Improved overdue-task eligibility by consistently using live member status.

- **Documentation**
  - Updated create-record response schemas and MCP reference documentation.

### DIFF
--- a/backend/internal/compose/agentcommandbothdoors_test.go
+++ b/backend/internal/compose/agentcommandbothdoors_test.go
@@ -453,7 +453,7 @@ func bothDoorsRegistry(staging agents.Approvals) *agents.Registry {
 		agents.WithTierFloor(func(string, string) (mcp.RiskTier, bool) {
 			return mcp.TierConfirmationRequired, true
 		}))
-	agents.RegisterCoreTools(reg, channelAnchor{}, nil, nil, nil, nil)
+	agents.RegisterCoreTools(reg, channelAnchor{}, nil, nil, nil, nil, nil)
 	agents.RegisterEnrichTool(reg, channelAnchor{}, nil)
 	agents.RegisterLifecycleTools(reg, channelAnchor{}, nil, nil, nil)
 	agents.RegisterCommsTools(reg, bothDoorsComms{}, channelAnchor{})
--- a/backend/internal/compose/agentcommandstagedrow_integration_test.go
+++ b/backend/internal/compose/agentcommandstagedrow_integration_test.go
@@ -228,7 +228,7 @@ func archiveRegistry(e *integration.Env) *agents.Registry {
 	// than latent — the tool refuses on the terms an unreadable list refuses
 	// on, so an accidental call here would fail loudly instead of deriving a
 	// company from the compiled-in baseline.
-	agents.RegisterCoreTools(reg, native, native, nil, fieldOwnership{pool: e.Pool}, nil)
+	agents.RegisterCoreTools(reg, native, native, nil, fieldOwnership{pool: e.Pool}, nil, nil)
 	return reg
 }
 
--- a/backend/internal/compose/agentgate_dealreopen_test.go
+++ b/backend/internal/compose/agentgate_dealreopen_test.go
@@ -209,7 +209,7 @@ func TestRelinkingOntoAProjectReachesAHumanAndOtherDestinationsDoNot(t *testing.
 		records: versionedDeal{stageID: stage, version: 3},
 	}
 	reg := agents.NewRegistry(nil, auth.NewGate(fullSeat{}))
-	agents.RegisterCoreTools(reg, deps.records, deps.stages, nil, nil, nil)
+	agents.RegisterCoreTools(reg, deps.records, deps.stages, nil, nil, nil, nil)
 	// Nil dependencies are enough: resolving a tier reads the arguments and
 	// invokes no handler.
 	agents.RegisterLifecycleTools(reg, deps.records, nil, nil, nil)
@@ -272,7 +272,7 @@ func TestEveryDynamicTierRouteHasACommandThatAnswersItsTier(t *testing.T) {
 		records: versionedDeal{stageID: stage, version: 3},
 	}
 	reg := agents.NewRegistry(nil, auth.NewGate(fullSeat{}))
-	agents.RegisterCoreTools(reg, deps.records, deps.stages, nil, nil, nil)
+	agents.RegisterCoreTools(reg, deps.records, deps.stages, nil, nil, nil, nil)
 	// relink_activity is a lifecycle tool, and it is dynamic — the walk below
 	// derives the routes to check from the policy table, so a registry missing
 	// this set would report the route as unresolvable rather than skip it.
--- a/backend/internal/compose/agentgate_test.go
+++ b/backend/internal/compose/agentgate_test.go
@@ -21,7 +21,7 @@ import (
 // list.
 func TestContractTierNeverBelowRegistryTier(t *testing.T) {
 	registry := agents.NewRegistry(stubApprovals{}, nil)
-	agents.RegisterCoreTools(registry, nil, nil, nil, nil, nil)
+	agents.RegisterCoreTools(registry, nil, nil, nil, nil, nil, nil)
 
 	for route, pol := range agentPolicies {
 		if pol.Access != accessTool {
@@ -52,7 +52,7 @@ func TestContractTierNeverBelowRegistryTier(t *testing.T) {
 // staging, so both artifacts are pinned.
 func TestUpdateRecordIsAutoExecuteOnBothArtifacts(t *testing.T) {
 	registry := agents.NewRegistry(stubApprovals{}, nil)
-	agents.RegisterCoreTools(registry, nil, nil, nil, nil, nil)
+	agents.RegisterCoreTools(registry, nil, nil, nil, nil, nil, nil)
 
 	spec, ok := registry.Spec("update_record")
 	if !ok || spec.Tier != mcp.TierAutoExecute {
@@ -161,7 +161,7 @@ func TestGovernanceOperationsAreHumanOnly(t *testing.T) {
 // tool fails closed.
 func TestOperationSpecTightenOnly(t *testing.T) {
 	registry := agents.NewRegistry(stubApprovals{}, nil)
-	agents.RegisterCoreTools(registry, nil, nil, nil, nil, nil)
+	agents.RegisterCoreTools(registry, nil, nil, nil, nil, nil, nil)
 
 	spec, _, ok := operationSpec(agentPolicy{Op: "archivePerson", Access: accessTool, Tool: "update_record", Tier: tierConfirmationRequired}, registry)
 	if !ok || spec.Tier != mcp.TierConfirmationRequired {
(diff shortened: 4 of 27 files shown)


LarsGradion:
Rebased onto `bc9c38484`. The org360 fix in this branch conflicted with #2636, which fixed the same two fitness failures in parallel — main's version taken, they are equivalent.

Post-rebase: `check-backend` fails on `TestEveryGateDeclaresItsShape`, which is **not from this branch**. `retainedcolumncases_test.go` arrived with #2565 and carries no `//gate:kind` line. This branch touches no file in that area. Everything else in the gate is green, and the duplicate-report integration tests pass on the new base.

PR #2641, opened 2026-08-25
test(records): the name lane partitions its cases instead of skipping one

dungbanguyen:
`main` is red, and every integration run of shard 5 fails on it.

`TestEveryNameGoCallsEqualReachesTheNameLane` (added by #2637) guards each case with `t.Skipf` when `normalizeName` does not equate the pair. The `apostrophe` case takes that branch — `normalizeName("Sean O'Brien") != normalizeName("Sean OBrien")` — and the integration lane refuses a skip outright, for the reason it gives itself:

```
FAIL: integration tests must not skip — provision the env/service, do not skip:
    --- SKIP: TestEveryNameGoCallsEqualReachesTheNameLane/apostrophe
```

That is CLAUDE.md's rule working as intended: *"a skipped security gate looks exactly like a passing one"*.

## The fix keeps the test's own judgement, and moves where it is made

The obligation the test names is **"every name GO CALLS EQUAL reaches the lane"**. A pair `normalizeName` does not equate is genuinely outside that — the lane owes it nothing, which is exactly what the original `Skipf` message said. The judgement was right; the place was wrong.

It is now made **once, before the subtests**, so no case is started and abandoned:

- pairs `normalizeName` equates run the lane assertion, unchanged — five of the six, all still passing;
- pairs it does not are **named in the log**, so the tested set stays visible rather than silently smaller;
- an empty inside half now **fails** — a case table that proves nothing should not pass quietly.

## What this deliberately does not decide

Whether `normalizeName` *should* equate `"Sean O'Brien"` and `"Sean OBrien"`. That is a real dedupe question — an apostrophe is exactly the kind of thing two records of one person differ by — but it is a question about the normalization, and it belongs to whoever wrote the case rather than to whoever unblocked the lane. Filed separately so the intent is not lost.

I have not touched `normalizeName`, the lane, or any other case.

## Verified

- `TestEveryNameGoCallsEqualReachesTheNameLane` passes with **no skip**, five subtests asserted, the apostrophe pair reported.
- Full `people` integration lane green, `0 skips`.
- `make check-backend` green, `craft static --strict` green (0 blocker, 0 major, 0 minor).

### DIFF
--- a/backend/internal/modules/people/creatededupe_integration_test.go
+++ b/backend/internal/modules/people/creatededupe_integration_test.go
@@ -434,23 +434,47 @@ func TestANameNobodySharesLeavesTheQueueEmpty(t *testing.T) {
 // does not. The trigram arm happens to rescue it, which is exactly why this test
 // exists — the lane must not depend on one approximate arm covering for another.
 func TestEveryNameGoCallsEqualReachesTheNameLane(t *testing.T) {
-	for _, tc := range []struct {
+	type namePair struct {
 		name      string
 		incumbent string
 		second    string
-	}{
+	}
+	candidates := []namePair{
 		{"identical", "Ada Lindqvist", "Ada Lindqvist"},
 		{"case differs", "Ada Lindqvist", "ADA LINDQVIST"},
 		{"untrimmed", "Ada Lindqvist", "  Ada Lindqvist  "},
 		{"accents", "Renée Bär", "Renee Bar"},
 		{"sharp s", "Anna Straße", "Anna Strasse"},
 		{"apostrophe", "Sean O'Brien", "Sean OBrien"},
-	} {
+	}
+	// The obligation is "every name GO CALLS EQUAL reaches the lane", so a pair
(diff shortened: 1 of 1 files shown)


dungbanguyen:
Superseded by #2647, which landed first and is the better fix.

Both of us moved the judgement out of the started subtest. The difference is what happens to a row the key does not equate: I partitioned it out and logged it, #2647 makes it a `Fatalf` — *"a row the key does not call equal is a row this table has no business carrying: it asserts nothing and says so to nobody"*.

That is the stronger rule and the one more in keeping with this tree. Mine tolerated an aspirational row indefinitely; theirs refuses it, so the table cannot quietly shrink to nothing. Their comment also names the failure I had not thought through — a skip is how the five REAL rows would go quiet if the key ever stopped folding case or accents, which is this file's own warning aimed at the test rather than the lane.

Closing unmerged. The apostrophe question outlives the row that was dropped and lives on in #2642.

Issue #2443, opened 2026-08-23
Frontend paste artifacts: dayMonth ×4 drops locale and zone, a third card spelling, and friends


LarsGradion:
Found by a tree-wide duplication audit (the six-probe scan in `docs/principles/one-source-of-truth.md`, run against origin/main). Umbrella for the frontend paste artifacts. Item 2 is the only one with a user-visible consequence today, and it is cosmetic; item 1 has since been fixed elsewhere.

**1. ~~The abbreviated date is hand-rolled four times past `formatDateAbbrev`, dropping locale and zone.~~ FIXED by #2454 (`f5d7b2a04`).** Every site named here — `personmemory.tsx` (`dayMonth`), `personrail.tsx` (`dayMonth`), `personcards.tsx` (`shortDate`), `personstrip.tsx` (inline), and the `clock` zone gap — is gone; `toLocaleDateString(undefined` no longer appears anywhere in `frontend/src/`. That PR closed this under the audit's own item **B15** without referencing this issue, and its census was wider than the one above: 34 call sites in three classes, including 11 that hand `new Intl.*` the app's own language code (`"en"` resolves to en-US inside a product pinned to en-GB) — a class no `toLocale` grep can see. It also fixed a live `assertIanaZone` divergence in `scheduledsends.tsx`, where a fixed-offset zone passed the local probe and then threw in `formatDateTime`. Left numbered so items 2–6 keep the numbers they are referred to by.

**2. `.fact-card` / `.legal-card` — the same choice card spelled twice in one stylesheet, third spelling of a design-system control.** `screens/onboarding.css:1235` (`.legal-card`) and `:1300` (`.fact-card`) have byte-identical ten-declaration base rules and identical `.selected` rules; both call sites (`onboarding-company-form.tsx:190` and `:287`) render the same aria-pressed button skeleton and even share `.fact-check`. The design system already has this control (`RecordPicker` option treatment, `design-system/atoms.css:696`), and the pair re-implements `.card`'s ground minus the `--shadow-card` — so these cards already sit a shadow away from every other card. The twins have also drifted from each other (secondary-text color, small-text margins). This is the third occurrence of the "second spelling of a card" the frontend rulebook warns about.

[...]

Issue #2442, opened 2026-08-23
Backend helper paste artifacts: the $N arg closure ×164, retryAfter ×4, backoff pair, deref ×6, and friends


LarsGradion:
Found by a tree-wide duplication audit (the six-probe scan in `docs/principles/one-source-of-truth.md`, run against origin/main). Umbrella for the backend helper paste artifacts — each item is small; together they are the "same 10–30 line helper re-typed per module" pattern.

**1. The `$N` placeholder closure — 164 copies in 111 files.** The exact line `arg := func(v any) int { args = append(args, v); return len(args) }` is re-typed once per query-building function (people 22, activities 16, org360 15, deals 9, …). storekit already owns the consumer type — `storekit.CustomFilterClauses(…, arg func(any) int)` at `platform/database/storekit/listquery.go:208` — and one blessed producer (`ListPrelude.Arg`, `listpage.go:51`). Three incompatible spellings already exist: int-returning (the 164), `"$%d"`-string-returning (`compose/datasweep.go:318`), and `strconv.Itoa` (`modules/privacy/auditlog.go:280`). Suggested fix: a `storekit.ArgList` with `Add`/`Args`; adopt opportunistically per touched file rather than one big sweep.

**2. `retryAfter(resp)` — 4 byte-identical copies under capture/**: `capture/gmail/client.go:382`, `capture/googleconn/googleconn.go:282`, `capture/graph/transport.go:38`, `capture/oauthflow/oauthflow.go:272`. None handles RFC 9110's HTTP-date form of `Retry-After`; adding it means editing four files. All in one module tree, so a `capture/internal/httpretry` package suffices.

**3. Jittered backoff — identical pair across siblings**: `capture/syncstate.go:77` and `overlay/syncbackoff.go:81` are character-for-character identical apart from constant names, down to the same `//nolint:gosec` comment. Siblings cannot import each other, so this needs a `platform/backoff` home. (webhooks/delivery.go, blobstore/s3.go, integrations/handoff.go are genuinely different policies — leave them.)

[...]

PR #2684, opened 2026-08-25
refactor(capture,platform): one reading of Retry-After, and it reads the whole header

tinnhlt-gra:
Wave 6 of the duplication audit, first of three. Item 2 of #2442 — and the census was short by two, in the direction that matters.

## The census

The issue names four byte-identical copies of `retryAfter(resp)` under `capture/`. There are **six** readings of the header, and the two the census could not see are the two that are different:

| site | shape |
|---|---|
| `capture/gmail/client.go` | `retryAfter(resp)` — delta-seconds only |
| `capture/googleconn/googleconn.go` | byte-identical |
| `capture/graph/transport.go` | byte-identical |
| `capture/oauthflow/oauthflow.go` | byte-identical |
| `platform/geocode/geocode.go` | `retryAfter(header string)` — **both RFC forms**, takes the value not the response |
| `capture/telegram/api.go` | `retryAfterOf(resp, env)` — envelope first, header as fallback, capped |

The census grepped for the four-argument shape, so the two that took a different argument were invisible to it. Same blind spot as #2443's month cut: a census keyed on the spelling it was written for cannot see the respelling.

## This is a defect, not a tidy-up

RFC 9110 §10.2.3 gives `Retry-After` **two** forms — `delta-seconds` and an HTTP-date. The four capture copies parsed only the first, so a provider answering

```
Retry-After: Wed, 21 Oct 2026 07:28:00 GMT
```

was read as having said **nothing at all**. The connector fell back to its own backoff and came back when *it* guessed rather than when the provider asked. The whole point of honouring the header is not to guess, and coming back early on a rate limit is how a throttle becomes a ban.

The geocoder's copy was the only complete one. Fixing the other four meant editing four files — which is precisely the arithmetic that leaves one of them unedited, and why the issue filed it as a duplication finding rather than a bug.

## Where it lives

`internal/shared/kernel/retryafter`, stdlib-only, below every tier that talks to a provider. The callers are four capture connectors, a platform package and a module — and a module never imports a sibling (ADR-0054), so kernel is the only home that serves all six.

`HeaderAt(value, now)` is the whole rule with the **one reading of the clock lifted out**, so the date form is stated exactly in a test rather than asserted as a range around a real one. `Header` and `Of` are that, at `time.Now()`.

Two decisions worth naming, both in the code:

- **Zero and negative delta-seconds are "nothing to wait for", not "wait none."** A caller that treated `Retry-After: 0` as an instruction would hammer a provider that has just asked it to slow down.
- **The date is measured against OUR clock, not the response's `Date` header.** A clock skewed against the provider's makes the wait wrong in whichever direction the skew runs, and ours is the clock the sleep is actually measured against — so a wait computed from it lands where it was meant to, skew or no skew.

## What did NOT move

Telegram's `parameters.retry_after`. That is Telegram's value in Telegram's envelope, not this subject; it stays where it is, with the header underneath it as the fallback it always was, and its five-minute cap is its own decision about its own retry ladder. The gate's header says so, because a reader who greps `retry_after` should find out why it is exempt without having to ask.

## The gate

`TestTheRetryAfterReadingIsSpelledOnce` sweeps the negative space: any file outside `kernel/retryafter` that reads the header fails. The subject is `<expr>.Get("Retry-After")` read **off the syntax**, so the header name inside a comment is not a finding — this tree names it in nine prose comments, and a gate that reported those would be waived away in a week. Casing is the wire's business, so `retry-after` matches too.

Mutation-verified: a probe reading the header in `graph/transport.go` fails the gate, naming the file. Near-miss fixtures pin the four ways it must stay silent — the name in a comment, a different header, a provider's own envelope field, and a `Get` with the wrong arity.

Stated honestly in the header: it cannot see a read through a variable holding the name, and it is not asked to see a provider-specific interval carried somewhere else entirely.

## Verification

`make check-backend` green. 92 lines deleted, 18 added.

Part of #2442 — two more to follow (`platform/backoff` for the identical jittered pair, and the `storekit` cursor envelope).

---
## Summary by cubic
Unifies Retry-After handling across capture and platform and fixes ignored HTTP-date values that led to early retries. Previously we parsed only delta-seconds; now we parse both forms and centralize the logic to schedule correct waits under rate limits.

**Bug Fixes**
- Parses HTTP-date and delta-seconds per RFC 9110 so providers’ dates are honored.
- Treats zero or negative values as “no wait,” not “retry immediately.”
- Measures date waits against our clock instead of the response Date header.

**Refactors**
- Centralizes parsing in `internal/shared/kernel/retryafter` via `retryafter.Of`.
- Replaces inline reads in Gmail, Google connector base, Graph, OAuth flow, and Geocode; Telegram keeps its envelope value with the header as a capped fallback.
- Adds gate `backend/oneretryafter_test.go` to block header reads outside `kernel/retryafter` and updates the gate inventory.

<sup>Written for commit 3989b207e978ff7f190e6cfc479d601aa0be8bee. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/2684?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added consistent handling of `Retry-After` responses across supported integrations.
  * Improved support for both delay values and HTTP date formats.
  * Preserved provider-specific retry delays and capped Telegram delays at five minutes.

* **Bug Fixes**
  * Improved behavior for invalid, missing, expired, or non-positive retry instructions.

* **Tests**
  * Added comprehensive coverage for retry-delay parsing and response handling.

* **Documentation**
  * Updated the gate inventory to reflect the new retry handling coverage.

### DIFF
--- a/backend/internal/modules/capture/gmail/client.go
+++ b/backend/internal/modules/capture/gmail/client.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/gradionhq/margince/backend/internal/modules/capture/googleconn"
 	"github.com/gradionhq/margince/backend/internal/modules/capture/oauthflow"
+	"github.com/gradionhq/margince/backend/internal/shared/kernel/retryafter"
 	"github.com/gradionhq/margince/backend/internal/shared/ports/connector"
 )
 
@@ -361,9 +362,9 @@ func (a *httpAPI) Watch(ctx context.Context, accessToken, topic string) (string,
 func classifyStatus(resp *http.Response, op string, body []byte) error {
 	switch {
 	case resp.StatusCode == http.StatusTooManyRequests:
-		return &connector.RateLimitedError{RetryAfter: retryAfter(resp)}
+		return &connector.RateLimitedError{RetryAfter: retryafter.Of(resp)}
 	case resp.StatusCode == http.StatusForbidden && googleconn.RateLimitBody(body):
-		return &connector.RateLimitedError{RetryAfter: retryAfter(resp)}
+		return &connector.RateLimitedError{RetryAfter: retryafter.Of(resp)}
 	case resp.StatusCode == http.StatusUnauthorized || resp.StatusCode == http.StatusForbidden:
 		return &connector.ProviderError{
 			Op: op, Status: resp.StatusCode, Reason: googleconn.Reason(body), Class: ErrAuthRejected,
@@ -374,18 +375,6 @@ func classifyStatus(resp *http.Response, op string, body []byte) error {
 		}
 	}
 	return nil
-}
-
-// retryAfter parses the provider's Retry-After (delta-seconds form; Google
-// does not send HTTP-dates here). Zero when absent — the caller's own backoff
-// takes over.
-func retryAfter(resp *http.Response) time.Duration {
-	if s := resp.Header.Get("Retry-After"); s != "" {
-		if secs, err := strconv.Atoi(s); err == nil && secs > 0 {
-			return time.Duration(secs) * time.Second
-		}
-	}
-	return 0
 }
 
 // Read caps for the Gmail JSON responses. The metadata calls (profile, id
--- a/backend/internal/modules/capture/googleconn/googleconn.go
+++ b/backend/internal/modules/capture/googleconn/googleconn.go
@@ -20,12 +20,12 @@ import (
 	"io"
 	"net/http"
 	"net/url"
-	"strconv"
 	"strings"
 	"time"
 
 	"github.com/gradionhq/margince/backend/internal/modules/capture/oauthflow"
 	"github.com/gradionhq/margince/backend/internal/shared/kernel/principal"
+	"github.com/gradionhq/margince/backend/internal/shared/kernel/retryafter"
 	"github.com/gradionhq/margince/backend/internal/shared/ports/connector"
 	"github.com/gradionhq/margince/backend/internal/shared/ports/datasource"
 	"github.com/gradionhq/margince/backend/internal/shared/ports/mcp"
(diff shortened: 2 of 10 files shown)


tinnhlt-gra:
## Review pass — Codex, Fable and CodeRabbit

**A claim of mine was false, and it is the kind this audit exists to find.**

The comment said measuring the HTTP-date against our own clock made the wait right "skew or no skew". It does not. It is wrong by exactly the skew, and wrong in the direction that costs: a clock running five minutes fast comes back **five minutes early** on a rate limit, which is how a throttle becomes a ban. Worse — the four capture readers this replaced rejected the date form outright and fell back to their own ladders, so consolidating them introduced skew sensitivity where there had been none.

It measures against the response's **own `Date` header** now, when it carries one. Both instants then come from the provider, so the interval between them is what it meant however far apart the two clocks are. Without a `Date` header ours is the only reference available, and the docblock says that rather than claiming more.

Mutation-verified: pointing it back at the local clock turns a ten-minute wait into **1364 hours**.

## A delta-seconds count too large for a Duration wrapped negative

`time.Duration(seconds) * time.Second` overflows above ~9.2e9, and every caller reads a non-positive wait as *"the provider said nothing"* — so the most extreme instruction a provider can give became **no instruction at all**, and the dispatcher fell back to its own guess. All four copies this replaced had the identical hole; consolidating was the moment to close it.

Clamped to the longest wait a `Duration` can hold rather than refused, because the provider did say something and the longest expressible wait is nearer to what it meant than our fallback. Mutation-verified: without the clamp, `Retry-After: 9223372037` returns **−2562047h**.

## Census blindness, declared rather than fixed

Both reviewers named respellings the subject cannot see: `resp.Header.Values("Retry-After")`, the canonical-key map index `resp.Header["Retry-After"]`, and a `const` holding the header name. Real, and the docblock now admits the shape rather than only the one case it originally named.

Widening the predicate to any string equal to the header name would report the **nine prose comments** in this tree that mention it, and a gate that fires on a comment is one nobody reads twice. The narrow syntactic subject is the trade, made deliberately and written down.

## What review confirmed rather than found

`HeaderAt` accepts nothing harmful the four old copies rejected — the only new tolerance is surrounding whitespace, which `net/textproto` already strips on a real response. Telegram's envelope-first precedence and its five-minute cap are exactly preserved on both paths, a nil response cannot panic, and the geocoder's behaviour is unchanged.

`make check-backend` green.

PR #2699, opened 2026-08-25
fix(design-system): the accent as text lifts in dark, and the axe sweep stops guessing which palette it is in

tinnhlt-gra:
## The lane was not flaky. It was two palettes.

The `uat` axe sweep read green locally and red on CI within one hour, on one tree. Three different answers were seen for one screen: 12 violations, 1 violation, 0.

`frontend/playwright.config.ts` pins `locale: "de-DE"` and `timezoneId: "Europe/Berlin"`, each with a comment saying that an unpinned environment input makes the suite depend on where the runner sits. **`colorScheme` is the same class of input and was not in that list.** Every run measured whichever palette the machine happened to report, and the two palettes do not have the same contrast.

Reproduced on main's tip, same spec, same tree, only the scheme emulated:

```
PROBE[light] violations(0)=[]
PROBE[dark]  violations(1)=["color-contrast: .lt-vtab.on"]
```

## Pinning alone would have deleted the defect rather than found it

Sweeping all 18 `CORE_SCREENS` under dark surfaced **8 live findings across 6 screens** — every one of them accent-coloured text on an accent-tinted fill:

| Screen | Finding |
|---|---|
| companies, contacts | `.lt-vtab.on` |
| home | `.glance-count-flat` |
| settings | `.badge` |
| settings/data-model | `.stage-row > .badge-accent.badge` ×3 |
| settings/people | `.panel-accent > .panel-head > .panel-head-text > h2` |

One cause. `--accent` lightens to `#16a34a` in dark, and that tone reading as **text** on the `--accentLight` tint derived from it measures **4.03:1** against a 4.5:1 requirement.

## The fix follows the pattern already in the file

The fill cannot move — `#16a34a` is the ADR-0040 mandate — so the text does. `--accentText` splits the accent-as-text from the accent-as-fill, which is exactly the shape `--tealText` and the `--mono*Text` pairs already have in `tokens.css`, for the same stated reason: *lift the text without shifting the fill*.

Its dark value is `--mono0Text`'s own `#4ade80`, so emerald text on an emerald tint reads as one colour across the product rather than as a second green. **Light keeps `--accent`'s value, so the light theme renders identically to before at all seventy sites.**

### Seventy sites, not six

All 70 `color: var(--accent)` declarations across 31 files move to the new token — not just the six that happened to fail on a swept screen. A screen outside `CORE_SCREENS` has the same defect and no test to report it, and leaving two spellings of "the accent as text" means the next author picks whichever one they grep first.

## Dark is now gated, not sampled

A second `describe` sweeps every `CORE_SCREENS` entry under `colorScheme: "dark"`. It is a second block rather than a tag on the first because the palettes are not the same subject: a screen can pass in one and fail in the other with no markup between them different.

## Verification

- `make frontend-e2e` — **exit 0**, 112 passed (was 94; +18 dark sweeps), 0 failed
- `make check-fe` — **exit 0**, 382 files / 4797 tests, plus 2 files / 36 tests on the extension surface
- Re-ran after rebasing onto current main, and re-checked that no new `color: var(--accent)` site arrived in the meantime — 0

**Mutation-verified.** Reverting the dark `--accentText` to `#16a34a` (confirmed applied by grep) turns red exactly the six screens measured beforehand, and nothing else:

```
✘ no AA violations on #/home in dark
✘ no AA violations on #/contacts in dark
✘ no AA violations on #/companies in dark
✘ no AA violations on #/settings in dark
✘ no AA violations on #/settings/data-model in dark
✘ no AA violations on #/settings/people in dark
6 failed | 106 passed
```

---
## Summary by cubic
Fixes dark‑mode contrast failures and stabilizes the axe sweep. Accent-colored text now uses a dedicated `--accentText` token (lifted in dark), and Playwright pins `colorScheme: "light"` with an explicit dark sweep to stop sampling palettes at random.

- Bug Fixes
  - Introduces `--accentText` (light equals `--accent`; dark `#4ade80`) to meet AA contrast on `--accentLight`.
  - Migrates ~70 `color: var(--accent)` text sites to `var(--accentText)`; light visuals stay the same; `--accent` remains the fill token.

- Tests
  - Pins `colorScheme: "light"` in `frontend/playwright.config.ts` and adds a full dark sweep of all `CORE_SCREENS`.
  - Removes palette-induced flakiness in axe and expands coverage (+18 dark sweeps).

<sup>Written for commit 329e1257b3beeb665bd5f527e977df2279dee90d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/2699?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Accessibility**
  - Improved dark-mode accent text contrast across the application to support WCAG 2.2 AA requirements.
  - Updated links, labels, controls, icons, selections, and navigation states for clearer dark-theme readability.

- **Bug Fixes**
  - Added automated accessibility checks for every primary screen in dark mode.
  - Standardized test runs to use the light theme unless dark-mode coverage is explicitly enabled.

### DIFF
--- a/frontend/e2e/ac.spec.ts
+++ b/frontend/e2e/ac.spec.ts
@@ -1231,6 +1231,32 @@ const ADDRESSED_VIEWS = [
   "reports/forecast",
 ];
 
+// The SAME sweep under the dark palette, which the suite measured by accident
+// for as long as the scheme went unpinned: a run inherited whichever palette
+// the machine reported, so one afternoon read light locally and dark on CI and
+// the lane presented as flaky. It was two palettes.
+//
+// It is a second block rather than a second tag on the block above because the
+// palettes are not the same subject. tokens.css publishes dark as its own set
+// of values — the accent lightens to the ADR-0040 mandate and the surfaces
+// darken — so a screen can pass in one and fail in the other with no markup
+// between them different. Eight findings across six screens were live here
+// when this block was written, every one of them accent-coloured text on an
+// accent-tinted fill, and none of them visible to the light sweep.
+test.describe("B-EP09.21: WCAG 2.2 AA (axe), dark palette", () => {
+  test.use({ colorScheme: "dark" });
+
+  for (const screen of CORE_SCREENS) {
+    test(`no AA violations on #/${screen} in dark`, async ({ page }) => {
+      await page.goto(`/#/${screen}`);
+      await page.waitForLoadState("networkidle");
+      await expectShellRendered(page);
+      await settleAnimations(page);
+      await expectNoAaViolations(page, `${screen} (dark)`);
+    });
+  }
+});
+
 test.describe("B-EP09.21: WCAG 2.2 AA (axe)", () => {
   for (const screen of CORE_SCREENS) {
     test(`no AA violations on #/${screen}`, async ({ page }) => {
--- a/frontend/playwright.config.ts
+++ b/frontend/playwright.config.ts
@@ -48,6 +48,16 @@ export default defineConfig({
     // so the record-clock and viewer-clock surfaces agree here and a test that
     // means to tell them apart has to say so itself.
     timezoneId: "Europe/Berlin",
+    // Pin the PALETTE for the same reason as the locale and the zone above:
+    // tokens.css publishes a light and a dark set behind
+    // `prefers-color-scheme`, so an unpinned scheme measures whichever one the
+    // runner happens to report — and the two do not have the same contrast.
+    // The axe sweep read light here and dark on CI within one hour, which
+    // presented as a flaky lane and was two palettes being sampled at random.
+    // Light is the default the product is built to (§2: "Build light-first");
+    // dark is swept explicitly by its own block in ac.spec.ts, so pinning here
+    // narrows what a test measures rather than what the suite covers.
+    colorScheme: "light",
     // the SW would compete with the network-edge seed mocks
     serviceWorkers: "block",
   },
--- a/frontend/src/app/account.css
+++ b/frontend/src/app/account.css
@@ -269,7 +269,7 @@
   align-items: center;
   justify-content: center;
   width: 14px;
-  color: var(--accent);
+  color: var(--accentText);
 }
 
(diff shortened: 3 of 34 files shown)


tinnhlt-gra:
Rebased onto `main` (was CONFLICTING/DIRTY).

**The conflict was in `frontend/e2e/ac.spec.ts`, and it was mine.** #2722 renamed the axe settle helper `animationsSettled` → `settleAnimations`; this branch's new dark-palette block calls it by the old name. Both sides were additions at the same point, so I kept both and renamed the call. Nothing was dropped: `ADDRESSED_VIEWS` and the dark `test.describe` are both present.

**One thing I checked rather than assumed.** `settleAnimations` now calls `page.emulateMedia({ reducedMotion: "reduce" })`. If that reset `colorScheme`, this PR's dark sweep would silently measure the *light* palette and prove nothing — a green lane covering nothing. Probed it directly inside a `colorScheme: "dark"` block:

```
before={"prefersDark":true,"reduced":false,"bgPage":"#0e1b16"}
after ={"prefersDark":true,"reduced":true, "bgPage":"#0e1b16"}
```

`prefersDark` holds and `--bgPage` stays the dark token while `reduced` flips. Playwright leaves unspecified `emulateMedia` keys alone, so the dark sweep is genuine.

Full suite on the rebased head: **127 passed / 14 skipped** (141 collected, up from 123 — the new dark screens).

Note CI here may still show the backend gate red for reasons outside this diff: main is currently failing `TestEveryUniquenessClaimNamesItsGateOrIsRegisteredDebt` and `TestTheIdleBaseIsSpelledOnce` from #2740, and `dco` on an unsigned commit in #2728. Neither is this PR's doing.

PR #2746, opened 2026-08-26
fix(migrations): the local-day replay holds its ids in what the component may say

tinnhlt-gra:
main is red on `backend gate (main)` — [run 32916607681](https://github.com/margince/margince/actions/runs/32916607681). The failure is arch-lint:

```
Component migrations shouldn't depend on github.com/google/uuid
in backend/migrations/briefrunlocalday_integration_test.go:27
```

`.go-arch-lint.yml` gives the component `canUse: [pgx]` and nothing else.

## The fix

The five fixture ids were never used as a *type* — they are opaque values handed to pgx and scanned back, and pgx binds a string to a `uuid` column and scans one back. So they are `string` constants now, and `mustParseUUID` goes with them.

That is also what this package already does elsewhere: `schema_integration_test.go` leaves generation to `uuidv7()` and `gen_random_uuid()` and never names a Go uuid package. The `ARRAY[]::uuid[]` cast stays — that is SQL text, not a Go dependency.

## The alternative I did not take

Adding `uuid` to the component's `canUse` list would also have made the gate green. It widens an architecture rule to buy one test's convenience, and the rule is what keeps this package compiling against a schema rather than against the application — the exception would be permanent and the convenience momentary.

## Verification

- `go-arch-lint check` → **OK - No warnings found**.
- Proved that run actually reads this file rather than skipping the build-tagged package: re-introduced the import on top of the fix and arch-lint reported `total notices: 1`. Restored and re-verified clean.
- `go vet -tags integration ./migrations/` clean; `gofmt` clean.

The test's own assertions are untouched — this changes how the ids are spelled, not what the replay proves.

## Not fixed here

main is also red on `dco (main)`, `uat (main)`, `integration shard (4/6)`, `integration shard (5/6)` and `integration (RLS + erasure + HTTP e2e)`. Separate causes, separate diffs:

- **`uat`** — #2743, open.
- **`dco`** — commit `2736937c0` (#2728) has no `Signed-off-by`; only its author can remediate that.
- **shard 4/6** — `read_brief({}): not found` in the tool-schema conformance census. `read_brief` is new and its fixture is unseeded.
- **shard 5/6** — `TestConcurrentCallsOfOneTaskAllReachTheProjection`.

## Summary by CodeRabbit

* **Tests**
  * Updated integration test fixtures to use string-formatted identifiers.
  * Maintained compatibility with PostgreSQL UUID storage and retrieval.

### DIFF
--- a/backend/migrations/briefrunlocalday_integration_test.go
+++ b/backend/migrations/briefrunlocalday_integration_test.go
@@ -24,7 +24,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/google/uuid"
 	"github.com/jackc/pgx/v5"
 )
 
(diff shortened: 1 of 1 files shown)


tinnhlt-gra:
Superseded — the fix landed on main by another route.

#2748 (`b11d5d179`) removed the `github.com/google/uuid` import from `briefrunlocalday_integration_test.go` while this PR was open, which is the same change this made. Verified at `origin/main` (792f0a455):

```
go-arch-lint check
OK - No warnings found
```

and main-health's own run on that commit reports `backend gate (main): success`.

Nothing here is still needed. Closing rather than rebasing an empty diff.

Worth recording why this sat open so long: it was deadlocked against #2750 — each PR was red only on the other's subject, so neither could go green alone. We were arranging a carrier branch to break that when it turned out main had already fixed both halves underneath us. **A deadlock that dissolves looks exactly like one you solved**, so I checked the base directly rather than inferring it from either PR going green.

read_brief, the one red in that knot nobody else was working on, is now #2753.

PR #2750, opened 2026-08-26
fix(deals,attention,migrations): one idle rule, four claims that name what holds them, and the arch-lint finding they were deadlocked with

tinnhlt-gra:
## #2740 left three gates red on main

Two of them are bookkeeping. **The first is a real duplication defect** — the
kind this tree's gates exist to catch, caught working.

### A second answer to "how long has this deal been quiet"

`internal/compose/attentionlanesseam.go` grew its own idle-base computation:

```go
since := deal.CreatedAt
if deal.LastActivityAt != nil {
    since = *deal.LastActivityAt
}
```

under a doc comment saying it counts *"from the same base the idle rule itself
measures from"* — an agreement nothing held. `TestTheIdleBaseIsSpelledOnce`
named it exactly:

```
internal/compose/attentionlanesseam.go holds a site this gate must judge
but lies outside every root (internal/shared/kernel/idlebase)
```

The number a reader is shown and the number the rule decides on are the **same
fact**. A copy lets the morning queue say "quiet 19 days" about a deal
`IsQuietFor` does not think is quiet at all — and the two agree in every test
written the day the copy is made. It now calls `idlebase.Since`.

**Routed, not ratified.** Adding the file to the scope's `Exempt` set would also
have gone green. That is the "a gate's exception is its attack surface" trade,
and the exception outlives the convenience that bought it.

### The two `formulas.go` claims are now held

> *"the one rule is spelled once"* — `QuietThresholdDays`
> *"Every rule bar the number is identical, so the two cannot drift"* — `IsQuietFor`

Both true, and both resting on one structural fact: the **threshold-named entry
points carry no rule of their own**. So that is what the new gate asserts, on
both surfaces:

| entry point | must be a single call to |
|---|---|
| `IsStalled` | `IsQuietFor` |
| `StalledSQL` | `QuietSQL` |

Delegation rather than text, so a body that grows a second condition fails
whatever it is spelled like.

**Mutation-tested**, each mutant grepped to prove it landed:

| mutant | result |
|---|---|
| inline a status+idle predicate into `IsStalled` | FAIL — *"must be a single call to it; its body returns `""`"* |
| gate lists a subject the file no longer declares | FAIL — *"this gate judged nothing about it, which is indistinguishable from it delegating correctly"* |

The second arm exists because a rename **empties** a gate rather than failing it,
which reads exactly like a clean tree.

### The other two claims

- **`TestAWithheldLaneIsNamedOnce`** — self-bound. The test counts the namings,
  which is precisely the claim its doc comment makes.
- **`renderEach`** — loses its claim instead. Nothing held *"the loop between
  them is spelled once"*, so the sentence keeps the reason and drops the
  assertion.

I deliberately did **not** register any of the four. The register is closed to
new entries; registering is the cheap fix that launders the debt and makes the
number stop meaning anything. Each claim got a test or lost the sentence.

### Two meta-gates caught the new gate on the way through

`delegations` read as a waiver map until declared `gatekit:fixture`, and the
fixture count is itself pinned — so adding one is a line somebody agrees to.
Both are the same discipline the gate itself applies, applied to it.

## How #2740 got three gates past everyone

The backend gate was **already red** on #2725's arch-lint finding. A lane that is
already red hides the next break in the same lane: nobody re-reads the log of a
job they already know is failing, and a stale red reports *identically* whether
it has one cause or four. Worse than a skip, which at least announces itself.

Worth a follow-up on `main-health`: key the filed issue on the set of failing
**test names** rather than the lane, so a new name appearing under an
already-red lane opens a new issue instead of being absorbed into the old one.

## Verification

- `make check-backend` — green except `arch-lint`, which is #2746's subject
  (`migrations` → `github.com/google/uuid`, from #2725) and fails identically at
  `origin/main`. Not this diff.
- `craft static --strict`, diff-scoped: **0 blocker / 0 major / 0 minor**.
- Root gate suite green; `deals` and `compose/attention` green.

> **Note on sequencing:** this PR and #2746 are mutually blocking — #2746 is red
> only on the gates fixed here, and this is red only on the arch-lint finding
> fixed there. One of the two needs to carry both commits. Coordinating.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved consistency when calculating how long deal activity has been idle.
  - Updated expected annotation results to reflect current behavior.

- **Tests**
  - Expanded coverage for quiet and stalled deal thresholds.
  - Added safeguards for consistent threshold evaluation.
  - Confirmed withheld lanes are named only once.
  - Improved fixture compatibility for database identifiers.

- **Documentation**
  - Clarified guidance for rendering lane data and maintaining threshold-related checks.

### DIFF
--- a/backend/gatecensus_test.go
+++ b/backend/gatecensus_test.go
@@ -52,7 +52,7 @@ const (
 	// Permitting the marker without counting it would let it spread quietly and
 	// reopen the class these rules close; pinned, every new one moves a number a
 	// reviewer sees.
-	wantFixtureAnnotations = 24
+	wantFixtureAnnotations = 25
 )
 
 // censusDecl is one package-level declaration this census governs — a map from
--- a/backend/internal/compose/attention/feed_test.go
+++ b/backend/internal/compose/attention/feed_test.go
@@ -972,6 +972,8 @@ func TestEveryLaneIsReadOncePerFeed(t *testing.T) {
 
 // A withheld lane is named ONCE. Naming it twice is what a duplicate read looks
 // like on the wire, and a client rendering the list would say it twice.
+//
+// Held by: TestAWithheldLaneIsNamedOnce (backend/internal/compose/attention/feed_test.go) — this test, which counts the namings.
 func TestAWithheldLaneIsNamedOnce(t *testing.T) {
 	svc := NewService(
 		stubApprovals{}, stubDuplicates{}, &stubTasks{}, stubReceipts{}, stubBriefing{},
--- a/backend/internal/compose/attention/optionallanes.go
+++ b/backend/internal/compose/attention/optionallanes.go
@@ -97,8 +97,8 @@ func (s *Service) optionalLanes(
 	}
 }
 
-// renderEach turns one producer's rows into wire items. The lanes differ in
-// what they read and how a row is drawn; the loop between them is spelled once.
+// renderEach turns one producer's rows into wire items, so a lane supplies only
+// what differs: the rows it read and how one of them is drawn.
 //
 // Never nil: the empty slice rather than a null the contract promised was a list.
 func renderEach[T any](rows []T, draw func(T) crmcontracts.AttentionItem) []crmcontracts.AttentionItem {
--- a/backend/internal/compose/attentionlanesseam.go
+++ b/backend/internal/compose/attentionlanesseam.go
@@ -24,6 +24,7 @@ import (
 	"github.com/gradionhq/margince/backend/internal/modules/agents"
 	"github.com/gradionhq/margince/backend/internal/modules/people"
 	"github.com/gradionhq/margince/backend/internal/shared/apperrors"
+	"github.com/gradionhq/margince/backend/internal/shared/kernel/idlebase"
 	"github.com/gradionhq/margince/backend/internal/shared/kernel/ids"
 	"github.com/gradionhq/margince/backend/internal/shared/kernel/principal"
 )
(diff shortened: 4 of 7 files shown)


tinnhlt-gra:
### Now carries #2746's commit too — the two PRs were mutually blocking

`git cherry-pick -x 9ebed5e0a`, authored by the original author and with the
source commit recorded in the trailer. #2746 is superseded by this.

The trap was worth naming: **two PRs, each red *only* on the other's subject**,
both gated on a required check. Nothing was wrong with either diff. It is the
merge-order cousin of the union problem — except a union break is untested,
while this one is *untestable* from either side. The tell is "my PR is red on
something my diff does not touch" showing up on both PRs at once.

`make check-backend` is now **fully green** on the combined branch.

### The arch-lint green is falsifiable

`go-arch-lint check` printing `OK - No warnings found` does not by itself prove
it read a **build-tagged** file — the integration-tagged package could simply
not have been compiled. Re-introduced the import on top of the fix:

```
Component migrations shouldn't depend on github.com/google/uuid
in backend/migrations/briefrunlocalday_integration_test.go:27
total notices: 1
```

then restored. So the gate is reading the file it is credited with judging.

tinnhlt-gra:
Superseded — these fixes landed on main independently while this was being assembled.

Verified at `origin/main` 792f0a455: `TestEveryUniquenessClaimNamesItsGateOrIsRegisteredDebt` and `TestTheIdleBaseIsSpelledOnce` both pass, arch-lint passes, and `formulas.go` carries `// Held by: TestIsQuietForIsTheSameRuleAtADifferentWindow`. The claims were **held with a test rather than registered**, which is the same call made here — nothing was laundered into the register, and it is still 638.

The one thing in this branch not on main is the read_brief conformance fixture, which belongs to its author and ships as its own PR now that arch-lint and the claim gates no longer block it.

Two notes worth keeping, since neither of us was careless and it still cost real time:

1. **A filed finding is a snapshot.** These four claims were genuinely broken when re-verified at `origin/main` before this work started. They were fixed underneath it inside the hour. With several agents on one tree the verification expires, and the cheapest re-check turned out to be the rebase itself — it said in one command what a grep would not have.
2. **A deadlock that dissolves is indistinguishable from one you solved.** Pushing twenty minutes earlier would have made the cherry-picks look like the thing that broke the knot.

PR #2753, opened 2026-08-26
fix(ci-tests): the conformance brief run is stamped with the day the read asks for

tinnhlt-gra:
main is red on `integration / integration shard (4/6)` — [main-health 32922529439](https://github.com/margince/margince/actions/runs/32922529439) on `792f0a455`:

```
--- FAIL: TestToolAnswersReachableWithoutApprovalSatisfyTheirSchemas/read_brief_{}
    toolschema_conformance_integration_test.go:227: read_brief({}): not found
```

**No commit broke this.** It is a time bomb, which is why blame and bisect find nothing.

## The mechanism

`brief_run` is keyed by LOCAL DAY — #2727's own migration. `briefseam.go` reads:

```go
run, err := engine.LatestRun(ctx, time.Now().UTC())
```

and `LatestRun` resolves that instant to a calendar day and selects `WHERE user_id = $1 AND local_day = $2`.

The lane's fixture is not missing — it already seeds through the real engine. But it pinned the assembly to a **fixed instant**:

```go
engine.SnapshotRun(ctx, time.Date(2026, 8, 8, 6, 0, 0, 0, time.UTC))
```

So the sweep asks for today's day and the only run on file is stamped a morning that has passed. The test certified an answer on 2026-08-08 and a not-found every day since — meaning the lane has been reporting on a tool whose schema it never actually read.

Two clocks in one fixture: a frozen seed and a wall-clock reader.

The part worth naming is that the fixed instant was chosen **for determinism** — the comment says so. The instinct was right; it became the defect the moment the read grew a day key underneath it. What the sweep needs is not a stable day but the *same* day at both ends.

## Why not fix the other end

Serving an empty brief for an unassembled one would break an existing gate and contradict the tool's own contract:

- `TestBriefReadServesTodaysRunAndNotAnOlderOne` asserts a day with no run answers not-found, because *"a stale ranking under today's date is the defect this filter exists to prevent"*.
- `tools_brief.go` argues separately that an empty `items` must not stand in for an unread queue: *"An agent reading `null` has to decide whether it means 'nothing is queued' or 'the queue was not read'."*

The not-found is correct and already held. The fixture is what was wrong.

## A bound I am not closing

The fixture and the reader now agree by both asking `time.Now().UTC()` — by asking the same question, not by sharing one answer. A reader that moves to an injected clock leaves this behind, silently and in exactly the same way. That is stated in the comment rather than papered over; a shared helper would mean a test reaching into the seam, which is a larger change than this red warrants. Credit to the review that raised it.

## Verification

`go vet -tags integration` clean, `gofmt` clean. The integration lane needs a real Postgres, so CI is the proof here — the failing test is in this PR's own lane.

## Also on main, not fixed here

`dco (main)` fails: commit `2736937c0` (#2728) carries no `Signed-off-by`. A sign-off is an attestation only its author can give, so no patch can clear it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated brief snapshot timestamps to use the current UTC time, ensuring daily brief lookups consistently return the correct current-day snapshot.
  * Improved validation of query vocabulary responses to help ensure consistent tool behavior.

### DIFF
--- a/backend/internal/compose/integration/toolschema_conformance_integration_test.go
+++ b/backend/internal/compose/integration/toolschema_conformance_integration_test.go
@@ -390,10 +390,20 @@ func TestTheConformanceCheckFailsAgainstAMisdeclaredSchema(t *testing.T) {
 func snapshotBriefRun(ctx context.Context, t *testing.T, e *Env) {
 	t.Helper()
 	engine := briefs.NewBriefEngine(e.Pool, people.NewStore(e.DB()))
-	// A fixed instant. The run only has to EXIST for read_brief to have
-	// something to re-read, and ranking against the wall clock would make what
-	// this lane certifies depend on the day it ran.
-	if _, err := engine.SnapshotRun(ctx, time.Date(2026, 8, 8, 6, 0, 0, 0, time.UTC)); err != nil {
+	// The reader's own instant, because brief_run is keyed by LOCAL DAY and
+	// read_brief asks for today's: briefseam.go reads `LatestRun(ctx,
+	// time.Now().UTC())`, which resolves a calendar day and matches
+	// `WHERE user_id = $1 AND local_day = $2`. A run stamped with any other day
+	// is a run that read cannot see, so a fixed instant here does not freeze
+	// what the lane certifies — it dates the fixture, and the sweep certifies a
+	// not-found from the morning after.
+	//
+	// Sharing the clock is what keeps the two ends honest; nothing here rests on
+	// WHICH day it is, only that one run exists under the day the read will ask
+	// for. The two agree by asking the same question, not by holding one answer
+	// between them — so a reader that moves to an injected clock leaves this
+	// behind, silently and the same way. Follow it here when it does.
+	if _, err := engine.SnapshotRun(ctx, time.Now().UTC()); err != nil {
 		t.Fatalf("assembling a brief run for the acting rep: %v", err)
 	}
 }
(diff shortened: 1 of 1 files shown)


tinnhlt-gra:
Second commit: the census's next finding, which the first one uncovered.

With `read_brief` answering, `TestToolAnswersReachableWithoutApprovalSatisfyTheirSchemas` still failed — but on the parent, not a subtest, and `read_brief_{}` itself now **PASSES**:

```
--- FAIL: TestToolAnswersReachableWithoutApprovalSatisfyTheirSchemas (0.71s)
    --- PASS: .../read_brief_{} (0.00s)
    toolschema_conformance_integration_test.go:217: describe_query_vocabulary is
      registered but never invoked here, so nothing holds its result to the
      schema or the envelope.
```

`describe_query_vocabulary` arrived with #2728 and was never added to the sweep. It takes no arguments, carries read scope and auto-execute, and needs no seam this lane declines to stand up — so it is a **call**, not a waiver. `unreachableInThisLane` is for handlers that cannot be reached without a provider the lane has no business composing; listing a tool there because nobody had got to it yet would spend the census's one exception on precisely the case it exists to catch.

Worth naming the pattern, because it is the same one that let #2740 put three gates past everyone earlier tonight: **a failing assertion hides the next failing assertion in the same test.** The census ran on every one of these runs and reported this every time — nobody could see it because the log was already going to say FAIL. Fixing one red is how you find the next, and "the test still fails" is not evidence the fix did not work.

Issue #2714, opened 2026-08-25
gate-inventory.md is a committed whole-tree manifest with no merge queue: every gate-adding PR breaks every other one


tinnhlt-gra:
`docs/reference/gate-inventory.md` is generated from every `//gate:kind` in the tree, carries per-kind **counts**, is **committed**, and is enforced by `TestTheGateInventoryPageIsCurrent`. With no merge queue, that combination makes any two gate-adding PRs mutually breaking even when their diffs share no file.

Observed repeatedly today across eight PRs from one author plus several from others.

## The mechanism

CI builds `refs/pull/N/merge` — the branch merged with main **as it stands when the job runs**. So:

1. Branch A regenerates the page against main at time T. Correct.
2. Main merges branch B at T+1, adding a gate.
3. Branch A's next CI run builds A + main-at-T+1, whose tree has B's gate and A's page. The drift gate fails.

Branch A did nothing wrong and has no green it can reach except by luck: rebasing at T+2 is only correct until the next gate lands. Today that window was as short as ten minutes.

## Three ways it goes wrong, all seen

**It breaks main, briefly and invisibly.** #2684 and #2650 each added a gate and merged eleven minutes apart. The first left main's page one short — `backend gate (main)` red, taking two integration jobs with it — and the second's regeneration silently repaired it. Had the second landed an hour later, main would have sat red for a reason no individual PR caused, and the repair diff looks like a no-op to anyone bisecting.

**A clean merge is not a current manifest.** If two branches touch different sections, git reports no conflict and the page is stale anyway. A rebase loop that regenerates only on conflict leaves the branch red.

**Regenerating from a stale base DELETES rows.** A session regenerating from an older tip produced a plausible one-line count diff that was also about to remove two gates that had landed in between. Destructive, and it reads as routine in either direction.

The same shape hits `backend/uniquenessclaims.txt`, whose `registeredDebt` ceiling is a count of the whole file: two PRs each retiring one claim leave both ceilings one too high.

## What would actually fix it

Roughly in order of cost:

1. **A merge queue** on main. Builds each PR against the tree it will actually land on, which is the general fix for this whole class and not only for this file.
2. **Drop the counts from the page.** The `## Parity (29)` headings are what collide; the rows themselves rarely do. A page without them still lists every gate and still drifts when a gate is added — but two PRs adding gates in different sections would no longer conflict.
[...]

PR #2769, opened 2026-08-26
feat(agents): a rep's standing authority is their own, and remembered

LarsGradion:
## What this adds

A rep can now hold **standing authority** for an overnight agent: a recorded
decision plus the passport that carries it. This is what the nightly fan-out
needs before it can run for anyone — today every scheduled occurrence is
enqueued with no passport and refused at execution time.

## Whose authority — the invariant, and the gate that now holds it

The grant **records** a decision and never confers one. What lets an agent act
is a passport, and there is exactly one production statement that mints one:

```sql
INSERT INTO passport (on_behalf_of, granted_by, ...)
VALUES ($1, $1, ...)   -- both from the session user
```

No admin-mint path exists anywhere in the tree. So a rep is only ever acted for
by a credential they minted themselves — structurally, not by a check.

**Nothing held that**, and the whole feature rests on it. A second `INSERT INTO
passport` taking a user id from a request body would compile, pass every test,
and quietly turn standing authority into something one person creates for
another. `backend/passportmint_test.go` now fails on either shape: a mint
outside the identity module, or a mint whose two user columns differ. Both
mutation-checked.

## Why the decision is its own row

The product asks **once**. A rep who declined has no passport, and neither has a
rep nobody asked — from the passport table alone the two are identical, and a
product that cannot tell them apart asks the declining rep again every night.

## Why there is no stored `lapsed` state

My first cut had one, and it was wrong. Whether a credential still works is the
**passport's** fact, and it changes when nothing writes to the grant table — a
passport expires at a moment nobody observes. Two sources for one fact disagree
the first time an expiry passes with no sweep running, and then every reader
holds a row claiming authority that stopped working hours ago.

So the state records the **answer**, which only the rep changes, and both reads
resolve liveness by joining the passport in the same statement. The rep's card
and the scheduler's enumeration cannot disagree about whether their authority
still works. Removing the state also removed a function with no caller.

## The read a rep gets takes no user id

`MyGrant` selects on the acting principal, so there is no argument by which a
handler could reach a colleague's decision — the bound is the signature rather
than a grant. A principal with no human behind it is refused, not answered:
answering would invite a caller to offer a machine the confirmation.

`LiveGrantsFor` is the one cross-rep read, ratified in `ungatedEntryPoints` with
what actually bounds it — it returns no credential and no record, only which
reps to run for, and each run authenticates that rep's own passport before
acting.

## Fixed along the way

`agents/doc.go` claimed the module owned no tables while `tableownership_test.go`
already assigned it `agent_run` and `runner_job`. Stale before this change; my
table makes it three.

Per-rep fan-out. `TriggerRef` carries no user segment, and both
`agent_run_trigger_unique` and `runner_job_trigger_unique` are keyed per-day for
the whole workspace. Fixing only the enqueue side would silently run at most one
seat, so the trigger ref and both constraints change together in the next PR.

## Verification

- `make check` green, `craft-static` 4/4, `rls-store-path`, `no-jurisdiction`.
- Migrations lane green; `head_catalog.txt` regenerated in this commit.
- 9 integration tests over real Postgres, seeding the passport through the real
  `POST /v1/passports` endpoint rather than inserting a row.
- Mutation-checked: a fan-out that trusts the stored state instead of the
  credential, a read that ignores the passport join, a second mint path, and a
  mint with differing user columns each fail a test.

## Summary by CodeRabbit

* **New Features**
  * Added standing-grant support for overnight agents, allowing decisions to be granted or declined per agent.
  * Granted authority is linked to valid passports, required credentials, and the correct owner.
  * Grant status reflects passport expiration or revocation.
  * Users can view only their own decisions, while overnight processing discovers eligible live grants.
  * Updated decisions replace previous responses while preserving the latest state.

* **Bug Fixes**
  * Prevented unauthorized, unasked, declined, expired, revoked, or incorrectly owned grants from being used.
  * Prevented non-human principals from recording standing decisions.
  * Improved audit tracking for grant and decision changes.


tinnhlt-gra:
This PR's `deterministic-gates` / `extension-reference` / `integration unit coverage` failures are all one test — `TestTheGateInventoryPageIsCurrent` — and it is a known structural collision, not something wrong with the diff's logic. Tracked as #2714.

I checked main before concluding that, because #2747 had just landed new gate arms and a whole-tree manifest is exactly what that would break: the gate **passes** at `origin/main` (verified at `eff498c77` and again at `9442e9b83`). So main is fine and this PR is the second of a pair.

The fix here is to regenerate `gate-inventory.md` and commit it in the same change.

Two properties worth knowing, because both make this read as something other than what it is:

1. **The colliding PRs can share no file in common.** The manifest counts the whole tree, so "our diffs do not overlap" is not evidence you are safe — which is the check most people would actually run before assuming they are.
2. **It self-heals if the other PR lands soon enough**, so a PR that was red on this and is green after an unrelated merge teaches the wrong lesson. That is why it presents as flake.

For anyone sequencing gate work behind this: the page renders from each gate's `//gate:kind <kind> <hardness>` line plus the opening line of its doc comment (`backend/gateinventory_test.go`). So it moves when a gate file is added or removed, when a `//gate:kind` line changes, or when a gate's first doc-comment line is edited — and not for changes inside a gate's body. Credit to @margince-next-31 for reading that derivation out; it narrows who is actually in the race considerably.

PR #2812, opened 2026-08-26
perf(check): the fitness gates run once, in parallel, and say where the time went

tinnhlt-gra:
## What this does

`make check` runs the fitness-gate package **twice** and the 30 root script gates **one at a time**. Neither is load-bearing. This PR stops both, and gives the gate a way to say where its time goes so the next person to optimize it starts from a measurement rather than a comment.

It also moves the 163 fitness gates out of the module root into `backend/gates/`.

## The two runtime fixes

**1. The fitness package ran twice.** `backend/Makefile`'s `test` target ran `go test ./...` — which includes the root package and `./internal/modules/ai/` — and then immediately re-ran both with `-count=1`. The uncached pass is the one that counts (both packages' subjects live *outside* the package, so Go's test cache cannot key them). The cached pass computed a result that was thrown away seconds later: **316 s of a 648 s `check-backend` in CI**, all of it that one package, and a third time in the `extension-reference` job.

The two packages are now named once, in `UNCACHED_TEST_PKGS`, and subtracted from the wildcard. The subtraction is **asserted, not assumed** — the recipe fails if it did not remove exactly as many packages as the list names, so a renamed package cannot leave the exclusion silently matching nothing:

```
go test: 143 cached packages + 2 uncached
```

**2. The 30 root script gates ran serially.** They are independent readers — each walks the tree and judges what it finds, or builds fixtures under `mktemp`; none writes into the tree. Nothing but the list itself ordered them. They now fan out at `-j4`:

| | before | after |
|---|---|---|
| 30 root script gates | **155 s** | **60 s** |

(8-core darwin, warm caches, all 30 green in both.)

## The move

`backend/` held 163 files, every one of them a `*_test.go` in `package backendarch`, and no production code at all. They are now `backend/gates/`, `package gates`.

Every one of those gates names its subject relative to the **module root** — `"internal/modules"`, `"api/crm.yaml"`, `"../docs/reference/gate-inventory.md"`, `"../extensions"`. 117 files carry such a literal, and many of those same strings are *import*-path fragments a gate compares against rather than file paths, so a mechanical rewrite would have corrupted the comparisons it did not change. Instead one `TestMain` restores the working directory — and **refuses to run** unless it lands somewhere with a `go.mod`, because a gate that walks `internal/modules` from the wrong directory finds nothing and reports the empty result as a clean tree.

**`docs/reference/gate-inventory.md` changes by two preamble lines and none of its 163 rows.** The page names a gate by its path relative to `gateDir`, so the files moved without the page moving. That matters because the page has no merge queue and its counts are tree-wide — a 163-row rewrite would collide with every other gate-adding PR in flight (#2714).

## Per-phase timing

Every `make check` now ends with a table:

```
  where the time went (4m18s total)
     60s   23.3%  root script gates (x30, -j4)
     ...
```

A phase that fails is never closed, so the table only ever describes a run that reached the end. The ledger is reset by whoever owns the run, so a stale `make check-go` cannot donate its seconds to the next full check.

## One gate widened, deliberately

`backend/gates/frontendlaneparity_test.go` reads `$(MAKE)` recipe lines to find the frontend lane's legs, and **refuses** a line containing `$(...)` because it cannot know what it expands to. It was right to refuse the new fan-out: `$(ROOT_SCRIPT_GATES)` genuinely hid 30 legs from it.

Rather than weaken the refusal, the parser now expands **literal** file-local make variables. A value built from another variable is still refused — half-expanding one would invent a leg list nobody wrote.

The widening is mutation-tested (8 mutants, each proved to have actually landed; 6 caught, 2 are inert-guard cases covered from the other side). That exercise caught a flaw in the first version of the new test: it asserted the fan-out *contains* the names in `ROOT_SCRIPT_GATES`, but `pulls` **is** the expansion of that variable — `expand(X) ⊇ X` holds however wrong the expansion is. It now checks each expanded word against an independent reading (does the Makefile declare a rule by that name?), with a floor that catches the read collapsing to empty, and the expected set is derived rather than pinned to a hardcoded `30`.

## Fixed along the way

- **`backend/.go-arch-lint.yml` declared `root: { in: . }`.** A component scoped to the module root matches nothing once the files move; `go-arch-lint` emitted 163 × "not attached to any component". Renamed to `gates` and repointed.
- **`backend/internal/platform/config/config.go` cited a gate that does not exist** — `backend/configseam_test.go`, which is nowhere in the tree and shows no sign of ever having been. The obligation it describes is real and held by `scripts/check-env-reads.sh`. Pre-existing; a comment naming a guarantee that nothing holds is worse than silence.
- `declaresATest` counted `TestMain` as a test, which would have demanded a `//gate:kind` tag for a file that judges nothing. Go runs `TestMain` *instead of* the tests; it is not one.

## Verification

- `make check` green.
- The gate suite went 12 → 4 → 0 failures across the move. **Every one of the 12 was loud** — the empty-corpus defences in that package all fired rather than passing over a tree they could no longer see.
- All 30 fanned-out gates confirmed `.PHONY`, so `rc=0` means every recipe ran rather than being skipped for a same-named file.

---
## Summary by cubic
`make check` now runs the fitness-gate package once instead of twice and the 30 root script gates in parallel at `-j4`, roughly halving the backend gate's wall time. Every green run ends with a per-phase timing table, and the 163 fitness gates moved from `backend/` to `backend/gates/`.

**Refactors**

- `UNCACHED_TEST_PKGS` names the two uncacheable packages once; the recipe asserts the wildcard subtraction removed exactly as many packages as the list names, so a renamed package cannot silently drop the uncached run.
- The script gates fan out through `ROOT_SCRIPT_GATES` at `-j4`; the recipe stays a bare `$(MAKE)` line so `frontendlaneparity_test.go` can still read the legs — it now expands literal file-local make variables and still refuses values built from other variables.
- One `TestMain` restores the module-root working directory and refuses to run unless the `go.mod` declares this module, so a gate that walks `internal/modules` from the wrong directory fails instead of reporting the empty result as clean.
- `gate-inventory.md` keeps all 163 rows because rows name a gate relative to `gateDir`, avoiding merge collisions with other gate-adding PRs.
- A `*_test.go` recreated at the module root or nested under `gates/` would run cached and appear on no page — a new census asserts both shapes stay empty.
- `AGENTS.md` now names the narrow lanes (`check-go`, `check-gates`, `check-fe`, `test-it DIR=<pkg>`) as the iterate loop, and states they are not a substitute for `make check`.

**Bug Fixes**

- `.go-arch-lint.yml` declared `root: { in: . }`, which matched nothing after the move; renamed to `gates`.
- Removed a comment citing `configseam_test.go`, a gate that does not exist.

<sup>Written for commit 7762ccfd6f905b22283169643ed176b23e0f807d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/2812?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Refactor**
  - Reorganized backend quality and architecture checks into a dedicated gate suite.
  - Centralized check execution with configurable parallel processing.
  - Added phase timing to clarify validation progress and duration.

- **Bug Fixes**
  - Improved uncached test handling so designated checks always run freshly.
  - Added validation that locally available frontend checks are covered by required CI checks.

- **Documentation**
  - Updated contributor, CI, architecture, and testing references to reflect the new gate locations and commands.


tinnhlt-gra:
## UAT

Run against this branch locally. No video attached — the browser on this machine is signed in to a different account, and a UAT comment on my own PR should not be attributed to a teammate.

### 1. Every package is tested exactly once

`make -n -C backend test` — the wildcard now excludes the two packages the last line re-runs uncached, so the pass that used to be computed and thrown away is gone:

```
go list ./gates ./internal/modules/ai/     <- the uncached list
go list ./...                              <- everything
go test $cached                            <- everything minus that list
go test ./tools/...
go test -count=1 ./gates ./internal/modules/ai/
```

And the run says what it covered, rather than leaving you to trust it:

```
go test: 143 cached packages + 2 uncached
```

### 2. The subtraction is asserted, not assumed

```
PKGS=[./gates ./internal/modules/ai/]  every=145  cached=143  want=2  got=2
assertion passes
no leak
```

Pointed at a package outside `./...`, it fails instead of silently matching nothing:

```
PKGS=[./gates ./tools/gen-composition]  every=145  cached=144  want=2  got=1
ASSERTION FIRES
```

### 3. All 30 fanned-out gates are `.PHONY`

Which is what makes `rc=0` mean *every recipe ran*, rather than *make skipped a goal that shares a name with a file*:

```
gates=30 missing-phony=0
```

The probe carries its own floor — with the fan-out line unreadable it refuses rather than reporting a green zero:

```
refusing: read 0 gate(s) out of `make -n check-backend`, expected at least 25 —
          the fan-out line moved or changed shape, so this probe was about to
          check nothing and report success.
```

(That floor exists because the probe *did* silently break once, when the phase-timer lines changed which line `make -n` prints first.)

### 4. Where the time went

Printed by every `make check`:

```
  where the time went (2m31s total)
     53s   35.1%  root script gates (x30, -j4)
      6s    4.0%  backend: build (host + windows)
     86s   57.0%  backend: vet/lint/arch-lint/test/test-extensions (-j4)
      5s    3.3%  backend: contract drift
      1s    0.7%  backend: composition reproducibility
```

The `-j4` analysis fan-out is now 57% of the gate. The two remaining fixes from the original measurement — `t.Parallel()` across this package, and one shared cached tree-walk instead of ~160 private `WalkDir` calls — both live inside that 86 s. Not attempted here: at least one gate in the package is reported order-dependent through a shared waiver set, which under parallelism would pass for the wrong reason.

### 5. The drop point the move opened, planted and caught

A `*_test.go` recreated at the module root compiles as its own package, runs **cached**, owes no `//gate:kind`, and reaches no page. Nothing fails. Same for a package nested under `gates/`, which the flat glob misses. Both planted live:

```
--- baseline, nothing planted ---
SURVIVED  clean tree (must survive)
--- a test file recreated at the module root ---
caught    backend/planted_root_test.go
          planted_root_test.go sits at the backend module root, where this census does not
          look and the `test` target does not run it uncached.
--- a package nested under gates/ ---
caught    backend/gates/nested/planted_test.go
          gates/nested/planted_test.go is nested inside gates/, and this census globs that
          directory flat — the file is a package of its own that no gate here judges.
--- restored ---
SURVIVED  clean tree again (must survive)
```

**This was not hypothetical for long.** Rebasing onto current main turned up `backend/seatfanout_test.go` — a real gate added at the module root tonight, at exactly the location this guard refuses. Without the guard it would have gone quiet under this change. It is moved into `gates/` here, and the inventory page did not move a line, because the census names rows relative to the directory.

tinnhlt-gra:
### CI status

The last red run was **mine, and it was one mistake with four faces**: I pushed a `git mv` that was staged but never committed, so the remote kept `backend/seatfanout_test.go` at the module root while my worktree had it under `gates/`. Every one of `deterministic-gates`, `extension-reference`, `integration unit coverage` and `integration` failed with the same line:

```
./seatfanout_test.go:181:17: undefined: exprText
```

— the stray file compiling as its own package, without the siblings that define its helpers. Fixed and force-pushed.

Two things worth recording rather than quietly fixing:

**The guard added in this PR caught it.** `TestNoGoSourceHidesWhereThisCensusCannotSeeIt` fired in CI with exactly the message it was written for:

```
seatfanout_test.go sits at the backend module root, where this census does not look
and the `test` target does not run it uncached.
```

`seatfanout_test.go` is a gate main added at the old location while this branch was in flight. That is the guard's whole premise — that `backend/` becomes a place a gate can sit and be invisible — arriving for real within hours, and then catching the author of the guard.

**One break hid the next.** In `deterministic-gates` the compile error aborted the job before it ever reached the gate suite, so that job never reported the *other* failure present in the same tree. Only `integration unit coverage` got far enough to show it.

### The one remaining failure is not from this branch

```
--- FAIL: TestEveryHandWrittenDocsPageFitsItsBudget
docs/explanation/data-enrichment-position.md is 499 lines, over the 450-line budget
This budget is not waivable for a new page
```

That page is **499 lines on `origin/main`** (`git show origin/main:docs/explanation/data-enrichment-position.md | wc -l`), added by 810b36bd4 (#2830). This branch does not touch it. `main-health` has not run past 6756ea85f, so nothing has reported it yet — it will fail every PR that rebases onto current main, not just this one.

Local `make check` on this branch is green apart from that single inherited failure.

PR #2814, opened 2026-08-26
fix(desktop): the start message names the admin the installation was created with

dungbanguyen:
## The problem

`announce()` carried the bootstrap admin address as a literal while `ensureConfig()` wrote it into `margince.yaml`. One fact, spelled twice.

`margince.yaml` is write-once and the organization is bootstrapped from it, so the first run is the only moment that address can be chosen. An installation created under any other address had **every launch** offering an account that does not exist:

```
  Sign in as  owner@margince.local     ← always, whatever the file says
```

A reader types what the console tells them and is refused. Nothing else on screen or in the folder contradicts it, so there is nothing to work back from.

## The change

Both sides now derive from one `defaultAdminEmail`, and `announce` reads the file.

The reader is deliberately narrow rather than a YAML parse. `desktop/launcher/go.mod` states the module is stdlib-only and outside `go.work` by design, and a parser dependency bought for one line of a start message would not pay for itself. It finds `email:` under `bootstrap_admin:` and stops.

Every shape it cannot follow — no file, no block, a commented-out or empty value — answers with the default. The address is informational, so reading it must cost a reader precision, never the launch.

## Tests

`TestAnnounceNamesTheConfiguredAdmin` asserts on the console text itself, because `configuredAdminEmail`'s own tests pass just as well while `announce` carries a literal. I verified that by reverting the one line and watching it fail. It captures `os.Stdout` because `say()` is `fmt.Printf` on purpose and the console has no seam to inject.

The helper's own tests cover a quoted value, a trailing comment, an `email:` belonging to a different block, and five fallback shapes.

## Verification

- `go build`, `go vet`, `gofmt`, `go test` on `desktop/launcher` — clean
- `craft static --strict --root ../../desktop` — 0 blocker, 0 major, 0 minor

`make check` ran green except `TestTheSweptCorpusIsEveryModuleThatCouldCarryAClaim`, which fails while walking the tree — my working tree carries symlinks it correctly refuses to follow, before it reaches any file in this diff. CI walks a clean checkout.

## AI disclosure

**Generated** — AI produced substantial portions, reviewed and owned by me.

## Summary by CodeRabbit

* **Bug Fixes**
  * The launcher now displays the configured administrator email from `margince.yaml`.
  * Custom administrator emails support quoted values, inline comments, and `#` characters in email addresses.
  * Nested administrator contact values are ignored unless directly configured.
  * Unsupported or invalid configuration safely falls back to `owner@margince.local`.
  * First-run configuration uses the shared default administrator email.
* **Tests**
  * Added coverage for email parsing, fallback behavior, nested configuration handling, and displayed administrator details.

### DIFF
--- a/desktop/launcher/layout.go
+++ b/desktop/launcher/layout.go
@@ -4,12 +4,14 @@
 package main
 
 import (
+	"bufio"
 	"crypto/rand"
 	"encoding/base64"
 	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 )
 
 // layout is the installation directory — everything is relative to it, so the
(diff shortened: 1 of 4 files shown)


dungbanguyen:
Thanks — all three were real, including one I had waved away when writing it. Addressed in 9661223a.

**`#` inside the address.** You're right that `#` is legal in the local part (RFC 5322 atext), so `admin#demo@demo.test` was being truncated to `admin` and announced — the exact failure this PR exists to prevent. A `#` now opens a comment only where YAML says it does, at the start of a line or after whitespace, and a quoted scalar is read to its closing quote instead.

**`email:` at any depth.** Fixed. The block's first field now fixes the column its own keys sit in, so `bootstrap_admin.contact.email` is not read as `bootstrap_admin.email`.

**The unasked `scanner.Err()`.** Fixed by removing the scanner: `os.ReadFile` replaces it, since the file is small and write-once, and one read has one error to answer where a scanner has a second that is easy to leave unasked. The deferred close goes with it, so there is no discarded error left in the function.

Both new shapes are held by tests that I verified fail against the previous reader — `a hash inside the address` and `only a nested email`.

dungbanguyen:
Both addressed in 958f79c4.

**Escaped values.** Agreed — `email: "admin@demo.test"` is legal YAML for an ordinary address, and returning the literal would announce an account nothing bootstrapped. I took the fallback rather than decoding: escapes are a parser's job and this is not a parser, so a double-quoted value containing a backslash is declined and the default stands. The default is a reader's own address; a half-decoded one is nobody's.

**The discarded copy error.** Fixed properly rather than waived. The error now travels with the captured text and fails the test on its own terms — a partial copy would otherwise have read as an absent line and failed the caller's assertion for a reason unrelated to what was printed.

Both are held by tests I verified fail without the change.

dungbanguyen:
Correct — `io.Copy` does not close its source, so `capture` leaked the read end on every call. Closed in the goroutine now, with the close error reaching the caller unless the copy already had something more informative to say. Pushed.

PR #2846, opened 2026-08-27
perf(ci): the shared build cache stops accumulating

tinnhlt-gra:
## What this fixes

The Go build cache every CI job restores had grown to **8.4 GB**. Restoring it was **6 m 35 s of a 13-minute merge-gate job**, and eleven Go jobs restore it on every run — so each gigabyte is paid eleven times, before any of them does work.

## It was growing, not merely large

Measured on the `plain` entry across nine hours, with no dependency change:

| time | size |
|---|---|
| 16:40 | 7.43 GB |
| 19:55 | 7.72 GB |
| 01:43 | 7.82 GB |
| 03:04 | **8.40 GB** |

Monotonic, roughly a gigabyte a day. A fixed working set would be flat.

That distinction was worth establishing rather than assuming, because `make check-backend` legitimately compiles the host build, a Windows cross-compile, two `vet` passes under different tags, the test binaries and golangci's own tree. Eight gigabytes could plausibly *have been* an honest working set, and the fix would then have been wrong.

## Why it grows

Go trims its own build cache — entries untouched for five days — but **only when its `trim.txt` marker is more than 24 hours old**. `cache-warm` runs every three hours and restores that marker along with everything else, so the marker is never old enough and **the trim can never fire**. The entry is restored, added to, and re-saved eight times a day with nothing ever removed.

## The fix: the writer stops reading

`cache-warm` exists to *produce* the entry the consumers restore. Restoring the previous entry before compiling is what turns "what this build needs" into "everything any build has ever needed".

Compiled from cold, what it saves is exactly the working set of the lane it mirrors — bounded by the build rather than by history. The cold compile costs only that job, which is scheduled and which nobody waits on.

The action gains a `restore` input so the writer can skip the restore while still getting its key from the one place the key is spelled — a writer that spelled its own key could write one no restore looks for. Both warm jobs set it; **all seven consumers are unchanged** and still restore by default.

## What I tried first, and why it was wrong

A `find -mtime +5 -delete` before the save, reproducing Go's age rule by hand.

Probed against a real 79 GB cache before shipping it: of **195,369 files, two** were older than five days. Go refreshes an entry's mtime whenever it is used, and the restore rewrites them anyway — so the predicate selects nothing. It would have deleted two files and logged an unchanged size while reading, in the diff and in the logs, exactly like a fix.

Deleting by age cannot work here. Not reading in the first place does not depend on mtime at all.

## Also

The action's note said *"a warm build cache for this repo measures ~550 MB"*. It was fifteen times that, and the sentence read as current fact — which is how the growth stayed invisible. Corrected, with the thing that now bounds it stated beside it.

## Verification

- `make check-backend` green on this branch, which is the half that judges workflow files (`ci-doc-parity`, `test-laneorder`, `check-image-pins`, `test-lanes`, and `frontendlaneparity_test.go`, which parses `ci.yml`). The diff is two YAML files and touches no Go or frontend source.
- Both files re-parsed as YAML after editing.
- Confirmed only the two warm jobs carry `restore: "false"`; the seven consumers were checked individually and still restore.

The effect is not verifiable from this PR alone — it shows up on the next scheduled `cache-warm` run, as the size of the entry it writes. Worth checking that number rather than assuming this worked.

---
## Summary by cubic
Stops the shared Go build cache from growing without bound, fixing the 8.4 GB entry that cost 6m35s to restore on the merge gate. The `cache-warm` writer now saves only the working set of its own build: it skips the action's restore and runs `go clean -cache` to clear the ~25 MB cache `setup-go` restores before compiling. Adds a `restore` input to the action (defaults to true), set to false on both warm jobs; all seven consumers are unchanged. Also corrects the stale comment that claimed a warm cache measured ~550 MB.

<sup>Written for commit 18fd51f638f8ffa4eb124aa1c2022b590b1a3ba4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/2846?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added an option to control whether the Go build cache is restored, while keeping cache keys and outputs available.
- **Improvements**
  - Updated cache documentation to reflect larger warm-cache capacity and automatic trimming.
  - Cache-warming workflows now rebuild from a clean cache before saving refreshed entries, helping maintain reliable and up-to-date build caches.

### DIFF
--- a/.github/actions/go-build-cache/action.yml
+++ b/.github/actions/go-build-cache/action.yml
@@ -6,8 +6,8 @@ description: >-
 # Why this exists: actions/setup-go caches ~/.cache/go-build under a key that
 # hashes only go.sum, so the entry is written ONCE and then every later run logs
 # "Cache hit occurred on the primary key ... not saving cache" and restores that
-# first, near-empty snapshot forever. A warm build cache for this repo measures
-# ~550 MB; the blob setup-go restores is ~25 MB. The result was that every Go
+# first, near-empty snapshot forever. A warm build cache for this repo is
+# multiple GB; the blob setup-go restores is ~25 MB. The result was that every Go
 # job in the pipeline — the integration shards, the merge gate, the
 # composed-build lane, the coverage pass, live-boot, govulncheck — compiled the
 # module from scratch on every run. setup-go keeps caching the MODULE cache
@@ -20,8 +20,26 @@ description: >-
 # last fallback drops the dependency hash entirely: after a dependency bump a
 # mostly-warm cache still beats a cold one, and the stale entries age out under
 # Go's own trimming rather than being served in error.
+#
+# "Age out under Go's own trimming" is the part that needs watching: Go only
+# trims when its `trim.txt` marker is over 24 hours old, and cache-warm both
+# restores that marker and runs every three hours, so the trim cannot fire on
+# its own. cache-warm therefore does it explicitly before saving. Without that
+# the entry grew about a gigabyte a day — it was 8.4 GB when this was found,
+# and every one of those gigabytes is restored by eleven jobs per run.
 
 inputs:
+  restore:
+    description: >-
+      Whether to actually restore. A CONSUMER wants the cache and leaves this
+      alone. The scheduled writer sets it to `false`: it exists to produce the
+      entry every consumer then restores, and an entry built on top of the
+      previous one accumulates instead of describing a build. Compiling from
+      cold makes what it saves exactly the working set of the lane it mirrors.
+      The key is still computed and returned either way, because a writer that
+      spelled its own key could write one no restore looks for.
+    required: false
+    default: "true"
   flavor:
     description: >-
       Which set of compiled artifacts this job produces, and therefore which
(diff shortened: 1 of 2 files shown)


tinnhlt-gra:
Right, and it was the load-bearing half of the fix — thank you.

`restore: "false"` on the composite action is necessary and **not sufficient**: `actions/setup-go` restores a build cache of its own into `GOCACHE` before anything else in the job runs. Confirmed from this repo's own logs — `Cache Size: ~25 MB (26132280 B)` on the `setup-go-Linux-x64-ubuntu24-go-1.26.6-…` key. Small, but the entry saved afterwards would have carried those artifacts forward, which is the same accumulation this PR exists to stop, just with a smaller seed. Worse, the comment I wrote claimed a cold compile, so the code would have read as correct while being wrong.

Fixed in 18fd51f63, taking a narrower cut than `cache: false`:

```yaml
- name: Start from an empty build cache
  run: go clean -cache
```

`cache: false` on `setup-go` would also disable the **module** cache, making this job re-download every dependency to produce the same artifacts. Only the *build* cache has to be cold, so `go clean -cache` targets exactly that and leaves module downloads warm. Applied to both warm jobs, and the comment now says the restore-skip is necessary but not sufficient and why.

PR #2847, opened 2026-08-27
fix(frontend): the leads surfaces read as the same product as contacts

luit-nfq:
## What this is

UI/UX work on the two lead surfaces, plus the global table fix they made
visible. Three tasks, and what each one turned out to be.

---

## 1. The column grip (`ListTable`, every list screen)

`.lt-grip::after` drew 1px of `--borderStrong` at 25%–75% of the header's
height, inset 4px from the cell edge. The cell's own `border-right` was already
at `right: 0`. So every column of every table carried **two parallel vertical
lines, 4px apart, in different colours and different heights** — and the
trailing column, whose divider `th:nth-last-child(2)` clears, carried the
grip's line by itself, against nothing. That is why it read as a rendering
fault rather than as an affordance.

- The grip draws nothing at rest; the divider is the line.
- It lights that same edge (2px, `--accent`, full height) on hover, on
  `:focus-visible`, and for the length of a drag — `is-dragging` rather than
  `:hover`, because a pointer dragged past the neighbouring column is no longer
  over the element it is moving.
- No grip on the trailing column: no divider to light, no neighbour to take
  width from.
- A drag now dresses the whole grid — `user-select: none`, resize cursor held —
  and a cancelled pointer ends it as surely as a released one, so the table
  cannot stay dressed for a gesture nothing is driving.

Held by two new cases in `listtable.test.tsx`. Blast radius is `ListTable`
only (10 product screens); `DataTable` is a different component and has no grip.

---

## 2. The leads queue

The list was already `ListTable` + `useListQuery`, so this was not a rebuild —
it was the chrome the other record lists have and this one did not.

- **Saved views.** Leads was the only record list without them, while the
  contract has carried `leads` in `SavedViewResource` all along and the
  frontend's `ViewResource` union simply did not name it. Same tab rail and
  same Save view verb as contacts, deals, organizations and projects.
- **The board/table toggle moves into `tools`**, where the deals screen's
  pipeline picker already lives: it changes how the same rows are *drawn*, so
  it belongs with the drawing dials rather than floating above the surface in a
  wrapper div. The segregation notice and the toast lose their wrappers too —
  the page is a stack, so a dismissed notice leaves no gap behind it.
- **Sortable identity column** on `full_name`, the A–Z view, and a Created
  column. All three are in the server's own `leadListFields`, so no header
  claims an ordering `/leads` cannot answer.
- **Bulk disqualify now asks.** It fired straight from the toolbar press while
  closing *one* lead opens a dialog. The batch is the less reversible of the
  two — the reader sees a count, not the rows — so it asks the same question
  and names the reason it is about to write on all of them.

---

## 3. The lead record page

It already used `RecordView`, but with neither rail nor aside — `zonesShape`
resolved to `"single"` and the page had **no grid at all**. Everything a rep
consults and everything they do was stacked in one column, each reading pushing
the next action further down.

It is now the two-column record page the contact page is:

- **Band** — a `StatStrip` of readings (score with what moved it, status,
  source, company) in place of a row of six badges styled inline with literal
  px. Six pills of equal weight read as decoration and get skipped; ruled slots
  with labels read as a table of what is true.
- **Rail** — what a rep *consults*: who owns this lead, and why it scores what
  it scores.
- **Work column** — the ladder, the details grid, the composer, the custom
  fields, in a `.record-stack`.
- **The ladder is the page's one tinted `Panel`** — the only surface here
  asking for a MOVE — and turns `warn` when the first response is already
  breached, because that is the finding rather than the layout.

Along the way, and stated because they are the parts a reviewer should push
back on if they disagree:

- **`RecordView` gains a `tabs` slot and `composed.css` a `.record-stack`**,
  both promoted from the person page, which had them as `.pe-tabs` and
  `.pe-overview-stack` — a screen holding the record page's own geometry, which
  the lead page was about to spell a third time. Person consumes both now; the
  two `pe-` rules are gone. This is the one change that touches every record
  page.
- **The page's ONE write is `useLeadPatch`**, shared by both columns: one PATCH
  shape, one If-Match, one invalidation, so an edit in the rail and an edit in
  the grid cannot send two versions of the same row.
- **All 27 inline `style` objects are gone.** What they spelled — a column, a
  row, a bare list — is four classes in `leads.css` on the space tokens.
- **All three lead dialogs sit on `ConfirmModal`** instead of hand-rolling a
  heading and a footer inside a bare `Modal`. `ConfirmModal` gains
  `confirmReason` so a refused confirm can still say why, which
  `confirmDisabled` alone cannot.
- **The ellipsis comes off the CTAs** — Qualify, Disqualify, More — in en, de
  and vi. The progress lines that end in one keep it: they report, they do not
  offer.

---

## Fixed along the way

- **`--color-text-muted` in `leads.css` is a token this tree has never
  defined**, so the ladder's separator had been falling back to `inherit` since
  it was written. Now `--textMeta`.
- **`.lead-ladder-how` and `.lead-ladder-step.is-terminal`** were referenced
  from the stepper and defined nowhere. Defined now. `is-current`, which
  nothing styled and which `aria-current` and the button variant already carry,
  is gone.
- **A real WCAG 2.2 AA failure.** `leads` was in *neither* axe sweep — not
  light, not dark, not the 390px no-horizontal-scroll pass — and the lead
  record was unswept too. Adding them found it: the row-selection checkbox is a
  native 13×13 box where §2.5.8 asks for 24×24, and it is the one checkbox in
  this product with no visible label to stand in for the missing area. Leads is
  the only list that passes `selection`, which is why no other sweep had met
  it. The box is scaled rather than overlaid — the rule measures the focusable
  control itself, so a pseudo-element would satisfy a finger and not the rule.
- **A refused confirm dragged its dialog footer out of line.** `Button` renders
  a refused control and its sentence as a column, which is right in a
  left-aligned row and wrong in a modal's right-aligned one: the confirm sat on
  a different line from Cancel with the explanation hanging off the edge. Every
  confirm dialog takes the fix, not only the lead ones — the shape was always
  wrong and no surface had passed `confirmReason` before to show it.
- **The leads queue showed ONE row at 390px.** `.lt-head` measured 146px
  against the contacts list's 87px, because eight view tabs wrapped to two
  lines — and this branch added the eighth. The view rail now scrolls sideways
  on a phone, the same rule the record page's tab strip keeps, taking the room
  left on its line rather than a line of its own so a three-view list keeps the
  height it had (contacts measures 87px before and after).
- **The citation register** drops `leads.tsx migration 0038` — build-process
  residue in a comment, which the rulebook asks a comment not to be, and which
  the rewrite removed.

---

## Stories

New, because the surfaces had none and `fe-uat` fails a changed component
without one: `leads.stepper.stories.tsx`, `leads.qualify.stories.tsx`,
`leads.disqualify.stories.tsx` — covering the refused, loading, empty and
failed states rather than only the happy one. The queue gains `Loading`,
`Failed` and `MorePages`, which contacts had and it did not.

---

## Gates run

| Gate | Result |
|---|---|
| `make check` | pass |
| `make frontend-check` | pass — 4985 tests, ds-gates, drift, lint, build |
| `make fe-uat` | pass — 137 stories captured |
| Browser, seeded stack | both themes, 390px + desktop, grip rest/hover/drag, Escape out of both dialogs, console clean |
| `make native-controls` | pass — 35 |
| `make frontend-e2e` | pass — 143 passed, 14 skipped, incl. 3 new leads axe sweeps |
| `make fe-clock-drift` | pass at +200 days — 4985 |
| `make bench-mobile` | p95 52 ms against a 300 ms budget |
| `check-ds-spacing` / `check-ds-purity` / `check-space-tokens` | pass |

---

## Left out, and why

- **`make verify-demo` fails on two rules on this machine, and it is not this
  branch.** `people work somewhere` (three people) and `accounts have a
  lifecycle` (one account) are all `seed-dev.sql` fixtures rather than
  demo-dataset rows, so `make seed-demo` cannot satisfy them — it re-ran and
  converged (192 companies, everything `0 new`).
- **The `bench-mobile` measurement record is not committed.** The run held the
  budget, but it happened on a laptop with three API processes and a Docker
  stack running, and overwriting the recorded baseline with a noisier number is
  not an improvement.
- **Merge and Share on a lead** — contacts offers both and leads offers
  neither. That is a product decision about whether a lead is a record you
  share, not UI love, so it is written up here rather than built.
- **Tab order on every record page is work column, then rail** — the rail sits
  visually left and comes second in the sequence. `PageZones` documents that as
  deliberate ("the story stays first"), it is what contacts and companies do
  too, and changing it would reshape every record page. Flagged, not touched.
- **The segregation callout costs 77px on a phone** until it is dismissed,
  which on a 390px screen is a third of the list. It is a first-run notice,
  localStorage-backed and gone for good after "Got it", and it predates this
  branch — but it is the biggest remaining reason the leads queue is shorter
  than the contacts one at that width.
- **The rep's default queue offers "Save view" on arrival**, because that
  default *is* a narrowing (`owner_id = me`) while the contacts default is not.
  Correct by the component's own rule and mildly odd to read. Worth a separate
  look at whether an owner-scoped default counts as narrowed.

## Summary by CodeRabbit

- **New Features**
  - Added saved-view tabs and controls to the Leads list.
  - Added sortable lead names and creation dates.
  - Added overview and history tabs to lead records.
  - Added confirmation dialogs for single and bulk lead disqualification.
  - Promoted leads retain their record view with promotion history and demotion controls.
  - Added scrollable record tabs and improved mobile layouts.
  - Added resizable list columns with persisted widths.

- **Bug Fixes**
  - Improved lead updates, timeline refreshes, and error handling.
  - Prevented accidental text selection during column resizing.

- **Accessibility**
  - Expanded accessibility coverage for lead records and mobile navigation.


luit-nfq:
Rebased onto `main` (21 commits) and pushed. Three conflicts, all resolved by keeping both sides:

- `leadbulk.tsx` — import line; `leadIdentityName` (main) and `usePlural` (this branch) are both used.
- `leads.qualify.stories.tsx` / `leads.disqualify.stories.tsx` — add/add. Main added a story each for the empty-`full_name` naming rule; this branch added the state matrix (preview outcomes, loading, error, phone). Kept this branch's file and re-added main's `NamedByItsAddress` story on top of it, so neither invariant is lost.
- `leads.tsx` — main's `leadIdentityName(lead) || t("lead.unnamed")` heading and the `RecordHistoryTab` `restore` prop landed in a block this branch had extracted into `LeadRecord`. Both ported into the new component.

`make frontend-check` after the rebase: 5069 passed.

### Review feedback

Four findings accepted and fixed in `3a2354b74` (now `ea5e0c6f1` after the rebase):

1. The shared lead mutation announced its error and its success on controls that did not start the write. The write now carries which control started it, so a half-typed score override survives a save from the details grid, and only the originating control reports.
2. The refused write on the History tab had no statement — the rail still writes there, so the refusal is now stated where the reader is.
3. Missing `mutationFn` variables on the disqualify and bulk mutations (repo rule: a `mutationFn` never closes over render state). Both now take their closure/run as a variable.
4. Bulk disqualify fired straight from `onClick`; it now goes through `ConfirmModal` like every other destructive verb.

Both regressions are held by new tests in `frontend/src/screens/leads.rail.test.tsx` (kept out of `leads.test.tsx`, which is already past the 1000-line ceiling).

One finding declined: the `vi.ts` wording suggestion. The Vietnamese string faithfully mirrors the `en`/`de` phrasing, and changing only `vi` desyncs the three catalogues — if the phrasing should change it should change in all three, which is a separate change.

The `fe-clock-drift` nitpick is already satisfied: `make fe-clock-drift` passes at +200 days.

PR #2850, opened 2026-08-27
perf(gates): the fitness gates run in parallel

tinnhlt-gra:
## What this does

The fitness gates run in parallel. The package goes **102 s → 18 s**, and it was the pole of the `-j4` fan-out that was **71% of `check-backend`** in CI.

Locally, `check-backend` goes **2m31s → 1m38s**:

| phase | before | after |
|---|---|---|
| `-j4` vet/lint/arch-lint/test/test-extensions | 86 s (57%) | **31 s (32%)** |
| root script gates (x30, -j4) | 53 s (35%) | 53 s (54%) |
| build, drift, composition | 12 s | 14 s |

## Measured first, because the fix depended on the shape

Each leg timed **alone and warm** — in the real run they contend, so the longest pole is invisible there:

| leg | cost |
|---|---|
| **test** | **91 s** |
| lint | 45 s |
| vet | 18 s |
| arch-lint | 2 s |
| test-extensions | 2 s |

Inside `test`, the gates package was a *flat tail*: 287 tests, 102 s, slowest single test 5.4 s. No hot spot to fix — the cost is ~100 independent tree walks (**78 `filepath.WalkDir`** and **106 `parser.ParseFile`** sites) at about 600 ms each.

Caching the parse inside `gatekit.Scope` was the other candidate. It would have returned **~13 s of the 102**, because only 22 of those walks go through `Scope`. Parallelism returns the wall time across all of them without editing one.

## Why it is safe, checked rather than assumed

These gates are readers. The state they share is each `gatekit.Waivers` set, and:

- its `matched` map is **already mutex-guarded**, with a comment naming `t.Parallel()` as the reason;
- `AssertAllMatched` **already** has to be called from exactly one place per declaration, held by `TestEveryWaiversDeclarationIsSweptForStalenessExactlyOnce`;
- getting that wrong under parallelism reports staleness that is not there — a **loud failure**, not a gate quietly passing.

Verified across three runs, two of them shuffled: **296 tests, identical set, identical verdicts, zero failures.**

## A gate holds it, because the regression is silent

A gate added without `t.Parallel()` still **passes**. It just runs alone while the others share the machine, and the suite gets a little slower with nothing failing to say so.

`TestEveryGateRunsInParallelWithTheOthers` requires the call to be **first** — Go runs a parallel test's body up to the call synchronously, so work above it is serial work that reads as annotated.

Mutation-tested, each mutant proved to have landed:

| mutant | |
|---|---|
| a gate loses `t.Parallel()` | caught |
| the call is present but not first | caught |
| an exemption names a gate that is gone | caught |
| the staleness sweep is dropped | caught (by the existing exactly-once gate) |
| the census sees no gates at all | caught (floor) |

## One exemption

`TestTheSchemaAndTheParserAgreeOnEveryInputDeclaration` calls `t.Setenv`, which Go forbids in a parallel test — the environment is process-wide. It is ratified in a `gatekit.Waive` set rather than a bare map, so its reason meets the same bar as every other waiver here and a stale entry is reported by the mechanism that already does that. (The tree's own `TestEveryPackageLevelReasonMapIsAWaiverOrADeclaredFixture` caught my first version, which was a hand-rolled map with a hand-rolled staleness check — a second copy of machinery that already exists.)

## What this does not fix

The script gates are now the pole at 54%, and they are near their own floor: at `-j4` the wall cannot drop below the slowest single gate. Timed cleanly, with nothing else of mine running:

| gate | |
|---|---|
| `test-money-scale` | 34 s |
| `money-scale` | 32 s |
| `test-one-spelling` | 20 s |
| `one-spelling` | 18 s |

So ~34 s is the floor against a 53 s fan-out. (A first reading put `test-money-scale` at 75 s; that was taken while other jobs of mine were running and disagreed with a clean baseline, so it is not the number — worth saying, because the contended figure would have argued for a much bigger prize than is actually there.)

Going further means making those scans faster — `money-scale` awks every Go and TS file through a comment stripper — which is a change to a gate whose failure mode is *going blind*. That wants its own PR and its own care rather than riding this one.

## Summary by CodeRabbit

- **Tests**
  - Enabled parallel execution across the gate test suite to reduce validation time.
  - Added checks that verify parallel-test coverage across supported build configurations.
  - Improved handling of build-specific and sequential tests.
  - Preserved existing assertions and validation behavior.

- **Documentation**
  - Updated the gate inventory to reflect expanded parallel-execution coverage and revised census totals.

### DIFF
--- a/backend/gates/agentgrantscopes_test.go
+++ b/backend/gates/agentgrantscopes_test.go
@@ -33,6 +33,7 @@ import (
 )
 
 func TestEveryGrantFundsTheToolsItsAgentDeclares(t *testing.T) {
+	t.Parallel()
 	granted := grantedScopes(t)
 	required := toolScopes(t)
 	declared := agentToolLists(t)
--- a/backend/gates/aggregategatereach_test.go
+++ b/backend/gates/aggregategatereach_test.go
@@ -68,6 +68,7 @@ type aggregateJobs struct {
 }
 
 func TestEveryAggregatedJobCanRunOnTheMergeQueue(t *testing.T) {
+	t.Parallel()
 	raw, err := os.ReadFile(aggregateWorkflow) // #nosec G304 -- a fixed repo-relative path
 	if err != nil {
 		t.Fatalf("reading %s: %v", aggregateWorkflow, err)
@@ -162,6 +163,7 @@ var callerConcerns = []string{
 // the fan-ins need `always()` precisely so a failed sibling cannot skip them into
 // a green.
 func TestNoLaneWorkflowDecidesItsOwnScope(t *testing.T) {
+	t.Parallel()
 	lanes, err := filepath.Glob(filepath.Join(workflowDir, "_lane-*.yml"))
 	if err != nil {
 		t.Fatalf("listing lane workflows: %v", err)
--- a/backend/gates/aiactivitycatalogparity_test.go
+++ b/backend/gates/aiactivitycatalogparity_test.go
@@ -42,6 +42,7 @@ import (
 const documentReadingKind = activities.ExtractionAITask
 
 func TestEveryKindSomethingProducesIsOneTheContractCanExpress(t *testing.T) {
+	t.Parallel()
 	declared := crmYAMLNamedEnum(t, "AiActivityKind")
 	if len(declared) == 0 {
 		t.Fatal("AiActivityKind declares no enum; this gate would pass vacuously")
@@ -110,6 +111,7 @@ func producedKinds() []string {
 // The reverse: a kind nothing can produce is copy three locales carry, a line
 // no reader will see, and a promise the server cannot keep.
 func TestEveryContractKindHasSomethingThatProducesIt(t *testing.T) {
+	t.Parallel()
 	produced := producedKinds()
 	for _, kind := range crmYAMLNamedEnum(t, "AiActivityKind") {
 		if !slices.Contains(produced, kind) {
@@ -125,6 +127,7 @@ func TestEveryContractKindHasSomethingThatProducesIt(t *testing.T) {
 // a string a strict client rejects; a smaller one truncates below what the
 // contract promised a reader would get.
 func TestTheReadsTextCapsAreTheOnesTheContractPublishes(t *testing.T) {
+	t.Parallel()
 	for _, b := range []struct {
 		property string
 		cap      int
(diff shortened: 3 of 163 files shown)


tinnhlt-gra:
Real, and mine — `t.Parallel()` on both put a reader and a writer of one file in the same window. Confirmed in the source: `TestEveryCitedMigrationExists` calls `pruneCitationRegister` under `-update-citations` (line 156), and `TestTheCitationRegisterIsSortedAndUnique` reads the same `citationRegisterPath`.

Fixed in 52239bd94, but **sequential rather than skipped**.

Go resumes parallel tests only after every sequential one has finished, so leaving the reader sequential orders it strictly before any prune. From the trace:

```
=== RUN   TestEveryCitedMigrationExists
=== PAUSE TestEveryCitedMigrationExists
=== RUN   TestTheCitationRegisterIsSortedAndUnique
--- PASS: TestTheCitationRegisterIsSortedAndUnique
=== CONT  TestEveryCitedMigrationExists
--- PASS: TestEveryCitedMigrationExists
```

The skip would have closed the race too. I avoided it because it leaves the register **unexamined in the one mode that rewrites it** — and an unrun gate reads exactly like a passing one, which is the failure this repository spends a lot of effort avoiding elsewhere. Sequential keeps the verdict and costs one short file read.

It is named in `parallelExempt` with that reason, so `TestEveryGateRunsInParallelWithTheOthers` does not flag it and the next reader learns why it is the odd one out.

tinnhlt-gra:
Confirmed and fixed in 738e9e4c7 — the numbers are 167 against 154:

```
go list -f '{{len .TestGoFiles}}' ./gates                  → 167
go list -tags integration -f '{{len .TestGoFiles}}' ./gates → 154
```

Thirteen gates carry a `!integration` constraint, so with the tag set the census was judging functions the build does not contain.

I took the first of your two options — ask Go for the active set rather than re-implement the constraints:

```go
ctx := build.Default
if builtWithIntegrationTag {
    ctx.BuildTags = append(ctx.BuildTags, "integration")
}
pkg, err := ctx.ImportDir(gateDir, 0)
```

`builtWithIntegrationTag` is a sentinel pair (`buildtag_default_test.go` / `buildtag_integration_test.go`, four lines each) because Go gives a binary no way to read its own build tags, and `build.Default` alone would answer for the default build regardless of what compiled the test.

**Doing it properly surfaced a second thing**, which is the part worth flagging: with the census now honest about the active set, two of the three `parallelExempt` entries live in `!integration` files — so under the tag they name subjects the package does not contain, and `AssertAllMatched` reported them as stale. Absence there means "excluded by a build constraint", not "ratifies code that is gone". The staleness sweep is therefore scoped to the build that contains every gate, with that reason stated at the call site. Verified green under both `go test ./gates` and `go test -tags integration ./gates`.

PR #2903, opened 2026-08-27
fix(capture): relinking an activity merges its displaced participants

### LINKED ISSUE #436: activities: relinking an activity can leave the displaced participant row behind
## What

`RelinkActivity`'s participant repoint carries a `NOT EXISTS` guard that skips the rewrite when the target person already has a participant row on that activity:

```sql
AND NOT EXISTS (
    SELECT 1 FROM activity_participant other
     WHERE other.activity_id = ap.activity_id AND other.person_id = $target)
```

The guard exists to avoid violating the per-(activity, person) uniqueness constraint. But when it fires, the displaced participant row is not rewritten and not deleted either, so after the relink the activity still names the OLD contact as a participant even though its link now points at the new one. The participants and the links stop telling one story, which is exactly what the repoint exists to prevent.

There is a second shape worth checking in the same statement: when the target has no participant row and several displaced ids do, every displaced row qualifies and each is rewritten to the target, which can collide on that same uniqueness constraint.

## Why it matters

`activity_participant` is registered PII (`piicoverage_test.go`) and it is what the interaction-edge projection derives its `(user, person)` pairs from. A stale participant row therefore keeps feeding a relationship-strength signal for a person the human explicitly said the conversation was not with.

## Status

Pre-existing. The statement is byte-identical to `main`; surfaced during review of #425, a behavior-preserving refactor that moved it into a named helper without changing it.

## Fix sketch

One set-based merge that keeps or creates exactly one target row and removes the displaced duplicates, rather than a conditional rewrite that silently no-ops. Two regression cases are needed and neither exists today: target already a participant, and multiple displaced participants in one relink.

## Related

Distinct from the stale interaction-edge gap already documented in-source at the same call site (the ADR-0078 `relinked_from` payload the graph consumer needs). That one is a public-event contract change; this one is local to the repoint statement.

dungbanguyen:
`RelinkActivity` repoints the activity's participants along with its link, so the two keep telling one story: a human saying *"this conversation was not with her, it was with him"* must not leave her named on it. The repoint was a conditional rewrite with a `NOT EXISTS` guard, and it was wrong in **both** directions.

**Skipped.** When the target already had a participant row, the guard fired and the displaced row was neither rewritten nor deleted — so the activity's link named him and its participants still named her. `activity_participant` is registered PII and is what the interaction-edge projection derives its `(user, person)` pairs from, so that row kept feeding a relationship-strength signal for somebody the human had just ruled out.

**Collided.** With several displaced participants and no target row, every one of them qualified and each was rewritten to the target, violating `uq_activity_participant`.

**The guard also asked the wrong question.** Uniqueness is per `(activity, role, user, person, address)`, not per person — so a target already on the activity in a *different role* was enough to skip a rewrite that would not have collided, and two displaced rows in the *same* role collided where the guard saw nothing.

One set-based merge replaces it, keyed on the tuple the index actually uses: within each group exactly one displaced row is promoted to the target — and only where the target holds no row of that shape already — and the rest are deleted. Either way the target is named once and no displaced row survives.

Both shapes have a case, and both fail against the old statement:

```
the displaced contact is still named on the activity (1 row(s)) — its link says one thing and its participants another
relink with two displaced participants: ERROR: duplicate key value violates unique constraint "uq_activity_participant" (SQLSTATE 23505)
```

Out of scope, and unchanged: the stale interaction-edge gap documented in-source at the same call site is a public-event contract change and stays its own slice.

`make check-backend` green.

Closes #436

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved activity relinking when participants are reassigned.
  * Prevented duplicate participant records and uniqueness conflicts during relinking.
  * Removed stale participant entries left behind after multiple participants are merged.
  * Ensured displaced participants are correctly removed when the destination already contains a matching participant.

### DIFF
--- a/backend/internal/compose/integration/relinkparticipants_integration_test.go
+++ b/backend/internal/compose/integration/relinkparticipants_integration_test.go
@@ -1,0 +1,98 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+//go:build integration
+
+package integration
+
+// Relink repoints the activity's PARTICIPANTS along with its link, so the two
+// keep telling one story: a human saying "this conversation was not with her,
+// it was with him" must not leave her named on it. activity_participant is
+// registered PII and is what the interaction-edge projection derives its
+// (user, person) pairs from, so a row left behind keeps feeding a
+// relationship-strength signal for somebody the human just ruled out.
+//
+// Both shapes below used to break that, in opposite directions.
+
+import (
+	"testing"
+
+	"github.com/margince/margince/backend/internal/modules/activities"
+	"github.com/margince/margince/backend/internal/shared/kernel/ids"
+)
+
+func seedRelinkActivity(t *testing.T, e *Env, subject string) ids.UUID {
+	t.Helper()
+	id := ids.NewV7()
+	e.WsExec(t, `INSERT INTO activity (id, kind, subject, occurred_at, source, captured_by)
+		VALUES ($1, 'email', $2, now(), 'manual', 'human:x')`, id, subject)
+	return id
+}
+
+func namedOn(t *testing.T, e *Env, activity, person ids.UUID) int {
+	t.Helper()
+	return e.WsCount(t, `SELECT count(*) FROM activity_participant
+		 WHERE activity_id = $1 AND person_id = $2`, activity, person)
+}
+
+// The target is ALREADY a participant. The repoint's skip left the displaced
+// row standing, so the activity's link named him and its participants still
+// named her.
+func TestRelinkRemovesTheDisplacedParticipantWhenTheTargetIsAlreadyOne(t *testing.T) {
+	e := Setup(t)
+	admin := e.As(e.Rep1, nil, AdminPerms)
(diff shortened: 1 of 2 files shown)


dungbanguyen:
Taking the review point on the existing-target case — it seeds the target in a **different role**, which is a different tuple under `uq_activity_participant`, so it reaches the promote branch rather than the delete branch it is named for.

I did not swap the role, because that would lose what the case does cover: the old guard asked only about the *person*, so a target present in **any** role was enough to skip the repoint and leave the displaced row standing. That is a real regression worth holding.

So the same-role case is added beside it as `TestRelinkDeletesTheDisplacedParticipantWhenTheTargetHoldsItsShape`. Against `main`'s statement all three now fail, each for its own reason:

```
RemovesTheDisplacedParticipantWhenTheTargetIsAlreadyOne   the displaced contact is still named  (different role, skipped)
MergesSeveralDisplacedParticipantsWithoutColliding        duplicate key ... uq_activity_participant (23505)
DeletesTheDisplacedParticipantWhenTheTargetHoldsItsShape  the displaced contact is still named  (same role, skipped)
```

PR #2905, opened 2026-08-27
fix(capture): raw_capture keeps the provider's bytes

### LINKED ISSUE #437: capture: raw_capture loses byte-fidelity for a provider payload that is not valid UTF-8
## What

The raw-evidence insert stores the provider original via `json.Marshal(string(rec.Raw))`. Go's `json.Marshal` replaces each invalid UTF-8 byte with U+FFFD, so a payload that is not valid UTF-8 is stored altered. `raw_capture` then no longer holds what the provider actually sent, which is the one thing that table exists to hold — the rawcapture and mailmap comments both describe it as the provider original.

## Why it matters

The raw copy is the fallback when mapping or classification is later found wrong: it is what a re-parse or a dispute reads. Mail is the realistic trigger — a message carrying a non-UTF-8 charset body or a malformed header is not exotic, and the corruption is silent: the insert succeeds, the row looks fine, and only a byte-level comparison against the provider would reveal it.

## Status

Pre-existing (`main` has the identical call); surfaced during review of #425, which moved this insert into a named helper without changing it.

## Fix sketch

Store the payload byte-preserving rather than through a JSON string encode — either the dedicated raw-capture insert path or a format that round-trips arbitrary bytes (bytea, or base64 with the encoding recorded). Whichever lands wants a regression test that writes an invalid-UTF-8 payload and asserts the read-back is byte-identical, since no current test would notice the substitution.

### LINKED ISSUE #2548: capture: one email containing a NUL byte permanently stops a mailbox's sync
### Summary

A single unauthenticated email containing a NUL byte (`0x00`) or invalid UTF-8 in its body permanently halts all mail capture for the mailbox that receives it. The capture transaction aborts on a PostgreSQL type rejection, the failure is not classified as a per-message skip, and the sync watermark is therefore never advanced past the offending message. Every subsequent sync re-fetches the same message and fails identically. No retry, reconnect, or backoff clears the condition.

Anyone who can send mail to a synced mailbox can trigger it. Mailbox addresses are published by design.

### Affected component

`backend/internal/modules/capture` — the IMAP, Gmail and Microsoft Graph ingest path.

### Details

Four links, each verified independently:

1. **No sanitiser.** `mailmap.extractText` copies decoded MIME part bytes straight into the message body (`string(content)`), and nothing between the parser and the database coerces invalid UTF-8 or strips NUL. There is no `ToValidUTF8`, `utf8.Valid` check, or NUL strip anywhere in `internal/modules/capture`.

2. **The database rejects it.** `storeRawCapture` (`internal/modules/capture/sinkraw.go:29`) JSON-encodes a non-JSON original with `json.Marshal(string(rec.Raw))`. Go's encoder renders a NUL as a six-character Unicode escape (backslash-u-0-0-0-0), which PostgreSQL's `jsonb` input refuses (`22P05`, "unsupported Unicode escape sequence") because it cannot be represented in the internal text form. `raw_capture.payload` is `jsonb NOT NULL`. Independently, `activity.subject` and `activity.body` are `text` and reject an embedded `0x00` outright.

3. **The error is fatal to the pull, not to the message.** `storeRawCapture` is called at `internal/modules/capture/sink.go:177`, inside the single `s.db.Tx(...)` callback, with no savepoint. Its error aborts the whole transaction and propagates out of `Sink.Upsert`. The connectors only treat `connector.ErrSkip` as a per-message drop — `internal/modules/capture/imap/imap.go:196-202` returns every other error as fatal, and the Gmail and Graph connectors have the same shape. A database data-shape rejection is not `ErrSkip`.

4. **The watermark never advances.** `Registry.SyncOnce` (`internal/modules/capture/registry.go:250-257`) takes the failure branch, calls `recordSyncFailure`, and returns — so `commitSyncCursor` at `:259` is never reached. The persisted cursor stays where it was before the poison message. The ADR-0063 backoff ladder classifies and degrades to a daily probe, but never quarantines the offending UID or advances past it.

The result is a permanent, self-sustaining capture outage for that mailbox from one message.

### Why the existing containments do not cover it

The codebase already recognises this failure class. `internal/modules/capture/sinkchannel.go:122-130` names it directly:

> a refusal there fails the entire capture and hands the connector a deterministic error it retries forever — the poison pill `sinkensure.go`'s savepoint exists to contain.

Two mitigations exist for it, and neither reaches this path:

- `admitCounterpartyShape` runs at the edge, before the transaction opens, so an unclassifiable counterparty shape costs nothing. A NUL byte in the body is not a counterparty-shape problem, so this gate never sees it.
- `sinkensure.go` uses savepoints so a failed statement does not poison the surrounding transaction. `storeRawCapture` is not wrapped in one.

Relatedly, `sink.go:164-166` states that the pre-store gates run before the raw store so that "a message that gets that far has been kept whatever happens next" — but the raw store is itself the statement that fails.

### Impact

Availability, unauthenticated, remote. One email permanently stops ingest for the targeted mailbox; all later mail for that mailbox is never captured. Repeating it against each connected mailbox disables capture workspace-wide.

Diagnosis is unusually hard: the message body is never stored, so nothing appears on the timeline to correlate the outage with. The pipeline-trace feature does not help either — `traceActivity` (`sink.go:326`) joins the same transaction and rolls back with it, and in any case `storeRawCapture` fails at `:177` before the trace is ever attempted. The visible symptom is a connector sitting in `error` with no indication of the cause.

Suggested severity: **High**. `CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H` (7.5) — offered as a starting point, not a claim on your rating.

Scope note: this is not one of the enumerated in-scope bullets in `SECURITY.md`, but it is not the excluded case either — the exclusion is denial-of-service against a **dev** deployment (`MARGINCE_ENV=dev`), whereas this affects production ingest.

### Reproduction

Send one message to any address a workspace syncs. The base64 payload decodes to `Hello\x00World`; every octet on the wire is printable ASCII, so it traverses standard relays unmodified.

```
MAIL FROM:<reporter@example.test>
RCPT TO:<monitored@target-workspace.example>
DATA
From: reporter@example.test
To: monitored@target-workspace.example
Subject: quarterly update
MIME-Version: 1.0
Content-Type: text/plain; charset=utf-8
Content-Transfer-Encoding: base64

SGVsbG8AV29ybGQ=
.
QUIT
```

Expected: the connector's next sync fails, its cursor does not advance, and no later message for that mailbox is captured. A raw `0xFF` byte in an 8-bit body reaches the same `text`-column rejection.

There is currently no test in `backend/` covering a NUL byte or invalid UTF-8 in captured content.

### Suggested remediation

1. Sanitise in the one parser every adapter shares, before the record leaves `mailmap`: coerce subject, body and counterparty name with `strings.ToValidUTF8` and strip `0x00`. Store the raw-capture original as `bytea`, or sanitise before `json.Marshal`.
2. Make the Sink defensive regardless: classify a write refused for a data-shape reason (SQLSTATE class 22) as `connector.ErrSkip` plus a breadcrumb, so a single unstorable message can never hold the watermark back — extending the containment `sinkensure.go` already applies elsewhere.
3. Add a regression test asserting the cursor advances past a message the database refuses.

dungbanguyen:
The raw-evidence insert stored a non-JSON original as `json.Marshal(string(rec.Raw))`. Go's encoder replaces every invalid UTF-8 byte with U+FFFD, so a payload that is not valid UTF-8 was stored **altered** — and `raw_capture` then no longer held what the provider sent, which is the one thing that table exists to hold.

Mail is the realistic trigger: a body in a non-UTF-8 charset, or a malformed header, is not exotic. The corruption is silent — the insert succeeds, the row looks fine, and only a byte-level comparison against the provider would reveal it. The raw copy is what a re-parse or a dispute reads, so a silent substitution is the worst shape this could take.

The same encode has a second edge with a sharper end: a NUL becomes the six-character escape that jsonb refuses outright (`22P05`). That one does not corrupt quietly — it fails the capture transaction.

## Three shapes, only the third is new

- JSON the column will take is stored **as itself**, unchanged;
- text that is valid UTF-8 and NUL-free keeps the **JSON-string spelling** every existing row already uses;
- anything else travels **base64 in an envelope that names its own encoding**.

The encoding travels *with* the payload rather than being inferred, because a reader guessing at a blob is how a re-parse silently reads the wrong bytes. The one reader that unwraps a stored original — `decodeStoredOriginal` — recognises the envelope by its declared encoding rather than by shape, so a provider payload that happens to carry those two keys cannot be mistaken for one.

**No migration.** Every existing row keeps its spelling, and the two shapes that already worked take the same path they took before.

## Coverage

The cases walk every shape a provider can send and require it back byte for byte. Against the old spelling, exactly the two the issue names fail:

```
--- FAIL: .../a_body_that_is_not_valid_UTF-8
--- FAIL: .../a_body_carrying_a_NUL
```

## Relationship to #2548

This removes the mechanism behind that report's poison pill **at its source** — the NUL no longer reaches the column as an escape, so the insert no longer fails and the sync watermark no longer sticks. It does **not** close #2548: that asks additionally for the Sink to classify a data-shape rejection (SQLSTATE class 22) as `connector.ErrSkip`, so that *any* future unstorable message can never hold a mailbox's cursor back. That defence-in-depth half is still worth having and is left open.

`make check-backend` green.

Closes #437

## Summary by CodeRabbit

* **New Features**
  * Raw capture now preserves non-UTF-8 data and content containing NUL characters without altering the original bytes.
  * Base64-encoded raw captures are supported when storing and reading captured payloads.
  * Existing JSON and text payload formats remain compatible and unchanged where applicable.
  * JSON containing literal escaped text continues to be preserved correctly.

* **Bug Fixes**
  * Invalid base64-encoded captures are now rejected with an error.

### DIFF
--- a/backend/internal/compose/participantreplay.go
+++ b/backend/internal/compose/participantreplay.go
@@ -26,6 +26,7 @@ package compose
 
 import (
 	"context"
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -213,13 +214,28 @@ func replayOne(ctx context.Context, tx pgx.Tx, c replayCandidate) (string, error
 
 // decodeStoredOriginal unwraps what the sink put in raw_capture.payload.
 //
-// The column is jsonb and a provider's original need not be JSON: capture
-// stores a non-JSON payload (an RFC822 message) as a JSON *string*, and a JSON
-// one (a calendar event resource) as itself. Both spellings arrive here, so
-// the string case is unwrapped and anything else is handed on as it stands.
+// The column is jsonb and a provider's original need not be JSON, so three
+// spellings arrive here: a JSON payload (a calendar event resource) as itself,
+// text (an RFC822 message) as a JSON *string*, and bytes jsonb cannot hold as
+// text — invalid UTF-8, or a NUL — in a base64 envelope that names its own
+// encoding. The envelope is checked before the string case because it IS a
+// JSON object, and it is checked by its declared encoding rather than by shape
+// so a provider payload that happens to carry those two keys cannot be
+// mistaken for one.
 func decodeStoredOriginal(payload []byte) ([]byte, error) {
 	if len(payload) == 0 {
 		return nil, errors.New("compose: the stored original is empty")
+	}
+	var envelope struct {
+		Encoding string `json:"encoding"`
+		Data     string `json:"data"`
+	}
+	if err := json.Unmarshal(payload, &envelope); err == nil && envelope.Encoding == capture.RawCaptureBase64Encoding {
+		raw, err := base64.StdEncoding.DecodeString(envelope.Data)
+		if err != nil {
+			return nil, fmt.Errorf("compose: decoding the stored original: %w", err)
+		}
+		return raw, nil
 	}
 	if !strings.HasPrefix(strings.TrimSpace(string(payload)), `"`) {
 		return payload, nil
--- a/backend/internal/modules/capture/rawcapturepayload_test.go
+++ b/backend/internal/modules/capture/rawcapturepayload_test.go
@@ -49,6 +49,11 @@ func TestARawCapturePayloadComesBackByteIdentical(t *testing.T) {
 		{"a body that is not valid UTF-8", []byte("Subject: Gr\xfc\xdfe\r\n\r\nlatin-1 \xff\xfe bytes")},
 		{"a body carrying a NUL", []byte("Subject: quarterly update\r\n\r\nHello\x00World")},
 		{"JSON whose own string escapes a NUL", []byte(`{"body":"Hello\u0000World"}`)},
+		// The lookalike: an escaped BACKSLASH followed by the text u0000. It
+		// decodes to a backslash and five characters, carries no NUL, and jsonb
+		// takes it — so it must stay stored as itself rather than be sent down
+		// the base64 path by a search that cannot tell the two apart.
+		{"JSON carrying the TEXT of a NUL escape", []byte(`{"body":"Hello\\u0000World"}`)},
 	} {
 		t.Run(tc.name, func(t *testing.T) {
 			stored, err := rawCapturePayload(tc.raw)
@@ -117,4 +122,21 @@ func carriesNUL(v any) bool {
 		}
 	}
 	return false
+}
+
+// The lookalike above decides between two spellings, so it gets its own
+// assertion rather than only riding the round-trip.
+func TestTheTextOfANULEscapeIsStoredAsJSON(t *testing.T) {
+	lookalike := []byte(`{"body":"Hello\\u0000World"}`)
+	stored, err := rawCapturePayload(lookalike)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if string(stored) != string(lookalike) {
+		t.Errorf("stored as %q, want the document unchanged — it carries the TEXT of an escape, not a NUL", stored)
+	}
+	escaped := []byte(`{"body":"Hello\u0000World"}`)
+	if stored, err := rawCapturePayload(escaped); err != nil || string(stored) == string(escaped) {
+		t.Errorf("a real NUL escape stored as %q (err %v), want it off the as-itself path", stored, err)
+	}
 }
--- a/backend/internal/modules/capture/rawcapturepayload_test.go
+++ b/backend/internal/modules/capture/rawcapturepayload_test.go
@@ -1,0 +1,120 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+package capture
+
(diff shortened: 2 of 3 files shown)


dungbanguyen:
Taken — the substring test could not tell an **escape** from the **text** of one. `{"body":"Hello\\u0000World"}` decodes to a backslash followed by five characters, carries no NUL, and jsonb takes it; the search saw the byte sequence and sent it down the base64 path for nothing.

Not a fidelity bug — the round-trip stayed exact either way — but a real one, because the whole point of keeping the two common shapes on their old paths is that they stay there.

**Parity decides it now**: only a `u0000` preceded by an *odd* run of backslashes is an escape, because the run's last backslash is what escapes the `u`.

The lookalike has its own case, in both directions: the text of an escape stays stored as itself, and a real escape leaves the as-itself path.

PR #2908, opened 2026-08-27
feat(frontend): toasts as one global region, with a real Undo where one exists

luit-nfq:
## What this is

Passive, transient feedback for writes whose result a reader cannot otherwise
see — with a real Undo where the product genuinely has an inverse write behind
it, and deliberate silence where it does not.

## The starting point was not nothing

`useToast` / `ToastRegion` already existed in `src/design-system/toast.tsx`,
catalogued and tested. It was a **hook**, so eight screens each minted their own
state and rendered their own region. All three failure modes of a global surface
held locally were live at once:

- `screens/commissiondecide.tsx` called `useToast` and rendered **no region** —
  every approve/pay/void confirmation went into a `useState` nobody read.
- `screens/deals.tsx` carried a comment explaining the instance belongs to the
  caller, because a hook minting its own shows its messages to nobody.
- The region being a **sibling** of the page's cards forced two layout rules to
  be written around a fixed box that takes no space — `enter.css` and the
  `.wrap:has(> .lt)` rule in `app/shell.css`. Both are deleted here.

Work that landed on `main` mid-branch showed the cost again: `approvalrow.tsx`
had grown an optional `toast` prop, a fallback to its own, and a conditional
region — three moving parts, all because the undo offer outlives the row that
shows it. One region answers that by construction; all of it is gone.

## Task 1 — the toast

One `ToastProvider` and one `ToastRegion`, mounted in `main.tsx` beside
`AttentionProvider` and portalled to `document.body` like every other overlay in
the design system. A screen only ever asks for `show`.

- **`action`** is the verb the message carries, drawn by the region rather than
  hand-placed in the body. Makes the toast sticky, gives it a dismiss, withdraws
  the message once pressed.
- **A verb outranks a report.** A message carrying an action is never evicted by
  one that only reports — that queues behind it.
- **The clock stops while you read.** Pointer over it or focus inside it holds
  the countdown; leaving hands back the full life (WCAG 2.2.1). Focus is never
  taken: in by Tab, out by Escape.
- **Mobile.** The region reads `--stickyBottomInset` instead of `bottom: 20px`.
  It previously sat *on top of* the fixed phone nav bar (toast z-index 60, bar
  30), covering the reader's way out of the screen.
- **Long content** wraps; the box stops at its own ceiling; the mark and verbs
  keep their width.

## Task 2 — the analysis

Full inventory of ~200 frontend write sites plus an audit of what the server can
actually reverse. Two findings shaped the design:

**Most destructive verbs have no undo.** Every `DELETE` in the contract is a soft
archive and **none has a restore endpoint** — 21 of them. The record-history
put-back refuses archives outright (`not_a_replayable_verb`,
`compose/undoability.go:51`), and the archive cascades hard-delete the
`list_member` and `taggable` rows, so even a future unarchive would be lossy. **A
"deleted — Undo" toast has nothing behind it, so this PR does not offer one.**

**A real send-undo exists and is unused.** `scheduled_at` is accepted by both send
endpoints with no minimum lead time, and `POST /scheduled-sends/{id}/cancel`
returns 204 while pending. See *Proposals*.

The policy for *which* writes get a toast lives in the story, where the next
author is working — not here.

## Task 3 — global, documented, tested

- One provider, one region, one stylesheet (`toast.css`, moved out of the
  40-component `atoms.css`).
- **New fitness gate** in `conformance.test.ts`: provider and region each mount
  in exactly one file, the shape `UnsavedGuard` is already held to. This is what
  would have caught the dead `commissiondecide` toast — no existing gate could,
  because the suite only ever rendered the toast's own harness, and a
  confirmation shown to nobody looks exactly like one nobody triggered.
- 18 unit tests: queue rule, deadline, pause-on-hover, pause-on-focus, Escape
  from a body-carried control, the action, the mark, portalling, timer cleanup.
- Story rewritten with five states and the adoption policy as its docs.

## Adoption

Shared write hooks, so one change reaches every screen going through them —
**20 call sites across 12 screens, compiler-enforced** because the message is a
required prop:

- `useUpdateRecord` — a dialog closing is the caller dismissing its own form,
  not the server agreeing to anything.
- `useArchiveRecord` — destructive, no restore endpoint, and the closing dialog
  was all a reader got.

Direct sites: capture settings, mail sharing, overnight agent grant, own email
domains (add/remove), consumer mail domains (add/remove), capture exclusions
(add/remove), user role change, tag apply, add-to-list.

**Undo, only where the inverse is exact:** tag apply (`DELETE /tags/{id}/apply`,
which the contract itself calls *"Undo for applyTag"*), deal-document hide, user
deactivate, team archive (its restore half did not exist in the UI and is added
here), and the approvals offer.

**Deliberately silent, with the reason in the code:** list membership (no removal
in the contract), role change (reversing needs the previous role, and the roster
has already been replaced), id-only removals (nothing to name them by), and every
archive.

## Fixed along the way

- **`customfields.tsx`** showed "A picklist needs at least one option" with the
  **green completion dot** — a refusal wearing a tick, because it reached the
  toast through a string-only `onToast` prop that could not say otherwise.
- **A shipped WCAG failure, pre-dating this branch.** `EntityRef` inside a toast
  measured `#0a6f4b` on `#13231d` = **2.63:1** against 4.5:1. The
  lead-qualified confirmation has put a control on the rail's dark plate in
  page-tuned accent green for as long as it has existed. Nothing caught it
  because no sweep in the tree had ever looked at a toast while one was up.
- **Escape belonged to the wrong element.** It was wired to the toast's two
  buttons, so a reader who tabbed to a control the *message* carried was inside
  a toast whose documented way out did nothing. It belongs to the region.
- **`--stickyBottomInset` did not reach portalled surfaces** — published on
  `.app`, which the portalled region is outside of, so it fell back to the plain
  ledge and covered the nav bar again. Now on the root.
- **`fe-uat` failed on `src/main.tsx`** — the app entry exports nothing and can
  have no story, the same false gate its `.d.ts` skip already avoids. Derived
  from `index.html` rather than named, so it follows a rename.
- `leads.list.tsx` and `approvalrow.tsx` hand-placed a `Button` inside the toast
  body; both go through `action` now.

## Left out, deliberately

- **The long tail of adoption** — `share` grant/revoke, saved views, DSR status,
  retention, Google app, AI provider keys, overlay user-map, lead vocabulary,
  deal-room title. All mechanical, no design questions; easier to review as one
  later pass than sprinkled through this branch.
- **Project attach/detach.** Re-attaching loses the role the detach knew, so the
  Undo would be quietly lossy. Wants a deliberate decision.
- **The person research drawer** — #2934. Its Save posts `claims: []` and has no
  `onSuccess`, so it writes nothing and says nothing. It deliberately got **no**
  toast: a "Saved" message on a write that stores nothing is worse than silence.
  The two claim shapes do not map and no UI bridges them, so it needs a product
  decision rather than a fix.

## Proposals (not done here)

1. **A restore endpoint for archives.** 21 archive endpoints, no way back from
   any, and the cascade drops list memberships and tags on the way out. Until it
   exists, Undo is unavailable for the verbs users most expect it on.
2. **A send-undo window.** `compose.tsx` already speaks `scheduled_at` and
   `ScheduledSendsScreen` already cancels; `scheduled_at = now + N` on a send-now
   would give a genuine Gmail-style Undo with no backend work. Held back because
   it changes composer behaviour — a scheduled send writes no activity and no
   delivery row, so the message stays off the timeline for N seconds. Note
   `send-message` (Telegram/Slack) has no `scheduled_at` and can never be undone.

## Gates

All run locally on the rebased branch. Note that **draft PRs skip every CI job**
in this repo (`ci.yml:73` and ~10 more gate on `draft == false`), so until this
was marked ready the green tick on it covered nothing.

| Gate | Result |
|---|---|
| `make check` | pass |
| `make frontend-check` | pass |
| `make native-controls` | pass — 35 tests |
| `make fe-uat` | pass — 186 stories captured |
| `make frontend-e2e` | pass — 144 passed, 0 failed |
| `make fe-clock-drift` | not run — no touched test reads a date |

## Verified in a browser

Against a freshly seeded stack, signed in as the demo admin: the toast renders
and is portalled outside `main`; the tag Undo round-trips for real (tag added,
sidebar count 2 → Undo → count 1, followed by its own confirmation); long content
wraps at 390px with no horizontal overflow; the toast clears the phone nav bar by
24px measured; console clean, no errors or warnings.

## Summary by CodeRabbit

* **New Features**
  * Added centralized confirmation notifications for saves, archives, updates, additions, and removals across the application.
  * Added actionable notifications with Undo support for tag, deal-file, team, member, and assignment changes.
  * Notifications pause while hovered or focused, support keyboard dismissal, queue actionable messages, and adapt to responsive layouts.
  * Added sticky error notifications for failed actions.
  * Added localized confirmation messages in English, German, and Vietnamese.
* **Accessibility**
  * Improved focus management, status announcements, Escape dismissal, and automated accessibility coverage for notifications.


luit-nfq:
Thanks — four of the five were real and are fixed in 160f07fa6. Notes per finding:

**`settings.tsx` / `products.tsx` — pre-edit name.** Confirmed and the worst of the five: renaming a stage announced the *old* name. `savedMessage` now also accepts `(updated) => string` and every site that names an instance reads the returned record. Two things were needed to make that typecheck, both worth stating: `NoInfer` on the parameter, because inference reads every position at once and a callback there dragged `Updated` down to its constraint; and the generic named explicitly at six call sites, because the `update` arrow cannot supply the type either — it is itself contextually typed by the generic being inferred.

**`users-access.tsx` — team name for Undo.** Confirmed. The restore now takes `{ id, name }` and uses the captured name, so a roster that has already dropped the archived row cannot make the confirmation call a team by its uuid.

**`toast.test.tsx` — focus pause on a sticky toast.** Confirmed, and the sharper version of your point is that the test could not fail: sticky means no timer, so it would have passed with the pause removed entirely. It drives a *timed* toast carrying a link now — the shape the lead-qualified confirmation actually has — and asserts both halves, that focus holds the deadline and that blurring hands back the full life rather than its remainder.

**`fe-uat.mjs` — attribute order and quoting.** Fixed. Each script tag's attributes are read independently, covering either order, single or double quotes, and a case-insensitive `type`. Worth noting the failure mode you identified was fail-*closed* rather than silent — an unmatched entry excludes nothing and the gate then fails on `main.tsx` — but a gate that is correct only for today's byte order is still the thing the rulebook warns about, so it is derived properly now.

**`edit.test.tsx` — no provider (nit).** Taken. The region is mounted, so what a save says is observable there at all, and it now pins the returned-name behaviour above plus the case that matters most beside it: a refused write stays silent.

All gates re-run green locally after the change: `make check`, `make native-controls`, `make fe-uat` (186 stories), `make frontend-e2e` (144 passed).

luit-nfq:
Both taken in 749959138 — and the Major one had a sibling.

**Failed Undo was silent.** Correct, and worth naming precisely: an Undo is the one press whose failure a reader cannot see, because the message it was offered from is consumed the moment it is pressed. A rejected restore left them watching a confirmation disappear and concluding the write came back.

Rather than patch the reported site, I grepped the other write that an Undo runs and found the same hole in `dealfiles.tsx`'s unhide. Both report their refusal now through the same `problemMessageOf` path the tag undo in `companyactions.tsx` already used — which is exactly why that one was right and these two were not. Sticky, since a refusal is not a courtesy to withdraw after 3.5s.

**`data-type` false positive.** Confirmed with your repro. `\b` sits between the hyphen and the `t`, so `data-type="module" data-src="…"` read as the entry module. Anchored on whitespace or the tag's start now. Checked against five shapes: both attribute orders, single and double quotes, uppercase `TYPE`, the `data-*` case, and a non-module script — the last two correctly yield nothing.

`make check` and `make fe-uat` green locally.

luit-nfq:
SonarCloud's quality gate failed on new-code coverage: 79.2% against a required 80%, 31 of 141 new lines uncovered. The uncovered lines were almost all the Undo arms — the inverse write, the confirmation that follows it, and the refusal.

`681c033e6` covers them by driving the real thing: the toast region is mounted the way `main.tsx` mounts it, the Undo button is pressed, and the assertion is on the route the inverse write reached plus the sentence that followed. Six tests across `companyactions`, `users-access` and `dealfiles`, each with both arms.

Fixed along the way, and the reason the refusal test would not pass at first: `takeTagOff` was handing the raw RFC-7807 body to `problemMessageOf`, which only unwraps a `ProblemError`. Every refusal it reported therefore said "something went wrong" instead of the server's own words. It is now a `useMutation` like the team restore and the document unhide beside it — the error is minted once in the `mutationFn` and arrives at `onError` intact.

PR #2909, opened 2026-08-27
fix(ai): a voice-source weight has to be a number

### LINKED ISSUE #438: ai: a voice-source weight of NaN or Infinity passes the 0..2 range check
## What

Voice-source weight validation is a pair of comparisons:

```go
if weight < 0 || weight > 2.0 {
```

Every comparison against NaN is false, so `NaN` passes both and is stored. JSON decoders in this path accept `NaN` / `Infinity` as numbers, so the value is reachable from the wire, and `+Inf` fails the upper bound but `-Inf`/`NaN` behave surprisingly.

## Why it matters

The weight is a multiplier in voice-corpus scoring. A single NaN weight propagates through any sum or average it participates in, turning the whole computed profile into NaN rather than producing an obviously-wrong number. That is a silent, hard-to-attribute failure rather than a rejected request.

## Status

Pre-existing (`main` has the identical check); surfaced during review of #425, which moved this validation into a named helper without changing it.

## Fix sketch

Add a finiteness check (`math.IsNaN` / `math.IsInf`) alongside the range check, and apply the same guard at the update site — the ingestion and update paths validate separately today, so fixing only one leaves the other open. A table test over `NaN`, `+Inf`, `-Inf` and the boundary values 0 and 2 would pin it.

dungbanguyen:
A weight is a **multiplier** in voice-corpus scoring, and the range check could not refuse one that is not a number: every comparison against NaN is false, so `weight < 0 || weight > 2.0` admitted it and stored it. The JSON decoders on this path take `NaN` and `Infinity` as numbers, so the value is reachable from the wire.

The cost is not a rejected request, it is a silent one. A single NaN weight propagates through every sum and average it takes part in and turns the whole computed profile into NaN — an obviously-wrong number would at least be attributable.

**Both write paths validated separately**, so fixing one would have left the other open: the ingest's declared-field check and the manifest patch each carried their own copy of the bounds. Both now answer through one function and say one sentence, so a caller cannot be told two different things about one rule.

Both infinities were already refused by the bounds — `+Inf` fails the upper, `-Inf` the lower. They are named in the check anyway so it says what it means rather than catching them by accident, and the table pins that they stay refused.

Against the range check alone, exactly the NaN case fails:

```
--- FAIL: TestAVoiceWeightMustBeANumberInRange/not_a_number
--- FAIL: TestBothVoiceWritePathsRefuseANaNWeight
```

`make check-backend` green.

Closes #438

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for voice-source weights during updates and corpus ingestion.
  * Invalid values, including values outside the 0–2 range, NaN, and infinities, are now rejected with a consistent error message.

* **Tests**
  * Added coverage for valid, out-of-range, and non-finite voice-source weights.

### DIFF
--- a/backend/internal/modules/ai/voice_source_mutations.go
+++ b/backend/internal/modules/ai/voice_source_mutations.go
@@ -66,8 +66,8 @@ func (s *VoiceStore) UpdateSource(ctx context.Context, profileID, sourceID ids.U
 }
 
 func validateSourceUpdate(in UpdateSourceInput) (*bool, error) {
-	if in.Weight != nil && (*in.Weight < 0 || *in.Weight > 2.0) {
-		return nil, &CorpusIngestError{Field: voiceKeyWeight, Reason: "must be between 0 and 2"}
+	if in.Weight != nil && voiceWeightRefused(*in.Weight) {
+		return nil, &CorpusIngestError{Field: voiceKeyWeight, Reason: voiceWeightRange}
 	}
 	excluded := in.Excluded
 	if in.Included != nil {
--- a/backend/internal/modules/ai/voicesources.go
+++ b/backend/internal/modules/ai/voicesources.go
@@ -10,6 +10,7 @@ package ai
 // visibleProfile gate.
 
 import (
+	"math"
 	"strings"
 	"time"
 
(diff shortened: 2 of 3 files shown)


dungbanguyen:
SonarCloud's quality gate is red here on `new_maintainability_rating`, and it is not this change's debt: both findings are pre-existing cognitive-complexity smells on `DeleteSource` and `ClearCorpus` (created 2026-07-23 and 2026-08-21), neither of which this PR touches. They surface because the PR edits two lines elsewhere in the same file, and a rebase onto current `main` did not clear them.

Leaving them: refactoring two ~60-line transaction bodies inside a PR about NaN weights mixes two pieces of work, and the complexity is real debt worth its own change rather than a drive-by.

`ci` — the required check — is green.

PR #2922, opened 2026-08-27
fix(worker): a role resolves its key vault once

### LINKED ISSUE #440: worker: the keyvault is resolved three times at boot, and the runner lane ignores the unconfigured flag
## What

`cmd/worker` boot calls `keyvault.FromEnv(pool)` in three places: the Surface-B runner lane, the connector-credential backfill, and the job runner. Each interprets the result differently.

The job runner honours the `configured` flag and nils the vault when it is false, so its channel and overlay wiring get an explicit nil. The runner lane does not: it passes the raw vault into `OverlayIncumbentResolver` regardless. On a deployment with no vault configured, the two paths therefore disagree about whether a vault exists — the runner's write-back attempts an unseal that fails per call, where the job runner's equivalent surface is wired off outright.

## Why it matters

Not a crash (the resolver's nil path is documented as returning a clean refusal), but the inconsistency means "is overlay write-back available?" has two different answers inside one process depending on which lane asks. That is the kind of divergence that makes an operator report unreproducible.

## Status

Pre-existing — `main` resolves the vault in the same three places with the same asymmetry. Surfaced during review of #427, which moved the boot sequence into named phases without changing what any phase does.

## Fix sketch

Resolve once at a single boot ownership boundary and pass `(vault, configured)` to every lane that needs it, so one decision reaches all of them. Worth confirming while doing it whether an unconfigured vault should leave overlay write-back declared unsupported rather than failing per call.

dungbanguyen:
Closes #440.

## What was there

`cmd/worker` resolved the key vault in three places — the Surface-B runner lane, the connector-credential backfill, the job runner — and each read the answer in its own way: two nil-ed the vault on the `configured` flag, the runner lane passed through whatever `FromEnv` returned. `cmd/api` had the same shape across three phases: the license binding, the connector-credential surface, the outbound relay password.

## Nothing actually diverged

`keyvault.FromEnv` already returns a **nil** vault when nothing is configured, so the flag and the nil say the same thing. The runner lane's `OverlayIncumbentResolver` therefore reported the same clean "no incumbent" the job lane's wired-off surface does.

That also answers the ticket's open question: on an unconfigured deployment overlay write-back is **already declared unsupported**, not failing per call.

## What was wrong

"Does this deployment have a vault?" had three answers in one process, and only their agreeing kept it from mattering. That is the shape a fourth lane gets wrong, and the symptom is an operator report nobody can reproduce.

## The fix

`keyvault.ForRole` resolves once, hands back a vault or nil, and lets no flag out. Each role calls it at boot and passes the value down: one deployment fact, read once, nothing to keep in step.

- the redundant `if !configured { vault = nil }` blocks go with it
- `backfillConnectorCredentials` returns nothing now — the vault resolution was the only error it ever passed up, and that is the boot's to refuse before any lane starts
- api's `run` gained `openInstallation`, which opens the pool and the vault together (both settled before any phase runs) and keeps the function under the length gate

## Held, not described

- `keyvault.FromEnv` is unreachable outside the vault package except from ratified sites that say why. One entry: compose's per-workspace BYOK lookup, which resolves per routing change rather than per boot, over a caller-supplied lookup, and deliberately warns rather than refusing.
- each role may call `ForRole` once — a second call fails the gate.
- the equivalence the roles now rely on (nil vault exactly when unconfigured) is asserted in both directions rather than assumed, since reading "unconfigured" off the value is what removing the flag rests on.

The census is module-wide rather than `cmd/`-only: boot wiring that moved into `internal/` would otherwise walk out of reach with the gate still green.

Both gate arms mutation-checked — adding a second `ForRole` call and a stray `FromEnv` to `cmd/worker/boot.go` fails each of them by name.

## Summary by CodeRabbit

- **Improvements**
  - Streamlined application and worker startup by initializing secure configuration once and reusing it consistently.
  - Improved handling when secure configuration is unavailable, allowing optional features such as email delivery to remain disabled without unnecessary startup failures.
  - Reduced disruption from credential backfill issues by treating individual failures as non-blocking.
  - Improved error context when secure configuration cannot be initialized.

- **Tests**
  - Added coverage to verify consistent secure-configuration behavior across application roles and deployment scenarios.

### DIFF
--- a/backend/cmd/api/boot.go
+++ b/backend/cmd/api/boot.go
@@ -134,7 +134,7 @@ func fileReleaseHandbook(ctx context.Context, pool *pgxpool.Pool, logger *slog.L
 //
 // Returns the loaded deployment config every later boot phase reads, and the
 // license watcher whose posture baseComposeOptions reports.
-func bindInstallation(ctx context.Context, cfg apiConfig, pool *pgxpool.Pool, logger *slog.Logger) (deployconfig.Config, *licensecheck.Watcher, error) {
+func bindInstallation(ctx context.Context, cfg apiConfig, pool *pgxpool.Pool, vault keyvault.Vault, logger *slog.Logger) (deployconfig.Config, *licensecheck.Watcher, error) {
 	deployCfg, err := deployconfig.Load(cfg.configPath, cfg.posture)
 	if err != nil {
 		return deployconfig.Config{}, nil, err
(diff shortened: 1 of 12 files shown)


dungbanguyen:
Rebased onto main. The gate immediately caught a **fourth** resolution that landed while this was in flight — `weeklyMailConfig` resolving its own vault to unseal the relay credential. It now takes the boot's, and its "vault unavailable" degradation goes with it: a vault that could not be built refuses the boot long before the weekly mail is considered, so what is left for that function to degrade over is the credential inside the vault, not the vault.

PR #2926, opened 2026-08-27
fix(metrics): a gauge does not carry the suffix that means counter

### LINKED ISSUE #460: metrics: the sweep gauges carry a _total suffix on a gauge, in both pairs
Raised in review of #457 by cubic, and deferred there deliberately.

## What

`margince_sweep_workspaces_total`, `margince_sweep_workspaces_failed`,
`margince_sweep_units_total` and `margince_sweep_units_failed` are all declared
`# TYPE … gauge`, and two of them carry the `_total` suffix Prometheus
convention reserves for counters.

They are levels, not counters: each reports how many workspaces (or units) a
fleet pass currently covers, read from `river_job` at scrape time. A number that
can go down.

## Why it matters

A `_total` suffix on a gauge invites `rate()` and `increase()`, which are
meaningless over a level and will render as noise rather than as an error. The
`# TYPE` line says gauge, but a dashboard author reading only the name will not
see it.

## Why it was not fixed in #457

That PR added the `units` pair as a deliberate sibling of the `workspaces` pair,
which shipped in #446 with the same shape. Renaming one half would leave an
operator reading `margince_sweep_workspaces_total` beside `margince_sweep_units`
— a worse surface than the convention it satisfies.

The convention point applies to both families or to neither, so it wants one
deliberate move rather than a side effect of an unrelated change.

## Shape of a fix

Rename all four, e.g. `margince_sweep_workspaces` / `margince_sweep_workspaces_failed`
and `margince_sweep_units` / `margince_sweep_units_failed`. This is a **breaking
change** for any existing dashboard or alert, which is the reason to do it as
one change with a note rather than incrementally.

## Scope

Metrics naming only. No behaviour changes; the readings are correct as they
stand.

dungbanguyen:
Closes #460.

## The rename

| before | after |
| --- | --- |
| `margince_sweep_workspaces_total` | `margince_sweep_workspaces` |
| `margince_sweep_units_total` | `margince_sweep_units` |

The two `_failed` halves already read correctly and are unchanged.

All four are **levels** — how many workspaces or units a fleet pass currently covers, read out of `river_job` at scrape time, numbers that go *down*. Prometheus reserves `_total` for monotonic counters.

The `# TYPE` line always said `gauge`. It is not what a dashboard author reads. The **name** is, and this one invited `rate()` and `increase()` over a level — neither of which looks wrong on a graph. They render as plausible noise, at exactly the moment somebody is deciding whether tenants are being missed.

Both pairs move together, which is why this is its own change rather than a side effect of #457. Renaming one half would have left an operator reading `margince_sweep_workspaces_total` beside `margince_sweep_units`.

## Breaking

Any dashboard or alert on the old names has to be updated. Noted in the changelog. **Nothing is aliased** — one number published under two names is a worse surface than a rename with a note.

## The gate

`TestOnlyACounterCarriesTheTotalSuffix` reads the `# TYPE` lines out of the exposition writers themselves rather than judging a list of family names, so a family added tomorrow is judged the day it is written and a deleted one takes its row with it. It covers both spellings this tree assembles headers in — the `Fprintf`/`WriteString` literals and compose's `writeFamilyHeader`, whose type is not in the format string.

The reverse arm is not decoration: a counter *without* the suffix is read as a level and alerted on as one — the same error with the roles swapped.

Both arms mutation-checked: putting `_total` back on `margince_sweep_workspaces` and taking it off `margince_relay_published_total` each fail the gate by name.

## Summary by CodeRabbit

* **Breaking Changes**
  * Renamed sweep workspace and unit gauge metrics by removing the `_total` suffix.
  * No compatibility aliases are provided; update dashboards, alerts, and integrations accordingly.

* **Documentation**
  * Updated configuration and metric references to reflect the new names.
  * Documented validation requiring `_total` for counters and excluding it from gauges.

* **Tests**
  * Added repository-wide validation to enforce Prometheus metric naming consistency.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,28 @@ numbers appear here when releases start.
 
 ### Changed
 
+- **BREAKING (metrics): the sweep gauges no longer carry a `_total` suffix.**
+  `margince_sweep_workspaces_total` is now `margince_sweep_workspaces`, and
+  `margince_sweep_units_total` is now `margince_sweep_units`. The two `_failed`
+  halves are unchanged. Any dashboard or alert reading the old names has to be
+  updated; nothing is aliased, because a series published under two names is a
+  worse surface than a rename with a note.
+
+  All four are levels — how many workspaces or units a fleet pass currently
+  covers, read out of `river_job` at scrape time, numbers that go DOWN — and
+  Prometheus reserves `_total` for monotonic counters. The `# TYPE` line always
+  said `gauge`; the NAME is what a dashboard author reads, and this one invited
+  `rate()` and `increase()` over a level, which render as plausible noise rather
+  than as an error.
+
+  Both pairs move together. Renaming one half would have left an operator
+  reading `margince_sweep_workspaces_total` beside `margince_sweep_units`,
+  which is a worse surface than the convention it satisfies.
+
+  A gate now holds the rule in both directions, over what the exposition writers
+  actually emit rather than over a list of names: a gauge may not carry the
+  suffix, and a counter must.
+
 - **The schema describes margince.yaml, not just the routing block.**
   `config/ai-routing.schema.json` is gone; `config/margince.schema.json` takes
   its place and covers the whole file — all fifteen sections — with the model
--- a/backend/cmd/worker/observe_test.go
+++ b/backend/cmd/worker/observe_test.go
@@ -154,8 +154,8 @@ func TestTheWorkerMetricsAreProcessLocalAndReServeNoFleetGauge(t *testing.T) {
 	for _, family := range []string{
 		"margince_job_queue_depth",
 		"margince_job_declared_info",
-		"margince_sweep_workspaces_total",
-		"margince_sweep_units_total",
+		"margince_sweep_workspaces",
+		"margince_sweep_units",
 		"margince_outbox_unpublished",
 	} {
 		if strings.Contains(body, family) {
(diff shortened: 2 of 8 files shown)


dungbanguyen:
Checked, and this one does not hold: the walk root is the **backend module root**, not `backend/gates`.

Go runs a test binary with its working directory set to the package's source directory — but this package is compiled and run through `go test -C backend ./gates`, and the module's own test setup puts the process in the module root. Measured rather than argued:

```
cwd: /Users/dung.nguyen/margince-work/repo/backend
internal/compose/jobmetrics.go IS reachable from the walk root
```

The mutation check in the PR body is the same evidence from the other end. Putting `_total` back on `margince_sweep_workspaces` — which lives in `backend/internal/compose/jobmetrics.go`, the file this finding says is out of reach — fails the gate by name:

```
metricsuffix_test.go:83: margince_sweep_workspaces_total is a gauge and carries
the _total suffix Prometheus reserves for counters.
```

It could not do that if the walk stopped at `backend/gates`. The sibling gates in this package walk the same way (`integrationmigrateonce_test.go` walks `"."`, `vaultwriters_test.go` walks `"internal"`), and the `minFamilies` floor is what would catch the scan going blind for any other reason: 20 families is far below what the tree carries, and a walk rooted at `backend/gates` would find zero and fail rather than pass quietly.

Rerooting at the parent would take the scan outside the module, which is the thing to avoid here.

Issue #2785, opened 2026-08-26
Commit 7482eb939 (#2766) lacks its author's DCO attestation


luit-nfq:
The squash of #2766 landed on `main` as `7482eb9392534159a60400f486b69294907de6ee` without a `Signed-off-by` trailer, so `dco (main)` is red again.

This is the same cause #2778 documented and closed off at the merge step: replacing the squash message body drops the branch commits' trailers, and they are the only sign-off a squash carries. AGENTS.md step 6 and CONTRIBUTING.md now say so.

**What is owed, and by whom.** `7482eb939` is authored by @LarsGradion, and #2778 made the gate enforce author-only remediation — the attesting commit must be signed off, authored by the same person, under the address its own sentence claims, and descend from the commit it names. So nobody else can clear this one. The line, in any later signed-off commit of theirs:

```text
DCO-Remediation-Commit: I, Lars Jankowfsky <lars@gradion.com>, hereby add my Signed-off-by to commit: 7482eb9392534159a60400f486b69294907de6ee
```

**What is NOT needed:** moving `DCO_MAIN_BASELINE`. That was the only tool available before #2778, and every advance is a stretch of `main` the gate stops asking about. The baseline now sits at `c4a230155` and covers the three previously unsigned squashes, all three passing on their attestations. One attestation closes this hole in place.

Deferred because the attestation cannot be given by anyone else.

tinnhlt-gra:
**Two more, so the census this issue names is now three — and `dco` is red on
every `main-health` run because of them.**

From run `33040273257` (`6c80fbac`), the gate's own output:

```
DCO: commit d600a6f643bec30e213126ab98b5900c5aa1df7d missing Signed-off-by trailer: fix(docs): re-measure the explanation page budget (#2840)
DCO: commit eb0d5dc76122dde993a830fd3f9c38f88f4cbc54 missing Signed-off-by trailer: test(e2e): the six deck use cases, end to end over MCP (#2837)
DCO: commit 7482eb9392534159a60400f486b69294907de6ee missing Signed-off-by trailer: feat(agents): account_coverage names the stakeholder it flags (#2766)
```

The two new ones landed 2026-08-27, three days after #2778 shipped the
attestation path — so the mechanism is in place and the practice that produces
these is not. All three are authored by `lars@gradion.com`.

**Nobody else can clear them.** `check-dco.sh` requires the attesting commit to
be authored by the same person under the same address the trailer names, which
[...]

PR #2929, opened 2026-08-27
fix(integrations): the enrichment fixture selects something an automatic run may buy

tinnhlt-gra:
## main is red on the integration lane, and every branch inherits it

`main-health` run `33065390617`, tip `6709b22b1`:

```
FAILED: integration / integration shard (1/6)
FAILED: integration / integration (RLS + erasure + HTTP e2e)
```

```
persondataenrich_integration_test.go:198: 0 runs exist for a captured counterparty
  with automatic_import on, want 1 — the customer switched capture enrichment on
  and got nothing
persondataenrich_integration_test.go:201: 0 submit jobs were committed, want 1
```

It presents as each author's own red, because CI builds the merge rather than
the head — three unrelated branches hit it inside one hour, each looking local.
The census that settles it is across branches at a point in time, and `main`'s
own run above confirms it.

## Cause

#2923 introduced `runCategories`: an automatic run may take only the categories
the provider gives away, because nobody weighed the purchase. The fixture
connection selects `professional_email` alone, and that costs one email credit.
Measured rather than reasoned — `Free()` is derived from the cost table:

```
Free()                        = [linkedin_profile current_employment job_history]
fixture connection categories = [professional_email]
intersection                  = ∅  →  ErrNothingFreeToBuy  →  skipped run
```

**The rule is right and stays.** The fixture is what has to move.

## The fix

The connection now selects one **priced** category and one **free** one. Both
halves carry weight: the free one is what lets an automatic run exist at all,
and the priced one keeps the narrowing observable — a connection carrying only
free categories cannot tell a run that narrowed to the free set from one that
took everything.

## Why this is a positive control rather than a broken assertion

Those two cases are the only ones in the file that reach `runCategories`. Every
other case is refused earlier — the agent by `triggerFor` in the consumer, the
toggle by connection admission before `queueOne`. I checked that before writing
it: my first draft of this claim said the negatives had been made vacuous, and
that was wrong.

What is true is narrower and worse. The refusal is silent — a skipped run, no
error — so a fixture that drifts out of the free set presents as *"the toggle
bought nothing"*, which is indistinguishable from the toggle being wired to a
constant false. The file's own comment on the second case already names that
risk:

> Without this case the refusal below passes against a gate that admits nobody —
> the toggle could be wired to a constant false and every assertion in this file
> would still be green.

So the guard reads the seeded selection rather than restating it, and runs
**before** `Setup(t)`: this is a contradiction between the fixture and the
provider's cost table, and no schema or seeded row changes the answer.

## Verification

Mutation-verified. Reverting the selection to `professional_email` alone — which
is exactly `main` today — fails on the fixture in 0.00s with no database:

```
--- FAIL: TestPersonCreatedQueuesAnEnrichmentRun (0.00s)
    the fixture connection selects [professional_email], and this provider charges
    for all of them — an automatic run takes only free categories, so it would queue
    nothing and the two cases asserting a run exists would fail as though a toggle broke
```

`make check-backend` green.

**What I could not run:** the integration lane itself. An unrelated Docker
project has held `:15432`/`:16379` on this machine for 27 hours, so `make db-up`
cannot bind and no local run of these two cases is possible. The lane on this PR
is the verification, and it is the same lane that is red on `main`.

## Note on how this landed

#2923 was merged at `10:27:21Z` past a red `ci` **and** past these two exact
integration jobs — they were failing on the PR before the merge. That is the
sixth such merge today; #2785 has the pattern. It also retires a bound I gave
earlier on that issue: I had said the bypasses cost DCO debt and nothing else a
health run can see. This one cost the integration lane.

---
## Summary by cubic
Fixes the integration lane on main by making the enrichment fixture select a category an automatic run can buy. The fixture previously selected only `professional_email` (priced), so automatic runs found nothing free, skipped, and failed two tests. Now it selects one free category (`linkedin_profile`) and one priced one, so runs proceed while still exercising the expected narrowing. A new guard fails the fixture early if the provider starts charging for all selected categories, so a future cost change surfaces as a fixture error rather than a phantom toggle break.

<sup>Written for commit 50e89afab8894daeebc24849a6d3aa294257c986. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/2929?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Tests**
  * Improved integration-test reliability by selecting multiple enrichment categories.
  * Added validation to skip or fail fast when selected categories require payment, preventing misleading test failures.

### DIFF
--- a/backend/internal/compose/integration/persondataenrich_integration_test.go
+++ b/backend/internal/compose/integration/persondataenrich_integration_test.go
@@ -44,17 +44,37 @@ type providerConsumerEnv struct {
 // mode, and wires the consumer over the REAL cross-module binding.
 func setupProviderConsumer(t *testing.T, mode string) *providerConsumerEnv {
 	t.Helper()
+	// Before the database: this is a contradiction between the fixture and the
+	// provider's own cost table, and neither a schema nor a seeded row can
+	// change the answer.
+	seeded := []string{"professional_email", "linkedin_profile"}
+	requireAnAutomaticRunCanBuy(t, seeded)
+
 	e := Setup(t)
 	c := &providerConsumerEnv{env: e, enqueued: new(int), personID: seedSubject(t, e)}
 
+	// One PRICED category and one FREE one, and both halves are load-bearing.
+	// An automatic run may take only what the provider gives away, so a
+	// connection selecting nothing free never queues anything — and the two
+	// cases here that assert a run EXISTS are the only ones that reach that
+	// code, every other case in this file being refused earlier by the actor
+	// or the toggle. Those two are this file's positive control: without them
+	// the toggle could be wired to a constant false and every assertion here
+	// would still be green. The priced category stays because narrowing to the
+	// free set is the behaviour under test, and a connection carrying only free
+	// categories cannot tell narrowing from taking everything.
+	//
+	// Named once above and passed to both the INSERT and the guard, so the
+	// check reads the selection this connection actually carries rather than a
+	// second copy of it that can drift from it silently.
 	if err := database.WithWorkspaceTx(e.Admin(), e.Pool, func(tx pgx.Tx) error {
 		_, err := tx.Exec(context.Background(), `
 			INSERT INTO provider_connection
 			       (id, provider, status, mode, preset, categories, automatic_individual_create)
-			VALUES (gen_random_uuid(), 'surfe', 'connected', $1, 'full',
-			        ARRAY['professional_email'], true)
+			VALUES (gen_random_uuid(), 'surfe', 'connected', $1, 'full', $2, true)
 			ON CONFLICT (provider) DO UPDATE
-			   SET status = 'connected', mode = $1, automatic_individual_create = true`, mode)
+			   SET status = 'connected', mode = $1, categories = $2,
+			       automatic_individual_create = true`, mode, seeded)
 		return err
 	}); err != nil {
 		t.Fatal(err)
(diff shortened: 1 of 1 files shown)


tinnhlt-gra:
**The gap named above is closed.** A second session ran the lane on an isolated
database, both directions, same harness and same command with only the checkout
differing:

```
ON this branch (50e89afab)
  PASS  TestPersonCreatedQueuesAnEnrichmentRun              0.10s
  PASS  TestACapturedContactIsBoughtOnceTheImportToggleIsOn 0.04s

ON main (6709b22b1)
  FAIL  persondataenrich_integration_test.go:143
          the consumer queued nothing for a newly created person: no rows in result set
  FAIL  persondataenrich_integration_test.go:198
          0 runs exist for a captured counterparty with automatic_import on, want 1
        persondataenrich_integration_test.go:201
          0 submit jobs were committed, want 1
```

Both directions matter here. A green run on the branch alone cannot tell the fix
from the harness; the red on `main` through the identical setup is what says the
change is load-bearing. It also picks up a third case — `:143`,
`TestPersonCreatedQueuesAnEnrichmentRun`'s own first assertion — that CI's shard
output had not surfaced separately.

## The lane does not actually require `make db-up`

Worth recording, because the port conflict that blocked me will block the next
person. `infra/docker-compose.dev.yml` hardcodes `15432:5432`, so an unrelated
project holding that port stops `make db-up` for everyone on the machine. But
`lib-testdb.sh` only DEFAULTS to it, and takes an override:

```sh
docker run -d --name margince-itest -p 15433:5432 \
  -e POSTGRES_USER=margince_owner -e POSTGRES_PASSWORD=dev -e POSTGRES_DB=margince \
  pgvector/pgvector:pg16@sha256:…
docker exec -i margince-itest psql -U margince_owner -d margince < scripts/db-init.sql

MARGINCE_TEST_DSN=postgres://margince_owner:dev@localhost:15433/margince \
MARGINCE_TEST_APP_DSN=postgres://margince_app:margince_app_dev@localhost:15433/margince \
  make test-it DIR=backend/internal/compose/integration RUN=…
```

Deliberately NOT pointed at the other project's Postgres: `make test-integration`
creates and drops databases, and doing that inside a container somebody else is
using is not a risk to take on their behalf.

tinnhlt-gra:
**Correcting a sentence in the description above, because acting on it would
cause an outage.**

I wrote that "an unrelated Docker project has held `:15432`/`:16379` on this
machine for 27 hours". The conclusion — don't touch it — was right. The reason
was wrong, and the wrong reason is the dangerous half: it invites the next
reader to reap what looks like somebody else's stray project.

It is not unrelated. It is **this repository's own shared dev stack**, running
under a non-default project name:

```
$ docker compose ls
NAME              STATUS        CONFIG FILES
margince-poc-v1   running(3)    …/margince-next/infra/docker-compose.dev.yml
```

Postgres, Redis and MinIO, from this repo's `infra/docker-compose.dev.yml`, owned
by the PRIMARY worktree — and there are fifteen linked worktrees on this machine
pointed at it. A `docker compose down` there is a fifteen-session outage, and its
failure mode is the worst kind for a shared box: the API keeps answering until it
doesn't, so every session's first suspicion is its own diff.

So the accurate note for anyone hitting the same port conflict is: `make db-up`
cannot bind because the shared stack already holds those ports, and the answer is
a private container on a free port with `MARGINCE_TEST_DSN` / `MARGINCE_TEST_APP_DSN`
pointed at it — which is what was done to verify this PR. Never the shared one,
and never `compose down` to free the port.

The verification of this change is unaffected; only my account of why the port
was busy was wrong.

PR #2931, opened 2026-08-27
fix(httpserver): one posture for a refused write, in the writer rather than in each section

### LINKED ISSUE #461: httpserver: /metrics and /readyz discard response-write failures, inconsistently with the job section
Raised in review of #457 by CodeRabbit, and deferred there as out of scope.

## What

`internal/platform/httpserver/observe.go` writes its exposition and its probe
body with `_, _ = fmt.Fprintf(...)`, discarding every write error — in `Readyz`,
in the outbox/relay/pool sections of `Metrics`, in `writeRuntimeMetrics` and in
`writeOverlayMetrics`.

The job section beside them does not: `writeJobMetrics` returns its error, and
`Metrics` logs it and stops.

## Why the split exists, and where it stops being principled

The split is deliberate as far as it goes. A truncated job section *parses as a
smaller fleet*, which an alert acts on, so it must not be written past. A failed
write on a probe body means the client hung up, and there is no channel left to
report on — the same rationale several `//craft:ignore swallowed-errors`
waivers in this tree state at other response-write sites.

What is not principled is that the distinction is invisible. Two postures sit in
one file with no marker saying which is which or why, so the next section added
there inherits whichever its author copied. The runtime section added in #457
inherited the silent one; nothing said it should.

## Shape of a fix

Either:

- make every section return its write error and have `Metrics` stop at the
  first failure, which is one posture and needs no marker; or
- keep the split and make it explicit — a named helper for the
  hung-up-client case carrying the reason once, so a new section has to choose
  rather than inherit.

The first is cleaner and touches every role on the chassis (`extra`, the AI
counters, the overlay section, `Readyz`). The second is smaller and keeps the
argument.

## Scope

Pre-existing; #457 changed which visibility lines `Readyz` renders, not how any
of them are written. No known failure in production — the endpoints are scraped
over short-lived local connections.

dungbanguyen:
Closes #461.

Two postures sat in `observe.go` with nothing saying which was which:

- the **job section** returned its write error and the handler stopped — a truncated job section *parses as a smaller fleet* and an alert acts on it
- **every other section** discarded its errors — a refused write means the scraper hung up and there is no channel left to report to

Both readings are defensible. What was not is that the distinction was **invisible**: a section added later inherited whichever its author copied. The runtime section from #457 inherited the silent one, and nothing said it should.

## The posture belongs to the writer now

`exposition` remembers the first refused write and makes every write after it a no-op. No section can half-write, nothing downstream is pushed into a socket that is gone, and the handler asks **once**, at the end, whether what it assembled actually left.

It is an `io.Writer`, which is what lets it reach the sections this package does not own — the injected `extra` families and the job section — without changing what they are handed. Their own discarded errors become sound here: the first failure is recorded whether or not the caller looked.

The short-circuit returns the **remembered** error, not nil: a writer that claimed success after failing would let a section that *does* check its writes carry on assembling into nothing.

## Measuring stops too

`gone()` is separate from `printf`'s silence, because they answer different questions. On a scrape whose reader hung up during the first section, the old handler went on to query the outbox, the job table and the mirror for a body nobody would receive. Rendering that into a no-op writer is free; the database reads are not.

## /readyz

Same writer, smaller reason. Nothing is logged there — a probe whose reader is gone has nowhere to report, and the orchestrator's own timeout says so — but a half-written answer is still not worth assembling.

## Not a new contract

`Metrics`' doc comment already said it: *"the error return is reserved for a refused WRITE… The handler logs it and stops rather than writing further sections into a socket that is not there."* One section kept it. Now the writer does, for all of them.

## Held

- nothing is measured for a scrape that is already gone (fails by name if the `gone()` guards are removed — verified)
- the exposition keeps refusing after the first failure, with the remembered error
- `/readyz` stops writing when its reader hangs up
- the existing `TestMetricsStopsWritingWhenTheJobSectionRefusesAWrite` still passes unchanged

`exposition` lives in its own file — `observe.go` passed the 500-line cap.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of disconnected clients during metrics and readiness responses.
  - Stopped writing and collecting additional metrics when a connection is unavailable.
  - Preserved the original connection error across subsequent write attempts.
  - Added logging for incomplete metrics scrapes caused by client disconnects.

- **Tests**
  - Added coverage for disconnected metrics and readiness requests, including verification that unnecessary metric collection stops promptly.

### DIFF
--- a/backend/internal/platform/httpserver/exposition.go
+++ b/backend/internal/platform/httpserver/exposition.go
@@ -1,0 +1,66 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+package httpserver
+
+import (
+	"fmt"
+	"io"
+)
+
+// exposition is the one thing every /readyz and /metrics section writes
+// through, and the reason none of them handles a write error.
+//
+// Two postures used to sit in this file with nothing saying which was which.
+// The job section returned its write error and the handler stopped, because a
+// truncated job section PARSES as a smaller fleet and an alert acts on it.
+// Every other section discarded its errors, on the equally sound reasoning that
+// a refused write means the scraper hung up and there is nothing left to report
+// to. The distinction was invisible, so a section added later inherited
+// whichever its author copied — the runtime section inherited the silent one,
+// and nothing said it should not have.
+//
+// There is one posture now, and it belongs to the writer rather than to each
+// section: the FIRST refused write is remembered and every write after it is a
+// no-op. A section cannot half-write, nothing downstream is pushed into a
+// socket that is gone, and the handler asks once, at the end, whether the
+// exposition it just assembled actually left.
+//
+// It is an io.Writer, which is what makes it reach the sections this package
+// does not own — the injected `extra` families and the job section — without
+// changing what they are handed. Their own discarded errors become sound here:
+// the first failure is recorded whether or not the caller looked.
+type exposition struct {
+	w   io.Writer
+	err error
+}
+
+// Write records the first failure and refuses everything after it.
+//
+// The short-circuit returns the REMEMBERED error, not nil: a writer that
+// claimed success after failing would let a section that does check its errors
+// carry on assembling into nothing.
+func (e *exposition) Write(p []byte) (int, error) {
+	if e.err != nil {
+		return 0, e.err
+	}
+	n, err := e.w.Write(p)
+	e.err = err
+	return n, err
+}
+
+// gone reports that a write was already refused, so this scrape is not going to
+// be delivered. The pure-rendering sections need no such check — printf is
+// already a no-op by then — but a section that MEASURES does: a database read
+// for a socket that is not there is work nobody will see the result of.
+func (e *exposition) gone() bool { return e.err != nil }
+
+// printf is how the sections in THIS file write. The error is not dropped — it
+// is held on the exposition, which is the only place any section's write error
+// was ever going to be actionable.
+func (e *exposition) printf(format string, a ...any) {
+	if e.err != nil {
+		return
+	}
+	_, e.err = fmt.Fprintf(e.w, format, a...)
+}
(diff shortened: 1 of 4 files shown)


dungbanguyen:
Taken. `printf` goes quiet after a refusal, but Go evaluates its arguments first — so `published()` and the three overlay counters were still read to build values for writes that went nowhere.

All four are atomic loads on process-local counters, so the cost is not the reason to fix it. The reason is the rule: **"nothing is measured for a scrape that has gone" is either true of every supplier in this file, or it is a claim a reader has to check one section at a time** — which is exactly the shape this change exists to end. Leaving four unguarded suppliers because they happen to be cheap would have recreated the invisible distinction in miniature.

The relay block is now behind the same `!out.gone()` guard its neighbours have, and `writeOverlayMetrics` returns after the lag section — which is the section that may be what discovers the writer is gone.

`TestNoCounterIsReadForAScrapeThatHasAlreadyGone` reads whether each supplier ran at all, so the guards cannot be dropped back out silently.

dungbanguyen:
Taken — and it is the same rule I had just applied to the exposition's counters, left unapplied one function over. After the `ai` line is refused `body.gone()` is true and `printf` goes quiet, but `embedState` still ran to resolve a marker for a body that cannot be delivered.

The probe returns there now, and `TestReadyzStopsWritingWhenItsReaderHangsUp` asserts the callback was **not reached** rather than only that the writes stopped — verified from the other end: dropping the guard fails it by name.

dungbanguyen:
Taken. `ReadMemStats` and `NumGoroutine` ran before any write could report the reader was gone — and this is the exposition's **first** section, so nothing earlier can discover it. The section has to probe for itself.

Its first family's header is that probe: two static lines that cost nothing, sent before anything is measured. If they land the reader is there and the readings are worth taking; if they do not, the section returns without paying a stop-the-world pause on behalf of a scrape nobody will receive — and the process does not pay it either, on every replica, while nobody is reading.

**Not asserted, and I want to be straight about why.** `printf` short-circuits after a refusal either way, so the body is byte-identical with and without the guard: there is nothing observable from outside the section to test. Adding a case that passes in both directions would read as coverage of a property nothing holds. What the existing case asserts is unchanged — nothing *after* this section measures anything — and the `hangUp` writer now says how many writes it accepts, so the probe's own two are what its count is about.

PR #2946, opened 2026-08-27
perf(test): one fixture where the cases share one, and a reason where they do not

### LINKED ISSUE #535: Table-driven integration tests re-seed Postgres per case (~44s, and it hides that the property is pure)
## Summary

Several table-driven integration tests call their database harness **inside** `t.Run`, so every case re-seeds Postgres to vary inputs that are pure in-memory arguments. The measured cost is **~44s across 87 parent tests / 405 subtests** — 9% of the lane's summed 493s.

## The shape

`activities.TestSendEmailRefusesAMessageWhoseAddresseeLineIsEmpty` is the clearest instance, at **5.15s** for three cases:

```go
for _, tc := range []struct{ name string; recipients, cc []string }{
    {name: "nobody at all", ...},
    {name: "cc only", ...},
    {name: "the one addressee is also cc'd in another case", ...},
} {
    t.Run(tc.name, func(t *testing.T) {
        e := setupSend(t)          // <- dials Postgres, creates a workspace, seeds users
        anchor := e.seedAnchor(t, "", "")
        ...
    })
}
```

What varies between cases is `Recipients` and `Cc`. What is re-created between cases is a workspace, its users, and an anchor activity. The subject under test — To/Cc normalization deciding the addressee line is empty — does not read any of it.

## Why it is worth fixing beyond the seconds

The three cases are asserting one pure property, and the fixture makes that hard to see. A reader has to check whether each case depends on the seeded state before concluding they do not. Hoisting the setup out of the loop states the answer in the code: one fixture, three inputs, one property.

## Suggested fix

Two options, in preference order:

1. **Hoist the harness above the loop** where the cases share a fixture. Smallest diff, keeps the coverage exactly as it is, and reads better.
2. **Move the case table to the unit lane** where the property really is pure and the function is reachable without a store — the addressee-line normalization above is a candidate. See #536 for the general version of that judgement.

Do not do this blindly: a case that mutates the fixture, or that asserts a constraint/trigger, genuinely needs its own seed. The 87 parents should be triaged, not rewritten in bulk.

## Finding the rest

```
# parent tests with subtests, ranked by cost, from a -v lane run
awk '/^--- (PASS|FAIL): /{cur=$3; d=$4; gsub(/[()s]/,"",d); dur[cur]=d}
     / +--- (PASS|FAIL): /{kids[cur]++}
     END{for(t in dur) if(kids[t]>0) print dur[t], kids[t], t}' lane.log | sort -rn
```

Then check each hit for a `setup…(t)` call inside the `t.Run` closure.

## Measurement context

Profiled on `main` (`d140000d`), full lane, 1828 tests / 493s summed: 87 parents with subtests = 43.8s (9%); 1741 without = 449.1s (91%).

dungbanguyen:
Closes #535.

Table-driven integration tests that call their harness **inside** `t.Run` re-seed Postgres to vary inputs that are pure in-memory arguments. The cost is real, but the reading is the point: a reader has to check whether each case depends on the seeded state before concluding it does not, and the fixture is what makes that work.

## Hoisted where the cases genuinely share one

| where | what changed |
| --- | --- |
| `activities` — the addressee-line refusals | the ticket's named example. What varies is To/Cc; the subtraction that empties the line reads none of the seeded state. **5.15s → 0.08s** |
| `comms` — the unknown-outcome property | both halves stage their own delivery and assert by that id |
| `overlay` — the mirrored-type round trip | one fixture for the five update cases |

Two details worth naming:

- the outbound count in the activities case becomes **cumulative**, which is a stronger assertion rather than a weaker one: a case that staged a delivery would be caught by its own assertion and by every later case's
- the overlay **archive** loop keeps a fixture of its own, and that is not tidiness — it seeds a *second* person and a second organization, and `firstListedID` cannot say which of two records at one path it means

## Triaged, not rewritten

The remaining hits keep their per-case seed, and now **say why in the case itself** — which is the same fix as hoisting, answering the same question from the other direction.

| where | why the seed is load-bearing |
| --- | --- |
| `comms/store_recordsent` | a staged message identity is unique per workspace. Hoisting it *passes*, and only because the previous case stamps the identity away first — a dependency between cases that reads as none. **Found by trying it**: the second loop fails on the conflict outright. |
| `activities/messageidentity_absorb` | every case seeds a survivor under the same minted identity, and the reconcile *searches* for its echo — an earlier case's bystander is a candidate for a later case's survivor |
| `privacy/auditlogboundary` | each case scrubs the same subject and asserts the pre-scrub image is gone; shared, the second reads an image the first already put behind a boundary, and a verb that does nothing passes |
| `compose/integration/csvimportbyid` | each case seeds a different ambiguity into one table, and the subject is which record an id resolves to among them |

Every hoist verified by running the suite it belongs to; every refusal to hoist verified the same way where it could be (the comms one is a real failing run, not an argument).

## Not in scope

The issue's census names 87 parents with subtests. This PR covers the ones whose harness call sits inside the loop — 27 sites, of which 5 are now hoisted (27 → 22) and 4 more carry the reason their seed cannot move. The rest of the 87 spend their time elsewhere and are not this shape.

## Summary by CodeRabbit

* **Tests**
  * Improved integration test reliability by isolating test cases and reducing dependence on execution order.
  * Reused shared fixtures where appropriate while preserving independent data setup for each scenario.
  * Added comments clarifying fixture isolation and the distinct behaviors covered by ambiguity, identity, and audit-boundary tests.

### DIFF
--- a/backend/internal/compose/integration/csvimportbyid_integration_test.go
+++ b/backend/internal/compose/integration/csvimportbyid_integration_test.go
@@ -99,6 +99,11 @@ func TestCSVImportByIDUpdatesTheNamedCompany(t *testing.T) {
 
 // The three cases that destroyed data under name matching, each harmless here.
 func TestCSVImportByIDIsImmuneToEveryAmbiguityThatBrokeNameMatching(t *testing.T) {
+	// Each case seeds a DIFFERENT ambiguity into the same table, and the
+	// subject is which record an id resolves to among them. A shared fixture
+	// would leave the first case's two Kestrel Data rows in place while the
+	// third case is about a trading name that is somebody else's registered
+	// name — a different ambiguity from the one it is named for.
 	t.Run("two companies sharing a name", func(t *testing.T) {
 		e := setupImportApp(t)
 		wanted := createOrgReturningID(t, e, map[string]any{"display_name": "Kestrel Data"})
--- a/backend/internal/compose/integration/overlay/overlay_write_shadow_integration_test.go
+++ b/backend/internal/compose/integration/overlay/overlay_write_shadow_integration_test.go
@@ -409,9 +409,17 @@ func TestOverlayWriteShadowsRoundTripEveryMirroredType(t *testing.T) {
 		{"lead", "/v1/leads", "9304", map[string]any{"full_name": "Grace Lead"}, apptest.AnyMap{"full_name": "Grace Lead 2"}, "full_name", "Grace Lead 2"},
 		{"activity", "/v1/activities", "9305", map[string]any{"kind": "call", "subject": "Intro Call"}, apptest.AnyMap{"subject": "Intro Call 2"}, "subject", "Intro Call 2"},
 	}
+	// One fixture for the five update cases: each seeds a different type at a
+	// different path, so firstListedID below still names exactly the record
+	// that case seeded.
+	//
+	// The archive loop keeps a fixture of its own, and that is not tidiness:
+	// it seeds a SECOND person and a second organization, and firstListedID
+	// cannot say which of two records at one path it means.
+	updateEnv := setupOverlayWrite(t)
 	for _, tc := range updates {
 		t.Run("update/"+tc.entityType, func(t *testing.T) {
-			e := setupOverlayWrite(t)
+			e := updateEnv
 			e.seed(t, tc.entityType, tc.externalID, tc.seedFields)
 			id := firstListedID(t, e.AppEnv, tc.path)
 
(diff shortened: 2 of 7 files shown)


dungbanguyen:
Right, and the correction matters more than the wording: a dispatch **writes** — the status, the park reason, the in-flight marker — so "a store which neither of them changes" was false. What isolates the two halves is that each stages its own delivery and reads it back by that id.

Said correctly now. A reader checking whether a hoist is safe has to be able to check the reason it gives, and this one gave a reason that was not the one holding.

PR #2954, opened 2026-08-27
fix(audit): an absent image cannot reach the column as the JSON scalar null

### LINKED ISSUE #589: storekit.marshalOrNil only detects an untyped nil, so an absent audit image can silently become JSON null
`marshalOrNil` writes SQL NULL when an audit image is nil:

```go
func marshalOrNil(v any) ([]byte, error) {
	if v == nil {
		return nil, nil
	}
	return json.Marshal(v)
}
```

A caller that declares its image as a concrete map and passes it in while nil defeats that: `var beforeImage map[string]any` arriving in an `any` parameter is a **typed** nil, so `v == nil` is false, `json.Marshal` runs, and the column gets JSON `null` instead.

The consequence is quiet. The obvious query for "this write had no prior state" is `before IS NULL` — the spelling `retention_integration_test.go` already uses — and it misses every row written that way. `coalesce(before, after)` is wrong for the same reason: it returns the JSON null rather than falling through.

Three stores had it, fixed in #514: capture's own-domain registry, capture's freemail classification, identity's onboarding state. `people/coldstartprofile.go` looks identical but assigns in both branches, so it was never affected.

Nothing stops the next caller repeating it — the correct shape is invisible at the call site, and the wrong one compiles and passes tests that read `after`.

Options, in rough order of preference:

1. Harden `marshalOrNil` to treat a nil map/pointer/slice as absent (a `reflect.Value.IsNil` check guarded by `Kind`). This changes the stored representation for any existing caller with the same bug, so it wants a sweep of what currently writes JSON null before it lands.
2. Change the `Audit`/`AuditWithEvidence` signature so an absent image cannot be expressed as a typed nil.
3. A fitness test that fails on any `audit_log` row whose `before` or `after` is the JSON scalar `null`.

Found while landing #514 (ADR-0082/A127); not fixed there because option 1 changes audit representation for every module at once, which is wider than that PR.

dungbanguyen:
Closes #589.

SQL `NULL` and the four bytes `null` look alike and answer differently. Every "there was no prior state" query in this tree is `WHERE before IS NULL`, and a column holding `null` **is not null** — it misses the row, silently, in exactly the place a reader is trying to establish that nothing preceded a change. `coalesce(before, after)` is wrong for the same reason: it returns the JSON null rather than falling through.

## Option 1 is already in

The issue's preferred fix — harden `marshalOrNil` against a typed nil — landed with the audited-update work (#2653) as `storekit.AbsentImage`, and it is unit-tested across every kind of nil that can carry one.

## What was left is what does not go through that door

`AbsentImage` binds the seam, not the table. The approvals module `INSERT`s its own audit row, and a caller that marshals an image itself hands over bytes the seam never sees. So this is the issue's option 3, as a constraint rather than a test — it binds every future writer instead of sampling what one suite happened to write.

The predicate reads false only for the scalar: `before <> 'null'` is **NULL** when `before IS NULL`, and a CHECK admits NULL. An absent image passes, a `null` image is refused.

## NOT VALID, and it stays that way

`audit_log` is append-only, so rows written before storekit was hardened cannot be rewritten — and validating would fail the migration on exactly those installations, blocking an upgrade over history nobody can change. That is also the sweep the issue asked for before landing option 1, answered by not needing one: the constraint is for the next write.

The test template carries 0 offending rows today.

## Held

`TestAnAuditImageCannotBeTheJSONScalarNull` inserts **directly**, which is the shape that does not pass storekit at all.

The honest absence is asserted **first**: a constraint that refused an absent before-image would have broken the case every reader's `IS NULL` is written for rather than the case they miss. And the refusal is matched by constraint name, so a case cannot pass because something else rejected the row.

Mutation-checked — relaxing the CHECK to `true` fails both arms by name.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved audit log validation so missing audit images are represented as absent values.
  - Prevented invalid JSON `null` values from being stored for “before” and “after” audit images.
  - Continued supporting valid existing audit records while enforcing consistency for new data.

- **Tests**
  - Added integration coverage confirming that absent images are accepted and JSON `null` values are rejected.

### DIFF
--- a/backend/internal/compose/auditimagenull_integration_test.go
+++ b/backend/internal/compose/auditimagenull_integration_test.go
@@ -1,0 +1,76 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+//go:build integration
+
+package compose
+
+// An absent audit image reaches the column as SQL NULL, and the JSON scalar
+// `null` cannot get in at all.
+//
+// The two look alike and answer differently. Every "there was no prior state"
+// query in this tree is `WHERE before IS NULL`, and a column holding the four
+// bytes `null` is not null — it misses the row, silently, in exactly the place
+// a reader is trying to establish that nothing preceded a change.
+//
+// storekit.AbsentImage answers this correctly at the door it owns. The
+// constraint is here because not every writer goes through that door: the
+// approvals module INSERTs its own row, and a caller that marshals an image
+// itself hands over bytes the seam never sees.
+
+import (
+	"context"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/jackc/pgx/v5"
+
+	"github.com/margince/margince/backend/internal/compose/integration"
+	"github.com/margince/margince/backend/internal/platform/database"
+	"github.com/margince/margince/backend/internal/shared/kernel/ids"
+)
+
+// writeAuditImage inserts one audit row with the given before/after literals,
+// as a direct INSERT — the shape that does not pass storekit at all, which is
+// the whole reason the refusal lives in the table.
+func writeAuditImage(t *testing.T, e *integration.Env, before, after string) error {
+	t.Helper()
+	return database.WithWorkspaceTx(e.Admin(), e.Pool, func(tx pgx.Tx) error {
+		_, err := tx.Exec(context.Background(), `
+			INSERT INTO audit_log (actor_type, actor_id, action, entity_type, entity_id,
+			                       before, after, occurred_at)
+			VALUES ('human', 'user:'||$1::text, 'update', 'person', $2, nullif($3,'')::jsonb, nullif($4,'')::jsonb, $5)`,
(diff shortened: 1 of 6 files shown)


dungbanguyen:
**The constraint caught a real one on its first run**, in exactly the shape the issue describes.

Both field-history seeders `INSERT INTO audit_log` directly, so they go around storekit and its `AbsentImage` answer. They marshalled the image unconditionally — and `json.Marshal` of a nil map is the four bytes `null`. The `export` row, which genuinely has no before-image, was therefore written carrying the JSON scalar, and every `WHERE before IS NULL` reader misses it while the column claims to carry an image.

That is the issue's own argument, found in the tree rather than reasoned about: *"the correct shape is invisible at the call site, and the wrong one compiles and passes tests that read `after`."* These tests read `after`.

Fixed with one helper that answers nil bytes — SQL NULL — for an image that says nothing, and says why a fixture writing the row itself has to give the same answer the door gives.

The whole `TestFieldHistory*` family passes.

Issue #2924, opened 2026-08-27
uat: Back from a record reaches the list's sort without its search — one observation, 72 local runs green


tinnhlt-gra:
## What failed

`main-health` run `33060231911`, tip `aa7167ea6`, job `uat (main)`. One test of
143:

```
e2e/history.spec.ts:96 › Forward returns to the list Back left

Error: expect(page).toHaveURL(expected) failed
  Expected pattern: /[?&]q=brandt/
  Received string:  "http://localhost:4317/#/companies?sort=-created_at"
  13 × locator resolved to <html lang="de" data-theme="light">…</html>

> 111 |   await page.goBack();
      |   await expect(page).toHaveURL(/[?&]q=brandt/);
```

Back from the record landed on the list with its **sort** but without its
**search**. Since the search is written with `history.replaceState`, that means
the entry the reader returned to never carried `q` — not that it lost it.

## What I could not reproduce

Against `origin/main` (`758233b93`), built and served the same way the lane does:

| what | result |
|---|---|
| the case alone, `--repeat-each=12 --workers=4` | 12 passed |
| the case alone, `--repeat-each=30 --workers=8` | 30 passed |
| full suite, `--workers=8`, three times | 143 passed each |

72 executions, no failure. So this is one observation, and I am filing it rather
than guessing at it.

## The obvious explanation is wrong, which is why this needs a real look

The natural reading is that the test's synchronisation point is too weak: it
waits for the search REQUEST and then clicks, so the address might not be written
yet. Its sibling three tests above does assert the address first, with a comment
naming the rule:

```js
// The request is the proof the dial reached the server; the address is the
// proof it can be returned to.
await page.waitForRequest(…);
await expect(page).toHaveURL(/[?&]q=brandt/);   // <- present at line 41
```

`Forward returns to the list Back left` omits that line and clicks straight
after the request. But reading the writer, the order is the other way round:
`setQuery` (`screens/listquery.tsx:592`) calls `setParams` → `replaceParams`,
which does `history.replaceState` **synchronously** inside the debounce
callback, then `announceAddressChanged()` — and the re-render that follows is
what issues the request. The address is written BEFORE the request exists, so
waiting on the request should already imply it.

**Which means adding that assertion would probably be a placebo, and might be
worse than one.** `toHaveURL` polls, so if the address is briefly wrong and then
corrected it would wait out the transient and click on a good URL — turning a
visible failure into a green run over a live defect. That is the "raise the
[...]

PR #2958, opened 2026-08-27
fix(integration): edgereverse names AnyMap from the package that declares it

tinnhlt-gra:
## main does not compile

`main-health` run `33081704383`, tip `c1584ea00` — **10 failed, 6 success, 1
skipped**. Nine of the ten are one cause:

```
internal/compose/integration/edgereverse_integration_test.go:58:11: undefined: apptest.AnyMap
FAIL github.com/margince/margince/backend/internal/compose/integration [build failed]
```

| job | why |
|---|---|
| `backend gate (main)` | golangci's typecheck reads the `integration` tag |
| `integration / integration shard (1..6/6)` | the package will not build |
| `integration / integration (RLS + erasure + HTTP e2e)` | fan-in over the shards |
| `integration / integration unit coverage` | same package |
| `sonarcloud (main)` | skipped — its dependency never produced coverage |
| `uat (main)` | **unrelated** — #2924, third occurrence |
| `dco (main)` | **unrelated** — `188598a90`, #2785 |

## The fix

`AnyMap` is declared in `integration/anymap.go` as a package-level alias:

```go
type AnyMap = map[string]any
```

`edgereverse_integration_test.go` is in that same package, so it needs no
qualifier. Eight sites, and a tree-wide grep confirms they are the only
`apptest.AnyMap` anywhere.

## Nobody could have caught this

Two changes that each compiled on its own base:

- one moved the JSON alias so that it is **not** the app fixture's;
- the other added this file, naming it where it used to live.

Their branches share no line. The merge is the first tree in which both are true,
and `apptest.` stayed valid right up to the moment the other side landed. Every
one of the nine commits in this window merged with **no red check** — I checked
each before writing this, precisely because "somebody bypassed a gate" has been
the right answer several times today and is the wrong answer here.

This is the same class as the `gate-inventory.md` collision in #2944 earlier: two
PRs that share no file, colliding on a fact that only exists in the merged tree.
There the shared fact was a count; here it is a symbol's package.

## Verification

`make check-backend` green, including the golangci typecheck that was failing.
`go vet -tags integration ./internal/compose/integration/` clean — that is the
build the shards do.

Not verified locally: the integration lane itself. The port that `make db-up`
needs is held by this repository's own shared dev stack (see the note on #2929),
and the failure here is a **compile** error, which `go vet -tags integration`
answers completely.

## What this does not fix

`uat` and `dco` stay red, and neither is in reach of this change:

- `uat` — #2924, now its **third** occurrence, and the second failed the test
  that already carries the assertion once proposed as its fix. It needs the two
  address writers instrumented, not another test change.
- `dco` — `188598a90` needs a `DCO-Remediation-Commit` trailer from its own
  author; `check-dco.sh` accepts one from nobody else.

---
## Summary by cubic
Fixes a merge collision that broke compilation of the `integration` package on `main`, failing nine CI jobs. `edgereverse_integration_test.go` referenced `apptest.AnyMap`, but `AnyMap` is declared in the same `integration` package, so the qualifier was invalid after another change moved the alias. Removes the `apptest.` prefix from all eight call sites.

<sup>Written for commit a7b3d1e0dbdf43672102de58cd4cefac42822220. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/2958?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Tests**
  * Updated integration test request data handling without changing test behavior or assertions.

### DIFF
--- a/backend/internal/compose/integration/edgereverse_integration_test.go
+++ b/backend/internal/compose/integration/edgereverse_integration_test.go
@@ -55,23 +55,23 @@ func seedEmploymentOverHTTP(t *testing.T, e *apptest.AppEnv) linkedPair {
 	role := seededEdgeRole
 	var person personRecord
 	if status := e.Call(t, "POST", "/v1/people",
-		apptest.AnyMap{"full_name": "Ada Employed"}, nil, &person); status != 201 {
+		AnyMap{"full_name": "Ada Employed"}, nil, &person); status != 201 {
 		t.Fatalf("create person → %d", status)
 	}
 	var org struct {
 		ID string `json:"id"`
 	}
 	if status := e.Call(t, "POST", "/v1/organizations",
-		apptest.AnyMap{"display_name": "Employer GmbH"}, nil, &org); status != 201 {
+		AnyMap{"display_name": "Employer GmbH"}, nil, &org); status != 201 {
 		t.Fatalf("create organization → %d", status)
 	}
-	return linkedPair{person: person.ID, org: org.ID, edge: linkEdge(t, e, apptest.AnyMap{
+	return linkedPair{person: person.ID, org: org.ID, edge: linkEdge(t, e, AnyMap{
 		"kind": "employment", "person_id": person.ID, "organization_id": org.ID,
 		"role": role, "source": "manual",
 	})}
 }
 
-func linkEdge(t *testing.T, e *apptest.AppEnv, body apptest.AnyMap) relationshipRecord {
+func linkEdge(t *testing.T, e *apptest.AppEnv, body AnyMap) relationshipRecord {
 	t.Helper()
 	var rel relationshipRecord
 	if status := e.Call(t, "POST", "/v1/relationships", body, nil, &rel); status != 201 {
@@ -264,10 +264,10 @@ func TestEndToEnd_reversingAProjectCompanyRefusesByNameAndWritesNothing(t *testi
 		Version int64  `json:"version"`
 	}
 	if status := e.Call(t, "POST", "/v1/organizations",
-		apptest.AnyMap{"display_name": "Client GmbH"}, nil, &org); status != 201 {
+		AnyMap{"display_name": "Client GmbH"}, nil, &org); status != 201 {
 		t.Fatalf("create organization → %d", status)
 	}
-	if status := e.Call(t, "POST", "/v1/projects", apptest.AnyMap{
+	if status := e.Call(t, "POST", "/v1/projects", AnyMap{
 		"name": "Joint rollout", "organization_id": org.ID, "source": "manual",
 	}, nil, nil); status != 201 {
 		t.Fatalf("create project → %d", status)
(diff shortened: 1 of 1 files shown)


tinnhlt-gra:
**Naming the second PR, which the description above left blank, and one detail
that makes this sharper than a merge-order accident.**

A second session reached the same diagnosis independently and identified the
adder:

```
#2930  b6a1af7fd  feat(history)   ADDED edgereverse_integration_test.go, using apptest.AnyMap
#2951  789750c29  refactor(test)  MOVED AnyMap out of apptest into integration.AnyMap
```

Confirmed against `origin/main` rather than a checkout: `apptest/` contains no
`AnyMap` at all, and `edgereverse_integration_test.go` is the only file in the
tree still naming one.

## The detail worth keeping

#2951's own docblock states its census:

> *fourteen files were importing that package for this alone*

Fifteen, as of the merge. That number was **correct when it was written and false
when it landed** — #2930's file did not exist in the tree #2951 was measured
against, and #2951's rename did not exist in the tree #2930 was tested against.

So this is not a stale census that somebody should have refreshed. It is a census
that could not have been right: the fifteenth caller was written concurrently, and
no measurement of either branch could see the other. That is a different failure
from the `gate-inventory.md` collision in #2944 — there the count was a committed
artifact both PRs wrote to, so the conflict was at least *representable*. Here the
number lives in prose, nothing compares it to anything, and the only thing that
notices is a compiler run over the merged tree.

## Note on the import

`apptest` stays imported. The file has 14 remaining `apptest.` references
(`AppEnv`, `SetupApp` and friends) — only the `AnyMap` qualifier was wrong. `go
vet -tags integration` passing is the proof: an unused import would have failed
it.

tinnhlt-gra:
Closing as redundant: #2956 landed the identical fix first (`5ea9ba1ea`, *edgereverse follows AnyMap to its new home*), and this branch's diff against `main` is now empty.

Same eight sites, same qualifier. No disagreement to record — just two people fixing one break inside twenty minutes, which is what a cheap and correct diagnosis costs. A third was nearly filed as well.

The lesson is procedural rather than technical, and it is mine: **say you are taking it before you take it.** When a break is one file and eight mechanical sites, the race is to the fix rather than to the diagnosis, so announcing intent is worth more than working fast. I announced to one session and not to the repository.

What survives here and is not in #2956 is the account of *why* it happened, which is worth keeping findable — #2930 added the file using `apptest.AnyMap` while #2951 moved `AnyMap` out of `apptest`; each compiled on its own base, and #2951's docblock census of "fourteen files" was correct when written and false when it landed. That number was not stale, it was unrightable: the fifteenth caller was authored concurrently and no measurement of either branch could see the other.

PR #2959, opened 2026-08-27
fix(ai): navigation is not evidence

### LINKED ISSUE #596: Site read: page chrome reaches the reader as legal-entity evidence, and language switchers become facts
Two extraction-quality problems, both visible to the reader on the onboarding review, both from one live read of a real multi-entity site.

**1. Navigation chrome presented as evidence**

The legal-entity decision asks the reader to pick which company the installation is for, and shows the evidence behind each candidate. For one candidate the evidence snippet was the site's navigation menu, flattened:

> Imprint | Gradion | Gradion Solutions Products Industries About English Contact Us Solutions Products Industries About English Deutsch Tiếng Việt ไทย العربية 日本語 Contact Us Imprint …

The surface is doing its job — it is showing, verbatim, what the read actually captured. The snippet just carries nothing that supports the candidate, which undermines the one decision on that screen: the reader is asked to choose between entities on evidence that does not distinguish them.

**2. Language switchers extracted as company facts**

The same read produced facts of category `language` whose values were the switcher labels — `Tiếng Việt`, `ไทย`, `العربية`, `日本語` — each cited to `/en/about` and each scored 100. They are true statements about the *website*, not about the business, and they are pre-selected into the company context along with everything else.

**Suggested direction**

Boilerplate suppression before a snippet is stored as evidence — a block repeated across most pages of a crawl is chrome, and a crawl already knows what it fetched. For the language lane specifically, a switcher's own labels are recognisable from their position in the same repeated block.

Both are read-side; the review surface renders whatever the read hands it, which is the correct behaviour for that surface.

dungbanguyen:
Closes #596.

One live read of a multi-entity site produced two findings that were true and useless.

**A legal-entity candidate cited to the mega-menu.** The evidence snippet was the site's flattened navigation — a passage that names *every* candidate the site has, so it distinguishes none of them — on the one screen whose whole purpose is choosing between them.

**Language-switcher labels as company facts.** `Tiếng Việt`, `ไทย`, `العربية`, `日本語`, each of category `language`, each scored 100, each cited to `/en/about`. True about the **website**, not about the business, and pre-selected into the company context with everything else.

## One rule, because they are one mistake

A finding whose evidence lies inside a block the crawl carries on most of its pages is supported by the site's *furniture* rather than by anything the company said.

## Why it runs where it does

The profile lane already strips that chrome before it builds its prompt. The per-page fact lane **cannot** — it reads a page as it commits, which is before the crawl it would have to be measured against exists. So it answers for what it found once the corpus is whole, which is the issue's own suggestion: suppression before a snippet is stored as evidence.

The blocks come from the **same measurement** the profile lane's excerpt uses (`stripSharedPrefixBlocks`, which is `stripSharedPrefix` also answering what it cut) rather than from a second reading of what furniture is.

## The failure direction is deliberate

A crawl too small or too large to measure yields no blocks and nothing is dropped. A missing finding is harder to notice than a noisy one.

## Dropped, not stripped of its evidence

A candidate with no support is worse on that screen than one fewer candidate: the decision is which entity this installation is *for*, and an option a reader cannot judge makes the judgement harder. Every drop is reported through the same trail a gate refusal takes, so nothing disappears without saying so.

A finding with **no** evidence at all is left alone — it has a reason of its own, and this is not it.

## Held

Three cases, on a fixture built from the real menu that prompted the issue: the drops happen and are reported; a crawl with no measurable chrome keeps everything; a finding with no evidence is not judged here. Mutation-checked — disabling the suppression fails the first by name.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved extraction accuracy by filtering out page facts supported only by shared navigation or menu content.
  * Preserved valid page evidence while retaining findings without evidence for separate citation handling.
  * Added detailed reporting for findings removed as navigation-derived content.
  * Applied consistent behavior across streaming and non-streaming extraction.

### DIFF
--- a/backend/internal/compose/siteboilerplate.go
+++ b/backend/internal/compose/siteboilerplate.go
@@ -92,10 +92,24 @@ const (
 // Returns the pages in the same order, with Text trimmed where a shared
 // opening was found. The input is not modified.
 func stripSharedPrefix(pages []crawlPage) []crawlPage {
+	out, _ := stripSharedPrefixBlocks(pages)
+	return out
+}
+
+// stripSharedPrefixBlocks is stripSharedPrefix, and also answers WHICH blocks
+// it cut.
+//
+// The blocks leave with the pages because two callers need the same answer for
+// opposite reasons. The profile lane wants the text without them, so its
+// excerpt budget is spent on the company's own words. The evidence filter
+// wants the blocks themselves, so a finding cited to nothing but the menu can
+// be recognised — and a second measurement, taken separately, would be a
+// second definition of chrome free to disagree with this one.
+func stripSharedPrefixBlocks(pages []crawlPage) ([]crawlPage, []string) {
 	out := make([]crawlPage, len(pages))
 	copy(out, pages)
 	if len(pages) < boilerplateMinPages {
-		return out
+		return out, nil
 	}
 	total := 0
 	for _, page := range pages {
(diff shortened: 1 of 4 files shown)


dungbanguyen:
Taken, and it is the failure this lane is most at risk of. A menu repeats the site's own vocabulary, so a phrase sits in **both** the navigation and the prose — `Contact Us`, a company's name, a product's. Dropping on the chrome match alone took real evidence with it, which is exactly the over-trimming `siteboilerplate.go` warns about: *a missing fact is harder to notice than a noisy one.*

A finding cited to the page's own words is a finding however many times the header also says them. The test is now **chrome AND NOT prose**, against what that page still says once its furniture is off — which `stripSharedPrefixBlocks` already hands back, so nothing is measured twice.

The exemption is about the **page**, not the phrase: the same words on a page that does not state them are furniture again. That is one of the four cases.

**Asserted against the rule itself rather than through a crawl**, and that was a correction along the way. My first attempt built a fixture where the same text sat in menu and prose — but the boilerplate measurement has thresholds of its own (`boilerplateMinSurvivingRunes`, the stub share), the fixture did not trip them, and the case passed with the fix *and* without it. A test that cannot fail is worse than none, so the rule is now exercised directly. Verified: dropping the prose clause fails it by name.

PR #3074, opened 2026-08-28
fix(ci): close the new_security_rating gate scheduled.yml keeps failing

tinnhlt-gra:
## Summary
- `sonarcloud quality gate (main)` (the `scheduled` workflow) has failed on `new_security_rating GT 1 (actual 3)` for the last four scheduled runs — six open VULNERABILITY issues, all in new code, none in a file the merge-time scan touches, so no PR has ever hit this.
- Two genuine fixes in `.github/workflows/scheduled.yml`: `pnpm install --frozen-lockfile` in the "use cases, driven by a model" job now carries `--ignore-scripts` (S6505), and the Claude CLI install is now pinned to `2.1.236` (S8543) instead of tracking whatever npm resolves at run time.
- The remaining four issues are false positives, waived in `sonar-project.properties` with the same reasoning already on record for `ci.yml`/`main-health.yml`: `scheduled.yml` is the third workflow that runs `pnpm exec playwright install`, which S6505 can't tell apart from an npm lifecycle-script vector; the Claude CLI's own `postinstall` is load-bearing (places the binary `bin/claude` execs), so `--ignore-scripts` there would ship a CLI that can't run; and `e2e/llm/check.py`'s two `open(path)` calls trip S8707's "an LLM chose this CLI argument" model even though the assistant under test never reaches this script — every path it opens is built by `scripts/e2e-llm.sh` from a fixed scenario glob and its own `mktemp` workdir.

## Test plan
- [x] `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/scheduled.yml'))"` — YAML still parses
- [x] Confirmed no workspace `package.json` declares `postinstall`/`prepare`, so `--ignore-scripts` on the root `pnpm install` is safe
- [x] `git push` pre-push hook (`craft static --strict`) passed — no Go files changed
- [ ] Next scheduled run of `.github/workflows/scheduled.yml` (or a manual `workflow_dispatch`) shows `sonarcloud quality gate (main)` green

---
## Summary by cubic
Fixes the `scheduled` workflow's failing SonarCloud security gate — the last four runs missed the `new_security_rating` target — and the PR's own `new_coverage` rejection from `e2e/llm/check.py`.

**Bug Fixes**
- Adds `--ignore-scripts` to the root `pnpm install`; safe because no workspace package declares `postinstall` or `prepare`.
- Replaces the unpinned global Claude CLI install with a pinned `2.1.236` composite action that installs project-local and runs `npm audit signatures`, failing closed if the registry signature doesn't verify; its own `postinstall` is load-bearing, so `--ignore-scripts` would ship a CLI that can't run.
- Waives S6505 on the Playwright install step, the same false positive already waived for `ci.yml` and `main-health.yml`; the CLI action keeps its own separate waiver since a whole-workflow waiver would hide genuine findings elsewhere in `scheduled.yml`.
- Waives S8707 on `e2e/llm/check.py`; its `open()` calls route through a shared `_open_checked` helper that refuses paths outside the actual repo root or the system temp dir — the root previously resolved to `e2e/`, wrongly refusing scenarios elsewhere in the repo — and the assistant under test never reaches this script.
- Excludes `e2e/llm/check.py` from the coverage gate; no Python coverage report is wired into the scan, and the stdlib-only judge script shouldn't gain a coverage dependency.

<sup>Written for commit 1fbeaf972fc018e77a44ade9d5a3557ded963b77. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/3074?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Chores**
  * Updated scheduled automation with a pinned Claude CLI version.
  * Improved dependency installation reliability for automated frontend checks.
  * Added a reusable, verified process for installing the required CLI tools.

* **Bug Fixes**
  * Strengthened end-to-end test file handling to prevent access outside approved locations.

* **Refactor**
  * Updated code-quality scanning rules to account for approved automation and end-to-end test behavior.

### DIFF
--- a/.github/actions/install-claude-cli/action.yml
+++ b/.github/actions/install-claude-cli/action.yml
@@ -1,0 +1,41 @@
+name: install-claude-cli
+description: >-
+  Install a pinned, signature-verified build of the Claude CLI onto PATH.
+
+# Isolated in its own file so the S6505/S8543 waivers this step needs
+# (sonar-project.properties) name only this action, not the workflow that calls
+# it. A waiver scoped to a whole workflow file also hides a genuine finding
+# anywhere else in that file; a waiver scoped to this one-step action cannot.
+#
+# Installed project-local into a scratch directory rather than `-g`: a global
+# install has no lockfile and `npm audit signatures` refuses to run against one
+# (EAUDITGLOBAL) — there would be nothing here to verify. A local install gets
+# its own throwaway package-lock.json and IS auditable, so this step fails
+# closed if the registry's signature over the exact tarball installed does not
+# verify, which is the integrity guarantee CodeRabbit asked for on a global
+# install that has none of its own.
+#
+# --ignore-scripts is not an option, unlike the pnpm installs elsewhere in CI:
+# the package's own postinstall (install.cjs) is what places the platform
+# binary bin/claude execs, so skipping it ships a CLI that cannot run. The
+# mitigation is the version pin (S8543 — an unpinned install tracks whatever
+# npm resolves at run time) plus the signature audit above.
+
+inputs:
+  version:
+    description: The exact @anthropic-ai/claude-code version to install.
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Install and verify the Claude CLI
+      shell: bash
+      working-directory: ${{ runner.temp }}
+      run: |
+        set -euo pipefail
+        mkdir -p install-claude-cli
+        cd install-claude-cli
+        npm install "@anthropic-ai/claude-code@${{ inputs.version }}"
+        npm audit signatures
+        echo "$PWD/node_modules/.bin" >> "$GITHUB_PATH"
--- a/.github/workflows/scheduled.yml
+++ b/.github/workflows/scheduled.yml
@@ -393,10 +393,18 @@ jobs:
       - name: Install the frontend
         if: steps.funded.outputs.configured == 'yes'
         working-directory: .
-        run: pnpm install --frozen-lockfile
+        # --ignore-scripts (githubactions:S6505): no lifecycle scripts run at
+        # install; esbuild ships prebuilt platform binaries as optional deps.
+        run: pnpm install --frozen-lockfile --ignore-scripts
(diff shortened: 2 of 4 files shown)


tinnhlt-gra:
Addressed all three CodeRabbit findings in dd99eb5:

- **Claude CLI integrity** (thread on scheduled.yml:407): moved the install into `.github/actions/install-claude-cli`, which now installs project-local (so it has its own lockfile) and runs `npm audit signatures` against it, failing the step if the registry signature over the installed tarball doesn't verify. A global install can't be audited this way at all — confirmed locally (`npm audit signatures -g` returns `EAUDITGLOBAL`).
- **Over-broad S6505 waiver** (thread on sonar-project.properties:342): the CLI install's own S6505 finding is now waived on its own file (`e15`), separate from the Playwright false-positive waiver on `scheduled.yml` (`e13`) — a workflow-wide waiver no longer masks anything else added to that file.
- **check.py path validation** (thread on sonar-project.properties:353): both `open()` calls now go through a shared `_open_checked` helper that refuses any path resolving outside the repo tree or the system temp directory. Verified: in-repo scenario and /tmp transcript paths still open; `/etc/passwd` is refused with a clear error.

tinnhlt-gra:
Good catch — fixed in 1fbeaf9. `_REPO_ROOT` only walked up two directories from `e2e/llm/check.py`, landing at `e2e/` instead of the actual repo root; every default path happens to be nested under `e2e/` so this was invisible until you flagged it. Now resolves correctly (verified against `git rev-parse --show-toplevel`, and against a scenario file placed outside `e2e/` but inside the repo, which previously would have been wrongly refused).

Issue #3122, opened 2026-08-29
The breadcrumb's current-page name collapses to nothing at 1280px, past the ellipsis it is designed to show


tinnhlt-gra:
At the default desktop width the top bar reads **"Contacts /"** and then stops — separator drawn, name gone. Not truncated with an ellipsis, not hoverable: zero pixels wide.

## Observed

Person record as `rep`, core `3b6a60c04`, Chromium, one record, one page load, viewport width stepped:

| Viewport width | `.crumbs-label` width | Visible? |
|---|---|---|
| **1280** | **0.0px** | **no** |
| 1366 | 40.5px | yes (ellipsis) |
| 1440 | 77.5px | yes |
| 1536 | 125.5px | yes |
| 1600 | 157.5px | yes |
| 1920 | 159.3px | yes (full name) |

The element is in the DOM with the right text throughout — `<span class="crumbs-label">UAT Dedupe Merge Person</span>` — it just has no box.

Screenshots are in the QC repo, `reports/evidence/defects/breadcrumb-topbar-1280.png` (reads `Contacts /` and nothing) and `breadcrumb-topbar-1920.png` (reads `Contacts / UAT Dedupe Merge Person`).

1280 is not an unusual width — it is the standard 13" laptop viewport, and Playwright's default.

## Why this is a defect rather than a layout preference

The intended behaviour is stated in the product's own source, and this is not it. `frontend/src/design-system/breadcrumb.css:82-85`:

> Only the last label clips, and it needs the whole chain to allow it: a flex item does not shrink below its content unless it is told it may. The tip `useTruncationTooltip` hangs on this element is what carries the rest of the name, on hover and on focus alike.

The design is: clip the current page, show an ellipsis, carry the full name in a tooltip. **At zero width there is no glyph to hover**, so the tooltip that makes truncation acceptable is unreachable. The fallback fails exactly where it is needed.

## Cause

Two individually reasonable rules:

- `breadcrumb.css:30` — ancestor crumbs are `flex: none`, so they never yield width. Deliberate, and the comment says why: shortening them "would hide the way back rather than the overflow."
- `breadcrumb.css:86-88` — the current page's `.crumbs-label` is `min-width: 0` with `overflow: hidden; text-overflow: ellipsis`, and **no positive floor**.

Flexbox honours `min-width: 0` literally, so once the unshrinkable ancestors, separators and the top bar's other tracks have taken the row, the single shrinkable item is driven to 0px — past the width an ellipsis needs to paint into.

[...]

PR #3123, opened 2026-08-29
fix(breadcrumb): give the current-page label a floor before it clips

tinnhlt-gra:
## Summary
- Improves, but does NOT close, margince#3122 (stays open — see below). Two real, browser-verified fixes to the breadcrumb primitive itself:
  1. `min-width: 2ch` on the current-page label — was `min-width: 0` with no floor, which let it collapse to a literal 0px box (invisible, not even an ellipsis) wherever the ancestor crumbs already filled the row.
  2. `overflow: hidden` on `.crumbs` — without it, the 2ch-floored box painted uncropped past the trail's own edge instead of clipping there (caught by cubic's review of the first commit).
- **Both are correct and both are confirmed working from 1366px up** (cross-session QC pass, real browser, three rounds of measurement including catching and correcting an instrument artifact in Playwright's `hover()` auto-scroll).
- **Neither fixes 1280px, and no floor value in `breadcrumb.css` can.** At that width `.crumbs`'s own grid cell (`.topbar-lead`, one of `topbar.css`'s three `minmax(0, 1fr)`/`auto`/`minmax(0, 1fr)` tracks) is measured at 72px — "Contacts /" alone already fills it. Any min-width on the label just makes content wider than a 72px box: overlapped without clipping, or clipped away with it. There is no room to give within the breadcrumb component; the trail's cell itself is too narrow, and that is a `topbar.css` grid-allocation question (should the search track shrink before the trail does? should the avatar-only right cell yield its unused share?) — a bigger, more visually-iterative call than this PR, left for margince#3122 to track.

## Test plan
- [x] `pnpm vitest run src/design-system/breadcrumb.test.tsx` — 8/8 pass, unchanged (jsdom does no real flex layout, so this class of bug is invisible to it — consistent with why it shipped in the first place)
- [x] `pnpm exec biome check` on the changed file — clean
- [x] Real-browser verification across 1280/1366/1440/1600/1920 (cross-session QC pass) — 1366+ confirmed fixed (ellipsis renders, tooltip reachable by real hover and focus); 1280 confirmed NOT fixed (`crumbs.clientWidth: 72, scrollWidth: 92` — the label is clipped away entirely, no hover target exists for a pointer)

---
## Summary by cubic
Fixes the breadcrumb current-page label collapsing to 0px at 1280px (margince#3122) so it now shows an ellipsis and the hover/focus tooltip instead of nothing.

- Sets the label's `min-width` to `2ch`, enough room for the ellipsis glyph alone without promising how much of the label survives.
- Adds `overflow: hidden` on `.crumbs` so the floored label clips at the trail's edge instead of bleeding under the top bar's search box at 1280px.
- The tooltip popup is unaffected because it renders via `createPortal`.
- No unit test added: jsdom does no real flex layout (the 8 existing tests pass unchanged); re-verify visually across 1280/1366/1440/1600/1920 before merge.

<sup>Written for commit b4f9463807fd1acbf638b02cfff20a8eb213878e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/3123?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved breadcrumb truncation so the final label consistently displays an ellipsis when space is limited.
  * Prevented overflowing breadcrumb content from extending beyond its container.

### DIFF
--- a/frontend/src/design-system/breadcrumb.css
+++ b/frontend/src/design-system/breadcrumb.css
@@ -7,6 +7,16 @@
    unchanged on screen. */
 .crumbs {
   min-width: 0;
+  /* The trail owns its own overflow rather than trusting a caller's wrapper
+     to. Without this, the last label's 2ch floor (below) is still painted in
+     full even when the row has no room for it — the ancestors' `flex: none`
+     never yields, so the floored box is pushed past the trail's own right
+     edge and bleeds, uncropped, into whatever sits beside it (measured: the
+     top bar's centred search box, at 1280px — the label was visible but
+     unreachable, sitting under a control that intercepted every hover and
+     click meant for it). Clipping here keeps the bleed inside the primitive
+     that owns the promise, instead of asking every consumer to remember it. */
+  overflow: hidden;
 }
 .crumbs > ol {
   display: flex;
--- a/frontend/src/design-system/breadcrumb.css
+++ b/frontend/src/design-system/breadcrumb.css
@@ -82,9 +82,17 @@
 /* Only the last label clips, and it needs the whole chain to allow it: a flex
    item does not shrink below its content unless it is told it may. The tip
    `useTruncationTooltip` hangs on this element is what carries the rest of the
-   name, on hover and on focus alike. */
+   name, on hover and on focus alike.
+
+   The floor is 2ch, not 0: `min-width: 0` alone has no lower bound, and at a
+   width where the unshrinkable ancestor crumbs already fill the row (measured:
+   1280px in the shell's own top bar), flexbox drove this label to literally
+   0px — past the point `text-overflow: ellipsis` has anything to paint the
+   glyph into, so the current page read as nothing at all rather than as
+   truncated (margince#3122). 2ch is enough room for the ellipsis on its own;
+   it does not promise any of the actual label survives. */
 .crumbs > ol > li:last-child .crumbs-label {
-  min-width: 0;
+  min-width: 2ch;
   overflow: hidden;
   text-overflow: ellipsis;
 }


tinnhlt-gra:
Re-ran the width table against this branch (`622fc49e6`) in a real browser, as asked. **The reported symptom is fixed; the design intent is not yet met at 1280.**

## The measurements

Same record, same page load, viewport stepped. Before → after:

| viewport | before | after | visible after |
|---|---|---|---|
| **1280** | **0.0px** | **17.8px** | yes |
| 1366 | 40.5px | 40.5px | yes |
| 1440 | 77.5px | 77.5px | yes |
| 1600 | 157.5px | 157.5px | yes |
| 1920 | 159.3px | 159.3px | yes |

So `min-width: 2ch` resolves to 17.784px and does exactly what it says: the label is no longer zero-width, and no other width regressed. `clientWidth: 18`, `scrollWidth: 159`, `text-overflow: ellipsis` — it is genuinely clipping rather than collapsing now.

## But the tooltip is still unreachable at 1280

The point of the floor, per the comment right above the rule, is that `useTruncationTooltip` "carries the rest of the name, on hover and on focus alike". It still cannot:

```
locator.hover: Timeout 8000ms exceeded
```

Hit-testing the centre of the label's own rect explains why:

```
width=1280  rect={x:386.5, w:17.8}  elementFromPoint -> BUTTON.topbar-search   (not the label)
width=1366  rect={x:386.5, w:40.5}  elementFromPoint -> SPAN.crumbs-label      (the label)
```

At 1280 the search button **overlaps** the crumb's box. The label has a rect, but nothing can point at it, so hover and focus never fire and the full name stays unreachable — which is the same end state as before, reached a different way.

On screen it reads `Contacts / U` — one character, with no ellipsis glyph visible in the 17.8px that survives. A reader cannot tell the name is truncated rather than just short.

## What I think this means

Not a reason to hold the fix, if you want it in as a strict improvement — 0px and invisible is worse than 17.8px and clipped, and nothing regressed. But margince#3122 should stay open, or a follow-up should be filed, because the acceptance criterion in the CSS's own comment is not satisfied at the width the issue was filed about.

The remaining cause is the one I noted on the issue: `topbar.css`'s `minmax(0,1fr) auto minmax(0,1fr)` holds the trail's cell symmetric with a right-hand cell containing only an avatar, and at 1280 the search track is allowed to sit over the trail rather than the row reflowing. A floor on the label cannot fix an overlap.

Happy to re-measure any follow-up — the repro is a few lines and I have it set up.

tinnhlt-gra:
cubic's read was exactly right, and sharper than mine — pushed b4f946380: `overflow: hidden` on `.crumbs` itself so the 2ch-floored label clips at the trail's own edge instead of bleeding into whatever sits beside it (confirmed safe: the tooltip renders via `createPortal`, unaffected by ancestor overflow). This should also close the overlap a cross-session QC pass found in a real browser at 1280px (search button intercepting hover/click). Re-requesting that same verification before merge.

tinnhlt-gra:
**Correction to my previous comment, and a hold on "Closes #3122".** Re-measured `b4f946380`. It does not fix the user-visible symptom at 1280, and my earlier "hover works, tooltip shows the full name" was an artifact of my own instrument.

## On arrival at 1280, the bar still reads `Contacts /` and nothing

No hover, no focus, no interaction:

```
crumbs.clientWidth = 72    crumbs.scrollWidth = 92    crumbs.scrollLeft = 0
```

The label is ~18px and sits past the 72px clip, so `overflow: hidden` now removes it from view entirely. Same visible outcome as the original defect, by a third mechanism: zero-width → overlapped → clipped away.

## My passing hover was Playwright, not a person

`hover()` calls `scrollIntoViewIfNeeded` first. That scrolls the newly-clipped `.crumbs` box horizontally, brings the label into view, and *then* the tooltip fires. It also explains the oddity in that screenshot I nearly reported as a regression — the trail read `tacts / U…` because the container had scrolled right and cut "Con" off the ancestor. That was my instrument, not your change.

A person with a mouse cannot do it: nothing is painted, so there is nothing to move the pointer onto, so no hover and no tooltip. `elementFromPoint` was the honest signal and I should have trusted it over the passing hover:

```
width=1280  crumbWidth=17.8  hit=BUTTON.topbar-search  hitsLabel=false   <- still false on b4f946380
width=1366  crumbWidth=40.5  hit=SPAN.crumbs-label     hitsLabel=true
```

1366 and above were already correct before either commit and remain correct.

## Why a floor cannot work at this width, whatever value it takes

`.crumbs` is only **72px wide** at 1280 — that is the trail's whole cell, and "Contacts" plus the separator already fills it. Any `min-width` on the last label just makes the content wider than a 72px box; it is then either overlapped (no clip) or clipped (with clip). The label cannot be shown because the trail's cell has no room to give, which is upstream of `breadcrumb.css`: `topbar.css`'s `minmax(0,1fr) auto minmax(0,1fr)` hands a symmetric share to a right-hand cell holding one avatar, while the search track takes `min(420px, 34vw)` from the middle.

If 1280 should work, the lever is that grid — let the trail claim the slack the avatar cell isn't using, or let the search track shrink below 420px before the trail does. That is a `topbar.css` change and a bigger call than the issue I filed. Closing #3122 as "improved; ellipsis correct from 1366 up" is a perfectly reasonable outcome — provided the issue says that, rather than implying 1280 is fixed.

To be clear on the parts that are right: `overflow: hidden` on `.crumbs` is correct regardless, and the diagnosis of the bleed was correct — the box really was painting past the trail's edge. It is just not sufficient here.

PR #3151, opened 2026-08-29
One entry for one FK — main does not compile with two

dungbanguyen:
`main` has not compiled since #3145 and #3144 both landed.

The two fixed the same red gate in parallel — `TestFK_rowScopedTargetsHaveVisibilityDecision` on `relationship.counterparty_person_id`, which #3140 introduced — and each added its own entry to `rowScopedFKDecisions`. A duplicate key in a map literal is a compile error, so the `migrations` test package has not built since: every branch fails the lint step with a typecheck error that names neither PR.

One entry now, keeping what each of them said that the other did not — which kind the far end belongs to (`worksWithKind`), and the function that walks it (`ensureRelationshipEndpoints`).

`make check-backend` exit 0.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected relationship mapping for person counterparties by consolidating duplicate and conflicting entries.
  * Preserved the existing organization counterparty behavior.

### DIFF
--- a/backend/migrations/schema_fitness_integration_test.go
+++ b/backend/migrations/schema_fitness_integration_test.go
@@ -251,9 +251,8 @@ var rowScopedFKDecisions = gatekit.Waive(map[string]string{
 	// the finder's ScopeClauseFor), never a request body.
 	// Client-supplied edge endpoints — every one probed at the store:
 	"relationship.person_id":                     "gated: auth.EnsureLinkTarget in CreateRelationship (H1)",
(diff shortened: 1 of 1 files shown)


LarsGradion:
Superseded by #3150, which merged first and removed one of the two duplicate census entries — main compiles and main-health run 33253652353 is green. This diff was cut against the pre-#3150 tree (it removes both old lines), so landing it now would conflict; if the merged entry's wording should carry the ensureRelationshipEndpoints sentence from this PR, that's a one-line follow-up. Reopen if I've misread the intent.

Issue #3179, opened 2026-08-29
notices.Store.Create has no idempotency key, so an at-least-once redelivery duplicates the notice


tinnhlt-gra:
## What's wrong

`workflow.Workflow.Apply`'s own interface doc says a handler is \"Idempotent on IdempotencyKey(ev)\" (`backend/internal/shared/ports/workflow/workflow.go:39`) — the same event redelivered must produce the same net effect, not a second one.

`leadSLAEscalation.Apply` (backend/internal/compose/leadsla.go) honors this for the task half: `activities.Store.LogActivity` resolves an existing row by `(source_system, source_id)` before inserting (`activities/activity.go:205` `replayedActivity`), so two deliveries of one breach log exactly one task — proven by `TestLeadSLAEscalationLogsOneTaskOnTheLead`, which calls `Apply` twice and asserts `tasks == 1`.

The notify half added in #3173 does not: `notices.Store.Create` (`backend/internal/modules/notices/store.go:56`) always inserts a fresh row with a new `ids.NewV7()` id — no dedup key, no `ON CONFLICT`, nothing checked before the insert. Two deliveries of the same `lead.sla_breached` event write two rows in the `notice` table, and the recipient sees the same "A lead's first response is overdue..." line twice on their Worklist.

## Where this surfaced

Found while fixing a nil-pointer panic in `TestLeadSLAEscalationLogsOneTaskOnTheLead` (the test wasn't updated to inject the new `notices` field #3173 added to `leadSLAEscalation`, separately fixed). Deliberately did not add a "notice count == 1 after two deliveries" assertion to that test while fixing the crash — writing one honestly would fail today, and fixing it needs a design decision, not a 3-line patch alongside an unrelated crash fix.

## Why this needs a decision rather than a quick patch

`activities`' dedup is a real (source_system, source_id) lookup-before-insert, which the `notice` table has no equivalent of (columns: id, recipient_user_id, kind, subject, body, captured_by, created_at, read_at — nothing shaped like a dedup key). Adding one is a schema change and an API decision that affects every future notice producer, not just this one caller:

- Should every `Create` call require a caller-supplied dedup key (mirroring source_system/source_id), or should the store derive one from (recipient, kind, subject) — and is that derivation actually collision-safe for two DIFFERENT real notices that happen to share text?
[...]

Issue #3174, opened 2026-08-29
A notify firing is at-most-once and its recipient's liveness is unchecked


LarsGradion:
Three connected gaps around the new notice transport (#3173), all rooted in pre-existing engine semantics rather than the transport itself:

1. The automation engine claims a run before Apply, so a transient failure creating the notice records a terminal 'failed' run and redelivery never retries — the notification is permanently lost. (The failed run is at least visible: an automation's failure reaches the Worklist's automation_health lane; a system workflow's does not.) This is the same claim-before-effect shape as #3038.
2. The lead-SLA escalation writes its task and its notice in two separate transactions, so a failure between them lands the task without the notice, and per (1) never retries.
3. Neither producer checks the recipient's seat is active: a deactivated owner's notice is recorded as delivered but will not be read until reactivation — the same shape task assignment already has.

Direction: make the notify effect retryable (effect claim inside the apply transaction, per #3038's direction), fold the lead-SLA pair into one transaction or make the notice a claimed effect of its own, and decide whether producers should skip (with a reasoned run) or still record for inactive recipients.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

navdeepplaya-alt:
Picking this up, bundled with #692/#1820 — the notification-system build reshapes the delivery pipeline these three gaps live in, so retryability of the notify effect, the lead-SLA two-transaction split, and the inactive-recipient rule get settled in that same design pass.

navdeepplaya-alt:
-

PR #3180, opened 2026-08-29
fix(compose): give TestLeadSLAEscalationLogsOneTaskOnTheLead a notices.Store

tinnhlt-gra:
## Summary
`main-health`'s integration shards were panicking with a nil-pointer dereference:

```
panic: runtime error: invalid memory address or nil pointer dereference
notices.(*Store).Create(0x0, ...) at notices/store.go:67
compose.leadSLAEscalation.Apply(...) at leadsla.go:101
compose.TestLeadSLAEscalationLogsOneTaskOnTheLead(...) at leadsla_integration_test.go:62
```

`leadSLAEscalation` gained a `notices *notices.Store` field in #3173 (the notify half of RC-5's `notify_and_task` default, now that a durable transport exists). Production wiring (`workflows.go:90`) sets it correctly, but this test's own construction of `leadSLAEscalation{}` wasn't updated — a nil pointer, and the test's payload always names an `EscalationTarget`, so `Apply`'s new `notices.Create` call panicked every run.

Fixed by constructing the store the same way `workflows.go` does.

**Not fixed here, filed as #3179 instead:** `notices.Store.Create` has no idempotency key (unlike `activities.LogActivity`'s own `(source_system, source_id)` dedup), so this test's existing two-delivery loop would create two duplicate notice rows — an honest "notice count == 1" assertion would fail today. Fixing that needs a schema change and an API decision for the notices module (what the dedup key is, whether every producer must supply one), not a patch bundled with this crash fix.

## Test plan
- [x] `go build -tags=integration ./...` — clean
- [x] `go vet -tags=integration ./internal/compose/...` — clean
- [x] `craft static` — 0 blocker, 0 major, 0 minor
- [ ] Not run locally: `make test-it` hits the same pre-existing stale `extensions/notes` directory noted in #3149's fix, on a shared machine I'm not touching. Relying on CI for the real integration run.

---
## Summary by cubic
Fixes the nil-pointer panic in `TestLeadSLAEscalationLogsOneTaskOnTheLead` by giving `leadSLAEscalation` the `notices.Store` it now requires, matching production wiring. The test doesn't yet assert on notice rows because `notices.Store.Create` lacks an idempotency key, which would create duplicate notices in the existing two-delivery loop; that's tracked in #3179.

<sup>Written for commit e6951a15ed249c8bff3b4aa63866c88bc84a2813. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/3180?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Tests**
  * Updated lead SLA escalation integration coverage to include notices alongside activities.
  * Improved test setup for scenarios involving escalation notices.

### DIFF
--- a/backend/internal/compose/leadsla_integration_test.go
+++ b/backend/internal/compose/leadsla_integration_test.go
@@ -21,6 +21,7 @@ import (
 	"github.com/margince/margince/backend/internal/compose/integration"
 	crmcontracts "github.com/margince/margince/backend/internal/contracts"
 	"github.com/margince/margince/backend/internal/modules/activities"
+	"github.com/margince/margince/backend/internal/modules/notices"
 	"github.com/margince/margince/backend/internal/shared/kernel/ids"
 	"github.com/margince/margince/backend/internal/shared/kernel/principal"
 	"github.com/margince/margince/backend/internal/shared/ports/datasource"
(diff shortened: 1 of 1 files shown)


tinnhlt-gra:
Superseded by #3178, merged independently for the same bug around the same time — same root cause (leadSLAEscalation's test wasn't updated when #3174 added the notices field), and their fix is better: a shared newLeadSLAEscalation(db, now) constructor both production and the test call, which prevents this exact class of bug (partial struct literal reading as complete) from recurring rather than just patching this one instance. Closing mine in favor of theirs.

Issue #3041, opened 2026-08-28
A confirmed close date can never be undone, and blames 'superseded'


LarsGradion:
## What is wrong

Undo refuses every `close_date_correction`, and gives a reason that is not true: `superseded (expected_close_date)`. Nothing superseded it. The refusal fires on the very first attempt, immediately after the change is applied.

## Why

`deal.expected_close_date` is a `date` column (`backend/migrations/core/0001_baseline.up.sql:1235`), but the audit image records a full timestamp.

The effect at `backend/internal/compose/closedate.go:175` parses the proposal with `time.Parse(time.DateOnly, ...)` and passes the resulting `time.Time` to `UpdateDeal`, so the audit `after` image holds `"2026-12-01T00:00:00Z"`.

`fieldsThatMovedSince` (`backend/internal/compose/superseded.go:117-123`) then compares that image against `to_jsonb(r) -> 'expected_close_date'`, which for a `date` column is `"2026-12-01"`. The two are distinct as JSON, so the field reads as moved and the evaluator refuses with `ReasonSuperseded`.

Observed directly: the audit row is `after={"expected_close_date": "2026-12-01T00:00:00Z"}` while the deal holds `2026-12-01`.

## Scope

This is not specific to automatic application — a close-date correction a human approves by hand has the same un-undoable audit row. Any `date`-typed column written from a `time.Time` is exposed to the same mismatch, so the fix is worth making at the write/image boundary rather than at this one call site.

## Why it was deferred

Found while building auto-apply, but the defect is in the close-date effect and the supersession comparison — neither of which that change touches. Fixing it properly means deciding where a date-typed value is normalized (the store's update path, or the image the write shape records), which reaches beyond the diff under review.

## What it means if nobody fixes it

Undo silently does not work for close dates, and tells the user something false about why. That matters more now: auto-apply is allowed to change records without asking precisely because the change can be put back, so a kind whose Undo always refuses undermines the bargain the feature rests on.

PR #3190, opened 2026-08-29
fix(deals): a refused close date is not proposed again tomorrow

LarsGradion:
## The bug

The nightly close-date sweep staged its correction with plain `Stage`. Its only guard was "is one already pending" — and a rejection clears `pending`. So the next night staged the same proposal again, and the night after that. A rep could only end it by giving in.

## The identity

Staging now goes through `StageUnlessDeclined` keyed on the deal and **the date it currently holds**.

Not the proposed date: that is recomputed against "today" on every pass, so an identity carrying it makes each night a different question and the memory recognises nothing.

Not the deal alone either, for the opposite reason. A refusal never expires, so a deal-only key would let one "no" end close-date hygiene on that deal permanently — the rep declines, sets their own date, it slips in turn, and nobody ever tells them.

## The keep-alive branch needed its own answer

When a deal is provisional and unflagged, the sweep re-offers the date **it wrote itself**, to keep the confirm card alive if the last staging expired undecided. Expired, not refused — and it could not tell the difference. After a rejection each night's card names a different date, so the staging's own memory reads it as a fresh question every time.

That branch now asks whether a correction on this deal was refused at all, and holds off if one was. The finer per-date answer stays with the branch that actually proposes a change.

## Payload

`CloseDateCorrection` gains `standing_close_date` beside the nullable `previous_close_date`. An identity field must be a string the payload carries with the same value, and "no date at all" is a real case — the `missing` flag exists for it. The card still renders the nullable one; the new field is not in the display allowlist.

## Verification

Four integration tests against a real database:

- a refused date is not proposed again the next night
- a refusal on one deal does not silence another
- a refusal survives the proposal moving under it
- **a date the rep set themselves is still corrected when it goes stale** — the test for the failure that hides

Mutation-checked in both directions. Keying on the proposed date fails all four. Keying on the deal alone passes three and is caught only by the fourth, which is exactly the point of writing it.

Plus a unit test that marshals the real struct and fails if either JSON tag moves away from the identity constants. Without it the nightly sweep discovers the mismatch, in a worker log, after shipping.

`closedatesweep.go` crossed the 500-line cap, so the proposal shape and its seams move to `closedateproposal.go`.

All local gates green: `check-backend`, `check-fe`, `craft-static`, `rls-store-path`, `no-jurisdiction`, and the nine existing close-date integration tests alongside the four new ones.

---
## Summary by cubic
Fixes the nightly close-date sweep so a date a rep declined is not proposed again the next night. Refusals are remembered against the deal and the date it currently holds, so a rep who sets their own date is still asked again when that one goes stale.

- Staging goes through `StageUnlessDeclined` keyed on the deal and its standing close date.
- A per-pass refusal check compares the proposed date on every arm, since the sweep writes its new date onto the deal before staging and would otherwise outrun the identity; the old deal-wide hold-off in the keep-alive branch is gone.
- `CloseDateCorrection` gains `standing_close_date` for the identity; the card still renders the nullable `previous_close_date`.
- Moved the proposal shape and stager seams into `closedateproposal.go`.
- Added integration tests for refusals persisting across nights and deals, a refusal surviving a moved proposal, and a rep-set date going stale.

<sup>Written for commit 5b3ddf8c3b3a5709b1c38910739308911fb67fa5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/3190?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Close-date correction proposals now include the deal’s current standing close date for reliable tracking.
  * Previously refused corrections are remembered and won’t be proposed again unless the deal’s standing date changes.
  * Correction payloads now require valid deal and date information.

* **Bug Fixes**
  * Prevented rejected corrections from resurfacing during later review cycles.
  * Improved separation of correction history between deals and changing proposals.

### DIFF
--- a/backend/internal/compose/closedate.go
+++ b/backend/internal/compose/closedate.go
@@ -57,20 +57,39 @@ func (s closeDateStager) HasPendingCorrection(ctx context.Context, dealID ids.UU
 	return s.svc.HasPendingKind(ctx, deals.CloseDateCorrectionKind, dealID)
 }
 
-// HasRefusedCorrection reads the memory's own record: any rejected correction
-// against this deal.
+// RefusedCloseDate answers the seam's question from the payloads the memory
+// stores.
 //
-// It asks whether ONE was refused rather than which, because its caller is the
-// keep-alive branch and that branch proposes no change to weigh — it re-offers
-// the date the sweep already wrote, which is the date the rep declined to have
-// set. The staging path keeps the finer answer for the branch that does propose
-// a change, where a new standing date is a new question.
-func (s closeDateStager) HasRefusedCorrection(ctx context.Context, dealID ids.UUID) (bool, error) {
+// RejectedChangesFor hands back payloads rather than answering a containment
+// query because only the caller knows what makes two of its proposals the same
+// question. This walks them and lets the probe decide — the judgment is
+// RefusalProbe.SameQuestionAs, in the module that owns close dates, because it
+// is a fact about corrections rather than about staging.
+//
+// The read commits before the caller stages, so it can lose a race to a decision
+// landing in the gap. That costs one extra offer and never an unasked write —
+// what the caller goes on to do is stage, and staging re-checks under its own
+// lock.
+func (s closeDateStager) RefusedCloseDate(ctx context.Context, dealID ids.UUID, proposed deals.RefusalProbe) (bool, error) {
 	refused, err := s.svc.RejectedChangesFor(ctx, deals.CloseDateCorrectionKind, dealID)
 	if err != nil {
 		return false, err
 	}
-	return len(refused) > 0, nil
+	for _, payload := range refused {
+		earlier, err := deals.UnmarshalCloseDateCorrection(payload)
+		if err != nil {
+			// A payload an older version of this stager wrote may not decode
+			// today, and a decision a human made does not expire because the
+			// shape moved. Reading it as "not this one" costs one card they
+			// have seen before; failing the sweep would cost every deal's
+			// date hygiene tonight.
+			continue
+		}
+		if proposed.SameQuestionAs(earlier) {
+			return true, nil
+		}
+	}
+	return false, nil
 }
 
 func (s closeDateStager) StageCorrection(ctx context.Context, dealID ids.UUID, targetVersion int64, summary string, proposal deals.CloseDateCorrection) error {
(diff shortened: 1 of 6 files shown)


LarsGradion:
## Do not merge — this does not work, and its tests pass for the wrong reason

A second review pass found the fix carries the same defect it was written to remove.

**The memory only holds for one calendar day.** `proposedCloseDate` is `today.AddDate(0, 0, stages*velocityDays)` ([closedate.go:197](backend/internal/modules/deals/closedate.go#L197)). A deal whose situation has not changed is offered `today+14` tonight and `today+15` tomorrow. `SameQuestionAs` compares proposed dates, so a refusal matches on the day it was made and never again. The rep is asked again the next morning — which is the bug.

**The tests cannot see it.** `NewCloseDateCorrector` pins `now: time.Now` ([closedatesweep.go:62](backend/internal/modules/deals/closedatesweep.go#L62)) and exposes no seam to override it, so both sweeps in every test run within milliseconds and compute the same "today". `TestARejectedCloseDateIsNotProposedAgainTomorrow` never reaches a tomorrow. I verified the design against a live database, but both probe passes ran on one wall-clock day, which is exactly why the observed pair looked stable.

Two further findings, both real:

**`AfterHumanEdit` is not "a person chose this date".** It is derived from `close_date_provisional`, which has two writers — a human edit ([deal.go:145](backend/internal/modules/deals/deal.go#L145)) and the sweep's own auto-apply arm ([closedatesweep.go:278](backend/internal/modules/deals/closedatesweep.go#L278)) running as the system principal.

**The flagged and keep-alive arms cannot recognise each other's refusals.** The flagged arm proposes `today+N`; the keep-alive arm proposes the date the deal holds. A rep who refuses one gets the other. No test covers the crossing case; all four seed deals that stay flagged.

## What the fix needs

The key must be something that does not move with the calendar. The candidate is the standing date the *rejected offer itself recorded* — the rep's own last chosen date, which changes only when they change it — compared against what the deal held before this pass's write.

It also needs a clock seam on the corrector so a test can advance a day, because without one no test in this tree can prove the property this PR is named for.

The bug this PR describes is real and still unfixed on main: a rejected close-date correction is re-staged by the nightly sweep. Reopening as draft rather than merging a fix that does not work.

LarsGradion:
Superseded by #3196, which fixes this properly. The key here moved with the calendar and the tests never crossed a day, so they passed for the wrong reason — recorded in #3192, and #3196 closes it along with the close-date Undo defect (#3041) that made the same records irreversible from the other end.

PR #3214, opened 2026-08-30
fix(records): the duplicate queue serves decisions that still have subjects

### LINKED ISSUE #3185: The duplicate queue keeps serving decisions about archived records, crowding live work out of the top 10
Filed by QC, measured on a dev stack at `d5c0eeb61`.

## What a person sees

Ops opens the "Needs you" duplicate lane. It serves the ten open decisions with the highest confidence. On our workspace, **eight of those ten are pairs where BOTH organizations have been archived** — they are asking somebody to decide whether to merge two records that no longer exist.

Measured, not estimated:

```
open organization candidates:        25
  BOTH sides archived (dead work):   21
  at least one side live:             4
  of the TOP 10 the lane serves:      8 dead
```

Because the lane orders by confidence and archived pairs keep whatever score they were filed with (0.917 here), the dead ones outrank live work indefinitely. Four real decisions sit behind eight dead ones.

## Cause

`dedupe_candidate` carries its own `archived_at`, and a merge sweeps it (`internal/modules/people/leadmerge.go:183`). Nothing sweeps it when the **subject** is archived, and the serving query does not check:

`internal/modules/people/dedupequeue.go:362-368`

```sql
SELECT ...
  FROM dedupe_candidate
 WHERE disposition = $1 AND archived_at IS NULL
   AND (left_org_id = $2 OR right_org_id = $2)
```

`archived_at IS NULL` is the candidate's own flag. There is no join against `organization.archived_at` (or the person/lead equivalents), so a decision outlives both the records it is about.

Verified end to end on one pair:

```
candidate 01a04e47-33aa-...  status: open   confidence: 0.917
  left  01a04e47-3383-...  "UAT Relationship Company"       archived_at 2026-08-30T00:37:22
  right 01a04e47-33a9-...  "UAT Relationship Co-sell Co"    archived_at 2026-08-30T00:37:23
```

## Why this matters outside a test workspace

Archiving one of two duplicates is a reasonable way for a person to resolve a duplicate — arguably the most natural one, since it needs no merge decision. Doing that today leaves the decision **open forever**, permanently occupying a slot in a lane that only shows ten. A workspace that resolves duplicates by archiving will slowly fill its own queue with its own completed work, and the more diligent the user, the faster it fills.

Nothing fails while this happens. The endpoint answers 200, the lane renders ten rows, every test passes. It is only visible if you ask whether the rows still point at anything.

## Suggested direction

Not prescribing the fix, but the two obvious shapes:

1. **Filter at read** — join the subject and require it live. Cheapest, and self-correcting for candidates already stranded.
2. **Sweep at archive** — set `dedupe_candidate.archived_at` when either subject is archived, the way `leadmerge.go` does on merge. Tidier, but leaves existing rows stranded and needs a backfill.

(1) alone fixes what a person sees today. (2) alone does not, without the backfill.

## How QC found it

Not by a failing assertion — by triaging one. Our `dedupe-merge-lifecycle` setup refuses to run when its own fixture cannot reach the served page, and says so:

> this fixture's dedupe candidate is not on the served "Needs you" page. The lane serves 10 of the workspace's open decisions, ordered by confidence, and a bare name collision scores low — so the walk cannot reach it.

The easy reading was "our test stack is dirty, clear it and move on". Asking *why* it was full found 21 of 25 entries pointing at archived records. Worth saying because the same symptom in a customer workspace would read as "the dedupe queue is noisy" rather than as a bug.

---
Found by [margince-qc](https://github.com/margince/margince-qc) while triaging a red in `UAT-MERGE-1` (`dedupe-merge-lifecycle`). Its setup refuses when the fixture pair cannot reach the served page, and the easy reading was "our test stack is dirty". Asking why it was dirty was the finding.

LarsGradion:
Archiving one of two duplicates is a reasonable way to resolve a pair. For a company entered twice it is the most natural one — it needs no merge decision about which fields survive. It left the decision open forever.

## What was happening

`dedupe_candidate` carries its own `archived_at`, and one path sweeps it: a lead merge retires the pairs naming the loser. Nothing sweeps it when a **subject** is archived, and the serving query never asked — `archived_at IS NULL` is the candidate's own flag, with no join to the record it names.

So a decision outlived both records it was about. And because the lane orders by confidence and an archived pair keeps the score it was filed with, dead work held its rank indefinitely. On the workspace where this was measured: 21 of 25 open organization candidates named two archived companies, and 8 of the 10 the lane serves were dead. Four real decisions sat behind them.

The more diligently a workspace resolves duplicates by archiving, the faster its queue fills with its own finished work.

Nothing failed while that happened. The endpoint answers 200, the lane renders ten rows, every test passes. It is visible only if you ask whether the rows still point at anything.

## Fixed at the read

Seventeen code paths archive a person, organization or lead — plain archives, merges, lead disqualification and promotion, erasure, two retention sweeps, import undo. Exactly one of them retires the naming candidates. Putting the rule at each of the other sixteen means the next archiver added has to remember it.

So both sides of a pair must now EXIST and be LIVE, in one predicate the queue, the badge and the by-record lookup all share.

**The liveness term cannot ride the row-scope clause**, and `auth.EnsureVisibleLive` already says why in its own words: erasure anonymizes a person in place and stamps `archived_at` while *leaving `owner_id` alone*, so a scope predicate answers "yes, still yours" for a record every live read path refuses. It is a separate term.

**It is also emitted unconditionally now.** That was not the bug — `person` and `organization` are capture-private, so even an all-scope human is bounded on them and the clause was built anyway. But the old shape skipped the whole predicate when *no* record type narrowed the caller, and a system principal is unbounded on all three. I measured the clauses rather than assuming: person and organization come back non-empty for an all-scope rep, `lead` comes back empty.

## One more reader, found on the way

`DuplicateCluster` in `enrichment.go` — the fence that stops two records of the same human each buying the same enrichment — filtered neither the candidate's own flag nor its twin's. It went on clustering a person the product had erased. It now requires both live.

`orgPairAlreadyFiled` in `renamerecheck.go` deliberately keeps spanning every disposition: it is a uniqueness-suppression probe, and `uq_dedupe_candidate_pair` is not partial, so narrowing it would let a second copy of an already-filed pair be inserted and collide.

## Verification

- `make check-backend` green, `make craft-static` clean.
- The whole `internal/modules/people` package and the `internal/compose` integration lane green, plus `integrations` (which `DuplicateCluster` feeds).
- Five integration tests, all failing against the old code. Four cover the three record types and the badge; the fifth is the control — a pair of live companies, with an *unrelated* company archived — because a predicate that closed everything would pass the other four while emptying the queue of exactly the work it exists to show.

Reported by QC. Closes #3185.

---
## Summary by cubic
Fixes the duplicate queue serving decisions about archived records, so closed pairs stop holding rank in the lane.

- Open pairs now require both subjects live; merged pairs stay listed on `status=merged`.
- By-id reads and dispose actions now refuse open pairs naming an archived record, so a kept link can't open the evidence snapshot.
- The liveness predicate is emitted unconditionally, so a system principal can't read archived pairs.
- `DuplicateCluster` now requires both candidate and twin live; archiving one side of a pair closes the decision without write-path changes.
- Added integration tests covering archived organizations, persons, disqualified leads, a live-pair control, and merged-pair visibility.

<sup>Written for commit b4946522ba2771b738d50554c0764ca23b778fc6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/3214?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Excluded archived or disqualified records from open deduplication queues and duplicate-cluster results.
  * Ensured candidates appear only when both paired records are active.
  * Updated candidate counts and visibility when records are archived or disqualified.
  * Preserved valid open and merged decisions involving active records.
  * Prevented archived duplicate pairs from being read or acted upon.
  * Applied write-authority checks consistently when viewing deduplication candidates.


LarsGradion:
**Query plan, measured rather than assumed.** Two correlated EXISTS per candidate row is the kind of thing that reads as a plan risk, so I ran `EXPLAIN` over the real predicate the queue builds:

```
Limit
  -> Incremental Sort  (Sort Key: confidence DESC, id DESC)
    -> Index Scan using idx_dedupe_candidate_open on dedupe_candidate
```

The partial index — `(confidence DESC) WHERE disposition = 'open' AND archived_at IS NULL` — still drives the ordered scan, and the sort stays incremental, so the `LIMIT` still stops early rather than materialising the queue.

Every subject probe underneath is an index scan, never a sequential one:

```
SubPlan -> Index Scan using idx_org_updated_keyset on organization
SubPlan -> Index Scan using idx_lead_ws_live on lead
```

`idx_lead_ws_live` is itself already live-filtered, so the lead side costs nothing extra. No new index is needed.

LarsGradion:
**Review round.** Three findings, two of them real. Both have a test that failed before the fix.

**A regression I introduced.** Applying "both sides live" to *every* status hid every successfully merged pair from `status=merged` — a filter the contract advertises — because a merge archives the loser by definition. The rule is about work still to do; a decided pair is a record of what somebody decided. Liveness now binds the open lane only.

**A gap the list fix did not reach.** A pair is addressable by id, and the review route and both dispose verbs enter through that read, which probed visibility but not liveness. So the queue stopped serving a dead pair while a kept link still opened it, evidence and all. Acting on it was worse than reading it — `merge_organization.go` already names the outcome: "a pair no human can ever dispose of", because merging answers `AlreadyMerged` and that reopens it.

`auth.EnsureVisibleLive` is the probe for that, and its own doc says why the plain one will not do.

**One thing worth flagging rather than burying.** Choosing the probe through a value means the write-authority gate can no longer see it — that gate reads the probe *spelled inside* a mutating function's body. This is a read, so it was waived there before and needs no waiver now; I deleted the waiver rather than re-keying it, and confirmed the gate passes for that reason and not because the probe vanished. The behaviour is held by a test over the route a reader actually takes instead.

**The third finding was the index question**, which the plan measurement in the comment above answers: the partial index still drives the ordered scan and the sort stays incremental. The reviewer could not reach Postgres, so it flagged this as structural rather than measured.

Also added: a test for `DuplicateCluster`. It had none at all, which is how it kept both its missing filters — and it is a spend fence.

PR #3216, opened 2026-08-30
feat(privacy): the audience means something to the system readers too

LarsGradion:
## What this changes

`activity.audience` says who may read a message's content. Every colleague-facing read path composed it; five system readers did not, so a message limited to its participants was still embedded into the shared vector index, mined for signature fields written onto a workspace-readable person, labelled onto a shared worklist, summarised into a signal, and exported whole in a subject-access package.

Each reader now composes it:

| Reader | Before | After |
|---|---|---|
| `search/embedgen.go` | indexed subject+body of every activity | `audience = 'workspace'` only; the vector is deleted when a row narrows and rebuilt when it widens |
| `people/enrichsignature.go` | mined the newest inbound mail whatever its audience | open mail only, plus the statutory availability test |
| `activities.ClassifyBacklogPredicate` | labelled every captured email | open mail only; the label is cleared on narrowing |
| `compose/signalextractread.go` | admitted a limited single-owner thread and sent its bodies to the model | refuses any thread carrying a limited message |
| `privacy/sar.go` | exported subject, body, attachment name and raw provider payload | lists the activity with id, kind, date, source and the mailbox holding it; content withheld |

`capture_classify` and `signal_extract` are pinned `no_payload`.

## The leak this closes

`TestARepsOwnAuditTrailStopsAtALimitedMessage` fails on `main`. The audit export admitted a row two ways — the target record is visible, **or** `actor_id = me` — and the second arm asked nothing about the activity. A colleague who touched a message before its audience was limited kept exporting its before-and-after image afterwards.

The polymorphic clause's unbounded branch is corrected in the same file, and **nothing was leaking through it**: `person` and `organization` are owner-private, so `UnboundedFor` is false for every human including an admin, and an admin export already faced the audience through the bounded arms. The change keeps that true if the owner-private set changes; the comment says so rather than claiming a fix.

## Signal extraction: why refusal, not owner-scoping

A limited thread used to fall back to the mailbox owner, the way a capture-private *record* does. The two look alike and are not. A record's owner visibility says one person is the reader, so a summary addressed to that person discloses nothing new. A limited audience says the content is withheld from readers who can still see the records it is filed against, and an owner-scoped signal is a durable, searchable restatement that outlives the message's own limit. There is no owner for whom extracting it is free.

## The gate

A second dimension on the existing statutory-hold census (`restrictedreaders_test.go`) rather than a second census: same corpus walk, same call graph, same per-function granularity, same waiver discipline, different markers. Its corpus is narrower — readers that project `subject`, `body`, `raw`, `filename`, `evidence_snippet` or `payload`, not every reader — because the hold is an availability rule while the audience is a content rule, and asking every reader produced 68 waivers, which is a list long enough to hide a real finding. Five readers are waived by name with the cost stated.

## The pipeline trace

The attention-label ladder gains `audience_limited`. Without it a member whose message was skipped would be told the batch simply had not reached it, and would wait for a pass that will never run. The trace read is itself audience-gated, so only the mailbox owner gets far enough to be told why.

## Verification

- `make check` green; `make craft-static` clean on all four roots.
- Full integration lane: **3,872 tests, 48 packages, 0 skips**.
- Every new test was mutation-checked: reverting the gate it guards makes it fail. The census was mutation-checked the same way (removing `embedgen`'s clause reports it, and surfaced a second reader through the same query map).
- Each acceptance test was run against `origin/main`'s copy of the file it guards, and fails there.

`privacy/sar.go` crossed the 500-line ceiling, so the gather list moved to `sarsections.go` — which tables hold data about a subject, apart from how the export runs. Three gates name the file their subject lives in; all three fail loudly on a stale key rather than passing quietly.

## What this does not do

This is the first of twelve PRs. It makes the audience column mean something before anything is built on it. Nothing yet writes a limited audience automatically — that is the capture posture, the sender ladder and the confidentiality verdict in the PRs that follow.

---
## Summary by cubic
`activity.audience` now means something to the system readers: five internal readers that ignored it compose it, so a message limited to its participants is no longer embedded, mined for signature fields, labelled onto a shared worklist, summarised into a workspace-visible signal, or exported whole. The audience is re-tested where each write lands, so a limit set between a reader's select and its write-back cannot be undone; no writer sets a limited audience yet, so this is groundwork for the capture and confidentiality work that follows.

**Bug Fixes**
- The embedding generator indexes only `workspace` rows; narrowing deletes the vector and clears the attention label, widening re-indexes it, and the write re-checks the audience, with every path locking the activity first so a concurrent narrowing cannot deadlock against the indexing write.
- The rescope consumer corrects toward the row it finds, not the event that announced it, so an overtaken audience change cannot leave derived data on a row that is workspace again.
- Signature enrichment and the attention-label backlog skip limited mail, re-test at write time under FOR SHARE so a concurrent narrowing cannot slip by, fail closed on a source erased or archived mid-pass, and report `audience_limited` instead of a misleading batch status.
- Signal extraction refuses a thread carrying a limited message instead of falling back to the mailbox owner, and the thread window composes the audience itself.
- The subject-access export lists limited activities with id, kind, date, source, and mailbox, withholds subject, body, attachment name, and raw payload, and folds case on both sides so a capitalised import address is no longer missing.
- The admin export no longer admits limited messages through its own-actor audit arm (attachment rows included) or its unrestricted record branch.
- The lead card's next-task title and deadline select the same row, and the extraction offer refuses a thread holding a hand-logged limited message the window would see.
- The label write also guards the statutory hold, and `capture_classify` and `signal_extract` are pinned `no_payload`.

**Refactors**
- The subject-access export's table list moved to `sarsections.go`; the gates that name it were repointed.
- The audience-reader census walks per file and strips comments, marking only a comparison or a write.

<sup>Written for commit 9525bc174d9fa6cd6a6c935fa15987be7172ed0a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/3216?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Privacy Enhancements**
  - Audience-limited activity content is excluded from exports, search indexing, embeddings, signal extraction, classification, and enrichment.
  - Attachments, audit images, raw captures, and message bodies now respect activity visibility.
  - Narrowing an activity’s audience removes previously generated derived data and prevents stale updates from restoring it.
  - Restricted or limited messages receive clearer processing status indicators.

- **Data Access**
  - Subject-access exports provide appropriate metadata while withholding restricted content.

- **Documentation**
  - Updated privacy gate inventory and coverage records.


LarsGradion:
---

## Adversarial review round 1

An adversarial pass found five further defects, all fixed in `12de689d`. Four are one bug in four places: **a clause in the SELECT does not bind the write.**

Each of these readers selects its subject in one transaction, spends a model call, and writes in another. Gating only the read leaves a window exactly one model call wide in which the message is limited and the write lands anyway — and nothing clears it afterwards, because the narrowing has already run.

| Writer | What landed on a limited message | Fix |
|---|---|---|
| `SetCaptureLabel` | the attention label, re-applied after the retraction cleared it | `AND audience = 'workspace'` in the UPDATE |
| `UpsertEmbedding` | the vector, re-inserted after the narrowing deleted it | audience added to the existing write-time restriction arm |
| `ApplySignatureFields` | title, phone, employer — **permanently**, since the rescope does not retract profile fields | source re-tested at apply time |
| `threadWindow` | the body, sent to the model | audience composed in the window read |

The row lock I added in the first commit did **not** close the label race: the classifier never takes the lock, so serialising made the losing schedule deterministic rather than safe. The fix belongs in the writer.

The `threadWindow` waiver claimed the offer settles it. It does not — `dueThreads` commits and `readThread` opens its own transaction, so under read-committed the window is a later statement against a newer snapshot. The waiver is deleted, and the gate reported it stale on the very next run.

### The SAR clause was fail-open

`NOT EXISTS (… audience <> 'workspace')` is **true when nothing joins**. A raw original whose activity was erased, or that never correlated, was disclosed in full — absence of a limiting row read exactly like permission. It is a positive match now, pinned by `TestASubjectAccessExportWithholdsARawOriginalNothingVouchesFor`, which seeds an orphan alongside the open and limited cases.

Mail correlates exactly: `sinkraw.go` stores the original under the activity's own natural key. **Telegram cannot** — raw is `bot:update_id`, the activity is `bot:chat_id:message_id` — so a Telegram original is withheld here whatever its activity says. Withholding is the side to be wrong on, and it is not the end of Art. 15 access: the row is still listed, and the message itself travels in the Activities section under its own test. Making the keys correlate is a change to what Telegram stores, not a predicate to widen.

### The census could pass on a comment

The marker was a bare `\baudience\b`, which any mention satisfied — including a SQL comment. It would have read this tree's own explanations of the obligation as the obligation being met. Comments are now stripped and the marker requires a comparison or a write. A planted reader projecting `a.audience` beside an ungated body is reported; under the old marker it passed silently.

### A waiver whose premise was false

The correspondence gate's waiver said it reads "the owner's own outbound". It does not: there is no `captured_by` constraint, so mailbox B's sink reads text from mailbox A's limited outbound, and A's limit is observable in B's result as a record that does or does not get created. The waiver now states that cost rather than denying it, and says why narrowing it is a product decision about what "corresponded" means rather than something to fix inside a waiver.

### Re-verified

- `make check` and `make craft-static` green.
- Full integration lane: **48 packages, 0 skips**.
- Every fix mutation-checked. The `threadWindow` test was checked in both directions: it passes with the offer clause removed (so the window is what holds it) and fails with the window clause removed.
- One of my own tests passed alone and failed in the lane — it assumed an empty backlog in a package that shares a database. Fixed to assert about its own row.

LarsGradion:
## Review round 2

Five findings, all fixed in `ff38612b`. Two were real user-visible defects.

### The lead card showed one task's title beside another task's deadline

`leadColumns` has two sibling subqueries for "the next open task": one takes its subject, one takes its `due_at`. I gated the subject on the audience and left the due date alone, so on a lead whose next open task is limited they picked **different rows** under the same `ORDER BY`. The card then rendered task A's title next to task B's deadline, with nothing to say so.

They describe one task, so they select one row. My earlier comment claiming the sibling subqueries "owe nothing" was wrong the moment they had to agree with each other.

### The thread offer was connector-scoped, the window read is not

`dueThreadsQuery` filters `captured_by LIKE 'connector:%'` before the audience test, so a **hand-logged** limited message on a captured thread was invisible to the offer while the window read — which has no such filter — could see it.

No limited text reached the model: the window clause added in round 1 excludes that message's body. What survived was the shape — the pass summarising a conversation while one of its messages is withheld from the summary's readers, with nothing on the summary saying a part is missing.

The test is now a correlated subquery over every email on the thread, written as `NOT EXISTS (… audience <> 'workspace')` rather than `bool_and(audience = 'workspace')`. The two differ on a thread whose messages are all somehow absent, and only this direction refuses it. Pinned by `TestAHandLoggedLimitedMessageTakesItsThreadOutOfThePass`, mutation-checked by re-adding the connector filter inside the subquery.

### Three smaller ones

- **`SetCaptureLabel` also guards `restricted_at IS NULL`.** The trigger already refuses a label arriving *after* a hold; this covers the other order, which commits the hold with a derived label on the row that the trigger then refuses every attempt to remove.
- **The census accepted `audience IS NOT NULL`.** The column is `NOT NULL`, so that is a predicate every row passes. `IS` is dropped; only a comparison against a value counts. Same fail-short class as the comment-matching bug from round 1.
- **The raw-capture fixture used fixed `source_id` values** under a `UNIQUE (source_system, source_id)`. A rerun against a database the suite did not rebuild would have failed on a duplicate key rather than an assertion.

### And one the tree caught on me

The signal-extraction suite crossed the 1000-line test ceiling, so the audience cases moved to `signalextractaudience_integration_test.go` — what the pass may **read**, apart from what it concludes. The new fixture there seeded with `now() - interval` while the suite freezes its clock; `TestAFrozenClockFixtureDoesNotSeedAgainstTheDatabaseClock` failed it in the lane, and the timestamp is derived from the frozen value now.

`make check`, `make craft-static` and the full integration lane (48 packages, 0 skips) are green, and re-verified after rebasing onto `3de8dc35`.

LarsGradion:
## Review round 3

Round 3 of 3. Anything found after this becomes an issue rather than another push.

### One of my own tests was vacuous, and the reviewer was right

`TestAThreadReadAgainAfterANarrowing…` claimed to pin the `threadWindow` audience clause. It did not. My mutation check removed the offer clause **and** the window clause together, so the failure came from the offer; with only the window clause removed the test still passed.

The window cannot be reached through the pass at all: `dueThreads` refuses a thread any of whose messages is limited, so by the time one is offered its window has nothing to exclude. It is asserted against the statement now, in `compose/signalextractwindow_integration_test.go`, calling `threadMessages` directly with a thread the offer would have refused — which is exactly the state the interval between the two transactions produces. **That** test fails when the window clause alone is removed. The original test keeps the claim it does prove, under a name that says so.

### The rescope consumer corrected towards the event, not the row

The bus is at-least-once and events arrive out of order. A message narrowed and widened again before the consumer ran had its vector deleted and its label cleared on a row that was `workspace` — and nothing rebuilt them, because the widening's own event was handled first and found the row already indexed. It reads the row now; the event still decides that a correction is *due*, never which direction.

Two existing fixtures changed with it: they sent the event without narrowing the row, which is a state `SetAudience` never produces. The lane caught that, which is the "seed the way the writer writes" rule doing its job.

### The audit own-trail arm had a second route

I excluded `entity_type = 'activity'` and left `attachment` through. An attachment's audit image carries its filename, and a filename states what the message is about — the same content by the longer route. Both shapes take the parent message's audience now, and an attachment of an **open** message still travels, which the test asserts alongside.

### Two write predicates gain a `FOR SHARE`

Read-committed evaluates a subquery against the snapshot taken when the statement begins, so a narrowing committing a moment later is invisible to it, and the retraction that follows deletes a vector the insert has not yet written. The lock makes the two order.

Same shape for the signature apply — whose fields are the ones no later correction reaches, since the rescope does not retract profile fields. Its subject lock moves above it (`HoldWritableLive`): the eraser is subject-first, and `subjectlockorder_test.go` failed my first attempt for exactly that reason. The source check also covers `archived_at` now, since the candidate query requires it.

### The census corpus could fail short

It required the content column and the `FROM activity` to sit in the **same string literal**. This tree assembles one query out of several fragments — which is why the sibling census walks `activityFragmentsIn` at all — so a reader whose projection and whose `FROM` live in different literals was never selected, and never examined. Per file now, with comments stripped on the corpus side too.

### A comment claimed atomicity that does not exist

`RetractDerivedForActivityTx` said the narrowing and the retraction commit together. They do not: every caller is an async consumer reading a committed event, and between the two there is a real interval. Corrected, and now **held by a gate** (`audienceretractioncallers_test.go`) — the write-side re-checks are shaped for the async case, so a caller added inside the narrowing's transaction would quietly invalidate them. The gate was mutation-checked with a planted bare in-package caller, which its first version missed.

### One unrelated fix taken along the way

A LinkedIn connection whose address carries a capital was missing from the Art. 15 package: `person_email` stores folded, an import does not, and the comparison folded only one side.

### Not taken

Two `source_system`/`source` labelling notes on the SAR are pre-existing and describe what that export deliberately names. No change.

`make check`, `make craft-static` and the full integration lane (48 packages, 0 skips) are green.

LarsGradion:
## Review round 4 — a deadlock I introduced

The `FOR SHARE` I added in round 3 could **deadlock**, and that outranks the round cap: a deadlock in the rescope consumer is a narrowing whose residue is never collected.

The retraction deleted the embedding row and *then* locked the activity; the upsert locked the activity and *then* wrote the embedding. Opposite orders. Every path takes the activity first now — the retraction, the upsert, and the rescope read, which also stops correcting from an unlocked snapshot (a widening committing between the read and the retraction would be undone by it, and the widening's own event would find the row already indexed and do nothing).

## The audit scope's second route

An audit row reaches a message two ways: it may target the **activity**, and it may target an **attachment** of one, whose image carries the filename. The own-trail arm excluded only the first, so the same content was reachable one join further out. Both branches take one clause now, written as explicit arms per shape rather than a `CASE` picking the id — that shape's correlation was easy to get subtly wrong and hard to read.

## `ApplySignatureFields` fails closed on a missing source

The fields it writes were read *out of* that message, so a source erased since candidate selection means writing erased content onto a workspace-readable person — into fields the rescope does not retract, and an audit trail that is append-only.

Nine fixtures were handing it a fabricated activity id. They seed the message the production caller always has now, which is what the apply is actually about; a fixture supplying its own version of production proves nothing about production.

## The census widening reached nothing

Round 3 widened the corpus to file scope, and the per-declaration judge still matched **per literal** — so a reader whose `FROM activity` and whose `SELECT subject, body` sit in different constants had no matching literal and was skipped entirely. Both censuses now match and judge a declaration's assembled text.

That immediately found four readers the narrow walk never saw: the retention selectors and the statutory-hold stamper. Each is waived for the reason the noise redaction is — they name content in order to **destroy** it, and gating them would keep a colleague's limited mail forever rather than protect it. A planted split-fragment reader is reported now; under the old walk it passed.

## Two things I got wrong, stated plainly

**My round-3 mutation check was invalid.** I removed the offer clause and the window clause together, so the failure came from the offer. With only the window clause removed the test passed — it was vacuous. Fixed in round 3 with a direct statement-level test; noting it here because the same mistake is easy to repeat.

**A test I wrote this round asserted a precondition that was never true.** It checked that an admin's bundle carries an open message's attachment audit image. The fixture's "admin" is a human, and `person` is owner-private, so `UnboundedFor` is false and it takes the bounded path — where an attachment audit row only ever passed through the own-trail arm, and that principal was not the actor on the row. The row was never exported, before or after the change. The test is deleted rather than patched; the reps' own-trail test covers the route that is actually reachable.

The same reasoning corrects the unbounded branch's comment: that branch belongs to the system principal, not to an admin. It carries the message test regardless, so it stays honest if the owner-private set changes.

`make check`, `make craft-static` and the full integration lane (48 packages, 0 skips) are green, re-verified after rebasing onto `2db3769b`.

PR #3238, opened 2026-08-30
feat(capture): a mailbox owner's other addresses are not contacts

LarsGradion:
## What this changes

A seat's **other** addresses — a send-as alias, a private domain the same person reads, an address they forward from — are declared per user and read by the capture sink twice: to decide a message is wholly internal, and to decide who the creation ladder is even about.

An alias is not a contact, and mail among a person's own addresses is not correspondence with anybody.

## Why the connector cannot do this

The connector decides direction by comparing the `From` header against the one address the grant was made for. A message **sent from an alias** therefore reads as inbound mail from a stranger, and the creation ladder then decides whether to mint a record for the mailbox owner. That is how a founder's own private domain became a contact record every seat could read.

The connector holds one address and no database — `mailmap.Parse` takes a single owner string, and the owner comes from the sealed connection state. So the correction belongs in the sink, where the acting seat and their declared identities are both in reach. The sync principal carries `UserID: grantedBy`, so "the acting seat" is exactly the mailbox owner.

## Two gates, reached by different messages

| Gate | What it sees | The message that reaches it |
|---|---|---|
| `internalOnlyTx` | every party is the seat themselves → drop before storage | owner ↔ own alias |
| `ladderSubjectTx` | the derived counterparty **is** the seat → find the real subject | owner's alias → customer, owner Cc'd |

The second is the one the drop gate cannot help with: that message names a customer, so it is external and captured, and its counterparty is still the alias.

## Per user, never per workspace

One seat's alias says nothing about another seat's mail, and a workspace-wide list would let anyone silence a colleague's counterparty by claiming their address. `TestAColleaguesIdentityIsNotMine` pins that: Rep3 claims an address, Rep1's mailbox syncs, and the mail is captured and the person minted exactly as if no claim existed.

## The audit trail names no declared address

The image carries the claim's id and kind and **never its value**. This is the ruling the user-scoped exclusion already follows, and it applies more strongly here: an owner identity is always one person's, and its whole purpose is keeping a private address out of the CRM. Writing it into the trail would put it back in, where nothing erases it and every admin reads it.

## Design notes

- **`SelfSet` composes `InternalDomains`** rather than restating the suffix rule. A second domain test would disagree with the first on exactly the subdomains nobody writes a fixture for.
- **The `provider` source is declared and unwritten.** Reading Gmail's `sendAs` list needs a scope the connection grant does not request, and asking for a wider scope is a decision about what the product asks of a user — not a detail of this table. The enum exists so that decision does not become a migration.
- **Not retroactive.** Mail already captured under the old reading stays, and a contact already minted from an alias stays until it is merged or removed. `TestWithdrawingAnIdentity…` pins the same in reverse.

## Verification

- `make check`, `make craft-static`, full integration lane (48 packages, 0 skips), all green.
- Eight acceptance cases drive the **production** registry → mailmap → Sink path, because both gates are SQL inside the capture transaction.
- Each gate mutation-checked against its own case. The first version of the ladder test was **vacuous** — the drop gate stopped its fixture before the ladder ran, so removing the ladder's test changed nothing. It now asserts who the ladder is asked about, read from the pending ledger, on a message only the ladder can answer for.

## Next

This is PR 2 of 12. It gives the later PRs a seat-level notion of "that is also me" — the ladder rework (PR 5) and the Senders page (PR 9) both read it.

---
## Summary by cubic
Stops the capture pipeline from treating a mailbox owner's other addresses — a send-as alias, a private domain they read, an address they forward from — as contacts or correspondence. Previously a message sent from an alias read as inbound mail from a stranger and the creation ladder could mint a person record for the owner's own private domain; now a seat declares their identities and the sink drops mail wholly between the owner's own addresses, resolves the real counterparty on threads the owner wrote from an alias, and never stamps the alias as the owner's other end.

**Notes**
- Identity claims are per user, never per workspace, and the gates resolve the acting seat the way the creation ladder does (`OnBehalfOf`, else `UserID`).
- The connected mailbox's primary address counts as the owner automatically via `capture_connection.account_label`, and declaring an identity settles the seat's own pending verdicts about that address.
- The store admits only the human seat themselves — a connector or system principal carrying the seat's `UserID` is refused, and a colleague's identity answers not-found rather than forbidden.
- Address claims fold domains through IDNA; plus-addressing stays a separate address.
- The audit image records the claim's id and kind, never its value.
- The `provider` source is declared but unwritten; reading a provider's send-as list needs a scope the grant does not request.
- Not retroactive: mail already captured and contacts already minted from an alias stay until merged or removed.
- Adds the `capture_owner_identity` table and three human-only endpoints: `GET`/`POST /capture/owner-identities` and `DELETE /capture/owner-identities/{id}`.

<sup>Written for commit 02cb2ba1963b5421b4d98d8a63b79853168b541a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/3238?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added support for declaring personal email addresses, aliases, and domains as owned identities.
  - Added API operations to list, create, and remove owned identities.
  - Identity creation is idempotent and restricted to human users.
- **Bug Fixes**
  - Prevented personal addresses and aliases from being recorded as external contacts.
  - Improved handling of internal messages, counterparty selection, and pending decisions.
  - Audit records no longer expose declared identity values.
  - Added validation and normalization for identity values.


LarsGradion:
## Adversarial review round 1

Six findings, four of them holes in the feature itself. Fixed in `e0d1052c`.

### The mailbox's own address was not in the set

Nobody declares it — the grant established it — so the gates protected a seat's **aliases** and left their **primary address** exposed. On a consumer mailbox (`owner@gmail.com`) the workspace's own domains do not cover it either, so a message between the owner's two addresses stood the owner in as their own counterparty, and the ladder was asked whether to create a record for them.

It comes from `capture_connection.account_label`, which every mail connector reports at grant. `TestTheConnectedMailboxsOwnAddressIsNeverTheLaddersSubject` pins it.

### A claim did nothing about questions already open

`Add` only inserted the row, so the gates bound mail arriving from then on. An address deferred moments before the claim kept its pending row, a verdict landed on it afterwards, and the seat's own address became a contact through exactly the door the claim was made to close.

Claims now settle the seat's own open ledger rows to `suppressed` — **not deleted**: the Senders surface reads them, and a decision that vanished would be one nobody could see or reverse. Scoped to the claimant: `TestAClaimSettlesNoColleaguesOpenQuestion` pins that a colleague's pending decision about the same address is theirs.

### The gates read `UserID`; the ladder reads `OnBehalfOf`

They agree in every principal the registry builds today. A gate that depends on that agreement without asking for it is one refactor away from applying one seat's claims while ingesting another's mail. Both gates resolve the owner through `capturePrincipal` now, the same way the creation ladder does.

### `auth.RequireHuman` was not the boundary this store claimed

It refuses an agent and a buyer and **admits the connector and system principals** — correct for the reads it guards, wrong for a list that says which of a person's mail is theirs alone. A connector principal carries the granting seat's `UserID`, so under `RequireHuman` alone it could list, add and withdraw that seat's identities. The store tests for an actual human seat.

### Two smaller ones

- **IDNA.** An address claim now folds its domain, so `owner@bücher.example` and `owner@xn--bcher-kva.example` are one address. Domain claims already normalized; the two disagreeing about the same mailbox was the inconsistency. **Plus-addressing stays unfolded on purpose** — `me+news@` is a different address, and treating it as the same silences mail the seat never claimed.
- **Participant rows** took the connector's derived counterparty, not the ladder's corrected subject, so the owner's own alias was stamped as the other end of their own message. Nothing promotes such a row to a person (the promotion needs a `person_email`), so what this removes is a durable falsehood about the exchange rather than a contact.

### One finding kept as a stated boundary

A capture whose mailbox owner cannot be named leaves the set empty, and the gate then answers exactly as it did before this feature existed. The ladder already refuses to create records for a capture with no granting human, so this is the same direction every other gate here fails in: toward keeping a message rather than acting on an owner nobody can name. Said in the code rather than left to be rediscovered.

### Verification

- `make check`, `make craft-static` green; full integration lane green apart from `TestDeepReadCrawlsExtractsAppliesWhatItEvidencedAndFinishesDone`, which **fails on `origin/main` too** (verified by checking out main and running it) and has nothing to do with capture.
- Twelve acceptance cases through the production registry → mailmap → Sink path, plus unit coverage of `SelfSet` at the boundaries where membership tests go wrong: subdomains, the suffix trap (`notother.example` vs `other.example`), case, whitespace, IDNA both directions, plus-addressing, and the empty set.

LarsGradion:
## CI status

One failing check: `TestDeepReadCrawlsExtractsAppliesWhatItEvidencedAndFinishesDone` in integration shard 1. It **is not this branch**.

- It fails on `origin/main` — verified by checking main out locally and running the test alone.
- It is the only failure in the shard; the "integration (RLS + erasure + HTTP e2e)" check fails in 11 seconds with no test output because it depends on that shard.
- #3240 ("main is red: a technology lane that changed nothing announces nothing") is already fixing it, is green, and is mergeable. I have not written a competing fix.

This branch clears once #3240 lands and I rebase.

## Two things fixed since the last push

**The frontend contract types.** I ran `make check-go` and not `make check`, so `frontend/src/api/schema.d.ts` was never regenerated from `crm.yaml`. The frontend typecheck reads it, so the drift gate failed in screens this change never touches. Regenerated and committed; `make check` — both halves — is green now.

**The migration number.** Main gained `1788129035` while this branch was open, above mine. A migration below the highest on main still applies, but in a different place on each installation — before it on a fresh database, after it on one already past — and anything order-dependent then leaves two schemas different with nothing to report it. Renumbered to `1788130107`.

That rename has a trap worth naming: the old version stays in an existing ledger, so the renamed file re-runs and dies on `relation already exists`. A test template carrying it has to be dropped rather than migrated forward, which is what I did locally before re-verifying the lane.

LarsGradion:
## Review round 2

Five findings, all fixed in `6858ef30`.

**The account label can be display-form.** A provider may store `Rep <rep@co>` — which is exactly why `bareAddress` exists in the registry — and my SQL took it raw. The seat's **primary** address was therefore missing from the set with nothing to say so. Parsed in Go now, through the same helper.

**The ledger and the gates used different tests.** Retirement compared `split_part(email, '@', 2)` against the claim; the gates compare through `SelfSet`, which folds case, normalizes IDNA and covers subdomains. The two agreed on the ordinary cases and disagreed on exactly the ones a claim gets made for. Rows are chosen through `SelfSet` now — a ledger and a gate answering differently about the same address is the shape nobody notices until a record appears.

**`WithoutSelf` kept blank entries.** `AllInternal` already skips one, so letting it through made an owner-only message look like it still had somebody else on it. "Nothing external is left" has to mean the same thing in both places.

**The audit waiver said the opposite of the code.** It claimed the image names the declared address; the image carries the claim's id and kind precisely so it does not. A comment that inverts the behaviour is worse than none.

**The delete test proved a different claim.** It used a fabricated id, so it showed that a *missing* row is 404. It plants a colleague's row now and asserts both the 404 and that the row survives — the property is that an existing row belonging to somebody else is indistinguishable from one that is not there.

Plus the migration named `capture_own_domain`, which does not exist; the workspace list is `workspace_email_domain`.

### One finding taken as a decision rather than a fix

Three gates read the identity list per message. Each is one indexed lookup on a per-seat table holding a handful of rows, inside a transaction already writing an activity, its links, its participants and its trace. Threading a value through those three call sites would be an abstraction bought with no measurement, so it stays three reads with the reasoning in the code.

### Coverage

The handlers were the whole of this branch's uncovered new code (28 of 55 lines, 0%). They have an HTTP suite now — round-trip, refusal of a malformed claim, and the colleague case above — and SonarCloud's new-code gate passes.

All 26 checks green.

Issue #3282, opened 2026-08-31
main is red on check-fe: personprovider fixture is missing the required 'applied' field


LarsGradion:
`make check-fe` fails on `origin/main` (currently 7d50d0b05), independent of any open branch.

```
src/screens/personprovider.fixtures.ts(11,14): error TS2741:
Property 'applied' is missing in type '{ id: string; subject_kind: "person"; ... }'
but required in type '{ ...; applied: ...; }'
```

The contract gained a required `applied` field on the person-provider run shape (#3278, "one installation-wide answer on automatic lookups") and the frontend fixture was not updated with it. `frontend/src/api/schema.d.ts` on main declares it required; `frontend/src/screens/personprovider.fixtures.ts` does not carry it.

**Reproduce:** `git switch main && git pull && make check-fe`.

**Why deferred here:** it is not my change — found while gating an unrelated branch, and fixing another team's fixture inside a worklist PR would hide it. The fix is a one-line addition to the fixture, by whoever knows what `applied` should be for that case.

**If nobody fixes it:** every frontend PR is red on a gate that has nothing to do with it, and people learn to ignore `check-fe`.

LarsGradion:
Fixed on main: `personprovider.fixtures.ts` carries `applied: true` and `make fe-typecheck` passes against `c106da438`. The fixture was updated alongside the applier (#3292) — verified just now by running the typecheck on current main, which is clean.

PR #3288, opened 2026-08-31
Who is waiting for a reply is its own source on the Worklist

LarsGradion:
The concept's second ten-second question — **"which customer is waiting on me?"** — had no producer. This adds one.

## Why it could not come from the at-risk lane

The queue could previously reach a waiting buyer only through deals going quiet, and that lane cannot answer the question: **a fresh inbound makes a deal less quiet**, so the deal drops out of the candidate set exactly when somebody starts waiting on it. A person with no deal at all was unreachable.

## The read

`activities.WaitingReplies` asks it directly — per thread, the newest inbound with no later outbound. One row per thread, because a customer who wrote three times is waiting once. Oldest first, because the longest wait is the one that was forgotten.

Two gates, deliberately different:

- The **discover** clause decides which rows are listed.
- The **audience** arm decides whether their words come too, and rides on the row as `readable`. A withheld message still proves somebody is waiting; it just cannot be answered from here, so no draft is offered and the subject line does not travel.

The outbound side **ignores** the audience arm on purpose: a reply this reader may not read still answered the customer. Skipping it would report a message as unanswered because the answer was somebody else's to see — which sends a rep to write a second reply, the worst failure available here.

## Ranking

Level 1, above every promise, deal and decision. It is the one thing on the page whose cost of inaction falls on somebody else.

One unanswered message is **one row**: the deal it belongs to no longer also appears as drifting. The waiting row wins because it is more urgent and more actionable — it names the message to reply to.

## Two faults the security review caught

**The subject line leaked.** It went onto the row whatever the audience arm said. Being admitted to know somebody is waiting is not being admitted to read what they wrote.

**Thread matching joined every NULL to every other NULL.** `IS NOT DISTINCT FROM` meant one unthreaded outbound anywhere in the workspace would silence every unthreaded question at once. Unthreaded messages are now excluded — under-reporting by a row rather than by a customer.

Both pinned by tests.

## Seen working

Against a seeded stack with an inbound email from 83 days ago on a €160,100 deal:

```
#1 L1 customer_waiting  Re: pricing for the retrofit   because=[buyer_wrote_last, waiting_days]
#2 L4 task              Confirm the workshop date
#3 L4 deal_at_risk      Globex Renewal
```

The Acme deal does **not** appear as drifting — the dedupe fired. Before this change the buyer was invisible and the deal sat below nine hygiene decisions.

## Gates

`make check-backend`, `make craft-static` (0 findings) green. `make check-fe` is red on `main` for an unrelated fixture (#3282), not this branch.

---
## Summary by cubic
Adds `customer_waiting` as its own Worklist source so the queue can answer "which customer is waiting on me?". Previously a waiting buyer could only surface through the at-risk deal lane, which drops a deal exactly when a fresh inbound arrives (a new message makes it less quiet) and never reaches a person with no deal at all.

- Ranks Level 1 — above promises, deals, and decisions — one row per thread, oldest wait first.
- A deal with an unanswered message no longer also appears as drifting; the waiting row wins.
- A withheld message still proves somebody is waiting, but travels without its subject line or a draft action.
- Unthreaded messages are now excluded, fixing the `IS NOT DISTINCT FROM` join that let one unthreaded outbound silence every unthreaded question in the workspace.

<sup>Written for commit 8389d0bb1cb8f5a56707a7e7bfb780ada9c0480d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/3288?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added a customer-waiting source to CRM activity data.
  - Added a worklist lane for unanswered inbound messages.
  - Waiting items show wait duration, related records, timestamps, and available actions.
  - Waiting items are prioritized and duplicate deal entries are removed.
  - Message subjects remain hidden when unread or not permitted.

- **Bug Fixes**
  - Excluded unthreaded messages from waiting-reply results.
  - Improved permission and error handling when loading waiting items.

### DIFF
--- a/backend/api/crm.yaml
+++ b/backend/api/crm.yaml
@@ -31249,7 +31249,7 @@ components:
         id: { type: string, description: "The owning record's id, as its own endpoint spells it." }
         source:
           type: string
-          enum: [approval, dedupe_candidate, task, brief_item, conversation_claim, deal_at_risk, meeting, relationship_decay, failed_approval, dsr, sync_health, capture_health, ai_work_health, bounce, automation_run, notice]
+          enum: [approval, dedupe_candidate, task, brief_item, conversation_claim, customer_waiting, deal_at_risk, meeting, relationship_decay, failed_approval, dsr, sync_health, capture_health, ai_work_health, bounce, automation_run, notice]
           description: "Which producer raised it, and therefore which endpoint its verbs go to."
         kind: { type: string, description: 'The producer''s own sub-type (an approval kind, a dedupe entity type) — for the icon and the label, never for authority.' }
         title:
@@ -31444,7 +31444,7 @@ components:
         id: { type: string, description: "The owning record's id, as its own endpoint spells it." }
         source:
           type: string
-          enum: [approval, dedupe_candidate, task, brief_item, conversation_claim, deal_at_risk, meeting, relationship_decay, failed_approval, dsr, sync_health, capture_health, ai_work_health, bounce, automation_run, notice]
+          enum: [approval, dedupe_candidate, task, brief_item, conversation_claim, customer_waiting, deal_at_risk, meeting, relationship_decay, failed_approval, dsr, sync_health, capture_health, ai_work_health, bounce, automation_run, notice]
           description: 'Which producer raised it, and therefore which endpoint its verbs go to.'
         category:
           type: string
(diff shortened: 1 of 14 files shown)


LarsGradion:
Superseded by a rebase onto current main — the same work, plus the fixes from the Codex and security reviews, is in the replacement PR. Closing rather than force-pushing this branch.

PR #3290, opened 2026-08-31
feat(people): the account roster leads with the contact worth writing to

LarsGradion:
## The bug

The company page's People section carries 25 of an account's contacts. It took the first 25 **by person id** — the roster read's own deterministic order, which is arbitrary as a reading order.

On a 105-contact account the one person who answered last week sits wherever their id falls, so the section that is supposed to open the page could omit them entirely while reporting nothing worse than `has_more`. The reader is told there are more contacts; they are not told the one that mattered was among them.

## What changed

**`people.EngagementOf`** — `answered` / `no_reply` / `untried`, read off the 90-day direction counts the §4 fold already produced. The rule previously lived in the browser (`coverage.ts` `reachOf`), where it could only classify the contacts a page had already been sent — which is the 25 chosen by id. It belongs where the whole account is in hand.

**`people.RankContacts`** — engagement, then strength, then id. Answered leads because they are the way in. Untried comes *second*: on an account where everyone has gone quiet, the person nobody has written to is the only move left that is not a fourth follow-up. The id tie-break is what makes a paged read stable.

**The strength fold gains three values** — the last inbound date, the last outbound date, and the latest inbound activity id. The counts say *that* a contact answered; the dates say which way the conversation is owed, and a contact who replied in March and was chased in August has both counts non-zero. The activity id is the anchor a follow-up opens a reply against rather than starting a fresh thread.

`contactsSection` ranks a **copy** before truncating: the same slice is folded by the account strength roll-up, and reordering it under that reader would make the summary depend on which section ran first.

## Cost

No contract change and no new query. The fold gains its columns inside the statement it already ran, in **both** the batch and per-person paths so the two cannot disagree. `org360_querycount_integration_test` still passes — the read is flat in the account's contact count.

The reply anchor is a correlated `LIMIT 1`, deliberately not `array_agg(...)[1]`: the aggregate form materialised every inbound id in a contact's entire history in order to use one of them, which is the payload growing with history that this fold exists to avoid.

## The anchor is windowed; the dates are not

These answer different questions and the asymmetry is on purpose.

A date is history — "they last wrote in March" stays true however old it gets, and the page prints it. The anchor is an **action**, and it has to agree with the state the same read reports. A contact whose only message is eighteen months old reads `untried`; carrying an anchor for them would let the page call someone never approached and simultaneously offer to follow up on a thread from last year.

Both were defects in the first commit here, found in review and fixed in the second.

## Tests

Unit (`internal/modules/people`):
- `TestEngagementOfReadsTheThreeStatesOffTheDirectionCounts` — including the case that separates the two rules: inbound *and* outbound both non-zero still reads `answered`.
- `TestRankContactsOrdersAnsweredThenUntriedThenNoReply`, `...PutsEngagementAboveScore` (the highest-scoring contact is the silent one, so score-led ranking would open the page on the person whose next move needs a reason), `...BreaksAScoreTieByStrengthThenID`, `...IsStableAcrossInputOrder`.

Integration (`internal/compose/integration/org360`):
- `TestOrganization360RanksContactsBeforeTruncatingTheSection` — 30 contacts, the only one who answered is last in id order. Fails against the previous behaviour.
- `TestStrengthFoldDatesEachDirectionSeparately` — two inbound messages, not one, so the anchor must pick the newer.
- `TestStrengthFoldLeavesTheInboundAnchorEmptyWhenNobodyWroteIn`
- `TestStrengthFoldDropsAnInboundAnchorOlderThanTheWindow` — the correctness defect above.

Every guard test here was mutation-checked: the rule was inverted or removed and the test confirmed to fail. That is what caught the single-inbound-message fixture, which passed against a deliberately reversed sort order and proved nothing until a second, older reply was added.

## Verification

`make lint` 0 issues; `internal/modules/people` and `internal/compose/integration/org360` integration suites green; `go test ./gates/` green apart from two failures that reproduce on clean `origin/main` and are caused by untracked local files (`.pnpm-store/`, an untracked docs page) that do not exist in the repo.

## Scope

This is the data layer plus the ranking fix. Nothing is visible to a user yet — the redesigned People tab reads this in a later PR.

---
## Summary by cubic
The people section now ranks an account's contacts by engagement before the 25-row cut, so the page opens with the contact worth writing to instead of the first 25 by person id.

**Ranking**
- Contacts sort by engagement (answered, then untried, then no_reply), then strength, then id; the id tie-break keeps paged reads stable.
- `people.EngagementOf` replaces the browser-side `reachOf` rule, which could only classify the 25 contacts already sent.
- `contactsSection` ranks a copy so the account strength roll-up's folding order isn't affected.

**Reply anchor**
- The strength fold gains last inbound date, last outbound date, and the latest inbound activity id, in both the batch and per-person paths.
- The anchor is bounded by the same 90-day window as the counts, so a stale thread is never offered as a follow-up; the dates stay unwindowed.
- The anchor is read as a correlated `LIMIT 1` rather than an array aggregate, so the payload doesn't grow with a contact's history.
- No contract change and no new query: the fold gains columns in the statement it already ran.

<sup>Written for commit 08a0f4ecd3fd9d88c9c8236e25013be44209e816. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/3290?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Organization contact lists now prioritize contacts based on recent engagement and relationship strength.
  * Contact engagement is classified as answered, no reply, or untried.
  * Recent inbound and outbound activity are tracked separately, with inbound activity limited to the past 90 days.
* **Bug Fixes**
  * Ensured the most relevant 25 contacts appear first when lists are truncated.
  * Preserved historical engagement dates while excluding outdated inbound activity.

### DIFF
--- a/backend/internal/compose/integration/org360/org360_contactrank_integration_test.go
+++ b/backend/internal/compose/integration/org360/org360_contactrank_integration_test.go
@@ -1,0 +1,68 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+//go:build integration
+
+package org360
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/margince/margince/backend/internal/compose/integration"
+	"github.com/margince/margince/backend/internal/shared/kernel/ids"
+)
+
+// The people section carries 25 of an account's contacts, and which 25 is a
+// product decision the roster read does not make: it returns them ordered by
+// person id, which is arbitrary as a reading order. Cutting that order keeps
+// the first 25 ids rather than the 25 contacts worth looking at.
+//
+// The fixture puts the ONLY contact who has answered at the end of the id
+// order, which is where the old cut dropped them. A reader opening the account
+// then saw twenty-five untried strangers, no way in, and a has_more flag that
+// truthfully reported more contacts while saying nothing about the one that
+// mattered. Person ids are UUIDv7 and therefore ascend in creation order, so
+// seeding the answered contact last is what places them outside the old cut.
+func TestOrganization360RanksContactsBeforeTruncatingTheSection(t *testing.T) {
+	e := integration.Setup(t)
+	ctx := e.Admin()
+
+	owner := integration.OwnerConn(t)
+	org := e.SeedOrg(t, "Brandt GmbH", nil)
+
+	const contacts = 30
+	var answered ids.UUID
+	for i := range contacts {
+		person := e.SeedPerson(t, fmt.Sprintf("Contact %02d", i), nil)
+		employ(t, e, person, org, "Fleet")
+		answered = person
+	}
+
+	// One inbound message: the single fact that separates a way in from a
+	// roster. It lands inside the 90-day window ending at the pinned clock.
+	mail := integration.AccountMailDirectedAt(t, owner, e.WS, "Re: your proposal",
+		"inbound", org360Clock.AddDate(0, 0, -3))
+	integration.LinkActivity(t, owner, mail, "person", answered)
+
+	view, err := org360Service(e).Assemble(ctx, ids.OrganizationID{UUID: org})
+	if err != nil {
+		t.Fatalf("assembling the 360: %v", err)
+	}
+	if view.People == nil {
+		t.Fatal("the people section is absent for a caller who may read it")
+	}
+	rows := view.People.Data
+	if len(rows) != 25 {
+		t.Fatalf("the section carried %d contacts, want the 25-row cut", len(rows))
+	}
+	if !view.People.Page.HasMore {
+		t.Fatal("a 30-contact account must report has_more on a 25-row section")
+	}
+
+	// The one who answered leads, because they are the way in.
+	if got := ids.UUID(rows[0].PersonId); got != answered {
+		t.Fatalf("the section opens on %s; the contact who answered is %s and must lead",
+			got, answered)
+	}
+}
--- a/backend/internal/compose/integration/org360/org360_lasttouch_integration_test.go
+++ b/backend/internal/compose/integration/org360/org360_lasttouch_integration_test.go
@@ -1,0 +1,129 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+//go:build integration
+
+package org360
+
+import (
+	"testing"
+	"time"
+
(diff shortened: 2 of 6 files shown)


LarsGradion:
## CI is red, and none of it is from this branch

Three integration failures. Each one reproduces on clean `origin/main` (`616977826`) with an empty working tree — verified locally by checking out main with no changes and running the same test:

| Shard | Test | On this branch | On clean main |
|---|---|---|---|
| 1/6 | `TestAConnectorCreatedPersonIsGovernedByTheImportToggle` | fail | **fail** |
| 2/6 | `TestPersonCreatedWithAutoEnrichOffBuysNothingAndDoesNotError` | fail | **fail** |
| 3/6 | `TestFK_rowScopedTargetsHaveVisibilityDecision` | fail | **fail** |

All three concern provider enrichment and `provider_applied_field` — tables and code paths this branch does not touch. This branch changes the §4 strength fold and the 360 people section only.

Two of them are guards about **spending a customer's credits against their own setting**:

- *"a captured counterparty triggered 1 paid runs under automatic_import=false — connecting a mailbox would buy data for every sender in its history"*
- *"1 runs exist with auto-enrich off — the customer's credits were spent against their own setting"*

The third says `provider_applied_field.person_id -> person` has no visibility decision, so a row-scoped record can be referenced without a read gate.

That looks like a real regression already on main rather than test flake, and it is worth someone owning separately from this PR. I have not merged this branch on top of it.

### What is green

`dco`, `deterministic-gates`, `craftsmanship`, `craft-residue`, `govulncheck`, `secret-scan`, `live-boot`, `extension-reference`, integration shards 4/6, 5/6, 6/6, integration unit coverage, and CodeRabbit.

Locally on this branch: `make lint` 0 issues, and the `internal/modules/people` and `internal/compose/integration/org360` suites pass in full, including the four new tests.

### One correction to my own earlier read

I first reported `TestPersonCreatedWithAutoEnrichOff…` as passing on main. That run reused a cached test-database template; with a fresh one it fails on main consistently across three runs. The earlier "pass" was the wrong answer, not the later failure.

Issue #3296, opened 2026-08-31
First-run setup cannot be completed without a Google OAuth app


tinnhlt-gra:
Found during UAT of the login-methods work, on a stack with no Google credentials configured.

The first-run onboarding gate demands, in order, (a) an AI provider and API key, and (b) a **Google OAuth client id and secret** — with **no skip on either**. Until both are supplied the gate hijacks every route: `#/settings/users` redirects to `#/onboarding/company`.

An operator deploying without a Google app therefore cannot finish onboarding through the UI at all. The only way past it during UAT was to create the anchor organization directly through the API (`PUT /v1/company`).

Not introduced by the login-methods branch, but adjacent to it, and that work makes it stand out: an installation is now fully usable with password sign-in and no external provider, so requiring a Google app to reach the product contradicts what the product supports.

**Expected:** the Google step is skippable, and the app can be configured later from Settings → General, where the card and its redirect URIs now live.

PR #3300, opened 2026-08-31
An invitation is redeemed into an active account, and an admin chooses the sign-in methods

tinnhlt-gra:
## What this fixes

An invited member clicked **Continue with Google** and got a neutral failure with no way
forward. The API said:

```
oidc sign-in refused  provider=google  reason="resolve/link account"
```

Google itself worked — token exchange and ID-token verification both passed, and they refuse
with their own reasons. The refusal came from `resolveFederatedUser`, whose email fallback
requires `password_hash IS NOT NULL`. That gate is **correct and stays**: without it an
unredeemed invitation is claimable forever by whoever controls that address at the IdP — no
token, no expiry, unlike the invitation mail itself.

Four things around it were wrong.

## 1. An invitation now says it is one

`app_user.status` has always admitted `'invited'` in its CHECK constraint. Nothing ever wrote
it, so an invited member read **Active** in Users & teams while being unable to sign in by any
method.

`InviteUser` writes `'invited'`; `RedeemPasswordReset` flips it to `'active'` and emits a new
`user.activated` event with its audit row, in the same transaction — every other status change
in the module already does, and a subscriber holding `user.invited` would otherwise never learn
the invitation completed. One statement still serves both flows: for the forgotten-password
caller the assignment is a no-op.

**This closes a latent lockout.** An invited admin was written `'active'` and therefore counted
toward the last-active-admin guard — so an operator could deactivate the last admin who could
actually sign in, on the strength of an account nobody can enter. The test reproduces it as
`err = <nil>` when the invite status is reverted.

Three surfaces had to widen with it, or the change would have been a regression rather than
half a fix:

- **Issuance** (`supersedeSetPasswordTokens`) refused any non-active target, in **Go**, so no
  SQL census found it. Unwidened, an expired invitation becomes permanently unrecoverable: the
  member has no password, the self-service reset refuses them, and the admin-issued link was
  the only route left.
- **`SetTeamMember`** refused with "reactivate them first" — nonsense for someone who has never
  been active, and `InviteUser` itself assigns teams at invite time.
- **`ReactivateUser`** now restores to `'invited'` when there is no password, rather than
  recreating the inconsistent row this change removes.

The seat ceiling needed **no change**: `fullSeatsInUseQuery` counts `status NOT IN
('suspended','deactivated')` and its doc comment already anticipated exactly this. A regression
test now holds that, because the reasoning rests on it.

## 2. An admin chooses which providers are offered

A **Sign-in methods** card in Settings → General lists every provider this deployment holds
credentials for, each with a switch. Password is listed too and is permanently on — rendered as a
control that exists and refuses rather than an absent row, because an admin who cannot find it
cannot tell whether password sign-in is configured at all.

Behind the card is a new setting, `identity.enabled_oidc_providers`, intersected with what the
deployment actually composed — an operator cannot invent a client id from a settings screen, so this can only ever
narrow.

**Password is not a member of the set.** That is why the entry is named for providers: there is
no value of it that turns password off, so "cannot be disabled" is a property of the shape
rather than a validation a later change could relax.

One seam, `WithOIDCProvidersEnabledFn`, is read by all three consumers — the capabilities probe,
`StartOidcSignIn`, and `OidcSignInCallback`. Checking only the first would leave the routes live:
the button disappears while the endpoint still mints sessions for anyone holding the URL.
Checking the callback is what stops flows **already in flight**. A disabled provider answers the
same 404 as an unknown one; a failed policy read does **not** — 404 would report an outage as
absence and send an operator hunting a configuration bug that is not there.

## 3. The Google app card stops lying

`GET /installation/google-app` read only the store, but the connector resolves **two** sources at
call time — stored wins, environment is the fallback. So with credentials in the environment the
card said *"No app stored. Gmail and Calendar cannot be connected until one is."* The second
sentence was **false**, and `GET /installation/setup` reported the same installation configured,
because it already carries an `envGoogleApp` flag for exactly this reason.

The response now reports `source: stored | environment | none` and shows the client id in use
(not a secret — it rides in every authorization redirect). The card moves to Settings → General,
because the same OAuth client now serves sign-in as well as mailbox connection, and filing it
under Capture said it belonged to one of them.

It also lists **both** callback URLs an operator must register. They are not derivable from each
other — sign-in uses a base that already carries `/v1`, while the connector prefers `apiBaseURL`
and falls back to `publicBaseURL`, so on a split deployment they do not even share an origin.
Registering one and not the other fails with `redirect_uri_mismatch`, which names nothing
actionable. A fitness test holds the advertised URI equal to the one the flow actually sends,
failing in both directions.

## 4. The login screen says what to do

The refusal copy is generic about state and specific about remedy: *"Sign-in with Google didn't
work. If you were invited, open the link in your invitation email to finish setting up your
account."* Identical for an unknown address, an invited one and an active one, so it adds no
signal to probe with — held as a property of the string, since the notice is driven by the
redirect's hash and never by a response.

## The census gate

`ActivatableMemberSQL` is a second named predicate in `livemember.go`, and
`backend/gates/livemember_test.go` was taught about it — necessarily, and for two reasons:

- An **unregistered** helper call flattens to its arguments, so every statement using it would
  render with no liveness halves and pass **unjudged**. That would not have failed anything; it
  would have silently shrunk the census.
- The default failure message reported a hand-spelled activatable pair as a *half-spelling*,
  whose advice is to add `status = 'active'` — precisely the edit that makes redemption refuse
  the invitation its own link was minted for. It now has its own verdict saying so.

The teaching is narrow by construction: it recognises the one predicate the helper emits, and a
**third** status pair is still reported. Both were verified by planting the defect and watching
it fail.

## Review

Reviewed by Fable and Codex twice — once on the design, once on the code.

**On the design**, both found the issuance lockout above, and both found that a first draft would
have edited the licensing meter for no reason (the seat ceiling already handled this, and its own
doc comment said so). Fable additionally found that the widened predicate would red the liveness
census as a half-spelling — which is what forced a named second helper after the seat-ceiling
justification collapsed.

**On the code**, Fable found the new census arm shipped with no probes and a probe scope that did
not know the helper's name, so a regression in the detector would have stayed green; and that a
predicate spelled into a bare declaration was invisible to it — the original org360 incident,
reproduced for the new predicate. Both are fixed and probed. Fable also caught that
`resolveFederatedUser`'s doc comment still asserted "there is no `invited` status", at the most
security-sensitive read in the file; that claim appeared in three places and all three are
corrected. The correction is substantive rather than cosmetic: both halves of that predicate are
still load-bearing, but `password_hash IS NOT NULL` now earns its place by refusing the **agent
seat**, not the invitation.

**Codex** found that the backend accepted the provider policy while no admin could reach it —
the Sign-in methods card above is that gap closed.

The integration lane found three more breaks that `make check` cannot see, all of them this
change behaving correctly against test setups that predated it: an assignee roster that no longer
offers an invited member, an ownership transfer that refuses one, and a consent fixture whose
second admin was invited and therefore could not hold the installation open.

## Testing

`make check` green on the merged tree. Integration lane green. Every new test was watched
failing before being trusted — the last-admin lockout, the provider-policy enforcement, the
redirect-URI drift, the frontend invited-row gates, and both arms of the taught census.

No prompt files are touched, so `aicert` needs no re-certification; verified against the final
diff.

---
## Summary by cubic
Invited members now carry an `invited` status instead of `active` until they redeem their set-password link, fixing a lockout where an invited admin counted toward the last-active-admin guard while being unable to sign in anywhere. Admins can now choose which OIDC providers the login screen offers, and the Google app card reports the app the connector actually uses.

**Invitation activation**
- `InviteUser` writes `invited`; `RedeemPasswordReset` flips it to `active`, emitting `user.activated`; `user.reactivated` now carries the status restored to.
- Admin-issued set-password links now admit invited members, making an expired invitation recoverable.
- `SetTeamMember` and `ReactivateUser` accept invited members; invited rows show as "Invited" in Users & teams.
- Refused federated sign-ins now tell invited users to open their invitation email.

**Sign-in methods and Google app**
- The new `identity.enabled_oidc_providers` setting can only narrow the deployment's configured providers; password is not in the set and cannot be disabled.
- Disabled providers return 404 at both sign-in start and callback, and the anonymous capabilities probe is rate-limited.
- The Google app card reports `source: stored | environment | none`, the client ID in use, and both callback URLs an operator must register; the card moved from Capture to Settings → General.
- Settings cards render defensively, so a response missing a contract-required field no longer takes the settings screen down.

<sup>Written for commit 638df4fc5998d7b6b6adc4c5042567337fa682da. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/3300?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added organization contact listing with search, filtering, sorting, and pagination.
  - Added manual organization VAT checks with queued processing and rate-limit responses.
  - Added invited-member status and invitation redemption with password setup.
  - Added configurable external sign-in providers; password sign-in remains always enabled.
  - Added `user.activated` event support.
  - Google OAuth settings now show configuration sources and authorized redirect URIs, with copy controls.
- **Bug Fixes**
  - Improved handling of invited, suspended, and deactivated members.
  - Disabled external providers are now blocked consistently during sign-in.
  - Added protection against repeated anonymous authentication-capability requests.
- **Localization**
  - Updated English, German, and Vietnamese authentication and settings text.


tinnhlt-gra:
## Manual UAT — executed on a live stack, not simulated

Run by a subagent driving Chrome against a real dev stack (api `:18082`, SPA `:8082`), seeded with `seed-dev`. Recording attached below.

### What was verified in the browser

| Step | Result |
|---|---|
| Invite shows an **Invited** badge, not Active | PASS — grey pill; zoomed side by side against Active's green to confirm it is neutral, not amber |
| Both row actions offered on an invitation | PASS — `Get set-password link` **and** `Deactivate` |
| Set-password link redeems | PASS |
| Member then reads **Active** | PASS |
| Sign-in methods card | PASS — password switch `aria-checked=true`, `disabled=true`; clicked it directly and it did not move |
| No Google button when no provider configured | PASS |
| Refusal copy | PASS — verbatim: *"Sign-in with Google didn't work. If you were invited, open the link in your invitation email to finish setting up your account."* |

### What the UAT caught, and the commit that fixes it

The Google app card showed **no redirect URIs at all**. The first draft published a URI only when its flow was already composed — which hides both of them on exactly the deployment where they are needed, since an operator creating their first OAuth client has no client id yet. That is the chicken-and-egg the section exists to solve.

Fixed in `ed4d7378d`: a URI is published whenever the deployment knows the origin that purpose would use, because that is a property of the installation rather than of credentials nobody has entered yet.

The fitness test earned its place immediately — the first attempt at that fix published a sign-in URI missing the `/v1` the composed routes append, and the test failed with both strings side by side. That rule now has one spelling used for both wiring and display.

### Verified directly against the running API

```
invite                      -> "status":"invited"
assignee roster             -> absent      (an invited member is not assignable)
admin roster                -> present     (an admin can still revoke the invitation)
password login pre-redeem   -> 401
admin mints a link          -> 201         (the expired-invitation recovery path)
redeem                      -> 204 -> status "active"
login post-redeem           -> 200
assignee roster             -> present
event_outbox                -> user.activated, with its linked audit_log_id
/auth/capabilities          -> {"oidc_providers":[],"password":true}
```

The last line is the short-circuit working: no provider composed means the policy seam is never wired, so the anonymous endpoint touches no database.

### Found but NOT fixed here

- **#3296** — first-run setup demands a Google OAuth client with no skip, so an operator without one cannot finish onboarding through the UI at all. Pre-existing and a different surface; this branch makes it more visible rather than causing it.
- Navigating to `#/login?oidc=failed` from an already-loaded SPA does not render the notice; only a full page load does. A real OAuth redirect **is** a full load, so the shipped path works.

### One expectation that was wrong, rather than the product

Redemption does not sign you in — it says *"Password updated… sign in with the new password"* and revokes every session. That is correct for a credential change, so the expectation was corrected, not the code.

tinnhlt-gra:
Recording of the UAT run described in the comment above.

uat

tinnhlt-gra:
## Review findings — all applied, one pushed back

Thanks both. Two of these were genuine P1s in code I had already convinced myself was right.

### P1: the agent seat could be sent to `invited`
It carries a NULL `password_hash` **by construction** — it holds no credential because it does not sign in — so my passwordless rule in `ReactivateUser` caught it, and extension dispatch would then fail to find the seat it requires, on a row nobody can redeem. Excluded now (`AND NOT is_agent`), with `TestReactivatingTheAgentSeatRestoresItToActive`.

### P1: the connector redirect URI named a route that does not exist
The Gmail/Calendar connectors are served under `gmail` and `gcal`, never `google` — an operator would have registered a URL nothing answers. Worse, my own fitness test compared the builder against **itself**, a tautology that passed while the value was wrong. It now asserts against the path template each provider is actually served on, and I verified it fails on the original bug. This also surfaced that one Google app backs **three** callbacks, so the contract gained `calendar_connect`.

### Security review — no authz/tenant/injection defect, two real gaps
- **Unthrottled DB work on the anonymous `/auth/capabilities`.** Once the policy is wired it takes a pool connection, a transaction and a row per request. My own diff had moved the limiter *before* the identical read on the OIDC routes for exactly this reason and left this one unbounded. Now budgeted — and exceeding it withholds the provider buttons, not the page: password still renders, so a flood costs the attacker the buttons and the installation nothing.
- **A failed policy read leaked internals with the wrong status.** The wrapped error reached `httperr`, so an unregistered settings key would answer an anonymous caller **404 carrying the setting's name** — the status my own comment three lines above forbids. Logged server-side now, bare error written.

Also: the provider list is bounded in the contract **and** the entry validator (the entry binds the non-HTTP writer, which the contract cannot reach); the policy survives a data reset, because absent means *all providers* and a wipe would silently re-open a sign-in method an admin closed; and a comment claiming the invitation link survives deactivation was backwards — deactivation consumes every unused token.

### Craft review — six, all applied
The two that mattered: `WithKeyvault` replaced the whole `googleAppHandlers` struct, silently zeroing the redirect URIs and env client id on any option order but the current one, with nothing holding that order; and the intersection rule was written twice, so the settings screen and the login screen could disagree with no test failing. Both now share one resolver. Plus a wire-visible array built from map iteration (rows swapped between boots), a mid-flight disabled provider answering JSON to a **browser navigation**, and two comments that overclaimed what their tests hold.

### Pushed back: the lock-order deadlock
> `RedeemPasswordReset` already locks `auth_token` before this query locks `app_user`. `supersedeSetPasswordTokens` locks `app_user` first…

The premise is not the case. `supersedeSetPasswordTokens` takes an **advisory** lock, not `FOR UPDATE` on `app_user`, and `lockMemberForTokenIssue`'s doc comment (`userpasswordlink.go:131-140`) says so explicitly — deadlock avoidance is the stated reason it is advisory, and the member is read "WITHOUT a row lock". My change preserves the existing token→app_user order, so it adds no new cycle. The security review reached the same conclusion independently. No change made.

### Recorded, not done
Extracting `activateIfInvited` from `RedeemPasswordReset` — a readability judgment call rather than a defect, and not worth restructuring a security-sensitive transaction at this stage.

One consequence worth flagging: the fixes pushed `ssologin.go` past the 500-line cap, so the account resolve/link half moved to `federatedidentity.go` — a cut that file's own header already described as two concerns.

PR #3326, opened 2026-08-31
feat(capture): a contact capture mints belongs to the mailbox owner until something judges its sender

LarsGradion:
Connecting a mailbox with a year of history names every correspondent in it. On the founder's own Gmail that was 184 person records — a lawyer, a travel agent, a clinic, 19 gmail.com addresses — every one of them visible to every colleague on the strength of one email.

**A person the capture sink mints now belongs to the mailbox owner until something judges their sender a business counterparty.** The verdict path is that something, and it is the only ensure caller that asks for the workspace.

## The change

`EnsureCounterpartyInput` gains `OwnerScoped`, and the two callers disagree about it:

- the **capture sink** (`compose/captureensurer.go`) asks for owner scope — it runs while the sender is still unjudged;
- the **verdict path** (`compose/captureverdict.go`) does not — something judged the sender, and that is the moment the record becomes the workspace's.

`false` is the default, so a caller that says nothing gets the behaviour that existed before this field. The wrong direction to fail in is silently narrowing a record somebody expected to see.

`promoteIfWorkspaceScoped` (in the new `people/capturescope.go`) is what moves a record when the verdict arrives. It runs on the exact-match branch of `ensurePerson`, before the quarantine return — an impersonation tell is a reason to learn nothing NEW from a message and not a reason to withhold a promotion something already decided.

## One direction, and what enforces it

The UPDATE writes `owner → workspace` and has no spelling that goes the other way. That matters because the sink runs on **every** message a sender ever writes: if this could narrow, a contact somebody promoted would be un-published the next time that person wrote.

The `ownerScoped` early return above it is an optimisation — it saves a statement per captured message, which is worth having on the sink's path — and the comment says so rather than implying it is the guard. I found that out by mutating it away and watching both tests still pass; the SQL is what holds the invariant, and one clause enforcing it is better than two that can drift.

The visibility pin is also the concurrency guard, checked through `RowsAffected`: a second ensure over the same person matches nothing and writes nothing.

## What this does not change

A person a **human** creates is still the workspace's from the moment they exist. Somebody typed them in, and that is the decision — there is nothing left to judge, and withholding such a record would be the product arguing with its author.

Nor does it change the audience of the correspondence: that is a property of the activity (`platform/auth ActivityContentClause`) and was settled in the four PRs before this one.

## Tests

- `TestAPersonTheSinkMintsBelongsToTheMailboxOwnerAlone` — the contact is `owner`, owned by the capturing seat.
- `TestAVerdictPromotesTheContactItJudged` — a workspace-scoped ensure promotes it, and a **later capture does not narrow it back**.

Both were mutation-checked. The first failed against unchanged code before the fix landed, which is the check for free.

`TestCaptureAutoCreatesTheCounterpartyBehindAThread` asserted the old behaviour and now asserts the new one, pointing at the promotion test for the other half. The file-level doc comment on `ensure.go` claimed connector rows are workspace-visible; this PR makes that false, so it says what is true instead.

`ensure.go` crossed the 500-line ceiling, so the promotion moved to `capturescope.go` — named to distinguish it from the `promote.go` beside it, which converts a lead into a person rather than widening who may read one.

## Still to come in this sequence

The plan's PR 5 also covers the tier-ladder changes — freemail and machine-localpart senders deferring rather than minting, the transactional registry suppressing even when corresponded, and the `personal`/`advisor` verdict kinds. Those are a separate concern from who may SEE a minted record, and they land next rather than being bundled here.

`make check` green (lint verified directly: 0 issues). Integration lanes: `backend/internal/compose/integration/capture`, `backend/internal/compose`. `make craft-static` clean across backend, extensions, fixtures and desktop.

---
## Summary by cubic
Contacts the capture sink mints now belong to the mailbox owner alone until a verdict judges their sender a business counterparty. Connecting a mailbox with a year of history no longer puts every correspondent — a lawyer, a clinic, 19 gmail.com addresses — in front of every colleague on the strength of one email.

- `EnsureCounterpartyInput` gains `OwnerScoped`; the capture sink passes it, the verdict path does not.
- A workspace-scoped ensure over an owner-scoped contact promotes it, before the quarantine return, via `promoteIfWorkspaceScoped`.
- The promotion is one-directional and pinned on the row's visibility in SQL, so a later sink capture never narrows a promoted contact back and a second ensure over the same person writes nothing.
- Persons a human creates stay workspace-visible from the start; `OwnerScoped` defaults to false, so callers that say nothing get the old behavior.

<sup>Written for commit c21321d4485e740ddd6dbb9b6ce681105436df26. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/3326?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Newly captured contacts remain private to their owner until they are confirmed.
  - Confirmed contacts can be promoted to workspace visibility automatically.
  - Repeated captures preserve broader visibility and do not make confirmed contacts private again.
  - Archived contacts and repeated promotions are handled safely without unintended changes.

- **Bug Fixes**
  - Corrected connector-created contact visibility so unconfirmed identities are not exposed to the workspace prematurely.

### DIFF
--- a/backend/gates/writeauthorityreach_test.go
+++ b/backend/gates/writeauthorityreach_test.go
@@ -53,10 +53,11 @@ var writesWithoutARowProbe = gatekit.Waive(map[string]string{
 	// leave every other subject's record untouched — for retention that is the
 	// storage-limitation failure the pass exists to prevent, and for the rest it
 	// would make a hygiene pass silently partial.
-	"internal/modules/privacy:anonymizePerson": "the retention engine anonymizes a person in place because their retention clock expired, under a policy that names the table. The subject is chosen by age; there is no caller asking for this row and no authority to test beyond the policy itself",
-	"internal/modules/privacy:anonymizeLead":   "the lead half of the same retention pass, chosen the same way",
-	"internal/modules/privacy:archiveDeal":     "the deal half of the same retention pass: a deal past its window is archived because of when it closed, not because somebody asked",
-	"internal/modules/deals:SweepWorkspace":    "the close-date hygiene pass reads every open deal in the workspace and corrects or stages the ones whose expected close has gone stale. It runs as the workspace's own system principal with no human behind it, and narrowing it to one seat's rows would leave every other rep's forecast uncorrected while the sweep reported success",
+	"internal/modules/privacy:anonymizePerson":         "the retention engine anonymizes a person in place because their retention clock expired, under a policy that names the table. The subject is chosen by age; there is no caller asking for this row and no authority to test beyond the policy itself",
+	"internal/modules/privacy:anonymizeLead":           "the lead half of the same retention pass, chosen the same way",
+	"internal/modules/privacy:archiveDeal":             "the deal half of the same retention pass: a deal past its window is archived because of when it closed, not because somebody asked",
+	"internal/modules/people:promoteIfWorkspaceScoped": "a row probe here would refuse the write it exists to make. The row is owner-scoped, which platform/auth reads as capture privacy — invisible to every principal but its owner, an admin included — so EnsureWritable answers no for exactly the rows this promotes. What decides the promotion is not the caller's authority over the row but the verdict that judged its sender a business counterparty, and the surrounding ensurePerson already takes auth.Require(person, create) for the caller who may mint one at all. The write is one-directional and idempotent: it moves owner to workspace and never back",
+	"internal/modules/deals:SweepWorkspace":            "the close-date hygiene pass reads every open deal in the workspace and corrects or stages the ones whose expected close has gone stale. It runs as the workspace's own system principal with no human behind it, and narrowing it to one seat's rows would leave every other rep's forecast uncorrected while the sweep reported success",
 
 	// The shared fill, probed by BOTH of its callers rather than by itself.
 	// ApplySitePersonFields takes EnsureWritableLive and SKIPS an out-of-scope
--- a/backend/internal/compose/captureensurer.go
+++ b/backend/internal/compose/captureensurer.go
@@ -72,6 +72,10 @@ func (p peopleEnsurer) EnsureCounterparty(ctx context.Context, in capture.Ensure
 		Source:      in.Source,
 		CapturedBy:  in.CapturedBy,
 		SuppressOrg: in.SuppressOrg,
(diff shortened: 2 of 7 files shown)


LarsGradion:
This merged as `3c0e1b146` with the subject `wip: capture ladder` and an empty body. That is my mistake: I amended onto a `wip:` placeholder commit while separating two branches and never corrected the subject before merging, and the squash took it verbatim.

`main` cannot be rewritten, so the commit stays as it is. Recording here what its message should have said, because `git log` is where the next reader looks and this PR is now the only place the reasoning lives:

---

**feat(capture): a captured contact belongs to the mailbox owner until something judges its sender**

Connecting a mailbox with a year of history names every correspondent in it — a lawyer, a doctor, a school. One email is not a reason to put any of them in front of every colleague.

`EnsureCounterpartyInput.OwnerScoped` is what the two ensure callers disagree about. The capture sink asks for owner scope: it runs while the sender is still unjudged. The verdict path does not, because something judged the sender a business counterparty, and that is the moment the record becomes the workspace's.

`promoteIfWorkspaceScoped` moves a record when the verdict arrives. One direction only: the UPDATE writes `owner → workspace` and has no spelling that goes the other way, which matters because the sink runs on every message a sender ever writes — if it could narrow, a promoted contact would be un-published the next time that person wrote. The `ownerScoped` early return above it is an optimisation on the sink's hot path, not the guard; the SQL is the guard.

A person a human creates is still the workspace's from the moment they exist.

---

Two things worth knowing for whoever picks up the next PR in this sequence:

1. The tier-ladder half of the plan's PR 5 (freemail deferral, machine localparts, the `personal`/`advisor` verdict kinds) is NOT in this change. It is a separate concern from who may see a minted record, and it lands next.
2. `ensure.go`'s file-level doc comment used to say connector-created rows are workspace-visible. This change made that false, and the comment now says what is true. If you are reading an older copy, it is wrong.

PR #3335, opened 2026-08-31
A sender the mailbox owner writes to is not always a person

LarsGradion:
Part of the mailbox-privacy work. PR 3 gave each importing mailbox its own provenance row, PR 4 made a mailbox say what it asks of its mail, PR 5 made a captured contact the owner's until judged. This one changes **which senders become records at all**.

## The problem

The tier ladder's T1 evidence is that the workspace WROTE to an address. That is honest evidence of intent and says nothing about whether a person is behind the address. A mailbox owner books flights, files expenses and answers their own notification robots — so a real import minted `Receipts` and `Plans` as contacts, and put nineteen private correspondents of one mailbox in front of every colleague.

## What changed

**T1 asks a second question.** `Sink.recordWorthy` refuses two shapes, both meaning "no person is reachable here" rather than "this person is uninteresting": a machine local part wherever it sends from, and a domain that exists to talk to its own users. It never yields to correspondence, because correspondence with an address nobody answers is not correspondence.

The operator's `transactional_never` allowlist still wins, through the same `TransactionalList` the registry consults — a declaration that a domain is legitimate has to cover both refusals or it is not an escape hatch. The local part is asked **before** the allowlist: vouching for a domain says its people are real, not that `noreply@` on it answers.

**The personal-service domains land in their own list**, not in `transactionalBaseline`. That baseline is read by `IsMachineAddress`, which two surfaces now use to drop rows — the attention queue, and (as of #3330) the Worklist's "machine sender" group. These are companies with salespeople; putting them there would file a named human waiting on a reply into a pile refused without thought, which that reader cannot recover from because nothing on the page says somebody was hidden.

**T3 defers instead of minting.** A consumer mailbox settles the ORGANIZATION question by itself and settles nothing about the person: a customer writing from their private address and a founder's sister arrive in the same shape. Free mail now goes to the verdict. That is safe for the company question because the refusal has two writers — the verdict's own create path reaches `deferOrgToTriage`, which asks the same consumer-mail question at the chokepoint every writer passes. `TestAVerdictOnAFreeMailSenderMintsThePersonAndNoCompany` is what says so; without it the two writers were one comment apart from disagreeing.

**Two sender kinds the taxonomy had no word for.** Every family letter and every message from a founder's lawyer previously had to be answered with one of six business kinds, and the engine picked one because the vocabulary is closed.

- `personal` makes no record. The mail is left alone: destroying it belongs to the purge behind its undo window. `hideNoise` is deliberately not called — its scope excludes every address the workspace has written to, which is every address this kind is about, so calling it would be a no-op that read like a hide.
- `advisor` makes the record and keeps it owner-scoped. A founder's lawyer is a genuine contact, and publishing them to the workspace announces that the founder has one.

**The vocabulary is spelled once.** `verdictKindNames()` feeds the response schema, the validator's message and two test assertions that each held their own copy.

**`capture_counterparty_verdict` drops to a local-only ladder.** Its prompt carries subject and body, and it is the task that decides whether a sender is personal — a cloud rung would send a founder's family mail off the machine in order to ask whether it was family mail.

## A defect this PR's own review caught

The first draft of the two new kinds **did nothing at all**. `verdictSchema()` still listed only the six original kinds, and on a grammar-constrained local rung that enum is what the model can physically emit — so neither kind was reachable. `apply()` also had no effect arm for them, which would have errored the transaction the moment anyone fixed the schema, burning the retries and retiring the row into the human review queue: exactly the sender the kind exists to withhold. Every test passed and `make check` was green.

The comment above that switch said "exhaustive by construction". That claim is what stopped anyone checking.

Two gates now hold it, both mutation-checked against the real defect: `TestEveryVerdictKindHasAnEffect` reads `apply()`'s switch and asserts an arm per kind, and `TestTheModelMayAnswerEveryKindTheTaxonomyDefines` asserts the schema admits every kind the taxonomy defines.

## Other fixes from review

- The down migration mapped `advisor` to `person`. Both are status `real`, but the sink reads `real AND kind <> 'person'` as "decided, mint nothing" — the rewrite would have flipped every advisor row to "known counterparty", so the next message from a founder's lawyer would mint and publish the contact on rollback. Now `role_mailbox`, and the migration says the rollback is lossy instead of calling it honest.
- A suppressed decision carries `create=false` explicitly rather than relying on `registrySuppresses` only ever suppressing an address T1 left alone. That holds today and is easy to make false from the other file; two readers act on the flag.
- `TestBackfillCountsOnlyTheCounterpartiesItsOwnPagesCreated` had lost its free-mail arm, so its org-count assertion passed for a trivial reason. Restored as the discriminator.
- Two certification fixtures carried real names; replaced with invented ones on `.example` domains.

## Not taken

An earlier draft gave `Suppress` a `SuppressionClass` so correspondence would spare only the prefix guess. Mutation testing showed `recordWorthy` already refuses every address that change would have caught — `IsMachineAddress` tests the infra domain as well as the local part — so shipping both would have been two answers to one question. Reverted; the registry keeps its boolean answer.

## Tests

Unit: `TestRecordWorthyRefusesAnAddressNoPersonAnswers`, `TestAnOperatorsAllowlistOutranksTheRecordWorthyRefusal`, `TestTheAttentionQueueStillSeesHumansAtPersonalServiceCompanies`, `TestEveryVerdictKindHasAnEffect`, `TestTheModelMayAnswerEveryKindTheTaxonomyDefines`, `TestTheTaxonomyDefinesTheKindsCaptureDeclares`.

Integration: `TestAnAdvisorVerdictMakesTheRecordAndKeepsItTheOwnersAlone`, `TestAPersonalVerdictMakesNoRecordAtAll`, `TestAVerdictOnAFreeMailSenderMintsThePersonAndNoCompany`, plus two new subtests on `TestCaptureTierGateLetsCorrespondencePrecedeSuppression` for the senders correspondence must **not** spare.

Every guard above was mutation-checked: the guard reverted, the test observed to fail, the guard restored.

## Found along the way

`walkAtOnce` in the backfill-yield harness never called `AttestSentByOwner`, so the parallel walk did not match the sequential one. Invisible while those fixtures used free mail; fixed here.

## Verification

`make check` green on the rebased tree, `make craft-static` clean, the full capture integration package green, the head catalog regenerated from migrations, and the down migration dry-run against a database holding rows with the new kinds.

---
## Summary by cubic
Fixes capture minting contacts for senders nobody answers, and keeps a mailbox owner's private and professional correspondence from being published to the workspace.

**What changed**

- T1 creates a counterparty only when `Sink.recordWorthy` also accepts the address: machine local parts (`noreply@`, `receipts@`) and registrable mail-infrastructure domains are refused, and correspondence never spares them.
- The operator's `transactional_never` allowlist still wins, but it vouches for a domain, not for `noreply@` on it.
- Personal-service product domains (`expensify.com`, `tripit.com`, `docusign.com`, …) live in their own list, not `transactionalBaseline`, so the attention queue's machine-sender drop still shows named humans there.
- Free-mail senders now go to the verdict ledger instead of minting a person on sight, and extensions share that deferral through the same sink; the verdict's own create path asks the same consumer-mail question, so gmail.com can never become a company.
- Two new verdict kinds: `personal` makes no record and leaves the mail alone, and `advisor` makes the record but keeps it owner-scoped.
- The kind vocabulary is spelled once in `verdictKindNames()`; two mutation-checked tests hold the effect switch and the response schema to that same list, closing the gap that let the new kinds reach the prompt while the schema still refused them.

**Migration and rollout**

- The migration `1788170900_a_sender_can_be_the_owners_alone` adds `personal` and `advisor` to the kind CHECK.
- The down migration rewrites `advisor` rows to `role_mailbox` (not `person`, which would re-mint and publish on the next message) and `personal` to `spam`; the rollback is deliberately lossy.
- `capture_counterparty_verdict` is now local-only, because its prompt carries the message body and a cloud rung would ship family mail off the machine to ask whether it is family mail.

<sup>Written for commit db70a081ca851ca8035c062cf5a14f1576b2ed35. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/3335?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added **Personal** and **Advisor** correspondent classifications.
  * Advisor messages can create private, owner-only contacts while preserving correspondence history.
  * Personal messages remain in activity history without creating contacts.
  * Added worklist suggestions for drafting replies to specific activities.
  * Free-mail and service addresses are deferred or excluded from contact creation when appropriate.

* **Privacy**
  * Personal correspondence classification runs exclusively on-device.

* **Bug Fixes**
  * Prevented infrastructure and notification addresses from appearing as contacts.
  * Improved suppression and pending-review behavior for uncertain senders.

### DIFF
--- a/backend/api/ai-tasks.yaml
+++ b/backend/api/ai-tasks.yaml
@@ -85,14 +85,14 @@ tasks:
     no_payload: true
 
   capture_counterparty_verdict:
-    ladder: [local_small, cheap_cloud]
+    ladder: [local_small]
     execution_mode: background
     on_budget_exhausted: queue
     status: shipped
     sites: [verdict]
     no_payload: true
     company_context: none
-    doc: "ADR-0072/A118: the ambiguous first-time-sender creation gate — real|noise|unsure per address, floor 0.7 (below → unsure, never noise). Release-blocking false-noise eval threshold. The no-payload posture is the `no_payload` field above (ADR-0074); it was formerly asserted in this prose and matched by substring."
+    doc: "ADR-0072/A118: the ambiguous first-time-sender creation gate — real|noise|unsure per address, floor 0.7 (below → unsure, never noise). Release-blocking false-noise eval threshold. The no-payload posture is the `no_payload` field above (ADR-0074); it was formerly asserted in this prose and matched by substring. Local-only ladder: the prompt carries subject and body, and this is the task that decides whether a sender is personal — a cloud rung would send a founder's family mail off the machine to ask whether it was family mail."
 
   signal_extract:
     ladder: [cheap_cloud, premium]
--- a/backend/api/crm.yaml
+++ b/backend/api/crm.yaml
@@ -17692,7 +17692,7 @@ components:
           enum: [pending, unsure, real, noise, rejected, suppressed]
         kind:
           type: [string, 'null']
-          description: 'Who wrote, when the verdict said: person | role_mailbox | organization_sender | newsletter | transactional | spam.'
+          description: 'Who wrote, when the verdict said: person | role_mailbox | organization_sender | newsletter | transactional | spam | personal | advisor. The last two belong to the mailbox owner rather than to the business: personal is a private correspondent, advisor a professional they engage personally.'
         resolved_at: { type: [string, 'null'], format: date-time }
 
     LicenseEntitlement:
--- a/backend/gates/uniquenessclaims_test.go
+++ b/backend/gates/uniquenessclaims_test.go
@@ -462,7 +462,7 @@ func TestTheRegisterHoldsNoEntryThatIsNoLongerAClaim(t *testing.T) {
 var shapeCensus = map[string]int{
 	"cannot-drift":   156,
 	"once":           155,
-	"one-of-a-kind":  155,
+	"one-of-a-kind":  154,
 	"is-every-named": 92,
 	"only-noun":      16,
 	"no-second":      10,
--- a/backend/internal/compose/aicert/corpus/capture_counterparty_verdict/advisor_01.yaml
+++ b/backend/internal/compose/aicert/corpus/capture_counterparty_verdict/advisor_01.yaml
@@ -1,0 +1,38 @@
+name: the_owners_own_lawyer_is_not_the_workspaces_contact
+task: capture_counterparty_verdict
+site: verdict
+source: hand_authored
+sanitized_by: hand_authored/claude-opus-5
+fixture:
+  display_name: Dr. Katrin Bauer
+  email: k.bauer@kanzlei-bauer.example
+  subject: Entwurf Gesellschaftervereinbarung - vertraulich
+  body: |
+    Sehr geehrter Herr Weidner,
+
+    anbei der überarbeitete Entwurf der Gesellschaftervereinbarung nebst meiner
+    Anmerkungen zu Ziffer 7 (Vesting). Bitte prüfen Sie insbesondere die
+    Regelung zum Bad-Leaver-Fall, bevor wir sie an die Gegenseite geben.
+
+    Mit freundlichen Grüßen
+    Dr. Katrin Bauer
+    Rechtsanwältin
+expect:
+  outcome: accepted
+  answer: advisor
+  rubric: >
+    A lawyer writing to the mailbox owner about the owner's own shareholding.
+    This is real correspondence with a real named professional, which is what
+    makes it the hard case: every surface signal says "person". The distinction
+    is whose business it is. The engagement is the owner's personally, and the
+    subject matter — a shareholder agreement, vesting, a bad-leaver clause — is
+    exactly what a colleague must not be handed by an import. "advisor" scores
+    high. "person" is the failure this scenario exists to catch: it is not a
+    misread of the sender, it is a disclosure, announcing to the workspace that
+    the founder has a lawyer and what they are negotiating. It scores zero above
+    the floor. "personal" is wrong but safe — it withholds rather than
+    publishes — and scores in the middle.
+  bands:
+    certified_min: 70
+    degraded_min: 50
+    floor: 40
--- a/backend/internal/compose/aicert/corpus/capture_counterparty_verdict/personal_01.yaml
+++ b/backend/internal/compose/aicert/corpus/capture_counterparty_verdict/personal_01.yaml
@@ -1,0 +1,33 @@
+name: a_private_correspondent_is_not_a_business_contact
+task: capture_counterparty_verdict
+site: verdict
+source: hand_authored
+sanitized_by: hand_authored/claude-opus-5
+fixture:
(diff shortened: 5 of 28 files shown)


LarsGradion:
A background security review flagged authorization in `captureverdictcreate.go`. I checked it and it does not hold — recording why, since the file is new in this PR and the question is a fair one to ask of it.

That file is a pure code move. `captureverdict.go` had grown past the 500-line ceiling once the advisor effect landed, so the two creation effects were extracted into their own file. Diffing the split for anything auth-shaped shows only the `OwnerScoped` plumbing added — no `auth.Require`, `EnsureVisible` or principal line was added, removed or moved.

The gating is where it always was: `people.EnsureCounterpartyTx` calls `auth.Require(ctx, entityPerson, principal.ActionCreate)` at its entry point, and both creation paths go through it.

The substantive question is whether the engine can scope a record to an arbitrary user. It cannot. `OwnerID` reaches `createOwnerScopedCounterparty` from the ledger row, which `derivationStart` populated from `capturePrincipal(ctx)` — the authenticated connector principal's `OnBehalfOf`. Nothing a sender writes reaches it, so an advisor record is scoped to the seat whose mailbox actually delivered the message.

LarsGradion:
Second look at the authorization question in the verdict engine, this time at the `advisor` arm specifically. The conclusion is unchanged but my earlier reasoning was incomplete, so recording the sharper version.

I said the path is protected by `people.EnsureCounterpartyTx` calling `auth.Require(ctx, entityPerson, principal.ActionCreate)`. That gate is present, but it is **not** what protects this path: `apply()` runs under `PrincipalSystem` (`workspaceCtx`, captureverdict.go:119), and `auth/rbac.go:68` gives that principal an unconditional pass. Object RBAC is bypassed here by design — a background sweep has no human grant to check.

What actually confines the record is the owner on the ledger row. `createOwnerScopedCounterparty` passes `row.OwnerID` through to `EnsureCounterpartyInput.OwnerID` alongside `OwnerScoped: true`, and that column was written by `recordDisposition` from `in.OwnerID`, which `derivationStart` took from `capturePrincipal(ctx)` — the authenticated connector principal's `OnBehalfOf`. Nothing a sender writes reaches it.

So an advisor verdict scopes the record to the seat whose mailbox actually delivered the message, and the confinement comes from provenance rather than from an RBAC check. `TestAnAdvisorVerdictMakesTheRecordAndKeepsItTheOwnersAlone` asserts the visibility, and I mutation-checked it by flipping `OwnerScoped` to false — the test fails with `an advisor's record is \"workspace\", want owner`.

PR #3377, opened 2026-08-31
An Outlook calendar is a connection of its own

dungbanguyen:
## Why

Google has always brought a pair — a mailbox and a calendar, each with its own consent, its own scope and its own disconnect. Microsoft brought only the mailbox, and `capture_connection_provider_check` admitted nothing calendar-shaped, so an Outlook calendar could not be stored at all.

Meanwhile the connect dialog's own copy said, in all three locales, that Microsoft brings *"mail **and calendar** on a Microsoft work account"*. It did not. After this it does.

## What

`graphcal` — a standing Outlook-calendar connection over Graph's `calendarView` delta, mirroring `gcal`.

A **second connection** rather than a widened `graph` one, mirroring gmail/gcal. One row would mean one consent carrying both permissions, and then a person who wants their calendar in the CRM but not their mail has no way to say so — and disconnecting either would take the other with it. Same Entra app, `Calendars.Read` alone.

### The meeting rules moved rather than being copied

`gcal/event.go` carried the classification (all-internal skip, cancelled skip, booked rooms are not guests) and the activity mapping, with a comment saying it stayed there *"because gcal is the only calendar connector today (ADR-0054 §3: flat by default; grow a subpackage only when a second concrete caller appears)"*. This is that second caller.

They are now `capture/meetingmap`. Whether a booked room counts as a guest, whether a meeting with no outside party is worth logging, and which addresses the writer judges against the workspace's own domains are answers about **meetings**, not about Google — and the way two copies fail is that a workspace's own colleagues quietly start reading as customer touches on one vendor and not the other. Each connector keeps only its vendor's decode. gcal's existing fixtures still run end-to-end through it.

### The window is the sync

90 days back, a year forward, sliding. That is what the standing connection *watches* rather than a first pull's bound — Graph's `calendarView` delta tracks the range it was opened on. It is why there is no `Backfiller` (the backwards half already is one), and it is stated in the docs rather than left for somebody to discover from an absent meeting.

## Two pre-existing defects found on the way, fixed here

1. **The participant replay pass never knew `graph`.** It handled `gmail`/`imap`/`gcal`, so every Outlook message was recorded `unreadable` rather than re-read. Graph hands over the RFC822 original, so `mailmap` already served it — only the case arm was missing. Both Microsoft sources are now wired.
2. **The connect transport's provider guard was a `!=` chain** that had to be widened in step with the refusal *message* beside it, with nothing making them widen together — a provider admitted by the guard but missing from the message is a refusal that names the wrong set. Both now read one list (`connectorproviders.go`), which also brought `connectors.go` back under the 500-line ceiling.

## Migration

`1788170959_an_outlook_calendar_is_its_own_connection` widens the provider CHECK — additive, every existing row already satisfies it. `head_catalog.txt` updated in the same commit.

## Verification

`make check-backend`, `make check-fe`, `make craft-static` green, plus the real-Postgres lanes `backend/migrations` (which proves the migration against the catalog) and `backend/internal/compose/integration/capture`.

## Summary by CodeRabbit

* **New Features**
  * Added Outlook Calendar as a separate connection from Outlook mail.
  * Supports calendar synchronization, incremental updates, meeting details, attendees, rooms, cancellations, and time zones.
  * Added organization intro-request drafts without sending messages or changing records.
  * Added buying-team role suggestion and confirmation messaging.
* **Improvements**
  * Updated connector names, descriptions, OAuth guidance, and localized text.
* **Documentation**
  * Documented Outlook Calendar setup, permissions, sync windows, and connection behavior.


dungbanguyen:
Review round addressed — resolving the threads with what changed, since three were real behaviour.

**The window did not move (P2, and the sharpest catch).** Graph's `calendarView` delta reports only events starting inside the range it was opened against, and a valid `deltaLink` is resumed forever — so a meeting booked past the forward edge was never reported and nothing said so. The comment claimed the window "slides on every re-anchor" while steady state performed no re-anchor at all. The cursor now dates its window and the connector reopens it every 30 days; `TestAStaleWindowIsReopenedEvenThoughTheDeltaStillWorks` covers fresh, stale, and the incremental round that must *not* redate the anchor.

**Time zones.** Better fix than mine, taken: the request now asks for UTC outright (`Prefer: outlook.timezone="UTC"`) rather than relying on the mailbox's own zone and hoping Go can resolve the name. The zero-time fallback stays for a zone that still will not resolve; it is no longer the common path.

**Unknown is not internal — for the organizer too (P2).** An organizer address with no parseable domain was read as internal and could drop an otherwise multi-party meeting, while the same shape in an *attendee* already counted as external. One rule now, and `meetingmap` has its own tests rather than being proved only through each vendor's decode.

**The rollback deleted connections (P1).** It removed every `graphcal` row to make room for the narrower CHECK, which strands each row's vault secret — nothing able to read it and nothing able to destroy it, the one custody guarantee disconnect exists to keep. It refuses now and says to disconnect first.

**The continuation check (P2).** A link that merely shared the base string was accepted and fetched with the bearer token. It must now be the base, or the base followed by a path or a query.

**Duplication.** `capture/graphconn` is the Microsoft twin of `capture/googleconn` — the credential bundle, the connect payload and the scope rendering are the same handshake, and two copies drift on `Granted`, which is exactly what a refresh must ask for.

**Not changing: the `NOT VALID` split.** `capture_connection` holds one row per connected mailbox, so the validation scan is not the cost here — the `ACCESS EXCLUSIVE` lock is taken by the `DROP`/`ADD` either way, `lock_timeout` already bounds the wait, and a second migration to maintain forever would buy nothing measurable at this table's size.

Plus the copy: the test comment's provider count, the German strings that had switched to the formal address beside informal neighbours, a connect guard that said "mailbox" while refusing a calendar, and doc lines that undercounted the connectors or described the old sync.

PR #3384, opened 2026-08-31
Margince can send from an Outlook mailbox too

dungbanguyen:
## Why

Gmail has been the only connection Margince could send from — and the connect dialog said exactly that, in three locales. An installation on Microsoft 365 could bring its mail in and then had to leave the product to answer any of it.

## What

`graph` now implements `connector.EmailSender` and `connector.AttachmentCarrier`, on `Mail.Send` riding the **same consent** as the read permissions. Microsoft, like Google, will not add a permission to an existing refresh token, so asking later would mean a second connection for the same mailbox. **A mailbox connected before this lands captures normally and refuses every send by name until it is reconnected** — cheap now, expensive after go-live, which is why it is this change and not a later one.

### The bytes are the shared renderer, not a second one

`gmail/rfc822.go` moved to `capture/mailwire`. Microsoft's `sendMail` accepts the same complete RFC822 message Gmail's `messages.send` does, and what a `multipart/alternative` puts first, whether an attachment's filename rides in one header or two, and where base64 folds each have **one** right answer. Two renderers would answer them twice and diverge exactly where one vendor's recipients get a blank message or a truncated PDF.

### The retry guard is better here than on Gmail, and the difference is worth stating

Gmail discards a client-supplied Message-ID and files the message under one of its own, which leaves its prior-send lookup searching for an identity that was never filed — `gmail/send.go` says so plainly. Graph exposes the identity as the filterable `internetMessageId` property, so the lookup reads the message's **own field** rather than a full-text index.

It still does not *close* the window: Exchange Online may rewrite the identity on submission depending on tenant configuration, in which case the filter matches nothing and a retry mails twice. That is the same failure Gmail has always had, but conditional rather than certain — and it is written down rather than left to be discovered. Mutation-tested: remove the `Attempt > 0` guard and the test fails with *"the recipient was mailed a second time"*.

### Microsoft carries less than Gmail, and `Carriage` says the real number

Graph's MIME submit ceiling is 4 MB of base64 — about 3 MiB of actual bytes — against Gmail's full 25 MiB inbound cap. Declaring the true number is what makes an over-large message **park with an honest reason** instead of leaving as a request Microsoft answers with an opaque refusal that no retry can get under.

### Drift protection

`comms.SendScopeFor` became a table rather than an `if`, and `compose/sendscope_test.go` now runs **per provider** over the three spellings (what the consent requests, what comms demands, what the connector re-checks) plus a new arm asserting every provider given a send scope actually implements `EmailSender`. A third vendor added to two of the three places fails here rather than parking every delivery with a connector limitation that does not exist. The frontend's send-grant badge (`connector-status.ts`) follows the same shape.

### Copy that had gone false

- *"the only connection Margince can send from"* (Gmail) — no longer true.
- *"Capture only"* (Microsoft) — no longer true.

Both fixed in `en`/`de`/`vi`.

## Verification

`make check-backend`, `make check-fe` (Node 24) and `make craft-static` all green.

> Note: this branch and #3377 both touch `connectors.addGraphBrings`. Whichever merges second takes a one-line conflict.

## Summary by CodeRabbit

* **New Features**
  * Microsoft Graph mailboxes can now send email, including attachments.
  * Safe retries help prevent duplicate deliveries.
  * Connection status identifies missing send permissions for Gmail and Microsoft Graph.
  * Setup guidance and connector descriptions reflect sending across supported OAuth mailboxes.

* **Bug Fixes**
  * Improved handling of failed or incomplete post-send lookups.
  * Standardized email formatting across supported mail providers.
  * Added clearer handling for messages exceeding Microsoft Graph’s size limit.

### DIFF
--- a/backend/gates/outboundanonymity_test.go
+++ b/backend/gates/outboundanonymity_test.go
@@ -78,6 +78,7 @@ var anonymousOutbound = gatekit.Waive(map[string]string{
 	"internal/modules/capture/googleconn/googleconn.go:func Get":    "runs inside an OAuth grant the person made to this app, which names the caller to the provider more precisely than an agent could",
 	"internal/modules/capture/graph/client.go:func GetMIME":         "runs inside an OAuth grant the person made to this app, which names the caller to the provider more precisely than an agent could",
 	"internal/modules/capture/graph/transport.go:func get":          "runs inside an OAuth grant the person made to this app, which names the caller to the provider more precisely than an agent could",
+	"internal/modules/capture/graph/sendclient.go:func SendMIME":    "runs inside an OAuth grant the person made to this app, which names the caller to the provider more precisely than an agent could",
 	"internal/modules/capture/oauthflow/oauthflow.go:func token":    "exchanges a code against a token endpoint using this app's registered client id, which is the identity that endpoint checks",
 	"internal/modules/capture/telegram/api.go:func request":         "calls a bot API under the bot's own token, and the bot IS the identity there",
 	"internal/modules/capture/telegram/sendfiles.go:func SendFiles": "calls a bot API under the bot's own token, and the bot IS the identity there",
--- a/backend/internal/compose/capture.go
+++ b/backend/internal/compose/capture.go
@@ -50,11 +50,23 @@ var gmailScopes = []string{
 	gmail.SendScope,
 }
 
-// graphScopes are the Microsoft identity platform scopes the read-only Graph
-// capture connector requests: mail read + the signed-in user's profile (the
-// mailbox owner lookup) + offline_access (the refresh token). No send, no
-// modify.
-var graphScopes = []string{"offline_access", "User.Read", "Mail.Read"}
+// graphScopes are the Microsoft identity platform permissions a Graph mailbox
+// connection requests: mail read for capture, the signed-in user's profile (the
+// mailbox owner lookup), offline_access (the refresh token), and send for the
+// governed outbound path. No modify, no delete, no shared mailboxes.
+//
+// They ride ONE consent for the same reason Gmail's pair does — Microsoft will
+// not add a permission to an existing refresh token, so asking later would mean
+// a second connection for the same mailbox. A mailbox connected before the send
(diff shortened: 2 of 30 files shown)


dungbanguyen:
Review round addressed — resolving the threads with what changed, since several pointed at the same two defects.

**The scope defect (P1).** Correct and serious. Microsoft carries the scope in the refresh form, and the form was built from the deployment's *configured* list — so adding `Mail.Send` asked every existing read-only grant for a permission it never consented to, and Microsoft refuses the **refresh**. An older mailbox would have stopped *capturing*, not merely declined to send. A refresh now asks for the scopes the grant holds, or omits the parameter (RFC 6749 §6, "the scope originally granted"); the exchange still spells the deployment's consent, which is the one place that is the right answer. `TestAnOlderGrantRefreshesForItsOwnScopesNotTheDeploymentsList` drives sync and the health check through a read-only grant.

**The size arithmetic (P1-adjacent).** Also correct — the payload is base64'd **twice**, once per attachment inside the MIME and once for the whole message on the wire, so a file costs ~1.8× not 1.33×. Per-file cap is 2 MiB, and the whole-message ceiling is now *enforced* rather than described: several files each under the per-file cap still exceed the request limit, and no gate above the connector can see that. It refuses with `ErrFilesNotCarried`, which the dispatcher parks on — a message this size is this size on every retry.

**The asynchronous send.** `sendMail` returns 202 and Exchange files the sent copy afterwards, so the immediate read-back missed almost every time. Removed; the receipt is empty, which `SendReceipt` already reads as "no re-key owed" — true for a MIME submit, where the identity we minted is the one in the body. The retry guard, which runs later, still finds the message.

**Two tests that proved nothing.** The carriage test compared `Carriage()` against the constants that produced it (so it could never catch the numbers becoming Gmail's — it asserts literals now), and the send-scope arm was a second hardcoded list checked against the first. `comms` exports its own census now and the gate derives from it, so a provider added there and nowhere else fails.

Plus the copy: the consent step names `Mail.Send`, and the matrix agrees with its prose on three send-capable connectors.

PR #3391, opened 2026-08-31
main is red: a deliberate qualifier reads as a redundant one

dungbanguyen:
`main`'s lint fails:

```
internal/compose/serveroptions.go:153:5: QF1008: could remove embedded field "captureExclusionHandlers" from selector (staticcheck)
```

## Why not just take the suggestion

The file already answers this, forty lines above the flagged one:

> an ambiguous selector — each call is qualified through its own embedded field instead.

Every assignment in `WithBlobstore` is written through its own embedded field deliberately, so that a second embed growing the same field name never turns one of these into an ambiguous selector — a failure that would surface as a compile error in a file nobody was editing. Dropping the qualifier on this one line because it happens to be unambiguous *today* keeps the linter quiet and loses the property the convention exists for.

So it is suppressed at the line with that reason, in the `//nolint:<linter> // <why>` shape the rest of the tree uses.

## Verification

- `make check-backend` (with a cleaned golangci cache), `make craft-static`

## Summary by CodeRabbit

* **Chores**
  * Added a targeted static analysis suppression without changing runtime behavior.

### DIFF
--- a/backend/internal/compose/serveroptions.go
+++ b/backend/internal/compose/serveroptions.go
@@ -150,6 +150,7 @@ func WithBlobstore(store blobstore.Store) Option {
 		// no objects has no purge, which is honest — destroying the rows and
 		// leaving the files would report mail as gone while its attachments sat
 		// in the bucket.
+		//nolint:staticcheck // QF1008: the embedded field is named on purpose, and the file says why forty lines up — every call here is qualified through its own embedded field so that a second embed growing the same method never turns one of them into an ambiguous selector nobody wrote a test for.
 		s.captureExclusionHandlers.purger = NewCapturePurger(pool,
 			NewRetentionServiceFor(InstallationDB(pool), store, slog.Default()))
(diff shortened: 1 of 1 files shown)


dungbanguyen:
Superseded — somebody fixed this on `main` first, by taking staticcheck's suggestion and writing `s.purger`. `main` is green, so this is redundant and I am closing it rather than reopening a settled line.

One thing worth leaving behind, since it is now only in a closed PR: `WithBlobstore` states a convention forty lines above that assignment — "each call is qualified through its own embedded field instead" — and it exists so that a second embed growing the same field name can never turn one of these into an ambiguous selector. That property is now weaker by one line. It costs nothing today, because `purger` is unambiguous today; it is the kind of thing that is discovered by a compile error in a file nobody was editing. If anyone wants the convention held rather than remembered, the durable answer is a fitness test over `WithBlobstore`, not a comment.

PR #3418, opened 2026-08-31
The mailbox privacy UI, and the projection that dropped the hold reason

LarsGradion:
## What this adds

The mailbox-privacy backend shipped over nine PRs and almost none of it was reachable from the product. This is the owner and admin UI: the posture a mailbox asks for, the organization's opt-in behind `shared`, the Senders page, and the hold reason on the record timeline.

### The mailbox posture

`MailPostureRow` sits under each connected mailbox on Settings → Connections, a `Select` over the three answers. `shared` is drawn and **refused** rather than hidden when the organization has not allowed it: a missing option would tell a reader the product has two postures, a greyed one with the reason beside it says there is a third and who unlocks it. The server refuses it too (422 `shared_posture_not_allowed`), so this is the same rule said early, never the only place it is held.

Narrowing opens a confirm whose checkbox also narrows mail already captured. Widening saves on the spot, because `apply_to_history` only ever narrows.

### The organization's opt-in

A second row on the mail-sharing card, saved by its own button so a reader cannot flip one setting while meaning the other. Turning it ON raises the works-council warning — it is the one capture setting whose default withholds, so the warning belongs to `true`, unlike the sharing switch above it.

### The Senders page

Every address a mailbox brought in, what the classifier concluded, and the seat's own answer where they gave one. The endpoint has shipped since the sender-override work; nothing rendered it, so the decisions were visible only in SQL. Two verbs per row, both from the contract: `business` readmits and is offered only where the classifier did not already admit them; `keep_out` destroys what that sender brought in, so it confirms first and says what survives.

An overruled row is drawn differently on purpose — a filled badge and "you decided", against the quiet dots a machine verdict carries — because the reader auditing this list needs to see which answers are their own.

### The timeline, and a backend defect behind it

A record page seeds its timeline from the person 360 read, and that read assembles its own SELECT instead of going through the shared activity projection. The shared one carries `audience_reason`; this copy did not. So the one screen where an owner decides whether to share a thread was the one screen that could not say why the thread was held.

Nothing failed, because the field is optional on the wire — a row with no reason and a row whose reason was dropped are the same bytes. It surfaced only by opening a contact on a dev stack and finding a badge that never appeared.

`activityprojectionfields_test.go` holds it from here: a file that scans into a `crmcontracts.Activity` and selects `a.audience` owes `a.audience_reason` beside it. The corpus is derived from what a file does rather than listed, so the next 360 or export joins the census on the commit it is written. Mutation-checked by removing the column again — the gate names the file.

A captured row now offers the **thread** verbs instead of the per-message dialog. That dialog is refused on captured mail (`audience_is_derived`, 422), so every press on a held row was an error the reader could not act on.

## Three things only the browser could show

Each of these passed typecheck and unit tests and was wrong on screen:

1. **The refusal reason was clipped.** It rode on the `shared` option's label, and a listbox row ellipsises — in German the half that got cut was the half a reader needs. It moved to the row description.
2. **A three-button confirm overflowed.** Two long German verbs plus Cancel did not fit the compact dialog; both were clipped. It became one question with a checkbox.
3. **`.card-actions` is not a table cell's verb group.** Its top margin pushed buttons off the row baseline and its wrap turned two verbs into a two-line cell, doubling every row's height. `.cell-actions` joins the design system with its own catalog entry.

## Acceptance, performed

On a dev stack (own `DEV_SLUG`, not `:8080`), as the mailbox owner and as a colleague, light and dark, German and English:

| Item | Result |
|---|---|
| Posture row renders, `classified` preselected | ✅ both themes, de + en |
| `shared` refused with the opt-in off | ✅ option disabled, reason in the description |
| Admin opt-in shows the works-council warning and saves | ✅ `shared_posture_allowed` PATCH carries only itself |
| `shared` becomes selectable afterwards | ✅ `aria-disabled` false after reload |
| Narrowing offers to narrow history | ✅ checkbox; posture persisted to `capture_connection` |
| Senders table lists every classifier kind | ✅ all six, correct verbs per row |
| "Business" overrules and redraws the row | ✅ returns `overruled_kind`, row shows "you decided" |
| "Keep out" confirms before destroying | ✅ names the destruction and what survives |
| Timeline shows the hold reason and thread verb | ✅ five held rows; a shared row takes the ordinary control |
| Sharing a thread opens it | ✅ `shared:true`, `held_by_others:0`, DB agrees |
| **As a colleague, a held row shows nothing** | ✅ lock + "Inhalt nur für Beteiligte", no subject, no sender, **no reason** |

The last row is the property the whole feature rests on: a colleague learns a message exists on a date, and nothing about it — including why it is held.

## Not in this PR

`GET /capture/held-threads` and `GET /ai/health` have no backend, so the held-threads list and the admin AI-health page are not built. The outage banner needs the latter. Filed rather than stubbed: a page reading an endpoint that does not exist is worse than an absent page.

## Tests

- `mail-sharing.test.tsx` — the opt-in saves on its own; mutation-checked by making the PATCH carry both settings, which fails it.
- `capture-senders.stories.tsx` — every classifier kind, an overruled row, and the empty state.
- `backend/gates/activityprojectionfields_test.go` — the projection census; mutation-checked.

`make check-backend`, `make check-fe` and the compose integration lane are green.

---
## Summary by cubic
Adds the owner and admin UI for the mailbox-privacy feature that shipped backend-only: a posture selector per connected mailbox, the organization's opt-in that unlocks `shared`, and a Senders page to audit what the classifier decided. It also fixes the record timeline, which never received the hold reason, and replaces the per-message audience dialog on captured mail with thread-level verbs — that dialog was always refused (422 `audience_is_derived`), so every press on a held row was a dead error.

**New Features**
- `MailPostureRow` under each connected mailbox picks `classified`, `held`, or `shared`; `shared` renders disabled with the reason when the organization has not allowed it, rather than hidden.
- Narrowing to a stricter posture asks whether to narrow already-captured mail; widening saves immediately.
- The mail-sharing card gains a separate `shared_posture_allowed` switch, saved by its own button so one setting cannot be flipped while meaning the other; turning it on shows the works-council warning.
- The Senders page lists every sender, the classifier's verdict, and the seat's own answer, with Business, Keep out, and Undo verbs; an overruled row carries a filled badge and "you decided".
- A captured row's timeline actions now share or keep private the whole thread; the outcome reports how many other seats still hold it.

**Bug Fixes**
- The person 360 read built its own SELECT without `a.audience_reason`, so the timeline — the one place an owner decides whether to share — could not show why a thread was held; the optional wire field made it fail silently.
- A census gate now derives every writer of `Activity.AudienceReason` from the AST and requires each to have a behavior test proving the reason is withheld from a reader who may not see the content.

<sup>Written for commit 058fc26f45b272963cf56946bcc58e8f06657269. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/3418?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Review captured senders, classify them as business, exclude them, or withdraw decisions.
  - Manage sharing for entire email threads, including restricted-visibility reasons.
  - Configure mailbox sharing posture and apply changes to historical mail.
  - Control shared-posture permission separately from general mail sharing.
  - View profile-field evidence and feedback details in Person 360.
- **Improvements**
  - Added German and Vietnamese translations for privacy and sender controls.
  - Added consistent table-row action styling.
- **Bug Fixes**
  - Restricted activities no longer expose audience reasons or message content to unauthorized viewers.


LarsGradion:
## Review round: the gate proved the wrong thing

Both reviewers landed on the same defect and they were right. Fixed in `058fc26f4`.

**The gate I shipped was a text census.** It checked that a file selecting `a.audience` also mentions `a.audience_reason` — the SQL spelling, not the privacy rule. I reproduced the consequence: delete the line that nils the reason on the withheld arm, and the gate stays green while every held row tells a colleague it is held because `personnel`. The whole backend suite was green with that leak live.

**It also failed short two ways**, which AGENTS.md rule 8 calls the one way a census must not break:

| Spelling | Old gate |
|---|---|
| `a.audience AS audience,` with the reason dropped | passed |
| `act.audience,` with the reason dropped | passed |

The second matters because a non-`a` alias is ordinary style in this tree — `sarsections.go` uses `at`, `export_scope.go` uses `av`.

### What replaced it

Two behaviour tests, each mutation-checked one clause at a time:

- **`TestAWithheldRowCarriesNoAudienceReason`** drives the real scan through `scanActivity` with `content_available` false and every content column *set*, so the question is what `record()` does with a reason that is there. It also asserts the safe markers survive — a version that blanked everything would withhold the row's existence, which is a different bug.
- **`TestThePersonPageWithholdsALimitedMessagesReasonFromAColleague`** covers the second writer, which is where the bug was. No test anywhere covered a withheld row on the person 360 read. It asserts the **author's** page keeps the reason *before* asserting the colleague's does not, so it fails in both directions: a leak, and the original dropped column.

Mutation results, each reverting exactly one clause:

| Mutation | Result |
|---|---|
| Drop `a.AudienceReason` from the withheld arm (person360) | fails, naming the leaked value `"manual"` |
| Restore the original defect — column out of SELECT, scan and assignment | fails on the author arm |
| Drop `audienceReason` from the shared projection's withheld arm | fails |

### The gate now asks a question a text scan cannot fake

Who writes `Activity.AudienceReason`, derived from the AST, and does a behaviour test prove each one. Two writers today. Verified by adding a leaking export under a different table alias — the old gate passed it, this one names the file.

Its three failure modes are each checked: an undeclared writer, a census entry whose named proof no longer exists, and a stale entry for a file that no longer assigns the field.

### One reviewer note not acted on

The security review suggested `gatekit.FileReadsTable` for corpus derivation. The AST walk over assignments to the contract field is a tighter subject — it is the writers of the promise rather than the readers of the table — so I kept that. Both reviewers' primary finding is fixed.

`make check-backend` green; the audience integration tests and the new unit test pass.

PR #3422, opened 2026-08-31
main is red: a stray content type on an empty one-click body is still absent

tinnhlt-gra:
## What's broken

`main` is red: `TestPreferenceCenterOneClickUnsubscribe` fails with a 422 on the one-click unsubscribe POST. `main-health` run https://github.com/margince/margince/actions/runs/33450789404 caught it after #3417 landed (that PR fixed a different, unrelated red — a frontend type mismatch).

Reproduced on a clean `origin/main` worktree via `make test-it DIR=backend/internal/compose/integration RUN=TestPreferenceCenterOneClickUnsubscribe`.

## Root cause

`requireOneClickBody` (added by #3415) refuses any one-click unsubscribe whose `Content-Type` isn't one of the two RFC 8058 form encodings — even when the body is genuinely empty. That contradicts its own doc comment: *"An ABSENT body is admitted too... Only a body that is present and says something else is refused."*

The integration test's shared `publicCall` helper always sets `Content-Type: application/json`, even when the caller passes a nil body. That's a legitimate client shape (plenty of HTTP clients default a content type regardless of payload), and the handler's own contract says it should be admitted — so the fix is in the handler, not the test helper.

## Fix

`requireOneClickBody`'s default branch (unrecognized media type) now checks whether the body is actually empty before refusing. A stray content type on a genuinely empty body is admitted; a stray content type on real content is still refused (unit test for the JSON-with-real-content case is unchanged and still passes).

## Verification

- `go test ./backend/internal/modules/consent/...` — all pass, plus a new case: empty body + `application/json` content type is admitted
- `make test-it DIR=backend/internal/compose/integration RUN=TestPreferenceCenterOneClickUnsubscribe` — now passes (was 422, now 200)
- `go build ./...`, `go vet ./internal/modules/consent/...`
- `craft static` (pre-push hook) — 0 blocker, 0 major, 0 minor

---
## Summary by cubic
Fixes the one-click unsubscribe handler so an empty request body is accepted regardless of the declared content type, matching the documented absent-body contract. Previously, a stray `Content-Type` on an empty body returned a 422, while a junk body with no content type was admitted. Both the no-`Content-Type` and unrecognized-`Content-Type` paths now share a single bounded read, enforcing the same size limit and failing closed on read errors.

<sup>Written for commit 1a2e2acdfd5894646e64af75e8e28d96c9715beb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/3422?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved one-click request validation for missing or unsupported content types.
  - Empty request bodies continue to be accepted, while nonempty, oversized, unreadable, or invalid bodies are rejected with validation errors.
  - URL-encoded requests now follow consistent body-size and validation rules.
  - Existing multipart request handling remains unchanged.

### DIFF
--- a/backend/internal/modules/consent/oneclickbody.go
+++ b/backend/internal/modules/consent/oneclickbody.go
@@ -75,8 +75,28 @@ func requireOneClickBody(r *http.Request) error {
 	case "application/x-www-form-urlencoded":
 		return oneClickFromForm(r)
 	default:
-		return httperr.Validation("body", "not_one_click", "a one-click unsubscribe carries the RFC 8058 form body")
+		// A stray content type on a body that turns out to be empty is still
+		// the absent-body case the module comment promises: a caller that
+		// declares a type it never fills must not be refused for content it
+		// did not send.
+		return requireEmptyBody(r)
 	}
+}
+
+// requireEmptyBody admits a body only if it turns out to hold nothing,
+// regardless of what content type declared it.
+func requireEmptyBody(r *http.Request) error {
+	body, err := io.ReadAll(io.LimitReader(r.Body, oneClickBodyLimit+1))
+	if err != nil {
+		return httperr.Validation("body", "not_one_click", "could not read the one-click body")
+	}
+	if len(body) > oneClickBodyLimit {
+		return httperr.Validation("body", "too_large", "a one-click unsubscribe body is a single short field")
+	}
+	if len(body) == 0 {
+		return nil
+	}
+	return httperr.Validation("body", "not_one_click", "a one-click unsubscribe carries the RFC 8058 form body")
 }
 
 // oneClickFromForm reads the URL-encoded spelling.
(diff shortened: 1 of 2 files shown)


tinnhlt-gra:
Ran the required end-of-work review round (craft-reviewer + security-redteam) on the original fix. Applied their findings in the follow-up commit:

- **security-redteam**: no exploitable defect (fail-closed body reads, bounded size matches existing paths, no tenant/authz change). Two low-severity notes accepted as-is (no fix needed).
- **craft-reviewer**: caught a real gap — the no-Content-Type branch never actually checked for emptiness, so a junk body with no Content-Type was silently admitted while the same junk under a stray Content-Type was refused, contradicting the same "absent body only" contract this PR restores. Consolidated both branches onto one `requireEmptyBody`/`readOneClickBody` predicate, made read failures fail closed on both paths, and added the two missing test cases (junk body + no content type; oversized body under an unrelated content type).

PR #3439, opened 2026-09-01
main is red: the assignee-filter test never disowned its control task

tinnhlt-gra:
## What's broken

`main` is red: `TestTheActivityListNarrowsByAssigneeOnTheWire` fails — `GET /v1/activities?assignee_id=...` returns 2 records instead of 1. `main-health` run https://github.com/margince/margince/actions/runs/33462177861 caught it.

## Root cause

Not a filter bug — `openTaskAssigneeClause` and its wiring from the HTTP handler down to the WHERE clause are correct, verified by reading the whole path.

The test's control record ("Nobody's" task, posted with no `assignee_id`) was expected to stay unowned, but `taskAssignee` (`backend/internal/modules/activities/activity.go`) self-assigns any task a human posts with no explicit assignee to the caller — the intended behavior landed by #3431 ("Mine means mine"). So the control task silently became the caller's own task too, and the assignee filter correctly returned both records.

The test file already has a helper, `disownOverDB`, built for exactly this situation (its own doc comment: *"A create over the wire stamps the calling seat as owner when the body names none, so the unowned state these queue tests are about has to be produced after the fact"*), used by the person/lead unowned-queue tests in the same file — this one test just predates it and was never updated.

## Fix

- Generalized `disownOverDB` to take the owner column name as a parameter (`owner_id` for person/lead, `assignee_id` for activity) instead of hardcoding `owner_id`.
- Updated the two existing call sites (person, lead) and one more in `automation_runs_preview_integration_test.go` to pass `"owner_id"` explicitly.
- Called it on the "Nobody's" task in the failing test.

## Verification

- `go build -tags integration ./...`
- `make test-it DIR=backend/internal/compose/integration RUN='TestTheActivityListNarrowsByAssigneeOnTheWire|TestThePersonListNarrowsToTheUnownedQueueOnTheWire|TestTheLeadListNarrowsToTheUnownedQueueOnTheWire'` — all pass
- `craft static` (pre-push hook) — 0 blocker, 0 major, 0 minor

---
## Summary by cubic
Fixes `TestTheActivityListNarrowsByAssigneeOnTheWire` so the assignee filter is asserted against an unowned control task. The task posted without an assignee was self-assigned to the caller at write time, so the filter returned both records; the test now disowns it before asserting. This is a test-only change: `disownOverDB` now accepts the owner column name (`assignee_id` for activities, `owner_id` for people/leads), all call sites pass it explicitly, and the test reuses `callerUserID`.

<sup>Written for commit 600db9e036f6ff2096da28bab1e19f9f32e07053. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/3439?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Tests**
  * Improved integration test coverage for ownership and assignment workflows across activities, people, and leads.
  * Enhanced validation and diagnostics when records are disowned or ownership fields are cleared.
  * Streamlined test setup by reusing authenticated user context and explicitly handling ownership fields.

### DIFF
--- a/backend/internal/compose/integration/automation_runs_preview_integration_test.go
+++ b/backend/internal/compose/integration/automation_runs_preview_integration_test.go
@@ -225,7 +225,7 @@ func assertPreviewMeasuresWithoutWriting(t *testing.T, e *apptest.AppEnv, autoID
 	}
 	// The create stamps the admin as owner; the recipe counts leads nobody
 	// owns, so the lead is disowned to be the unrouted one the preview must find.
-	disownOverDB(t, e, "lead", lead.ID)
+	disownOverDB(t, e, "lead", "owner_id", lead.ID)
 
 	var rowsBefore int
 	if err := e.Owner.QueryRow(context.Background(), `
--- a/backend/internal/compose/integration/declaredfilters_http_integration_test.go
+++ b/backend/internal/compose/integration/declaredfilters_http_integration_test.go
@@ -30,15 +30,16 @@ import (
 // disownOverDB nulls a record's owner column at the database. A create over
 // the wire stamps the calling seat as owner when the body names none, so the
 // unowned state these queue tests are about has to be produced after the
-// fact. The column varies by table: person and lead use owner_id, activity
-// (a task's self-assignment, taskAssignee in activity.go) uses assignee_id.
+// fact. The package-local Env fixture spells this same idiom inline through
+// e.WsExec instead of sharing this helper — a different fixture type, not a
+// second copy of the invariant.
 func disownOverDB(t *testing.T, e *apptest.AppEnv, table, column, id string) {
 	t.Helper()
 	if err := apptest.InWorkspace(e, t, func(tx pgx.Tx) error {
 		_, err := tx.Exec(t.Context(), `UPDATE `+table+` SET `+column+` = NULL WHERE id = $1`, id)
 		return err
 	}); err != nil {
-		t.Fatalf("disown %s %s: %v", table, id, err)
+		t.Fatalf("disown %s.%s for %s: %v", table, column, id, err)
 	}
 }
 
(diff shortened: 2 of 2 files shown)


tinnhlt-gra:
Ran the required end-of-work review round (craft-reviewer + security-redteam). Applied findings in the follow-up commit:

- **security-redteam**: no exploitable defect. Confirmed the widened `disownOverDB` interpolation surface (table + column, was table only) stays a compile-time literal at all four call sites, `apptest.InWorkspace` runs on the fixture path not a tenant path, and the production behavior the test now exercises (`taskAssignee`, the assignee-filter WHERE clause) is correct. One coverage observation (not a security finding, no action taken): the wire test only has a "disowned" control arm, not a "different human's task" arm — but the store-level twin (`declaredfilters_integration_test.go`, using a third rep) already owns that arm, so nothing is uncovered overall.
- **craft-reviewer**: five small findings, all applied — moved the per-table rationale out of the call site and onto the helper doc (state the rule once), added the column name to the helper's failure message, noted why the package-local `Env` fixture's inline `e.WsExec` spelling of the same idiom isn't folded into this helper (different fixture type — AGENTS.md's "two writers... or say why they do not"), reused `callerUserID` instead of open-coding the `/me` read, and renamed `nobodys` to `unassigned` to match the person/lead sibling tests' naming.

dungbanguyen:
Heads-up: #3437 carries the same fix for this test, alongside the other two things main is red on — three datasource doubles that panic on `Provider.Read`, which is what the failing shards here are hitting. Whichever lands first, the other will want a rebase; happy to drop my half of this one if you'd rather keep yours.

tinnhlt-gra:
Superseded — #3437 was expanded (retitled to include "and an unassigned task is not unassigned") and merged the same fix independently as `nullOverDB` before I could land this. Closing to avoid the duplicate.

PR #3462, opened 2026-09-01
fix(contract): the generated types carry the description the contract states

dungbanguyen:
## Main is red, twice

`make drift` fails on a clean `origin/main` worktree:

```
--- a/backend/internal/contracts/api_gen.go
-  // PayloadCaptureEnabled The deployment's `capture.trace_payloads` posture. False means …
+  // PayloadCaptureEnabled The deployment's `capture.trace_payloads` posture, on unless the deployment file turns it off. False means …
```

`crm.yaml`'s description was reworded without regenerating `internal/contracts`, so the `deterministic-gates` lane is red on the tip and every open PR inherits it.

This is **regeneration only** — no contract edit. The generated text is now what `crm.yaml` says.

## Scope note

This PR originally also carried the enrichment-fixture fix for the *other* red-main failure. [#3470](https://github.com/margince/margince/pull/3470) landed the same fix independently and byte-for-byte, so that commit is dropped here rather than duplicated — this PR is now the drift half alone, which nothing else covers.

### DIFF
--- a/backend/internal/contracts/api_gen.go
+++ b/backend/internal/contracts/api_gen.go
@@ -24943,7 +24943,7 @@ type PipelineTrace struct {
 	// Connector The provider id that carried the message, never a display label; resolve it against `GET /channel-providers`. Empty when the trace rows have been swept.
 	Connector *string `json:"connector,omitempty"`
 
-	// PayloadCaptureEnabled The deployment's `capture.trace_payloads` posture. False means no rung carries `counterparty` or `subject` because the operator did not turn payload capture on — as against a rung that simply has none.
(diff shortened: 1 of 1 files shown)


dungbanguyen:
Note: #3470 carries this same regeneration, because `make drift` fails on every branch until it lands and that PR could not otherwise prove its own fix (the enrichment fixture). Whichever merges first, the other's copy of the regenerated file becomes a no-op on rebase.

dungbanguyen:
Superseded by [#3470](https://github.com/margince/margince/pull/3470), which landed both halves of the red main — the enrichment fixture and the stale generated contract. Nothing left here.

PR #3492, opened 2026-09-01
main is red: the meeting arm of the cohort repair has no implementation

dungbanguyen:
`main` does not compile. #3485 landed `cohortpromote.go` calling
`linkAttendedMeetings`, the ownership waiver naming `meetingcohort.go`, and
`meetingcohort_integration_test.go` asserting what that function does — but not
the file itself:

```
internal/modules/people/cohortpromote.go:94:19: undefined: linkAttendedMeetings
internal/modules/people/cohortpromote.go:284:23: undefined: maxMeetingLinksPerActivity
```

The `people` package does not build, so every branch is red and no test in it
runs at all.

## Written from what the tree already states

Not invented: `PeopleOwedACohortRepair`'s selector is the same predicate read the
other way round — connector-captured, unrestricted, unarchived meetings the
person attends through a participant row, not already linked to them, on an
activity below the per-activity ceiling — and the three integration tests say
what the write owes.

Three choices worth reading, each of which the tests pin:

- **Per person, not per activity.** The mail arm refuses an activity any person
  is already linked to, because attaching a message to a second party would
  relabel somebody's mail. A meeting is the opposite shape — everyone in it
  belongs on it — so the guard asks whether *this* person is filed under it, or
  the second attendee is locked out by the first.
- **The attendee row resolves through `merged_into_id`.** A database written
  before the merge repointed participant rows still holds rows naming a retired
  id, and reading them literally files the meeting under a record nobody can
  open.
- **`maxMeetingLinksPerActivity` is a ceiling on the activity, not on the pass**,
  which is why it is not `cohortRepairBatch`: an invitation can carry a room full
  of addresses, and a company all-hands on two hundred records is not what any of
  them means by "my meetings".

## Proof

The three tests #3485 shipped without their subject now pass:
`TestTheRepairFilesAMeetingUnderAnAttendeeWhoArrivedLater`,
`TestAMeetingReachesEveryAttendeeNotOnlyTheFirst`, and
`TestAMeetingReachesTheSurvivorWhenItsAttendeeRowNamesAMergedRecord`. Whole
`internal/modules/people` package green.

@margince — if the original file exists somewhere unpushed, prefer it to this and
close me; I wrote this to unblock every other branch rather than to claim the
design.

### DIFF
--- a/backend/internal/modules/people/meetingcohort.go
+++ b/backend/internal/modules/people/meetingcohort.go
@@ -1,0 +1,92 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+package people
+
+// The cohort repair's MEETING arm.
+//
+// linkCapturedCohort beside it finds work by counterparty_email, and a calendar
+// meeting has none: attendance is a LIST, so the mapper leaves that field unset
+// and the only thing naming an attendee is a participant row. For a while the
+// difference was invisible, because a meeting usually has mail about it and the
+// page reads the mail instead — a meeting nobody emailed about was reachable
+// from nowhere.
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/jackc/pgx/v5"
+
+	"github.com/margince/margince/backend/internal/shared/kernel/ids"
+)
+
+// maxMeetingLinksPerActivity bounds how many people one meeting is filed under.
+//
+// An invitation can carry a room full of addresses, and every one of them that
+// later becomes a contact would otherwise put another link on the same activity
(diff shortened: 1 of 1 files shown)


LarsGradion:
Superseded by #3493, already merged as 75eeec70d — main compiles again.

Same diagnosis and the same single file. The cause was mine in #3485: the split into meetingcohort.go was staged with `git add -u`, which stages tracked files only, so the new file never made it into the commit while the gate had already passed against a working tree that had it.

The version on main is the Codex-reviewed one — it resolves the attendee's own merge redirect via coalesce(merged_into_id, id) and excludes archived meetings. Worth diffing against this branch's copy if it was written from scratch, in case it is missing either.

Sorry for the duplicated effort.

Issue #3430, opened 2026-09-01
Limiting a message removes the control that would undo it: audience_reason routes to a thread control that needs a thread_key hand-logged mail never has


tinnhlt-gra:
## Observed

Measured on core `b3147236b`, walking `UAT-AUD-1` (`audience-limited-message`).

A salesperson opens a person's timeline, presses **Visibility** on one message
and limits it to the people on it. The message is limited correctly. The
**Visibility** button then disappears from that row, and nothing takes its
place. The person cannot re-open the dialog, cannot widen the message again, and
cannot see what they chose.

The two rows below are from one page, one run, adjacent in the same timeline.
They differ in exactly one column:

| Row | `audience` | `audience_reason` | `thread_key` | Controls drawn |
|---|---|---|---|---|
| "what the yard will settle for" (limited by hand) | `participants` | `manual` | *null* | Reply, Relink |
| "slipway slot for week 22" (untouched) | `workspace` | *null* | *null* | Reply, Relink, **Visibility** |

Both were logged by a person — the product labels both *"typed by you"* — so
neither has a `thread_key`. The only thing that changed is that one of them was
limited, and limiting it is what took the control away.

### Why

`compose.tsx:2847` chooses the control by whether the row has a reason:

```tsx
{activity.audience_reason ? <ThreadAudienceAction … /> : <AudienceAction … />}
```

`PATCH /v1/activities/{id}/audience` sets `audience_reason = manual`
(`activities/audience.go:108-110`, *"The reason moves with the audience, and
becomes `manual`"*), so **any** hand-limited message takes the first branch.
That branch then refuses to draw anything without a thread
(`compose.tsx:2950`):

```tsx
if (activity.content_state === "withheld" || !threadKey) {
  return null;
}
```

The dispatch asks "does this row have a reason?"; the destination requires "does
this row belong to a thread?" Those are not the same set, and every manually
logged message sits in the gap. `#3418` made `audience_reason` visible to the
screen for the first time, which is what turned a dormant mismatch into this.

The comment above the guard explains the `withheld` half — *"Withheld content
carries no reason either, so there is nothing to draw and no standing to change
it."* It does not explain the `!threadKey` half, and that is the half that bites.

### It is not only limited rows

`audience_reason` stays `manual` after the audience is widened back. Across the
62 rows this scenario's fixture has accumulated, **18 carry a reason and no
`thread_key`** — every one of them has lost the control. Ten of those 18 read
[...]

Issue #3225, opened 2026-08-30
Settings → Users offers ops enabled team controls the server refuses, on the same page where the neighbouring card withholds its own


tinnhlt-gra:
Filed by QC. Measured on `6ae7d233d`.

## What happens

`ops` is the only non-admin seat that reaches **Settings → Users** (`useHoldsOperatorSeat`). On that page:

- **`UsersAdminCard` gates itself correctly** — `users-admin.tsx:97` calls `useHoldsAdminRole()`, withholds its controls, and prints "Managing users is available to admins only."
- **`TeamsCard` consults no role at all.** `grep` for `useHoldsAdminRole|isAdmin|useCanWrite` in `users-access.tsx` returns nothing.

So the same page hands an ops seat **enabled** membership tick boxes and a **New team** button for writes the server refuses:

```
ops → POST   /v1/teams                          -> 403
ops → PUT    /v1/teams/{id}/members/{userId}    -> 403
ops → DELETE /v1/teams/{id}/members/{userId}    -> 403
```

`SetTeamMember` opens with `if !actor.hasRole(roleAdmin)`, and there is no `team` object in `coreObjects` — the authority is structural, not a grant, so no seat short of admin can hold it.

## Why this is worth a ticket

**The product already makes this argument itself.** `TeamsCard` deliberately filters agent and inactive seats out of the member list, on the reasoning — in its own comment — that offering either is offering a box that can only fail. It then offers the whole remaining list to a seat the server refuses.

And the asymmetry is on one screen: a reader sees one card explain politely why it is unavailable, and another present working-looking controls beside it. Ticking a box produces a refusal, and the roster is unmoved on reload — so nothing is corrupted, but the reader has to discover by failing that the control was never theirs.

## Also: two statements that are now stale

Both say membership confers row scope, and it does not.

- `users-access.tsx`, the comment above `TeamsCard`
- `crm.yaml:11604`, the `addTeamMember` description

`identity/internal/policy/defaults.go` gives `manager` and `rep` both `RowScopeOwn`, with a comment stating the intent plainly: *"membership of a team is not by itself permission to rewrite a teammate's records."* `settings.tsx:295` and the card's own on-screen subtitle agree ("Being in a team grants no access on its own"). So the code and the UI copy are right and those two descriptions are behind.

[...]

Issue #3184, opened 2026-08-29
The AI verdict surface reaches no screen: EnrichedFields is mounted only by unrouted PersonScreen


tinnhlt-gra:
Filed by QC. This is not a defect in the verdict logic — that logic looks right. It is that nobody can reach it.

## What is unreachable

`EnrichedFields` (the "What Margince read" card) is the only place a person can record or see an AI verdict on an enriched profile field: the buttons "That is right" and "Correct", and the markers "Confirmed" and "Corrected by you".

It is imported in exactly one place:

```
$ grep -rn 'EnrichedFields' frontend/src --include='*.tsx' | grep -v test/stories
frontend/src/screens/personcorrections.tsx:41:export function EnrichedFields({
frontend/src/screens/people.tsx:62:import { EnrichedFields } from "./personcorrections";
frontend/src/screens/people.tsx:672:        <EnrichedFields personId={id} view={view} />
```

`people.tsx:672` is inside `PersonScreen`. `PersonScreen` is exported at `people.tsx:702` and **nothing imports it** — `App.tsx:442` routes `PersonPageV2` from `./screens/personpage` for a contact. The tree already says so in its own words, at `frontend/src/screens/people.stories.tsx:208`:

> PersonScreen (the 360 view below) is UNROUTED DEAD CODE: App.tsx routes...

And the routed page does not carry the card. `personpage.tsx` mentions `verdict` only for consent (`consentVerdict`, `entry.verdict === "allowed"`), which is a different concept.

`POST /ai/feedback` has exactly one caller in the whole frontend, `personcorrections.tsx:103` — inside the unmounted card. The operation is `x-agent-access: human-only` with cookie auth, so an agent seat cannot reach it either.

**Net effect: no user of Margince can confirm or correct an AI-filled profile field, and no user can see either marker.** The backend ledger, the RBAC gate and the 360 overlay all work; they have no consumer.

## Why this seems worth raising rather than just deleting

Two issues have been filed against the correctness of this overlay, and both read as though the surface is live:

- **#2773** (closed) — "A human's correction outlives the value that replaced it: the 360 verdict overlay has no recency check" — fixed by #3171.
- **#3176** (open) — "A verdict recorded from a stale page overrides the value that replaced it while the page was open" — which describes a person having a page open, and a race between what they see and what they submit.

[...]

PR #3535, opened 2026-09-01
Seven small QC-report gaps, one shared root cause

tinnhlt-gra:
## Summary

First of a layered batch fixing easy items from the [margince-qc report](https://github.com/margince/margince-qc/blob/main/reports/qc-report.md) tested against `966bff358`. This PR is frontend-only — every item here had a working capability or correct logic already; only the UI wiring or a rendering condition was wrong.

- **margince#3184** — the confirm/correct control for a field Margince read off a signature (`EnrichedFields`) existed but was built on the dead `PersonScreen` and never carried onto the routed `PersonPageV2`.
- **margince#3225** — team-membership checkboxes and the archive button had no permission check at all; an ops seat saw enabled-looking controls that 403'd. Gated on `useHoldsAdminRole()`, following the same absent-write-affordance-plus-one-line-explanation pattern `UsersAdminCard` already uses.
- **margince#3316** — a worklist row's tie-break explanation printed two raw clock times under a "how many days" heading when the backend's own comparator fell back from a bucketed day count to the exact occurred-at instant (a deliberate backend behavior, not a bug). Converts that instant to elapsed days client-side so the line matches its own heading, without touching the backend contract.
- **margince#3360** — a notice's supporting line (the deal's name) was already in the payload (`WorklistItem.detail`) and never rendered.
- **margince#3430 and margince#3447 are one defect, seen from opposite sides.** Which audience control a timeline row shows was keyed off `audience_reason`, which is null on an untouched CAPTURED row (wrong per-message button shown first) and non-null on a NARROWED hand-typed row (right button vanishes for good, since a hand-typed row has no thread for `ThreadAudienceAction` to act on). Keying off `thread_key` instead fixes both — confirmed live against a running dev stack before and after the fix.
- **margince#3490** — the outcome (mark introduced / name used) and withdraw actions on an introduction ask had working hooks (`useCompleteIntroRequest`, `useCancelIntroRequest`) that nothing in the app called. Wired into `IntroAsksCard`'s per-row actions, gated on the same actor rules the backend's lifecycle table enforces.

**margince#3526 is not included** — PR #3529 landed the complete fix (including a backend moment-card flip this PR alone could not reach) while this branch was in flight. My own draft of that fix was dropped after the merge to avoid a redundant, less complete change.

**margince#3122 needs no code change** — already fixed by #3123 before the QC report's own tested commit; the GitHub issue just never auto-closed (likely a squash-merge trailer drop). Will close separately with evidence, not as part of this diff.

## Test plan

- [x] `make check` (backend + frontend) green locally, including a golangci-lint run with a private cache to avoid a sibling worktree's shared-cache pollution
- [x] Unit tests added/passing for every fix: `personpage.test.tsx`, `users-access.test.tsx`, `worklist.copy.test.ts` (new file), `worklist.test.tsx`, `compose.test.tsx`, `introasks.test.tsx`
- [x] #3430/#3447 reproduced live against a running dev stack (before: Visibility button vanished after narrowing a hand-typed email; after: reopens correctly, and captured mail shows the thread-level control from the start)
- [x] #3184, #3225's read-only render confirmed via unit test with role-scoped fixtures
- [ ] UAT video attached as a follow-up comment

https://claude.ai/code/session_01C9j2rMQXvSRTWPyrqvMpC8

---
## Summary by cubic
Fixes seven small QC-report gaps on the frontend where the underlying capability already worked; only UI wiring or a rendering condition was wrong.

**Bug Fixes**
- Signature-read field confirm/correct controls now render on `PersonPageV2`; they existed only on the dead `PersonScreen`.
- Team-membership checkboxes and the archive button are gated on `useHoldsAdminRole()`; a non-admin now sees a plain "admins only" line instead of controls that 403 or a list that would be empty.
- The worklist "how many days" tie-break line converts the backend's exact-instant fallback to elapsed days, computed against the server's snapshot instant rather than the client clock, and falls back to the bare sentence when both sides round to the same day — never printing a false tie, negative count, or raw clock times.
- Notice rows now show the supporting line sent in `WorklistItem.detail`, which was previously never rendered; scoped to the `notice` source, since a relationship-decay row's detail is a raw day count, not prose.
- Timeline audience controls now key off the `captured_by` connector prefix instead of `thread_key` or `audience_reason`: `thread_key` would have routed a rep's own hand-typed threaded reply to a control that 404s, and `audience_reason` is null on untouched captured mail and non-null on narrowed hand-typed mail — the reverse of each case's need.
- Introduction asks now call `useCompleteIntroRequest` and `useCancelIntroRequest` from per-row actions, gated on the same actor rules the backend's lifecycle table enforces. Mutation state is matched per row, so a failed outcome on one row no longer reads as failed on the next row acted on.

margince#3526 is excluded: #3529 already landed the complete fix, including the backend change this PR can't reach. margince#3122 needs no change; #3123 fixed it before the QC report's tested commit.

<sup>Written for commit 900b1855e7f3adf0c08dbb95aea74619e2e740d0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/3535?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added controls to complete, withdraw, and report errors for introduction requests.
  - Added admin-only team membership and team-management visibility guidance.
  - Added enriched profile fields to person overviews.
  - Added worklist disposition actions, completion messages, and undo handling.
  - Added localized labels in English, German, and Vietnamese.
  - Added supporting details to worklist rows and improved waiting-day comparisons.

- **Bug Fixes**
  - Corrected audience controls for threaded and non-threaded messages.
  - Notice rows now display server-provided details when no subject is available.


tinnhlt-gra:
UAT: margince#3430 / margince#3447 live round trip.

Recorded against a running dev stack (rep@demo.test), on a hand-typed email:
1. Press **Visibility** on a hand-typed email → choose **Participants only** → save.
2. The **Visibility** button stays on the row (previously vanished for good).
3. Reopen it → the dialog correctly shows **Participants only** as the current choice, not the default.
4. Switch back to **Everyone in the organization** → save → confirmed via `GET /activities/{id}` that `audience` reverted to `workspace`.

The other 5 fixes (#3184, #3225, #3316, #3360, #3490) are unit-tested with fixtures reproducing each reported scenario; this one specifically needed a live round trip since the defect only shows up in server-state that a unit test can approximate but not fully replace.

PR #3584, opened 2026-09-01
Money in the Worklist, and the FX conversion under it

### LINKED ISSUE #3552: The Worklist ranks a yen against a euro: expectedBase is not a base-currency figure
`ranked.expectedBase` in `backend/internal/compose/attention/rank.go` documents
itself as:

> expected revenue in the installation's base currency, which is the ONLY figure
> by which two deals may be compared. Raw minor units compare a yen to a euro and
> get it wrong.

**No conversion happens.** `expectedRevenue`
(`backend/internal/compose/attention/classifyrisk.go:100`) returns the deal's own
`AmountMinor` verbatim, and `AttentionDealFacts` carries only `amount_minor` in
the deal's own currency — no base-currency amount reaches this package at all.

So the field is named for a conversion that does not exist, and the ordering
compares raw minor units across currencies. The comment describes the correct
behaviour; the code does the thing that comment warns against.

## What it costs

The `expected_revenue` tie-break decides which deal leads when neither is
overdue. On a mixed-currency pipeline it ranks by whichever number is larger as
an integer: ¥1,000,000 (about €6,000) outranks €50,000. The reader is told
"expected revenue" decided it, with both figures shown in their own currencies,
so the page looks self-consistent while the order is wrong.

## Why it is invisible today

Every deal in the demo data is EUR. A single-currency installation is correct by
accident, which is why this has not surfaced — and why it will surface as a
ranking that "looks random" on the first multi-currency pipeline rather than as
an error anybody can trace.

`summarize()` already documents the same problem honestly for the material
threshold: it withholds `base_currency` because "the bar is the median of raw
amount_minor values — expectedRevenue converts nothing, and says so". The
ordering did not get the same treatment.

## The fix

Bind the FX seam (`backend/internal/compose/contractfxseam.go`) into the
attention feed the way `WithDealFacts` binds the deal figures, convert in
`expectedRevenue`, and let `summarize()` name `base_currency` once the figure is
genuinely in one. Until then the honest minimum is to correct the comment so the
next reader is not told the ordering is currency-safe.

Found while scoping a revenue-at-risk aggregate for the Worklist's stat strip,
which needs exactly this converted figure and cannot be built honestly without
it.

LarsGradion:
Closes #3552 and #3591.

Started as the Worklist ranking fix. Two review rounds found a real bug underneath it, and that is now the larger half.

## 1. The Worklist ranks money in one currency

`ranked.expectedBase` documented itself as "expected revenue in the installation's base currency, the ONLY figure by which two deals may be compared", and nothing converted it — `expectedRevenue` returned the deal's own `AmountMinor` verbatim. Every deal in the demo data is EUR, so a single-currency installation was correct by accident.

- **`attention/basemoney.go`** — a `BaseMoney` seam, the per-read pricing pass, and the option that binds it.
- **`compose/attentionfxseam.go`** — the wiring, through `deals.PriceAll`.
- `expectedRevenue` and `materialBarOf` now answer in the base currency, and `summarize` sends `base_currency` — a field `WorklistSummary` has promised since the queue shipped and never sent.

A deal the estate cannot price ranks as UNPRICED, below every priced deal, with no material verdict. Its own amount still shows in its own currency.

## 2. The FX engine ignored minor-unit scales

Codex and the craft reviewer caught this independently, and it is the more serious finding.

A stored `fx_rate` is a **major-unit** rate — the ingestion prompt reads `"1 <from> = <rate> <to>"` off a rates page — while `amount_minor` counts **minor** units. `deals.ConvertToBase` multiplied them directly, which is correct only when both currencies carry the same number of decimals. JPY has none and EUR has two, so ¥5,000,000 (€30,000) converted to €300.

Fixed in the engine, not at my call site: it reaches the hierarchy rollup, org360's pipeline and quota attainment as well as the Worklist.

**Two tests in the tree had the wrong answer baked in** — the deals module's own ¥100 integration case, and the Worklist fixtures from my first commit. Both now state real money. The material-bar test is rebuilt so a converted median and a raw one genuinely differ, mutation-checked by taking the median from raw amounts.

## 3. The SQL copies (#3591)

Fixing only Go made it worse, not better: `open_pipeline_minor_base` is published for the same organization by the company RECORD (SQL view) and the company PAGE (Go), so a yen deal came out a hundredfold apart on two screens naming one figure. One invariant spelled on both sides of a wire is one item.

SQL cannot reach the Go digit table, so the table comes to SQL: `currency_minor_digits`, seeded by migration `1788303400`, read by the rollup view and by `quotas.targetInBase`. The app is granted SELECT only — the schema default would otherwise let a runtime write break parity silently.

The view keeps everything `1787905200` gave it (the per-deal bound, the bound on the sum, the `conv` lateral); only the conversion inside gains the scales.

## 4. An unrelated red on main, fixed here

`registryWithGate` gained a `MeetingBriefReader` and a `*slog.Logger` in #3596, and `composedRegistry` in `agenttierfloor_integration_test.go` was not updated. The file is integration-tagged, so the untagged build stays green and only the integration lane sees it.

**This is red on clean `origin/main`**, and it takes down `extension-reference` and `integration unit coverage` on every branch cut from it — including this one. Fixed here rather than filed, because it is one argument and this branch cannot go green without it. `TestATierFlooredForOneRecordTypeBindsTheToolDoorToo` compiles and passes for the first time since #3596.

## What holds all of it

- `TestOnlyOnePlaceMultipliesAnAmountByAStoredRate` (new gate) — fails when a second Go site multiplies by a stored rate. Its first version matched only a bare `rate.Rate` and reported PASS over a planted `rate.Rate.Int.Int64()`; it now searches the whole operand subtree, and that missed mutation is the case it is checked against. It says in its own comment what it cannot see.
- `minorunits_integration_test.go` — the live database table against the Go map, both directions.
- `openpipelineagreement_integration_test.go` — one yen deal through both surfaces, asserted equal as well as correct. Fails against either half alone.
- `deals/fxconvert_test.go` — the scale rule both directions, mutation-checked.
- `attention/basemoney_test.go` and `basemoneywiring_test.go` — the ordering inversion, the converted median, the unpriced deal, the unbound seam, a failed conversion failing the read; the last two driven through `Worklist` itself and mutation-checked.

## Still open

**#3592** — the frozen-rate readers. I audited it rather than leaving it as a to-do; findings are on the issue. Short version: there is exactly ONE wrong multiply, `deal.amount_minor_base`, a stored generated column computing the same unscaled product. Every reader I listed when filing is downstream of it and needs no change. Because the value is stored, correcting it is a table rewrite and it hits a real constraint — a generated column cannot reference another table, so it cannot read `currency_minor_digits`. That design question wants deciding before anyone writes the migration, which is why it stays its own change.

## Notes for review

- Rebased twice. #3590 independently landed the same `reader.worklistFrom` scope fix I had made here, plus a leads lane; my wiring test still earns its place because it drives `Worklist` end to end and carries the money assertion.
- `attention/feed.go` and `worklist.go` crossed the 500-line ceiling once main's leads lane and this branch's money code both landed. Moved by concept — `WithBaseMoney` and `priceDayOnto` to `basemoney.go`, `materialBar.stateOn` to `material.go`. Both files are now smaller than main left them.
- `make check-go` is fully green on the rebased branch.

## Summary by CodeRabbit

* **New Features**
  * Added currency-aware pricing for attention queues and expected-revenue ordering.
  * Converted amounts now account for differing currency decimal scales and display the base currency.
  * Added support for partially priced results when rates are unavailable.
  * Extended currency conversion consistency across pipeline totals, quotas, weekly reviews, and organization reporting.

* **Bug Fixes**
  * Improved rounding and overflow handling for cross-currency amounts.
  * Ensured unpriced deals do not receive misleading materiality or value assessments.


LarsGradion:
Correcting something I said imprecisely when triaging the craft review, since it is the kind of claim a reviewer would reasonably check.

The reviewer questioned whether the unbound-seam path (`fx == nil` → raw minor units, no claimed currency) earns its place, given production always binds it. I kept it and said "32 test sites across seven files rely on the optional shape." The count was right, the framing was not, and the two halves are different things:

- **32 sites pass `dayMoney{}` to `classifyDay`** across 3 files (`counts_test.go`, `worklist_test.go`, `worklistincidents_test.go`). That is an inert argument, not the degradation path — none of those tests assert anything about money.
- **31 sites construct a bare `&Service{}` and call `worklistFrom`.** Those genuinely depend on an unbound `fx` not exploding, and that is the real reason the optional shape is load-bearing.

Having checked rather than asserted: the degradation itself is one clause, `s.fx == nil`, sharing a guard with `day.AtRisk == nil` — a case that has to exist regardless. Making `fx` a required `NewService` argument would mean rewriting 31 test constructions to bind a stub, for no behavioural change. So the decision is unchanged, but it rests on the guard being nearly free rather than on a file count.

The reviewer flagged this as low-confidence and said they would not block on it. Recording it here so the reasoning is checkable rather than taken on trust.

LarsGradion:
Four findings from cubic and CodeRabbit, all checked against the code rather than taken on trust. Two were real, one I had genuinely gotten wrong, one is declined with evidence. Pushed as `7e9378893`.

## Fixed: a bound seam with nothing to convert read as unbound (cubic P1, CodeRabbit)

Both reviewers found this independently and it is the real one. If every at-risk deal carries an amount but no currency, `priceTheDay` builds an empty `amounts`, asks the estate nothing, learns no base, and returned the same zero `dayMoney` an **unbound** seam returns. `converted()` inferred its answer from `base != ""`, so `expectedRevenue` fell back to raw minor units and stated a `material` verdict over an amount whose units nobody knows — under a seam bound precisely to stop that.

I reproduced it as a failing test before touching anything: `TestADayOfUnitlessDealsIsConvertedWithNothingPriced` printed `"big-integer" states "material" over an amount with no currency`.

`dayMoney` now carries `ran` as an explicit field — a fact about the seam, not about the answer. `value()` also refuses to name a currency it does not have, rather than relying on that path being unreachable; relying on unreachability is what produced the bug.

## Fixed: PriceAll swallowed an estate fault as one unpriced row (CodeRabbit)

`ConvertToBase` returns two different refusals and `PriceAll` skipped both. Only one is about a single amount. A corrupt stored rate (NaN in `fx_rate`) makes *every* amount in that currency unconvertible until an operator fixes the row, so skipping it row by row reports a short total as a whole one for as long as the bad row survives. My comment justified the first case while the code swallowed both.

`ErrAmountOutOfRange` and `ErrRateNotFinite` are sentinels now. `PriceAll` skips the first, propagates the second. `TestTheTwoConversionRefusalsAreDistinguishable` pins it, mutation-checked in both directions.

## Fixed, and it was my error: the endpoint test proved nothing (cubic P2)

cubic is right and this is the finding I am most glad to have received. The fixture had ¥5,000,000 beating €4,000 on **both** readings — 5,000,000 > 400,000 as integers, €30,000 > €4,000 converted. I confirmed it by stripping the conversion out of `expectedRevenue`: the test still passed. A test that agrees with the defect it is named for.

Rebuilt so the two readings disagree: ¥9,000,000 (€54,000) against €80,000. Converted, the euro deal leads; raw, the yen integer wins. I then ran the raw-ranking mutation against the whole package — all five money tests fail against it now.

## Declined, with the check: the minor-unit table (cubic P2)

cubic flagged that a currency missing from `currencyMinorDigits` mis-scales by 100. Rather than assert the table is complete, I compared it against ISO 4217's published exception list: **33 codes, no omissions, no spurious entries, no digit disagreements.** `MinorUnitDigits` already documents why defaulting to two is correct rather than a fallback — refusing unknown codes would suppress amounts for nearly every currency on earth while admitting two dozen exceptions.

## Two things the gates caught that no reviewer did

The `REVOKE INSERT, UPDATE, DELETE` I added left the application role holding SELECT alone, and the workspace data sweep runs as that role — so `TestSweepWorkspaceDataClearsDomainKeepsIdentity` failed with `permission denied for table currency_minor_digits`. Adding it to `preservedResetTables` then tripped `seedresetparity_test.go`, which named a **third** mirror of that list in `scripts/seed-reset.sql` that I did not know existed.

## And a fifth surface, from main

Main's #3598 added a new `ConvertToBase` caller in weekly's pipeline comparison, with the old signature. It would have shipped the same yen bug on a sixth screen. Fixed here, and it is exactly the case the new gate exists for.

The migration is renumbered above main's `1788303969` so it applies in the same position on every installation rather than before it on a fresh database and after it on an existing one.

PR #3594, opened 2026-09-01
The audit boundary is green again on main: the waiver follows the file, and an unknown route withholds

joshbruh-lang:
## What

Two failures `main` has carried since #3581, both in the audit-log boundary, are fixed:

- `TestEveryReaderOfTheActivityTableExcludesRestrictedRows` passes again. The waiver for the compliance read's activity join is keyed on `auditaudienceboundary.go`, the file that now declares the join fragment, instead of `auditlog.go`.
- `TestTheCollateralTypesAreTombstonedSoTheBoundaryReachesThem/attachment` passes again. The `content_readable` predicate reads a route that cannot say whether a row is governed as governed, so the image is withheld instead of the predicate coming back NULL and the scan failing on the row.

## Why

#3581 moved the LATERAL join over `activity` into `auditaudienceboundary.go`. The restricted-readers gate reports a package-level fragment at the file that declares it, named by the function that composes it, so the waiver keyed on `auditlog.go:ListAuditLog` matched no subject and the reader was reported unwaived.

The same change made the route's `governed` column a subselect over the attachment row. For an attachment whose own row is gone — erased, which is exactly when the image left behind matters — that subselect is NULL, `NOT (governed_type AND NULL)` is NULL, and `ListAuditLog` failed the whole page with "cannot scan NULL into *bool". Unknown is not "not governed": the boundary is fail-closed, and the tombstone case already asks for that row to be there and withheld.

Both halves are red on `main` at 83d67e1 and on every PR merged with it: `extension-reference`, `integration unit coverage`, `deterministic-gates` and the integration shards. #3581's own final run shows the same four failures.

## How verified

- [x] `go test -count=1 ./gates -run TestEveryReaderOfTheActivityTableExcludesRestrictedRows` fails on `main` and passes with this change
- [x] `make test-it DIR=backend/internal/modules/privacy` against a local Postgres 16 with pgvector: the tombstone case fails on `main` with the scan error above and the whole package passes with this change
- [x] `craft static --strict` on the touched Go file: no findings
- [ ] `make check` in full: not run; the change is one map key in a gate test and one `coalesce` in a projection
- [x] `make test-integration` is green (or: this change does not touch tenant data / RLS / the write shape) — the privacy package, which owns the change, is green; see above

## AI involvement

Diagnosed from the failing checks' own messages and written by Claude Code, under the account holder's review.

## Accountability

https://claude.ai/code/session_01GHxSEFgDFYn3GvbmcfmHVU


joshbruh-lang:
Superseded: `main` now carries both of these fixes — the audit route's NULL answer landed through #3600, and the re-keyed waiver arrived with #3563. Closing this one; nothing left for it to do.

---

PR #3624, opened 2026-09-02
weeklyplan is a module the architecture knows about

dungbanguyen:
`main` is red again. Both failures are #3604's new `weeklyplan` module, which neither architecture gate had been told about.

## arch-lint: files attached to no component

```
File /internal/modules/weeklyplan/write.go not attached to any component in archfile
File /internal/modules/weeklyplan/weeklyplan_integration_test.go not attached to any component in archfile
```

The component entry says what the code already does: `weeklyplan` production imports `contracts` and `platform` and nothing else, so `mayDependOn: [contracts, platform]` — the same line `notices` and `introductions` carry beside it. `compose` gains it as a dependency because compose is what wires it.

## depguard: a module reaching into the composition layer

```
weeklyplan_integration_test.go:24:2: import 'internal/compose/weekly' is not allowed
  from list 'modules-common': modules never see the composition layer
```

The test moves to `internal/compose/integration`, where every other test that needs the `Env` harness already lives, and drops the now-redundant package qualifier. That is what makes the import legal rather than waived — it was crossing a boundary because it was sitting on the wrong side of one — and it leaves the module importing no composition at all, which is the property the rule exists to keep.

`make check-backend` green; the moved test green in its new home.

## Summary by CodeRabbit

- **Refactor**
  - Updated internal architecture configuration to include the weekly planning component and its approved integrations.

- **Tests**
  - Streamlined weekly planning integration test setup while preserving existing environment, permissions, and database access behavior.

- **Impact**
  - No changes to the end-user experience or publicly available functionality.

### DIFF
--- a/backend/.go-arch-lint.yml
+++ b/backend/.go-arch-lint.yml
@@ -107,6 +107,7 @@ components:
   comms:       { in: internal/modules/comms }
   notices:     { in: internal/modules/notices }
   introductions: { in: internal/modules/introductions }
+  weeklyplan:  { in: internal/modules/weeklyplan }
   compose:     { in: [internal/compose, internal/compose/**] }
   migrations:  { in: migrations }
   cmd:         { in: cmd/** }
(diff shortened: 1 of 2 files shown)


dungbanguyen:
Superseded. #3628 landed the test move — with `package integration_test` rather than `package integration`, which is the better call — and this branch's version of it conflicts with no purpose.

The archfile half is still missing on main, so arch-lint still reports `weeklyplan/write.go` and its siblings attached to no component. That alone is now #3625.

PR #3627, opened 2026-09-02
main is green again: three things #3563 and my own #3613 left red

LarsGradion:
Three breaks on `main`, none of which any single PR's own gate could see. All three merged green on their own branches and only collide in the tree.

## The backend did not compile

`waitingScanCap` was exported to `WaitingScanCap` by #3563, and one call site kept the old name. A comment two hundred lines up kept the old spelling too, so the rename was half-applied in both senses.

## `waiting.go` crossed the size cap

The same commit grew it to 563 lines. Rather than a waiver, the three functions it added move to `waitingrecord.go` — they answer a different question from the rest of the file: narrowing the who-is-waiting read to **one record**, for the record page's own panel, rather than to the workspace.

## And one of mine

#3613 put a module integration test where depguard forbids it. A module may never import the composition layer, and that test drives the compose harness. It moves beside the harness, which is where every other module test needing a seeded workspace already lives.

**Why I missed it:** the golangci lock was contended when I gated that branch, so I shipped on a scoped `golangci-lint run ./internal/modules/weeklyplan/...`. A scoped run does not evaluate depguard's module list — the rule is about which package the import comes *from*, and the scoped invocation had nothing to compare against. That is a real gap in how I have been working around the lock, not a one-off.

## Verified

`build`, `vet`, `arch-lint`, the full test suite, `go-file-length` and `lint` (0 issues) all green, plus the integration lane for the moved test.

---
## Summary by cubic
Fixes three breaks that only collide on `main`: the backend didn't compile, `waiting.go` crossed the file size cap, and an integration test sat where depguard forbids it.

**Bug Fixes**
- Updates the one call site and comment still using the old `waitingScanCap` spelling after the rename to `WaitingScanCap`.
- Moves the three per-record waiting functions from `waiting.go` into a new `waitingrecord.go` to get back under the 500-line cap.
- Moves the weeklyplan integration test beside the compose harness, where module tests needing a seeded workspace already live.

<sup>Written for commit a0ffadab0c6a715ccaf0323bfb235bb950525276. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/3627?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved visibility filtering for waiting replies in activity records.
  * Ensured activity lists consistently apply authentication, link visibility, and waiting-reply limits.
  * Improved error handling when applying activity visibility filters.

* **Tests**
  * Updated integration test organization and documentation for Compose-based database and integration coverage.

### DIFF
--- a/backend/internal/compose/integration/weeklyplan_integration_test.go
+++ b/backend/internal/compose/integration/weeklyplan_integration_test.go
@@ -3,9 +3,14 @@
 
 //go:build integration
 
-package weeklyplan_test
+package integration_test
 
 // The plan over real migrated Postgres.
+//
+// HERE rather than in the module, because it drives the compose harness and a
+// module may never import the composition layer — depguard says so. Every
+// module test that needs a seeded workspace lives beside the harness for the
+// same reason.
 //
 // What these defend is the access model and the settle, neither of which a unit
 // test can see: a plan is one person's, its second reader is gated on a live
--- a/backend/internal/modules/activities/waiting.go
+++ b/backend/internal/modules/activities/waiting.go
@@ -30,7 +30,6 @@ import (
 	"github.com/margince/margince/backend/internal/platform/auth"
 	"github.com/margince/margince/backend/internal/shared/kernel/ids"
 	"github.com/margince/margince/backend/internal/shared/kernel/principal"
-	"github.com/margince/margince/backend/internal/shared/ports/datasource"
 )
 
 // WaitingReply is one inbound message nobody has answered.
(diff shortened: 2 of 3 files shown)


LarsGradion:
Superseded: #3615 fixed the compile and the file size while this was in flight, and the remaining depguard violation ships as #3628 off current main.

Issue #3670, opened 2026-09-02
main is red again: the tag governed-vocabulary migration seeded no role a grant on it


tinnhlt-gra:
\`TestEveryRBACBackfillConvergesOnTheSeededMatrix\` fails on current \`origin/main\` (confirmed on PR #3663's CI, a merge with no changes to identity/collections):

\`\`\`
rbacseedparity_integration_test.go:236: every backfill in order: role "admin" holds no grant on "tag", so every tag route answers 403 for it — permanently, on any installation that took this path
rbacseedparity_integration_test.go:236: every backfill in order: role "management" holds no grant on "tag", so every tag route answers 403 for it — permanently, on any installation that took this path
rbacseedparity_integration_test.go:236: every backfill in order: role "manager" holds no grant on "tag", so every tag route answers 403 for it — permanently, on any installation that took this path
rbacseedparity_integration_test.go:236: every backfill in order: role "ops" holds no grant on "tag", so every tag route answers 403 for it — permanently, on any installation that took this path
rbacseedparity_integration_test.go:236: every backfill in order: role "read_only" holds no grant on "tag", so every tag route answers 403 for it — permanently, on any installation that took this path
rbacseedparity_integration_test.go:236: every backfill in order: role "rep" holds no grant on "tag", so every tag route answers 403 for it — permanently, on any installation that took this path
\`\`\`

## Why

\`1788320100_a_tag_is_governed_vocabulary\` introduces \`tag\` as a new RBAC object (governed vocabulary, per its own name), but carries no backfill migration granting any EXISTING seeded role a grant on it. A fresh install presumably gets the object through the seed defaults directly; an installation that upgrades through the migration replay never does — every tag route 403s for every role, permanently, on that path. The gate that catches this (\`TestEveryRBACBackfillConvergesOnTheSeededMatrix\`) is the same one #3497/#3500 fixed for exactly this shape (a migration widening/adding to the seeded matrix with no corresponding backfill) — this is a new instance of that class, not a regression of the fix itself.

## What it needs, and why I have not just done it

The mechanical repair is a new backfill migration granting each role its \`tag\` object grant — but which grant each role should get (read-only for \`read_only\`? create+read for \`rep\`? full for \`admin\`/\`management\`?) is a product decision about a **governed-vocabulary object**, the same category of judgment call #3497 deferred to the migration's own author rather than guessing. I'd rather the author of #3660 decide the matrix than have this issue guess it.

[...]

Issue #3497, opened 2026-09-01
main is red: an RBAC backfill replayed alone no longer converges on the seeded matrix


dungbanguyen:
`TestEveryRBACBackfillConvergesOnTheSeededMatrix` fails on clean `origin/main`, in four subtests:

```
1787449829_seeded_roles_own_scope:  role "manager" has row_scope "own", want "team"
1787818335_knowledge_corpus_rbac:   role "manager" has row_scope "own", want "team"
1788230541_an_introduction_is_a_grant_of_its_own: same
1788244324_a_team_lead_reads_their_team: role "manager" carries document key
    "row_scope_set_by", which this comparison does not model: "1788244324"
```

## Why

`1788244324_a_team_lead_reads_their_team` moved the manager's `row_scope` from
`own` to `team` and stamped the change with a `row_scope_set_by` key. The gate
rewinds to each backfill's starting state, replays **that one migration**, and
asserts the result equals the *current* seeded matrix — so every backfill written
before the scope moved now produces the old answer and fails, and the newest one
fails on a key the comparison does not know about.

Both failures are the same shape: the matrix moved, and a gate that replays
history in isolation has no way to know which matrix each migration was written
against.

## What it needs, and why I have not just done it

The mechanical repair is to replay the writes **in version order, cumulatively**,
and compare once at the end — which is what a real installation does. It also
weakens what the gate catches: an individual backfill that diverges and is
corrected later would stop failing, and the gate's own doc comment says
convergence is the point.

That is a judgement about what the gate is for, on a migration whose subject is a
**row-scope boundary** — the thing whose wrong value the test itself calls "a
disclosure or a blackout". I would rather the author of #3486 decide it than
guess, so this is a report rather than a PR.

It does block everything: `main-health` is red and every branch inherits the
failure. If nobody picks it up shortly I will take the cumulative-replay reading
and say so in the PR.

Confirmed on a clean `origin/main` worktree, `internal/compose/integration`.

PR #3671, opened 2026-09-02
A tag's governed color is one of the fixed set, not a hex value

tinnhlt-gra:
## Summary

`TestTagsLifecycleAndApplication` sends a hex color the new `tag_color_check` constraint (`1788320100_a_tag_is_governed_vocabulary`) refuses, so this one test has been turning `main-health` red for every branch that merges past that migration — confirmed via #3663's CI.

The contract already types `CreateTagRequest.Color` as the closed enum (`amber`/`rose`/`slate`/`teal`); the test bypassed it through the generic `AnyMap` helper and was never updated when the migration landed alongside it.

The same migration also left a real gap — no seeded role got a backfill grant on the new `tag` object, so `TestEveryRBACBackfillConvergesOnTheSeededMatrix` fails too — but which grant each role should get is a product decision, the same shape #3497/#3500 deferred to the migration's author rather than guessing. Filed as #3670 rather than fixed here.

## Test plan

- [x] `make test-it DIR=backend/internal/compose/integration RUN=TestTagsLifecycleAndApplication` passes
- [x] `craft static --strict` clean on the diff
- [x] `go build ./...` clean

https://claude.ai/code/session_01C9j2rMQXvSRTWPyrqvMpC8

---
## Summary by cubic
Fixes a failing integration test by sending a governed tag color (`amber`) instead of a hex value, so `TestTagsLifecycleAndApplication` passes the new `tag_color_check` constraint.

The known RBAC seed-matrix gap for the new `tag` object is tracked separately in #3670 and not addressed here.

<sup>Written for commit 82e66719a9d302238e1e001e3c3f41880dd6f063. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/3671?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Tests**
  * Updated the tag lifecycle integration test to use the governed `amber` color value instead of a hexadecimal color.


dungbanguyen:
Heads up — the integration lane has been red on `main` since the tag work landed, partly on the same line your PR touches: `collections_integration_test.go` still posts `"color": "#ff6b00"`, which `tag_color_check` now refuses, so `TestTagsLifecycleAndApplication` fails at the create and none of the lifecycle below it runs.

I have taken just that one line in https://github.com/margince/margince/pull/3682 to get the lane green — it also carries the missing RBAC backfill for the `tag` object, which is the other half of the same breakage. Nothing else of the colour work is touched, so your broader change here is unaffected; if #3682 lands first you may get a trivial conflict on that line.

tinnhlt-gra:
Superseded — landed independently on main via #3674 (`Tags reach the roles an installation already had`), which also fixed the RBAC backfill gap I'd filed as #3670 (now closed). Same fix, different session; closing without merging to avoid a duplicate/conflicting commit.

PR #3723, opened 2026-09-02
The harness's admin can see the words a rep can see

dungbanguyen:
`main` is red: `integration / integration shard (4/6)` fails on `TestThePersonListNarrowsByTagID` — "the tag filter returned 0 people, want only the tagged one". Reproduced on plain `origin/main` and at `edd89a55e`, the commit that added the test.

## Cause

`AdminPerms` in the integration harness carries no `tag` grant. `storekit.TagFilterClause` refuses a caller without `tag.read` by rendering `FALSE` rather than by erroring — deliberately, since an error would tell an outsider the tag exists — so `e.Admin()` asked "who carries VIP" and was answered "nobody". The suite read that as the store's answer.

The real seed (`identity/internal/policy`'s admin row) grants admin `crud` on both `tag` and `list`. The fixture was narrower than `AccountRepPerms`, a **rep** fixture, which already carries `tag.read` and `list.read`.

## Fix

`AdminPerms` gains the `tag` and `list` grants the seed gives it.

## Gate

`harnessperms.go` already says in prose that "the harness's admin stands in for every seat in most suites". The new gate is its contrapositive: no fixture for a seat the **seed** grants strictly less than admin may hold a grant `AdminPerms` lacks. It compares *seats*, not fixtures — a fixture is often deliberately narrower than its own seat, and several here say so where they are declared. A companion census parses the fixture names out of `harnessperms.go`, so an unregistered fixture fails rather than being exempt by omission.

It reads `migrations/testdata/rbac_seeded_defaults.json`, which `identity/rbacfixture_test.go` pins to `policy.MustDefaultJSON` on every unit pass — `policy` itself sits behind `identity/internal/`.

**Stated limit, in the doc comment:** a seat the seed grants exactly what it grants admin is never compared. `OpsPerms` holds the admin grid, and `AdminWithSignals` is an admin fixture carrying a `signal` grant `AdminPerms` lacks. Nothing holds that narrowing honest.

## Evidence

- `TestThePersonListNarrowsByTagID` — fails on `origin/main`, passes here.
- Mutation: drop `tag` from `AdminPerms` → the gate names it (`AccountRepPerms (rep) holds tag.read and AdminPerms does not`).
- Mutation: unregister `AccountRepPerms` → the census names it.
- `make check-backend` exit 0; the whole `compose/integration` package green (135s).

## Summary by CodeRabbit

* **Bug Fixes**
  * Administrators now have consistent create, view, update, and delete access for tags and lists.
  * Tag-filtered results now correctly reflect administrator permissions.
  * CRM-originated correspondence records unresolved sender visibility questions while still creating the associated person record.

* **Tests**
  * Added checks to keep administrator permissions aligned with configured policies.
  * Expanded integration coverage for correspondence processing, organization tags, query budgets, and permission behavior.

### DIFF
--- a/backend/internal/compose/integration/capture/capture_tiergate_integration_test.go
+++ b/backend/internal/compose/integration/capture/capture_tiergate_integration_test.go
@@ -377,9 +377,19 @@ func TestCaptureTierGateHonorsCorrespondenceFromACRMOriginatedSend(t *testing.T)
 		WHERE pe.email = 'team@news.prospect.example'`); n != 1 {
 		t.Fatalf("%d persons for a sender the CRM had written to, want 1 — a CRM send must count as correspondence", n)
 	}
+	// A ledger row here is expected, and it is not a deferral. #3719 made both
+	// create tiers open the same question the deferred tier opens, because a
+	// sender created on sight and never judged stayed the mailbox owner's
+	// permanently. T1 still decides STORAGE — the person above exists, which is
+	// what "nothing defers" was ever about. What the row asks is whose record it
+	// is, and `advisor` is a legitimate answer that keeps it private, so the
+	// question cannot be skipped for a corresponded-with sender either.
 	if n := countRows(t, e, `
-		SELECT count(*) FROM capture_pending_counterparty WHERE email = 'team@news.prospect.example'`); n != 0 {
-		t.Fatalf("%d ledger rows for a corresponded-with sender, want 0 — T1 decides, nothing defers", n)
+		SELECT count(*) FROM capture_pending_counterparty
+		WHERE email = 'team@news.prospect.example'
+		  AND status = 'pending' AND resolved_at IS NULL`); n != 1 {
+		t.Fatalf("%d open visibility questions for a corresponded-with sender, want 1 — "+
+			"T1 creates the record on sight and still asks whose it is", n)
 	}
 }
 
--- a/backend/internal/compose/integration/harnessperms.go
+++ b/backend/internal/compose/integration/harnessperms.go
@@ -199,6 +199,15 @@ var (
 			objInstallSettings: {Read: true, Update: true},
 			"project":          {Create: true, Read: true, Update: true, Delete: true},
 			objRelationship:    {Create: true, Read: true, Update: true, Delete: true},
+			// tag and list are crud for admin in the real seed
+			// (identity/internal/policy's admin row), and this fixture had
+			// neither. That drift was not inert: the tag FILTER refuses a
+			// caller without tag.read by rendering FALSE rather than by
+			// erroring, so an admin missing the grant asked "who carries
+			// VIP" and was answered "nobody" — and this fixture was
+			// narrower than AccountRepPerms, which already carries both.
+			"tag":  {Create: true, Read: true, Update: true, Delete: true},
+			"list": {Create: true, Read: true, Update: true, Delete: true},
 		},
 		RowScope: principal.RowScopeAll,
 	}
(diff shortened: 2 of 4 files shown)


dungbanguyen:
Two further failures came back from CI. They are not the same thing, so splitting them:

**`org360` (mine).** `TestOrganization360CostDoesNotGrowWithTheAccount` — 44 queries against a budget of 43. Caused by this branch: with `tag.read`, `assemble.readTags` runs where it was previously refused, adding one statement for the account's chips. It is flat in the size of the account, so the flatness assertion still holds; only the recorded cost moved. Budget raised to 44 with the running commentary the file uses. Verified on plain `origin/main`: **passes** there, so this is mine to own.

**`capture` (main's).** `TestCaptureTierGateHonorsCorrespondenceFromACRMOriginatedSend` — "1 ledger rows for a corresponded-with sender, want 0". Verified on plain `origin/main`: **fails** there too. #3719 deliberately made both create tiers open the verdict question the deferred tier opens — its own commit message says so — and added `capturepromote_integration_test.go` without updating this suite. The assertion now reads the row as what it is: an *open* question (`status = 'pending' AND resolved_at IS NULL`) about whose record it is, not a deferral. T1 still decides storage; the person exists, which is what "nothing defers" was ever about.

Folded into this PR rather than split, because `main` is red on all six shards and this is the change that makes it green.

**Evidence**
- Mutation on the capture assertion: cut the `askWhoseRecord` call → `0 open visibility questions … want 1`.
- Every `compose/integration` subpackage run locally: `agentaccess`, `capture`, `channels`, `collections`, `customfields`, `jobfanout`, `org360`, `overlay`, `webhooks` — all `ok`; the parent package `ok` (135s); `e2e` has no test files.

dungbanguyen:
Valid, fixed — with one adjustment to the suggested remedy.

Restricting the parse to `principal.Permissions` **composite literals** would have exempted `AdminWithSignals`, which is `withFullSignalGrant(AdminPerms)` — a call, not a literal. That is the one fixture in the file assembled rather than written out, so admitting only literals would introduce the exact "exempt by omission" hole the census exists to close, while closing the one you found.

So it now narrows by exclusion instead: a composite literal of some *other* type is provably not a fixture and is skipped; everything else — literals of `principal.Permissions`, calls, identifiers — still has to be registered. Blank identifiers are skipped too, since `_` names nothing a suite can act through. Deciding this properly wants the type checker, and the module has no `go/packages` dependency; the doc comment says so rather than implying the AST answer is complete.

**Evidence**
- `var probeGrant = principal.ObjectGrant{Read: true}` added to `harnessperms.go` → census passes (your case).
- `AdminWithSignals` removed from the registry → census fails by name (the case the narrowing must not lose).
- `make check-backend` exit 0; `compose/integration` green.

dungbanguyen:
Both correct, both fixed — they are the same failure in two places: an exclusion that drops a fixture with nothing failing.

**Per name.** The exclusion ran per *declaration*, so in `var A, B = principal.Permissions{…}, principal.ObjectGrant{…}` the second value dropped the whole spec and exempted `A`. It now decides per name against the value at the matching index. The multi-return form (`var a, b = f()`) has no per-name value to inspect, so it is asked for rather than excluded.

**The qualifier is resolved.** Matching the bare identifier `principal` would read an aliased import as some other package and exclude every `Permissions` literal in the file — silently. It now walks the file's imports, maps the path back to whatever identifier qualifies it, and matches on that. A file that does not import the package, or dot-imports it so no qualifier survives, fails loudly rather than passing on an empty census.

**Evidence** — three mutations, each on the branch it exercises:
- Alias the import in `harnessperms.go` to `p` → census still passes, all seven fixtures found.
- `var ProbeSeat, probeGrant = principal.Permissions{…}, principal.ObjectGrant{…}` → fails by name: `harnessperms.go declares ProbeSeat, which harnessSeatFixtures does not name`.
- Point `principalPackage` at a path the file does not import → `harnessperms.go does not import … — an empty census would pass while governing nothing`.

`make check-backend` exit 0; `compose/integration` green.

Issue #3736, opened 2026-09-02
TestDraftEmailFiringLandsTheComposedDraftOnTheRunRecord tests a case #3727 made impossible


tinnhlt-gra:
## What's red

`TestDraftEmailFiringLandsTheComposedDraftOnTheRunRecord`
(`backend/internal/compose/integration/workflow_action_executors_integration_test.go:183`)
fails deterministically on current `main`, blocking the shared `ci` required
check (confirmed via `make test-it DIR=backend/internal/compose/integration
RUN=TestDraftEmailFiringLandsTheComposedDraftOnTheRunRecord` against a clean
checkout of `main`, not a flake — reproduced twice):

```
workflow_action_executors_integration_test.go:200: workflow task11a_draft_email_probe:
this automation drafts an email but has no owner; a drafted message goes out under
one person's name and is released by that person, so assign an owner before enabling it
```

## Why

#3727 ("A held draft is released by the person it goes out as") added a
compose-time refusal: any firing that drafts email with no owner refuses,
because a drafted message goes out under one person's name
(`engine_run.go`'s `withSendingOwner`, `handlers_actions.go:206`
`principal.SendingHuman`).

The test's own registration path makes an owner structurally unreachable.
`draftEmailProbe` is registered via `engine.RegisterSystemWorkflow(...)`, which
runs through `engine.go`'s `system` handler loop — that loop calls
`e.runOne(runCtx, h, ev)` directly with no `OwnerID` assignment at all,
unlike the catalog/instance loop below it (`iev.OwnerID = inst.owner`). The
contract is explicit and intentional
(`internal/shared/ports/workflow/workflow.go:83-87`, `Event.OwnerID` doc):
*"Zero for a system-seeded automation... and for a system handler (no
instance behind it at all)."*

So this is not a missing test fixture value (I first assumed the seeded
deal just needed an owner — `SeedDeal(..., &e.Rep1)` — and confirmed that
does **not** fix it, since `RegisterSystemWorkflow`'s path never reads the
deal's owner or `iev.OwnerID` in the first place). The test's premise — a
bare system handler successfully drafting email — is exactly the case
#3727 closed. #3727's own commit message names this same fix for sibling
fixtures in `handlers_actions_test.go`/`handlers_event_test.go` (routed
through a new `ownedFiring()` helper, package `automation`), but that
helper only reaches tests calling `ApplyActions`/`Apply` directly in the
`automation` package — it does not cover this integration test's
[...]

PR #3738, opened 2026-09-02
main is red: the draft probe still expects a draft an owner now gates

dungbanguyen:
`TestDraftEmailFiringLandsTheComposedDraftOnTheRunRecord` fails on the tip. It passes at `62faafd39` and fails at `5f1b3ceda`; the change between them is https://github.com/margince/margince/pull/3727.

```
workflow task11a_draft_email_probe: this automation drafts an email but has no owner;
a drafted message goes out under one person's name and is released by that person,
so assign an owner before enabling it
```

## Why the probe cannot pass as written

That refusal shipped deliberately — `MissingDraftOwnerError` names `owner_id` as the field to set, and its own doc says a catalog template seeded without one "drafts nothing".

Every **system-registered** firing is ownerless by construction: `engine.HandleEvent` builds its event from a `kevents.Envelope`, which carries no owner, and `OwnerID` is populated only from an automation *instance* (`timescan.go`). The probe is registered with `RegisterSystemWorkflow`, so no arrangement of the test gives it one.

The test asserts what shipped instead: the firing fails, and the returned error names the missing owner. The run row records which automation stopped and not yet why — asserted as it is rather than asserted away, because that gap is worth someone's attention.

## What this costs, stated rather than dropped

`workflow_run.applied` was asserted against a real database **here and nowhere else**. The regression it guarded is a real one the original author called out: `recordApplyOutcome`'s staged arm wrote only `detail` for its whole life, so an automation could report having drafted a reply while holding nothing to show for it. That is now unguarded at the database level.

Restoring it needs a firing that carries an owner, which needs the owner question settled. **This PR does not settle it** — https://github.com/margince/margince/issues/3605 carries the two readings:

- `withSendingOwner` documents a system firing as one that "drafts unvoiced";
- `MissingDraftOwnerError` makes an ownerless draft a refusal.

Both sentences are on `main` and they cannot both be true. Whoever settles that should restore the `applied` coverage with it.

## Verification

Whole package against real Postgres, not the named test:

```
ok  github.com/margince/margince/backend/internal/compose/integration  285.375s
```

`make check-backend` green. Diagnosed on a clean `origin/main` worktree first, so the diff under review is not what broke it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Workflow executions without an assigned owner now fail with a clear missing-owner error.
  * Failed workflow runs are correctly recorded with a `failed` status.
  * Draft email workflows no longer persist drafts or approval requests when execution fails.

### DIFF
--- a/backend/internal/compose/integration/workflow_action_executors_integration_test.go
+++ b/backend/internal/compose/integration/workflow_action_executors_integration_test.go
@@ -17,6 +17,7 @@ package integration
 import (
 	"context"
 	"encoding/json"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


tinnhlt-gra:
Filed the same root cause independently as #3736, with a fix in #3741 that takes the other direction: routes the probe through the catalog/instance path (RegisterWorkflow + a seeded owned automation instance) instead of asserting the refusal, so it keeps the `workflow_run.applied` coverage you correctly flag this approach drops. Not arguing for mine over yours — yours is simpler and explicitly defers the real #3605 question rather than papering over it, which is a defensible call. Flagging so whoever merges doesn't do both.

tinnhlt-gra:
#3741 merged (routes the probe through the catalog/instance path, keeping the workflow_run.applied coverage). This should be superseded now — closing is your call since it's your PR, just flagging so it doesn't sit stale.

dungbanguyen:
Superseded by https://github.com/margince/margince/pull/3741, which is the better fix — and I am glad it landed rather than mine.

Mine asserted the refusal and accepted that `workflow_run.applied` lost its only real-database assertion, because I could find no way for a system-registered firing to carry an owner. #3741 routes the probe through the catalog/instance path instead, so the firing HAS an owner, the compose-and-park contract still runs, and the `recordApplyOutcome` staged-arm regression stays guarded. That is the coverage my PR would have given up.

Verified on a clean `origin/main` worktree at `dc471bd9a`: `TestDraftEmailFiringLandsTheComposedDraftOnTheRunRecord` passes and the `applied` assertions are intact.

The design tension I raised on https://github.com/margince/margince/issues/3605 is now answered by the code as well: a system firing routed through an instance carries an owner, so "drafts unvoiced" is the stale sentence rather than the rule. Worth a follow-up to correct `withSendingOwner`'s comment, which still says otherwise.

PR #3757, opened 2026-09-02
Wave 4: a vCard naming no company can be staged, and its import stops retrying forever

### LINKED ISSUE #3410: A vCard naming no company can never become a review: the identity asserts organization, the payload omits it, and the retry it advises fails identically for ever
## Observed

Measured on `cfbbb3fdc`, extending `UAT-VCARD-1`.

A near-match vCard carrying no `ORG` line can never be staged as a review. The import reports it as:

> …the review could not be queued; import the card again to retry

and re-importing fails identically, every time, for ever.

## Cause

Two correct-looking decisions that contradict each other.

**The identity asserts `organization`.** `compose/vcardcreateproposal.go:114` explains why, and the reasoning is sound:

> Identity is the card's own addressing — who this card claims to be — not the whole payload: a re-export that reorders phone numbers or tweaks a title is still the same question… **Organization joins the identity as the discriminator for the card**

**The payload omits it when empty.** `compose/vcardcreateproposal.go:62`:

```go
Organization string `json:"organization,omitempty"`
```

**Staging refuses the mismatch, deliberately.** `approvals/staging.go:105-113`:

```go
// Membership is checked separately from value: a missing key and an
// explicit null both read back as a nil any, but jsonb containment
// treats {"k":null} as present-and-null — an identity asserting a field
// the payload omits would pass and then never containment-match.
got, ok := payload[field]
if !ok || got != want {
    return nil, fmt.Errorf("crmapprovals: Identity field %q is not carried by ProposedChange", field)
}
```

So a card with no company produces `organization: ""` → `omitempty` drops the key → the identity asserts a field the payload does not carry → refused. Each of the three is defensible on its own; together they make an ORG-less near-match unstageable.

**Measured directly:** the same probe card refused with no `ORG`; adding one `ORG` line staged successfully.

## Cost

The person is told to do something that cannot work. "Import the card again to retry" is advice that fails identically on every attempt, and nothing tells them why or that retrying is futile — so the natural response is to retry, then to assume the file is at fault.

Cards without a company line are not an edge case. Personal contacts, cards exported from a phone, and anything from an individual rather than a business commonly carry none.

The lost contact is a *near-match* — one the product judged similar enough to an existing person to want a human's decision. So the cards this silently drops are exactly the ones where getting it wrong matters most.

## Why our own suite recorded it as a mystery

`UAT-VCARD-1` had carried a note that one fixture card failed staging with the cause not established. It is this. The scenario is green either way, because it asserts the reported outcome rather than the reachability of a review — so the failure was visible in the report's text and invisible to every assertion.

## Not asserted

We deliberately did **not** add an assertion pinning the current behaviour. Baking in "an ORG-less card cannot be reviewed" would make a fix fail our suite for doing the right thing, and this is a defect we expect to be fixed rather than a state to guard. It is written up in `SCENARIO.md` and the report instead, replacing the old "cause not established" bullet.

Visible in the scenario's refreshed evidence: Priya Raghunathan (no `ORG`) carries the retry sentence; Marisol Etxebarria (with `ORG`) does not.

---
Found by [margince-qc](https://github.com/margince/margince-qc) while adding a guard for margince#3407 — the guard is unrelated, and this turned up because the new fixture card needed an `ORG` line to stage at all, which raised the question of why.

tinnhlt-gra:
## Summary

margince#3410 bundled two fixes to the vCard import lane, per the QC plan's own scoping (same code path).

### The rule that bent: a card's identity may assert an empty string that the payload doesn't literally carry

Three individually-correct decisions contradicted each other:
1. The staged identity always asserts `organization` — including as `""` — because organization discriminates two same-named cards from different companies (`vcardcreateproposal.go:114`).
2. The staged payload struct tagged `Organization` as `json:"organization,omitempty"`, so an org-less card dropped the key entirely.
3. The approvals engine's containment check deliberately refuses an identity asserting a field the payload doesn't carry (`staging.go`) — a real invariant protecting every other `Stage` caller in the codebase from a silent future containment-match failure.

**The call made**: bend (2). `omitempty` was serving a genuine display purpose ("empty fields are omitted so the card carries facts, not blanks"), but that purpose is already satisfied at the *display* layer — `resolveDisplay`/`displayValue` in `approvalkind.ts` already treats an empty string identically to an absent key (`if (raw.trim() === "") return null`). So dropping `omitempty` from just this one field (the only payload field that also doubles as an identity discriminator) fixes the mismatch with zero UI change: an org-less card's review still shows no Company row.

I didn't touch the staging engine's containment check (3) — it's a load-bearing invariant shared by every `Stage` caller in the tree, not something specific to vCards, and weakening it would reopen exactly the class of bug it exists to prevent.

### Bundled: `vcard_ingest`'s retry ceiling

Same job's insert site, same issue. `vcardIngestInsertOpts` carried no `MaxAttempts`, so a deterministic failure (this bug, or any other) rode River's own 25-attempt exponential ladder for an answer that never changes — the job parses local files with no model call and no network read, so a repeat attempt against the same bytes can only repeat the same verdict.

`api/jobs.yaml`'s `max_attempts` field turned out to be the wrong place for this: that field only governs `opts_owner: fan_out` jobs (`compose.workspaceSweepOpts` is its only reader, per the YAML's own comment), and `vcard_ingest` is `opts_owner: caller`. The bound belongs at the insert site instead — matched the existing pattern in `commsjobs.go`/`geocodejobs.go` (a named `const` on the `*river.InsertOpts`).

## Test plan
- [x] TDD: `TestACardNamingNoCompanyCanStillBeStaged` (integration) written RED-first against the real approvals engine, now green; `TestVCardIngestDeclaresABoundedRetryLadder` (unit, 100% coverage on the new line)
- [x] `go build ./...`, `go vet ./...`, `gofmt -l .` clean
- [x] `make test-it DIR=backend/internal/compose` green (real Postgres), full existing vCard suite unaffected
- [x] `craft static --strict`: 0 blocker/major/minor
- [x] Live UAT: uploaded an org-less near-match card through the actual Contacts → Import cards flow on a real dev stack, confirmed the review appears with no phantom Company row, and approved it into a real contact — video attached below
- [x] No AI prompt touched, no aicert re-run needed
- [x] `make check` green before push, rebased onto latest `origin/main`

https://claude.ai/code/session_013oooEFASw2gFLzM2uAb64A

---
## Summary by cubic
Fixes margince#3410: previously, org-less vCard near-matches could not be staged because their empty organization field was omitted; now they can stage without showing a Company row. `vcard_ingest` now stops after 3 attempts instead of River’s default 25, limiting repeated retries for deterministic failures.

- Leaves the approvals containment check unchanged and preserves the empty organization value in the staged payload.
- Adds integration coverage for staging and unit coverage for the retry limit.

<sup>Written for commit 58c4b117a1ecce5a15ee81eeb0416af887af470f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/3757?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * vCard imports with missing organization information can now be staged for review successfully.
  * vCard import jobs are limited to three retry attempts, reducing repeated processing of failed imports.
  * Staged vCard proposals consistently include organization information, including when the value is empty.

### DIFF
--- a/backend/internal/compose/vcardcreateproposal.go
+++ b/backend/internal/compose/vcardcreateproposal.go
@@ -58,8 +58,12 @@ type vcardCreateProposal struct {
 	// The rest of what an approval RELEASES, flattened for the card: the
 	// decider is shown every field the create will write, or they are
 	// approving more than they were asked. Empty fields are omitted so the
-	// card carries facts, not blanks.
-	Organization string `json:"organization,omitempty"`
+	// card carries facts, not blanks — Organization is the one exception:
+	// it also joins the identity below, and the engine's containment check
+	// refuses an identity asserting a field the payload omits. A card
+	// naming no company still displays as nothing; omitempty just cannot
+	// be the reason it does, or that card can never be staged at all.
+	Organization string `json:"organization"`
 	Title        string `json:"title,omitempty"`
 	Phones       string `json:"phones,omitempty"`
 	URL          string `json:"url,omitempty"`
--- a/backend/internal/compose/vcardingest.go
+++ b/backend/internal/compose/vcardingest.go
@@ -53,6 +53,15 @@ import (
 // past a limit of five.
 const vcardIngestMaxCards = 5
 
+// vcardIngestMaxAttempts bounds the retry: parsing a handful of local files
+// and writing at most vcardIngestMaxCards people touches no model and no
+// network (api/jobs.yaml's timeout says so), so a failure here is either a
+// transient database hiccup, worth one or two more tries, or a defect in the
+// card or the code — neither of which a 25th identical attempt fixes. River's
+// own default would keep retrying a deterministic refusal for hours across an
+// exponential ladder, for no different an answer than the first attempt gave.
+const vcardIngestMaxAttempts = 3
+
 // vcardIngestInsertOpts is the trigger's insert, spelled here beside the worker
 // whose queue and attempt cap it names.
 //
@@ -68,8 +77,9 @@ const vcardIngestMaxCards = 5
 // their own.
 func vcardIngestInsertOpts() *river.InsertOpts {
 	return &river.InsertOpts{
-		Queue:      aiCaptureQueue,
-		UniqueOpts: river.UniqueOpts{ByArgs: true, ByState: activeSweepStates},
+		Queue:       aiCaptureQueue,
+		MaxAttempts: vcardIngestMaxAttempts,
+		UniqueOpts:  river.UniqueOpts{ByArgs: true, ByState: activeSweepStates},
 	}
 }
 
--- a/backend/internal/compose/vcardingest_test.go
+++ b/backend/internal/compose/vcardingest_test.go
@@ -1,0 +1,21 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+package compose
+
+import "testing"
+
+// A mailed card's import touches no model and no network, so a failure is
+// either a transient database hiccup or a deterministic refusal — a
+// malformed card, a bug — that answers identically on every attempt. River's
+// own default would retry the deterministic case 25 times across hours of
+// exponential backoff for no different an answer than the first attempt gave.
+func TestVCardIngestDeclaresABoundedRetryLadder(t *testing.T) {
+	got := vcardIngestInsertOpts().MaxAttempts
+	if got != vcardIngestMaxAttempts {
+		t.Fatalf("enqueued MaxAttempts = %d, want the declared ladder %d", got, vcardIngestMaxAttempts)
+	}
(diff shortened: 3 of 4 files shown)


tinnhlt-gra:
UAT: uploading an org-less near-match vCard through Contacts → Import cards. The card stages cleanly, shows in the review queue with no phantom Company row, and approves into a real contact.

vcard-org-less-review

Issue #3793, opened 2026-09-03
The weekly review shows "Promises kept" and "Promised, delivered" on one screen, over different quantities, in the same {n} of {m} shape


tinnhlt-gra:
## Observed

Measured on core `762cdb9cf`, on `#/home` → **Weekly** as an ordinary Member
seat, 2026-09-03. Read off the screen, in this order, six lines apart:

```
PROMISES KEPT            0 of 0        <- the outcomes strip (StatCard)
WON                      37
LEADS ANSWERED IN TIME   0 of 0
MEETINGS WITH A NEXT STEP 0 of 0
CARRIED OVER             0

Promised, delivered      0 of 0        <- the workings list (dl)
Moved                    161
Lost                     18
You decided              0 yes · 43 no
```

They are two different quantities:

| on screen | i18n key | counts |
|---|---|---|
| **Promises kept** | `home.weekly.promisesKept` | `commitments_kept / commitments_due` — commitments a rep wrote into their **weekly plan** and settled (`weeklyplan.CloseWeek`, `modules/weeklyplan/closeweek.go:49`) |
| **Promised, delivered** | `home.weekly.promised` | `tasks_done / tasks_due` — **tasks** with a due date in the week (`weekly_review.tasks_due/tasks_done`) |

Both render through the same `home.weekly.ofDue` → `"{done} of {due}"`, so they
are the same shape as well as nearly the same words. Nothing on the page names
the difference: no caption, no tooltip, no unit.

`28d085762` split the ten figures into five outcomes (the strip) and five
workings (the list below it), which is a good split for the reason its own
comment gives. It put these two on opposite sides of it, which is what brought
them onto one screen a few lines apart.

The vocabulary is contradicted a tab away, too. Home → **Morning**, in the
Promises & tasks rail:

> Promises made in conversation are not tracked yet — only tasks are listed here.

So on one tab "promise" is explicitly not-a-task, and on the next tab the
headline figure called *Promises kept* is not the task one and the task one is
called *Promised, delivered*.

## Cost

Small, and entirely about reading rather than arithmetic — both figures are
correctly computed and neither is wrong.

A rep or their lead reads the strip first; it is the read-across row the split
exists to protect. **Promises kept** is the leftmost card in it, and on a seat
that keeps no weekly plan it reads `0 of 0` for ever while the figure that
actually reflects the week's delivered work sits demoted in the list below under
a name one word away. The likely reading of `0 of 0` at the head of the strip is
[...]

Issue #3791, opened 2026-09-03
The Worklist's summary line counts the first page and the whole day in one sentence — "11 urgent" over 118


tinnhlt-gra:
## Observed

Measured on core `762cdb9cf` (`MARGINCE_BUILD_REVISION` read off the running
API), on `#/worklist` as an ordinary Member seat, 2026-09-03.

The line above the queue is one sentence with five figures in it
(`worklist.summary`, `frontend/src/i18n/en.ts:7746`):

```
11 urgent · 5 due · 14 in play · 0 routine — 165 in all
```

Four of those figures describe the **first page**. The fifth describes the
**whole day**. Nothing on the screen says so, and the reader has every reason
to read the sentence as a breakdown of the total, because that is the shape of
the sentence.

Read off the same screen, in the same session, pressing **Show more** six times
until the button disappeared:

| | the line says | the day actually holds |
|---|---|---|
| urgent (`level <= 2`) | **11** | **118** |
| due (overdue) | 5 | 5 |
| in play (`level 3-5`) | **14** | **27** |
| routine (`level >= 6`) | **0** | **9** |
| in all | 165 | 165 (154 rows, 2 of them folded groups standing for 13) |

`11 + 14 + 0 = 25`, which is the page size. The four figures do not move as the
reader pages — after six presses and 154 rows on screen the line still read
`11 urgent · 5 due · 14 in play · 0 routine — 165 in all`.

The completeness line one tier down *does* track the reader (`25 shown` became
`154 shown · 1 sources have more`), and so do the readings strip and the filter
pills, which is what makes the lead sentence's stillness read as a fact about
the day rather than as a fact about one response.

### Why

`summarize` is handed the page, deliberately:

```go
shown, more, reached := pageFrom(rows, limit, cursor)
ordered := rankAll(stampAsOf(shown, day.AsOf))
...
Summary: summarize(ordered, materialBarOf(day, s.money)),
```

(`backend/internal/compose/attention/worklist.go:268-306`), and its own comment
says why:

> Every figure counts items the queue actually CARRIES, so a number above the
> list and the rows below it cannot disagree — which is the defect the lane feed
> shipped, reporting a twelve-item page as a total.

That is a good rule and it is not the problem. The problem is that the browser
then prints those page-scoped figures beside a **day-scoped** total it computes
itself:

```ts
function dayTotal(day: Worklist): number {
  return day.counts.reduce((total, count) => total + count.considered, 0);
}
```

[...]

Issue #3700, opened 2026-09-02
A read-only seat is offered "Plan my week" and the refusal reaches no part of the screen


tinnhlt-gra:
`PlanSection` draws "Plan my week" for every reader whose plan read came back empty, including a `read_only` seat that the server refuses. Pressing it does nothing and says nothing.

## Observed

Measured on core `6dab0e508`, on `#/home` as `uat-readonly@demo.test` (`read_only`).

The panel's empty state draws the button unconditionally — `plan.data === null` is the only condition, and no permission is consulted (`screens/brief.plan.tsx:158-163`):

```tsx
{plan.isSuccess && plan.data === null && (
  <PanelBody>
    <p>{t("plan.none")}</p>
    <Button onClick={() => start.mutate()} disabled={start.isPending}>
      {t("plan.start")}
    </Button>
  </PanelBody>
)}
```

The server refuses it:

```
POST /v1/weekly-plans/current   (read_only)
403 {"code":"permission_denied","detail":"weekly_plan.create: permission denied"}
```

because `read_only` holds `{"create": false, "read": true, "update": false, "delete": false}` on `weekly_plan` (`migrations/core/1788308622_weekly_plan_rbac.up.sql`), which is deliberate and right.

`useStartWeeklyPlan` has an `onSuccess` and no `onError` (`screens/weeklyplan.queries.ts:84-98`), so the rejection reaches nothing that draws. On screen, 2.5s after the press: the panel still reads "You have not planned this week yet.", the button is still there, and the page carries no `[role="alert"]` and no `[role="status"]` — measured, count 0.

So the reader presses a button, nothing happens, and the product never tells them why.

## Cost

Small and narrow — one seat, one button — but it is the failure mode #3573 was filed for on the account page ("The account page offers 'Log activity' and 'Add task' to a seat that may not create activities"), and it is the same shape: a control offered without consulting the permission the server will apply. A read-only auditor concludes the app is broken rather than that they may not do this.

## Workaround

None. The button is inert for this seat.

## Suggested fix

Either gate the control on the reader's `weekly_plan.create` grant — the panel already receives the authorization map through `/v1/me` — or give the mutation an `onError` so the refusal is drawn. The first is better: a control that exists to fail is worse than no control.

---
[...]

Issue #3690, opened 2026-09-02
Overlay mode: a live-looking write control can silently do nothing


tinnhlt-gra:
Surfaced during review of PR #3663 (cubic AI review), two instances of the same shape: a control renders as pressable in overlay mode, but the write it starts is refused entirely — silently, with no explanation — because the surface backing it (\`LogActivityAction\`) returns \`null\` in overlay rather than opening.

1. **\`companyheader.tsx\`'s \`WriteEmailAction\`** — not gated on overlay at all, unlike its sibling \`LogActivityAction\` in the same header strip. If email composition is genuinely unavailable through the overlay mirror (unconfirmed — needs someone who knows the overlay/mirror contract to say), it should hide the same way the log-activity verbs do; if it IS available, no fix is needed there and the finding is moot.
2. **\`persontoday.tsx\`'s \`ActionVerb\`** (the moment card's recommended/secondary actions) — generic across every \`PersonMomentAction\` kind, driven purely by \`action.state\` (\`available\`/\`will_confirm\`/\`blocked\`), which the *backend* computes with no concept of overlay mode. A moment recommending \`activity_log\` or \`task\` in overlay renders as a live, pressable button whose destination drawer (\`PersonActivityDrawer\`) silently renders nothing when pressed. This is pre-existing for \`activity_log\` — margince#3572's PR only added a second surface (\`task\`) with the identical shape, it did not introduce the pattern.

## Why filed rather than fixed

Both need a product/design call this issue doesn't try to guess:

- Does overlay actually refuse email composition, or was `WriteEmailAction` correctly left ungated? (Determines whether #1 is a bug at all.)
- What should `ActionVerb` show for an action it cannot honour for an overlay-only reason the server never sent — reuse the "blocked" treatment with new client-side copy, or something else? `readiness()` currently only ever echoes the server's own `blocked_reason`; teaching it an overlay-specific reason with no server signal behind it is a new client-side authority over "why is this refused" that the rest of the file deliberately keeps server-side.

## Where

- \`frontend/src/screens/companyheader.tsx\` — \`WriteEmailAction\` mount, \`CompanyPrimaryActions\`
- \`frontend/src/screens/persontoday.tsx\` — \`ActionVerb\`, \`readiness()\`

dungbanguyen:
## Decision, 2026-09-02: give the server the overlay signal — do not teach the client a new authority

Re-triaged against `origin/main` `4fd58793`.

### The ruling

[...]

PR #3805, opened 2026-09-03
Frontend QC batch: schedule times, plan permissions, worklist scope, peers, plurals, reach

### LINKED ISSUE #3792: Home's "Today's schedule" draws every meeting with a blank time — the panel reads occurred_at, the lane sets due_at
## Observed

Measured on core `762cdb9cf` (`MARGINCE_BUILD_REVISION` read off the running
API), on `#/home` as an ordinary Member seat, 2026-09-03.

A meeting was booked for later today through the product's own API:

```
POST /v1/activities  {"kind":"meeting","subject":"… schedule rail probe",
                      "occurred_at":"2026-09-03T03:40:00Z","source":"manual"}
201 → occurred_at 2026-09-03T10:40:00+07:00
```

The **Today's schedule** panel then drew it with an empty time cell:

```
Today's schedule
              qcm-… schedule rail probe   [Needs prep]
```

`.rail-schedule-when` read `""` — the row is there, the badge is there, the
time is not. The screenshot is one row with a blank gutter where the time
belongs.

The time is nowhere else on Home either. The same meeting in the **Do next**
card is described as *"starting shortly"*, which is a relative phrase, not the
10:40 the meeting is at. So a rep reading Home is told they have a meeting today
and is never told when.

### Why

`whenOf` reads `occurred_at`:

```ts
function whenOf(item: WorklistItem, locale: Locale, zone: string): string {
  return item.occurred_at ? formatTimeOfDay(item.occurred_at, locale, zone) : "";
}
```

(`frontend/src/screens/home.schedule.tsx:127-131`, added by `9d2d0ba13`), and
its docstring says `occurred_at` "is when it is".

The meeting lane never sets it. `meetingItem` puts the start on `due_at` and
says why:

```go
DueAt: &starts,
```

(`backend/internal/compose/attention/meeting.go:32-42` — "the start time as
`due_at` because that is what a reader is racing"). Nothing downstream copies
it across: `stampAsOf` sets a private ranker field, not the item's
`occurred_at`.

Confirmed on the wire for that row:

| field | value |
|---|---|
| `source` | `meeting` |
| `due_at` | `2026-09-03T10:40:00+07:00` |
| `occurred_at` | **absent** |

So the fallback branch is the only branch a meeting row can take, and it draws
nothing by design — the design just guarded the wrong field.

### Why no test sees it

`home.fixtures.ts:254-265` builds the schedule fixture with neither
`occurred_at` nor `due_at` on the meeting rows, so the panel's own tests render
the empty-time branch and assert around it. There is no fixture in which a
meeting has a time, which is why a panel that can never show one is green.

## Cost

The panel is a list of times and it has no times in it. That is the whole of
what "Today's schedule" is for; without the time it is a duplicate of the
meetings count already in the readings strip.

Narrow, though, and worth saying so: the meeting itself is not lost, it is
ranked correctly, the *Needs prep* badge is right, and the link opens. It is one
missing field on one rail panel, and the exact time is still available by
opening the meeting. Nothing is misfiled and no number is wrong — a rep is
simply not told when to be somewhere, on the screen that exists to tell them
what their day looks like.

## Workaround

Open the meeting from the schedule row; the activity carries its own
`occurred_at`. Or read the Worklist row's deadline rather than Home's rail.

---
Found by [margince-qc](https://github.com/margince/margince-qc), following a
static lead about `9d2d0ba13`'s new schedule rail that nobody had been able to
confirm because no meeting had reached a test seat's day that week. Booking one
through `POST /v1/activities` and opening `#/home` settled it in two minutes.

luit-nfq:
The small, self-contained frontend issues from the last QC pass, in one branch. The larger ones (#3614, #3450, #3690, #3587) are deliberately left for their own PRs.

## What changed, per issue

**#3792 — Home's "Today's schedule" drew every meeting with a blank time.** `whenOf` guarded on `occurred_at`, which is when something *happened*; a meeting still ahead of the reader has no answer for that, so every row took the empty branch and a rail whose whole job is to say when to be somewhere printed no times at all. The start is `due_at` — the contract's own field for "when this is due, or when the meeting starts", and the one the meeting lane fills. The fixture's meetings carried neither field, so the panel's tests rendered the empty branch and asserted around it; giving them a start is what makes the gap expressible.

**#3700 — a read-only seat was offered "Plan my week" and the refusal reached nothing.** Every control in the panel is now bound to the grant it asks for, and the two differ: opening a week is `weekly_plan.create`, everything after it `update`. A refused seat keeps the plan it may read, loses every write affordance, and is told once why rather than on each row. The panel's tests stubbed `/me` with the default fixture, which grants nothing at all — so a control drawn without consulting the reader's grants passed them.

**#3519 — the rebuilt Network tab placed no peer, so `Record: works with` reached no screen.** Peers get their own lane, under their company in the right column. `Record: works with` is the only writer of a `works_with` edge in the product and was gated on `suggest_edge`, which the server sets on peer nodes alone — so it had no reachable call site at all, and the `relationship.create` gate on it had become unobservable from any screen.

**#3793 — "Promises kept" and "Promised, delivered" over different quantities.** Now "Plan commitments kept" and "Tasks delivered", in all three locales, leaving the word "promise" to the Morning rail that already reserves it. Each name is pinned to its own quantity on a fixture whose two fractions differ, and no weekly reading may spell "promis" again. The weekly mail's copy moved with it, which the mail-copy parity gate caught.

**#3791 — the Worklist's summary counted the first page and the whole day in one sentence.** Both halves in one change, because fixing either alone is worse than neither. The server counts every figure over the assembled day — the same snapshot `counts` and `reach` are taken from — and the browser prints `summary.total` instead of deriving a second answer; `dayTotal` is gone. The paging test's fixture reported `total: 1` over a one-row page, so it could not express the difference it was named for.

**#3589 — a rep can log a call, and arrive ready to.** The API has accepted `kind: "call"` all along and no frontend surface could send one. It is in the picker for everyone now, and `#/leads/<id>?action=call` opens the lead's standing composer on that kind and brings it into view; the composer is the lead's own card, so the status-ladder refresh it already owns fires for a call logged this way. In overlay, an address that asked for the composer is answered rather than ignored.

**#2964 — a count interpolated into a single key printed "1 steps".** The four keys a reader can actually reach at 1 become `_one`/`_other` pairs through `usePlural`. `personcards.loops.test.tsx` asserted `"overdue 1 days"` verbatim, so the defect had a test holding it in place.

**#3304 — four capabilities shipped with no way to reach them.** The four instances are closed (`nav.ts` now carries `worklist`), so what this adds is the gate the issue asked for: a screen the router answers is reachable if the nav carries it or a shipped module links to it — an `#/…` address or a `screen:` property, the latter followed through a constant. Six screens are genuinely entered from outside the app and say so beside their own entry in `router.tsx` with a reason, the way `craft:ignore` and `plural-rule:allow` do.

## Fixed along the way

- **An unfilled field's placeholder was illegible.** `.fieldgrid-value [data-empty]` painted its "Add …" text in `--textTertiary`, a *mark* tone: 2.49:1 against the card, a serious WCAG AA failure on every record carrying an unfilled field. `--textMeta` is the role the contrast law gives all 11–13px meta text. Found by sweeping this branch's routes with axe.
- **Two type errors the composed build catches and `tsc --noEmit` does not** — `SurfaceState`'s required `emptyLabel`, and a grant map typed as `Record<string, string[]>` where `GrantSpec` keeps a misspelled object a compile error.

## Two issues that do not reproduce

- **#1250 (`priority: critical`) — fresh org triggers a Shell-wide render-loop crash.** Walked it on a throwaway worktree with its own empty stack: bootstrap admin, first login, onboarding gate satisfied, then every route both by address and by clicking the rail. No crash, no "Maximum update depth exceeded", no "This view stopped working", and zero console errors beyond #2907's 501. **Recommend closing.**
- **#3675 — a react-query cache leaks between App suites.** The pair the report names passes; all six App suites pass together sequentially in one process; the whole 505-file suite is green. The file's `KNOWN GAP` comment told the next reader to read a real failure as a known leak, so it is replaced by a case: the app is mounted, unmounted and mounted again on a different record, which is the shape a surviving cache would fail. **Recommend closing.**

## Left out, and why

- **#2964's gate half.** Widening `one-plural-rule.test.ts` to see a key with a count poured into it is its own change: swept over `en.ts`, 51 keys interpolate a count beside a bare plural noun with no sibling, and most need their call site read one at a time rather than a blanket conversion.
- **#3589's caller.** Nothing links to `?action=call` yet. `WorklistMove.action` carries only `draft_reply`, so a worklist row cannot offer a call without the server saying it recommends one — a contract change and a product call, not wiring.
- **The mocked axe sweep does not cover record subroutes.** `CORE_SCREENS` in `e2e/ac.spec.ts` is top-level screens, so the person Network tab this branch changes is swept only by hand. Teaching `e2e/seed.ts` a `/people/{id}/graph` payload would close it.
- **`backend/internal/compose/attention/worklist.go` is four lines under the 500-line ceiling.** It wants splitting by concept before the next change to it.

## Verification

Against the seeded stack in a real browser (Chromium), as admin and as a purpose-made `read_only` seat:

| | measured |
|---|---|
| Worklist header | `1 urgent · 1 due · 11 in play · 5 routine — 17 in all` — the bands sum to the total, and the page carried 5 rows |
| Home schedule | the booked meeting draws `14:43`, from a payload carrying `due_at` and no `occurred_at` |
| Home weekly | `PLAN COMMITMENTS KEPT` in the strip, `Tasks delivered` in the list |
| Plan panel, read-only seat | the read-only sentence, the week still reported, zero buttons in the panel |
| Network tab | peer node placed under "Who they talk to", reachable by arrow keys, `Record: works with QC Peer Suggested` offered, pressed, and the offer gone on the next read |
| Lead record | `?action=call` opens the composer on Call; the same lead without the param and a *different* lead both open on Note |

States, keyboard and viewports: 390px and desktop with no horizontal overflow on any touched route; the map's focus follows the arrow keys with a visible 2px ring, Enter opens the detail, Escape clears it; light and dark.

Accessibility: axe WCAG 2.2 AA over every touched route (person network, worklist, home morning, home weekly, the lead record), at 1440px and 390px, in light and dark — **zero serious or critical violations** after the placeholder-contrast fix above.

Console: clean apart from two pre-existing lines — the signed-out `/me` probe's 401, and #2907's `EmbedReindexStatus` 501, which reproduces on a vanilla install in two places.

## Gates

| gate | result |
|---|---|
| `make check` | pass |
| `make frontend-check` | pass — 505 files, 6032 tests |
| `make fe-uat` | pass — 133 stories captured, 0 missing, 0 unresolved (`.tmp/fe-uat/manifest.json`) |
| `make native-controls` | pass — 35 tests |
| `make frontend-e2e` | pass — 224 passed, 36 skipped, including the axe sweep |
| `make fe-clock-drift` | pass — 505 files at +200 days |

New stories, all captured by fe-uat: `Schedule`, `ScheduleUndated`, `ScheduleClear`, `Promises`, `Plan`, `PlanNone`, `PlanReadOnly`, `ObservedPeers`, `ObservedPeersNoWrite`, `OpenedOnACall`, `LeadArrivedToLogACall`, `LeadCallRefusedInOverlay`.

Closes #3792, #3700, #3519, #3793, #3791, #3589, #3304.

## Review round

Four findings, each a case the tests did not hold. All four now have a test that
fails without the fix.

1. **The composer read its kind once.** `initialKind` was consumed by a
   `useState` initializer, so a link naming a kind — pressed on the record the
   reader is already on — changed the address under a form that never
   unmounted, and handed them the composer they asked for set to the wrong
   verb. The prop is renamed `askedKind`, for what it is, and the form follows
   it: only `kind` moves, exactly as far as the reader's own Type control moves
   it, so a half-typed subject survives. Fixed in the form, not at the lead
   call site, so the person page's `?log=task` address is fixed with it.
2. **The plan panel asked for the union of two grants.** The grant a reading
   needs is one of them: a week with no plan asks for `weekly_plan.create`, a
   week with one asks for `update`. A seat holding create alone read an
   existing plan with every control gone and no sentence saying why. The
   decision moved into `statesReadOnly`, which also waits on the week, and both
   directions are now covered.
3. **A negative case could pass before the seat answered.** Every capability
   hook reads false while `/me` is in flight, so an ungranted seat and an
   unanswered one draw the identical page — the assertion would have kept
   passing had the control stopped consulting the grants. The stub signals when
   `/me` answered and the case waits for that plus a render.
4. **The reach gate counted a mention as a door.** A string literal has to
   BEGIN with the address, which is what every href here looks like; prose
   naming a screen is not a door, and reading it as one would have excused the
   screen most likely to have none — the one the copy tells you to type.

## Not green, and not this branch

`main` is red on the integration lane independently of this PR, so this PR's
integration shards fail with it. Verified against an `origin/main` worktree
before this branch existed, and the latest `main-health` run fails the same
shards:

- consent / double-opt-in / confirm-mail — DOI issuance answers `409 "this
  installation cannot yet mail one"` where the lane expects `422`/`404`, and
  `confirmmail_integration_test.go:82` reports `delivered = false with a working
  relay wired`.
- `toolschema_conformance_integration_test.go:228` — `forecast_movement` and
  `forecast_readings` are registered but never invoked in the lane.

Both areas are backend work outside this change's scope (frontend, plus the
attention worklist summary). Nothing here touches consent, mail or forecasting.

## Summary by CodeRabbit

- **New Features**
  - Added call activity logging from lead links, with messaging when calls aren’t supported in overlays.
  - Added read-only plan views with explanatory messaging.
  - Added peer relationships and a dedicated peer lane to person network views.
  - Added email activity drawers from activity timelines.
  - Meeting schedules now display planned start times in the viewer’s time zone.
  - Added forecast assurance and exception-resolution capabilities.

- **Improvements**
  - Worklist totals and priority counts now represent the full assembled day across pagination.
  - Improved singular/plural wording and weekly “Tasks delivered” terminology.

- **Bug Fixes**
  - Corrected data isolation when switching people after remounting the app.


luit-nfq:
Second review round addressed — all three findings, each with a case that fails without the fix.

**`brief.plan.tsx:142` — retained edit state.** Real: `editable` can go false under a reader mid-stage (the weekly job closes the week, or the grant is withdrawn), and the rows lose their checkboxes when it does while the Save footer and the mounted add form stayed. Both reached a write the server refuses, over ticks nobody can see any more. Both are now gated on `editable`; `takes Save away when the week closes under a staged tick` closes the week under a staged tick via a focus refetch and asserts the footer goes and nothing is sent.

**`reach.test.ts:234` — alias scope.** Real, and in the direction that matters: a tree-wide table of constants lets an unrelated module's same-named `const` answer for a name, which reports a door where none exists and excuses exactly the screen nothing reaches. A name now resolves in the module that wrote it, then through that module's imports (renames followed, bare specifiers refused). This is load-bearing rather than theoretical — three screens in the tree (`scheduled`, `ext`, `x`) are reached through an imported constant, and dropping the import hop reds the gate on all three. Two census cases added: the imported constant either file first including an `as` rename, and an unrelated file's same-named constant opening nothing.

**`reach.test.ts:169` — absolute hash URLs.** Taken. The regex now admits an optional origin, so `"https://app.example/#/planted"` is a door like `"#/planted"`, with a census case. Prose is still not a door: the anchor means `"Open #/planted to see it"` still counts for nothing.

Gates after the change: `make check-fe` pass, `make fe-uat` pass (139 stories), `craft static --strict` clean on the push.

**Still not mergeable, and not because of this branch.** `main`'s integration lane is red on its own: the consent / double-opt-in / confirm-mail family and the SAR export family both fail on an unmodified `origin/main` worktree here — `AssembleSAR: ERROR: column "captured_at" does not exist (SQLSTATE 42703)`, and DOI issuance answering `409 "this installation cannot yet mail one"` where the lane expects `422`/`404` — plus `toolschema_conformance_integration_test.go:228` for `forecast_movement` and `forecast_readings` registered but never invoked. Every other lane here is green (`ci` reports `1 of 9 job(s) did not pass`, and that one is integration). Nothing in this PR touches consent, mail, SAR or forecasting.

luit-nfq:
CI on `9fa1a048c`: every lane green except integration — `ci` reports `1 of 9 job(s) did not pass`, and that one is integration. The 32 failing tests are byte-for-byte the set that fails on an unmodified `origin/main` worktree; nothing in this branch is involved (frontend, plus the attention worklist summary).

Four independent causes on the tip, now all tracked:

- #3828 — the DOI / confirm-mail group. Needs a product decision on the withdrawn endpoint, so it cannot be cleared by a fix here.
- #3818 — the privacy FK visibility gate.
- #3868 (filed now) — `sarsections.go:214` selects `captured_at` from `consent_qualifying_event`, which carries `created_at` and `captured_by` and has never had `captured_at`. One statement fails, `AssembleSAR` returns the error, and the whole SAR/erasure family (~20 tests) fails with it.
- #3869 (filed now) — `forecast_movement` and `forecast_readings` are registered with neither a call in the tool-schema conformance lane nor a named seam.

So this PR is done and cannot go green on its own: merging waits on `main`, and `main` waits on the decision in #3828. Holding here rather than merging past a red required check.

luit-nfq:
Rebased onto `main` (`7bba710be`). Two things changed in the process, both worth naming.

**The schedule commit is gone.** #3879 landed the identical fix — `whenOf` reads `due_at`, for the same reason — so my version was a duplicate and is dropped rather than merged. Main's suite has the positive case (`draws each meeting's start time`); it does not have the negative one this branch carried, that a meeting the server sent with no start is drawn without a time rather than with an invented one. Say the word and I re-add that single case here, or leave it to a follow-up.

**The plan panel's footer merged with main's refusal Callout.** #3853's neighbour work added a `plan.saveRefused` sentence where this branch gated the footer on `editable`. The two halves are now gated differently on purpose: the refusal STAYS, because a week closing is exactly what refuses the save and hiding the reason with the button would leave the reader neither; the Save button goes, because the rows lose their checkboxes with it. The merged footer pushed `PlanSection` past its complexity ceiling, so it is a `PlanFoot` component now. Main's new refusal test had no `/me` stub and drew no checkboxes under this branch's grant-aware panel, so it got the same seat every other case in the file uses.

**The three red things on the tip are not this PR**, and each is tracked:

- `craftsmanship` — `approvals/authority.go` and `mailcopy/catalog.go` are one line over their caps on main. #3899 fixes both. (This branch already split `weeklyLines` for its own diff-scoped push, so half of that overlaps.)
- `integration shard (3/6)` — `merge_tags` is registered with neither a call nor a waiver in the tool-schema census. #3886 fixes it.
- `integration shard (5/6)` — the 360's flat-cost budget is one short of the attachment read #3897 added. #3908 fixes it.

Everything else on this run passed. `make check` is green locally on the rebase.

PR #3825, opened 2026-09-03
createTag stores the description the contract accepts

### LINKED ISSUE #3820: createTag drops the description the contract accepts
`CreateTagRequest` declares `description: [string, 'null']` and `Handlers.CreateTag` (backend/internal/modules/collections/handlers.go) decodes the body and calls `h.store.CreateTag(ctx, req.Name, color)` — the description is never passed and is dropped in silence.

A caller that sends one gets a 201 and a tag without it, with nothing saying so. `updateTag` does honour the field, so a description can be set on a second call; the create path is the gap.

**Why deferred:** found while adding the MCP `create_tag` tool. The tool's schema deliberately omits `description` so it does not advertise a write that does not happen, which keeps the two transports honest with each other. Fixing the HTTP door is a change to that handler and its tests, not to the tool.

**If nobody fixes it:** the published contract keeps promising a field the server throws away.

dungbanguyen:
Closes #3820.

`CreateTagRequest` declares `description`, and `Handlers.CreateTag` decoded the body and then called `store.CreateTag(ctx, req.Name, color)` — the field was dropped in silence, so a caller that sent one got a 201 and a tag without it. `updateTag` already honoured the field; the create path was the gap.

**The change**

- `Store.CreateTag` takes `description` and writes the column. The call sites that coin a word without describing one (`NewTag`, the `create_tag` seam, three integration fixtures) pass `nil` explicitly.
- The empty-string sentinel gets one spelling. `tagUpdateFrom` already read `""` as "no description" — that rule now lives in `tagDescriptionOrNone` and both doors call it, so a tag's description cannot depend on which door wrote it.
- The MCP seam keeps passing `nil`, with a comment saying why: `create_tag`'s schema deliberately omits the field (#3820's own framing), and the comment is there so the next reader does not read it as the same drop being fixed here.

**Tests**

`TestCreatingATagKeepsTheDescriptionItWasGiven` posts a description and then re-reads the tag — the read-back is deliberate, because a handler that echoed its own request body would pass an assertion on the 201 alone without ever writing the column. `TestCreatingATagWithAnEmptyDescriptionStoresNone` pins the sentinel.

Both mutation-checked:

- passing `(*string)(nil)` from the handler again → `the 201 carries description = <nil>, want the text sent`.
- dropping the `*description == ""` arm of `tagDescriptionOrNone` → `description = "", want absent`.

**Verification**

`make check-backend` green. `make test-it DIR=backend/internal/compose/integration/collections RUN='TestCreatingATag|TestRenamingRecolouringAndDescribingATag'` green.

No contract change: `crm.yaml` already published the field, which is why the server throwing it away was the defect.

## Summary by CodeRabbit

* **New Features**
  * Tags can now include optional descriptions.
  * Non-empty descriptions are preserved when tags are created or updated.
  * Empty descriptions are treated as absent.

* **Bug Fixes**
  * Improved consistency in how tag descriptions are stored and returned.


dungbanguyen:
One follow-on this PR creates, noted here so it is not a surprise in review.

`create_tag`'s tool schema in `backend/internal/modules/agents/tools_tagvocab.go` carries this comment, added by #3822:

> No `description`, though CreateTagRequest declares one and update_tag takes one. The REST handler decodes it and never passes it to the store (collections/handlers.go CreateTag), so the field is dropped in silence today — offering it here would advertise a write that does not happen.

After this PR the write **does** happen, so the comment's stated reason is void — and a comment that stops the next author looking is exactly what the rulebook warns about.

Deliberately not touching it here. Whether `create_tag` should now *offer* `description` is a question about the tool surface (a model composing an explanation for a word a workspace will read is a different act from a curator typing one), and answering it in this PR would turn a one-line HTTP fix into a change to the agent surface. #3820's own framing draws that line too: *"Fixing the HTTP door is a change to that handler and its tests, not to the tool."*

If a reviewer would rather the stale half of that comment go now, say so and I will strike the sentence — but the field stays off the tool schema either way until somebody decides it belongs there.

PR #3830, opened 2026-09-03
Classify the send-authorization tables' row-scoped FKs

### LINKED ISSUE #3818: Privacy gate red on main: communication_basis and communication_suppression FKs have no visibility decision
`TestFK_rowScopedTargetsHaveVisibilityDecision` (backend/migrations/schema_fitness_integration_test.go) fails on clean `origin/main`. Six FK columns have no entry in `rowScopedFKDecisions`:

- `communication_basis.lead_id`, `.person_id`, `.source_activity_id`
- `communication_suppression.lead_id`, and its person/activity columns

Those tables arrived in #3815. The gate's rule is that a reference to a row-scoped record is a READ of it, so each FK is either gated at the store (`auth.EnsureLinkTarget` / `EnsureVisible`) or classified with the reason it needs no gate.

**Why I am filing this rather than fixing it.** Each entry in that map is a claim about how a specific write path gates a specific reference, and the map's own comment says its completeness is the invariant. Writing six entries for a subsystem I have not read would produce exactly what the rulebook warns about — a claim that stops the next author looking, which is worse than the silence it replaced. The author of #3815 knows whether those references are gated; I would be guessing.

It only fires in the integration lane against a real database, so `make check` stays green with it broken. Any branch running `make test-it DIR=backend/migrations` is blocked until it is classified.

dungbanguyen:
Closes #3818. **This unblocks the integration lane on `main`** — `TestFK_rowScopedTargetsHaveVisibilityDecision` fails on a clean `origin/main`, so any branch running `make test-it DIR=backend/migrations` is red through no fault of its own.

**What the gate found.** `communication_basis` and `communication_suppression` arrived with the send-authorization record (core `1788407500`, #3815) carrying five FK columns that name a row-scoped record and no entry in `rowScopedFKDecisions`. The ticket named six; the gate names five — `communication_suppression` has no activity column, so its FKs are `person_id` and `lead_id` only.

**What the research found.** The ticket was filed rather than fixed because "the author of #3815 knows whether those references are gated; I would be guessing." The answer turns out not to need that knowledge, because there is nothing to know yet: **nothing in the tree inserts a row into either table.** Every Go statement that names them is one of

- a READ — consent's `liveSuppression`, whose own comment says it is "the only reader in the tree that applies these rows";
- a REMOVAL — privacy's erasure and retention paths, which delete the row or null the subject link (`retentionactions.go`, `erasure_consent.go`, `erasure_leadtwins.go`). None of them ever creates a reference.

So the honest classification is the one this map already spells twice — `conversation_claim.task_activity_id` and `provider_applied_field.person_id` — `PENDING WRITER`: the entry states the obligation on whoever writes the first row, and names the sibling whose shape that writer has to match.

**The five entries**

| column | obligation |
| --- | --- |
| `communication_basis.person_id` | a path that already holds the subject carries that path's gate (`consent_qualifying_event.person_id`'s shape); a caller-named subject takes `auth.EnsureVisibleLive` in the insert's transaction (`lead_manual_signal.lead_id`'s shape) |
| `communication_basis.lead_id` | the lead arm of the same; `communication_basis_one_subject` makes the two exclusive |
| `communication_basis.source_activity_id` | `conversation_claim.source_activity_id`'s obligation — `auth.EnsureActivityContentVisibleLive`, and **live** rather than merely visible, because a basis that outlived its evidence would keep authorizing sends on a message nobody can now open |
| `communication_suppression.person_id` | the subject arm, with the difference written down rather than rediscovered: a hard bounce arrives from a provider and not from a caller, and a system writer that names no principal has no scope to probe |
| `communication_suppression.lead_id` | the lead arm; the target CHECK admits an address-only row, so the probe binds only the arm a writer fills |

The address column is deliberately not classified: an address is a fact about a mailbox and names no row-scoped record, so it is outside this gate rather than exempted by it. That is said in the block comment so the next reader does not look for the missing entry.

**Mutation-checked, both directions**

- deleting the `source_activity_id` entry → `FK communication_basis.source_activity_id -> activity has no visibility decision`
- adding a `communication_basis.no_such_column` entry → `waiver communication_basis.no_such_column matched no subject: delete it, or correct the key if the subject was renamed`

**Verification.** `make test-it DIR=backend/migrations` green (all five tests). `make check-backend` green.

**Known limitation, filed as #3831.** A `PENDING WRITER` entry is a claim that no writer exists, and nothing fails when one appears — the classification just goes quietly false. That is AGENTS.md's "a census that can fail short has already failed". The holder belongs in `backend/gates`, where `collectTableWrites` already walks every SQL write target in the tree, so it is that gate's corpus rather than this one's and would roughly double this diff.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved data integrity checks for communication preferences and suppression records across visibility, activity, system-generated, provider-generated, and address-only scenarios.
  - Added more precise classification for person acquisition evidence and related records, including system-originated suppression handling and server-derived write paths.
  - Added safeguards to ensure these records are correctly classified during validation.

### DIFF
--- a/backend/migrations/schema_fitness_integration_test.go
+++ b/backend/migrations/schema_fitness_integration_test.go
@@ -157,6 +157,30 @@ var rowScopedFKDecisions = gatekit.Waive(map[string]string{
 	"conversation_claim.person_id":          "gated: RecordConversationClaim opens with auth.RequireHuman + auth.Require(person, update), then auth.EnsureWritableLive on the person inside the write's own transaction — a claim may only be recorded against a person the caller could already change",
 	"conversation_claim.source_activity_id": "gated: the same writer takes auth.EnsureActivityContentVisibleLive on the cited activity in that transaction, so a claim can never quote a message the caller may not open — and LIVE rather than merely visible, because a claim must not outlive its evidence",
 	"conversation_claim.task_activity_id":   "PENDING WRITER: the column has no writer. The task an extracted commitment creates is written through the tasks substrate's own gated path, and this entry is replaced with that gate when the routing edge lands",
+	// The send-authorization record's two capability tables (core 1788407500)
+	// are SCHEMA ONLY: nothing in the tree inserts a row into either. What
+	// exists today READS them — consent's liveSuppression, which its own
+	// comment names as the only reader that applies a suppression — and
+	// REMOVES them, on privacy's erasure and retention paths, which null the
+	// subject link or delete the row and never create a reference. So the five
+	// entries below are the obligation on whoever writes the first row rather
+	// than a record of a gate already taken, and each names the sibling whose
+	// shape that writer has to match.
+	//
+	// The subject columns are one fact spelled twice, because the CHECKs make
+	// them alternatives: communication_basis_one_subject admits a person XOR a
+	// lead, and communication_suppression_names_a_target admits either, or
+	// neither when the row is scoped to a bare address. An address is a fact
+	// about a MAILBOX and names no row-scoped record, so it is outside this
+	// gate rather than exempted by it.
+	"communication_basis.person_id": "PENDING WRITER: the obligation on whoever writes it — a basis recorded from a path that already holds the subject (a captured inbound, an inquiry) carries that path's gate, as consent_qualifying_event.person_id does; a basis whose subject a CALLER names takes auth.EnsureVisibleLive on that person in the insert's own transaction, as lead_manual_signal.lead_id does, because recording why writing to somebody is lawful would otherwise confirm they exist",
+	"communication_basis.lead_id":   "PENDING WRITER: the lead arm of the entry above, under the same obligation — the CHECK makes the two exclusive, so a writer fills exactly one and the arm it fills carries that arm's probe",
+	// LIVE rather than merely visible, for the reason conversation_claim's
+	// evidence column is: a basis that outlived its evidence would keep
+	// authorizing sends on the strength of a message nobody can now open.
+	"communication_basis.source_activity_id": "PENDING WRITER: the obligation is conversation_claim.source_activity_id's — auth.EnsureActivityContentVisibleLive on the cited message, in the same transaction as the insert, so a basis can never rest on a thread the writer's caller may not read",
+	"communication_suppression.person_id":    "PENDING WRITER: the subject arm of a suppression, under communication_basis.person_id's obligation — with one difference worth writing down rather than rediscovering: a hard bounce arrives from a provider and not from a caller, and a system writer that names no principal takes no row-scope probe because there is no scope to probe. It is the CALLER-named row this binds",
+	"communication_suppression.lead_id":      "PENDING WRITER: the lead arm of the entry above; the target CHECK admits a row naming only an address, so a writer may fill neither subject column and the probe binds only the arm it fills",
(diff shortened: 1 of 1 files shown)


dungbanguyen:
Merged `origin/main` (`ab319cb38`) in — and the gate found a **sixth** column while this sat.

#3852 landed `person_acquisition_evidence`, whose `person_id` names a row-scoped record and had no decision. Classified as a child row, on the same terms as `person_email.person_id` beside it and for a stronger reason: `recordAcquisition` takes the id `createPerson` has just minted, on that same transaction, so **no caller can name a subject here at all**.

I checked for a second writer rather than assuming there was none, and there is one: `mergerelink.go` moves the rows onto a merge survivor. It supplies neither id from a request body — both come from the pair `mergePair` has already gated — so it does not change the classification, and the entry says so rather than leaving the next reader to re-derive it.

The five original entries are unchanged. `make test-it DIR=backend/migrations` green (all tests), `make check-backend` green on the merged tip.

Note for whoever reviews the two together: this PR and #3848 are both "unblock `main`'s integration lane" and they touch different packages — `backend/migrations` here, `compose/integration` + `modules/consent` there. Neither needs the other; the lane needs both.

dungbanguyen:
Rebased onto `65538d84a` — it was behind `main` by enough that its checks were failing on the stale base rather than on its own diff (the comparison showed ~2500 deletions that were just main's newer files).

Rebased, the diff is exactly 33 lines in `schema_fitness_integration_test.go`, and `TestFK_rowScopedTargetsHaveVisibilityDecision` passes. `make check-backend` is green apart from the contract-frontend-drift leg, which only refused because my scratch worktree has no `node_modules` — an environment gap, not a finding.

I did this because `main` is red on exactly these FKs and it is blocking other PRs, including https://github.com/margince/margince/pull/3838. Content is untouched; if you were mid-edit, apologies for the collision — say so and I will keep out of it.

dungbanguyen:
Closing: superseded by #3837, which classified all six columns while this was in review.

`main` at `18786238d` carries entries for `communication_basis.person_id`/`.lead_id`/`.source_activity_id`, `communication_suppression.person_id`/`.lead_id` and `person_acquisition_evidence.person_id`, and `make test-it DIR=backend/migrations` is green there. #3818 is fixed on `main`.

Worth noting one difference, because it is a claim about the code rather than a wording preference. The landed entries classify the five communication columns as **server-derived**, on the grounds that only privacy's subject-addressed statements name them. That is true of every statement in the tree today — but those statements only ever DELETE rows or null the subject link, and none of them creates a reference. What the gate asks about is the reference, so what those five columns actually are is not "server-derived" but "no writer yet": the tables are schema only.

The difference matters in one direction. "Server-derived" reads as a decision that has been made and holds; "pending writer" — the spelling this map already uses for `conversation_claim.task_activity_id` and `provider_applied_field.person_id` — says the obligation is on whoever writes the first row, and names the sibling whose gate they have to match. When a writer lands, the first reading stops being true silently and the second one is a note addressed to the person landing it.

Not reopening this for it. The gate is green, the columns are classified, and rewriting somebody's just-merged reasons is not worth a CI run — but I filed #3831 for the underlying hole either way: nothing fails when a column classified as having no writer gets one, whichever words the entry uses.

PR #3838, opened 2026-09-03
main is red: three suites still drive a retired double opt-in

dungbanguyen:
The real-Postgres lane has been red on the tip for days. [#3605](https://github.com/margince/margince/issues/3605) recorded a different cause each time it recurred; this is the one that is actually failing now.

```
consent_integration_test.go:336: issue DOI token → 422
consent_integration_test.go:394: DOI issuance for a non-DOI purpose → 409, want 422
```

## What happened

Operator-held double-opt-in issuance was **retired**, and for a real reason — `MissingDraftOwnerError`'s neighbour in `consent/handlers.go` says it plainly:

> "This endpoint returned the plaintext token to the authenticated operator, who could paste it back into RecordConsent — so the round trip the proof stands on could be closed without the subject's mailbox ever taking part, and the resulting consent_event recorded a confirmation that had not happened."

The endpoint refuses for every purpose now, and `double_opt_in_token` is gone from the contract entirely. Three test sites still drove it: two in `consent_integration_test.go`, and `grantMarketingConsent` in `comms_send_integration_test.go`.

## What they do now

They take the round trip as it actually works — the **confirm-details link**: a single-use token minted for the subject's own live primary address, answered on the anonymous page it opens.

The token is read from the store seam rather than the contract, deliberately. The contract hands it to nobody, and that IS the property the retired endpoint lacked — a test that could ask for the plaintext over the wire would be exercising exactly the hole this change closed.

The issuance test now asserts the **retirement**: it refuses whether or not the purpose requires double opt-in, and the answer names why. A malformed request is still malformed, because the handler probes the required id before refusing for its own reasons — an endpoint that says no must not become the one place a missing purpose goes unnamed.

## The required-body-id sweep

`TestAnOmittedRequiredIDIsNamedAndASuppliedOneStaysHidden` failed on the same endpoint, and the honest answer is not to exempt it.

That suite defends the property that a caller cannot distinguish an id naming nothing from one they may not see. A uniform refusal satisfies that **more** completely than a 404 does — every id gets the same answer, so there is nothing to compare — but it cannot answer 404 without inventing a lookup it does not perform. Making it 404 for invisible ids and 409 for visible ones is what would enumerate. So the case carries an explicit `suppliedStatus` and stays in the sweep.

## Four uncovered tools

The conformance sweep refuses a registered tool it never invokes, and four had arrived:

- `create_tag`, `update_tag` — need tag writes this lane's seat does not hold;
- `forecast_readings`, `forecast_movement` — need `forecast.read`, which it also does not hold. I tried invoking `forecast_readings` first (all its arguments are optional) and it answered `forecast.read: permission denied`, so it is waived with that reason rather than a guessed one. `forecast_movement` additionally needs two snapshots, which only the period's own producer mints.

Granting any of those here would widen the authority every other tool in the sweep runs under, which is the reason the existing tag waivers give.

## On the `uat` lane

[#3504](https://github.com/margince/margince/issues/3504) is **already fixed** — the 04:38 scheduled run has `uat (main): success`, and all five AC-shell specs pass locally. Only the 00:15 run failed. That issue can be closed; nothing here touches it.

## Verification

```
ok  github.com/margince/margince/backend/internal/compose/integration  292.038s
```

Whole package, plus `make check-backend` green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated marketing consent flows to use confirmation links and require fresh confirmation after withdrawal.
  * Operator-issued double opt-in requests now return clearer conflict responses when unsupported.
  * Refined validation for supplied but inaccessible identifiers to return endpoint-appropriate status codes.

* **Tests**
  * Expanded integration coverage for consent confirmation, identifier handling, and tool-schema behavior.
  * Documented supported exceptions for tools requiring additional setup or permissions.

### DIFF
--- a/backend/internal/compose/integration/comms_send_integration_test.go
+++ b/backend/internal/compose/integration/comms_send_integration_test.go
@@ -341,10 +341,17 @@ func TestAMarketingSendRendersBothOneClickUnsubscribeHeaders(t *testing.T) {
 	}
 }
 
-// grantMarketingConsent takes the recipient through the double-opt-in round
-// trip marketing_email requires — the server mints the token, the confirming
-// grant presents it — so the send under that purpose is lawful at both the
-// request-time gate and the dispatcher's.
+// grantMarketingConsent takes the recipient through the round trip
+// marketing_email requires, so the send under that purpose is lawful at both
+// the request-time gate and the dispatcher's.
+//
+// Through the CONFIRM LINK, which is how a double-opt-in purpose is confirmed
+// now: a single-use token minted for the subject's own live address, answered
+// on the anonymous page it opens. The token comes from the store rather than
+// the contract because the contract hands it to nobody — an operator holding
+// the plaintext could close the round trip with the subject's mailbox never
+// taking part, which is why the issuance endpoint that once returned it
+// refuses.
 func (p *preflightEnv) grantMarketingConsent(t *testing.T) {
 	t.Helper()
 	var purposes struct {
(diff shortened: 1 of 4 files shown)


dungbanguyen:
Superseded — every fix this carried reached `main` from other people while it sat in review, and I have verified all of it rather than assuming:

| what it carried | state on `main` (`d176bb932`) |
|---|---|
| the three suites driving the retired DOI issuance | fixed — `grantMarketingByConfirmLink` and `TestDOIIssuanceMintsNothingWhicheverPurposeIsNamed` are there, and both consent tests pass |
| the required-body-id sweep's expectation for that endpoint | passes |
| the four uncovered conformance tools | passes |

Whoever fixed the consent side reached the same conclusion by the same route — the confirm-details link, with the token taken from the store seam because the contract hands it to nobody.

Whole package on a clean `origin/main` worktree: `ok … 299.124s`, zero failures.

Nothing here is worth carrying forward, so this closes rather than rebasing to an empty diff.

PR #3840, opened 2026-09-03
main is red: five suites still describe a door #3807 closed

dungbanguyen:
`main` is red on five of six shards. Reproduced on plain `origin/main`; bisected to `a264d922b` (#3807).

## What #3807 did, and why it stays

It removed double-opt-in issuance. The endpoint minted a token and returned the plaintext to the **authenticated operator**, who could paste it straight back into the consent write — one person closing both halves of a round trip whose entire evidentiary value is that the *data subject* closed it from their own mailbox. The `consent_event` recorded a confirmation that had not happened.

That is correct and nothing here reopens it. What it left behind is five suites in the compose lane still driving the removed door, and two censuses no longer satisfied.

## Consent

The compose lane's own question is the **send gate** — no recorded decision suppresses, a recorded grant releases — and it never needed a double-opt-in purpose to ask it. It rode `marketing_email` only because the operator could mint that purpose's token.

- `TestConsentDefaultDenySuppressesSends` and the one-click header test now grant a **consent-class purpose they can legitimately grant**. For the header test that is exactly right on its own terms: `compose/preferenceunsub.go` emits the one-click surface for every purpose that is not *locked*, which is the property under test.
- The DOI purpose is now held to the refusals that are the whole of its surface here — bare grant, forged token, and the mint itself, each refused, **with the send still suppressed after all three**. A closed door that let the message out anyway would be decoration.
- The confirm-details round trip that *does* grant marketing is covered where it lives: `consent/confirmflow_integration_test.go`.

The required-id suite **keeps** its DOI case rather than dropping it. Issuance refuses before resolving anything, so the supplied arm cannot reach a lookup to answer 404 from — it expects the conflict and still checks the refusal names no field. The omitted arm is unchanged and is the reason to keep the case at all: *an endpoint that refuses for its own reasons must not become the one place a missing id goes unnamed* — which is what the handler's own comment says.

## Tool conformance

Five tools from #3822 and #3829 were registered with nothing holding their answers to their schemas.

- **Called:** `create_tag` and `update_tag` (a fixture mints a tag and renames it, so the id `update_tag` needs comes from the answer `create_tag` gives — the same shape `create_record` already uses), and `forecast_readings`, twice.
- **Waived, with the seam named:** `forecast_movement` needs two snapshots only the nightly pass writes; `forecast_input_checks` answers not-found until that pass has run, which is the correct answer to a question nobody has asked and not a result shape to hold.

## One more instance of a drift already fixed twice

The forecast tools were *also* unreachable because `AdminPerms` carried no `forecast` grant while the real seed gives admin `create`+`read`. That is exactly the `tag`/`list` drift from #3723, third occurrence. Fixed the same way, with the same reasoning in the same place.

## Evidence

- Bisected: passes at `a264d922b~1`, fails at `a264d922b`.
- `compose/integration` green (146s), and `capture`, `org360`, `collections`, `agentaccess`, `channels` subpackages green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected consent event exports to use the available creation timestamp.
  * Included person acquisition evidence in privacy export provenance.
* **Tests**
  * Expanded consent authorization, withdrawal, unsubscribe, and double opt-in safeguards.
  * Improved identifier validation while preventing sensitive information disclosure.
  * Added tag creation and update schema coverage.
  * Expanded forecast tool conformance and permission coverage.
  * Strengthened capture backfill coverage for eligible counterparties.
  * Added schema-fitness coverage for acquisition evidence relationships.
  * Verified privacy export queries against the live database schema.
  * Updated delivery checks to distinguish provider acceptance from relay failure.


dungbanguyen:
Four more red lanes landed on `main` while this was in CI, all verified on plain `origin/main` and folded in here — `main` is red on five of six shards, so splitting them into separate PRs would mean each waiting on the others.

**Visibility decisions (#3815).** `communication_basis` and `communication_suppression` arrived with foreign keys to `person`, `lead` and `activity` and no classification, failing the census on five columns. Nothing inserts into either table yet — privacy reads them for a subject-access package and deletes on erasure, the send path declares a decision against them, but no caller supplies any of these ids. Each entry records the obligation the writer inherits, in the shape the census already has for exactly this (`PENDING WRITER`, beside `conversation_claim.task_activity_id`). The evidence column is called out separately: a basis's `source_activity_id` is what the basis *stands on*, so its writer owes the stronger check `conversation_claim.source_activity_id` already takes.

**No access package could be assembled at all (#3841).** A new Art. 15 section selected `captured_at` from `consent_qualifying_event`, which records `occurred_at` and `created_at` and has never had a `captured_at`. Every export failed with `SQLSTATE 42703` — not a thinner package, *none*, because the gather runs its sections in order and the first failure aborts the assembly.

The column is corrected, and the gate is the point: `TestEverySARSectionStatementResolvesAgainstTheSchema` PREPAREs every section's statement against head. Preparing resolves each table and column name and checks argument arity — the whole class of defect — without seeding a subject or reading a row. It asks `sarSections` for the real list rather than listing queries of its own, so a section added to any chapter is covered the day it is added. Checked both ways: restoring `captured_at` fails naming section 16 and 42703; misspelling a table fails naming section 19 and 42P01.

**A flake that had cost its test's subject.** `TestConcurrentCreationsAreAllCounted` raced sixteen messages at once — eight sends and their eight replies — and asserted at least *two* creations resulted. Two races were running and only one is ever its subject: eight creations recomputing one run's reach is what the lock is about, while a reply overtaking the send it answers is the ladder deferring correctly. CI ran the second to completion, deferred every counterparty, and reported `the page ledgered 0 creations, so there was no race to lose` — the test failing for the reason its own comment gives. The sends are now the connector's serial prelude and only the replies race, so all eight are creatable and the floor is eight rather than two.

**One thing I could not close, reported rather than left implied:** dropping the run row's `FOR UPDATE` does *not* fail that test, before or after this change. Making it deterministic did not make it catch the lost update it names. Filed as #3851 with the mutation evidence.

**Evidence:** `make check-backend` exit 0; `compose/integration` green (189s); `migrations`, `privacy`, and the `capture` subpackage green.

dungbanguyen:
**Scope has shrunk a lot** — #3837 and the FK/tool-conformance work landed on `main` while this was in CI, fixing the same things. I took `main`'s version everywhere the two overlapped, including where theirs is better: `grantMarketingByConfirmLink` drives the real confirm-link round trip rather than sidestepping to a purpose that needs no double opt-in, which exercises the grant path that actually exists. My consent rewrite is gone in favour of it.

**What is left is what `main` does not have:**

- **`TestEverySARSectionStatementResolvesAgainstTheSchema`** — the durable half of the SAR fix. `main` corrected the column; nothing stops the next one. This PREPAREs every section's statement against head, which resolves each table and column name and checks argument arity without seeding a subject or reading a row. It asks `sarSections` for the real list, so a section added to any chapter is covered the day it is added. Checked both ways: restoring `captured_at` fails naming section 16 and 42703; misspelling a table fails naming section 19 and 42P01.
- **`TestConcurrentCreationsAreAllCounted` made deterministic** — it raced eight sends against their own eight replies and asked for at least *two* creations. CI ran that second race to completion, deferred every counterparty, and reported `the page ledgered 0 creations, so there was no race to lose`. Sends are now the connector's serial prelude, only replies race, and the floor is all eight.
- **The `forecast` and `data_coverage` fixture grants** — third and fourth instances of the drift `tag`/`list` had in #3723. `data_coverage`'s comment says plainly that nothing exercises it yet.
- **A comment on `confirmmail`** explaining why the field name moved.

**Re the CodeRabbit thread:** correct and fixed before the merge took that file. The withdrawal arm withdrew `marketing_email`, which `setupConsent` never grants — already suppressed by default deny, so the 409 it asserted was the default answering again and the arm passed whether withdrawal worked or not. `main`'s version of the file grants and withdraws the same purpose, so the same defect is closed there.

**Still open, reported not hidden:** dropping the run row's `FOR UPDATE` does *not* fail the capture test, before or after this change. Filed as #3851 with the mutation evidence.

`make check-backend` exit 0; `compose/integration`, `consent`, `privacy`, `migrations` and `capture` green.

PR #3848, opened 2026-09-03
main is red: the integration lane, from two merges that skipped it

dungbanguyen:
`origin/main` at `7ffbea6bf` fails five of six integration shards. **No open PR was fixing it**, so this is that fix. It blocks every branch: `ci` is the only required check and the integration lane is inside it, so three of my own PRs (#3825, #3830, #3834) are red for reasons that have nothing to do with their diffs.

Both causes are the same shape — a change landed with its unit lane green and its integration lane never run. That lane needs a real Postgres, so `make check` says nothing about it.

## Cause 1 — #3807 retired the double-opt-in token, four suites still drive it

#3807 removed the operator-held DOI token for a good reason: it was minted for the authenticated operator, who could paste it straight back into `recordConsent`, so the round trip whose whole evidentiary value is that the **subject** completed it from their own mailbox could be closed without the subject taking part. Issuance now refuses with 409 and mints nothing; the only thing that confirms a double-opt-in purpose is the confirm-details link.

Five tests still spoke the old contract:

| test | was |
| --- | --- |
| `TestConsentDefaultDenySuppressesSends` | granted marketing with a minted token |
| `TestConsentDoubleOptInNorm` | the whole mint/redeem round trip |
| `TestDOIIssuanceSupersedesAndValidatesPurpose` | issuance validates purpose, supersedes prior token |
| `TestAMarketingSendRendersBothOneClickUnsubscribeHeaders` | `grantMarketingConsent` minted a token |
| `TestAConfirmRequestMailsTheSubjectTheirOwnLink` | read `delivered`, renamed to `provider_accepted` |

**The fix reaches for the wiring rather than around it.** A granted marketing consent now comes from the real path: the harness composes `WithOperatorMail`, calls `confirm-request`, reads the link **out of the mail the product sent** — the response deliberately does not carry the token, and that is the property the whole mailbox-as-evidence claim rests on — and spends it anonymously at `POST /v1/public/confirm/{token}`.

That lives in one new helper (`confirmlink_integration_test.go`) used by both harnesses. Two spellings of "grant marketing" would be two readings of what the product accepts, and only one of them would be corrected the next time that path moves.

`TestDOIIssuanceSupersedesAndValidatesPurpose` had no subject left, so it is replaced by `TestDoubleOptInIssuanceRefusesAndMintsNothing`: the endpoint answers 409 (it is still in the public contract, so a caller deserves to be told why), for a DOI purpose and a non-DOI one alike, and **the audit log shows no mint** — a refusal that still wrote a row would leave a redeemable credential behind every 409.

`TestConsentDoubleOptInNorm` keeps everything that survived and gains the single-use half: a spent link replayed answers 404, and the send after it still refuses.

### One census needed a real change, not just an update

`TestAnOmittedRequiredIDIsNamedAndASuppliedOneStaysHidden` drives each body twice and asserts `omitted → 422 naming the field`, `supplied-but-invisible → 404`. The DOI row now gets 409 on the supplied arm, because the endpoint refuses everybody before it looks at any row.

Rather than delete the row — which would let the handler's body-id probe rot, and that probe exists precisely so *"an endpoint that refuses for its own reasons must not become the one place a missing id goes unnamed"* — the case gains a `suppliedStatus` override with the rule for using it stated on the field: an exception is only sound when the route **cannot vary its answer by whether the row exists**. That is true here, and the half this row cannot see is held next door: `TestDoubleOptInIssuanceRefusesAndMintsNothing` drives the same endpoint with two **visible** purposes and gets the same 409.

## Cause 2 — #3817 and #3822 added four tools the conformance sweep never invoked

`TestToolAnswersReachableWithoutApprovalSatisfyTheirSchemas` fails a registered tool that nothing holds to its schema. Four were uncovered: `create_tag`, `update_tag`, `forecast_readings`, `forecast_movement`.

- `create_tag` and `update_tag` are reachable — the lane's seat holds the tag write grants even though it lacks `tag.read` (which is what the existing `apply_tag`/`list_tags` waivers are about). `coinTagThroughTheToolSurface` coins a word and holds `create_tag`'s answer to its schema and envelope, exactly as `createThroughTheToolSurface` does for `create_record`, and `update_tag` edits that word from the table. `create_tag` joins `create_record` in the census's excused-by-fixture seed, with the reason beside it.
- The two forecast tools are refused: `forecast.read: permission denied`. Waived in the shape the tag family already uses — granting the seat that object would widen the authority every other tool in the sweep runs under. `forecast_movement` carries a second reason: its two required ids name stored snapshot rows, so without a snapshot producer the call would exercise the not-found path rather than the waterfall its schema describes.

## Mutation checks

The diff is tests, so the mutations break production and confirm the tests notice:

- `MailboxProof: MailboxProvenByConfirmLink` → `MailboxUnproven` in `recordMarketingAnswerTx` → all three marketing suites fail with `spend the confirm link with marketing_choice="granted" → 422`.
- `if err := httperr.RequireBodyID(…); err != nil && false` in `IssueDoubleOptIn` → the census's `omitted names the field` arm fails while the supplied arm still passes, which is the pair working as designed.

## Verification

- `make test-it DIR=backend/internal/compose/integration` — green, full package (138s).
- `make test-it DIR=backend/internal/modules/consent` — green, full package.

No production code changed. Once this lands the three PRs above want a rebase.

## Summary by CodeRabbit

- **Bug Fixes**
  - Privacy reports now identify consent event timestamps as `created_at`.
  - Confirmation-mail reporting now distinguishes relay acceptance from sending capability.
  - Direct double-opt-in token issuance requests are validated as conflict responses without token creation or auditing.

- **Tests**
  - Expanded integration coverage for mailed consent confirmations, single-use links, tags, forecasts, and data coverage.
  - Clarified expected error handling for unavailable or invalid resource identifiers.


dungbanguyen:
**Rebased onto the moving tip, and the scope grew —  went redder while this sat.** Merged `origin/main` (`ab319cb38`) in; the two original causes are unchanged and still fixed. Two more arrived with the eight commits that landed meanwhile:

**Cause 3 — #3841 selects a column `consent_qualifying_event` does not have.**

`sarsections.go`'s qualifying-events section reads `captured_at`; that table spells the recording instant `created_at` (`occurred_at` is the other date it carries, and it is the event's own). Every SAR and erasure suite failed with `ERROR: column "captured_at" does not exist (SQLSTATE 42703)` — eleven tests, and the first **production** change in this PR:

```
TestTheAssembledPackageIsSerializable
TestTheAccessPackageCarriesWhatIsHeldAboutTheSubject
TestErasureRemovesPIIEverywhereAndSticksViaSuppression
TestSARIncludesTheSubjectsSentMessages
TestSARExportsCustomFieldValues
TestSARHandsBackTheProviderClaimsAndTheRunHistory
TestASubjectAccessExportListsAHeldActivityWithoutItsText
TestASubjectAccessExportWithholdsARawOriginalNothingVouchesFor
TestASubjectAccessExportCarriesTheMailNobodyHasSentYet
TestABlindCopiedSubjectSeesTheirOwnMailAndNobodyElsesAddress
TestTheQueueGateIsWhatKeepsTheExportAdminOnly
```

The exported key changes with the column, which is right: `SARPackage.ConsentQualifyingEvents` is `[]map[string]any` keyed off the column names, nothing in `crm.yaml` declares its shape, and an export that renamed the column to a word the table does not use would be inventing a field.

**Cause 4 — #3829 added a third forecast tool the sweep cannot reach.** `forecast_input_checks` joins `forecast_readings` and `forecast_movement` in `unreachableInThisLane`. Verified rather than assumed: invoked in the lane it answers `forecast.read: permission denied`, the same refusal as its two siblings, and its waiver says so — it takes no arguments, so the grant is the only thing between this lane and the call.

**Re-verified on the merged tip:** `make test-it DIR=backend/internal/compose/integration` green, full package (205s). `make test-it DIR=backend/internal/modules/consent` green, full package. `make check-backend` green.

**Still red on `main` and not fixed here:** `TestFK_rowScopedTargetsHaveVisibilityDecision` in `backend/migrations`. That is #3818, and it belongs to #3830 — which #3852 has just given a sixth column (`person_acquisition_evidence.person_id`) to classify. I am taking that there rather than splitting it across two PRs.

dungbanguyen:
Third pass, and this one found the reason the conformance lane keeps failing rather than another instance of it.

**`AdminPerms` was short two grants the seeded admin role gives.** `harnessperms.go` opens by saying what it is for — *"the one part of the harness that has to mirror something outside the test tree: identity's seeded role documents. A fixture that drifts from that seed still passes, while proving nothing about production"* — and it had drifted. The real admin holds `forecast` (create+read) and `data_coverage` (read); the fixture held neither.

That is why every forecast tool answered `forecast.read: permission denied`, and why I waived three of them last round with a reason that sounded right and was wrong: *"granting it here would widen the authority every other tool in the sweep runs under"*. It would not have widened anything — it would have matched production. The tag and list entries a few lines above record exactly this drift, with exactly this consequence, and I read them as precedent for the waiver instead of as a warning.

With the grant:

- `forecast_readings` and `list_input_checks` are **called**, not waived.
- `forecast_input_checks` and `data_coverage` still cannot answer, for a real reason this time: both read last night's input-check run before anything else, and this lane runs no nightly pass. The store answers not-found — a different shape from the empty-findings case `list_input_checks` proves. Waived on the producer, like `decide_approval_bundle` beside them.
- `forecast_movement` keeps only its true half: it needs two stored snapshots to difference.

**And the census caught five stale waivers I would not have looked for.** The tag family (`apply_tag`, `remove_tag`, `list_tags`, `get_tag`, `get_record_tags`) was waived for want of a `tag.read` the fixture has held since *that* drift was corrected. Adding the calls made the census fail with "is both invoked here and excused as unreachable — one of the two is stale", which is the gate doing precisely its job. The waivers are gone and the five tools are called, so five schemas nothing was holding are now held.

Also merged `origin/main` (`65538d84a`) — `list_input_checks` and `data_coverage` arrived while this sat, and are covered above rather than one-off waived.

**Re-verified:** `make test-it DIR=backend/internal/compose/integration` green, full package (221s). `make test-it DIR=backend/internal/modules/consent` green, full package. `make check-backend` green.

One thing worth saying out loud, because it is the reason this PR has been rewritten three times: every cause here — the DOI tests, the SAR column, and each new tool — is a change that merged with its unit lane green and its integration lane never run. `make check` excludes the real-Postgres lane by design, so a branch that was behind `main` can be green on its own PR and still red `main` on merge. That is a process question rather than a patch, and I am not filing it as an issue from a session; it belongs to whoever owns the merge queue.

dungbanguyen:
Closing: superseded by #3837, which landed the same fix while this was in review.

#3837 does what this PR did for the two original causes, and does it the same way — the harness composes the operator relay, the confirm link is read out of the mail the product actually sent, and the subject spends it through the anonymous public edge. `TestDOIIssuanceSupersedesAndValidatesPurpose` is likewise replaced by a test that the retired endpoint refuses alike for a DOI and a non-DOI purpose and writes no row.

The SAR column is fixed there too, as `created_at AS captured_at` rather than the rename this PR made. Theirs is the better call: every other section in `sarsections.go` reports the recording instant under `captured_at`, so aliasing keeps the export's own vocabulary consistent where my rename would have made one section speak differently from its neighbours.

Verified on `main` at `18786238d`: `make test-it` green on `backend/internal/compose/integration` (full package), `backend/internal/modules/consent` (full package) and `backend/migrations`. The lane is no longer red, so there is nothing left here to land.

One finding from this PR was **not** superseded and is now #3893 on its own: `AdminPerms` is short the `forecast` and `data_coverage` grants the seeded admin role holds, which is why every forecast tool arrives in the conformance lane as a waiver. #3837 waived the new ones on that reasoning too, so the drift is still there and still producing waivers.

Nothing here was wasted work to chase down — but the duplication was real, and it happened because neither of us had claimed the red publicly before starting. I commented on #3605 with the diagnosis before opening this; a comment on a ticket is evidently not where somebody fixing `main` looks.

PR #3893, opened 2026-09-03
The conformance lane's admin holds the grants its real seed gives

dungbanguyen:
Seven agent tools were waived out of `TestToolAnswersReachableWithoutApprovalSatisfyTheirSchemas` for want of grants the seeded admin role actually holds. Nothing was checking their schemas or their envelopes, and the reason recorded for that was wrong.

`harnessperms.go` opens by saying exactly what it is for:

> They live beside the environment rather than inside it because they are the one part of the harness that has to mirror something outside the test tree: identity's seeded role documents. **A fixture that drifts from that seed still passes, while proving nothing about production.**

It had drifted. `identity/internal/policy/defaults.go`'s admin row grants `forecast` create+read and `data_coverage` read; `AdminPerms` held neither. The `tag` and `list` entries a few lines above record the same drift with the same consequence — that one was found when an admin missing `tag.read` asked "who carries VIP" and was answered "nobody".

## What that cost

The waivers read as facts about the lane and were facts about the fixture:

> "needs a seat holding forecast.read, which this lane's seat does not carry — measured, not assumed: the bare call comes back permission denied. **Granting it here would widen the authority every other tool in the sweep runs under**"

The measurement was right and the conclusion was not: granting it does not widen anything, it matches production. The tag family carried the same reasoning about `tag.read` — which the fixture has held since that earlier drift was corrected, so those five waivers were describing a seat that no longer existed.

## What is called now

With the two grants, **seven tools move from waived to invoked**, so their declared output schema and envelope are held against a live handler for the first time:

- `list_tags`, `get_tag`, `apply_tag`, `get_record_tags`, `remove_tag` — ordered so `apply_tag` leads and `remove_tag` follows, and the two reads in between see a record actually carrying the word. An empty answer keeps its shape whatever the read did, which is the case worth not proving.
- `forecast_readings`, `list_input_checks` — both take no arguments.

## What is still waived, and why it is not the grant

Three remain, each with the reason that actually stops it:

- `forecast_input_checks` and `data_coverage` read last night's input-check **run** before anything else, and this lane composes no scheduler, so the store answers not-found. That is a different shape from the empty-findings case `list_input_checks` proves above it — measured after the grant landed, not assumed.
- `forecast_movement` needs two stored snapshots to difference. Its siblings are called above, so the grant is not what stops this one.

## Verification

- `make test-it DIR=backend/internal/compose/integration` green, full package (139s).
- `make check-backend` green.
- Mutation-checked: setting `"forecast": {Create: true, Read: false}` fails `forecast_readings` and `list_input_checks` with `forecast.read: permission denied`, and nothing else.

Found while chasing the repeated conformance failures on `main` — every new forecast tool was landing straight into a waiver, because the first one had. The fixture was the reason, and it will keep producing them until it mirrors the seed.

## Summary by CodeRabbit

- **Tests**
  - Expanded integration coverage for tag management and forecasting workflows.
  - Added validation for forecast readings, input checks, data coverage, and forecast movement scenarios.
  - Updated test fixtures to reflect current access requirements for forecasting and data coverage features.
  - Refined test handling for scenarios where input-check runs or snapshot data are unavailable.

### DIFF
--- a/backend/internal/compose/integration/harnessperms.go
+++ b/backend/internal/compose/integration/harnessperms.go
@@ -208,6 +208,15 @@ var (
 			// narrower than AccountRepPerms, which already carries both.
 			"tag":  {Create: true, Read: true, Update: true, Delete: true},
 			"list": {Create: true, Read: true, Update: true, Delete: true},
+			// forecast is create+read for admin in the real seed and
+			// data_coverage is read (identity/internal/policy/defaults.go's
+			// admin row). The same drift the tag and list entries above record,
+			// and no more inert: every forecast tool on the agent surface came
+			// back `permission denied` in the conformance lane, which reads as
+			// "this lane cannot reach them" rather than as "this fixture is
+			// short a grant production gives".
+			"forecast":      {Create: true, Read: true},
+			"data_coverage": {Read: true},
 		},
 		RowScope: principal.RowScopeAll,
 	}
--- a/backend/internal/compose/integration/toolschema_conformance_integration_test.go
+++ b/backend/internal/compose/integration/toolschema_conformance_integration_test.go
@@ -225,6 +225,20 @@ func TestToolAnswersReachableWithoutApprovalSatisfyTheirSchemas(t *testing.T) {
 		// Renaming the word coined above. Both tag-vocabulary writes are reachable
 		// here: AdminPerms carries tag create/read/update/delete.
 		{"update_tag", `{"tag_id":"` + tag.String() + `","name":"conformance-renamed","color":"amber"}`},
+		// The rest of the tag family, reachable for the same reason: the
+		// fixture carries tag read as well as the writes. apply_tag leads and
+		// remove_tag follows, so the middle two read a record that is actually
+		// carrying the word rather than an empty answer that would keep its
+		// shape whatever the read did.
+		{"list_tags", `{}`},
+		{"get_tag", `{"tag_id":"` + tag.String() + `"}`},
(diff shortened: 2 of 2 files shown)


dungbanguyen:
Closing: #3886 covers this and goes further.

It reaches the same finding — the tag waivers' `tag.read` reason had expired, and the forecast tools were waived for want of grants the real admin role holds — and then does the part I waived: it seeds two `recheck` snapshots and a finished input check through the writers production uses, so `forecast_movement`, `forecast_input_checks` and `data_coverage` are **called** rather than waived on a missing producer. That is the better answer; a waiver naming a producer is still a schema nothing holds.

One substantive difference worth a reviewer's eye rather than a re-run of this PR. #3886 clones a seat instead of adding the grants to `AdminPerms`, reasoning that *"several suites read that fixture as a seat which cannot see the forecast, and widening it there would make them pass without testing anything."*

I looked for those suites before opening this and did not find one: nothing in `backend/internal/compose/integration` asserts a forecast refusal under `AdminPerms`. If that holds, the cloned seat leaves the fixture still diverged from the seeded admin — and `harnessperms.go` opens by saying that is the one thing it must not be: *"the one part of the harness that has to mirror something outside the test tree: identity's seeded role documents. A fixture that drifts from that seed still passes, while proving nothing about production."* The `tag`/`list` entries in that file record the last time this drift was found, and the expired `tag.read` waivers #3886 is deleting are what it cost.

Not blocking on it — the census is answered either way, which is what `main` needs. But if no suite actually depends on the narrower seat, the grants belong in the fixture, and leaving them out means the next forecast tool arrives as a waiver again for the same wrong reason.

PR #3946, opened 2026-09-03
The overlay flip runs on one workspace handle

### LINKED ISSUE #2561: The overlay flip runs on two database handles, so its import and its mode flip can target different workspaces
Found while reviewing #2521 (ADR-0091 §5). Not reachable today; filed so the design gap is not carried silently.

## What

`compose.FlipRunner` uses two different bindings for one operation:

- the import and the reconstruction go through `actingWorkspaceDB(ctx, f.pool)` — `database.BindTo` pinned to **the workspace in the request context**;
- the mode flip goes through `f.svc.CompleteFlip`, an `overlay.Service` built on `InstallationDB(pool)` — a handle that resolves **the installation's singleton** through `identity.InstallationWorkspace` and never consults ctx.

If those two ever name different workspaces, the flip imports an estate into one workspace and flips the other out of overlay mode.

## Why it is not reachable now

Every path into `FlipRunner.Execute` is HTTP-driven, and `identity/middleware.go` binds the request context from the same cached `InstallationWorkspace` resolver the handle uses. So the two agree on every live path. `actingWorkspaceDB` exists for the case where they would not — its own doc says the rebuild target "on a clean instance is a workspace the server never resolved" — and that case has no caller today.

## Why it is not fixed in #2521

Before that PR, `CompleteFlip` took the id from ctx (the acting workspace) while its transaction was bound by `s.db.Tx` to the installation's. The UPDATE and the transaction it ran in could already name different workspaces. #2521 makes the statement agree with its own transaction, which is strictly more consistent — but it does not make the *lane* consistent, because the transaction itself is still opened on the installation handle while the import runs on the acting one.

Making the predicate follow ctx again would only move the disagreement back inside one transaction. The fix belongs one level up: **the whole flip should run on ONE handle.**

## Options

1. Build the `overlay.Service` the runner uses over `actingWorkspaceDB`, so import, reconstruction and mode flip share a binding. Keeps the acting-workspace capability.
2. Drop `actingWorkspaceDB` from the flip lane and run everything on `InstallationDB`, accepting that the flip only ever targets the installation's workspace. This is the ADR-0091 direction — one installation, one organization — and would let `actingWorkspaceDB` be deleted once `flipbundle.go` follows.

Option 2 is the smaller end state and matches where ADR-0091 is heading; option 1 is right only if a rebuild really must write into a workspace the server cannot resolve. That is the question to settle first.

## Where

- `backend/internal/compose/flip.go` (`actingWorkspaceDB` at :172 and :281, `f.svc.CompleteFlip` at :255)
- `backend/internal/compose/flipbundle.go:175`
- `backend/internal/compose/dbhandle.go:45`
- `backend/internal/modules/overlay/flipstate.go` (`CompleteFlip`)

Raised by cubic on #2521 as P1; downgraded here to `priority: normal` because no live path reaches the divergence, and recorded rather than fixed in that PR because the fix is a change to how the flip lane is composed, not to a predicate.

dungbanguyen:
Closes #2561.

## It was three handles, not two

The ticket names two; there was a third.

| step | handle |
|---|---|
| import + reconstruction | `actingWorkspaceDB` — the workspace the operator is acting in |
| mode flip | `f.svc`, an `overlay.Service` over `InstallationDB` |
| **run records** | `f.runs`, a field built over `InstallationDB` at construction |

That last one matters on its own: a flip's run records could land in a different workspace from the estate it imported, and the record is what a **resumed** attempt reads to know where it got to.

## The fix

`Execute` resolves **one** handle and hands it down. Resolving it there rather than per step is the whole point — three callers each asking separately is what let two of them ask a different question.

- `overlay.Service.On(db)` rebinds a service with **every option kept**. That matters because `CompleteFlip` calls the mode-flip observer: a second service built fresh for the flip would have dropped it silently. The shallow copy is correct only because this struct caches nothing per handle, and its doc says so rather than leaving the next reader to verify it.
- The runner's run-store **field is gone**, not rebound. A field is a second answer available to a method that had already resolved a handle — exactly how two of the three lanes ended up on the installation binding.

## Option 1, deliberately, and option 2 is not foreclosed

Option 2 deletes `actingWorkspaceDB` and accepts that a flip only ever targets the installation's workspace. That turns on whether a rebuild must write into a workspace the server cannot resolve — and `dbhandle.go` currently asserts it must ("on a clean instance is a workspace the server never resolved"). That is the ticket's *"question to settle first"*, and it is a **capability ruling, not a composition one**.

Option 1 closes the divergence either way, so the bug does not wait on that ruling. If option 2 is later chosen, this change is what makes it a deletion rather than a rewrite.

## The gate is the point

The divergence was invisible and not reachable: every caller is HTTP-driven, and identity's middleware binds the request context from the same resolver the installation handle uses, so the two agreed on every live path. **No behavioural test could have caught them differing** — which is also why re-introducing it is one line and would be as invisible as it was the first time.

So the rule is shape, not behaviour: nothing in `flip.go` may reach `InstallationDB`; the lane must still reach `actingWorkspaceDB` somewhere (the positive floor, so a prohibition cannot pass by checking nothing); and no `RunStore` may sit on the runner.

**All three arms mutation-checked**, each failing with its own message:

```
M1 reintroduce InstallationDB → flip.go reaches InstallationDB 1 time(s). …
M2 put the RunStore back      → flip.go puts a RunStore on flipRunner. …
M3 stop resolving the acting  → flip.go reaches actingWorkspaceDB nowhere — this
   handle                        gate was about to pass having checked nothing
```

`make check-backend` exit 0; `modules/overlay`, the `overlay` integration subpackage, and `compose` green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workspace consistency during overlay migrations by ensuring related operations use the same workspace database.
  * Preserved the ability to resume migrations after failures using saved checkpoints.
  * Improved reliability for migration preflight, execution, previews, and run creation or resumption by keeping operations aligned to the active workspace.
  * Improved validation of workspace handling and consistency of diagnostic output.

* **Documentation**
  * Updated the gate inventory to include the new workspace-binding validation.

### DIFF
--- a/backend/gates/agentgrantscopes_test.go
+++ b/backend/gates/agentgrantscopes_test.go
@@ -28,6 +28,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"sort"
 	"strings"
 	"testing"
 )
@@ -197,10 +198,17 @@ func between(src, from, to string) string {
 	return rest
 }
 
+// sortedKeys is the set's keys in a stable order.
+//
+// It really does sort now. Both callers put the result in a FAILURE MESSAGE,
+// and map iteration order is randomised per run — so the same finding read
+// differently every time, which is how somebody comparing two runs concludes
+// the tree moved when only the map did.
 func sortedKeys(set map[string]bool) []string {
 	out := make([]string, 0, len(set))
 	for k := range set {
 		out = append(out, k)
 	}
+	sort.Strings(out)
 	return out
 }
--- a/backend/gates/fliponehandle_test.go
+++ b/backend/gates/fliponehandle_test.go
@@ -51,13 +51,21 @@ func TestTheOverlayFlipLaneResolvesOneWorkspaceHandle(t *testing.T) {
 	// silently matched no file at all, or if the lane stopped resolving a
 	// handle by any route.
 	//
-	// Counted PER FUNCTION, not per file. Two resolutions in flip.go is the
-	// correct shape and always was: the parity dry-run is its own operation —
-	// zero-write, no run id, no operator — and Execute is another. What must
-	// never happen is one operation resolving twice, because that is the
-	// divergence itself rather than a route to it. A file-wide count of one
-	// would fail today and say nothing about the invariant; a file-wide count
-	// of "any" would admit Execute asking twice.
+	// Counted against NAMED ENTRY POINTS, not per function and not per file.
+	// Each is finer than the last for a reason the coarser version misses:
+	//
+	//   - per FILE would demand one call and fail today. Two resolutions in
+	//     this file is the correct shape and always was — the parity dry-run is
+	//     its own operation (zero-write, no run id, no operator) and Execute is
+	//     another.
+	//   - per FUNCTION admits a helper. Execute resolves once, a helper it
+	//     calls resolves once, both counts are one, and the operation is back
+	//     to two handles — which is the defect, not a route to it.
+	//
+	// So resolution is a PRIVILEGE of the two operation entry points, and every
+	// other function in the lane must receive the handle. That is the rule that
+	// cannot be satisfied by moving the second call somewhere else.
+	resolvers := map[string]bool{"Execute": true, "parityPreview": true}
 	acting := map[string]int{}
 	installation := 0
 	for _, decl := range file.Decls {
@@ -70,12 +78,12 @@ func TestTheOverlayFlipLaneResolvesOneWorkspaceHandle(t *testing.T) {
 			if !ok {
 				return true
 			}
-			// calleeName (retentionscope_test.go) already answers this for
-			// the package, and it handles a QUALIFIED call as well as a bare
-			// one — which matters here even though dbhandle.go is the same
-			// package today: a helper that later moves would arrive as a
-			// selector and slip past a gate that only looked at identifiers,
-			// and a prohibition that stops matching is one that passes.
+			// calleeName (retentionscope_test.go) already answers this for the
+			// package, and it handles a QUALIFIED call as well as a bare one —
+			// which matters here even though dbhandle.go is the same package
+			// today: a helper that later moves would arrive as a selector and
+			// slip past a gate that only looked at identifiers, and a
+			// prohibition that stops matching is one that passes.
 			switch calleeName(call) {
 			case "actingWorkspaceDB":
 				acting[fn.Name.Name]++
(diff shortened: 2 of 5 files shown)


dungbanguyen:
Both addressed — one as suggested, one differently, and I want to be explicit about which.

**Qualified calls — valid, fixed as suggested.** The prohibition matched only a bare `*ast.Ident`, so a `dbhandle.InstallationDB(...)` would have arrived as a `*ast.SelectorExpr` and slipped past. A prohibition that stops matching is one that passes.

It now uses `calleeName` — the helper this package already has for exactly this — rather than a second copy. That branch is not untested, and I checked rather than assumed: making it return `""` fails two existing gates, `TestRetentionPassCtxOnlyDrivesTheRetentionPass` and `TestEveryReaderOfTheRelationshipTableCarriesTheEdgeGateOrAVerdict`, both of which match genuinely qualified calls (`identity.LiveMemberSQL(…)` among them). I could not mutation-check it through *this* gate, because a qualified `InstallationDB` call does not compile — `dbhandle.go` is the same package — so the coverage is real but it is somebody else's gate proving it.

**Exactly one `actingWorkspaceDB` — the concern is right, the rule as stated is not.** It would fail today, and it would say nothing about the invariant.

Two resolutions in `flip.go` is the correct shape and always was: `parityPreview` is its own operation — zero-write, no run id, no operator — and `Execute` is another. What must never happen is **one operation** resolving twice, because that is the divergence itself rather than a route to it.

So it counts **per function**:

```
M4 make Execute resolve twice → flip.go: Execute resolves the acting workspace 2 times.
                                One operation, one handle — a second resolution inside one
                                function is the divergence #2561 closed, not a route to it.
```

The "reaches it nowhere" floor stays, so the prohibition still cannot pass by checking nothing.

`make check-backend` exit 0.

dungbanguyen:
Correct, and the hole is real — per-function counting admitted exactly that: `Execute` resolves once, a helper it calls resolves once, both counts are one, and the operation is back on two handles.

Fixed as suggested. **Resolution is now a privilege of the named operation entry points** — `Execute` and `parityPreview` — and every other function in the lane must receive the handle. That is the rule a second call cannot satisfy by moving somewhere else.

Each entry point is also *required* to resolve, which is the floor in the other direction: a prohibition whose subjects stopped existing reads exactly like one with nothing to find, and a renamed entry point would leave the list describing a function that is gone.

**Both arms checked:**

```
M6 a helper resolves its own handle →
   flip.go: sneakyHelper resolves the acting workspace, and only the operation entry
   points (Execute, parityPreview) may. A helper that resolves its own handle puts the
   operation back on two of them however few times each function asks (#2561) — take
   the handle as a parameter instead.

M7 an entry point stops resolving →
   flip.go: parityPreview resolves no workspace handle. It is one of the lane's
   operation entry points, so if it stopped resolving one this gate was about to pass
   having checked nothing — or the entry point was renamed and this list is now
   describing a function that does not exist
```

One thing found on the way and fixed since both callers need it: the package's `sortedKeys` **did not sort**. Both uses put the result in a failure message, and map iteration order is randomised per run — so the same finding read differently every time, which is how somebody comparing two runs concludes the tree moved when only the map did.

`make check-backend` exit 0.

Issue #2522, opened 2026-08-24
38 of the last 500 commits on main carry no Signed-off-by, and the DCO gate caught them


tinnhlt-gra:
## The artifact, not the process

CLAUDE.md: *"Sign off every commit (`git commit -s`) — the DCO gate rejects a PR containing any commit without a `Signed-off-by` trailer."* CONTRIBUTING.md says the same.

**38 of the last 500 commits on `main` carry no `Signed-off-by` trailer.** The oldest is from 2026-08-17; the newest landed today at 09:48 +0700.

This is distinct from #2504. That issue is about merges landing *before* their checks report — a timing hole where no verdict existed. This one is the opposite and worse: **the gate ran, produced the correct verdict, and the merge happened anyway.**

Worked example, `c3a203f` (#2516):

```
merged_at            2026-08-24T02:48:19Z
dco    completed_at  2026-08-24T02:16:58Z  failure   <- 31 min BEFORE the merge
ci     completed_at  2026-08-24T02:29:56Z  failure   <- 18 min BEFORE the merge
```

The `dco` job said exactly what it should:

```
DCO: commit 947408b missing Signed-off-by trailer:
     fix(agents): the import copy says an empty id means a new company
Every commit must be signed off (git commit -s). See CONTRIBUTING.md.
```

`ci` — the one required check — then failed *because* of it: `FAIL: dco reported 'failure'. FAIL: 1 of 9 job(s) did not pass`. The other eight passed, so nothing was broken except provenance. It merged 18 minutes later.

## Why this needs a decision rather than a fix

**The artifacts are permanent.** These 38 commits are on `main`. Rewriting history to add trailers would break every clone, every open PR's base, and every deployed ledger reference — the remedy is worse than the defect. So the question is not "how do we fix these" but "what do we do about a public BUSL-licensed repository whose stated contributor-provenance requirement has not held 38 times since 17 August".

Options, and I recommend the second:

1. **Nothing.** Accept the gap and keep the gate for future commits. Cheapest, and leaves the stated policy false about the tree's own history.
2. **Record it, then hold the line.** A note in CONTRIBUTING.md or a `dco-exceptions` record naming the range and why it was not rewritten, so the policy is true about what it claims going forward and honest about what came before. Then close the enforcement hole so the count stops growing.
3. **Rewrite history.** Not recommended, for the reasons above.

Whichever is chosen, the enforcement hole is the part that makes this recur — see #2504.

## The detector

```sh
[...]

Issue #2496, opened 2026-08-23
Two admin-merged PRs left main uncompilable; nothing checks the merge result


LarsGradion:
## What happened

`origin/main` did not compile for roughly 20 minutes on 2026-08-24.

- **#2486** made `dealstatus.promptVersion` derive from `statusSystemFor`.
- **#2490** gave `statusSystemFor` a `lang` parameter.

Each branch was green. Neither could see the other's half. Both merged with `--admin`, which lands a PR before any gate runs on the merge result — so the first thing to discover the breakage was a human running `go build` afterwards.

Fixed in #2495.

## Why this is worth an issue rather than a shrug

The two changes were **semantically related** — one derives a version from a prompt builder, the other changes that builder's signature — and they were merged within an hour of each other by different sessions. This is not a rare collision; it is what happens whenever two sessions work the same area, which is the normal mode in this tree.

Branch protection would have caught it, because the merge queue builds the combination. `--admin` is what skips that.

## Options

1. **Require a status check on `main`** — the merge stops until CI has run on the actual merge commit. Costs the wait that `--admin` exists to avoid.
2. **Post-merge `main` build alarm** — does not prevent a red `main`, but bounds how long it stays red and who finds out. Cheap.
3. **Accept it** — a red `main` is the known price of admin-merge, and whoever hits it fixes it forward.

This needs a product/process ruling, not a patch. Recommending (2) as the floor: the cost of a red `main` is not the breakage, it is that every other session's local gate result becomes untrustworthy until somebody notices.

## Note

`main` on this repo is currently unprotected, so `--admin` is a no-op there — the merge would have succeeded either way. That makes (1) a real configuration change, not just a habit change.

LarsGradion:
## B already exists, and it caught both of today's breakages

`main-health.yml` runs the full backend gate, the integration lane and the frontend lane against main's tip every two hours, then files one issue per broken lane naming the commits that landed since the last green run.

Its record over the last 30 runs: **24 green, 6 red.** Main goes red roughly one run in five, and this is what catches it each time.

Today it caught two, both mine:

- The #2486/#2490 signature collision this issue was filed about. It merged at 23:27 UTC; the alarm last ran at 22:38 and next fired at 00:17, so it **would** have caught it — I just found it 20 minutes earlier by hand.
- A nil pool in the offer-draft test fixture from #2490. It filed **#2509** three minutes after the merge, before I knew anything was wrong. Fixed in #2510.

[...]

PR #3981, opened 2026-09-04
A merge that landed without a verdict says so, at push time

### LINKED ISSUE #2504: Seven PRs merged before their required check reported at all — main went red and no gate could have stopped it
## What happened

Between 23:26 and 23:56 on 2026-08-23, **four PRs merged while their `ci` check was FAILURE**. Verified from the API, not inferred:

| PR | merged | `ci` | also failing |
|---|---|---|---|
| #2490 | 23:26:55 | **FAILURE** | **`dco`**, deterministic-gates, extension-reference, all 6 integration shards, unit coverage, RLS/e2e, govulncheck, live-boot |
| #2495 | 23:35:30 | **FAILURE** | all 6 integration shards, RLS/e2e |
| #2494 | 23:50:42 | **FAILURE** | all 6 integration shards, RLS/e2e |
| #2499 | 23:56:42 | **FAILURE** | all 6 integration shards, RLS/e2e |

`main` was red as a direct result. Reproduced on clean `origin/main` at `e5961b7`, from a detached worktree with no local changes:

```
--- FAIL: TestDraftOfferLinesStagesGroundedLinesAndDiscloses
panic: runtime error: invalid memory address or nil pointer dereference
  identity.BaseLanguageForPrompt(...)   settingsentry.go:215
  compose.offerDrafter.draftCandidates(...) offerdraft.go:307
```

Every `TestDraftOfferLines*` test dies at the same line. The defect itself is being fixed separately; **this issue is about the four merges, not the panic.**

## Why this matters more than one red main

[CLAUDE.md](CLAUDE.md) states the rule plainly, twice:

> **Merge only when everything is green** … Never merge over a red or still-running check.

and about DCO:

> **Sign off every commit** (`git commit -s`) — the DCO gate rejects a PR containing any commit without a `Signed-off-by` trailer.

#2490 merged with `dco` FAILING. A gate described as one that *rejects* a PR did not reject it. Whatever mechanism allowed that (a repo-role bypass is the likely one) means **`ci` being the only required check on `main` buys less than the documentation implies** — a required check is only required for those who cannot bypass it.

The second-order cost is the one to weigh. `main-health` runs 2-hourly, so it is not the first place a break shows up; it was still reporting the 22:38 green while main had been broken since 23:26. For 90 minutes the only signal available to anyone was their own PR going red against a base they did not break — which is exactly how this was found, twice, independently, by two sessions each spending a rebase and a full local lane to rule their own diff out.

**That is the real cost: every open PR pays to re-derive a fact somebody already knew.**

## What is not being claimed

- Not that the merges were done in bad faith. Four in thirty minutes, during a series of related fixes, reads like a deliberate push-through of a stack where each PR was expected to fix the previous one's red. #2495's own title is "main compiles again".
- Not that bypass should be removed. There are real cases for it — an infrastructure outage that reds every PR, a revert that must land now.
- Not that this is the panic's fault. #2490's blast radius is a separate, already-diagnosed defect.

## What would actually help

Ranked by how much they cost versus what they buy:

1. **Make a bypassed merge loud.** If a merge lands over a red required check, that fact should reach the same place a red `main-health` does. Right now it reaches nobody — the PR closes green-looking in the list, and the next signal is 2 hours later, or another session's rebase. This is the one I would do first and it needs no policy change at all.
2. **Shorten the gap for `main` specifically.** `main-health` at 2 hours is fine for catching drift; it is too slow to be the backstop for a bypass. A push-triggered run on `main` would have named this at 23:27 instead of 01:00.
3. **Decide, and write down, when bypass is legitimate.** The rulebook currently says "never merge over red", which is not what the repo permits. A rule the tooling contradicts teaches people to read rules as advisory. Either the rule or the permission should move.

Filed by a session whose own PR (#2498) went red against this base; I am holding that PR's push rather than spending another CI run on a base known to be broken.

dungbanguyen:
Closes #2504 — its **item 1**, "make a bypassed merge loud", which the ticket ranks first and notes *"needs no policy change at all"*.

## What lands

`merge-attest.yml`, on every push to `main`. One question: **did what just landed have a verdict behind it** — a pull request naming the commit, and a `ci` check that reported, was green, and finished *before* the merge.

It **runs no lane** — two API reads and a jq, which is what makes a push trigger affordable at ~85 merges a day. `main-health` still owns whether `main` actually works.

It is **not a gate and never will be**: it runs after the merge and blocks nothing. A repository role merging past `ci` stays deliberate — an infrastructure outage that reds every PR, a revert that has to land now. What changes is that the fact is loud and attributed at push time instead of surfacing two hours later as somebody else's PR going red against a base they did not break.

## The four findings

Each reads identically from outside, which is the reason for the split:

| | |
|---|---|
| nothing merged it | no PR names the commit — no review, no check ever applied |
| **the check never reported** | #2504's own case. An absent verdict is not a red one, and nothing asking *"was it green?"* can tell them apart |
| the check was not green | includes `skipped`, which GitHub counts as passing |
| it went green *after* the merge | decided on an unfinished run; the answer arriving later is luck |

The **oldest** matching run is judged, so a re-run after the merge cannot clear the record of the merge it was absent for.

## Evidence

**Against two real merges, end to end** — the actual `gh api` calls the workflow makes, re-run just now on current `main`:

- `57db6381` (#3952, merged green) → `ok: … merged behind a green ci (pull request #3952)`
- `c3a203f` (#2516 — the merge #2522 documents as landing over a red `ci`) → `pull request #2516 merged while its required ci check was failure`

**The judge** (`make test-merge-verdict`) — 12 cases, including every arm above, the skipped-counts-as-passing case, the re-run case, and two that must exit **2** rather than 1: an unset lookup and a missing sha are the alarm being *broken*, not a finding. Reporting the first as the second is how an alarm gets switched off.

**The wiring**, mutation-checked three ways — drop the report job's condition, drop an env line, or stop running the judge, and `test-scheduled-report` names which. A `dco` arm was once added with its cases and its env line and still filed nothing, because the report job's `if:` did not select it; this is that invariant asked of the new workflow.

`make check-backend` exit 0.

## Items 2 and 3, deliberately not here

- **Item 2** (a push-triggered *lane* on `main`) is a runner-budget call: `main-health` is ~15 jobs, and at 85 merges a day that is a different order of spend against a 20-concurrent org ceiling the PR lanes queue in. It is #2496's option 2 and belongs with that ticket's ruling.
- **Item 3** ("decide, and write down, when bypass is legitimate") is a policy ruling.

## Note

This sat finished-but-unpushable for a while: the token had no `workflow` scope, so GitHub refused any branch touching `.github/workflows/`. That is resolved.

## Summary by CodeRabbit

* **New Features**
  * Added automated verification that merged changes include a successful required check completed before merging.
  * Added reporting for missing or invalid merge verdicts, including pull request details, failure reasons, and run information when available.
  * Verification runs independently without blocking standard CI workflows.

* **Documentation**
  * Documented the merge verification workflow and updated workflow counts.

* **Tests**
  * Added coverage for successful, missing, failed, pending, late, and rerun checks, plus reporting behavior.


dungbanguyen:
All three valid, all three fixed — and they are one failure wearing three shapes: an alarm that files a wrong accusation is one people learn to ignore, which costs more than the silence it was built to replace.

**`filter=all`.** The endpoint defaults to `filter=latest` — the most recent attempt per check. The judge deliberately reads the **oldest** attempt so a re-run after the merge cannot clear the record of the merge it was absent for, and the default returned precisely the attempt that would clear it. The rule was unenforceable and nothing said so.

**One pull request, chosen once.** Exactly as described: the workflow took its head from `.[0]` while the judge reported on `min_by(.number)`, so when order and minimum disagreed it described PR A while judging B's check runs against A's merge time. Taking your suggested option — the workflow now narrows to one element using the judge's own rule, and both the number and the head come from it.

**A reason, not merely a failure.** The judge exits 2 with no reason when its input is missing (an unset lookup is the alarm being *broken*, not a finding), and a failed API step leaves the same shape. Either filed the default merge-accusing issue with nothing behind it.

**Wiring checks for all three**, since two of them are properties of the *query* rather than the payload, and a judge fed the wrong data agrees with it.

One thing worth reporting from writing them: my first version matched `filter=all` and `min_by(.number)` as plain strings, and **passed with the code removed** — because both phrases appear in the comments explaining them. Caught by mutating the code rather than trusting the check. They now match the query string and the jq expression:

```
M1 drop filter=all from the query  → FAIL: merge-attest asks for check runs without filter=all…
M2 revert to .[0] selection        → FAIL: merge-attest does not narrow the commit's pull requests…
M3 drop the reason requirement     → FAIL: …runs on a failed verdict without requiring a REASON.
```

`make check-backend` exit 0; `test-check-merge-verdict` and `test-scheduled-report` green.

The `uat` failure is unrelated and pre-existing — the vite dev proxy failing on `/.well-known/oauth-protected-resource`. That lane is advisory and deliberately outside the `ci` aggregate.

dungbanguyen:
Valid, and it defeats the fix from the last round rather than being independent of it.

`filter=all` makes every **attempt** of every check count toward `per_page`, so a pull request re-run a few times passes 100 — and the rows that get dropped are the tail, which is exactly what the judge reads. It takes the **oldest** attempt so that a re-run after the merge cannot clear the record of the merge it was absent for. Unpaginated, the rule `filter=all` exists to enforce was enforceable only up to the first hundred rows.

Now `gh api --paginate --slurp`, with `.[].check_runs[]` to flatten. `--slurp` because `--paginate` alone emits one JSON document per page, which is not a value jq can fold.

**One ceiling remains, and I have named it in the workflow rather than quietly inheriting it:** this endpoint stops at the 1000 most recent check *suites* for a ref. Reaching that needs a commit re-run into four figures, and the alternative — walking the check-suites endpoint — spends more request budget on every push to cover a case nothing here has approached. If that trade should go the other way, it is a one-line change and I will make it.

Held by a wiring check, since this is a property of the query rather than of the payload the judge sees:

```
mutation: drop --paginate --slurp →
  FAIL: merge-attest reads check runs without --paginate --slurp. per_page caps at 100 and
        filter=all counts every ATTEMPT, so a re-run pull request passes it — and the tail
        that gets dropped is the OLDEST attempt, which is the one the judge reads.
```

`test-check-merge-verdict` and `test-scheduled-report` green.

Issue #3710, opened 2026-09-02
main is red: App.tabkey.test.tsx "no tab is current" since #3683


LarsGradion:
`make check-fe` fails on `main` and `main-health` is red.

```
FAIL  src/App.tabkey.test.tsx > the routed subtree's key >
      keeps the record's chrome while a tab change swaps the panel
Error: no tab is current
```

Bisected by hand: the test passes at `d431f5bc2` and fails at current main. `6642bd709` (#3683) is the only commit between them touching `App.tabkey.test.tsx`, and its own message says the guard had been "passing for the wrong reason since the tab strip was extracted" — it repointed `currentTab()` from the slot to the strip that actually draws the buttons.

So the tightened assertion is almost certainly right and it has surfaced a real defect: after a tab change, no button carries `aria-pressed="true"` on the strip. Either the strip is not receiving the current tab, or it sets currency by some other attribute now.

Reproduce with `cd frontend && npx vitest run src/App.tabkey.test.tsx` — it fails in isolation, so this is not the load-flakiness of #3677.

Filed rather than fixed here: the branch that found it touches capture settings and privacy, nothing under `App.tsx` or the record tab strip, and the fix belongs with whoever repointed the guard.

dungbanguyen:
Same failure as #3971 and #3675, and PR #3982 fixes it. The measurement is on #3675: the /360 read settles in ~500-800ms against Testing Library's un-overridden 1000ms default, so the wait was late rather than absent, and `make check-fe` runs enough suites in parallel to cross the line.

Worth noting against this ticket's own framing: it reads as a regression from #3683. It is not — `App.tabkey.test.tsx`'s waits have ridden the default budget since the file was written, and #3683 only changed how often the machine was busy enough to expose it. That distinction matters if anybody is still considering a revert.

dungbanguyen:
Fixed by #3982, merged 2026-09-04. Verified against current `main`: `App.tabkey.test.tsx` passes alone and passes beside `App.test.tsx` — the pairing this ticket's own reproduction called out — 24 of 24.

The correction already recorded here is the part worth keeping: this reads as a regression from #3683 and it is not one. `App.tabkey.test.tsx`'s waits had ridden Testing Library's un-overridden 1000ms default since the file was written, against a /360 read that settles in ~500–800ms; #3683 only changed how often the machine was busy enough to cross that line. #3982 states the budget the waits actually need.

Closing as a duplicate of #3675, which carries the measurement and stays open for the cache-leak question its title names.

Issue #3675, opened 2026-09-02
App.tabkey.test.tsx: a react-query cache leaks between App suites, so the person page never loads


LarsGradion:
## Where this stands

Three separate defects were tangled here. Two are fixed; the third is this issue.

**Fixed (PR to follow): the suite stubbed one read where the page makes two.** The person page asks for `/v1/people/{id}` before `/v1/people/{id}/360`, and `personFetch` answered only the second. The first 503'd, so the page sat at "Loading…" and both cases failed on their first `waitFor`.

**Fixed: the test was querying a class that holds no buttons — and had been passing for the wrong reason.** `currentTab()` looked for `.record-tabs button[aria-pressed="true"]`. `.record-tabs` (composed.tsx:811) is the *slot* the record head wraps around whatever tabs it is handed; the tabs themselves are drawn by `design-system/recordtabs.tsx` under `.recordtabs-strip` / `.recordtabs-tab`. So that selector matched nothing.

Worse, `tabStrip()` returned the same `.record-tabs` wrapper — a node that survives any remount. Its identity assertion (`expect(tabStrip()).toBe(stripBefore)`) therefore proved nothing about the routed subtree's key. With the selector corrected, mutating `IDENTITY_DEPTH.contacts` from `2` to `WHOLE_ADDRESS` now fails both cases; before the fix, that mutation went undetected.

## What remains, and why it needs an owner

With both fixed, the file passes alone and still fails when another App suite runs first:

```
pnpm exec vitest run src/App.tabkey.test.tsx                    → 2 passed
pnpm exec vitest run src/App.test.tsx src/App.tabkey.test.tsx   → fails
```

**This is pre-existing.** On an unmodified tree that same pair already fails one of the two — so the leak is not introduced by the fixes above; they only make it visible in both cases instead of one.

What the failing run shows: the address is correct (`#/contacts/p-1/overview`), the record's name resolves ("Person p-1"), and the page still shows "Loading…". So the `/360` read never settles. `renderApp` builds a fresh `QueryClient` per mount, so a cache is being held somewhere above it that a fresh client does not clear.

That is a defect in the harness or in a module-level cache, not in this file, and it plausibly weakens every App-level suite rather than only this one — which is why it is worth someone's time rather than a local workaround.

## Why it matters beyond one file

[...]

PR #3982, opened 2026-09-04
The tab-key suite states the budget its waits actually need

### LINKED ISSUE #3971: App.tabkey.test.tsx is flaky: "no tab is current"
`src/App.tabkey.test.tsx` fails intermittently with `Error: no tab is current`.

Two identical runs of the same file on the same commit gave different results — three failures, then one. It also fails the same way on `78d8807db`, which predates the merges I was testing against, so it is not a regression from recent work.

It surfaces in `make check-fe` as a single failing test out of ~6215, which reads exactly like a real defect and costs a re-run to dismiss.

**To reproduce:** `cd frontend && pnpm exec vitest run src/App.tabkey.test.tsx` a few times.

The failure is in the tab-chrome assertion — the test asks which tab is current before the routed subtree has settled, so it is most likely a missing wait rather than a product bug. Worth confirming that before treating it as one.

dungbanguyen:
Closes #3971, and it is the same defect as #3675 and #3710 — with a different cause than any of the three recorded.

## What it was read as

`App.tabkey.test.tsx`'s own header said:

> KNOWN GAP, #3675: these two cases pass alone and fail when another App suite has run first. The person page then never leaves "Loading…" — the address is right and the record's name resolves, so **the /360 read is what does not settle**, and something above this file is holding a react-query cache across suites that `renderApp`'s fresh client does not clear.

The symptom is exactly that. Instrumented on a failing run, the DOM carries the breadcrumb `Contacts/Person p-1` — so the bare record read landed — and then `Loading…`, with no `.record-head` and no `.recordtabs-strip`.

## What it is

The read settles. It is late, not absent, and the two want opposite fixes.

Measured, with the wait instrumented:

| | time to the first current tab |
|---|---|
| the file alone | 532, 480, 494 ms |
| beside `App.test.tsx` | 602, 608, 678 ms |
| the failing run | died at 1032 ms |

Testing Library's `asyncUtilTimeout` is **1000 ms** and this repo overrides neither it nor `vi.waitFor`'s — `vitest.budget.ts` says so in as many words. So the budget sits a quarter of a second above the work, and `make check-fe` runs far more than one suite alongside it. Lifting the waiter to 8 s makes the same previously-failing run pass, which is what separates "never arrived" from "arrived late" — and `frontend/AGENTS.md` is right that the first is the usual answer, which is why I measured instead of assuming.

There is no cache leak. `renderApp` builds a fresh `QueryClient` per mount, and nothing survives the file boundary; the correlation with "another App suite ran first" is contention, not state.

## The fix, in the shape the repo already has for this

`vitest.budget.ts` records this exact arithmetic (#1144) and names the files that opt out of the default population by stating their own waiter and per-test budgets — `integrations-provider.test.tsx` among them. This file joins them:

- `SETTLE_MS = 10_000` on each wait, because a test may not ride a budget a quarter above its measured work.
- `TWO_SETTLES_MS = SETTLE_MS * 2 + SLOWEST_MEASURED_TEST_MS` as the per-test ceiling, since both cases wait twice and a test may spend the **sum** of its waiters without any one of them failing. The allowance for the work between the waits is the same measured constant the suite ceiling uses, so the two rest on one measurement rather than a number picked here.

`scripts/test-budget.test.ts` is what keeps that honest, and it is not decoration: dropping the per-test ceiling while keeping the waiters fails it with `src/App.tabkey.test.tsx:173 spends 20000ms of waiter budget under a 10437ms ceiling`.

The file's header note is rewritten, because a wrong diagnosis left in place is worse than none — the next reader would have gone looking for a cache.

## Mutation-checked

The flake needs real contention, so it was reproduced the way `vitest.budget.ts` measured its own numbers — under deliberate load (80 spin loops):

- **default budget, under load:** `Tests 2 failed | 20 passed (22)`
- **these budgets, same load:** `Tests 22 passed (22)`

## Also closes, or wants a look

#3675 and #3710 are the same failure under different headings. #3675's title states the cache-leak diagnosis this disproves; I have commented there rather than closing someone else's ticket from a PR body.

**Verification.** `make check-fe` green (6,116 tests). `make check-backend` green.

## Summary by CodeRabbit

- **Tests**
  - Improved routing test stability by allowing sufficient time for chained application reads to complete.
  - Added explicit per-wait and per-test time limits based on measured test performance.
  - Documented the diagnosis of loading delays to support more consistent test validation.

- **User Impact**
  - No changes to application functionality or user-facing behavior.


dungbanguyen:
Rebased onto `main`, which had moved under this: #3805 landed a third case on `App.tabkey.test.tsx` — *"resolves the record again on a second mount of the whole app"* — written for the cache leak.

**Kept it, and applied the budgets to it too.** The case is worth having whatever the cause was: nothing else in the tree asserts that a remount re-reads, and a module-level cache that survived one would be a real defect. What it is not is the reason this file flakes, and it rides the same un-overridden one-second default the other two do, so on `main` it is a third case that will fail under load rather than a fix for the two that already do.

The header note now says both things: why the second-mount case exists, and — separately — what the measurement showed. All three cases take `SETTLE_MS` per waiter and `TWO_SETTLES_MS` as their own ceiling.

Re-verified after the merge: `23 passed` for `App.tabkey.test.tsx` + `App.test.tsx` under deliberate load (80 spin loops), where the default budget gives `2 failed | 20 passed`. `make check-fe` green at 6,258 tests. `scripts/test-budget.test.ts` green, which is what holds the per-test ceilings against the waiters they now compose.

PR #4011, opened 2026-09-04
main is red: two breakages, both from PRs whose gates were not extended

dungbanguyen:
`main` is red twice over. Both failures are on every open PR, and each one blocks the other's CI, so they land together.

## 1. `compose/attention` — the cross-source fixtures predate the engagement rule

`go test ./internal/compose/attention` fails on `origin/main`.

#3995 (5de83df4d) added the unproven-wait demotion in `classifyWaiting`:

```go
unproven := !waiting.Engaged && !waiting.HasOpenDeal
if unproven {
	level = levelRoutine
}
```

Every fixture in the package meaning *a customer genuinely waiting on us* was updated to say `Engaged: true` — `worklist_test.go`, `waitingfreshness_test.go`. `crosssourcerank_test.go` builds its `WaitingCustomer` by hand and was missed, so both of its cases now describe an unproven wait:

- `TestTheDayIsOrderedAcrossItsSourcesAndNotOnlyWithinThem` — `ranked [meeting promise risk notice …a001], wanted [meeting …a001 promise risk notice]`.
- `TestSystemNewsNeverLeadsACustomerWaiting` — the page leads with the automation row, which is the thing that test exists to refuse.

Both rows span levels on purpose. Demoting one to routine quietly changed the subject of each test. Fixed by saying `Engaged: true`, as every sibling fixture does.

Production is unaffected: `attentionwaitingseam.go:113` carries `Engaged: row.Engaged` from the query, and it is the only construction of `attention.WaitingCustomer` outside tests.

## 2. `compose/integration` — two report tools were registered but never invoked

`TestToolAnswersReachableWithoutApprovalSatisfyTheirSchemas` fails naming them:

> `compose_analytics_report` is registered but never invoked here, so nothing holds its result to the schema or the envelope.

#3999 registered `compose_analytics_report` and `describe_report_blocks` without extending the conformance lane. Adding the two calls found a real defect behind the first.

**`ComposeAnalyticsReportResult.Blocks` advertised an object and answered with an array.** It held the rendered blocks as one `json.RawMessage`, and `schemaFor` describes a `json.RawMessage` as `{"type":"object"}` — deliberately, and honestly, for a document whose members are another module's. It is wrong for a list. So the published `OutputSchema` said `blocks` was an object while every response carried an array, and a client validating against the schema this server advertises rejects all of them.

Held as `[]json.RawMessage`, the derived schema becomes an array of opaque blocks — what a caller actually receives, with the block kinds still free to vary inside it, which is the property the raw type was chosen for. `testdata/outputshapes.json` and `docs/reference/mcp-info.*` are regenerated from the types, and carry only this change.

`compose_analytics_report` is invoked with a text-only `title` block: the figure-bearing kinds resolve against a run, and what this lane holds is the envelope, which is the same either way.

## Verification

- `go test ./internal/compose/attention` — was FAIL on `origin/main`, now `ok`.
- `make test-it DIR=backend/internal/compose/integration RUN=TestToolAnswersReachableWithoutApprovalSatisfyTheirSchemas` — was FAIL, now `ok`, with both new tools passing.
- `go test ./internal/compose/ ./internal/modules/agents/` — green, including the two golden-file gates.

https://claude.ai/code/session_019b1jSBftJqGvaQ7ejTgnQ6

## Summary by CodeRabbit

- **New Features**
  - Analytics reports now return rendered report blocks as an array, allowing multiple blocks to be handled consistently.
  - Report output schemas clearly describe each block as an object within the array.

- **Documentation**
  - Updated the MCP reference documentation and catalog to reflect the revised analytics report response format.

- **Bug Fixes**
  - Improved handling of individual report blocks, including clearer error reporting when a block cannot be encoded.

### DIFF
--- a/backend/internal/compose/analyticsreporttool.go
+++ b/backend/internal/compose/analyticsreporttool.go
@@ -54,15 +54,20 @@ func analyticsReportComposer(pool *pgxpool.Pool, floor analyticsquery.Floor) age
 			return nil, err
 		}
 
+		// Encoded one block at a time rather than as one array, because the
+		// result carries them as a slice — see ComposeAnalyticsReportResult
+		// for why the advertised schema depends on which of the two it is.
+		//
 		// Empty, never nil: a document that rendered no blocks cannot happen —
 		// Validate refuses one — but a nil here would marshal to null, which a
 		// model reads as "unknown" rather than as "none".
-		if blocks == nil {
-			blocks = []RenderedBlock{}
-		}
-		encoded, err := json.Marshal(blocks)
-		if err != nil {
-			return nil, fmt.Errorf("compose: rendering a composed report: %w", err)
+		encoded := make([]json.RawMessage, 0, len(blocks))
+		for _, block := range blocks {
+			raw, err := json.Marshal(block)
+			if err != nil {
+				return nil, fmt.Errorf("compose: rendering a composed report: %w", err)
+			}
+			encoded = append(encoded, raw)
 		}
 		return json.Marshal(agents.ComposeAnalyticsReportResult{Blocks: encoded})
 	}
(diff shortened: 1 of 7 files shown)


dungbanguyen:
Cross-reference: #4018 was opened independently from a parallel session with the same `crosssourcerank_test.go` fix. This PR carries that change plus the tool-conformance breakage, which also fails on every open PR; noted there that the two have to land together.

PR #4018, opened 2026-09-04
fix(attention): main is red since #3995 — the cross-source fixtures assert the demoted case

dungbanguyen:
`main` has been red since [#3995](https://github.com/margince/margince/pull/3995).
Two tests in `internal/compose/attention` fail on a clean checkout of the tip:

```
--- FAIL: TestTheDayIsOrderedAcrossItsSourcesAndNotOnlyWithinThem
    ranked [meeting promise risk notice …a001], wanted [meeting …a001 promise risk notice]
--- FAIL: TestSystemNewsNeverLeadsACustomerWaiting
    the page led with [automation capture sync …a001] over a customer waiting two hours
```

In both, the waiting customer ranks **last** — below a hygiene notice in the
first, below three system rows in the second.

## Root cause

Bisected: green at `5de83df4d^`, red at `5de83df4d` (#3995). I reproduced both
against `origin/main` in a separate worktree before touching anything, so this
is main's state and not a local artefact.

#3995 added the rule that a wait with nothing to prove we ever wrote, and no
money on the thread, leaves the top band (`classify.go`):

```go
unproven := !waiting.Engaged && !waiting.HasOpenDeal
if unproven {
    level = levelRoutine
}
```

`Engaged` is a bool, so **a fixture written before the field existed asserts the
demoted case by saying nothing.** That is what happened here: the two
`WaitingCustomer` fixtures in `crosssourcerank_test.go` predate the field, so
they now classify as routine and sort beneath hygiene — which is exactly what
the ordering below them says must never happen.

#3995 knew this. It updated the `WaitingCustomer` fixtures in `worklist_test.go`
for precisely this reason, one of them carrying the comment:

> A conversation we are already in. Only such a wait leads the day — one where
> nothing proves we ever wrote is demoted, and **a fixture leaving this false
> would be asserting the opposite claim.**

It just did not reach `crosssourcerank_test.go`. I checked all 41
`WaitingCustomer` construction sites in the tree: those two are the only ones
that both omit `Engaged` and assert the top band, which is why exactly two tests
fail and not thirty.

## Why the fixtures and not the rule

The rule is deliberate and well covered — `waitingfreshness_test.go` asserts the
demotion, the `no_reply_history` reason, and the engaged case that must not
carry it. Its comment accepts the cost explicitly: *"DEMOTED, never dropped …
the cost of being wrong is a scroll instead."* An unproven wait sitting in the
routine band is that scroll, by design.

So the product is right and the fixtures were stating the opposite of what their
own assertions demand. Both now say `Engaged`, with the reason beside them —
a reader meeting `Engaged: true` in a ranking test needs to know it is load
bearing, not decoration.

## Verification

- `make check` from the repo root — green, both halves, 8m54s, zero failures
- `go test ./internal/compose/attention/` — ok
- Nine added lines, all in one `_test.go`; no production code changed

## Summary by CodeRabbit

* **Tests**
  * Updated waiting-customer test scenarios to represent previously answered conversations.
  * Ensured customer-priority ranking assertions use the intended conversation state.

### DIFF
--- a/backend/internal/compose/attention/crosssourcerank_test.go
+++ b/backend/internal/compose/attention/crosssourcerank_test.go
@@ -57,6 +57,11 @@ func TestTheDayIsOrderedAcrossItsSourcesAndNotOnlyWithinThem(t *testing.T) {
 			Subject:    "Can you confirm the retrofit price?",
 			Since:      rankInstant.Add(-72 * time.Hour),
 			PersonID:   waitPerson,
+			// A conversation we are already in, which is what earns this row
+			// the top band at all: a wait with nothing to prove we ever wrote
+			// is demoted to routine, and the order below would then be
+			// asserting that instead of what it says.
+			Engaged: true,
 		}},
 		crmcontracts.Attention{
 			AsOf:        rankInstant,
@@ -90,6 +95,10 @@ func TestSystemNewsNeverLeadsACustomerWaiting(t *testing.T) {
 			Subject:    "Still waiting on that quote",
 			Since:      rankInstant.Add(-2 * time.Hour),
 			PersonID:   waitPerson,
+			// The customer in the name is one we have answered before. An
+			// unproven wait is routine by design, so leaving this false would
+			// make the test demand the opposite of the rule it is checking.
+			Engaged: true,
 		}},
 		crmcontracts.Attention{
 			AsOf:             rankInstant,


dungbanguyen:
Same diagnosis, same one-file fix — this overlaps #4011, which I opened at 02:50Z (this one at 03:17Z) from a parallel session working the backlog descending. Apologies for the collision; claiming a *ticket* does not reserve a *red build*, and this is the third time the two of us have arrived at the same red main independently.

#4011 carries the identical `Engaged: true` change to `crosssourcerank_test.go`, plus a **second** breakage that also fails on every open PR and is not covered here:

```
--- FAIL: TestToolAnswersReachableWithoutApprovalSatisfyTheirSchemas
    compose_analytics_report is registered but never invoked here, so nothing
    holds its result to the schema or the envelope.
    describe_report_blocks is registered but never invoked here, …
```

#3999 registered both tools without extending the conformance lane. Adding the two calls surfaced a real contract defect behind one of them: `ComposeAnalyticsReportResult.Blocks` was a `json.RawMessage`, which `schemaFor` describes as `{"type":"object"}`, while every response carries an **array** — so the tool advertised a shape it never answers with, and a client validating against the published schema rejects all of them. #4011 fixes that too and regenerates the two golden files.

Because each breakage fails the other's CI, they have to land together — #4011 green is what actually turns main green.

Suggest closing this one in favour of #4011, but it is your call: if you would rather carry both fixes here, say so and I will close #4011 instead. Either way the second breakage needs to be in whichever one lands.

dungbanguyen:
Superseded by #4011, which landed the identical fix 14 minutes after this opened — the same `Engaged: true` on both cross-source fixtures, with the same diagnosis, plus the `compose_analytics_report` / `describe_report_blocks` census gap that was blocking this PR's integration shard.

Two sessions raced the same red main. Nothing here to rescue: main carries the fix and this branch now conflicts with it.

Issue #4003, opened 2026-09-04
Decide: may a scheduled extension tick make a governed core write under a member's live authority?


tinnhlt-gra:
A scheduled extension tick answers as the **job**: `PrincipalAgent`, no `UserID`, no
`SeatType`, no `Permissions`, and the one scope the unit's manifest declared. It lands
records through `Runtime.Ingest(on: member)`, which resolves that member's live grants
per record and discards the tick's principal entirely. Governed **core** writes are
refused outright — `extcore.go` returns `ErrUnattendedCoreWrite`, and the empty
`Permissions` denies every object at `auth.Require` even if that refusal were relaxed.

**The question:** should a tick be able to make a governed core write under a member's
live authority — a `Runtime.ForMember(ctx, on, fn)` or equivalent?

Re-opening the door `extcore.go` deliberately closed is a product decision, not a
refactor, which is why this is filed rather than fixed.

## What a design has to answer

A first draft proposed exactly this and was withdrawn. Four findings, each verified
against the code, and each still standing:

1. **The scope never reaches the thing it would have to convince.** `extensionCore` is
   built from fields on the `callRuntime` at `Tx` time, and `callRuntime.scoped`
   overwrites the actor on whatever ctx a handler passes with the actor on `r.callCtx`
   — deliberately: *"what a handler passes can shorten a deadline but cannot change who
   is acting."* A `ForMember` that decorates the handler's ctx is erased by the next
   `rt.Tx(ctx, …)`. This is precisely why `Ingest` works today: `ingressAuthority` runs
   *after* `scoped`, inside the runtime. The only honest mechanisms are a derived child
   `callRuntime` handed to `fn` — which forces the signature `fn func(ctx, rt Runtime) error`
   — or a scope token invalidated under `r.mu` on return.

2. **The declared scope would enforce nothing.** `auth.Require` reads `Permissions` and
   never looks at `Scopes`; scope enforcement lives in `Gate.Admit`, which the core port
   does not go through. A bound stated as "the member's live grants ∩ the declared scope"
   would be the member's live grants, with the manifest scope as decoration.

3. **One boolean cannot answer two doors.** `extingress.go` refuses an *attended* ingest;
   `extcore.go` refuses an *unattended* core write. Both read the same `r.unattended`.
   Inside such a scope, `Core` would have to see an authority while `Ingest` keeps seeing
   none. The draft also never said what `Ingest(on: B)` inside `ForMember(A)` does — the
[...]

PR #4030, opened 2026-09-04
A scheduled tick names the job it is, not a person

tinnhlt-gra:
A scheduled extension tick used to resolve an `is_agent` app_user row per workspace, record its id on every child job row, and re-read it at execution to act as that seat. **The seat authorized nothing.** Every unit that ships scheduled work lands its records through `Runtime.Ingest`, which resolves the *member's* live grants per record and discards whatever the tick was carrying.

What the resolution actually bought:

- a round trip per tenant per dispatch, inside the dispatcher's declared wall clock;
- a workspace **silently skipped** whenever that seat was archived — visible only as a gauge;
- an operator who deactivated "the AI" stopping **inbound message capture for every connected member**, which nobody asked for and no screen explained.

The tick now answers as the job it is.

## The change

```go
func extensionJobPrincipal(decl extension.JobDeclaration) principal.Principal {
    return principal.Principal{
        Type:   principal.PrincipalAgent,
        ID:     "agent:" + decl.DispatcherKind(),
        Scopes: principal.NewScopeSet(principal.Scope(decl.RequestedScope)),
    }
}
```

No user, no seat type, no permissions — a pure function of the declaration, so the queued row carries no principal that can go stale and the dispatcher reads no identity at all.

### `PrincipalAgent`, and emphatically not `PrincipalSystem`

This is the load-bearing decision. There are **two independent locks** on the governed core-write door, and only one of them is `extcore.go`'s refusal:

| lock | where | what `PrincipalSystem` does to it |
|---|---|---|
| `refuseUnattended` | `extcore.go` | unaffected — still refuses |
| `auth.Require` on empty `Permissions` | `auth/rbac.go` | **bypassed** — system returns nil before `Permissions` is read |

A system principal would also take `RowScopeAll`. Reaching for it here looks like a simplification and would delete a security control. **Nothing asserted that second lock**; `extjobprincipal_test.go` now does, and both its arms were watched failing — swapping the type to `PrincipalSystem`, and granting the principal one object.

`PrincipalAgent` also keeps `audit_log.actor_type = 'agent'`, which the `captured_by_kind=agent` AI-review lane reads. `PrincipalHuman` would break both readers at once.

## What this PR does NOT do

**The seeded seat row is left standing, guards intact.** Retiring it needs this binary deployed everywhere first: an old worker against a seatless database skips every workspace *silently*. The migration, the four identity refusals, the two CHECK constraints and the licence-count change are a second PR, opened once this one is deployed. So `Settings → License` is unchanged here.

The fan-out key does not move from workspace to user, `is_agent` and its wire field stay, and no `pkg/extension` capability is added.

## Two judgement calls worth reviewing

**`SyncNow` stays on the `Runtime` interface.** Only the seat check and `extension.ErrNoUnattendedSeat` go. Removing a published surface is a separate decision from retiring the seat, and this PR otherwise needs no downstream coordination at all. `pkg-freeze` reports the one incompatible change — `ErrNoUnattendedSeat: removed` — as an advisory, which is what it is pre-v1.

**`docs/evidence/extension-tier/{DESIGN,README}.md` are deliberately untouched**, though both mention the seat and the removed gauge. Each opens with *"Historical record … Read it as 'how the tier was proved', never as 'what ships'"*. Editing them would falsify the record of what a review actually checked.

## A published metric disappears

`margince_extension_job_seatless_workspaces` goes with the skip it reported. **This is a gap for anyone alerting on it, not a zero.** Nothing replaces it — the condition it gauged can no longer occur. Release note.

## Three claims that were already false on `main`

Fixed while here, since each would mislead the next reader:

1. The `passportlessagents` waiver said the tick's authority was *"the job owner's, read live from identity"*. There was no owner.
2. `pkg/extension/runtime.go` said additive interface growth is safe *"because extensions consume Runtime and never implement it"* — every unit's test fake implements it.
3. `docs/how-to/add-an-extension.md` cited `0216_agent_seat_backfill`, removed by the baseline squash.

## What review found, and what it cost

Adversarial passes ran before this was opened. The security review found **no exploitable finding** — it walked the whole tick call graph and confirmed the new principal is strictly weaker than the one it replaces (`SeatType` `"full"` → `""` is a tightening; zero `RowScope` is `""`, not `all`; `Gate.Admit` already denied an agent with zero `OnBehalfOf`).

The craft review found **a test that proved nothing**, and it is the finding worth reading:

> `TestAnInstallationWithNoAgentSeatStillTicks` archived the seat with `UPDATE app_user … WHERE is_agent` — and the compose harness seeds no agent seat. The statement matched **zero rows**. The test passed byte-identically with it deleted, while its own comment called the archive the premise.

It now seeds a seat and asserts the archive touched exactly one row. That guard was itself watched failing. The second commit also fixes six places still describing the deleted mechanism — including the confirm dialog an admin reads *at the moment they deactivate the seat*, which said "stops every job that runs with nobody behind it, extensions included" in all three locales, with a test pinning the sentence.

## Verification

| lane | result |
|---|---|
| `make check` (both halves) | green |
| `make test-it` — `internal/compose` | green, 192s |
| `make test-it` — `internal/compose/integration` | green, 355s |
| `make test-it` — `internal/modules/identity` | green, 24s |
| `craft static --strict`, diff-scoped | 0 blocker, 0 major, 0 minor |
| coverage, new code | `extensionJobPrincipal` 100% |

Every new and inverted fixture was watched failing against the case it names, not merely observed passing.

**UAT on a real dev stack** — video below. The agent identity is deactivated and archived while `openchannel`'s drain is cadencing at 60s; the next tick completes, with zero retryable or discarded rows. That is the whole change in one action.

Deferred half — whether a tick should be able to make a governed core write under a member's live authority — is #4003, with the four findings that killed the first design.

---
## Summary by cubic
Scheduled extension ticks no longer resolve the installation's agent seat to act as; they now answer as the job itself — `PrincipalAgent`, id `agent:<dispatcher kind>`, no user, no seat, no permissions. Deactivating the agent identity no longer stops inbound message capture, and a workspace with no seat is no longer silently skipped.

**Security**
- The tick answers as `PrincipalAgent`, not `PrincipalSystem`, because system bypasses the `auth.Require` denial on empty permissions and takes `RowScopeAll`.
- A new test pins the empty grant map as the second lock on governed core writes.

**Side effects**
- `margince_extension_job_seatless_workspaces` is removed with the skip it reported — a gap for anyone alerting on it.
- `extension.ErrNoUnattendedSeat` is removed; `SyncNow` stays on the Runtime interface.
- The seeded agent seat row stays until this binary is deployed everywhere; retiring it is a second PR.
- The deactivate-agent confirm dialog in all three locales no longer claims extension jobs stop.

<sup>Written for commit 7619930323441ffcac9a18ac988e4ebc4c8d3ef8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/4030?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Changes**
  - Scheduled extension jobs no longer require an agent seat and continue running after the agent identity is deactivated.
  - Jobs run under an unattended job identity without user permissions; records use the connected member’s live authority.
  - On-demand extension jobs run without a user identity or grants.
  - Removed obsolete seatless-workspace metrics and related error messaging.

- **Documentation**
  - Updated help content and deactivation warnings to explain the revised scheduled-job behavior.


tinnhlt-gra:
## UAT — the agent identity is deactivated, and the tick keeps running

Run against a real dev stack (`make dev`, worktree-isolated database, API restarted after the last backend change), with `extensions/openchannel` composed and its drain cadencing at 60s.

https://github.com/user-attachments/assets/31736dce-67d2-4735-9e64-cf3e7aa80b99

**What the recording shows, in order:**

1. The installation's agent seat: `agent@demo-workspace.gradion.local`, `status=active`, not archived.
2. `ext_openchannel_drain` and `ext_openchannel_drain_ws` both completing on the clock.
3. The enqueued child row — **no `principal_id` field at all**:
   ```json
   {"job_kind": "ext_openchannel_drain_ws", "workspace_id": "01a06a50-8d16-71f2-937a-9b96377a8009"}
   ```
4. **The operator action.** The agent identity is deactivated *and* archived — the state that, before this change, silently stopped every scheduled extension job in the installation, inbound message capture included.
5. 41 seconds later the next drain tick **completes**. Completed ticks go 2 → 3 with no live agent seat in the table.
6. Zero retryable or discarded rows across both kinds. The tick did not fail-and-retry into looking like it worked; it ran.
7. The seat is restored, since retiring the row is the follow-up PR's job, not this one's.

The equivalent assertion is held in CI by `TestAnInstallationWithNoAgentSeatStillTicks` (`internal/compose`), which seeds a seat, asserts the archive touched exactly one row, and then asserts the handler ran and the actor it ran under carries no `UserID`.

PR #4061, opened 2026-09-04
Liveness gets the fitness function it never had

### LINKED ISSUE #2202: No fitness function asserts that a write refuses an archived row
A write in this repo owes four obligations. Three have fitness functions:

| Obligation | Gate |
|---|---|
| writes only tables it owns | `backend/tableownership_test.go` |
| carries a concurrency guard | `backend/updateguard_test.go` |
| probes *write* authority, not visibility | `backend/writeauthority_test.go` |
| audits **and** emits | `backend/writeshape_test.go` |

The fourth — **liveness: may this write touch an archived or erased row?** — has none. `satellite_lifecycle_test.go` gates the archive *cascade* (children follow the parent down); nothing gates the *refusal*. The tree is 38 `EnsureWritable` call sites to 10 `EnsureWritableLive`, and which is correct is decided per call site by hand.

That one gap produced 8 defects across `organization`, `person`, `person_consent`, `attachment`, `contract`, `commission_entry` and `site_read` — every one individually defensible where it sits. It has also been found independently before and patched as a point fix: https://github.com/gradionhq/margince-poc-v1/issues/1398 is the same divergence in `contracts`, filed as `fast-track-debt`.

### The obligation to assert

A function that mutates a table carrying an `archived_at` column either carries the live predicate, probes a `*Live` spelling, or is ratified in a waiver with a written reason. `updateguard_test.go` is a direct template — same shape, same `gatekit.Waive` discipline, same derive-the-table-set-from-the-migrations approach.

### Enforcement is a gate, NOT a trigger

This deliberately overrules step 1 of https://github.com/gradionhq/margince-poc-v1/issues/2145, which proposed pushing the rule into the schema on the geocode-trigger precedent. **That precedent does not generalise.** Art. 17 erasure and the retention sweep *must* write archived rows — `archived_at = coalesce(archived_at, now())` is throughout those statements — so a row-level refusal breaks erasure. Liveness is a property of the call site, not of the row.

Migration `0290`'s shape (refuse an *incomplete* write, rather than any write) is the one that does generalise, and is worth considering later for the organization profile columns specifically. Not part of this.

### The waiver families, already derived

So they don't have to be re-derived from scratch — full reasoning in the handoff on https://github.com/gradionhq/margince-poc-v1/issues/2145:

- **Erasure and retention must write archived rows** — `privacy/erasuretimeline.go`, `erasure_restrict.go`, `restrictionoverride.go`, `retentionrestricted.go`, and `retentionactions.go`'s executors. This is the family that makes a trigger impossible; each waiver should say so rather than restating "terminal absolute write".
- **The archive transitions themselves** and their cascades over child tables.
- **Merge relinks** — `people/merge.go`, `merge_organization.go`, `mergerelink.go`. `mergePair`'s liveness lives in the `read` closure it is handed, which a scanner cannot see.
- **The four `RefuseArchive*` paths** — they write nothing; they are the 409 refusal.
- **`promotableLead`** — reads `IncludeArchived` on purpose so a re-promote answers 409 with the outcome pointer rather than a misleading 404, then refuses explicitly on `lead.ArchivedAt != nil`.
- **`ensurePairWritable`** — writes the dedupe queue row, not either record.
- **`resolveOrCreateAnchor`** — guarded in its helper; `anchorOrganization` carries `WHERE is_anchor AND archived_at IS NULL`.
- **`upsertLead`** — `ON CONFLICT … DO NOTHING`; the replay branch writes no lead column.
- **`capturenoise.go`'s redaction** — strips content from rows it has already archived, by design.

### Blocked by

The 8 fixes land first, so the gate arms clean.

dungbanguyen:
Closes https://github.com/gradionhq/margince-poc-v1/issues/2202.
Part of https://github.com/gradionhq/margince-poc-v1/issues/2145 — the highest-value item on that tracker, and the last one that needed the holes closed first (ruling 6). https://github.com/gradionhq/margince-poc-v1/pull/4057 closed them.

## The gap

A write in this tree owes **four** obligations. Three have fitness functions:

| Obligation | Gate |
|---|---|
| writes only tables it owns | `tableownership_test.go` |
| carries a concurrency guard | `updateguard_test.go` |
| probes *write* authority, not visibility | `writeauthority_test.go` + `writeauthorityreach_test.go` |
| audits **and** emits | `writeshape_test.go` |
| **liveness — may this write touch a row somebody archived or erased?** | **nothing** |

That one gap produced eight defects across `organization`, `person`, `person_consent`, `attachment`, `contract`, `commission_entry` and `site_read`. What they had in common was not a module or a table — every one was individually defensible where it sat. Nothing asked the question.

## The gate

`TestEveryByIDWriteOfARetirableRowAnswersForLiveness`: a function that writes **one row of an archivable table by id** must do one of two things.

- **REFUSE** an archived row — an `archived_at IS NULL` predicate, a `*Live` probe, a `LiveOnly` lock or patch.
- **DECLARE** that it reaches one on purpose — `auth.EnsureRetractable`, or `storekit.IncludeArchived`.

Anything else is ratified in `livenessUnstated` with **where the liveness actually is**. No entry reads "known defect, see issue" — ruling 6 — so a waiver here states that the obligation is met some other way, never that it is owed and unpaid.

**The declaration half is what makes this hold a line a scanner otherwise cannot.** "No liveness here" is *correct* for Art. 17 erasure, for the retention sweep, for a merge retiring its source, and for every retraction. Before `EnsureRetractable` existed, that case and a forgotten filter were the same text.

**Enforcement is a gate and NOT a trigger**, deliberately — this is ruling 5 on the tracker overruling its own issue body. The geocode-trigger precedent does not generalise: Art. 17 and the retention sweep *must* write archived rows (`archived_at = coalesce(archived_at, now())` runs through those statements), so a row-level refusal would refuse the destruction the archive is a step of. Liveness is a property of the call site.

**Scope** is single-row-by-id UPDATE and DELETE, the same boundary `updateguard_test.go` draws. A set-based write takes its liveness from the parent that selected it. An INSERT is out for a different reason: a row being created has no `archived_at` to violate, and the liveness its *anchor* owes is a separate question with a separate answer (https://github.com/gradionhq/margince-poc-v1/issues/1405).

The table set is derived from the migrations — every `CREATE TABLE` carrying `archived_at` plus the tables that grew it in a later `ALTER` — so a table that becomes retirable widens the census on its own.

## What it found

145 functions judged; 52 ratified. The waivers group into the families the tracker's handoff had already derived: erasure and retention must write archived rows; the retirement transitions themselves; the liveness is one frame up (named per entry); the write cannot reach a retired row because of what it is.

**Two could say nothing true, and were fixed:**

1. **`SetCaptureLabel` labelled a message archived while the model was thinking.** Its docblock promises the exclusions are re-tested at write time and re-tests two of three — the audience and the statutory hold — and not the archive. A noise disposition or the retention sweep takes the message off the timeline during the model call; the label lands after, and **nothing revisits an archived row to clear it**, because the backlog excludes them. What it leaves is a hidden message on a shared worklist. Same defect as the narrowing case the file already tests, different cause.
2. **`revertColumn` now takes `auth.EnsureRetractable`.** Taking bought data back off a contact is a retraction, and `RevertProviderFills` already holds the row with `IncludeArchived` for exactly that reason. The two probes run the same statements, so **behaviour is unchanged** — what changes is that the call site says which side it took, and a live probe there would have refused the privacy control on every retired contact.

## Anti-vacuity

Rule 8 says a census that can fail short has already failed, so:

- Two floors: `byIDWriteFloor` (the walk or the statement reader going quiet) and `retirableTableFloor` (the migration derivation reading nothing, which would exempt the entire tree while reporting a clean run).
- `livenessUnstated.AssertAllMatched` names any ratified function the reader stops seeing.
- `writelivenesscases_test.go` drives both halves from **synthetic source**, because a reader proven only by "the tree is clean" keeps passing after it stops working. 14 cases: what the census must SEE (a by-id DELETE, an aliased `a.id = $1`, a statement hoisted to a package-level table — the shape the organization column writers actually use) and what it must NOT credit (`archived_at IS NOT NULL`, which is the *restore* arm's claim; a marker named only in a comment; a set-based write).

One credit is deliberately looser than `updateguard`'s: the marker is **not** attributed to the table being written. It cannot be — the liveness that governs a child write is frequently its anchor's, so requiring probe and statement to name one table would refuse the shape this rule is mostly about. The cost is written down where the code makes the trade.

## Tests

`TestTheLabelWriteLosesToAnArchiveThatLandedWhileTheModelThought` — beside its narrowing twin, same fixture, same race. Verified it fails against the unfixed statement and passes with it.

Verified: `go test ./gates/` (full suite), `go test ./internal/...`, the `activities` and `compose/integration` integration lanes, `craft static --strict` over the diff.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented labels from being applied to activities archived while processing is in progress.
  * Allowed retraction of titles on archived contacts where appropriate.

* **Tests**
  * Added coverage to verify archive-aware write behavior and liveness handling across updates and deletes.

* **Documentation**
  * Clarified archive-aware write rules and updated the gate inventory to reflect new validation coverage.

### DIFF
--- a/backend/gates/writeliveness_test.go
+++ b/backend/gates/writeliveness_test.go
@@ -247,37 +247,73 @@ func retirableTables(t *testing.T) map[string]bool {
 }
 
 // byIDWritesIn names the retirable tables these statements write one row of by
-// id. A table outside the set answers nothing about liveness because it cannot
-// hold an archived row.
+// id, and says of each whether some statement writing it REFUSES an archived row
+// in its own text. A table outside the set is absent: it cannot hold an archived
+// row, so it answers nothing.
+//
+// Per TABLE rather than per statement, twice over and deliberately.
+//
+// statementsJudged returns overlapping READINGS of one value — a concatenation
+// is recorded both folded and in parts — so a fragment carrying the SET without
+// the WHERE would read as an unguarded write of a statement that is guarded.
+//
+// And the refusal is credited from any statement NAMING the table, not only from
+// the write. The commonest correct shape resolves the row live in a SELECT and
+// pins the UPDATE to it with a version or an archive transition, so insisting the
+// predicate ride the write itself would ask thirteen correct functions to be
+// waived. What it still separates is the case that matters: a function that
+// guards its organization write and leaves an activity write bare names no
+// activity statement carrying the predicate, so the activity write is reported.
 func byIDWritesIn(statements []string, retirable map[string]bool) map[string]bool {
 	written := map[string]bool{}
+	refused := map[string]bool{}
 	for _, lit := range statements {
-		for _, m := range byIDWriteOf.FindAllStringSubmatch(lit, -1) {
-			if retirable[m[1]] {
-				written[m[1]] = true
+		if livePredicate.MatchString(lit) {
+			for _, table := range tablesUnderLivePredicate(lit) {
+				refused[table] = true
 			}
 		}
+		for _, m := range byIDWriteOf.FindAllStringSubmatch(lit, -1) {
+			if retirable[m[1]] {
+				written[m[1]] = false
+			}
+		}
+	}
+	for table := range written {
+		written[table] = refused[table]
 	}
 	return written
 }
 
-// statesLiveness reports whether this function answers the liveness question at
-// all, either way. It reads the function's own frame: the statements it runs
-// and the names it calls.
-//
-// The marker is deliberately NOT attributed to the table being written, which is
-// where updateguard's lock credit is stricter. It cannot be: the liveness that
-// governs a child write is frequently its ANCHOR's — a deal room's deal, a
-// contract's organization — so requiring the probe and the statement to name one
-// table would refuse the shape this rule is mostly about. What that costs is a
-// function whose marker belongs to some other row; what insisting would cost is
-// a waiver on every correct anchor-gated write in the tree.
-func statesLiveness(fn *ast.FuncDecl, statements []string) bool {
-	for _, lit := range statements {
-		if livePredicate.MatchString(lit) {
-			return true
-		}
+// statementTable follows every keyword that introduces one, so a live predicate
+// is credited to the tables its own statement reads or writes and to no others.
+var statementTable = regexp.MustCompile(`(?i)\b(?:UPDATE|FROM|JOIN|INTO)\s+(?:ONLY\s+)?([a-z_]+)`)
+
+// tablesUnderLivePredicate names the tables one statement touches, for a
+// statement that refuses an archived row.
+func tablesUnderLivePredicate(statement string) []string {
+	var named []string
+	for _, m := range statementTable.FindAllStringSubmatch(statement, -1) {
+		named = append(named, m[1])
 	}
+	return named
+}
+
+// statesLivenessInGo reports whether this function answers the liveness question
+// in its Go frame — a *Live probe or lock, or a declaration that it reaches an
+// archived row on purpose.
+//
+// A Go marker is NOT attributed to the table being written, which is where
+// updateguard's lock credit is stricter. It cannot be: the liveness that governs
+// a child write is frequently its ANCHOR's — a deal room's deal, a contract's
+// organization — so requiring the probe and the statement to name one table
+// would refuse the shape this rule is mostly about.
+//
+// The IN-STATEMENT half is attributed, in byIDWritesIn above, and that is the
+// half where it can be. A function that guards its organization write and
+// leaves its activity write bare answers for one table and not the other; the
+// two are asked separately, so the second is still reported.
+func statesLivenessInGo(fn *ast.FuncDecl) bool {
 	stated := false
 	ast.Inspect(fn.Body, func(n ast.Node) bool {
 		switch node := n.(type) {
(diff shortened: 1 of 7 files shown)


dungbanguyen:
Addressed the review finding in `f799e8bd2`, and it was right about a real gap.

**What it caught:** the credit was function-scoped in both halves, so a function updating `organization` with `archived_at IS NULL` and `activity` with a bare `WHERE id = $1` answered for both, and the activity write went unreported.

**What changed:** the IN-STATEMENT half is attributed now — a refusal credits the tables its own statement names, and no others. The Go half stays function-scoped, and that is not a concession to convenience: the liveness governing a child write is frequently its ANCHOR's, and a probe on a deal cannot be made to name the `deal_room` row it protects.

**One thing the suggested shape would have cost, and how it is kept:** attributing the refusal to the WRITE alone made thirteen correct functions into findings — `DeleteSource`, `automation.Archive`, `DeactivateUser`, `PromoteOrgNameTx` and nine more. The commonest right shape resolves the row live in a SELECT and pins the UPDATE to it with a version predicate or an archive transition, so the predicate sits one statement above the write it governs. Crediting from any statement NAMING the table keeps those out of the waiver list while still separating exactly the case the review found: a function that names no statement of the second table under a live predicate still answers nothing for it.

The falsification suite carries that case now, from synthetic source — one guarded write and one bare sibling in the same function, asserting the bare one is still reported.

Issue #2761, opened 2026-08-26
A statement hand-writes updated_at and version onto tables whose trigger already sets both


dungbanguyen:
## In one sentence

Statements hand-write `updated_at = now()` and `version = version + 1` onto
tables whose trigger already sets both, so the assignment is dead — and the next
writer of that table copies it, or worse, notices a sibling statement without it
and concludes the two arms differ.

## What is going on

`0001_baseline.up.sql` installs `set_updated_at()` and
`set_updated_at_bump_version()` as `BEFORE UPDATE ... FOR EACH ROW` triggers on a
long list of tables:

```sql
CREATE FUNCTION set_updated_at_bump_version() RETURNS trigger AS $$
BEGIN
  IF current_setting('margince.last_activity_move', true) = 'on' THEN
    RETURN NEW;
  END IF;
  NEW.updated_at = now();
  NEW.version = OLD.version + 1;
  RETURN NEW;
END;
$$;

CREATE TRIGGER trg_organization_updated BEFORE UPDATE ON organization
  FOR EACH ROW EXECUTE FUNCTION set_updated_at_bump_version();
```

A `BEFORE ROW` trigger runs after the statement's `SET` has been applied to
`NEW`, so it **overwrites** whatever the statement assigned to those two columns.
The early return is for last-activity moves only, and nothing in the Go tree sets
that GUC — it is set and cleared inside the trigger functions that move
`last_activity_at`.

So on any trigger-touched table, a statement's own `updated_at = now()` or
`version = version + 1` is dead. Not wrong — dead. Same value, assigned twice.

## Why this is a problem and not just untidy

Two confirmed instances, both in `internal/modules/people/`, both found while
working #2145:

- `coldstartprofile.go` — `writeOrgColumn`'s four overwrite statements carry
  `updated_at = now()`. Its sibling **fill** statements in `coldStartColumns` do
  not. Reading the two arms side by side, the difference looks meaningful. It is
  not, and I nearly preserved it as intentional while merging them.
- `company.go` — `resolveOrCreateAnchor`'s
  `UPDATE organization SET display_name = $2, version = version + 1`. This is
  #2145's item 2 verbatim ("one manual version bump"), and it turns out to be
  moot rather than a fourth concurrency mechanism.

That second one matters for how #2145 gets read: the tracker counts *four*
disagreeing approaches to concurrent editing across the organization writers, and
one of the four is not an approach at all.

The file already knows this shape is a hazard — `company.go` carries the note:

[...]

Issue #2574, opened 2026-08-24
A liveness probe and the write it guards are two statements, and nothing closes the gap


dungbanguyen:
Every `auth.EnsureWritableLive` call site in `backend/internal/modules/**` is probe-then-write inside one transaction. Under READ COMMITTED the probe reads a snapshot and the write happens later, so an archive or an Art. 17 erasure committing between the two lands anyway. Roughly twenty sites take that risk, and none of them says so.

Raised by adversarial review on #2559 and #2562, and filed rather than fixed there because the answer is systemic and those PRs are not the place to change a shared primitive.

## Why it is worth closing rather than noting

For most sites the residue is a stale write. For two it is not:

- `consent/doi.go` — a `consent_doi_token` minted in that window is a **working invitation to grant consent**, in the post, for an erased person.
- `people/searchpersonfields.go` — a `person_profile_field` row written there puts the subject's details back after an erasure cleared them.

Both now carry a `WHERE EXISTS (SELECT 1 FROM person WHERE id = $1 AND archived_at IS NULL)` predicate on the INSERT. **That is a narrowing, not a fix.** It closes probe→write; it does not close write→COMMIT, where an erasure still wins. It is also a novel idiom in this tree: those are the only two statements carrying one, so the invariant now has two writers and eighteen abstainers.

## The shape of the real fix

A row lock rather than a predicate, taken once in the primitive rather than spelled per statement — `SELECT … FOR NO KEY UPDATE` on the subject inside `EnsureWritableLive`, so a concurrent archive blocks until the write commits instead of racing it. `FOR NO KEY UPDATE` rather than `FOR UPDATE` because the probe must not block ordinary FK references to the row.

Two things to settle before building it:

1. **Lock ordering.** `people` already has a documented profile-then-build / name-then-row order, and `renamerecheck.go` records a deadlock found only by review when a row lock was taken before the workspace-wide name lock. A lock inside `EnsureWritableLive` would be taken at the top of many transactions and has to be checked against that ordering — this is the reason the same suggestion was declined on #2559 for the enrichment paths specifically.
2. **Cost.** It adds a lock to every live-probed write. Worth measuring against the twenty sites before adopting.

## Scope

- Decide lock-vs-predicate for `EnsureWritableLive`.
- If the lock lands, remove the two one-off predicates so the invariant has one writer again.
- If it does not, say so at the primitive and let the two predicates stand as the declared exceptions with their reason.

PR #4072, opened 2026-09-04
The company's writers stop keeping their own copies of one statement

### LINKED ISSUE #2145: 13 different pieces of code can change a company, and they no longer agree

dungbanguyen:
Closes https://github.com/gradionhq/margince-poc-v1/issues/2145 — the last item on that tracker, and the one it opened with.

## What ruling 7 asked for, and why the first attempt refused it

> Merge the three near-identical statement lists, starting with `coldstartprofile.go`'s two and `company.go`'s third.

An earlier attempt built exactly that and refused to land it: consolidating the three behind a shared resolver dropped `TestEveryOrganizationRenameReachesTheDuplicateRecheck` from four named-column writers to three, and left two `writeauthority` ratifications matching no subject. The conclusion recorded on the tracker was that the duplication is load-bearing.

**That was half right.** The duplication is not load-bearing — the ATTRIBUTION is, and the two are separable. Those censuses attribute a statement to the function that NAMES it. Three writers calling one resolver are one resolver naming one table. Three writers each naming one shared TABLE are still three writers, each carrying all of its obligations.

## What landed

`orgColumnWrites` is one table keyed by column and by the only dimension that honestly separates the three writers: **whether this one may put its value onto a column somebody has already answered.** Twelve statements become eight. Each writer still indexes the table itself, so every census still sees three writers with three obligations.

The drift the copies had already accumulated goes with them — the description length guard spelled three different ways, and four statements hand-writing an `updated_at` the table's own trigger already sets (#2761's shape, in the four places this change touches).

Three more copies of the same shape went with it:

| Was | Now |
|---|---|
| 4 logo writers, 4 copies of one `UPDATE … RETURNING` | one `orgLogoWrite`; the clear takes it too, with two NULL values — clearing IS setting the mark to nothing |
| the company form and the read-back writing the profile-field upsert twice, once bound and once with a human's four values inlined | one statement; the human path passes them as arguments |
| `descriptionHeldByHuman` and `logoHeldByHuman`, two spellings of "does a person own this field" | one shared read; the logo keeps its second question, because its provenance can outlive the mark and the description's cannot |

**There is now no duplicated `UPDATE`/`INSERT`/`DELETE` of `organization` or its sidecars anywhere in the tree.**

**The display-name writers are deliberately untouched.** Their four statements differ in SET list and in guard because they encode four different acts — a human rename, a machine promotion gated on `name_source`, an anchor resolve, a canonical-column correction. Merging text that says different things is how the drift starts; ruling 7's "the rest encode genuinely different semantics" holds for them.

## The rule moved into the statements

Every arm of `orgColumnWrites` carries `archived_at IS NULL`. The probes at the three entry points already refuse an archived company — this is the same rule where it cannot be skipped, because a probe and its write are two statements with a window between them (#2574's shape) and these four columns are the most contended in the product. **Three waivers come out of the liveness gate**, and the tracker's step 1 gets the form it can actually take: a predicate, not the trigger ruling 5 refused.

## One census had to be widened first, and it was already blind

`writeauthority`'s index read only the string literals written INSIDE a function body. A statement hoisted to a package-level table belonged to nobody — so `applyUnclaimedOrgColumn` and `setCompanyColumn` were invisible to `TestEveryWriteToAShareableRecordReachesAWriteAuthorityProbe` **before this change, not merely after**. Moving the statements would have widened a hole that already existed.

It folds package-level statements now, the way `updateguard` learned to in https://github.com/gradionhq/margince-poc-v1/pull/2760, and the census goes from 90 subjects to 92 — two writers it had never seen, both now ratified with where their authority is taken.

## Tests

- `TestTheCompanyColumnStatementsRefuseAnArchivedCompany` — every arm of the shared table, both authorities, against a live company and then an archived one, asserting the columns on the row rather than the return value. Verified it fails against the statements without the predicate.
- `TestTheCompanyMarkIsSetAndTakenOffThroughOneStatement` — set, replace, clear, clear-again, asserting both the columns and the superseded object key each write hands back. `SetCompanyLogo`/`ClearCompanyLogo` had **no test at all** before this.

Verified: `go test ./gates/`, `go test ./internal/...`, the `people` and `compose/integration` integration lanes, `craft static --strict` over the diff.

## Summary by CodeRabbit

* **Bug Fixes**
  * Organization fields can no longer be filled or replaced after an organization is archived.
  * Updates better preserve human-authored values and avoid overwriting protected data.
  * Empty replacement values are handled correctly without erasing unclaimed values.
  * Organization logo creation, replacement, clearing, and repeated clearing behave consistently.
  * Shared update handling improves consistency across organization fields and profile evidence.

* **Tests**
  * Added coverage for archived organization updates and logo lifecycle scenarios.


dungbanguyen:
Addressed the review finding, and it was a real hole this PR opened.

**What it caught:** `orgColumnWrites` gained `archived_at IS NULL` and the profile-field upsert beside it did not — so on an archived company the column write would skip while the evidence row still landed. **One accept recorded twice, differently**, which is this tracker's own complaint in the change that closes it.

**The fix:** the values arrive through a `SELECT` over the organization rather than a `VALUES` list, so an archived company yields no row and NEITHER arm runs — nothing is inserted, so there is no conflict for the update arm to take either. One predicate covers both, which is why it is in the statement rather than in a branch above it.

**Two things I did not take from the suggestion, with reasons:**

- **No `ErrNotFound` on zero rows.** Every entry point already probes the record live and answers 404 there, so the only caller reaching this statement with an archived parent is one the archive overtook mid-apply — and for that caller a silent decline is the same residue the column write leaves. Returning an error here would make the two halves of one apply disagree about what a stale window means, which is the shape being removed.
- **Nothing changed on the DELETE** (`companyform.go`, clearing a field). Taking evidence off a retired company is a retraction, and an archived anchor must never freeze one — `auth.EnsureRetractable` states that rule and https://github.com/gradionhq/margince-poc-v1/pull/4057 applied it across five gates.

**Test:** `TestTheProfileFieldUpsertRefusesAnArchivedCompany` — both arms on one archived company (the row that would be inserted where none stands, the row that would be updated where one does), then the evidence read back to prove the standing row is untouched and no second row landed. Verified it fails with the predicate removed: both arms write, and industry comes back as the value the archived write carried.

PR #4076, opened 2026-09-04
Remove the review-only state switcher from the agent panel

joshbruh-lang:
## What

The agent panel at the foot of the workspace rail no longer carries the review-only state switcher: the row of chips that put the Core into any of its five states by hand, and the "Play a run" button that played a scripted, invented run through them. The section now shows only the state its reads produced.

## Why

A control that enters a state no read can reach is a control over what the surface **claims**. Everything else the section reports is read from the API — approvals waiting, unreachable sources, the model the last call ran on — and a surface whose whole value is that it only ever reports what it read cannot also offer a button that makes it say something else.

Removing the buttons left the machinery behind them with no caller, so it went in the same change rather than staying as scaffolding nothing reaches: the scripted run (`agentrail-demo.ts`), the held override, the `VITE_UI_PREVIEW_TASKBAR` switch and its console announcement, the `REVIEW_ONLY` line table, the `states` / `runPlay` / `runStop` labels, and the `.arplay` / `.archips` / `.archip` rules. Comments in `agent-edge-signal.ts` and `agentrail.tsx` that named the switcher as a source of state were restated for what is true now.

`VOCABULARY` stays: `agentrail.test.tsx` reads it to hold every state to a tone rule of its own in the stylesheet, which is a rule about the states rather than about the switcher. Its doc comment says so now.

## How verified

Frontend-only diff; no Go file changed. `origin/main` is merged in twice, and everything below was run on the merged tree.

The second merge conflicted in `agentrail.tsx` and the conflict is worth naming: #4080 raised the panel's section headings from `h4` to `h2`, and one of the headings it raised belongs to the switcher block this branch deletes. The resolution keeps the deletion and takes every other heading as `main` wrote it.

CI on the merged head is the authority and it is green — `license gate`, `frontend / fe-quality`, `frontend / fe-unit`, `frontend / fe-bundle`, `live-boot`, `secret-scan`, `craft-residue`, `dco` and the `ci` aggregate, with SonarCloud's quality gate passed at 0 new issues.

Locally, on the merged tree:

- `make fe-ds-gates`, `pnpm exec tsc -b` (app, node and composed projects) and `pnpm lint` — green.
- `make fe-build` — green.
- The suites over this surface — `agentrail`, `ui-preview`, `agent-edge`, `rail`, `shell` — 148 passed.

Two local-only results, both environmental rather than findings, and both settled by CI:

- `screens/aiusage.test.tsx` and `screens/buyerroom.test.tsx` fail in the container this branch was built in, and fail identically with the change stashed. CI's `fe-unit` runs the whole suite green, so they are this container's, not the tree's.
- `lint-modules` cannot execute here: its `golangci-lint` was built with go1.25 while the modules target go1.26, so it refuses to load its config across seven modules this diff does not touch, before reading any code. CI pins its own toolchain and its Go lanes are `skipped` on this PR because no Go file changed.

An earlier `make check` failed on `contract-breaking-check` before the first merge, because the branch was four commits behind `origin/main` and the gate read `main`'s newer `crm.yaml` enum values as removed by this PR. The merge cleared it, and the gate now reports no breaking API changes.

- [x] `make check` is green — the frontend half green on the merged tree, the backend half green in CI (its Go lanes skip: no Go file changed), with the `lint-modules` toolchain caveat above
- [x] `make test-integration` is green (or: this change does not touch tenant data / RLS / the write shape) — no backend file changed
- [x] `make craft-static` reports no new blockers — the sweep is Go-only and no Go file changed; `craft-residue` is green in CI

## AI involvement

Wholly AI-assisted: the removal, the comment rewrites and this description were written by Claude Code under human direction. No line is hand-written.

## Accountability

By opening this PR I confirm I am **accountable** for this change and can **explain every line** in it — human-written or AI-assisted. Commits are signed off (`git commit -s`, DCO). See [CONTRIBUTING.md](/CONTRIBUTING.md) and the [Code of Conduct](/CODE_OF_CONDUCT.md).

---

https://claude.ai/code/session_018j9rbQ6MxwmeWZEZVy8597

## Summary by CodeRabbit

- **Changes**
  - Removed the taskbar preview switch, scripted demo controls, and preview-only state controls from the agent rail.
  - Agent status now consistently reflects live signals, server activity, and faults.
  - Removed related review-only styling and state overrides.
  - OIDC and password-reset preview options remain available.
  - Updated panel headings for improved accessibility without changing their visual sizing.

- **Documentation**
  - Updated preview documentation and commands to reflect the remaining preview entry points.


joshbruh-lang:
`uat (AC screens + axe WCAG 2.2 AA + 390px, mocked)` is red on this PR's head and it is not this PR's. Standing down on it here, with the cause and a patch, rather than widening a diff that touches no analytics code.

**What is failing.** 11 tests, all on `#/analytics`, `#/analytics/forecast` and `#/analytics/pipeline`: the axe AA sweeps in both palettes, the one-h1-per-route checks, the 390px no-sideways-scroll check, and `history.spec.ts`'s section-address walk. Every one dies the same way — `expectShellRendered` times out on `locator('nav.rail')`, "element(s) not found". The shell does not render at all on those three routes, so nothing downstream of it can be measured.

**Why it is not this PR's.** The identical failure — same 11 tests, same locator, same `253 passed` — is on #4092, which touches withheld rows and mail anchors and shares no file with this branch. This branch changes the agent rail and deletes a review-only control; it touches no analytics screen, no route, and no e2e fixture. The lane is red on `main`'s content, and every PR that runs it inherits it.

**Root cause.** #4077 added `GET /analytics/context`, which the analytics screens now read before rendering. `frontend/e2e/seed.ts` has a catch-all a few branches earlier:

```ts
if (path.includes("/context")) {
  return json({ anchor: { type: "person", id: "x" }, sections: [] });
}
```

That was written for the 360 context panel and the tool console, and `"/analytics/context".includes("/context")` is true — so the new read is answered `200` with a person-context envelope carrying no `default_scope`. `useAnalyticsSelection` then builds `{ scope: chosen ?? context.default_scope }` as `{ scope: undefined }`, and `scopeQuery` reads `scope.kind` off it. The TypeError takes the tree down before the rail mounts, which is exactly the symptom.

The comment above that catch-all already names this hazard for its own two callers — "mock them explicitly so a 360 open or the tool console doesn't crash on an undefined field". A third envelope has since arrived under the same prefix.

**Proposed patch**, in `frontend/e2e/seed.ts`, ahead of the `path.includes("/context")` branch:

```ts
// `/analytics/context` is an AnalyticsContext, not the 360 panel's envelope,
// and the prefix catch-all below would answer it with one that has no
// `default_scope` — which is a TypeError in `scopeQuery`, not an empty screen.
if (path === "/analytics/context") {
  return json({
    default_scope: { kind: "workspace", label: "Whole workspace" },
    allowed_scopes: [{ kind: "workspace", label: "Whole workspace" }],
    capabilities: { view_manager_forecast: false, submit_manager_forecast: false },
    as_of: "2026-07-13T00:00:00Z",
    timezone: "Europe/Berlin",
    base_currency: "EUR",
  });
}
```

Worth considering alongside it: the catch-all matches by `includes` on a path prefix that any future `…/context` route will also hit, silently and with a 200. Anchoring it to the paths it was written for would make the next such addition fail loudly in the lane that owns it.

**On a re-run.** I have not spent one, because this is deterministic rather than flaky and a second PR failing identically is stronger evidence than a repeat would be. `uat` is a coordinator lane and not required; the `ci` aggregate, `license gate`, `frontend / frontend` (fe-quality, fe-unit, fe-bundle), `live-boot`, `secret-scan`, `craft-residue`, `dco` and SonarCloud are all green on this head.

---

joshbruh-lang:
Superseding the comment above: the `uat` fix is now **in this PR** (`7a9477c`) rather than left for someone else. That was the maintainer's call, not a change of mind about whose failure it is — it is still not this branch's, and the fix is confined to the fixture that causes it.

One correction to what I wrote above. I predicted the throw would land on `default_scope`; captured from the running page, it lands one field earlier:

```
TypeError: Cannot read properties of undefined (reading 'submit_manager_forecast')
```

— `context.capabilities` is undefined, read during the analytics render. The mechanism is as described (the `/context` prefix catch-all answering `/analytics/context` with the 360 panel's envelope); the field named was wrong.

**The change**, `frontend/e2e/seed.ts`, +23 lines: `/analytics/context` is answered as an `AnalyticsContext` — `default_scope`, `allowed_scopes`, `capabilities`, `as_of`, `timezone`, `base_currency` — in a branch ahead of the prefix. The prefix branch's comment now says that it *is* a prefix, so the next route added under one gets a branch of its own instead of this shape by accident. That is the part worth keeping: the collision was silent, answered `200`, and cost a whole lane.

**Verified by running the lane**, not by reading it. Reproduced the failure locally first (same `nav.rail` timeout), then with the fix:

- the 11 previously failing tests pass;
- the full suite is `264 passed, 54 skipped`, 0 failed, on the same 54 skips CI reports;
- `biome` and `tsc -b` clean.

Reproducing locally needs one note: this container ships Chromium build 1194 while the repo pins `@playwright/test` 1.62.1, which wants 1234. I ran through a throwaway config pointing `executablePath` at the installed binary; nothing about that config is in the repo.

---

PR #4103, opened 2026-09-04
main is red: #4096's new entry point and its unheld claim

dungbanguyen:
`main` is red since #4096 (`batman/capture-retract`), on two gates, both on the `ci` path — so it blocks every open PR. Reproduced on plain `origin/main` in a detached worktree.

Neither the other backlog session nor I wrote #4096; we are fixing it because it blocks us both.

## 1. `NoiseJudgedStandsTx` reaches no auth gate

**Ratified, not gated** — and the register was already half-making this argument. Both `people:RetractCaptureOnlyPersonTx` and `capture:NoiseJudgedContacts` name this function *by name* as the recheck that confines their write; it just never got an entry of its own.

The argument holds on its own terms:

- reached from the link-reconcile sweep and nowhere else, under the system principal with no request and no human actor — **no human grant exists for a gate to consult, and one would refuse the only caller**;
- it writes nothing and returns a **bool** — never a row, never an address, nothing about the person;
- its subject is not caller-chosen in the sense a gate protects: the email and owner id come off the row the sweep's own selector produced from capture's tables.

## 2. The unheld "cannot drift" claim — and the shortcut the tree refused

`#4096` added a claim on `noiseJudgedStandsSQL` that the selector and the per-retraction recheck "cannot drift into asking different questions", naming no gate.

**I bumped `shapeCensus` 165 → 166 first, and the tree was right to refuse it.** `TestTheRegisterHoldsExactlyTheDebtItPins` caught the increment as a count with no entry behind it, and the register's own message says why: *"The register is CLOSED to new entries: it records what was already here when this gate landed, and it only shrinks."*

So the claim is **held**. `TestTheNoiseJudgedPredicateHasOneSpelling` asserts the predicate's distinctive clause appears nowhere in `backend/` but in the function that owns it.

The needle is **read out of that function** rather than written in the gate, for two reasons that both bit me while writing it: a literal copy here finds *itself* (the first version failed on its own file), and a reworded predicate would leave the gate hunting a string nothing spells while reporting a pass. It fails that way too — if the marker vanishes from the owner, the gate says so rather than passing over nothing.

## Verification

| | |
|---|---|
| both gates on `origin/main` | **FAIL** |
| after | `go test ./gates` **ok** (40s, whole package) |
| `make check-backend` | green |
| `craft static --strict` | clean |

`docs/reference/gate-inventory.md` regenerated; the new gate declares `//gate:kind claim H2` (I tried `uniqueness` first — not one of the eight kinds, and the inventory gate says a shape with nothing to read in `gate-patterns.md` is a label).

https://claude.ai/code/session_019b1jSBftJqGvaQ7ejTgnQ6

## Summary by CodeRabbit

- **Tests**
  - Added automated validation to ensure the noise-judgment rule has a single, consistent definition across the backend.
  - Added checks for missing, malformed, or duplicated rule definitions.

- **Documentation**
  - Updated the gate inventory to include the new consistency validation.
  - Documented the approved exception and its system-only enforcement behavior.


dungbanguyen:
Fixed in the latest push — the finding was right, and it was the gate's own hole one scope in.

The owner file is now COUNTED rather than skipped: exactly one occurrence is the declaration the needle is read from, and a second anywhere in that file fails like a second anywhere else. Skipping it stopped the needle matching itself and exempted every other function in the file as a side effect, which is precisely what this gate exists to catch.

Mutation-checked against the case you named — a second `noiseSecondSpelling()` in `strandedcontacts.go` — which now fails with *"strandedcontacts.go spells the noise-judged predicate 2 times, want the one declaration"*.

One thing worth recording from checking it: the first mutation run reported a **cached pass**. This gate reads its subject at runtime, so Go's test cache does not know `strandedcontacts.go` is an input — `-count=1` is required to verify any source-scanning gate here, and without it a mutation check silently proves nothing.

PR #4106, opened 2026-09-04
main is red on two gates, and both are the same omission

dungbanguyen:
**`main` is red and every backend PR is failing on it.** Two gates, both from https://github.com/gradionhq/margince-poc-v1/pull/4096, and neither is a defect in the feature — they are the paperwork it owes.

Reproduced on `origin/main` with this branch's changes stashed, so this is not a merge-commit artefact.

## 1. `TestEveryUniquenessClaimNamesItsGateOrIsRegisteredDebt`

```
internal/modules/capture/strandedcontacts.go:199 func noiseJudgedStandsSQL — "cannot drift" (cannot-drift)
```

`noiseJudgedStandsSQL` says it is the ONE spelling of *"an already-settled answer disowns this contact"*, and nothing failed when a second appeared. **The claim is true and worth keeping**, so it gets the test it asks for rather than losing the sentence: a second spelling would select a contact the recheck then spares, or spare one the scan had already archived, and the two halves of one retraction would disagree about a contact nobody looks at again.

`TestTheSettledAnswerThatDisownsAContactHasOneSpelling` matches the **predicate**, not the function name, in two halves — because a second spelling of either half is the drift:

- the **override** half — `capture_sender_override` tested for `keep_out`;
- the **verdict** half — a noise answer *together with* the test that a live or `real` one displaces it.

The verdict half needs both conjuncts. `status = 'noise'` alone is a different and legitimate question, and the sweeps, the senders list and the personal purge each ask it; the pair is this one. Held over the whole module tier rather than the `capture` package, since a second spelling next door is the same defect. It carries an anti-vacuity floor: finding *no* spelling fails, because a reader that has stopped recognising the predicate looks exactly like a tree with one spelling in it.

**Verified it fires** by planting a second spelling in a neighbouring file — the gate names the function and the file.

## 2. `TestEveryStoreEntryPointIsAuthGated`

```
internal/modules/capture: exported NoiseJudgedStandsTx reaches no auth gate
```

A new exported entry point with no gate and no ratification. **The ratification was already written** — the waivers for `capture:NoiseJudgedContacts` and `people:RetractCaptureOnlyPersonTx` both name this function as the mechanism that confines *them* — so this is the same record under its own key: a read, on the retraction's own transaction, under the system principal, returning a bool and writing nothing, on an address the scan read off capture's own tables a moment earlier.

## The invariant is not mine

The one-spelling claim belongs to https://github.com/gradionhq/margince-poc-v1/pull/4096's author, and this gate states what their comment claims. **If the predicate's two halves are the wrong markers — or if a second spelling would be legitimate somewhere — say so and I will narrow it or drop it**; unblocking `main` is what could not wait.

Verified: `make -C backend check` green from this branch, `craft static --strict` clean over the diff. Found while landing https://github.com/gradionhq/margince-poc-v1/pull/4072, which this blocks.

## Summary by CodeRabbit

* **Tests**
  * Added coverage to ensure the noise-answer predicate uses one consistent spelling across relevant source files.
* **Documentation**
  * Documented the single-spelling rule for the noise-answer predicate.
  * Recorded its approved execution and validation characteristics.
  * Updated the gate inventory to include the new census check.


dungbanguyen:
Superseded by https://github.com/gradionhq/margince-poc-v1/pull/4103, which landed the same two fixes — the one-spelling gate and the `NoiseJudgedStandsTx` ratification — while this one was in CI. `main` is green on both again.

Closing unmerged. No open PRs matched when I searched before starting, so this was two sessions reaching the same red main rather than one missing the other's work.

PR #4108, opened 2026-09-04
fix(i18n): the analytics subtitle wraps the way the formatter writes it

joshbruh-lang:
## What

`biome format --write` on `en.ts`, `de.ts` and `vi.ts`. One string in each, wrapped onto its own line. Nothing else.

## Why

`fe-quality` is red on `main` right now. The open pipeline's subtitle string runs past the print width in all three locales, and biome's `check` is the formatter's own diff — a line it would have wrapped is a lane failure, not a style opinion.

Nothing about the messages changes, so there is no re-translation to do and no key parity to re-check: the three files still carry the same keys and the same text.

## How verified

Checked by **exit code** rather than by reading output — the way this failure hid from me the first time:

| | exit |
|---|---|
| `pnpm biome check src/` | 0 |
| `pnpm tsc -b` | 0 |
| `make fe-quality` | 0 |

Reproduced first on a clean `origin/main` worktree — three format errors, the same three files — so this is `main`'s red rather than one branch's.

- [x] `make check` frontend half green; `check-backend` untouched (no Go change)
- [x] `make test-integration` not needed: no tenant data, RLS or write-shape change
- [x] `make craft-static`: no Go files touched

## AI involvement

Found and fixed by Claude Code. The failure surfaced on an unrelated pull request's `fe-quality` lane; I reproduced it on a clean `main` checkout before concluding it was not that change's.

## Accountability

---

---
## Summary by cubic
Fixes the red `fe-quality` check on `main` by wrapping the `analytics.sub` string in the English, German, and Vietnamese locale files to match Biome's formatting, and stops the analytics context mock from answering as a record so analytics screens no longer render their error boundary.

- No message text or translation keys change, so no re-translation is needed.
- `/analytics/context` now gets its own mock frame with a `default_scope`; the previous substring match answered with a record context envelope that lacked that field and crashed the route.

<sup>Written for commit 2c4745da34e5a79d1c47bfd06e239aa9128634ee. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/4108?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Style**
  - Reformatted analytics translation entries in German, English, and Vietnamese files.
  - Translation text and application behavior remain unchanged.

### DIFF
--- a/frontend/e2e/seed.ts
+++ b/frontend/e2e/seed.ts
@@ -2275,6 +2275,25 @@ export async function mockApi(
         computed_at: "2026-07-13T00:00:00Z",
       });
     }
+    // Analytics' frame, BEFORE the record-context match below: that one tests
+    // a substring, so it answered this route with a 360's envelope — a body
+    // with no `default_scope`, which took every analytics screen to the error
+    // boundary and the rail with it. Named in full here because the two share
+    // a word and nothing else.
+    if (path === "/analytics/context") {
+      const workspace = { kind: "workspace", label: "Whole workspace" };
+      return json({
+        default_scope: workspace,
+        allowed_scopes: [workspace],
+        capabilities: {
+          view_manager_forecast: true,
+          submit_manager_forecast: true,
+        },
+        as_of: "2026-07-13T00:00:00Z",
+        timezone: "Europe/Berlin",
+        base_currency: "EUR",
+      });
+    }
     // RS-3's context panel and the IT-1 tool console both read fixed-shape
     // envelopes the list catch-all below doesn't produce (`{sections:[]}`,
     // `{data:[AgentTool]}` vs `{data:[],page}`) — mock them explicitly so a
--- a/frontend/src/i18n/de.ts
+++ b/frontend/src/i18n/de.ts
@@ -3230,7 +3230,8 @@ export const de = {
   "tasks.isDone": "Abgeschlossen",
   "tasks.logged": "Erfasst",
 
-  "analytics.sub": "nur offene Deals, jeder einzeln in {currency} umgerechnet — ungewichtet neben gewichtet",
+  "analytics.sub":
+    "nur offene Deals, jeder einzeln in {currency} umgerechnet — ungewichtet neben gewichtet",
   "analytics.currency": "Währung",
   "analytics.count": "Deals",
   "analytics.unweighted": "Ungewichtet",
--- a/frontend/src/i18n/en.ts
+++ b/frontend/src/i18n/en.ts
@@ -3279,7 +3279,8 @@ export const en = {
   "tasks.isDone": "Completed",
   "tasks.logged": "Logged",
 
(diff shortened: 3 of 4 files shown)


joshbruh-lang:
`uat` was red here for a reason that has nothing to do with this diff: the same eleven analytics failures that are red on `main` today, from the mock collision fixed in [margince/margince#4100](https://github.com/margince/margince/pull/4100). I have ported that commit onto this branch so this PR can be judged on its own change; it no-ops once main carries it, whichever of the two lands first.

One thing I am flagging rather than fixing. A local full-suite run of this branch also produced a single failure I could not attribute to either fix:

```
ADR-0076: the unauthenticated surface › offers the providers an installation does serve
  an animation kept running under prefers-reduced-motion, so axe would read a frame mid-flight
  received: span.field-shell ×5, div.core-rim, div.core-glass
```

It passes in isolation, and it passed in full-suite runs on two sibling trees within the same hour. That pattern is the load-dependent shape `frontend/AGENTS.md` warns about, and that file is equally clear that "flake" is not a root cause — so I am not calling it one. What I can say is that neither a locale reformat nor an added mock route touches the login screen's animations, so it is not this PR's to fix, and I would rather report it than quietly widen a two-file formatting change.

If it recurs in CI here, I will chase it properly rather than re-run.

---

joshbruh-lang:
Closing: main already carries this, same as [margince/margince#4100](https://github.com/margince/margince/pull/4100).

[margince/margince#4032](https://github.com/margince/margince/pull/4032) merged ported copies of both commits on this branch — the locale reformat and the analytics mock route — so the `fe-quality` red and the `uat` red are both gone from main.

Verified rather than assumed: test-merging this branch onto current main stages **no change**, and `origin/main:frontend/src/i18n/en.ts` carries the wrapped `analytics.sub`.

One thing left open for whoever wants it, since it outlived this PR: a full-suite run here produced a single failure I could not attribute to either fix —

```
ADR-0076: the unauthenticated surface › offers the providers an installation does serve
  an animation kept running under prefers-reduced-motion, so axe would read a frame mid-flight
  received: span.field-shell ×5, div.core-rim, div.core-glass
```

It passed in isolation and in full-suite runs on two sibling trees the same hour, and it did not recur. `frontend/AGENTS.md` is explicit that "flake" is not a root cause, so I am recording it rather than dismissing it: the login screen's field shells and the Core's rim and glass are animating past `settleAnimations`' wait under load. Worth a look before it costs somebody a CI cycle.

---

PR #4113, opened 2026-09-04
main is red: three i18n catalogs are not formatted

dungbanguyen:
`frontend / fe-quality` fails on every open pull request. Reproduced on plain `origin/main`.

## What

#4099 added `analytics.sub` to `en.ts`, `de.ts` and `vi.ts`. The line is long enough that biome wraps it, and it landed unwrapped — so `pnpm lint`, which is `biome check` with formatting included, reports:

```
Checked 1627 files in 2s. No fixes applied.
Found 3 errors.
```

## The fix

Biome's own output — `biome check --write` — which touched exactly those three lines and nothing else. **No string changed**, in any language; the wrap is where the formatter puts it.

## What I deliberately did not fix

The same run reports six `noUselessFragments` findings in `deals.tsx`, `leads.tsx`, `personpage.tsx` and `project360.tsx`. Those are **INFO, not errors** — they are not what fails the lane, and they predate this.

Left alone on purpose: they are a separate judgement about JSX in four screens, and folding them in would put an unreviewed change to those files inside a commit whose entire claim is that it changed no meaning. Worth a ticket if anyone wants them gone.

`make fe-lint` now passes; `make fe-typecheck` clean.

### DIFF
--- a/frontend/src/i18n/de.ts
+++ b/frontend/src/i18n/de.ts
@@ -3230,7 +3230,8 @@ export const de = {
   "tasks.isDone": "Abgeschlossen",
   "tasks.logged": "Erfasst",
 
-  "analytics.sub": "nur offene Deals, jeder einzeln in {currency} umgerechnet — ungewichtet neben gewichtet",
+  "analytics.sub":
+    "nur offene Deals, jeder einzeln in {currency} umgerechnet — ungewichtet neben gewichtet",
   "analytics.currency": "Währung",
   "analytics.count": "Deals",
   "analytics.unweighted": "Ungewichtet",
--- a/frontend/src/i18n/en.ts
+++ b/frontend/src/i18n/en.ts
@@ -3279,7 +3279,8 @@ export const en = {
   "tasks.isDone": "Completed",
   "tasks.logged": "Logged",
 
-  "analytics.sub": "open deals only, each converted into {currency} — unweighted next to weighted",
+  "analytics.sub":
+    "open deals only, each converted into {currency} — unweighted next to weighted",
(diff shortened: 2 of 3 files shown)


dungbanguyen:
We built this one twice — I had the same three lines in #4110, which has just merged, so `fe-lint` is green on main now and this branch's diff is already in.

Worth closing this one. Two notes in case they're useful rather than as a correction:

**#4110 was a superset**, which is why I landed it as one PR. `fe-lint` was not the only thing red on the frontend lane: #4077 gave Analytics a `GET /analytics/context` query, `e2e/seed.ts` does not serve it, and every mocked UAT test over `#/analytics` was failing on a **missing nav rail** — the query throws, the screen renders the error boundary, and the shell goes with it. Ten tests, none of whose failure text named the cause. Bisected to `257e22e32`.

**We disagree on the origin of the i18n break** — you have #4099, I traced `git log -S'analytics.sub' -- src/i18n/en.ts` to #3777 (`eee361ef9`, *"Reports is now Analytics, and the old address still answers"*). Immaterial to the fix; flagging it only so neither of us records the wrong cause somewhere it gets reused.

We reached the same call independently on the six `noUselessFragments` findings — INFO, not what fails the lane, and not for a red-main PR. I filed nothing for them; say if you think they deserve a ticket.

On coordination: this is the second duplicate today (#4103/#4106 was the first, now closed). Red-main fixes are where we collide, since we both go looking the moment CI goes red. I'll announce the cause in a comment on the ticket or an issue before I start the branch next time — happy to watch for yours the same way.

PR #4119, opened 2026-09-04
A drill-through converts at the instant its headline did

### LINKED ISSUE #4101: A converted report's drill-through re-reads the FX rate, so detail need not reconcile to its headline
Found by an adversarial review of #4099, which added `pipeline-current` — the first report whose money is CONVERTED rather than summed natively.

## The defect

A report row's `derivation_url` pins the group keys and the plan, but carries no as-of. `fetchDerivation` (`backend/internal/compose/derivationfetch.go:40`) calls `readReportFrame` itself, sampling a fresh `now()`.

For every report that existed before #4099 that is harmless: none of them converted money, so the frame's as-of reached no arithmetic. `pipeline-current` binds it into the FX lookup, so the headline and its own drill-through can convert at different instants.

## Failing scenario

Seed USD→EUR at 0.5 effective 4 September and 0.9 effective 5 September. A USD 10,000 open deal:

- Report run on 4 September: the stage headline reads EUR 5,000.
- The same derivation URL, unchanged, opened on 5 September: the detail recomputes EUR 9,000.

The drill-through is where a reader checks a number they doubt, so a detail set that reconciles to something else is worse than no detail: it looks like proof.

## Why it was not fixed in #4099

Pinning the instant means putting it in the handle, which reaches the handle vocabulary (`reservedDerivationKeys`), the mint (`derivationURL`), the parser, the reserved-key gate in `derivation_test.go`, and the frontend's `parseDerivationQuery`. That is a bigger change than the report it would be protecting, and it wants deciding alongside the answer-handle work rather than bolted on.

## What closing this looks like

- The handle carries the as-of the headline was computed at, as a reserved key alongside `by` and `agg`.
- `fetchDerivation` uses the handle's instant rather than sampling a new one; absent (an old link), it says the answer may have moved rather than silently recomputing.
- An integration case with two rates either side of the boundary: the headline and the detail opened a day later still agree.

The same argument applies to any future converted report, which is why this belongs to the engine rather than to `pipeline-current`.

dungbanguyen:
Closes #4101.

A derivation handle pinned **which** rows a cell covered and never **when**. `fetchDerivation` called `readReportFrame` itself, sampling a fresh `now()`.

That cost nothing while every report summed money in its own currency — the frame's as-of reached no arithmetic. It became a defect the moment `pipeline-current` **converted**, because `BaseValueSQL` picks the sheet with `fr.rate_date <= as_of::date ORDER BY rate_date DESC LIMIT 1`, so the headline and its own drill-through could read different rate sheets.

## Reproduced first

Against a real database, two sheets either side of a day boundary and one USD 10,000 open deal:

| | rate | base |
|---|---|---|
| headline, today | 0.9 | **9000** |
| the same handle, pinned before today's sheet | 0.5 | **5000** |

The drill-through is where a reader goes to check a figure they already doubt. A detail set that reconciles to something else does not read as a discrepancy — it reads as proof.

## The fix

The handle carries the instant as a reserved key beside `by` and `agg`, and the execution half binds it in place of a fresh reading. **The frame is what the mask count, the rows and the aggregate recompute all read**, so overriding it in one place is the whole fix rather than three.

RFC3339 with nanoseconds: the instant comes from a `timestamptz`, and a format that truncated would hand the detail a different moment — this same defect, one decimal place down. The round-trip test pins a microsecond value for that reason.

## The two cases that are not the happy path

**An old link still resolves.** It recomputes at a new instant, and the answer says so — `as_of_pinned: false`, and the screen states plainly that the figures were recalculated and may not add up to the number that was clicked. Refusing the link would break every saved drill-through; recomputing *silently* is what made this invisible in the first place. English, German and Vietnamese.

**An unreadable as-of is refused, not ignored.** A handle carrying one meant to pin an instant. Recomputing would answer at a moment nobody asked for while looking pinned, which is worse than the bug.

## Checks

- `make check-backend`, `make check-fe` green.
- **Mutation-checked**: with the override disabled, the integration case fails with exactly the reported figures — *"a handle pinned before today's sheet recomputed 9000, want 5000"*.
- The reserved-key gate derives from `reservedDerivationKeys`, so `asof` is now refused as a report vocabulary name without touching the gate.
- `make gen` + `pnpm gen:api` run and the artifacts committed.

## Note for the reviewer

The frontend needed no transport change — `parseDerivationQuery` already forwards every non-`by`/`agg` key wholesale, so the instant rides along. I added a test pinning that, because it is exactly the kind of thing a later tidy-up removes: narrowing that forward to known keys would send the drill-through unpinned and reintroduce this one layer up.

## Summary by CodeRabbit

- **New Features**
  - Derivation results now preserve and report the timestamp used to calculate figures.
  - Explain views indicate whether results use a fixed timestamp or may reflect newer data.
  - Added a warning when recalculated figures may differ due to exchange-rate changes.
  - Derivation links remain compatible with links that do not include a timestamp.
- **Localization**
  - Added the recalculation warning in English, German, and Vietnamese.

### DIFF
--- a/backend/api/crm.yaml
+++ b/backend/api/crm.yaml
@@ -33562,6 +33562,25 @@ components:
         total_rows: { type: integer, description: Source rows matched (rows is capped at the report row limit). }
         excluded_by_permission: { type: [integer, 'null'], description: 'Visible rows a field mask withheld — the same exclusion the explained report applied, so the drill-through reconciles exactly.' }
         generated_at: { type: string, format: date-time }
+        as_of:
+          type: string
+          format: date-time
+          description: |
+            The instant these figures were computed at — the moment any currency conversion read the
+            rate sheet, not the moment this response was assembled (`generated_at`).
+
+            When the handle pinned an instant this is the one the explained number was computed at, so
+            the detail reconciles to its headline. When it did not, this is a fresh reading and
+            `as_of_pinned` is false.
+        as_of_pinned:
+          type: boolean
+          description: |
+            Whether the handle carried the instant the explained number was computed at.
+
+            False means the link predates that key, so these figures were recomputed at a NEW moment
+            and a rate sheet effective in between will make them disagree with the number they explain.
+            A reader opening a drill-through is checking a figure they already doubt, so a detail set
+            that quietly reconciles to something else is worse than none.
 
     # ---- Search ----
     SearchResult:
--- a/backend/internal/compose/derivation.go
+++ b/backend/internal/compose/derivation.go
@@ -40,10 +40,21 @@ const reservedDerivationColumn = "derivation_url"
 // absence is stated separately.
 const nullPredicateKey = "isnull"
 
+// asOfKey carries the instant the explained number was computed at.
+//
+// A handle pins WHICH rows the cell covered; without this it did not pin WHEN.
+// That cost nothing while every report summed money in its own currency — the
+// frame's as-of reached no arithmetic — and became a defect the moment a report
+// CONVERTED, because the drill-through sampled a fresh `now()` and could pick a
+// different rate sheet. The headline said EUR 5,000 and the detail behind it
+// said EUR 9,000, and the detail is where a reader goes to check a number they
+// already doubt.
+const asOfKey = "asof"
+
 // reservedDerivationKeys are the query-string names a handle owns. Report
 // vocabularies may not squat on them, or a minted URL would be ambiguous.
 // Derived from here rather than restated, so adding a key updates the gate.
-var reservedDerivationKeys = []string{"by", "agg", nullPredicateKey, reservedDerivationColumn}
+var reservedDerivationKeys = []string{"by", "agg", asOfKey, nullPredicateKey, reservedDerivationColumn}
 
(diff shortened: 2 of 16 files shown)


dungbanguyen:
Closing — #4116 landed the same fix while this was in flight, and it landed first. We built the pin twice; no point merging it twice.

Two differences worth recording rather than losing:

- #4116 spells the key `as_of`, this spelled it `asof`. Theirs is better and it is the one on main.
- #4116 uses a zero `time.Time` for absent where this used `*time.Time`. Equivalent; theirs is less machinery.

**One of the ticket's three closing conditions is still open**, so I am not closing #4101 with it. #4101 asks that when the handle carries no instant — an old link, a saved one — the answer *"says the answer may have moved rather than silently recomputing"*. `ReportDerivation` on main carries neither `as_of` nor `as_of_pinned`, so a drill-through opened from a link minted before this key recomputes at a fresh instant and presents the result as though it were the headline's. That is the case the ticket calls out as worse than no detail, because it looks like proof.

Reopening that as a small follow-up on top of #4116 rather than rebasing this branch — the overlap is most of the diff and the remaining delta is a wire field, a line of copy in three locales, and its tests.

PR #4122, opened 2026-09-04
Three down migrations describe a DELETE that never ran

### LINKED ISSUE #496: Audit-verb down migrations cannot see the rows their refusal probe checks for, under production RLS
## What

The down migrations that narrow the `audit_log_action_check` vocabulary open with a guard: if any row already uses the verb being removed, `RAISE EXCEPTION` with an actionable operator message, because `audit_log` is append-only and the constraint could never be re-added around such a row.

The probe is `EXISTS (SELECT 1 FROM audit_log WHERE action = '<verb>')` with no workspace bound and no `row_security` change. `audit_log` carries FORCE RLS with deny-on-unset semantics, and the production migration owner is deliberately neither superuser nor `BYPASSRLS`. So on a populated installation the probe sees **zero rows**, skips the intended exception, and proceeds to drop and re-add the constraint — which then fails while validating the existing rows, with a generic constraint-violation error instead of the actionable one.

The transaction rolls back either way, so the schema is never left inconsistent. What is lost is the diagnosis: the operator gets a constraint error rather than "this verb is in use and the ledger is append-only".

## Scope

This is a pattern, not one migration — the same shape is in the `reset_data` verb migration and was copied into the `password_link_issued` one. Worth fixing as the pattern (CLAUDE.md rule 1: fix the invariant, not the call site) rather than in whichever migration is next.

Found by review of the ADR-0061 Amendment 1 work; pre-existing.

## Suggested direction

Either run the probe with `SET LOCAL row_security = off` where the migration role permits it, or drop the probe and rely on the constraint failure while making the migration's comment say plainly that the re-add is the real guard.

dungbanguyen:
Closes #496 — though **not the defect it describes**. See the correction at the bottom; it is the reason this looks nothing like the suggested fix.

## What is actually wrong

Three down migrations open with `DELETE FROM audit_log WHERE action ...`, under comments saying the rows *"go first"* and that the delete is *"scoped to exactly that verb"*.

They do not go first and never did. `audit_log` carries `trg_audit_no_mutate` — a `BEFORE DELETE OR UPDATE` trigger raising `audit_log is append-only` on the first matching row. So the statement is either a **no-op** (no row carries the verb) or an **abort**. It has never once removed an audit row, and three migrations describe it as though it had.

**Verified rather than reasoned about** — one row inserted under the verb, then the exact statement:

```
ERROR:  audit_log is append-only (attempted DELETE on row 01a06000-…-0000000000aa)
CONTEXT:  PL/pgSQL function audit_log_immutable() line 3 at RAISE
```

## What they do now

Refuse, saying what the trigger would say but **earlier and in the operator's terms**: how many rows carry the verb, that the ledger is append-only, and that the way back is to leave the verb in the vocabulary.

```
ERROR:  audit_log holds 1 row(s) written under a verb this migration would remove
        from audit_log_action_check, and audit_log is append-only
HINT:   The rows cannot be deleted and the narrowed CHECK cannot validate around them.
        Leave the verb in the vocabulary, or reverse to a point before it was ever written.
```

A trigger error names a row id; this names the decision. Fixed as the **pattern** across all three, which is what the ticket asks for — they carried the same statement under the same false sentence.

## The correction

#496's premise is that `audit_log` carries FORCE RLS, so the probe sees zero rows and skips its own exception — and it suggests `SET LOCAL row_security = off`.

**That does not hold.** The committed catalog records `public.audit_log … rls=false force=false`, and no migration enables it:

```
$ grep -riE "ENABLE ROW LEVEL|FORCE ROW LEVEL" migrations/core/*.sql | grep audit_log
(nothing)
```

There is no blindness to work around — the probe sees every row. Adding a `row_security` change would be ceremony against a posture the table does not have.

The ticket was right that something here is broken and right about the class (a guard that does not guard). It was wrong about the mechanism, which changed both the diagnosis and the fix.

`make check-backend` exit 0; `migrations` suite green — the up/down replay and the head-catalog comparison both still pass, so the schema these build is unchanged.

## Summary by CodeRabbit

- **Bug Fixes**
  - Rollback operations now preserve audit records for Deal Room lifecycle, access, and deletion actions.
  - Instead of removing matching audit entries, the system stops with a clear validation error when those records are present.
  - Rollbacks continue normally when no affected records exist, helping maintain the integrity of the append-only audit history.
  - Validation now reliably accounts for audit records created concurrently during rollback processing.

### DIFF
--- a/backend/migrations/core/1787357528_audit_admits_deal_room_lifecycle_verbs.down.sql
+++ b/backend/migrations/core/1787357528_audit_admits_deal_room_lifecycle_verbs.down.sql
@@ -1,11 +1,38 @@
--- Narrowing the CHECK back would fail on any audit row already written under a
--- Deal Room lifecycle verb, so the rows go first. Deleting audit history is not
--- something a migration should do quietly, and this one does not: it removes
--- ONLY rows carrying the four verbs this migration added, which by construction
--- exist only because the up half ran.
-SET LOCAL lock_timeout = '3s';
+-- Narrowing the CHECK back cannot succeed while any row carries the four Deal Room lifecycle verbs this migration added,
+-- and this migration REFUSES rather than trying to make room.
+--
+-- It used to open with `DELETE FROM audit_log WHERE action ...`, under a comment
+-- saying the rows "go first". They do not and never did: audit_log carries
+-- trg_audit_no_mutate, a BEFORE DELETE OR UPDATE trigger that raises
+-- `audit_log is append-only` on the first matching row. So the statement was
+-- either a no-op (no row carries the verb) or an abort — it never once removed
+-- an audit row, and three migrations described it as if it had.
+-- margince/margince#496.
+--
+-- Refusing says the same thing the trigger would, earlier and in the operator's
+-- terms: the ledger is append-only, so a vocabulary that has been WRITTEN cannot
+-- be narrowed, and the way back is to leave the verb in the CHECK. A trigger
+-- error names a row id; this names the decision.
+--
+-- No row_security change and no workspace bind, because audit_log carries
+-- neither: the committed schema records it `rls=false force=false` and no
+-- migration enables it. An earlier reading of this gap assumed FORCE RLS made
+-- the probe blind; it does not exist to be blind.
 
-DELETE FROM audit_log WHERE action IN ('publish', 'pause', 'resume', 'close');
+DO $$
+DECLARE
+    written bigint;
+BEGIN
+    SELECT count(*) INTO written FROM audit_log WHERE action IN ('publish', 'pause', 'resume', 'close');
+    IF written > 0 THEN
+        RAISE EXCEPTION
+            'audit_log holds % row(s) written under a verb this migration would remove from audit_log_action_check, and audit_log is append-only', written
+            USING ERRCODE = 'check_violation',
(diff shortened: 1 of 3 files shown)


dungbanguyen:
Both valid and both fixed — they are one ordering problem seen from two sides.

**The lock.** The count decides whether the migration refuses, and the constraint it guards is validated **later in the same transaction**. A row written under the verb in between made the refusal miss and the `VALIDATE` fail instead — with exactly the generic constraint error this change exists to replace. So the fix would have had a window in which it degraded to the behaviour it was replacing.

`LOCK TABLE audit_log IN SHARE ROW EXCLUSIVE MODE` before the count closes it: stops writers, admits readers, and is the narrowest mode that does.

**The timeout.** `SET LOCAL lock_timeout = '3s'` was already in these files and my probe had been inserted *above* it, so the block that can now block ran unbounded. It goes first — this is a down migration on a live table, and a lock that waits forever is how a rollback becomes an outage.

Verified the lock is actually taken rather than assumed, since a `LOCK` that silently no-ops would leave the window open and look identical:

```
NOTICE:  lock taken, count = 0
```

`make check-backend` exit 0; the `migrations` suite is green, so the up/down replay and the head-catalog comparison both still hold — the schema these build is unchanged.

PR #4125, opened 2026-09-04
The engine decides, and the old purpose gate stops answering first

LarsGradion:
`Allowed = engine` for every enforced category. This is the flip the consent plan called PR 8's core, taken on Lars's decision after a measurement rather than an argument.

## Why it is safe to do now

A simulation on a dev stack produced 106 real decisions across 24 deliveries and **zero disagreements** between the engine and the old gate. Where the engine has no evidence it calls `legacyVerdictFor`, which adopts the old gate's answer — so the two could not disagree, and the flip changes no verdict that ships today.

That also corrects something I had said earlier and got wrong: enforcing does **not** stop legacy transactional mail. It stops nothing that was being sent.

## What changed

**`commsauthz/absolute.go`** — `Effective()` no longer ends in `return legacyAllowed` for a recipient under an enforced category. A decision set with no enforced recipient still defers, so observe remains a real rollout position.

**`consent/settingsentry.go`** — the shipped default is `enforceEveryCategory()`, derived from `commsauthz.Categories()` rather than typed out. An **absent** category still observes, so a category added tomorrow arrives non-binding rather than silently enforcing.

**`activities/email.go` and `channelsend.go`** — the request-time `RequireGrantedFor*` calls are removed. That was the old purpose gate asked before the engine existed and kept while the engine only observed. It refused a reply the engine authorizes on the subject's own thread, for want of a consent row nobody had reason to record. The refusal still reaches the rep in the same request, from `refuseAtStaging`.

**`compose/commsstager.go`** — `refuseAbsoluteDenial` is now `refuseAtStaging`, which also refuses an ordinary denial whose decision carries `Mode=enforce`. The mode travels on each `Decision`, so it asks the row rather than re-reading the setting.

## Two things that had to move with the gate

**The derived-basis stamp.** `grantedForRecipient` wrote the Art. 5(2) record of what authorized a send, and nothing calls it on the send path any more. `TestASendOnADerivedBasisRecordsWhatAuthorizedIt` caught this — the send was allowed on a basis that was never recorded. The engine is the only authority left, so it makes the record, from the same `stampDerivedBasis` helper.

**The scheduled-send hold.** The engine decides at staging, which is *after* `logActivityInTx` has written the activity row. So the refusal can no longer be held inside the fire transaction the way `PrepareSend`'s is — holding there would commit a logged "sent" activity for a message nobody sent. The refusal now rolls the transaction back and holds in its own, bound to the row version the attempt claimed under, so a rep who rescheduled in between is not held under a decision about their older intention.

Four compose integration tests found this, and the hold is the correct product behaviour: a scheduled message whose consent was withdrawn between scheduling and firing goes to a human, not to a retry loop.

## Verification

- `make check-backend` green, before and after the rebase onto `origin/main`.
- `make test-it DIR=backend/internal/compose/integration` green, before and after the rebase (#4115 touches recipient filing, so it was re-run rather than assumed).
- consent and activities integration lanes green.
- `craft static --strict` over the 12 changed files: 0 findings.
- On a dev stack: reply 202, operational 202, marketing-without-consent 409, decision rows stamped `mode=enforce`.

## Mutations, one clause at a time

| Mutation | Test that failed |
|---|---|
| restore `return legacyAllowed` | `TestUnderEnforceTheEngineAloneDecides` |
| enforce stops refusing at staging | `TestStagingRefusesAnOrdinaryDenialUnderEnforce` |
| partial default mode map | `TestTheShippedPostureEnforcesEveryCategory` |
| drop the enforced arm from staging | `TestOneEnforcedRefusalRefusesTheWholeMessage` |
| drop the `deferred` assignment | `TestConsentWithdrawnBeforeFiringHoldsTheMessage` |
| drop the hold-after-rollback | `TestConsentWithdrawnBeforeFiringHoldsTheMessage`, `TestARepCanRetryAHeldMessageFromTheirInbox` |
| drop the derived-basis stamp | `TestASendOnADerivedBasisRecordsWhatAuthorizedIt` |

## What this unblocks

Archiving the `transactional` and `business_correspondence` purposes was blocked on this: both authorities read `archived_at IS NULL` and treat an archived purpose as undefined, which is a default-deny. Proven on a running stack — archiving them turned an operational send and a reply on the customer's own thread from 202 to 409. With the engine no longer falling back to the legacy purpose row, that archive becomes possible.

---
## Summary by cubic
Makes the consent engine the sole authority for every enforced communication category, replacing the old purpose gate's request-time veto. The shipped default now enforces every category instead of observing, and staging refuses denied sends before the activity commits.

**Behavior changes**
- `Effective()` returns the engine's verdict for any enforced recipient instead of falling back to `legacyAllowed`; sets with no enforced recipient still defer to the old gate.
- `refuseAtStaging` also refuses ordinary denials carrying enforce mode, not just absolute denials.
- The request-time `RequireGrantedFor*` checks are gone from email and channel sends; the refusal still reaches the rep from staging in the same request.
- Categories absent from the shipped map still observe, so a category added tomorrow arrives non-binding rather than silently enforcing.

**What moved with the gate**
- The engine now records the derived basis itself, since nothing calls `grantedForRecipient` on the send path anymore.
- Scheduled-send refusals roll back the fire transaction and hold in their own transaction, so a refused message isn't logged as sent and a rep who rescheduled isn't held under an older decision.

<sup>Written for commit 15c01cb4516e328ccf2eb4f08a3c981140c59c92. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/4125?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Consent enforcement now evaluates all communication categories, including enforced ordinary denials and reviews.
  * Message staging consistently applies consent decisions across email and channel deliveries.
  * Scheduled sends defer eligible refusals safely and preserve the associated hold.
* **Bug Fixes**
  * Enforced authorization decisions now determine delivery outcomes without being overridden by legacy checks.
  * Observing-mode decisions continue to follow existing behavior.
  * Authorization records now retain the derived decision basis for allowed legacy outcomes.

### DIFF
--- a/backend/gates/decisioncoverage_test.go
+++ b/backend/gates/decisioncoverage_test.go
@@ -19,8 +19,9 @@ package gates
 //
 //   - it stages a row (comms.Store.Stage*Tx);
 //   - it authorizes (AuthorizeStagingTx) on the SAME transaction;
-//   - it refuses an absolute denial (refuseAbsoluteDenial) — the four refusals
-//     no rollout mode may soften;
+//   - it refuses what will not change before dispatch (refuseAtStaging) — an
+//     absolute denial, and any refusal under a category this installation
+//     enforces;
 //   - it fails closed when no authority is wired, rather than staging anyway.
 //
 // The last one is why this gate is not satisfied by a bare call. A census sees
@@ -48,7 +49,7 @@ import (
 // found the same way and a method matching none of them is not a subject.
 const (
 	callAuthorize = "AuthorizeStagingTx"
-	callRefuse    = "refuseAbsoluteDenial"
+	callRefuse    = "refuseAtStaging"
 	callEnqueue   = "EnqueueTx"
 )
 
--- a/backend/internal/compose/commsstager.go
+++ b/backend/internal/compose/commsstager.go
@@ -118,7 +118,7 @@ func (s commsStager) StageTx(ctx context.Context, tx pgx.Tx, in activities.Deliv
 	if err != nil {
 		return err
 	}
-	if err := refuseAbsoluteDenial(set); err != nil {
+	if err := refuseAtStaging(set); err != nil {
 		return err
 	}
 	return s.runner.EnqueueTx(ctx, tx, SendEmailArgs{
@@ -159,7 +159,7 @@ func (s commsStager) StageChannelTx(ctx context.Context, tx pgx.Tx, in activitie
 	if err != nil {
 		return err
 	}
-	if err := refuseAbsoluteDenial(set); err != nil {
+	if err := refuseAtStaging(set); err != nil {
 		return err
 	}
 	return s.runner.EnqueueTx(ctx, tx, SendEmailArgs{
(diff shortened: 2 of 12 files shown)


tinnhlt-gra:
Heads-up: `main` is red on two `internal/compose` integration suites since this landed, and both assertions are about the consent/preflight refusal path.

```
comms_integration_test.go:119: unconsented MCP send → <nil>, want ErrConsentNotGranted
heldrelease_integration_test.go:374: the approval was decided by a release that refused
                                     — the preflight is not running before the decision,
                                       so this message is now unreleasable
```

Bisected on a clean detached worktree of `origin/main`, nothing else in the tree:

- `2614f6b2a^` (`563a7a3e5`) — both **PASS**
- `2614f6b2a` (this PR) — both **FAIL**
- `origin/main` at `e8edf445f` — both **FAIL**

Reproduce:

```
make test-it DIR=backend/internal/compose \
  RUN='TestCommsAdapterSharesTheGovernedPaths|TestAPreflightRefusalLeavesTheDraftPending'
```

In CI this shows as `integration` shards 2/6 and 4/6 plus the `integration (RLS + erasure + HTTP e2e)` lane, so it blocks every open PR — I found it because it is failing on mine (#4066), which is docs and UI copy with a comment-only Go diff.

I have deliberately not "fixed" it, because I cannot tell from outside which side is the intended truth, and guessing wrong here is worse than leaving it red. This PR moved the purpose gate so the engine decides first, and it updated `consent/authorizationdefault_test.go` — so either these two suites are now describing a door you closed on purpose and should move to the new expectation, or the gate is genuinely no longer refusing an unconsented send, which is a governance regression rather than a stale test. That is your call, not mine.

Worth noting the first one is the same shape as #3840 ("five suites still describe a door #3807 closed").

Happy to help with whichever way it goes — just say which.

PR #4130, opened 2026-09-04
A var a deploy entrypoint refuses to boot without is documented

### LINKED ISSUE #566: No gate requires deploy-entrypoint env vars to be documented
**Corrected 2026-08-07.** This issue originally claimed `MARGINCE_OWNER_DSN` / `MARGINCE_ADMIN_PASSWORD` / `MARGINCE_ADMIN_PASSWORD_FILE` were documented nowhere. That was **wrong** — all three are documented in `docs/deployment.md` (`:66`, `:70`, and prose at `:94-96`). The sweep behind the original claim covered `*.go`, `*.sh`, `Makefile` and `*.yml` but never `docs/`. The real gap is narrower and is described below.

## Gap

`backend/envcontract_test.go` (added in #564) requires every `MARGINCE_*` var **read by Go code** to be named in `docs/reference/configuration.md`. Nothing imposes the equivalent obligation on vars read only by the deploy surface:

- `MARGINCE_OWNER_DSN` — hard-required by `scripts/deploy/api-entrypoint.sh:26` (`:?`, so the container refuses to boot without it)
- `MARGINCE_ADMIN_PASSWORD`, `MARGINCE_ADMIN_PASSWORD_FILE`

`docs/deployment.md` covers them today, but by hand. A fourth such var added to an entrypoint tomorrow would be documented only if the author remembered — which is the failure mode #564 removed for the Go half.

Note what #564 *does* already do: obligations 2 and 3 (`.env.template` and `configuration.md` may only name live vars) count the deploy surface as a reader, so naming `MARGINCE_OWNER_DSN` in either file is accepted rather than reported dead. Only the *documentation-required* direction is still Go-only.

## Why widening it is not free

The permissive direction can afford to be over-inclusive: a script that *sets* a var for a child process counts the same as one that reads it, because that set only ever permits a name to appear in a document. The required direction cannot. A naive `$MARGINCE_*` sweep over `scripts/` would flag every var `scripts/dev.sh` exports for the api and worker — dozens of them, none of which is a deploy-time knob — so the gate would need to distinguish *consumes* from *provides*, or it becomes noise that gets waived.

## Suggested approach

1. Decide the target: should deploy-only vars be documented in `configuration.md` alongside the binaries' config, or does `docs/deployment.md` stay their home and become a second accepted target for the gate? The latter matches how the docs are already organised.
2. Then derive the "consumed by a deploy entrypoint" set narrowly enough to be true — most likely from the `: "${VAR:?...}"` and `${VAR:-default}` forms in `scripts/deploy/**` rather than any mention anywhere.

Worth confirming the list is only these three first; a var referenced solely inside a heredoc or via indirect expansion could still be hiding.

dungbanguyen:
Closes #566.

## The gap

Obligation 1 requires every `MARGINCE_*` var **Go reads** to be named in the reference doc. Obligations 2 and 3 already count the deploy surface — but only in the *permissive* direction: they ask whether a name a document mentions is still real, never whether a real name is mentioned.

So a var an entrypoint hard-requires was documented if the author remembered. That is exactly the failure mode obligation 1 removed for the Go half.

## Where, and why that is not my choice to make

`docs/deployment.md`, not `configuration.md` — and the file already settles it. Its own opening says *"a var read solely by deploy shell or a workflow is documented in docs/deployment.md by hand"*, and `configuration.md` is *"the table of record for the binaries"*. These vars are read by no binary. Splitting them across two documents by which process happens to read them is how an operator ends up checking the wrong one.

The new obligation is what turns "by hand" into an obligation.

## Scope, inherited rather than re-invented

It reuses obligation 4's `${VAR:?}` set rather than sweeping for mentions, so it inherits that set's **stated limits exactly**: `:-default` is not a requirement (a var with a fallback is not one an operator must supply — requiring those would admit most of the deploy surface and leave the label meaningless), and an explicit guard or `set -u` is invisible to both.

It carries the same floor, for the same reason a prohibition needs one: a sweep that scans nothing passes exactly like a clean one.

| mutation | result |
|---|---|
| undocument a required var | `1 var(s) a deploy entrypoint refuses to boot without, absent from ../docs/deployment.md` |
| make the sweep find nothing | `no ${MARGINCE_…:?} requirement found under ../scripts/deploy — a sweep that scans nothing passes exactly like a clean one` |

**It passes today, which is the point** — this is preventive, not a repair.

## Two corrections to the ticket, both narrowing it

The ticket asks to *"confirm the list is only these three first"*. Confirming it turned out to be most of the remaining work:

- The required set is **`MARGINCE_DSN` and `MARGINCE_OWNER_DSN`** — two, not three.
- **`MARGINCE_ADMIN_PASSWORD` and `MARGINCE_ADMIN_PASSWORD_FILE` are not written in the `:?` form**, so no entrypoint refuses to boot without them. They are not hard-required in the sense this gate is about, and pulling them in would have meant widening the definition to the `:-` form the gate deliberately excludes.

Both of the two are already documented, so no prose needed writing.

`make check-backend` exit 0.

## Summary by CodeRabbit

- **Documentation**
  - Deployment requirements are now validated against the deployment guide to help ensure required environment variables are documented.

- **Tests**
  - Added automated checks that identify undocumented required deployment variables and flag missing deployment requirements.

### DIFF
--- a/backend/gates/envcontract_test.go
+++ b/backend/gates/envcontract_test.go
@@ -79,6 +79,12 @@ var quotedEnvVarName = regexp.MustCompile(`"(MARGINCE_[A-Z0-9_]+)"`)
 const (
 	configurationDoc = "../docs/reference/configuration.md"
 	envExample       = "../.env.example"
+	// deploymentDoc is where a var read only by the deploy surface is
+	// documented. This file's own opening already names it as their home —
+	// "a var read solely by deploy shell or a workflow is documented in
+	// docs/deployment.md by hand" — and obligation 5 below is what turns "by
+	// hand" into an obligation.
+	deploymentDoc = "../docs/deployment.md"
 )
 
 // deploySurfaceRoots are the non-Go trees that configure a deployment: the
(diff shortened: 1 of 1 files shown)


dungbanguyen:
Declining this one — `../docs/deployment.md` is correct, and `../../` would break it. Checked rather than argued:

**The gates package chdirs to the backend root.** `gates_test.go:42` — `os.Chdir("..")` in `TestMain`. So every path in this file is relative to `backend/`, not to `backend/gates/`. That is also why the constant beside it reads `configurationDoc = "../docs/reference/configuration.md"` with a single `../`; a double would have been wrong there for the same reason and that gate has been green for a while.

```
$ cd backend && ls ../docs/deployment.md
-rw-r--r--  16962  ../docs/deployment.md

$ cd backend && ls ../../docs/deployment.md
ls: ../../docs/deployment.md: No such file or directory
```

**And the mutation already proved it end to end.** Renaming `MARGINCE_OWNER_DSN` inside `docs/deployment.md` — the repo-root file — failed the new test by name:

```
1 var(s) a deploy entrypoint refuses to boot without, absent from ../docs/deployment.md
```

A path that did not resolve would have failed differently: `namesIn` reads the file with `os.ReadFile` and fails the test on error, so a wrong path is a read failure, not a missing-name report. Getting the *specific* message back is what shows the file was found, read, and searched.

Resolving as answered rather than actioned.

PR #4142, opened 2026-09-04
A held draft's preflight asks the engine that now decides

dungbanguyen:
Main has been red since #4125 on three suites, and one of them was reporting a real defect rather than drift.

## The regression

#4125 moved the send decision out of request-time preparation and into staging: `refuseUnsendable` stopped asking the purpose gate, and the engine now answers inside the transaction that writes the delivery. The held-draft release did not move with it.

Its preflight runs `PrepareSend` and throws the result away — that used to be how a human learned **now** that consent was withdrawn, while the approval was still pending and they could fix the cause and approve the same row again. After #4125 that preparation answers nothing about consent, so the refusal arrived from staging instead: *after* the approval had committed. An approved row nothing can decide again, a parked run, and a message a human said yes to that will never exist.

`TestAPreflightRefusalLeavesTheDraftPending` says exactly that in its failure text, and it has been saying it since #4125:

```
the approval was decided by a release that refused — the preflight is not
running before the decision, so this message is now unreleasable
```

## The fix

The preflight asks the engine. `PreviewStagingTx` answers the staging question without writing the Art. 5(2) record — a record describes a message that was staged, and this one may never be. It shares `stagingDecisions` with `AuthorizeStagingTx` rather than asking the same question a second way: a preview with its own code would be a second authority, and the one that stopped matching would look exactly like the one that still did.

It previews the **same request** staging will decide on. `PrepareSend` builds the authorization once and carries it, and `delivery()` takes it instead of rebuilding it from the origin, so a preview cannot answer about a different message than the one that would go out.

## The other two suites

Drift, of a kind worth naming. Both replace the delivery machinery with a recording double — and the authority #4125 moved lives **in** that seam. The double removed the gate and then reported that nothing was refused: a suite named for the governed path, passing because it no longer contained one. Both now stage through the real machinery and count the delivery rows it wrote.

The opt-out suite's agent principal gained the seat it acts for. Staging refuses a principal with no `app_user` identity — sending is a human act — and an agent reaches a send through a passport issued against a human's seat. It does not soften the suppression: the gate never consults the principal, which is that suite's whole point.

## Verification

`make check-backend` exit 0, plus the compose, consent, activities and comms integration suites.

Each fix mutation-checked:

| mutation | result |
|---|---|
| drop the preview from the preflight | reproduces the decided-approval failure verbatim |
| disable `refuseAtStaging` | all three suites fail |

## Summary by CodeRabbit

- **New Features**
  - Added a pre-send authorization preview to ensure messages are evaluated against delivery rules before staging.
  - Delivery requests now carry their authorization details consistently through the send process.
- **Bug Fixes**
  - Improved enforcement of consent and suppression rules across message delivery paths, including MCP and agent-initiated sends.
  - Prevented unauthorized or opted-out messages from being delivered while preserving approved sends.

### DIFF
--- a/backend/internal/compose/comms_integration_test.go
+++ b/backend/internal/compose/comms_integration_test.go
@@ -29,6 +29,7 @@ import (
 	"github.com/margince/margince/backend/internal/modules/consent"
 	"github.com/margince/margince/backend/internal/modules/search"
 	"github.com/margince/margince/backend/internal/platform/database"
+	"github.com/margince/margince/backend/internal/platform/jobs"
 	"github.com/margince/margince/backend/internal/shared/apperrors"
 	"github.com/margince/margince/backend/internal/shared/kernel/ids"
 	"github.com/margince/margince/backend/internal/shared/ports/datasource"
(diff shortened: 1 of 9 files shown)


tinnhlt-gra:
Independently verified this clears the red I bisected on #4125 — thank you for picking it up, and for pulling apart which of the three was drift and which was a real defect.

Ran the two failures from my side against your branch (`bb2c36101`) in a clean detached worktree, nothing of mine present:

```
make test-it DIR=backend/internal/compose \
  RUN='TestCommsAdapterSharesTheGovernedPaths|TestAPreflightRefusalLeavesTheDraftPending'

--- PASS: TestCommsAdapterSharesTheGovernedPaths (0.29s)
--- PASS: TestAPreflightRefusalLeavesTheDraftPending (0.20s)
ok      github.com/margince/margince/backend/internal/compose    1.899s
```

For the record, the same two fail on `origin/main` at `62406d2b9`, and the bisect was `2614f6b2a^` PASS → `2614f6b2a` FAIL.

Your reading of the held-draft release matches what the assertion was actually saying — the refusal arriving after the approval commits, leaving an approved row nothing can decide again. That is why I left it alone and asked on #4125 rather than updating the suites to match the new behaviour: from outside I could not tell a closed door from a broken one, and on a consent gate the cost of guessing wrong is asymmetric. Good that it turned out to be worth the pause.

This is currently blocking every open PR, mine (#4066) included, so no rush needed from my side beyond it landing.

dungbanguyen:
Valid — applied. The aliasing defeats a guarantee this file states in as many words: *"a caller outside this package may carry one from PrepareSend to SendPreparedTx and may not forge one."* Handing back the backing arrays is a forge without touching a field — staging decides on `p.authorization`, so rewriting a recipient through the previewed value would send to somebody the engine was never asked about.

Both slices are now cloned, and the regression test is mutation-checked. Dropping the clone:

```
sendcoreauthorization_test.go:39: the staged recipient is now "stranger@elsewhere.test"
    — a caller rewrote the send through the request it previewed
sendcoreauthorization_test.go:42: the staged links were rewritten through the previewed request
```

`make check-backend` exit 0.

dungbanguyen:
Flagging that this is **`DIRTY`** — main has moved under it and it needs a rebase to merge.

It matters more than usual right now: this is the only fix in flight for the red compose integration lane (#4145), and roughly fifteen open PRs are queued behind it. No unresolved review threads, so the rebase looks like the only thing between it and merging.

I have not touched the branch — it is not mine to rebase. I claimed #4145 before checking open PRs against the failing files, found this, and retracted; the diagnosis here is better than the ticket's, which stopped at "not #4143".

dungbanguyen:
Closing as redundant — #4149 landed the same fix while this was in CI, and it got there first.

We converged independently on the same design: a preview on `consent.Gate` sharing the decision body with `AuthorizeStagingTx`, `PreparedSend.Authorization()` exposing the request rather than rebuilding it downstream, both test suites moved onto the real delivery machinery, and the opt-out suite's agent principal given the seat it acts for. No point carrying a second copy.

One thing worth recording rather than losing. This branch made `Authorization()` carry a single pre-built request, which aliased its `Recipients` and `Links` backing arrays into the value handed out — a caller could rewrite the staged send through the request it had merely previewed. #4149 is not exposed to it: `p.message.authorization(p.origin)` rebuilds through `EmailRecipients` and `linkedRecordIDs`, both of which allocate, so every call returns independent slices. Noting it only so the property is known to be load-bearing — a later change to carry the request instead of rebuilding it would need to clone.

Verified main is green at bd76b461c: `backend/internal/compose` passes in full.

Issue #4075, opened 2026-09-04
The license gate is non-deterministic: syft's network enrichment fails per-package and reads as a denied license


dungbanguyen:
`make sbom-check` denies packages whose licences are perfectly ordinary and present. Which packages it denies **changes between runs of the same commit**.

## Evidence

Three of my PRs — all comment/workflow-only, none touching a dependency — had their license gates start within **four seconds** of each other:

```
3978 failure 2026-09-04T06:14:03Z
3981 failure 2026-09-04T06:14:06Z
3983 failure 2026-09-04T06:14:07Z
```

All three failed identically:

```
by compliance: 2 denied, 524 allowed, 9 ignored, 2 unlicensed
  @csstools/css-syntax-patches-for-csstree  1.1.8   (no licenses found)
  @esbuild/win32-arm64                      0.28.2  (no licenses found)
```

Then, with **no change to any commit**:

- Re-running 3978's gate alone → **passed**.
- Re-running 3981's gate alone → **failed on a different package**: `lightningcss-linux-arm64-gnu 1.32.0 (no licenses found)`.

## The licences are not missing

Both originally-denied packages carry a licence in the npm registry at exactly the pinned versions:

```
@csstools/css-syntax-patches-for-csstree@1.1.8 → MIT-0
@esbuild/win32-arm64@0.28.2                   → MIT
```

Both are already in `.grant.yaml`'s `allow`. And a local `make sbom` on the same tree resolves them correctly:

```
'pkg:npm/%40csstools/css-syntax-patches-for-csstree@1.1.8' | licenses=[{'license':{'id':'MIT-0'}}]
'pkg:npm/%40esbuild/win32-arm64@0.28.2'                    | licenses=[{'license':{'id':'MIT'}}]
```

## Cause

`.syft.yaml` sets `enrich: all`, and its own comment says so: *"syft pulls Go licenses from the module proxy and npm licenses from the registry, so `make sbom` needs network access — it is not an offline scan."*

When a registry lookup fails, the package resolves to **no licence** — and `grant` cannot tell "we could not ask" from "it has a forbidden one", so it denies. Three concurrent scans hitting the registry together is enough to trip it; the moving failure set is what a rate limit looks like and what a missing licence does not.

## Why it is worse than an ordinary flake

1. It **accuses a dependency of a licence problem it does not have**, in a supply-chain gate — the one place a false positive is most likely to be believed and acted on.
[...]

PR #4154, opened 2026-09-04
A lead-vocabulary change is published, not just audited

### LINKED ISSUE #1873: Lead source and disqualify-reason mutations write an audit row but no outbox event
Found by the Codex review of #1872 and deferred.

`backend/internal/modules/people/leadsource.go` (CreateLeadSource, UpdateLeadSource, DeleteLeadSource) and `leaddisqualifyreason.go` (the same three) write the `audit_log` row inside the transaction but stage no `event_outbox` row. The pipeline catalog emits `pipeline.created` for the same kind of configuration change ("config changes are first-class facts"); the custom-field catalog today writes audit only, which is the precedent these two followed.

Why deferred: the events catalog (`backend/api/public-events.yaml`, `shared/kernel/events/catalog.go`) has no type for a lead-vocabulary change, and adding one is a public event-contract change that deserves its own PR with the webhook fixtures.

What to do: add `lead_source.changed` / `lead_disqualify_reason.changed` (or one `lead_vocabulary.changed`) to the catalog, emit it from the six mutations via `storekit.EmitEvent`, and decide whether the custom-field catalog should follow.

dungbanguyen:
Closes #1873.

The six mutations behind the two administered lead vocabularies wrote an `audit_log` row and staged nothing on the outbox. An admin could rename or delete a lead source and no subscriber was told — a consumer caching the source list to render or group by it kept serving an entry that no longer existed, and the only record of the change was one nobody subscribes to. events.md §5.3b: config changes are first-class facts, which is why `pipeline.created` already exists for the same kind of edit.

The six waivers said the closed catalog defined no type and the closed-verb law forbids inventing one build-side. Both were true; neither is any more. The types are added **contract-first** in `public-events.yaml` with the generators following — the sanctioned path, not a build-side invention. All six waivers are deleted, which is what turned `TestEveryAuditedMutationEmitsAnEvent` green rather than a new exception.

## Two types, not one

A subscriber caching the source list and one reporting on disqualification reasons are different readers, and the catalog is fine-grained everywhere else (`organization.created`/`updated`/`archived`/`merged`). Both ride the **lead stream**, because their entries are values every lead carries.

## What each carries

| | identifier | follows a rename? |
|---|---|---|
| `lead_source.changed` | `key` | no — the key never changes, and it is what leads carry |
| `lead_disqualify_reason.changed` | `label` | yes — a reason has no key, the label *is* its identity |

Neither publishes a label beside a key. A label is mutable and still readable while the entry exists, so putting it on the wire only invites a subscriber to trust a copy that can go stale. After `deleted` the identifier is the only thing naming what went away — which is why the wire fixtures pin the `deleted` case.

Both entity types join `workspaceLevelEntities`, for the reason `pipeline` and `stage` are there: workspace configuration with no per-owner row scope to probe. Every seat that can see a lead sees the same source list.

## The ticket's open question

**The custom-field catalog should not follow.** Its waiver rests on two things that do not hold for the lead vocabularies: a cross-object catalog change has no single family stream to ride — the very thing these two *do* have — and `docs/explanation/custom-fields.md` ratifies audit-only in as many words ("Catalog changes are **audit-only by ratification**"). Left alone.

## Verification

`make check-backend` exit 0, plus the people and webhooks integration suites.

Mutation-checked:

| mutation | result |
|---|---|
| drop the delete emit | `lead_source.changed = [created:webinar updated:webinar]`, missing `deleted:webinar` |
| publish `before.Label` instead of `before.Key` on update | `updated:Webinar` instead of `updated:webinar` |

The second is the one that makes the key-not-label decision testable rather than merely stated.

## Summary by CodeRabbit

* **New Features**
  * Added webhook events for lead source and lead disqualification reason changes.
  * Events are published when these configurations are created, updated, or deleted.
  * Event payloads include the affected identifier, change type, and source key or reason label.
  * Workspace-level subscriptions can now receive these configuration events.


dungbanguyen:
Note: this branch picked up a `Merge branch 'main'` commit from outside this session. I rebased instead and force-pushed — the trees were byte-identical (`git diff --stat` between the two was empty), so nothing was dropped.

Rebasing rather than merging is deliberate here: `gh pr update-branch` produces a merge commit whose diff against base includes main's lockfile movement, which trips the `deps` path classifier and runs the license gate — currently non-deterministic (#4075). A rebase keeps the diff to what this PR actually changes.

dungbanguyen:
SonarCloud reports 6.0% duplication on new code (gate ≤3%). Recording rather than quietly merging past it: that duplication is `emitLeadSourceChanged` and `emitLeadDisqualifyReasonChanged`, which are structurally the same shape with different payload types, field names and error strings.

It is the direct cost of publishing two event types instead of one, which the PR argues for on its merits — a subscriber caching source keys and one reporting on disqualification reasons are different readers. Collapsing them behind one generic emitter would trade two short, readable functions for indirection over two payload structs that share no interface, and would not make the pair any easier to keep in step.

SonarCloud is advisory here; `ci` is the required check and is green.

Issue #4152, opened 2026-09-04
A connector-captured thread row offers "Visibility," and the server refuses it with a control that no longer exists


tinnhlt-gra:
## Observed

Measured on core `bd76b461c`, 2026-09-04, as `uat-rep@demo.test` (the member whose mailbox captured the thread).

A connector-captured, two-message conversation (`captured_by: connector:ext:openchannel:…`, `kind: message`, `thread_key: openchannel:…`) drawn on the contact's timeline offers three row actions: **Reply**, **Relink**, **Visibility** — the same three a plain, hand-logged row gets. Pressing **Visibility** opens the per-message audience dialog, and saving it (`PATCH /v1/activities/{id}/audience`) answers:

```
422 validation_error / audience_is_derived
"This message came from a mailbox, so who can read it follows from what each
importing mailbox asks for. Share or keep the thread private instead."
```

There is no "Share or keep the thread private" control anywhere on the page. `ThreadAudienceAction` — the thread-level control the error message refers to — no longer exists in the frontend source at all (`grep -rn ThreadAudienceAction frontend/src` returns nothing). It looks like it was removed by `47793146d` ("A contact says who can see it, and the edit controls fail closed", #4036).

The dispatch left behind, `frontend/src/screens/timelineactions.tsx`:

```tsx
{activity.kind !== "email" && (
  <AudienceAction activity={activity} entityType={entityType} entityId={entityId} />
)}
```

Its own comment states the assumption that makes this wrong for a captured thread: *"Every other kind keeps its control. A note, a call or a meeting has no drawer to be sent to, and its audience is never derived — there was never a question to guess at for those, only the one write."* A connector-captured `kind=message` row is exactly the counter-example: its audience **is** derived, the same way an email's is, and it has no drawer either — email's own `EmailAccessEditor` is gated `activity.kind === "email"` and reaches nothing else.

No content leaks — the write is refused before anything changes. This is a broken control, not a disclosure: a member trying to keep a captured conversation private gets a dialog that opens, accepts input, and then answers with a validation error naming an action the page does not offer.

## Cost

A colleague who captures a sensitive conversation through a connected mailbox has no way, from the product's screen, to limit who else in the workspace can read it — pressing the only control the row offers ends in a confusing, mailbox-jargon refusal that names a control the page does not have.

## Workaround

[...]

PR #4157, opened 2026-09-04
main is red: the flush search input names a size off the ramp

dungbanguyen:
`type.test.ts` fails on `main`, so every frontend lane is red.

#4026 gave `.input-icon-flush .input` a literal `font-size: 14px`. The ramp's neighbouring rungs are `--fs-sm: 13px` and `--fs-body: 13.5px`, and the gate's own words are the reason it fails: *"12.5px beside 13px is not a decision a reader can see"*.

`--fs-body` is what `.input` itself uses, so the flush variant now agrees with the control it modifies instead of disagreeing with it by half a pixel — which is also why this needed no judgement call about which rung to pick.

## Coordination

Not my ticket's work — I hit it running `make check-fe` for #4152. I checked open PRs touching `atoms.css` and `type.test.ts` before starting and found none, so I am fixing it here and saying so.

`make check-fe` green.

### DIFF
--- a/frontend/src/design-system/atoms.css
+++ b/frontend/src/design-system/atoms.css
@@ -1280,7 +1280,7 @@
   border-radius: 0;
   height: auto;
   padding: 0 0 0 24px;
-  font-size: 14px;
+  font-size: var(--fs-body);
 }
 .input-icon-flush .input:focus,
 .input-icon-flush .input:focus-visible {


joshbruh-lang:
The rung passes the gate, but the defect underneath it is still there — flagging it rather than sitting on it, since I hit the same line from the other side and my branch resolved the conflict the other way.

`.input-icon-flush .input` sits OUTSIDE the `@media (max-width: 35rem)` block and outranks `.input` by specificity, so whatever size it names beats the iOS floor a few rules above. `--fs-body` is 13.5px, so a flush field on a phone is still under 16px and still zooms the viewport on focus — which is exactly what that floor's waiver exists to prevent, and it says so in as many words:

```css
@media (max-width: 35rem) {
  .input {
    font-size: 16px; /* ds:ignore iOS zooms any field under 16px — a platform floor, not a rung */
  }
}
...
.input-icon-flush .input {          /* (0,2,0) — beats .input (0,1,0), media query or not */
  font-size: var(--fs-body);        /* 13.5px on a phone */
}
```

On `feat/onboarding-welcome-ground` I removed the declaration instead: the flush variant strips the field's *chrome* — border, ground, radius — and type is not chrome, so inheriting puts it back under both `.input`'s desk size and the phone floor. That satisfies the type gate too, since there is no size left to be off the ramp.

Not asking for anything here — #4157 is merged and green, and I am not going to re-litigate a fix that unblocked main. But if the flush field is meant to keep a size of its own, it needs the flush selector added to the floor's media block as well, or the zoom comes back on the next reader who reaches for a rung.

---

PR #4166, opened 2026-09-04
The leads board has both ends of the pipeline

### LINKED ISSUE #1886: Leads board: terminal columns (Qualified/Disqualified) with drop-to-qualify and drop-to-disqualify
Deferred from #1883 (the lead work-surface PR).

The board on `#/leads` still shows the three open columns (New · Contacted · Engaged). The planned shape has two more, collapsed by default with a count in the header: Qualified and Disqualified. Dropping a card on Qualified should open the qualify dialog (`leads.qualify.tsx`), dropping on Disqualified the disqualify dialog (`leads.disqualify.tsx`) — never a bare status PATCH, which the server refuses for the terminal states.

Why deferred: the list query excludes archived (terminal) leads by default, so the two columns have neither rows nor a count to show without a second, `include_archived=true` read or a server-side count per status. `PipelineBoard` (design-system/composed.tsx) also has no collapsed-column notion yet.

What to do: add a per-status count to `GET /leads` (or a small aggregate endpoint), teach `PipelineBoard` a collapsed column, and wire the two drops to the dialogs.

dungbanguyen:
Closes #1886.

The board showed New · Contacted · Engaged and stopped there. A lead that was qualified or disqualified simply vanished from it, and there was nowhere to drop one to say so — both transitions were reachable only from a lead's own page.

## Why the count had nowhere to come from

This is what deferred the ticket. A lead is **archived** the moment it reaches either terminal status, and every list this product serves excludes archived rows. The board's own page cannot see a single promoted lead, so a column counting its cards would read zero however many had passed through.

So the counts come from a report. `leads-by-status` is the one lead key that counts archived rows — the only spec in the registry with no archived predicate, which is the point rather than an oversight. Row scope still binds, so a count never includes leads the caller's own list would withhold.

## Two engine assumptions this uncovered

Every spec until now restricted *something*. `buildReportWhere` seeded its clause list with the base predicate and `reportSQL` always wrote `WHERE`, so a spec that restricts nothing rendered `WHERE GROUP BY` and answered 500:

```
report leads-by-status: ERROR: syntax error at or near "GROUP" (SQLSTATE 42601)
```

Both now treat an absent restriction as absent, rather than making every such spec spell a no-op `TRUE`.

## The board

`PipelineBoard` learns a **collapsed** column: head and count, no cards, still a drop target. The cards are not rendered rather than hidden with CSS, and the head becomes a button only on a board that passes `onToggleColumn` — a control that does nothing is worse than no control, and a deal board would otherwise have grown one on every stage.

**A drop on either terminal column opens its dialog.** Neither transition is a status change: qualifying promotes the lead into a person and maybe a deal, disqualifying records a reason. The server refuses a bare status PATCH into both, so a board that sent one would put a refusal in front of a reader who did the ordinary thing.

**Opening a folded column fetches its leads.** A column that expanded to nothing while claiming twelve would be worse than one that does not open — and these rows are archived, so this one read asks for them by name, and only when opened.

`lead.boardTerminalOnly` said *"the board shows open leads only"*, which this makes false. Reworded in both locales.

## Verification

`make check-backend` exit 0; `backend/internal/compose` integration green; full frontend suite 6358 passed; biome clean; `tsc -b` clean.

Mutation-checked:

| mutation | result |
|---|---|
| count the cards instead of the report | both terminal columns read 0 |
| let a terminal drop fall through to the status PATCH | no-PATCH and dialog assertions fail |
| draw the terminal columns expanded | 2 tests fail |
| give the spec `whereArchivedNull` | `counted 0 promoted, want 1` |

## Summary by CodeRabbit

- **New Features**
  - Added a leads-by-status report with counts grouped by status, owner, and source, including promoted and disqualified leads.
  - Terminal columns on the lead board are now collapsed by default, show status counts, and can be expanded to view archived leads.
  - Dropping a lead into a terminal column now opens the appropriate qualification or disqualification dialog.

- **Documentation**
  - Updated report and tool references to include the new report and revised catalog metrics.

- **Bug Fixes**
  - Improved report generation for reports without filtering conditions.


dungbanguyen:
All five valid — applied. Verified each against the code first; none were false positives, and two of them contradicted things the diff claimed about itself.

**Grid track.** Right, and the fold bought nothing because of it. `.board` gave every column an implicit `minmax(220px, 1fr)` track, and a track is sized by the board, not by the item in it — `width: max-content` on the item could not shrink it. The board is flex now: `1 1 220px` reproduces "at least 220, share what is left", and a folded column opts out with `0 0 auto`.

**Toggle props.** Right, and this is exactly what the comment beside it claimed the design already did ("a control that does nothing is worse than no control"). It was per *board*; it is per *column* now — a column declares it can fold by carrying `collapsed` at all, so the three open lead stages keep the plain div.

**Rows past fifty.** Right, and worse than truncation on its own: the head states the report's true count, so a column could say 12,000 and show 50. The read walks the cursor now and the open column carries a load-more.

**Error states.** Right — the counts fell back to `0` and the rows to an empty column, so "nobody was ever disqualified" was indistinguishable from "we could not ask". Both now name which read failed. They say it in their own words rather than reusing the generic problem text: the existing refused-move test caught three identical error lines stacking up, which is its own argument against the generic message.

**Terminal drag.** Right. The card no longer offers the gesture, and `dropLeadOn` refuses a terminal source anyway — a drop carries an id through a `dataTransfer` string that no card had to originate, so the guard belongs where the decision is taken too. With it in place the "is it already there" comparison became unreachable, which TypeScript pointed out:

```
TS2367: This comparison appears to be unintentional because the types
'"new" | "contacted" | "engaged"' and '"promoted" | "disqualified"' have no overlap.
```

Regression test added as asked, and mutation-checked: dropping either half of the guard fails it.

One thing your review led me to that it did not name — `vi.ts`. I had reworded `lead.boardTerminalOnly` in `en` and `de` only; the key-parity gate caught the third catalog when I added the two new error keys.

`pnpm lint` clean, `tsc -b` clean, 6359 frontend tests pass, `make check-backend` exit 0.

PR #4169, opened 2026-09-04
worklistFrom hands its two side sources their own ranking

dungbanguyen:
Main is red on `craftsmanship`:

```
../../backend/internal/compose/attention/worklist.go
  185: [MAJOR] long-func — worklistFrom is 85 code lines (> 80) — extract helpers
BLOCK — 0 blocker, 1 major, 0 minor
```

Reproduced on plain `main` (1efecf16d); #4161 grew the function past the cap.

The two blocks extracted are the ones that were already separable — the waiting customers and the overdue leads. Each collects rows from a source read *beside* the assembled day, sorts them oldest-first, and marks the ones past the lead as crowded; neither touches the day's own rows, and between them they were a third of the function.

Behaviour is unchanged, and one detail had to be preserved rather than tidied: `rankedWaits` narrows to scope **before** marking crowding, because crowding is a fact about position on the page — marking it over rows the scope then removes demotes a reader's second waiting customer for standing behind seven of a colleague's. The comment saying so travels with the code that depends on it.

`rankedWaits` assigns through the slice index rather than the loop's copy, which the original achieved by appending; `rankedLeads` returns its own slice for the same reason.

`make check-backend` exit 0, `make craft-static` PASS, and the attention unit + integration suites green.

### DIFF
--- a/backend/internal/compose/attention/worklist.go
+++ b/backend/internal/compose/attention/worklist.go
@@ -18,6 +18,7 @@ package attention
 import (
 	"context"
 	"sort"
+	"time"
 
 	crmcontracts "github.com/margince/margince/backend/internal/contracts"
 	"github.com/margince/margince/backend/internal/shared/kernel/ids"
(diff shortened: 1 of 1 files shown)


dungbanguyen:
Superseded by #4178, which carries this exact `worklist.go` extraction plus the second main repair (`integration shard (4/6)`, the forged meeting kind). Both were mine and I opened them against the same red main from two worktrees — my mistake, not two findings.

Closing this one rather than #4178 because main is red in two ways and #4178 fixes both; splitting them across two PRs would cost a second full CI run for no extra coverage.

PR #4178, opened 2026-09-04
Two main repairs: the forged meeting kind, and worklistFrom's length

dungbanguyen:
Main is red on `integration shard (4/6)`. `TestTheLimiterStillHoldsAMeetingThatNamesAJudgedSender` fails from #4164 — reproduced on plain `main`, passing at `1efecf16d` and failing at `438c95405` — and it is failing about something worth keeping.

```
a meeting-kind record naming a judged sender was born audience="workspace" reason="",
want participants/no_record — the kind is a word the caller chose, and the ladder
judged the person who sent it
```

## Why the two are in tension

#4164 made a meeting the fourth and strongest sense of *"we have dealt with this person"*, which is right — a founder mails forty people and hears from six, while nobody spends an hour by accident. It also let that meeting **outrank a settled judgement** about the sender, so a partner known only from a calendar stops being suppressed by the ESP registry.

The trouble is where `kind` comes from. A connector reports it beside the raw bytes, the extension ingress copies `Activity.Kind` straight off a third-party record with no vocabulary check, and the column carries no `CHECK` constraint. **"meeting" is a word a caller writes.** With the override in place, a suppressed sender could land one message that calls itself a meeting and have that same message be the evidence lifting the suppression standing against it — published to the whole workspace on the strength of its own label.

## The fix

The read answers two questions instead of one, in the same query:

| | counts | used for |
|---|---|---|
| `met` | every qualifying meeting, **including the row being captured** | creating the contact — a calendar record names a guest, and the guest is the point of it |
| `metElsewhere` | only meetings **other than the row being decided about** | outranking a settled judgement |

A meeting on some other row is evidence; the row under judgement is not evidence about itself. #4164's headline case — `TestTheGuestOnACapturedMeetingBecomesAContact` — is untouched.

## What I did not do

Reverting `positive()` to `corresponded` alone also makes both tests pass, and throws away the thing #4164 set out to fix. No test covered that arm — it was asserted in prose — which is exactly why the blunter fix looks equally good from the outside. Worth knowing when reviewing this.

## Verification

`make check-backend` exit 0; `internal/modules/capture` and `integration/capture` suites green.

Mutation: read `met` where `positive()` reads `metElsewhere` →

```
--- PASS: TestTheGuestOnACapturedMeetingBecomesAContact
--- FAIL: TestTheLimiterStillHoldsAMeetingThatNamesAJudgedSender
```

That pair is what shows the distinction is the load-bearing thing, rather than the meeting read as a whole.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved meeting-based contact evaluation by distinguishing the current captured activity from other qualifying meetings.
  - A meeting record can no longer independently serve as evidence to lift a communication suppression; separate meeting activity is required.
  - Updated activity processing to apply this evidence consistently when deciding whether a counterparty can be recorded.

### DIFF
--- a/backend/internal/compose/attention/worklist.go
+++ b/backend/internal/compose/attention/worklist.go
@@ -18,6 +18,7 @@ package attention
 import (
 	"context"
 	"sort"
+	"time"
 
 	crmcontracts "github.com/margince/margince/backend/internal/contracts"
 	"github.com/margince/margince/backend/internal/shared/kernel/ids"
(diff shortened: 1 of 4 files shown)


dungbanguyen:
Carrying the `worklistFrom` length fix (#4169) as a second commit, because the two were blocking each other: main is red on both, so #4169 failed shard 4/6 on the meeting bug this PR fixes, and this PR failed `craftsmanship` on the length break #4169 fixes. Neither could go green alone without bypassing a check.

Two unrelated fixes in one PR is not what I would normally do — flagging it plainly rather than quietly. Both are main-repair, each is its own commit with its own reasoning, and they are separable by revert.

#4169 will be closed as folded in once this lands. `make check-backend` exit 0 and `make craft-static` PASS with both applied.

PR #4182, opened 2026-09-04
A judgement is drawn below the day, not inside it

LarsGradion:
## What this changes

The queue mixed three jobs. A buyer waiting on a reply is work a seller executes; a duplicate pair is a judgement somebody makes before work can continue; a stopped mailbox is a source an administrator restores. Drawn in one list they competed for the same attention — a rep scanning for their next call stepped over the product's own housekeeping to find it.

**Judgements are drawn below the day now, in their own panel.**

## The split is the server's, not this client's

`destination` shipped in #4127 and nothing read it until now. That matters more than it looks: the counts above the queue are computed from the same field, so a client re-deriving the split from `source` or `category` would eventually disagree with the figures it is drawn beside. This reads the field; it does not reimplement the rule.

## Three decisions worth naming

**Below, never hidden.** Work nobody can see is the reason it goes undone, so the split moves the row rather than dropping it.

**A day with no judgements draws no panel.** An empty section headed "To review" reads as a screen that failed to draw its contents.

**A row an older server did not classify stays in the day.** Reading that silence as review would empty a rep's queue the moment they talked to an older server; reading it as seller work leaves the page exactly as it was before this split existed. One is a regression nobody would attribute to version skew.

**Ranks are still counted over the whole queue.** A row's number is its place in the day the server ranked, not its place within whichever panel it landed in — renumbering per panel would tell a reader the fourth-most-urgent thing is their first.

## Evidence

| Obligation | Evidence |
|---|---|
| User-visible outcome | `keeps a judgement out of the queue and still shows it` — asserts the pair is absent from Today **and** present in Review; presence alone would pass either way |
| Empty state | `draws no review panel on a day that holds none` |
| Version skew | `keeps a row an older server did not classify in the day` |
| Mutation strength | `unbandedRows(today)` reverted to `unbandedRows(queue)` — the split test fails |
| i18n | en/de/vi all carry `worklist.review` ("To review" / "Zu prüfen" / "Cần xem xét") |
| Full lane | `make check-fe` green |

## Noted, not fixed

`Panel` renders a bare `<section>` with no accessible name, so it is not a landmark and cannot be queried by role. The tests reach it through its heading, with a comment saying why. Giving `Panel` an `aria-labelledby` is a design-system change and belongs in its own PR.

---
## Summary by cubic
Splits the worklist into seller work and review work using the server's `destination` field, so review rows render below the day instead of inside the queue. Rows an older server doesn't classify stay in the day, and a day with nothing to review draws no review panel.

**Notes**
- The split reads `destination` instead of re-deriving it from `source` or `category`, so it can't drift from the summary counts.
- Rows are now numbered within their own panel, so both lists start at 1 and no panel shows a broken-looking sequence.
- The day's pane only fills from day rows, and the day no longer draws an empty heading for the review band that moved below it.
- Known limits: Show more sits inside Today and does nothing when the next page is all review, and the header still totals both kinds of work in one sentence.
- Tests cover the partition (no row lost), the empty state, version-skew rows, band headings, and per-panel ranks.
- `Panel` still has no accessible name, so tests reach it through its heading.

<sup>Written for commit 8d67bc97d71f19e0179a9423740e2cd4a35993c2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/4182?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/i18n/de.ts
+++ b/frontend/src/i18n/de.ts
@@ -7898,6 +7898,7 @@ export const de = {
   // for every fact the server sends as a closed vocabulary.
   "worklist.loading": "Dein Tag wird gelesen…",
   "worklist.queue": "Heute",
+  "worklist.review": "Zu prüfen",
   "worklist.more": "Mehr anzeigen",
   "worklist.more.failed": "Konnte nicht mehr laden. Bitte erneut versuchen.",
   "worklist.summary":
--- a/frontend/src/i18n/en.ts
+++ b/frontend/src/i18n/en.ts
@@ -7989,6 +7989,7 @@ export const en = {
   // for every fact the server sends as a closed vocabulary.
   "worklist.loading": "Reading your day…",
   "worklist.queue": "Today",
+  "worklist.review": "To review",
   "worklist.more": "Show more",
   "worklist.more.failed": "Could not load more. Try again.",
   "worklist.summary":
--- a/frontend/src/i18n/vi.ts
+++ b/frontend/src/i18n/vi.ts
@@ -7810,6 +7810,7 @@ export const vi = {
   // for every fact the server sends as a closed vocabulary.
   "worklist.loading": "Đang đọc ngày của bạn…",
   "worklist.queue": "Hôm nay",
+  "worklist.review": "Cần xem xét",
   "worklist.more": "Xem thêm",
   "worklist.more.failed": "Không tải thêm được. Hãy thử lại.",
   "worklist.summary":
--- a/frontend/src/screens/worklist.bands.test.tsx
+++ b/frontend/src/screens/worklist.bands.test.tsx
@@ -62,7 +62,9 @@ describe("a band holding nothing says so", () => {
     renderWorklist("en");
 
     expect(await screen.findByText("Nothing needs you today.")).toBeTruthy();
-    expect(screen.getByText("Nothing to review.")).toBeTruthy();
+    // A band that stays in the day. `review` is drawn in its own panel below
+    // now, so it declares no empty run here.
+    expect(screen.getByText("No new pipeline work waiting.")).toBeTruthy();
   });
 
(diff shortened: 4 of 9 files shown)


LarsGradion:
## Review round: three fixed, two recorded

**Fixed in `8d67bc97d7`** — all three were the same shape: the page was built around one list, and cutting it in two left things still describing the whole.

- **The day reported work clear that had moved.** `day.bands` counts every row the server ranked, so with judgements filtered out the review band drew an empty run *inside* the day — a heading saying "Nothing to review" while that work sat in the panel below. The tests could not have caught it: every fixture row omits `band`, so all of them fall through to the unbanded tail and the band path was never exercised.
- **The ranks read as broken.** Numbering across the whole day puts 1, 4, 7 in one panel and 2, 3, 5 in the other. Each panel numbers its own rows now.
- **The pane took rows from both panels.** A selected review row drew its context beside Today while the highlighted row sat below.

**Recorded, not guessed at** — both are product decisions rather than defects with one right answer:

1. **Show more sits inside the Today panel.** If the next page is all review-destined, pressing it changes nothing in the panel holding the button. Options: move the control outside both panels, give Review its own, or label it for the day. Wants a design call.
2. **The header totals both kinds of work in one sentence.** `summary.buckets` already partitions seller work from review, so the header could say "12 today · 5 to review" instead of one number spanning two panels. That is a copy change with a contract already behind it, but it changes the sentence every rep reads every morning.

Filing both rather than picking one inside a PR about the split.

Issue #1857, opened 2026-08-19
The scheduled-pass collapse changed the vocabulary but not the shape: tenant fan-out dispatchers still exist


LarsGradion:
Decided 2026-08-19 (option a), covering the gap named in both ADR-0103 and ADR-0063 — one piece of work, two records describing it.

**The decision (ADR-0103).** A scheduled pass over the installation is one job declaration, not two. The dispatchers that existed only to enumerate workspaces and enqueue one child each collapse into the pass itself.

**What is true.** The role vocabulary and the two response schemas shipped. The collapse did not. The fan-out dispatchers are still there, so the code says one thing and does another.

**The second half (ADR-0063).** Mailbox backfill is bounded and previewed — both built. The single nightly dispatcher it also called for was never built. Same shape, same area, and finishing one without the other leaves the confusion in place.

**Cost.** Nothing is broken. It costs a reader: the names describe a structure the code does not have, which is the kind of drift that makes the next person add a third layer rather than remove the second.

**Why one piece of work.** Both records describe the same unfinished cleanup in the job runtime. Doing them separately means opening the same files twice and reasoning about the same fan-out twice.

**Note for whoever picks this up:** `backend/internal/compose/jobmetrics.go` carries the `workspace_id` metric label that ADR-0080 licenses specifically because the fan-out exists. When the fan-out goes, check whether that label still earns its exception — the record says it goes with it.

dungbanguyen:
Picking this up. Two things I found while scoping that the ticket does not name, recording them before I start rather than deciding silently.

**Scale.** 31 kinds carry `role: dispatcher`. 27 of them fan out per **workspace** and are the vestigial ones this decision is about — `InstallationWorkspace` refuses more than one active workspace (`ErrMultipleWorkspaces`), so each of those enumerates exactly one row and enqueues exactly one child. The other 4 fan out per **connection** or **build** (`gmail_sync`, `gmail_watch_renew`, `graph_watch_renew`, `telegram_poll_sweep`, `voice_build_retry`) and are genuine one-to-many; those stay.

**Retiring 27 job kinds needs a migration, and there is no mechanism for it.** I could not find any retired-kind handling in `platform/jobs`. When the child kinds stop being registered, any `river_job` rows of those kinds still queued at deploy time become unworkable and sit there failing rather than draining. It is at most one tick's worth per kind, but it is permanent debris in `river_job` and it would page somebody.

[...]

PR #4193, opened 2026-09-04
Two more passes collapse: finance_sync and embed_drift

dungbanguyen:
Batch two of #1857, stacked on #4192 (base is that branch, not main).

Both follow the shape the first PR established: the child kind goes, the dispatcher absorbs the work through `runPerWorkspace`, and the child's queued rows join the drain migration.

**`finance_sync` folds the child's worker into the sweep's** rather than keeping two types. The child carried the pool, the logger, the sync principal and `providerFor`; the dispatcher carried only a pool. Keeping both would have left a type whose whole purpose was to be constructed once and called once.

## The deadline is not always the child's five minutes

This is what the second conversion taught, and it is worth flagging for the remaining 23.

`embed_drift_workspace` declared `timeout: {none: true}` deliberately: a negative River timeout keeps a row outside `job_rescuer`, and the pass walks a corpus bounded by its backlog rather than a wall clock. Giving the collapsed kind the five minutes the first two pairs used would have **re-armed the rescuer behind the longest pass in the tree**, and nothing about the diff would have looked wrong.

`TestTheTwoPassesDeclaredWithNoWallClockStillResolveToNegativeOne` caught it:

```
embed_drift_workspace is not declared in api/jobs.yaml
```

and then had to change. It asserted `embed_drift_sweep` **bounded**, on the ground that a dispatcher only enumerates and enqueues, and asserted the child unbounded. Those are now one kind, so it changed sides rather than leaving — which is the whole content of the collapse: there is no longer a row whose deadline can be short because it does nothing.

## Verification

`make check-backend` exit 0; `internal/compose` and `integration/capture` integration suites green. Kind count 91 → 89.

Remaining after this: 23 workspace pairs.

### DIFF
--- a/backend/api/jobs.yaml
+++ b/backend/api/jobs.yaml
@@ -246,14 +246,12 @@ kinds:
       rest due for tomorrow.
 
   finance_sync_sweep:
-    role: dispatcher
+    role: worker
     go_type: FinanceSyncSweepArgs
     queue: default
-    timeout: 2m
+    timeout: 5m
     opts_owner: caller
     cadence: 6h
-    fans_out_to: finance_sync
-    fan_out_unit: workspace
     reason: >-
       Four times a day, and at boot like every pass here. An accounting source
       is not a stream: invoices are issued and settled on a human timetable, so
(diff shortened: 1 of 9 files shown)


dungbanguyen:
Rebased onto main now that #4192 has landed, and retargeted the base from that branch to main.

One thing changed in the rebase beyond the mechanics. This PR originally added its two kinds to the drain migration #4192 shipped; that is wrong, because a migration that has already run does not run again — the two kinds would have been drained only on installations that had not yet applied the first one, which is to say none of the ones already on it. They now have their own migration, and #4192's file is restored byte-for-byte to what it shipped.

PR #4217, opened 2026-09-04
The thread is teased on one row, and folded until asked for

joshbruh-lang:
## What

On the account's 360, "what happened" no longer opens expanded. It is one
folded row that **teases** what is under it — the section's name, how much it
holds, and the newest exchange — and the whole row is the control that opens
it, with "Full history" at its end.

```
›  What happened · 2   Rate sheet sent · 2 June 2026          [ Full history ]
```

## Why

It was open on arrival, and the reasoning for that was sound as far as it went:
the thread is what the call above it was read *from*. But it cost the 360 its
whole lower half before a reader had asked for anything, and the 360 exists to
be taken in at a glance.

Folding it is only half an answer, though — a control that says just "show"
makes a reader open it to find out whether it was worth opening. So the row
carries the thing they wanted the thread for: **the newest exchange, in the
words its own row will use once the fold opens.** The decision to open is now
made before the click.

**The whole row is the control**, not a button beside the words describing it.
It is the design system's `Disclosure`, whose `<summary>` is the toggle, so the
chevron, the name, the count and the teaser are all pressable. "Full history"
goes in `action` — *outside* the summary, since a control inside a control both
fails `nested-interactive` and would fold the section it was meant to leave.

**One derivation for the headline.** The teaser reads the newest exchange
through `activityHeadline`, now exported from the list that draws the rows.
Two spellings of "what this exchange is called" is how a teaser comes to
promise a row the reader then cannot find under it.

### A latent bug in `Disclosure.action`

`action` had **no caller in the tree** until this one, and the first use found
the defect: the verb was absolutely positioned, so it reserved no width and a
summary wide enough simply ran underneath it. Reproduced at 900px — the
teaser's date sat behind the "Full history" button.

The wrap is a grid now: the `<details>` takes every column the verb does not.
That keeps what the old comment was protecting — the whole line except the
verb's own box is still the toggle — while making a long teaser ellipsise
*before* it reaches the verb instead of sliding under it. The CSS comment
explaining the old choice is rewritten rather than left asserting something
that is no longer true.

## How verified

- `make check-fe`: green, exit 0 — **557/557 files**.
- Playwright, full suite against the built preview: **274 passed, exit 0.**
- Two unit tests pin the behaviour through the real screen: the fold is closed
  on arrival and its name carries the count (including a subjectless call,
  which is what that test was always about), and the teaser shows the **newest**
  exchange while the section stays folded.
- Shot at 1440px and 900px, with a 360 carrying two activities. At 900px the
  name holds one line, the teaser ellipsises, and the verb keeps its own
  column; measured `clipped: true` on the teaser and no overlap.
- The old copy keys `co.360.readThread` / `co.360.hideThread` are removed in all
  three locales, replaced by `co.360.thread` / `co.360.threadCount`.

- [x] `make check` is green
- [x] `make test-integration` is green (or: this change does not touch tenant data / RLS / the write shape) — frontend only

## AI involvement

Written by Claude Code at Jo's request: fold it by default because it takes too
much space, but tease it rather than showing a bare button, and make the whole
row the control. The `Disclosure.action` overlap was not part of the ask — it
surfaced on the first narrow-viewport screenshot and is fixed here rather than
left for the next caller.

## Accountability

off (`git commit -s`, DCO).

---

---
## Summary by cubic
The account 360's "What happened" thread now arrives folded and teases its newest exchange on the row that opens it, so the 360's lower half stays free until a reader asks for the thread.

- The whole row is the toggle: section name, count, and teaser are all pressable, with "Full history" in the `Disclosure` `action` slot outside the summary.
- The teaser reads through `activityHeadline`, now exported from the recent list, so the folded row and the opened rows use one wording per exchange.
- `Disclosure.action` no longer runs under long summaries — the wrap is a grid, so the action keeps its own column and the summary ellipsises before reaching it.
- i18n keys `co.360.readThread` and `co.360.hideThread` are replaced with `co.360.thread` and `co.360.threadCount` in all three locales.

<sup>Written for commit d6687add98654b72e7d5fdeeddf7e1624207a3ce. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/4217?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Company activity threads now appear folded by default, showing the activity label, message count, and a teaser of the newest exchange.
  - Select the summary row to expand the full history.
  - Long activity summaries and thread actions now remain clearly separated without overlapping.

- **Improvements**
  - Activity headings now use the activity subject when available, with a localized fallback when not.
  - Updated English, German, and Vietnamese labels provide clearer thread and message-count wording.


joshbruh-lang:
On the `worklist.walk` singular/plural finding: **verified, real, and not this PR's.**

`noticeText` in `frontend/src/screens/worklist.walknotice.tsx` calls
`worklist.walk.arrived` whenever `arrived > 0`, so `arrived === 1` renders
"1 weitere seit Ihrem Start. Sie warten auf eine Aktualisierung" — wrong number
in German — and the English arm has the same defect one word later ("1 more
since you started. **They** wait for a refresh"). `worklist.walk.both` inflects
two counts and has no arms either. So it is not a German-only wording slip: all
three locales carry it, and the catalog already has the machinery for it —
`_one` / `_other` arms picked through `Intl.PluralRules` in `format/plural.ts`,
which `i18n.test.ts` and `one-plural-rule.test.ts` both hold.

It is on `main` and untouched by this branch. It entered the review's window
only because the merge commit re-presents `main`'s diff, `f965dfff` having
landed while this PR was open. Fixing it here would mean three locales and a
worklist screen riding into a change about the account 360's thread fold, so I
have left it alone rather than widen the PR.

Worth its own change, and small: two counts, two keys with arms, one caller
that stops formatting the number before the catalog can inflect on it.

---

PR #4229, opened 2026-09-04
A quiet contact can be set aside, and a brief item answered where it is ranked

LarsGradion:
Two verbs the server already offered and the row never drew, plus the defects a review pass found in them.

## Set aside a quiet contact

`relationship_decay` rows kept coming back every morning with no way to say "not this one, not now". The 30-day nudge-dismissal endpoints shipped without a caller; the row now calls them. `actionDismiss` is offered on the decay lane, whose row id **is** the person's id, which is exactly what `/people/{id}/nudge-dismissal` takes.

Undoable from the confirmation toast, like every disposition beside it — the row leaves the lane on success, so a misclick otherwise costs the reader the only address they had.

## Answer a brief item from the queue

The row drew only Pin while the server sent three verbs, because none of `act`/`set_aside`/`dismiss` was in `VERB_DESTINATION`. Reuses `useBriefItemMark` from `home.queries`, which already invalidated the worklist.

The double-submit guard is `isPending || isSuccess`, not `isPending`: an answered brief item is patched **in place** (unlike a completed task, which leaves), so the row stays on screen between the write settling and the refetch. Two POSTs were observed before the fix.

## The five defects found by reverting one clause at a time

1. **`BriefVerbs` drew all three verbs unconditionally.** The lane sends all three today, so no test could tell a client reading `actions` from one assuming the full set. The day one is withheld, the assuming client keeps drawing it and posts an answer the server did not authorise.
2. **The refusal handler read `mark.error`** — the state React last rendered, still null on a first failure. A rep whose colleague had answered the item in another tab was told "no cause reported", and the retry that invites hits the same 409.
3. **`dismiss` was added to the lane-wiring gate's global performable set.** The endpoint depends on the source: a decay row posts to the person's nudge dismissal, a brief item to the brief's own mark. A lane-blind entry passes the gate for a lane the client has no route for, and the advertised verb is simply not drawn — the failure the gate exists to catch. Now keyed by lane.
4. **A refused undo said nothing.** The dismissal removes the row, so by the time Undo is pressed the component is unmounted and React Query has dropped the observer per-call callbacks hang off. `mutateAsync` + `catch` instead.
5. **The test for #4 proved nothing.** The default five-minute `gcTime` kept the dismissal's observer alive for the length of the test, so a per-call `onError` still fired where production had long dropped it. `mutations: { gcTime: 0 }` makes it model a reader who takes seconds to press Undo.

## Evidence

| Obligation | Evidence |
|---|---|
| User-visible outcome | `worklist.dismiss.test.tsx` "is set aside for a month, and can be put back" — PUT `{days:30}`, then DELETE on undo |
| Server authority respected | "draws only the verbs the row was given" — a row sent `["act"]` draws Done alone |
| Failure behavior | "names why an answer was refused" (409 detail reaches the toast); "says so when the undo is refused after the row has gone" |
| Idempotency | "submits once however fast the reader presses" — guard is `isPending \|\| isSuccess` |
| Contract | `TestNoOptionalLaneOffersAnActionTheSurfaceCannotPerform` — lane-scoped, so a verb without a client route still fails |
| Mutation strength | All five clauses reverted individually; each fails its own named test and only that test |
| Full lane | `make check-backend` green (4m22s); `make check-fe` green (2m53s) |

---
## Summary by cubic
Gives the worklist two verbs it never drew: a decay row can now set a quiet contact aside for a month, and a brief row can be answered in place with `act`, `set_aside`, or `dismiss`. Also fixes five defects a review pass found while landing them.

**New Features**
- The decay lane's row now offers "Not now", which PUTs a 30-day nudge dismissal and can be undone from the confirmation toast.
- Brief rows now draw the `act`, `set_aside`, and `dismiss` verbs the server sends, each posted to the brief's own endpoint via the existing `useBriefItemMark`.

**Bug Fixes**
- Brief verbs are drawn only when the server's `actions` list includes them, so a withheld verb is never posted.
- Refused answers now name the server's cause from the mutation callback instead of the stale `mark.error`.
- The verb-wiring gate now scopes `dismiss` to its lane, since its endpoint depends on the row source.
- A refused undo now says so via `mutateAsync` + `catch`, because the row unmounts before Undo is pressed.
- The double-submit guard now checks `isPending || isSuccess`, since an answered brief item stays on screen until the refetch.

<sup>Written for commit a472f0eade4d342f363a4fd02455e20eb7a343af. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/4229?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


LarsGradion:
Superseded by #4230, which is the same feature over the same eleven files and is now on main.

#4230 additionally carries five fixes from an adversarial review pass, each held by a test that fails when the fix is reverted:

1. A refused undo said nothing — the dismissal removes the row, so the component owning the Undo button is unmounted before it can be pressed and a per-call `onError` is dropped with that component's observer. It uses `mutateAsync` and a catch, which is what `TaskComplete` beside it already does and explains.
2. `BriefVerbs` ignored the server's `actions` and drew all three verbs regardless.
3. The refusal toast read `mark.error`, the last-rendered state, still null on a first failure — so a 409 reported "no cause reported".
4. The lane gate treated `dismiss" as globally performable although its endpoint depends on the source.
5. The double-submit test's stub answered instantly, so `isSuccess` alone satisfied it and the `isPending` half of the guard could be deleted with everything still green.

Closing rather than merging: the two branches are the same work, and this one would conflict across all eleven files without adding anything.

PR #4243, opened 2026-09-04
WIP: A row says at most three reasons

LarsGradion:
## Why

The `because` line had no bound. Every reason the server sent was drawn, and the line wrapped freely — so the block is not a fixed cost. It grows by a line every time the reason vocabulary does, and the vocabulary is still growing.

That matters because the row is already too tall. Measured at 390×844 while investigating the skipped criteria in #4238, an **ordinary task row is 284px against the 176px ceiling** `AC-WORKLIST-SDR-07` asserts:

| Part | Height |
|---|---|
| rank | 19 |
| title (wraps to two lines at 390px) | 49 |
| when · **because** · consequence | **57** |
| dispositions | 44 |
| verbs (Pin) | 44 |
| gaps + padding | ~71 |

57px of that is this block carrying **only two reasons**. Without a bound, a reason added anywhere makes every row that carries it taller, and the person adding it has no way to know they did.

## The decisions

**The cap is a ceiling the row owes its height, not a judgement about which reasons matter.** The server already ranked them — `crm.yaml` says `because` arrives *"in the order they were weighed"* — so the first three are the three that put the row where it is. Taking a prefix keeps the server's ranking rather than inventing a second one in the client.

**Three rather than two**: a task that is due today *and* unassigned *and* overdue is saying three true things a rep acts on differently, and two would drop one of them on a row that fits today.

**Applied after the unspeakable reasons are dropped.** Capping the raw list first would let a row carrying three reasons this build has no sentence for say nothing at all, while the two it could have said sat just past the cut. That ordering has its own test.

## What this does not do

**It does not make AC-07 pass, and is not meant to.** The row measured there carries two reasons, so a cap of three changes nothing about it. Both geometry criteria stay `fixme`.

What it does is stop the ceiling being missed again by a change nobody connected to it. The decision about what a 176px row actually holds is still open, and still a product call.

## Evidence

| Obligation | Evidence |
|---|---|
| The cap holds | `says at most three reasons, keeping the ones the server weighed first` — asserts the first three appear and the fourth, a phrase this build **knows**, does not |
| Ordering is right | `still says the reasons it knows when unknown ones come first` — three unknown reasons ahead of two known ones; both known ones still appear |
| Mutation strength | Two clauses reverted one at a time, each observed failing its own test: removing `.slice(0, MOST_REASONS_SHOWN)`, and moving it before the `.filter` instead of after |
| Fixture is not vacuous | The four copy keys are distinct strings (`overdue`, `due today`, `nobody owns it`, `routine tidying`), so the absence assertion cannot pass by accidental substring match |
| Full lane | `make check-fe` exits 0 |

---
## Summary by cubic
Limits the worklist row's `because` line to at most three reasons so the row stops growing taller every time the reason vocabulary expands.

- Takes the first three reasons in the server's weighted order, preserving the ranking instead of re-sorting in the client.
- Capping runs after unknown phrases are dropped, so a row with three unspeakable reasons still shows the two known ones past the cut.
- Does not make `AC-WORKLIST-SDR-07` pass: that row carries two reasons, so the cap changes nothing there.

<sup>Written for commit 98bb593ff63d7b762610383d571b1be2d1ebba9c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/4243?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/screens/worklist.row.tsx
+++ b/frontend/src/screens/worklist.row.tsx
@@ -114,9 +114,15 @@ export function WorklistRow({
       : item.detail;
   // The badged reasons are drawn as badges above and left out here, so one
   // meeting does not report the same finding twice in two registers.
+  //
+  // CAPPED, and capped AFTER the unspeakable ones are dropped. Capping the
+  // reasons before `reasonText` runs would let a row carrying three reasons
+  // this build has no sentence for show nothing at all, while the two it could
+  // have said sat just past the cut.
   const because = phrasedReasons(item)
     .map((reason) => reasonText(reason, t, locale, zone))
     .filter((phrase): phrase is string => phrase !== null)
+    .slice(0, MOST_REASONS_SHOWN)
     .join(" · ");
   const above = comparisonText(item.above_next, t, locale, zone);
   const consequence = consequenceText(item, t);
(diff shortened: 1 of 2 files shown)


LarsGradion:
Closing this. The premise was wrong and the fix discards information the reader needs.

**The measurement I built it on was misread.** I attributed 57px to the `because` block. Measured properly at 390px, that 57px is three SEPARATE elements — `when`, `because` and `consequence` — at 19px each. The reason line itself is one line. Three reasons still fit on one line; the fourth costs 18px and the sixth another 19px. So the growth is real but small: capping at three saves at most 37px against a 108px overflow. It does not do what the commit message claimed.

**A blind prefix drops load-bearing facts.** An adversarial pass found five, and the pattern is one thing: reasons that matter most are appended LAST, because they are absorbed or applied late.

- `pinned` is appended by `applyPins` (pins.go:81) after everything else, and it drives the row to the top. The comment there names the exact harm — without it "the one row a reader definitely knows the reason for is the one the page cannot account for."
- `expected_revenue` is absorbed onto a waiting row last (classifydecay.go:138) and is a ranking comparator (ranksteps.go:152). The row is lifted above its neighbours by money and would then not say why. `decaydedupe_test.go:155` exists specifically to keep that fact visible.
- An absorbed at-risk deal appends all three of its grounds (`material`, `quiet_days`, `closing_soon`) after the waiting reasons, so a cap of three drops two of them.
- `classifyWaiting` can emit five reasons in evaluation order, so `asks_nothing` — a verdict that changed the row's level — falls off.

I tried a `DECIDING_REASONS` allowlist to protect the important ones. That was worse: it is a hand-written importance ranking in the client, and the rank steps do not map onto reason kinds (`band`, `crowded` and `occurrence` have no reason at all). It is a second answer to a question the server already answers, which is the thing this repo forbids outright.

**The tree already has an answer and it is not discarding.** `dealstatus.tsx:227` shows the first reason and puts the rest behind a disclosure — reachable, not dropped. A worklist row has no pane to put a remainder in, and building one is the row-anatomy decision that AC-WORKLIST-SDR-07 is already waiting on.

So this belongs with that decision rather than ahead of it. The measurements are recorded in the ledger and in #4238's comments; nothing is lost by closing.

PR #4270, opened 2026-09-05
A self-only proposal is matched as one member's, not the workspace's

### LINKED ISSUE #3758: vcard_create's identity-keyed supersession and decline memory are not subject-scoped
Surfaced during review of margince#3410 (the vCard org-less staging fix): `vcard_create` is a self-only approval kind — `decidable` (`internal/modules/approvals/authority.go`) restricts reads and decisions to `a.OnBehalfOf == p.UserID` — but the identity-keyed paths in the shared staging engine carry no equivalent subject filter:

- `lockPendingUnderIdentity`/`supersedePendingInTx` (`internal/modules/approvals/staging.go`) match and expire a pending row by `kind + target_entity_id + proposed_change @> identity` alone.
- `declinedProbeSQL(byIdentity=true)` (`internal/modules/approvals/stagingidentity.go`) matches prior decisions by the same containment, also with no subject filter.

For most `vcard_create` cards this is masked because the identity includes `organization` and usually `emails`, both of which are genuinely a fact about the CARD rather than about a shared name alone. But `internal/compose/vcardcreateproposal.go`'s `vcardCreateStager` now (post margince#3410) declines to build an identity at all for a card with neither an email nor an organization — deliberately, because a bare full name is not this card's own addressing, it is every card sharing that name. That narrows the reachable case to "two members independently upload a name-only card for two different people who happen to share a name," which the fix above defuses for that one caller.

The underlying engine gap is broader: any other identity-bearing `selfOnlyKind` (LinkedIn match match is the other current example) could produce the same cross-subject interference if two members' identities ever collide, and nothing in the engine stops it — it trusts each caller to build a collision-proof identity, which is a property no caller currently states or tests.

**The call to make**: scope `lockPendingUnderIdentity` and `declinedProbeSQL` by `on_behalf_of IS NOT DISTINCT FROM` the acting subject for every kind in `selfOnlyKinds`, so a self-only kind's identity-keyed paths agree with its own read/decide gate about whose proposal a row is. This is a shared staging-engine change affecting every self-only caller, not a single-caller fix, so it wants its own review rather than folding into a caller's PR.

dungbanguyen:
Closes #3758, to the ruling recorded on it.

## The gap

`decidable` says a self-only proposal belongs to ONE member: only they may read it and only they may decide it. The shared staging engine's matching statements did not know that — they matched on kind, target and either the logical identity or the diff hash.

So for exactly the kinds whose own gate says a row is one person's business, two members proposing the same thing:

- **joined** onto one row (`stageOrJoinPendingInTx`'s diff-hash step — the sibling [tinnhlt-gra found](https://github.com/margince/margince/issues/3758#issuecomment-5544117127), same root cause),
- **expired** each other's pending proposals (`lockPendingUnderIdentity`),
- **inherited** each other's declines (`declinedProbeSQL`, both discriminators).

Each is silent from the member's side: their card never appears in their queue and nothing says why.

## Why not leave it to the callers

The engine trusts each caller to build a collision-proof identity. **No caller states that property and none tests it**, and #3410 shows how subtle it is — a bare full name *"is not this card's own addressing, it is every card sharing that name"*, which took a review to notice. An unstated, untested property held by every future caller is a hope rather than a boundary.

## The shape

One predicate, `subjectScopedShape`, asked of the staging: it is the same question `decidable` asks of a stored row — a self-only **kind**, or a staged create against a table whose rows belong to one human each. **Derived, not listed at each statement**, so a kind added to `selfOnlyKinds` or a table enrolled in `probeOwnerOnly` inherits the narrowing rather than quietly staying shared.

`subjectScope` appends the subject to the argument slice and derives `$N` from where it landed, per the rulebook — a hand-typed placeholder is a number nothing checks.

**`IS NOT DISTINCT FROM`, never `=`.** `on_behalf_of` is nullable, and `=` against NULL is NULL — so an `=` here would match nothing at all for those rows and the memory would forget a refusal rather than hold it. `TestASubjectlessProposalStillRemembersItsOwnDecline` is what holds that, and it is the mutation `=` fails.

## The narrowing stops there

`TestASharedKindStillMatchesAcrossMembers` is the assertion protecting this from over-reaching: a shared kind still matches across members, which is the inbox working as designed — a manager triages what a rep staged. Without it the fix could have been "scope everything" with nothing to say the inbox had stopped being shared.

## Verification

`make check-backend` green; `make test-it DIR=backend/internal/modules/approvals RUN=…` green on the five new tests. Mutation-checked: the scope never applying, the scope applying to every kind, `selfOnlyKinds` no longer consulted, and `=` for `IS NOT DISTINCT FROM` — each fails with the test that names it, and the shared-kind assertion fails in the opposite direction.

## Two things about main, not about this change

1. **`make craft-static` is red on main** (`championcoverseam_test.go`'s naked `any`, from #4242) — every open PR fails the craftsmanship lane on a file it never touched. Fixed separately in #4269.
2. **`TestAnApprovalTheAgentNeverRedeemedIsMarkedRatherThanLeftSilent` fails when the approvals integration package is run whole, and passes when run alone** — reproduced on a pristine `origin/main` checkout with no local changes, so it is not this change. `MarkLapsedRedemptions` counts 2 where the test expects 1, which reads as cross-test interference in the sweep's own scope. Filed rather than fixed here: it is a different subject and would double this diff.

### DIFF
--- a/backend/internal/modules/approvals/selfonlysubject_integration_test.go
+++ b/backend/internal/modules/approvals/selfonlysubject_integration_test.go
@@ -1,0 +1,241 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+//go:build integration
+
+package approvals
+
+// A self-only kind's identity-keyed paths, over a real Postgres.
+//
+// decidable says a self-only proposal belongs to ONE member: only they may read
+// it and only they may decide it. The staging engine's matching statements did
+// not know that. They matched on kind, target and either the logical identity
+// or the diff hash — so two members proposing the same thing landed on ONE row,
+// expired each other's pending proposals, and inherited each other's declines.
+//
+// These are Postgres tests because the boundary IS the predicate: the clause
+// lives in SQL, and a Go double would be asserting against its own idea of what
+// the statement matches.
+
+import (
+	"context"
+	"testing"
+
+	"github.com/margince/margince/backend/internal/shared/kernel/ids"
+	"github.com/margince/margince/backend/internal/shared/kernel/principal"
+)
+
+// forMember is the connector principal that stages a card ON BEHALF OF one
+// member — the shape every self-only kind is staged in. A human staging
+// directly carries no OnBehalfOf, and a self-only row without one is decidable
+// by nobody (decidable fails closed), so it is not the shape under test.
+func (e *stagingEnv) forMember(member ids.UUID) context.Context {
+	ctx := principal.WithWorkspaceID(context.Background(), e.ws)
+	ctx = principal.WithCorrelationID(ctx, ids.NewV7())
+	return principal.WithActor(ctx, principal.Principal{
+		Type: principal.PrincipalConnector, ID: "connector:vcard",
+		UserID: member, OnBehalfOf: member,
+		Permissions: principal.Permissions{RoleKeys: []string{"admin"}},
+	})
+}
+
+// secondMember seeds another seat in the same workspace.
+func (e *stagingEnv) secondMember(t *testing.T) ids.UUID {
+	t.Helper()
+	id := ids.NewV7()
+	if _, err := e.owner.Exec(context.Background(),
+		`INSERT INTO app_user (id, email, display_name) VALUES ($1, $2, 'Second')`,
+		id, "second-"+id.String()+"@st.test"); err != nil {
+		t.Fatal(err)
+	}
+	return id
+}
+
+// nameOnlyCard is the card the ticket's reachable case is about: two members
+// upload a card for two DIFFERENT people who happen to share a name, so the
+// logical identity collides while the subjects do not.
+func nameOnlyCard() StageInput {
+	const sharedName = "Jan Mueller"
+	return StageInput{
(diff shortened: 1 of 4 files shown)


dungbanguyen:
Correction to the second note in the description.

I wrote that `TestAnApprovalTheAgentNeverRedeemedIsMarkedRatherThanLeftSilent` fails on pristine main. It does — **locally, when the approvals integration package is run whole** — but **CI's integration lane is green on it**, which I checked on #4269 (pristine main plus one line: all six shards and the RLS/e2e lane pass). So main is not red on this, and the note should not have implied it was.

What is actually true: the test passes alone and fails when the package runs whole, in a local `make test-it DIR=backend/internal/modules/approvals` run. `approval` carries no workspace column — ADR-0091 collapsed the installation to one workspace — so `MarkLapsedRedemptions` sweeps the table, which is right in production and means `setupStaging` minting a fresh workspace id per test does not isolate anything: rows from earlier tests in the same run are still in the sweep's reach.

Filed as #4271. Not this change either way — reproduced with no local modifications.

dungbanguyen:
The issue is #4273 — the number in the comment above is wrong.

PR #4348, opened 2026-09-05
fix(agents): the tool conformance lane invokes the analytics query it gained

dungbanguyen:
**main is red.** #4334 registered `run_analytics_query` without giving the tool-conformance sweep a call for it, and the census that exists to notice exactly that noticed:

```
toolschema_conformance_integration_test.go:298: run_analytics_query is registered but never
invoked here, so nothing holds its result to the schema or the envelope. Add a call above,
or name the seam it needs in unreachableInThisLane.
```

Reproduced on a pristine `origin/main` checkout before changing anything, so this is main's and not a PR's.

**Why it went unnoticed.** The post-merge run for #4334 failed; the next commit (#4336) was frontend-only, so the backend lane never re-ran and `main`'s latest run reads green. Every PR opened since inherits the failure through its merge commit — two of mine did, which is how I found it.

**Invoked, not waived.** `unreachableInThisLane` is for tools whose seam this lane does not stand up — a live calendar, an outbound mail provider, a drafting model. The analytics seam IS up here: `compose_analytics_report` and `run_report` already run over the same `deals-by-stage` population. A waiver would have excused a tool the sweep can reach, which is the failure mode that list is written to avoid.

Unsaved on purpose: `save` is what mints a citable `run_id`, and what this lane holds is the answer SHAPE — columns and rows arrive either way, and their empty-not-null slices are exactly what a schema alone cannot hold.

## Verification

Before, on `origin/main`:
```
--- FAIL: TestToolAnswersReachableWithoutApprovalSatisfyTheirSchemas (0.57s)
```
After:
```
--- PASS: .../run_analytics_query_{"entity":"deals-by-stage","measures":[{"fn":"count"}]} (0.00s)
ok  github.com/margince/margince/backend/internal/compose/integration
```

No open PR was fixing this.


dungbanguyen:
Closing in favour of #4368, which fixes the same break. We opened these within the hour and neither of us saw the other's first.

Theirs is the one to keep: it invokes the tool over `pipeline-current`, the population this fixture already holds an open deal in, so the answer is a row rather than an unproven empty page. Mine used `deals-by-stage` — also non-empty here, but `pipeline-current` is the better fixture for the reason their PR body gives.

Same diagnosis on both sides, independently: reproduced on a pristine `origin/main` checkout, and invoked rather than waived because the analytics seam is up in this lane, so a waiver would excuse a tool the sweep can reach.

PR #4371, opened 2026-09-05
fix(rollup): the won-deal fixture writes the base amount that stopped being generated

dungbanguyen:
**main is red**, and this is the deterministic half of it.

#4327 stopped `deal.amount_minor_base` being a `GENERATED` column — a generated expression cannot reach either currency's minor-unit scale — and made it one the close writer fills. That PR updated the sibling fixture in `organization_computed_integration_test.go` (it deleted `freezeDealFX`) and missed this one: `seedRollupWonDeal` inserts a won deal with raw SQL, so its frozen figure is now NULL and `SUM(d.amount_minor_base)` reads zero.

```
orgrollup_integration_test.go:206: tree closed-won = 0, want 30000
orgrollup_integration_test.go:372: closed-won = 0, want 5000 (in-quarter deal at its frozen 0.5 rate only)
```

Reproduced on a **pristine `origin/main` checkout** (`45d506d93`) before changing anything, so this is main's and not a PR's. Every open PR inherits it through its merge commit — five of mine did, which is how I found it.

## The fixture states the figure, it does not compute it

`baseMinor` is a new argument rather than arithmetic inside the helper. Repeating the conversion here would be a second implementation of the thing under test, and the sums would then agree with the fixture rather than with the product — the same reason the two USD cases already state their rate instead of deriving one.

- EUR against a EUR base: `30_000 → 30_000`, identity.
- USD at the frozen `0.5`: `10_000 → 5_000`, which is exactly what the assertion three lines down already claims.

## The other failures in those shards are not this

`passports`, `seatmatrix` and `aiactivity_e2e` also reported failures in shards 3 and 4. All pass in isolation here — they are the parallel lane's own interference ("admin sees 1 passports, want 2", "no actor bound to context"), not a code defect. Only the rollup one reproduces deterministically.

## Verification

`make test-it DIR=backend/internal/compose/integration RUN='TestOrgRollup'` — all green.

No open PR was fixing this. (The other main break — `run_analytics_query` missing from the conformance sweep — is #4368; I closed my duplicate #4348 in its favour.)

### DIFF
--- a/backend/internal/compose/integration/orgrollup_integration_test.go
+++ b/backend/internal/compose/integration/orgrollup_integration_test.go
@@ -69,13 +69,25 @@ func seedRollupOpenDeal(t *testing.T, e *Env, st rollupStages, org ids.UUID, amo
 
 // seedRollupWonDeal closes a deal with the frozen FX rate the
 // deal_closed_fx CHECK demands — the rate the quarter sum must reuse.
+//
+// baseMinor is the FROZEN converted amount, stated by the caller rather than
+// computed here. It used to need no argument at all: amount_minor_base was a
+// GENERATED column, and inserting the amount and the rate was enough to produce
+// it. 1788583500 stopped generating it — a generated expression cannot reach
+// either currency's minor-unit scale — so it is now a column the close writer
+// fills, and a fixture that inserts a won deal directly has to fill it too.
+//
+// Stated and not derived, deliberately. Repeating the conversion arithmetic
+// here would be a second implementation of the thing under test, and the sums
+// below would then agree with this fixture rather than with the product. What
+// the caller writes is the figure it expects the quarter sum to find.
 func seedRollupWonDeal(t *testing.T, e *Env, st rollupStages, org ids.UUID,
-	amountMinor int64, currency, fxRateToBase string, closedAt time.Time,
+	amountMinor, baseMinor int64, currency, fxRateToBase string, closedAt time.Time,
 ) {
 	t.Helper()
-	e.WsExec(t, `INSERT INTO deal (id, name, amount_minor, currency, fx_rate_to_base, pipeline_id, stage_id, organization_id, status, closed_at, source, captured_by)
-		VALUES ($1, 'Won Deal', $2, $3, $4, $5, $6, $7, 'won', $8, 'manual', 'human:test')`,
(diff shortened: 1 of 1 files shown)


tinnhlt-gra:
Found and fixed this same defect independently while working on margince#4368 (dropped my duplicate commit there in favor of this PR, which was already open and green). One correction worth flagging though: the passports failure ("admin sees 1 passports, want 2") is **not** parallel-lane interference — it reproduces deterministically in isolation (3/3 runs via `make test-it`). Root cause: 8996cef8e ("A settings surface is gated on the grant it named, not on being an admin") moved `ListPassports`'s admin widening off `id.hasRole("admin")` (reading `Identity.Roles`) onto `auth.Require(ctx, "user_admin", ActionRead)` (reading `Identity.Permissions`), and never updated `passports_integration_test.go`'s `identityFor`, which still only sets `Roles`. Fixed in margince#4368 alongside the `run_analytics_query` conformance gap. Might be worth a second look at `seatmatrix`/`aiactivity_e2e` in isolation too, in case they're the same class of gap rather than genuine flakes — I haven't checked those two myself.

tinnhlt-gra:
This landed via margince#4368 (merged as 5290d8419), which included the same `amount_minor_base` fixture fix alongside two other unrelated main-red fixes. Should be a no-op diff now if you rebase onto main — probably safe to close unless you want to keep it as a narrower, single-purpose reference.

dungbanguyen:
Closing in favour of #4368, which landed the same fix an hour after this went up. Confirmed green on `origin/main` `5290d8419`.

Theirs is the better one and worth saying why: it computes the frozen figure with `deals.ConvertToBase` — the exported arithmetic the real close writer itself calls — where this stated the expected number as a fixture argument. Both close the hole; only one cannot drift from production when the conversion changes again.

Same diagnosis independently: reproduced on a pristine `origin/main` checkout, #4327 stopped `amount_minor_base` being generated, updated the sibling fixture in `organization_computed_integration_test.go` and missed this one.

Also worth recording for whoever reads this thread later: `passports`, `seatmatrix` and `aiactivity_e2e` were failing in the same shards throughout and are NOT part of it — all three pass in isolation, and their symptoms ("admin sees 1 passports, want 2", "no actor bound to context") are the parallel lane interfering with itself. Only the rollup pair reproduced deterministically.

PR #4373, opened 2026-09-05
fix(compose): the published MCP surface names the vocabulary a client is served

luit-nfq:
## Why

`docs/reference/mcp-info.json` is the served MCP surface as a client receives it, and it is stale on `main` — `TestPublishedMCPSurfaceMatchesWhatAClientIsServed` fails on a clean checkout of `main` with no changes at all:

```
mcpinfo_test.go:159: ../../../docs/reference/mcp-info.json is stale —
it no longer matches what a client is served.
```

The analytics vocabulary's schema version reads `246ea65901f0ba2b` where a client is now served `940be79ffe683865`.

## What moved, and why the page looked unchanged

`schemaVersion` hashes every entity's field **names, SQL expressions and kinds** (`analyticsqueryseam.go`), while the page renders only the names grouped by kind. So an expression changing moves the version and leaves the rendered list byte-identical — which is exactly this diff: one hash, nothing else.

The cause is #4327, *a conversion applies both currencies' minor-unit scales*, which rewrote the expressions behind `amount_base_minor` and `weighted_base_minor`. That is the half of the hash the page cannot show, so nothing in review looked wrong.

## Why it matters

A compiled plan names the version it was planned against and is **refused** once that version moves — deliberately, so a plan naming a since-renamed field cannot render SQL against a column that is gone. A client reading this page to pre-plan a query would have every plan refused, and the page it read is what told it the wrong number.

## How

Regenerated through the artifact's own switch, never hand-edited:

```
go test ./internal/compose/ -run TestPublishedMCPSurface -update-mcp-info
```

`make check` is green on this branch.

## Scope

This is one of two things red on `main` right now. The other — the schema-conformance sweep not invoking `run_analytics_query` — is #4368, and this PR deliberately leaves it alone: they are separate defects with separate causes, and neither fixes the other.

---
## Summary by cubic
Regenerates the published MCP surface so the analytics schema version in `docs/reference/mcp-info.json` matches what a client is actually served, fixing a stale value on `main`.

- The version hashes field names, SQL expressions, and kinds, but the rendered page shows only names, so the prior currency-scale change (#4327) moved the version without changing the visible list.
- A compiled plan is refused once the version it names moves, so a client planning from the stale page would have every plan refused.
- The file was regenerated via `go test ./internal/compose/ -run TestPublishedMCPSurface -update-mcp-info`, not hand-edited.
- Leaves the other failing check on `main` (the schema-conformance sweep, #4368) alone; it is a separate defect.

<sup>Written for commit b57e59711c56e18fa2a59ead9adf82dbb0b658d9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/4373?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Documentation**
  - Updated the published analytics schema document version.
  - Analytics vocabulary and query definitions remain unchanged.


tinnhlt-gra:
This looks like a duplicate — margince#4356 already fixed this exact defect (regenerated `docs/reference/mcp-info.json` via `go test ./internal/compose/ -run TestPublishedMCPSurface -update-mcp-info`) and merged into main at `a4600eb6`. Worth checking whether this is now redundant, since main should already carry the fix.

luit-nfq:
Superseded by #4368, which grew to cover this too — `docs/reference/mcp-info.json` is regenerated on main and `TestPublishedMCPSurfaceMatchesWhatAClientIsServed` passes there. Nothing left in this branch: rebasing it onto main skips the commit as already applied.

PR #4379, opened 2026-09-05
Name the next step, and make the button write it

joshbruh-lang:
## What

An account with an open deal and nothing scheduled raised advice titled **"Set the next step"**, with an **"Add the next step"** button that did nothing at all — the company page had no surface to route `add_task` to, so pressing it left the account exactly as it was.

After this PR:

- The rule **names the step** in the words it will be written in — `Agree the next step on "Fleet retrofit 2026"` where exactly one deal is open, or the account's own name where several are and picking one would be a guess.
- The action carries the **prepared `POST /tasks` body**, and the button writes it. The client sends that body unchanged.
- The recommendation also appears on the record's **Tasks tab**, above the open work, drawn as the agent's rather than a person's. An archived account draws none.

## Why

Two failures, and the second is the one that costs the surface its credibility.

**The button was the control-that-does-nothing this surface is built to avoid.** `Organization360Suggestion.action` already promised that a rule which cannot name an action carries `null` so the card advises without a button — but `add_task` named an action the client then dropped on the floor. A reader who presses a button and sees nothing happen learns to stop pressing them, and that lesson does not stay local to one row.

**Advice that restates the finding is not advice.** The rule had already worked out that nothing is scheduled *and* which deals that is true of. Handing that back as "set the next step" spends the reader's attention and returns their own problem.

The invariant this establishes: **the sentence a rep accepts and the task they get are one sentence.** The step is composed once, server-side, and travels as the exact body the write takes. Composing a task out of the row's words client-side would put a second author on it, and the two would drift the first time either side was reworded; it would also give the client a say in what the task is linked to, which the rule already decided from records.

Three things it deliberately does **not** do:

- **No model.** The step is derived from the same open pipeline the reason sentence is read off, under the same row-scope predicates. Every rule in this file stays one a rep could make themselves.
- **Nothing staged, no invented deadline.** The rule fires on an absence, and an absence has no date; `due_at` stays unset for the same reason the suggestion's own does. Nothing exists until a rep presses the button, and the write goes through the same governed `POST /tasks` the task form uses — audited and undoable.
- **No invented work.** It names *agreeing* the next step, never the work itself. "Prep the expansion workshop" is the row the mockups drew and the contract already names as the thing the system has no basis for.

The title and the prepared subject are outside the fingerprint (PO-AC-N-13/14), so this reword retires no dismissal — the existing ones hold.

### Layout

`suggestions.go` crossed the 500-line cap, so the no-next-step rule and the step it prepares move to `internal/compose/org360/nextstep.go`. It is the one rule in the file that answers with work rather than with a reading, which is what the split is along.

## How verified

**CI is the authority here, and it has now run the two lanes this container could not.** An earlier revision of this description hedged on both; CI answered them, and the notes below are what it actually reported rather than what I expected.

- **Integration: green** — all 6 shards, `integration (RLS + erasure + HTTP e2e)`, and `integration unit coverage`. This is the lane I could not run locally (no Docker), so the checkbox below is CI's verdict, not mine.
- **Frontend: green** — `fe-unit`, `fe-quality` (ds gates + contract drift + biome + composed tsc), `fe-bundle` (vite build + storybook, which is what registers the new story).
- **Also green:** `live-boot`, `govulncheck`, `craft-residue`, `secret-scan`, `extension-reference`.
- **`deterministic-gates` caught one real defect of mine** and is fixed in `ed86e07`: `taskBody` has to name the anonymous struct oapi-codegen emits for `CreateTaskRequest.Links` in order to build one, and that struct's `EntityId` reads to staticcheck as an ST1003 naming mistake rather than as the mirror it is. Waived in source with the reason, identical in form and wording to the waiver `newSuggestionAction` has carried three functions up since before this branch. The `gates` package itself passed (277s) in the same job.

Locally: `go vet ./...`, `make arch-lint`, `craft static --strict` on every changed Go file (0 blocker / 0 major / 0 minor), `check-rls-store-path.sh`, `check-no-jurisdiction.sh`, `check-changelog-sections.sh`, and the `org360` unit tests. Contract regenerated through the `gen` targets and `pnpm gen:api` — after the merge from `main` as well as before, so no generated file here is a textual auto-merge.

New tests, each shown failing before the fix:

- `TestTheAskIsTheStepTheButtonWrites` — the title and the prepared subject are the same string, and the body carries no date.
- `TestTheStepHangsOnTheDealOnlyWhenThereIsOne` — including the case where the count outruns the names.
- `TestAnAccountStepCarriesTheAccountsName`, and the nameless-record fallback.
- `companytasks.test.tsx` — the Tasks tab posts the server's body byte-for-byte, and an archived account offers nothing.
- `company360.test.tsx` — an `add_task` carrying no body still draws no button.

**Two local failures that were NOT this change**, both since confirmed environmental by CI passing the same gates: the pinned golangci-lint refused to load against this container's newer Go toolchain, and `TestBaselineEraFixtureIsTheMatrixTheBaselineSeeded` needs full history that a shallow clone does not have.

A third is worth recording because it nearly cost a wrong push: `frontend/src/screens/home.readings.test.tsx` fails 2 tests in that container and fails identically on clean `main`, because `Intl.NumberFormat` with `compact` + `currency` + `maximumFractionDigits: 1` resolves `minimumFractionDigits` to 1 there and renders `€420.0k` where the test expects `420k`. CI runs Node 24 and `fe-unit` is green, so it is the local ICU and nothing is owed — but a fix pushed on that reading would have been a regression.

Not driven manually — no dev stack here. The Tasks-tab pairing has a new Storybook story (`NextStepRecommended`); `fe-bundle` confirms it registers. It is a state no seeded demo account can reach: every one of them already carries an open task, which is exactly the condition that silences the rule.

- [x] `make check` is green
- [x] `make test-integration` is green — via CI's integration lane; not run locally (no Docker)

## AI involvement

Written end to end by Claude Code (Opus 5) in one session: the contract change, the backend rule and its split, the frontend wiring, all tests, the story, the changelog entry, the merge from `main`, and the lint fix. No part of this diff was hand-written.

Three things worth a reviewer's attention:

- The first version of the Tasks-tab test left a React render queued past jsdom teardown, which surfaced as an unhandled error in an **unrelated** test file 500 tests later. It was caught by running the full suite against a clean tree for comparison, and fixed by waiting on the 360 re-read the mutation invalidates. Worth knowing that failure mode exists in this suite — `frontend/AGENTS.md` describes the neighbouring family.
- `taskBody`'s first doc comment said "spelled once", which tripped the uniqueness-claim gate correctly: the claim had no test holding it. The comment was reworded rather than the gate waived.
- The merge from `main` had one real conflict — `main` deleted `RecentActivityPanel` from `company360.stories.tsx` while this branch added `ProposedNextSteps` to the same import list. Every generated artifact was regenerated from the merged contracts rather than left as the line-merge git produced.

## Accountability

https://claude.ai/code/session_0166pfUGWcidks6Bja2HRPW8

---
_Generated by [Claude Code](https://claude.ai/code/session_0166pfUGWcidks6Bja2HRPW8)_


joshbruh-lang:
### `uat` is red here, and it is not this PR's

Standing down on it, with the reasoning so it can be checked rather than taken on trust.

**What is failing:** `uat (AC screens + axe WCAG 2.2 AA + 390px, mocked)` — 14 of 331 specs in `e2e/ac.spec.ts`. Every one is on `#/home`, `#/deals`, `#/analytics`, the command palette, the agent's panel, or the app shell (`nav.rail` never renders). The two signatures are `#/home` failing to render the shell at all, and `scrollable-region-focusable` on `section[data-stage="s3"|"s4"|"s5"]` — the deals board's empty stage columns.

**Why it is not this PR's**, on two independent grounds:

1. **It is already filed against `main`.** #4396 — *"UAT lane red on main: #/home crashes under the seed mock, #/deals fails axe on empty stage columns"* — was opened automatically by `main-health` at 10:04 today, ~20 minutes before this run, with that exact signature. #4397 covers the fourteenth failure (`nothing scrolls sideways at 390px on #/analytics`). Both are open.
2. **No failing screen can reach this diff.** The frontend half of this PR touches `company360.tsx`, `organizations.tsx`, three i18n files (one added key) and a stories file. Nothing under `home.tsx`, `deals.tsx`, `analytics.tsx`, `app/shell.tsx` or `app/agentrail.tsx` imports from `company360.tsx` — the importers are `assistant.tsx`, `companyfinance.tsx`, `companytoday.tsx`, `compose.tsx`, `organizations.tsx` and test/story files, none of which any of the 14 specs exercises.

**No fix exists to port.** #4396 and #4397 are both open with no PR referencing either, so there is nothing to bring in here that would go green. Widening this PR to fix the home seed mock and the deals board's axe violations would be fixing two unrelated defects under a title about the next-step suggestion.

**Not re-run.** The rule for a re-run is to confirm a failure that looks like it is not this PR's — that is already settled by #4396 being open against `main` for the same lane, which is the stronger form of the same evidence. These are deterministic axe violations and a render crash, not a flake, so a second run would consume a runner to reprint the same 14 lines.

`uat` is a coordinator lane and is not a required check (`frontend/AGENTS.md` states this). Every required check on this head is green: `ci`, `deterministic-gates`, `craftsmanship`, the frontend fan-in (`fe-unit`, `fe-quality`, `fe-bundle`), all six integration shards plus the RLS/erasure/HTTP e2e fan-in, `live-boot`, `govulncheck`, `extension-reference`, `secret-scan`, and SonarCloud's quality gate (0 new issues, 94.0% coverage on new code).

Worth one note for whoever picks up #4396: `main`'s frontend lane is red on its own tip right now (`fe-quality` and `fe-bundle` both fail at `079ae26`, filed as #4404), which is why `uat (main)` is currently *skipped* there and cannot produce a same-lane comparison. This PR's `fe-quality` and `fe-bundle` both pass, because its base predates that breakage.

---

joshbruh-lang:
### Ported #4498's forecasting fix so this PR can go green

`integration / integration shard (3/6)` failed on `./internal/modules/forecasting` once this branch merged `main`. That is the failure #4498 describes and is not this PR's: `testdb.AssertPoolsQuiesced` is registered as a per-test cleanup asserting the **shared** pool is empty, so under `t.Parallel()` the first sibling to finish reports the others' still-legitimate connections as a leak — naming whichever test happened to end first. #4498's own words: *"main is red, deterministically, on every PR whose merge commit reaches this package."* This branch's merge is what reached it.

Rather than wait for #4498 to land, its change is **ported here unchanged** (`23b6da9`): the nine tests that take the shared pool drop `t.Parallel()`, and `setupSnapshot` carries the reason so the next author does not put it back. The periods those cases work in were always what kept them independent, so running them in sequence costs nothing they relied on. It no-ops the moment `main` carries #4498.

Verified locally: `go vet -tags integration ./internal/modules/forecasting/` clean, `gofmt` clean, `craft static --strict` on both files — 0 findings.

### Also on this head

The `company360.tsx` conflict with #4448 is resolved (`5073b00`). `main` added a `t-caption` typography class to the dropped-count line while this branch changed that line's condition and value; the two are orthogonal, so both are kept. Typecheck clean, the new `type-one-spelling.test.ts` gate passes 12/12, and the four affected suites pass 169/169.

Worth recording for anyone who hits it: **that conflict is also why CI appeared stuck.** The workflow runs against `refs/pull/4379/merge`, which GitHub cannot compute while a PR is conflicted, so `ci` silently never started on the previous head — no run to re-trigger, and re-triggering would not have helped. Resolving the conflict is what started it.

---

PR #4387, opened 2026-09-05
Show a mailbox import on the orb, the lit edge and a top-centre chip

joshbruh-lang:
## What

A connect-time mailbox import now shows up everywhere the agent shows up. While a backfill is queued or running, the rail's Core goes to `ingest`, the lit window edge lights, the line under the orb reads "Importing mail history · 42%", and the import's share is drawn as the ring around the orb. A small chip floats top-centre of the content column for as long as mail is arriving — a compact Core with the same ring, the server's own counts, and a link to the connections card (absent on that card, which already draws the run). The connections read polls at 2.5s while an import is live and not at all otherwise, so the orb, the chip and the settings card all draw from one body.

## Why

After connecting a mailbox the product said "your mailbox is now capturing" and nothing in the chrome showed it: the agent's status surfaces derive from the AI-activity feed, and the feed learns of capture only as per-message classifications the router announces once each call is over. A two-hour import therefore reached the chrome as a trickle of settled lines and never as work in flight — the one state named for the agent taking something in was the one the longest intake could not cause.

The invariant: the chrome reports the agent's standing state, and every reading is the server's own. The import is read off `CaptureConnection.backfill` (the persisted-row counts the settings card already draws), never a client-side counter; without a preview there is no denominator, so no ring is drawn rather than a guessed one; a scan that outgrew its preview clamps to a full ring; and the ring goes under a fault, because a red orb wearing a progress ring would say the work is going well and badly at once.

One named exception: the chip draws a second Core, which the shell's own comments call a one-Core rule. It is drawn for one RUN, wears that run's ring, and exists only while the run does; the reason is stated at the component.

Scraping is covered from the other side: #4392 (now on `main` and merged into this branch) makes a company website read report itself to the feed as a `site_read` occurrence, so a crawl lights the orb and the edge through the rail's existing path. This PR adds nothing for it and needs nothing from it; the import is the one long run the feed still cannot carry.

## How verified

- `fe-typecheck` (`tsc -b`), `fe-lint` (biome), the design-system gates (`check-ds-spacing.sh`, `check-space-tokens.sh`, `check-ds-purity.sh`, `native-controls`, `catalog`, `type`, `corners`, `tokens`, `conformance`) and the Storybook build are green.
- Full `fe-unit` in CI is green on the head.
- New suites: `capture-progress.test.ts` (derivation: null without a live run, share from persisted counts, no fraction without a denominator, clamp, two mailboxes folded into one reading, poll cadence), `agentrail.capture.test.tsx` (orb state, ring, line, edge, ring kept under a feed-named run, ring dropped under a fault, nothing reported once settled), `capture-chip.test.tsx` (share and counts, count-only without an estimate, absent at rest, absent on the connections card).
- Stories: `Shell/Agent rail` gains `ImportingMail` and `ImportingMailUnpreviewed`; `Shell/Capture chip` covers importing, queued, unpreviewed and almost-done.
- Copy added to all three catalogs (`en`, `de`, `vi`); the i18n parity suite passes.
- `origin/main` merged in after #4401 (the `uat` lane fixes) and #4392 landed; the merged tree passes the same gates.

- [x] `make check` is green — the frontend half in full as above; no Go files change, so the backend half is unaffected
- [x] `make craft-static` reports no new blockers — no Go files in the diff

## AI involvement

The whole change was AI-generated in a Claude Code session from the product owner's description of the gap, and reviewed against the rail's existing doctrine (`agentrail.tsx`, `ai-activity-orb.ts`) and the frontend rulebook.

## Accountability

https://claude.ai/code/session_01Uo4JEvQcJemsi4osSv5qpD


joshbruh-lang:
**`uat (AC screens + axe WCAG 2.2 AA + 390px, mocked)` is red on this head, and the failure is not this PR's.**

The 13 failing cases in `e2e/ac.spec.ts` (`nav.rail` not found on `#/home` at 390px and in dark, the palette and agent-panel cases at 390px, the `#/analytics` 22px overflow, and the three `scrollable-region-focusable` axe findings on `#/deals`' stage sections) are the same set that failed on [#4378](https://github.com/margince/margince/pull/4378)'s `uat` run, which merged into `main` immediately before this branch's base (`5290d84`). That run had 14 failures; this one has 13, all of them in that set. None of the screens named (home, analytics, deals stages, the palette, the account menu) is touched by this diff, and the e2e seed carries no live backfill, so the capture chip never renders in that lane.

No open PR carries a fix yet, so there is nothing to port. I do not have the means to re-run the job from here; the lane is not a required check and the `ci` aggregate is green on this head. I will keep watching and bring in the fix when one lands on `main`.

---

joshbruh-lang:
**`frontend / fe-unit` went red on the merge commit `32e204a`, and the failure is `main`'s.**

The one failing case, `App.test.tsx › mounts the field builder on the Data model page`, fails identically on `origin/main` and on [#4419](https://github.com/margince/margince/pull/4419)'s own fe-unit run. #4419 moved the settings route constants out of the settings screen module, which made the settings screen a real lazy chunk; this test is now its first loader, and under vitest the chunk's first transform outlasts the finder's budget, so the test expires on a page still compiling.

Fix opened as [#4425](https://github.com/margince/margince/pull/4425) (warm the chunk before rendering, budget unchanged) and ported into this branch as `e55397d`; it no-ops once `main` carries it.

---

joshbruh-lang:
**`frontend / fe-unit` is red on `a08224e`, and the failure is not this PR's.**

Every test passed (7206 of 7206). The job failed on one unhandled error vitest caught after the run: `ReferenceError: window is not defined` from react-dom's scheduler (`performWorkUntilDeadline` on a Node `Immediate`), attributed to `src/screens/worklist.email.test.tsx` — a scheduler tick landing after that file's jsdom environment was torn down. That file renders the worklist screen alone and reaches nothing this diff touches; it passes three consecutive local runs on this head; and the same `main` base passed fe-unit on [#4425](https://github.com/margince/margince/pull/4425) and on this PR's previous head `e55397d`.

I have no means to re-run the job from here, and there is no robust fix within this PR's scope: the race is between the last test's settled `waitFor` and work React still has queued at file teardown, in a test file this PR does not own. The next push that carries a real change (a `main` merge, once `main` moves) will re-run the lane; I will keep watching until then.

---

PR #4384, opened 2026-09-05
Two sections of a record are two sections

joshbruh-lang:
## What

Record pages drew pairs of panels edge to edge. The account's Deals tab, its Partner tab, the overview's partial-read notice, and the seam between any tab's body and the record's chronology all had **no interval at all** between two boxes — which reads as one card with a rule through it rather than as two sections. Each takes the record's own step now. The project page, which ran both its columns at a third rhythm, reads at the record's steps like every other record.

## Why

A record's work column carries no interval of its own — `.arrive-stack` on it is the arrival, not a rhythm — so every tab body owes its panels a stack. Most take one; the ones that did not had nothing between them.

The invariant, in the three places it now lives:

- **A tab body that draws more than one panel wraps them in `.record-stack`.** Stated in the class's own comment and in the design-system catalog, where the next author meets it.
- **`.wrap`'s no-two-boxes-touching rule reaches the work column.** That rule's own comment already said any wrapper that comes to stand in the page's stack belongs in its list; the record's work column is one and was never added. It is the FLOOR under a body that has not taken the stack, not a second answer to how a record reads.
- **The columns' rhythms belong to `RecordView`, which draws them.** `.record-rail` / `.record-aside` were spelled twice, in the account's stylesheet and the contact's (one as a raw `12px`, one as the token), for classes the shell itself applies — so a record page importing neither sheet got no rail rhythm, and the project page had written a third answer beside them. `.record-timeline` is new and owns the break between a tab's bodies and the record's story, which was nobody's.

Every tab of all six record pages was read for this; the settings and AI-settings tabs were checked too and were already correct (`.settings-stack`).

| Surface | Before |
|---|---|
| Company → Deals | contract panel flush on the deals panel |
| Company → Partner | three cards edge to edge |
| Company → Overview, failed read | the partial-read notice flush on the first pane |
| Deal → Overview | a hand-set 16px where the record steps 24 |
| Deal and Project → body ↔ chronology | 0px |
| Project → both columns | 16px, against 24 for the work column and 12 for rails |

One record page keeps the defect and is not fixed here: `screens/people.tsx`'s `PersonScreen` is unrouted dead code (its own story says so), so the floor rule covers it rather than restructuring something nothing reaches.

## How verified

Measured rather than eyeballed — the gap between the work column's children, in a headless browser against the built catalog:

```
project page  before   DIV.record-stack  gapToNext=0    SECTION
deal page     before   DIV.record-stack  gapToNext=0    SECTION
[...]

Issue #4396, opened 2026-09-05
UAT lane red on main: #/home crashes under the seed mock, #/deals fails axe on empty stage columns


joshbruh-lang:
## What fails

The advisory `uat (AC screens + axe WCAG 2.2 AA + 390px, mocked)` lane fails 13 tests in `frontend/e2e/ac.spec.ts` on `main` at 5290d84 and every frontend PR based on it. Reproduced locally with `make frontend-e2e` against the unmodified tree; the failures are deterministic, not flaky.

Because the lane is advisory it blocks no merge today, which is exactly why it matters: every frontend PR's UAT verdict is red for a reason that is not the PR's, so a real regression in the same lane would read as more of the same.

### 1. `#/home` renders the error boundary under the seed mock

Every test that starts on home fails on `expectShellRendered` (`nav.rail` not found): the one-h1 census, both home axe sweeps (light and dark), the 390px overflow check on home and on analytics, the palette sheet, both agent-panel tests, and the account-menu test.

Page error: `TypeError: Cannot read properties of undefined (reading 'trim')` in `minorUnitDigits` (`frontend/src/format/minorunits.ts`).

Cause: since #4363 the home readings strip (`frontend/src/screens/home.readings.tsx`) calls `useForecastReadings`, which reads `GET /forecast`. `frontend/e2e/seed.ts` has no fixture for `/forecast`, so the list catch-all answers it with `{data, page}`, `data.base_currency` is undefined, and `formatMoneyCompact` throws mid-render. The same catch-all trap is already documented in `seed.ts` for `/admin/job-health` and `/analytics/context`.

Fix: a shaped `/forecast` fixture in `seed.ts` ahead of the catch-all. Required fields per the `/forecast` response in `frontend/src/api/schema.d.ts`: `period_start`, `period_end`, `scope_kind`, `won_minor`, `evidence_minor`, `best_case_minor`, `open_minor`, `weighted_minor`, `eligible_count`, `priced_count`, `confirmed_date_count`, `fx_missing_count`, `as_of`, `timezone`, `base_currency`.

### 2. `#/deals` fails axe with `scrollable-region-focusable`

Both palettes report three violations on `section[data-stage="s3"]`, `"s4"` and `"s5"`: the empty stage columns rendered by the board in `frontend/src/design-system/composed.tsx` are scroll containers with no focusable content. A column with nothing in it should not be a scroll region, or should carry focus only when it actually overflows.

## Where it was found

[...]

PR #4405, opened 2026-09-05
fix(frontend): the forecast mock, and the lint the lanes still fail on

dungbanguyen:
**Scope has shrunk since this opened, and deliberately.** Main is being repaired by several people at once, and three of the four causes this PR started with have landed through other PRs while it sat. Rebased onto current main and everything already fixed elsewhere is dropped — including two of my own changes that main now carries better versions of.

What is left is what main is still red on.

## 1. `/forecast` is not mocked, and the fallback answers the wrong shape

The morning strip's pipeline slot reads `/forecast`. `e2e/seed.ts` has no entry for it and its fallback answers a **paged envelope**, which this route's schema is not — so the card received `{data: [], page: {…}}`, which is truthy, and went on to format money against a body with no `base_currency`.

The card's own guard for that landed separately (it is on main now, and reads a currency-less body as an unanswered question). This is the other half: the harness should not hand a screen a body the server cannot produce. Without it the mock still lies about this route, and the next reader of that card debugs a shape that only exists in the tests.

## 2. `fe-lint` is red

`noUselessFragments` in `leads.tsx`, `personpage.tsx` and `analytics.test.tsx`, and an unused import in `person-record.spec.ts`. All machine-fixable, none introduced here.

## Dropped from this PR, because main fixed them first

- **The empty stage column's tab stop.** Main takes `tabIndex={0}` on every column with a `biome-ignore`; mine was conditional on the column being empty and uncollapsible. Theirs is simpler and cannot drift as a column's contents change — I took it and deleted my helper rather than push a competing spelling.
- **`recordtabs` `min-inline-size: 0`.** Main has it with a better comment than mine (it names the grid case as well as the flex one). My copy became a duplicate-property lint error; removed.
- **The `home.readings` currency guard.** Already on main.
- **`project360.tsx`.** #4384 rewrote the lines my lint autofix touched; took theirs.

## Verification

- `pnpm exec playwright test e2e/ac.spec.ts` — **195 passed, 1 skipped, 0 failed**.
- `make check-fe` green.

Original diagnosis of all four causes, for the record: the `#/home` crash was captured with a `pageerror` listener against the built preview (`TypeError: Cannot read properties of undefined (reading 'trim')` in `minorUnitDigits → toMajor → formatMoneyCompact → PipelineOutlook`), and eleven of the fifteen uat failures were the app's error boundary standing where the rail should be — none of them naming `#/home`.

## Summary by CodeRabbit

* **Refactor**
  * Simplified internal rendering and test syntax without changing user-visible behavior.
  * Removed redundant UI wrappers from lead and person record screens.
* **Tests**
  * Updated end-to-end and analytics test code while preserving existing assertions and matching behavior.
  * Added mock forecast data for an owner-scoped reporting period.


dungbanguyen:
Superseded and closing unmerged.

https://github.com/margince/margince/pull/4401 landed an independent fix for the same main-red episode while this sat unrebased, and it covers the ground this covered. Verified against a clean `origin/main` worktree at `85e324eb5`:

```
$ pnpm exec playwright test e2e/ac.spec.ts
  1 skipped
  195 passed (1.8m)
```

The uat lane is green on `main` with none of this branch applied, so there is nothing left here to land. Every check on this PR was red for a reason that had also already been fixed: all four failing jobs traced to the same three `TS2739` errors in `src/screens/personreadings.test.tsx`, inherited through the merge commit from a `main` that https://github.com/margince/margince/pull/4407 has since repaired.

One judgement in this branch was better than the one that landed, and I am recording it here rather than reopening it as a diff: `#4401` gives **every** stage column `tabIndex={0}`, where this branch gave one only to a column that is empty and not collapsible (`ownTabStop`). Both clear `scrollable-region-focusable`; the landed one puts five extra tab stops in front of a reader on a full board, since a column with cards is already reachable through them and a collapsible one through its head. Not worth a PR on its own — I have left it as https://github.com/margince/margince/issues/4396's closing note and it can ride along with the next change to that board.

https://github.com/margince/margince/issues/4396 is closed with the verification.

PR #4412, opened 2026-09-05
A compact money figure gets the unit test that would have caught it — merge after #4414

joshbruh-lang:
## What

Adds the unit test `formatMoneyCompact` never had. Four cases in `frontend/src/format/format.test.ts`, nothing else in the diff.

**This must land after #4414 and not before.** #4414 fixes the helper; this only pins the fix. On its own it is red for anyone whose ICU forces the decimal.

### History of this PR

It opened as a change to two assertions in `home.readings.test.tsx`, on the premise that `formatMoneyCompact` deterministically renders `€420.0k` and the test was wrong to expect `€420k`. That premise was false, and this PR's own CI disproved it: the runner rendered `€420k` while this container rendered `€420.0k`, off the same commit. The assertion change went red on CI for exactly the reason it was meant to fix.

The real defect was in the formatter, and #4414 — opened independently, 24 minutes after this one — has it, with a better account of the mechanism and a stronger fix for the second half. So the formatter change and the per-file `cleanup()` that were here have been dropped rather than duplicated:

- **The formatter.** #4414's `minimumFractionDigits: 0` is the same line this carried, and its comment is the better one: it names the clamp direction and catches a case missed here — in German, where 420k does not abbreviate at all, the figure rendered `420.000,0 €`.
- **The cleanup.** #4414 registers Testing Library's `cleanup()` globally in `vitest.setup.ts`, with a behavioural gate, so no file can forget. This PR fixed the one file that had — strictly the smaller fix.

What is left is the one thing neither #4414 nor `main` has.

## Why

`maximumFractionDigits: 1` with no minimum leaves the minimum to Intl, whose default under `style: "currency"` is the currency's own digit count — two for EUR — and ECMA-402 only clamps that DOWN to the maximum. So the option reads as "at most one digit" and resolved to *exactly* one on some ICU builds and to none on others. One stored figure, two renderings, decided by the reader's browser.

Nothing caught it because nothing tested the helper. The screen tests that did touch the figure asserted `/420k|420,000/`, which cannot tell the three spellings apart: it passes on `€420k`, on `€420.0k`, and on the uncompacted `€420,000.00` — three different claims to a reader, one match to the regex. That is why these assertions are exact strings.

The four cases, and what each would catch:

- A round figure drops the decimal — the divergence itself.
- A remainder keeps it: `€186.4k` and `€186.9k` are the same number on the page if the maximum ever goes to zero.
- Under the ten-thousand threshold the figure stays exact (`€8,332`), so the compact form is never used where it buys nothing.
- The abbreviation rides the **server's** ISO scale, not Intl's CLDR count: VND carries no minor unit, so 420,000 minor units is `₫420k`. Dividing by Intl's count instead would draw `₫4.2k`, a figure the record does not hold.

## How verified

- `pnpm exec vitest run src/format/format.test.ts` — 39 passed, with #4414's `format.ts` applied locally to stand in for the merged state.
- Mutation-checked in both directions, which is the point of a gate: without the fix the round-figure case fails `expected '€420.0k' to be '€420k'`; with it, green. Removing the fix from a tree carrying only these tests reproduces it.
- `pnpm exec biome check` clean on the one changed file.

CI on this branch alone will go green *for the wrong reason* — the runner's ICU already defaults the minimum to zero, so the gate passes there whether or not the helper is fixed. That is precisely why it should not merge ahead of #4414: landing first would put a test in `main` that is red on every newer ICU.

- [x] `make check` is green — `fe-unit` green (576 files); the backend lanes are untouched by a one-file frontend test change
- [x] `make test-integration` is green (or: this change does not touch tenant data / RLS / the write shape) — test-only frontend change
- [x] `make craft-static` reports no new blockers — `craft static` sweeps the Go trees; this diff is TypeScript

## AI involvement

Agent-written throughout: the original wrong diagnosis, the CI evidence that overturned it, the four test cases and their comments, the mutation check, and this description. The decision to slim this PR to the gate rather than duplicate #4414 was the human's.

## Accountability

https://claude.ai/code/session_01CWXrbtjmkFfmHhZnw84j7F

### DIFF
--- a/frontend/src/format/format.test.ts
+++ b/frontend/src/format/format.test.ts
@@ -8,6 +8,7 @@ import {
   formatDayMonth,
   formatDuration,
   formatMoney,
+  formatMoneyCompact,
   formatMoneyOrAbsent,
   formatNumber,
   formatSignedMoney,
@@ -79,6 +80,43 @@ describe("money formatting (B-EP09.17)", () => {
 
   it("still scales two-decimal currencies under vi", () => {
     expect(formatMoney(128_400, "EUR", "vi")).toBe("1.284,00\u00a0€");
+  });
+});
+
+// The KPI-slot formatter had no test of its own, and that is how it came to
+// render one stored figure two ways. `maximumFractionDigits: 1` with no minimum
+// leaves the minimum to Intl, whose default under `style: "currency"` is the
+// currency's own digit count — so a round €420,000 kept a forced ".0" on one
+// ICU build and dropped it on another. One figure, two renderings, decided by
+// the reader's browser rather than by this file.
+//
+// EXACT strings for that reason. `toContain("420")` is satisfied by both
+// spellings AND by the uncompacted "€420,000.00", which is the assertion shape
+// that let the divergence run: it could not tell the three apart.
+describe("a money figure abbreviated for a KPI slot", () => {
+  it("drops a decimal a round figure does not need", () => {
+    expect(formatMoneyCompact(42_000_000, "EUR", "en")).toBe("€420k");
+  });
+
+  // The maximum earns its keep here: rounded to whole thousands these two
+  // figures are the same number on the page.
+  it("keeps the decimal that tells two figures apart", () => {
+    expect(formatMoneyCompact(18_640_000, "EUR", "en")).toBe("€186.4k");
+    expect(formatMoneyCompact(18_690_000, "EUR", "en")).toBe("€186.9k");
+  });
+
+  // Below the threshold the long form is no wider and carries every digit, so
+  // abbreviating buys nothing — and the currency's own decimals go with it,
+  // because "€8,332.00" is what the slot has no room for.
+  it("stays exact under ten thousand", () => {
+    expect(formatMoneyCompact(833_200, "EUR", "en")).toBe("€8,332");
+  });
+
+  // The scale is the SERVER's ISO table, not Intl's CLDR one. VND carries no
+  // minor unit, so 420,000 minor units is ₫420k; dividing by Intl's count
+  // instead would draw ₫4.2k, a figure the record does not hold.
+  it("abbreviates on the stored scale, not Intl's own count", () => {
+    expect(formatMoneyCompact(420_000, "VND", "en")).toBe("₫420k");
   });
 });
 


joshbruh-lang:
`uat` is red on this head, and it is not this PR's. Recording why, and why I am not porting the fix.

**The failure.** One test of 276:

```
e2e/ac.spec.ts:1170 › §3.8: 390px mobile › nothing scrolls sideways at 390px on #/analytics
  + "the shell content scroller (.scroll): 107px past the viewport"
```

**Why it is not this PR's.** The diff here is one file, `frontend/src/format/format.test.ts`, and it adds four assertions about a formatter. It touches no component, no CSS and no route, so it cannot move a layout. The same test fails identically on #4414, whose diff is unrelated to both this one and to analytics — two different diffs, one shared failure, which puts it on the base.

**The fix that exists.** #4405 carries it: `.recordtabs { min-inline-size: 0 }`, so a tab strip in a flex header scrolls its own tabs instead of widening the page. That PR names this exact sweep as what caught it.

**Why it is not ported here.** #4405 measured eleven pixels past the viewport; this run is 107. Same test and plausibly the same cause, but not the same magnitude, so I cannot say that change alone would turn this lane green — and porting a fix I have not seen work would be a guess in a PR whose whole diff is four assertions. `uat` is also a coordinator lane rather than a required one, and the `ci` aggregate is green on this head, so this does not block the merge.

No re-run: this is deterministic rather than flaky, red on two unrelated diffs.

Worth noting for whoever picks the overflow up — the gap between 11px and 107px suggests something landed on `#/analytics` after #4405 was written, so that PR may no longer be the whole fix.

---

joshbruh-lang:
Folded into #4414 and closing this.

`9bb688a3` is now cherry-picked onto #4414's branch unchanged — same four cases,
same commit message, same author trailers. Nothing was rewritten.

**Why they had to be one PR rather than two ordered ones.** #4414's SonarCloud
was failing on *0.0% Coverage on New Code (required ≥ 80%)*: its new code is the
`format.ts` line, and the only test exercising that line was here, behind a
"merge after #4414" gate. So the fix could not satisfy the coverage gate while
its own test waited downstream of it — the split was self-blocking rather than
merely sequential.

Folded, #4414 now carries the fix and the gate that proves it, which is also the
shape the repo asks for: the change and the test that would have caught it land
together.

Verified on the merged branch (`9f85918`, with current `main` in):

- `format.test.ts`, `home.readings.test.tsx` and `scripts/autocleanup.test.ts` — 58/58.
- Full unit suite — **577 files, 6685 tests, 0 failed.**

Nothing is lost by closing this; every line of it is in #4414.

---

PR #4414, opened 2026-09-05
fix(frontend): a compact money figure carries no digit it has nothing to put in

joshbruh-lang:
## What

Two things, both in the frontend unit lane.

`formatMoneyCompact` now renders `€420k` where it rendered `€420.0k`, and in German `420.000 €` where it rendered `420.000,0 €`. A compact money figure keeps a fraction digit only when there is a remainder to put in it — `€185.5k` is unchanged.

Testing Library's cleanup is now armed for the whole suite, so a render no longer outlives the case that made it.

## Why

**The formatter.** `formatMoneyCompact` passed `maximumFractionDigits: 1` to a `style: "currency"` formatter and nothing else. For the currency style `minimumFractionDigits` starts at the currency's own minor-unit digits — two for EUR — and ECMA-402 only clamps that *down* to the maximum. So `min(2, 1) = 1`: the option reads as "at most one digit" and resolves to **exactly** one. The function's own doc comment states the intended output (`€428k`), so the implementation contradicted its contract, and every round compact money figure in the product carried a dead `.0`. German is the worse case — 420k does not abbreviate there at all, so the digit landed on a full number.

This is what made `src/screens/home.readings.test.tsx` red. Two cases wait on `/420k/` and `/420k|420,000/`; the figure was on the page the whole time, spelled `€420.0k`, which neither pattern can match. Both waits died at their full 1000ms `asyncUtilTimeout` — the signature `frontend/AGENTS.md` names under *"A wait that expires is usually waiting for something that never arrives"*, read the way that section says to read it. It is not load-dependent: the file failed **5 of 5** runs, three of them alone on an idle machine with identical timings.

**The cleanup.** Testing Library registers its own `afterEach(cleanup)` only when the runner has put `afterEach` on the global; `test.globals` is not set in `vite.config.ts`, so RTL's guard saw nothing and never fired. Nothing reports an unregistered hook. A render therefore outlived the case that made it — React tree still mounted, query observers still subscribed — so a case queried a document holding every tree before it, and `notifyManager` went on flushing renders into trees no test owned, past the file's own teardown. That is the reported `@tanstack/query-core` unhandled error after a test had finished.

Most files here call `cleanup()` themselves, which is a habit rather than a guarantee. The files that forgot leaked whole screens: `worklist.email.test.tsx` ended with **5 mounted containers**, four entire `WorklistScreen` trees. Registered once in `vitest.setup.ts` so a file cannot forget, imported dynamically inside the existing DOM guard so node-environment suites do not pay for react-dom.

The invariant is gated behaviourally rather than by asserting a hook exists: `scripts/autocleanup.test.ts` renders in one case and asks the next what is left in the document, because what breaks a suite is what a page-wide query can still find. Written without JSX so it is a `.ts` and falls inside `tsconfig.node.json`'s `scripts/**/*.ts` — a `.tsx` there would run in the suite and be typechecked by nothing.

No timeout was raised, and no test was skipped, disabled or quarantined.

## How verified

- `pnpm exec vitest run src/screens/home.readings.test.tsx` — **17/17**, where it was 2 failed / 15 passed on 5 consecutive prior runs. The file's test time falls from **2287ms to 313ms**: the two cases no longer sit out a full waiter budget.
- The formatter change checked across currencies and both locales: `€420.0k → €420k`, `US$1.0m → US$1m`, `420.000,0 € → 420.000 €`, and `€185.5k`, `€9,999`, `JP¥420k` unchanged. No test in the tree pins the `.0` form, and the `185.5k` / `186.4k` assertions in `personreadings` and `company360` are unaffected.
- `worklist.email.test.tsx` leaked containers measured with a probe: **5 → 0**, file still 4/4.
- `scripts/autocleanup.test.ts` proven fast-red by disarming the registration in `vitest.setup.ts`: it fails with `expected 'the case before this one rendered me' to be ''`. A gate that cannot fail would be worth nothing here.
- `pnpm lint` (Biome, 1729 files) clean — the remaining warnings are pre-existing and in files this diff does not touch. `tsc -p tsconfig.app.json` and `tsc -p tsconfig.node.json` both clean.
- Baseline recorded before the change: full `vitest run` on unmodified `main` (a58d0b8), `Test Files 1 failed | 575 passed (576)`, `Tests 2 failed | 6676 passed (6678)`, ~410s on 4 cores. Two full-suite runs, same file, same two cases both times — so on this commit the lane is not intermittent, it is deterministically red.

Full-suite runs on the fixed tree are still finishing as I open this; I will report their verdict on this PR and push anything they turn up.

- [ ] `make check` is green — not run: this container has no database or Go toolchain lane available, so the backend half could not be exercised. The frontend half (lint, both typechecks, the unit suite) is covered above.
- [x] `make test-integration` is green (or: this change does not touch tenant data / RLS / the write shape) — this change touches neither.
- [x] `make craft-static` reports no new blockers — `craft static` sweeps the Go trees only, and this diff has no Go.

## AI involvement

Entirely AI-assisted: the reproduction, the DOM capture at failure, the ECMA-402 diagnosis, both fixes, the gate and this description. The measurements quoted above were all run in this session rather than reasoned about.

## Accountability

https://claude.ai/code/session_011P2P9rBzVpxjJwRf13HvhT

---
_Generated by [Claude Code](https://claude.ai/code/session_011P2P9rBzVpxjJwRf13HvhT)_

---
## Summary by cubic
Fixes `formatMoneyCompact` so round compact figures no longer render a trailing `.0`, arms Testing Library cleanup so test renders don't outlive their cases, and keeps SonarCloud from counting the test harness as uncovered product.

**Bug Fixes**
- `formatMoneyCompact` now renders `€420k` instead of `€420.0k` (and `420.000 €` instead of `420.000,0 €` in German); a fraction digit appears only when there's a remainder, so `€185.5k` is unchanged.
- Added `minimumFractionDigits: 0`; the currency style defaults it to the currency's minor-unit digits, and ECMA-402 only clamps it down to `maximumFractionDigits`.
- `cleanup` from `@testing-library/react` now runs after every test via `vitest.setup.ts`, preventing leaked React trees and observer subscriptions from affecting later tests.
- Adds exact-string tests in `format.test.ts` and a behavioral cleanup guard in `scripts/autocleanup.test.ts`.
- Excludes `vitest.setup.ts` from Sonar coverage; the v8 provider never reports the harness, so the new-code gate read 0% once it gained a line.

<sup>Written for commit ea6cdda40868404ca9ba8d0695d9f8dcdb5f0478. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/4414?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


joshbruh-lang:
The full-suite runs the description said were still finishing. All green, and the last one answers the question this PR started from — whether the lane is sensitive to how busy the machine is.

**Before, on unmodified `main` (a58d0b8), 4 cores:**

| run | verdict |
|---|---|
| full suite | `Test Files 1 failed \| 575 passed (576)`, `Tests 2 failed \| 6676 passed (6678)`, 411s |
| full suite | same file, same two cases |
| `home.readings.test.tsx` alone ×3, idle machine | `2 failed \| 15 passed (17)` each time, durations within 14ms of each other |

Five for five, the same two cases, whether the machine was busy or idle. That is what settled the diagnosis: the file was never intermittent, so load was never the variable, and raising a budget would have been fixing the wrong thing.

**After, on `499006e`:**

| run | verdict |
|---|---|
| full suite ×3 | `Test Files 577 passed (577)`, `Tests 6680 passed (6680)` — 407s, 403s, 399s |
| full suite under 4 spin loops on 4 cores | `577 passed (577)`, `6680 passed (6680)`, 699s |

577 and 6680 are the previous totals plus `scripts/autocleanup.test.ts` and its two cases. **Zero unhandled errors in any of the four runs** — grepped for, not assumed; the `notifyManager` error that prompted this had no occurrence left to find.

The contended run is the one worth reading. Deliberate contention stretched the suite 1.75× (399s → 699s), which is the regime the original reports blamed, and nothing failed. The added `afterEach(cleanup)` costs nothing measurable on an idle box either — 403s median against a 411s baseline.

CI on this head agrees: `fe-unit`, `fe-quality`, `fe-bundle`, `live-boot`, `secret-scan`, `changes` and `craft-residue` are all green. The Go gates are skipped because the diff has no Go.

One limit worth stating plainly rather than leaving to inference: `make check` was not run end to end, because the environment this was developed in has no database or Go lane. CI has since covered the frontend half and `live-boot`; the backend unit and integration lanes are skipped by the `changes` router rather than passed. The unticked box in the description says the same thing.

---

joshbruh-lang:
## #4412 folded in, and `main` merged

Two changes on `9f85918`, neither touching the fix itself.

**#4412's test commit is cherry-picked here** (`9bb688a3`, unchanged). It was
opened as "merge after #4414", but the split was self-blocking rather than
merely ordered: SonarCloud failed this PR on *0.0% Coverage on New Code
(required ≥ 80%)*, and the only test covering the `format.ts` line was in #4412,
gated behind this one. The fix could not satisfy the coverage gate while its own
test waited downstream of it. #4412 is closed; nothing is lost.

**`main` merged in** (`570c4661`). The previous head's `uat` failure was the
lone `#/analytics` 390px case, which was never this PR's — #4416 fixed it on
`main` at 11:39, after this branch was cut.

### Verified on the merged branch

- `format.test.ts`, `home.readings.test.tsx`, `scripts/autocleanup.test.ts` — 58/58.
- Full unit suite — **577 files, 6685 tests, 0 failed.**

### One inherited red, and it is not this PR's

`fe-lint` fails on `main` right now, and every open PR inherits it:

```
src/design-system/recordtabs.css:38:3  lint/suspicious/noDuplicateProperties
```

`.recordtabs` declares `min-inline-size: 0` **twice**, at lines 18 and 38, with
two different explanatory comments. Two commits fixed the same 390px defect
within minutes of each other and neither saw the other: #4416 (`688cb837`) and
#4392 (`570c4661`). Both are merged.

It is a one-line deletion — keep one declaration and the comment that reads
better — but it belongs to whoever owns those two merges, not to a compact-money
fix, so I have not folded it in here. No open PR is fixing it yet.

Once that lands, this PR should be green on every lane.

---

joshbruh-lang:
Pushed the merge of `main`, plus one fix that belongs to this PR's lane and one report of a failure that does not.

### Fixed here: `fe-lint` was red on `main`

`.recordtabs` carried `min-inline-size: 0` **twice**. #4392 and #4416 each reached that block from a different direction — one from Analytics' flex header caught by the 390px sweep, one from the strip's own `overflow-x: auto` — and neither could see the other. Biome's `noDuplicateProperties` fails the second copy.

Proven to be `main`'s and not this branch's before touching it: Biome run against a clean `origin/main` worktree containing none of my changes exits 1 with the identical error. This PR only inherited it because CI lints the merge commit.

I fixed it here rather than reporting it, because it is one line, the two declarations are byte-identical so nothing behavioural moves, and it blocks this PR's own `fe-lint`. The surviving comment folds in what each of the two knew: the rule never bites inside a `work` container, because the block below pins an explicit `inline-size`; outside one, a flex header is where it does.

### Not fixed here: `fe-unit` is also red on `main`

`src/App.test.tsx > the custom-fields admin … > mounts the field builder on the Data model page` fails waiting for `heading level 2 "Custom fields"`, which never mounts. The rendered DOM has the settings nav — "Settings", "You", "Account" — and no section body under it. It dies at 1016ms, the 1000ms `asyncUtilTimeout`, so the section is absent rather than slow.

Three checks say it is not mine:

- It reproduces on a **pristine `origin/main` worktree**, same test, same error.
- It fails **identically with this PR's `afterEach(cleanup)` disarmed**, so the cleanup registration is not what withholds the heading.
- It did **not** exist in this PR's baseline: the full-suite runs recorded above, on `a58d0b8`, had `App.test.tsx` green.

It arrived with 70c0b73 (#4419), which moved the settings catalog, addresses and visibility predicate out of `settings.tsx` into `settingsnav.tsx`. The test grants an admin `custom_field: [read, create, update]`, and `data-model` is gated on `customField || pipeline || product || offerTemplate || tag`, so the entry ought to be visible — which points at the page body rather than at the predicate. I have not taken it further.

I am not fixing it in this PR, and the reason is scope rather than difficulty: it is a behavioural regression inside a 510-line module merged twenty minutes ago, and the plausible repairs — the section's mount condition, or the test's fixture — are a choice about what #4419 intended, which belongs to its author. No fix exists to port; I searched the open PRs, and #4405 *adds* the `min-inline-size` declaration rather than deduping it. @LarsGradion, flagging it since #4419 is yours.

So this PR's `fe-unit` will stay red on that one test until `main` is fixed, through no property of this diff. Everything else in the merged tree is green: **578 files, 7185 tests, one failure, and that one is the above.**

### Also merged

A sibling session pushed a unit test for `formatMoneyCompact` onto this branch (2764a49) — exact-string assertions covering the round figure, the remainder, the sub-threshold case and the VND scale. Good gate, and it closes a real hole: the helper had no test of its own.

I corrected its explanatory comment in `d4068dc`. It described the defect as an ICU-build difference — one reader's browser drawing `€420.0k` where another drew `€420k`. It is not environmental: ECMA-402 clamps an absent `minimumFractionDigits` down to the maximum and nothing else, so `min(2, 1) = 1` on every conforming runtime and every reader saw the same wrong figure. That distinction is the useful part of the diagnosis — an environmental defect is chased by collecting more environments, whereas this one reproduced 5 of 5 times on a single machine, idle and loaded alike, which is exactly what identified it. The assertions were already right and are unchanged.

---

joshbruh-lang:
## Conflict resolved, and this PR gives the `recordtabs` fix back to `main`

The PR went `mergeable_state: dirty` because `.recordtabs` was deduplicated in
two places at once. Both were correct and they kept **different halves**:

- **`main`** dropped the first declaration (the one #4392 added) and kept the
  second, with #4416's comment.
- **This branch** (`7f9c847b`) dropped the second and kept the first, under a
  comment merging both explanations.

Same end state, incompatible diffs — so git could not take either.

Resolved by taking `main`'s file verbatim: `git diff origin/main --
frontend/src/design-system/recordtabs.css` is now empty. That is the right way
round twice over — `main` is authoritative and already carries the fix, and it
returns this PR to its own scope. Nothing is lost: the surviving declaration is
the one that fixes the 390px `#/analytics` overflow, and `.recordtabs` still has
exactly one.

The diff against `main` is now just the four files it started as:

```
frontend/scripts/autocleanup.test.ts  | 41 ++++
frontend/src/format/format.test.ts    | 42 ++++
frontend/src/format/format.ts         |  9 ++
frontend/vitest.setup.ts              | 20 +-
```

### Verified on `0e0660b1`

`make check-fe` — **green, exit 0**, with current `main` merged in. That is the
whole frontend half: ds gates, drift, biome, both typechecks, the unit suite and
the build.

Worth noting for whoever merges: `fe-lint` was red on `main` for about half an
hour from that duplicate, so every open PR inherited it. It is clean on `main`
now, and with this PR's formatter fix the `fe-unit` red goes too — those were the
last two shared blockers in the queue.

---

joshbruh-lang:
Both `main`-red failures I reported above are resolved, and this PR is green. Correcting my previous comment, which said `fe-unit` would stay red until `main` was fixed — it no longer is.

- **`App.test.tsx`** — fixed on `main` by 865b054 (#4427), which pre-imports the lazy `screens/settings` module inside the two affected cases so the dynamic import's transform cost sits outside the `findBy` window rather than inside it. It reached this branch through a sibling session's merge. `App.test.tsx` is **21/21** here now.
- **`recordtabs.css`** — deduped on `main` independently, so my fix for it is a no-op and has dropped out of the diff. `.recordtabs` carries exactly one `min-inline-size` on both sides.

The effective diff against `main` is back to the four files this PR is actually about: `format.ts`, `format.test.ts`, `vitest.setup.ts`, `scripts/autocleanup.test.ts` — 111 insertions, 1 deletion.

**Full suite on the current head (0e0660b): `Test Files 579 passed (579)`, `Tests 7192 passed (7192)`**, rc=0, no unhandled errors. CI agrees: `fe-unit`, `fe-quality`, `fe-bundle`, `live-boot`, `secret-scan`, `changes` and `craft-residue` all green; the Go gates skipped, as the diff has no Go.

One note on 865b054's shape, since it lands next to the defect this PR is about and a later reader may wonder whether it is the same anti-pattern. It is not: `await import("./screens/settings")` does not extend a budget, it moves a one-off module transform out of the measured window, so the 1000ms wait then measures routing rather than Vite. The distinction matters because the rule in `frontend/AGENTS.md` is about waits that expire on something that never arrives — here the module does arrive, just after the budget when it has to be transformed first.

The description above still carries the sentence "full-suite runs are still finishing as I open this". They finished; the verdicts are in the earlier comment and this one. I have left the description otherwise as-is rather than rewriting around cubic's generated section.

---

joshbruh-lang:
## SonarCloud: 0.0% on new code, and the denominator was one unlisted file

The quality gate failed on *0.0% Coverage on New Code (required ≥ 80%)*. Pushed `67dd1ed`, which adds one line to `sonar-project.properties`. The measurement behind it, since "add a test" was the obvious reading and it is the wrong one:

| new file/lines | how the scan treats it | coverable new lines |
|---|---|---|
| `frontend/scripts/autocleanup.test.ts` | `sonar.exclusions` names `frontend/scripts/**` — not analysed | 0 |
| `frontend/src/format/format.test.ts` | `sonar.test.inclusions` — test code, coverage-exempt | 0 |
| `frontend/src/format/format.ts` | analysed; 8 comment lines + 1 object-literal property | 0 |
| `frontend/vitest.setup.ts` | analysed as **production**, absent from every lcov | all of them, none covered |

So the denominator is `vitest.setup.ts` alone, and the numerator can only be zero. That is why the earlier round — folding #4412's `format.test.ts` in to satisfy this same gate — could not move it off 0.0%: that test covers a line the scan was never counting.

**Measured, not inferred.** The lcov from `vitest run --coverage.enabled` carries 817 records and `vitest.setup.ts` is not one of them; `format.ts` is present at 69 of 71 lines. And the single product line this PR adds — `minimumFractionDigits: 0,` at `format.ts:113` — sits inside one `return new Intl.NumberFormat({...}).format(major)` statement, whose coverage v8 attributes to the statement's own line: `format.ts` has **no DA record between lines 98 and 146**. It is not a coverable line, so no test could have raised the ratio through it either.

**Why the file was unlisted.** It is the suite's own setup — the hooks and DOM stubs vitest runs *around* the modules it instruments. The v8 provider reports the code under test, never the harness that arranged it, so the file structurally cannot appear in a report this repository can produce: 0% however thoroughly the suite exercises it. `vite.config.ts` and `playwright.config.ts`, its two neighbours in the frontend root, are already coverage-excluded; `vitest.setup.ts` never was, and nothing had added a line to it since those lists were written — it and `sonar-project.properties` were last touched by the same commit, `e12ecf4`. It had therefore never appeared as new code, and the gap had no way to show until a change touched it. This is that change.

Put in `sonar.coverage.exclusions` rather than `sonar.test.inclusions` deliberately: the wrong metric is coverage and only coverage, so its issue and duplication treatment stays exactly as it was. The rationale in the config follows the one already written there for `*_windows.go` and the WebGL renderers — an honest exclusion beats a gap no work would close.

`check-lcov-paths.sh` still passes on the generated report (817 records, all resolving from the repo root), and its own test is green on all 7 cases. Full suite with coverage enabled: **579 files, 7192 tests, 0 failed.**

One thing I cannot verify from here: the gate is recomputed server-side, so whether this clears it will show on the next scan rather than in anything I can run locally.

---

joshbruh-lang:
## Green on every lane, and the backend half is now actually exercised

Both open items from my last two comments are closed on `ea6cdda`.

**SonarCloud passes.** The quality gate cleared once `frontend/vitest.setup.ts` stopped being counted as uncovered product code. That fix had never been evaluated before now: the `sonarcloud` job is gated behind `ci`, and `ci` was failing on the integration lane, so the scan was skipped rather than run and failed.

**The `./migrations` failure did not recur.** Shard 4/6 is green on a current base, as are all six shards, the integration aggregate and the unit-coverage pod. Nothing was changed to make that happen — this PR still contains no Go and no SQL — so the earlier red was a property of the base it was merged against, not of the diff. I did not re-run the job to get this; the merge of current `main` was owed anyway, and re-running the lane was its side effect rather than its purpose.

**Worth flagging for whoever merges:** this is the first head on this PR where the backend lanes ran at all. `sonar-project.properties` is a root file, so the `changes` router stopped classifying the PR as frontend-only — and six integration shards, `craftsmanship`, `govulncheck`, `deterministic-gates`, `extension-reference` and `live-boot` all went from *skipped* to *passed*. That closes the gap the description's unticked `make check` box describes: the backend half could not be run in the environment this was developed in (no database, no docker daemon), and it has now been covered by CI instead. The box stays unticked because `make check` itself was never invoked end to end, which is the honest reading of it.

### Every check on `ea6cdda`

`ci` · `changes` · `craft-residue` · `secret-scan` · `craftsmanship` · `govulncheck` · `deterministic-gates` · `extension-reference` · `live-boot` · `integration` (6 shards + aggregate + unit coverage) · `frontend` (`fe-unit`, `fe-quality`, `fe-bundle`) · `uat` · `sonarcloud` · **`SonarCloud Code Analysis`** · `cubic` — all green. `license gate` is skipped, correctly: the diff contains no Go file to carry a header.

Locally on the merged tree: **583 files, 7230 tests, 0 failed**, lint and both typechecks clean.

### The diff, unchanged in substance

```
frontend/src/format/format.ts         |  9 ++    the fix
frontend/src/format/format.test.ts    | 42 ++    its gate
frontend/vitest.setup.ts              | 20 ++    the cleanup registration
frontend/scripts/autocleanup.test.ts  | 41 ++    its gate
sonar-project.properties              | 14 ++    the harness is not product
```

No open review threads. Ready for a human.

---

PR #4417, opened 2026-09-05
fix(frontend): the agent's cost on the rail is the administrator's figure

joshbruh-lang:
## What

The month's cost figure on the agent rail (bottom-left sidebar), in the head of the agent card that opens from it, and in the card's meta row is now shown only to seats holding the `admin` role. Every other seat still sees the runtime row (served model, tool count, posture); the money is absent, the rail button's accessible name carries no spend, and the `/ai/usage` read is not issued at all for such a seat.

Fixed along the way: `frontend/src/design-system/recordtabs.css` declared `min-inline-size: 0` twice on `.recordtabs` (one copy each from #4416 and #4392), which fails Biome's `noDuplicateProperties` and had the `fe-quality` lane red on `main`. The second copy is removed here; the one with the fuller comment stays.

## Why

The three renderings share one hook, and it gated on the `automation:update` grant alone — the grant the server serves `/ai/usage` on. That grant is wider than the administrator: `ops` holds it by default and an operator-edited role may hold it, so the figure was drawn for seats it does not belong to. What the agent costs is the administrator's figure, so the role narrows the grant; the grant stays beside it because an admin whose role lost it would only be asking for a 403.

`useHoldsAdminRole`'s doc comment lists the fixed set of surfaces allowed to read the role rather than a grant; this PR adds the rail's cost figure to that list with its reason (it narrows a grant that exists, rather than standing in for one that does not).

## How verified

- `pnpm vitest run src/app/agentrail.test.tsx` — all passing, including the new spec: an `ops` seat holding `automation:update` gets the runtime row and no money on the rail, in the panel head, in the meta row, or in the button's aria-label, and never requests `/ai/usage`. Confirmed the spec fails with the gate removed.
- `pnpm lint` (Biome) clean after the duplicate-property removal; reproduced the error first on main's copy of the file. `pnpm exec tsc -b` clean.
- Full `pnpm vitest run`: 6677 passed, 2 failed in `src/screens/home.readings.test.tsx` (`/420k/` expected, `€420.0k` rendered). Those two fail identically on `main` in this environment without this diff, so they are not this change's; likely an ICU/Node difference in the sandbox.

- [x] `make check` is green — frontend half run directly as above (typecheck, lint, unit); no Go files touched, so `check-backend` and `craft static` have nothing in scope

## AI involvement

Entire change AI-assisted from the request "costs on the agent in the sidebar and the agent card should only be shown to admins": locating the shared hook, the gate change, the doc-comment update, the test, the lint fix and this description.

## Accountability

https://claude.ai/code/session_01NKLZd93PAUUz4dHPNUUtvL


joshbruh-lang:
**CI note: `frontend / fe-quality` was red on the first head for a failure that is main's, not this PR's.**

Biome's `noDuplicateProperties` fails on `frontend/src/design-system/recordtabs.css`: `.recordtabs` declares `min-inline-size: 0` twice. The two copies arrived on `main` through #4416 and #4392 (each brought one), after this branch was cut, and `pnpm lint` on main's own copy of the file reproduces the error. No fix PR for it existed, so I merged `main` into this branch and removed the second copy here (the one with the fuller comment stays) — one line, and it no-ops once main carries the same removal. CI is re-running on the new head.

---

PR #4418, opened 2026-09-05
fix(worklist): the deal card says whether anybody is carrying the deal

### LINKED ISSUE #4304: Worklist's `deal.no_champion` field is never populated, on any row, though the `because` reason beside it fires correctly
# Worklist's `deal.no_champion` field is never populated, on any row, though the reason line beside it fires correctly

**Upstream:** _pending — filed after this draft, id to be added_

## Observed

Measured on `qc-own`, 2026-09-05, core `47052e50d`.

`WorklistDealFacts.no_champion` (`backend/api/crm.yaml`) is documented at
length as a genuine tri-state — `true` only when a committee exists, was read
in full, and nobody engaged holds the champion seat; **absent** (never
`false`) for a deal with no committee, a covered committee, or one this reader
could not read in full:

> "Null is not `false`. It is sent only when the answer is both known and
> negative, and stays absent in the three cases that would otherwise be
> rounded down to a finding ... A client MUST NOT render an absent value as
> 'nobody is carrying this'." — `crm.yaml`, `WorklistDealFacts.no_champion`

The ONE function that builds `WorklistDealFacts` for a Worklist row,
`dealFactsOf` (`backend/internal/compose/attention/classifyrisk.go:90-112`),
copies `StageId`, `OwnerId`, `AmountMinor`, `Currency`, `ExpectedCloseDate` and
`QuietDays` from the item — and never `NoChampion`. Every field on
`WorklistDealFacts` this build sets is present except that one, on every row,
regardless of what the coverage read actually found.

The sibling read for `/attention` gets it right, which is what makes the gap
visible: `dealFacts` (`backend/internal/compose/attention/rendersilence.go:151-166`)
builds `AttentionDealFacts` for the same `RiskyDeal` and explicitly carries
`NoChampion` through, with a comment saying why —

> "Carried through, not re-decided. The seam that read the committee already
> resolved the tri-state ... a renderer that reapplied the rule would be a
> second judgement over a fact it cannot see the evidence for."

`classifyRisk` itself (`classifyrisk.go:56-58`) reads `item.Deal.NoChampion`
(the `AttentionDealFacts` field `dealFacts` populated) directly and appends
the `because: [{kind: "no_champion"}]` reason correctly — this is a separate
code path from `dealFactsOf`, and it is the one a reader actually sees, which
is why nothing on screen currently reads wrong.

Reproduced live, building `worklist-because-fold` (UAT-FOLD-1): a deal with a
`champion`-role stakeholder seated and never engaged reads, on the same
`GET /v1/worklist?filter=deals_at_risk` row —

```
"because": [..., {"kind": "no_champion"}, ...]   // present, correct
"deal": {"no_champion": ...}                      // key absent — always, on this row and every other
```

confirmed by asserting `deal.no_champion === true` in setup and watching it
fail with `Received: undefined` before any fixture-reasoning question was
even in play — the row's `because` array already carried `no_champion`
correctly at that point.

## Cost

None visible today: `frontend/src/screens/worklist.copy.ts` and
`worklist.row.tsx` read only the `because` array to draw the row's reason
line, and grepping the whole frontend tree for `no_champion`/`NoChampion`
finds no reader of `WorklistItem.deal.no_champion` at all — only the
`because` kind, and the unrelated deal-detail-page badge
(`deal.strip.people.noChampion`, driven by a different read entirely). So no
current screen shows a wrong answer.

The cost is a dead, richly-documented contract field: a future client reading
`deal.no_champion` directly (a card, a census, an agent tool projecting the
field) gets `absent` unconditionally, which this field's own doc says a
reader "MUST NOT render ... as 'nobody is carrying this'" — meaning the safe
reading is also the reading that makes the field useless, forever, until this
is fixed. It is the sibling struct doing the right thing one file over that
makes this a five-minute, precisely locatable fix rather than a design
question.

## Workaround

None needed on any current screen — read the `because` array's `no_champion`
entry, which is correct, rather than `deal.no_champion`, which is always
absent.

---
Found by [margince-qc](https://github.com/margince/margince-qc), building
`UAT-FOLD-1` (`worklist-because-fold`) — proving a drifting deal's uncovered
champion seat folds correctly behind the Worklist row's 4th `because` reason,
and that a genuinely covered one never reads as a gap.

LarsGradion:
`dealFactsOf` builds the worklist row's `WorklistDealFacts` and copied every
field except `NoChampion`, so `no_champion` was absent on every row regardless
of what the champion-coverage read actually found. The lane's own projection
(`dealFacts` in `rendersilence.go`) carries it deliberately; this one dropped
it. The `because: no_champion` reason fires from a separate code path, which is
why nothing on screen read wrong and the gap stayed invisible.

**No screen changes today** — nothing in `frontend/` reads
`WorklistDealFacts.no_champion` yet, only the reason beside it. What changes is
that a client drawing champion coverage from the facts object now gets the same
answer the reason already gave.

Second, smaller fix in the same area: **a named owner is one the scope filters
can read.** A ranked row states its owner twice — `ownerRef` answers the
client, `owner` answers the scope filters through `answersTo` — and only the
second decides whose Mine queue a row lands on. `TestEveryProducerStatesAnOwner`
reads `ownerRef` alone, so a producer naming a person to the reader while
leaving `owner` zero would print that name on the card and put the row on
nobody's queue, with the census green. Latent today (the three producers that
call `ownedBy` set both); the new census keeps it that way.

Closes #4304.

| Obligation | Evidence |
|---|---|
| User-visible outcome | `TestTheDealCardSaysNobodyIsCarryingIt` — the row's `deal.no_champion` is a stated `true` where the coverage read found an uncovered committee. Field-only; no rendered change yet. |
| Permission/row scope | Unchanged. The tri-state is resolved by the coverage seam under the caller's own read; this projection copies and never re-decides, which is what keeps a withheld committee absent rather than reported. |
| Failure behavior | `TestTheCardMakesNoChampionClaimAboutAWithheldCommittee` — a committee the reader could not read in full stays absent, never `false`. |
| Idempotency/concurrency | n/a — pure projection, no write. |
| Mobile | n/a — no rendered change. |
| Storybook | n/a — no component change. |
| Accessibility | n/a — no component change. |
| Contract | No contract edit. `make check-backend` drift + composition reproducibility green. |
| Write shape | n/a — read path only. |
| Mutation strength | Three mutations, three distinct failures, no redundant test: (1) drop the field → `TestTheDealCardSaysNobodyIsCarryingIt` + `TestTheCardRepeatsAStatedFalse…`; (2) round `nil` down to `false` → `TestTheCardMakesNoChampionClaimAboutAWithheldCommittee`; (3) swallow a stated `false` into absent → `TestTheCardRepeatsAStatedFalse…`. For the census: dropping `owner:` from the task producer fails `TestANamedOwnerIsOneTheScopeFiltersCanRead` while `TestEveryProducerStatesAnOwner` stays green — which is the proof the gap was uncovered. Emptying the fixture trips its own count guard. |
| Full lane | `make check-backend` EXIT=0, zero failures, on `570c4661ef`. |

Reviewed by `craft-reviewer`, which found four things, all fixed here: a
duplicate absent-case test whose comment claimed a route it never travelled
(deleted), a comment that paraphrased its sibling instead of stating the
invariant (rewritten to the one fact that matters — absent is not false), the
missing stated-`false` input (now a test), and a floating section comment
(folded into the doc comment).

Codex review not run: out of quota for `gpt-5.3-codex-spark` until Sep 12.

---
## Summary by cubic
Fixes the worklist deal card to carry the `no_champion` field from the coverage read, so the facts object now matches the reason line. Also guards that named owners are set in the scope-filter field, not just the client-facing one.

- `dealFactsOf` now copies `NoChampion` verbatim, preserving the tri-state: absent stays absent, stated true and false are passed through.
- `TestANamedOwnerIsOneTheScopeFiltersCanRead` enforces that producers naming an owner also set the `owner` field used by `answersTo`, so rows land on the right Mine queue.
- `TestAFoldedRowNamingAPersonWouldNotBeScopedToThem` pins the one gap the census cannot reach: a folded row naming a person would carry a zero `owner`, so it fails if a foldable producer ever names one.

<sup>Written for commit b63fdff14a34e93b07463a058d46493ad3fa276d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/4418?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Deal cards now accurately reflect whether a champion is confirmed, explicitly absent, or unknown because committee information is unavailable.
  - Attention rows with named owners now maintain consistent owner information, improving owner-based filtering and visibility.

- **Tests**
  - Added coverage for champion-status handling and owner assignment to help prevent regressions.

### DIFF
--- a/backend/internal/compose/attention/classifyrisk.go
+++ b/backend/internal/compose/attention/classifyrisk.go
@@ -96,6 +96,10 @@ func dealFactsOf(item crmcontracts.AttentionItem) *crmcontracts.WorklistDealFact
 		OwnerId:     item.Deal.OwnerId,
 		AmountMinor: item.Deal.AmountMinor,
 		Currency:    item.Deal.Currency,
+		// Copied verbatim: absent is not false. A committee this reader could
+		// not read in full arrives nil, and normalizing it to false would tell
+		// a rep nobody is carrying a deal whose champion they simply cannot see.
+		NoChampion: item.Deal.NoChampion,
 	}
 	// The close date rides on the lane item's own due moment, and the idle
 	// count on its own typed field. Both were already resolved; only this projection
--- a/backend/internal/compose/attention/nochampion_test.go
+++ b/backend/internal/compose/attention/nochampion_test.go
@@ -99,3 +99,67 @@ func withDealFacts(deal RiskyDeal) RiskyDeal {
 	deal.AmountMinor, deal.Currency = &minor, &currency
 	return deal
 }
+
+// The reason and the fact are two answers, and a client reads both.
+//
+// The tests above prove the `because` line fires; this proves the typed field
+// beside it says the same thing. They are separate code paths — classifyRisk
+// appends the reason from the lane item, dealFactsOf projects the facts object
+// — so one carrying the tri-state proves nothing about the other. A card
+// drawing a champion-coverage badge reads the field, not the reason.
+func TestTheDealCardSaysNobodyIsCarryingIt(t *testing.T) {
+	uncovered := true
+	row := oneRiskRow(t, withDealFacts(RiskyDeal{
+		DealID: ids.NewV7(), Name: "Fleet retrofit",
+		QuietDays: 19, NoChampion: &uncovered,
+	}))
+	if row.Deal == nil {
+		t.Fatal("the row carries no deal facts at all")
+	}
+	if row.Deal.NoChampion == nil || !*row.Deal.NoChampion {
+		t.Errorf("the card states no_champion=%v, want a stated true", row.Deal.NoChampion)
+	}
+}
+
+// Absent, never false — the distinction the contract spells out at length.
+//
+// `false` here would be the card claiming the committee is covered, over a
+// read that never saw the committee. A rep reading that stops looking for a
+// champion on a deal whose champion they simply may not see.
+func TestTheCardMakesNoChampionClaimAboutAWithheldCommittee(t *testing.T) {
+	row := oneRiskRow(t, withDealFacts(RiskyDeal{
+		DealID: ids.NewV7(), Name: "Fleet retrofit",
+		QuietDays: 19, NoChampion: nil,
+	}))
+	if row.Deal == nil {
+		t.Fatal("the row carries no deal facts at all")
+	}
+	if row.Deal.NoChampion != nil {
+		t.Errorf("the card answers %v about a committee it could not read, want no answer", *row.Deal.NoChampion)
+	}
+}
+
+// The third input, and the one the contract's prose and its schema disagree on.
+//
+// `noChampionOf` never returns a stated false today — a covered committee
+// arrives absent — so this pins the projection's behaviour if one ever does.
+// It passes it through rather than swallowing it, because a stated false is
+// the coverage seam's answer to give: this row copies, it does not judge. The
+// schema permits `false`; the field's prose says it is "sent only when the
+// answer is both known and negative". Resolving that is the contract's work,
+// and until it is resolved a silent change of behaviour here should cost
+// somebody this test.
+func TestTheCardRepeatsAStatedFalseRatherThanInventingAnAnswer(t *testing.T) {
+	covered := false
+	row := oneRiskRow(t, withDealFacts(RiskyDeal{
+		DealID: ids.NewV7(), Name: "Fleet retrofit",
+		QuietDays: 19, NoChampion: &covered,
+	}))
+	if row.Deal == nil {
+		t.Fatal("the row carries no deal facts at all")
+	}
+	if row.Deal.NoChampion == nil || *row.Deal.NoChampion {
+		t.Errorf("the card states no_champion=%v over a seam that answered false",
+			row.Deal.NoChampion)
+	}
+}
--- a/backend/internal/compose/attention/ownerscope_test.go
+++ b/backend/internal/compose/attention/ownerscope_test.go
@@ -1,0 +1,75 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+package attention
+
+// One row, two owner answers, and nothing holding them together.
+//
+// A ranked row states its owner twice: `ownerRef` answers the CLIENT, and
+// `owner` answers the SCOPE FILTERS through answersTo. The split is deliberate
+// — a zero `owner` is a row the filters simply keep, where a client told
+// `unassigned` would read a confident claim nobody owes it — and only `owner`
+// decides whose Mine queue a row lands on.
+//
+// So a producer that names a person in `ownerRef` and leaves `owner` zero puts
+// a row on nobody's queue while the card beside it prints that person's name.
+// The existing census (TestEveryProducerStatesAnOwner) cannot see that: it reads
+// `ownerRef` alone, which is the half that would still be right.
+//
+// Today the three producers that name a person set both. This test is what
+// makes that stay true.
+
+import (
+	"testing"
+
+	openapi_types "github.com/oapi-codegen/runtime/types"
+
+	crmcontracts "github.com/margince/margince/backend/internal/contracts"
+	"github.com/margince/margince/backend/internal/shared/kernel/ids"
+)
+
+func TestANamedOwnerIsOneTheScopeFiltersCanRead(t *testing.T) {
+	t.Parallel()
+	holder := ids.NewV7()
+	rows := classifyDay(dayWhoseRowsNameAnOwner(holder), rankInstant, dayMoney{})
+	rows = append(rows,
+		classifyWaiting(WaitingCustomer{Since: rankInstant, OwnerID: holder}, rankInstant),
+		classifyLead(OwedLead{OwnerID: holder}, rankInstant))
+	named := 0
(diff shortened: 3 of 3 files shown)


LarsGradion:
---

**Correction to an earlier version of this PR body.** It said "Codex review not
run: out of quota for `gpt-5.3-codex-spark` until Sep 12." The quota part is
true of that one model, but the conclusion was wrong: the 5.6 line is
reachable and spells its names WITHOUT a `codex` infix (`gpt-5.6-sol`,
`gpt-5.6-terra`, `gpt-5.6-luna`). I inferred "Codex is down" from a single
capped model. Reviews have been re-run; the claim above is withdrawn.

**Disclosure trace for the newly-populated field**, since `no_champion: true`
asserts three things at once (a committee exists, was read in full, and nobody
engaged holds the seat) and could otherwise be an existence oracle over person
rows the reader cannot see:

- `ChampionCoverFor` (`backend/internal/modules/deals/championcover.go:65`)
  takes the caller's own `ctx` and `tx`. Its single caller,
  `attentionAtRisk.championCover` (`backend/internal/compose/attentionchampioncover.go:37`),
  runs it inside `database.WithWorkspaceTx(ctx, a.pool, …)` on the serving
  request — so the committee is resolved under the same principal that receives
  the row. No system-principal path, no cached or shared lane assembly.
- A reader who cannot see the seats gets `Withheld: true`
  (`championcover.go:80-81`), and `noChampionOf`
  (`attentionchampioncover.go:74-79`) returns `nil` for `!ok || Withheld ||
  Covered`. So withheld and seatless both arrive absent, and this projection
  copies that absence rather than rounding it to a finding.
- Folded rows cannot inherit a verdict: `batches.go` sets no `Deal` field at
  all, so a group row carries no deal facts.

Net: the field is absent in exactly the cases where a stated value would tell
the reader something, and `dealFactsOf` has one call site on the per-item path.

LarsGradion:
**Codex reviews ran after all** — the earlier "out of quota until Sep 12" was
wrong about the CLI, not just about one model. The 5.6 line spells its names
without a `codex` infix, and three reviews now cover this diff.

**On the disclosure question** — whether newly populating `no_champion` makes
it an existence oracle over person rows the reader cannot see — three
independent traces agree there is no path, and each found the same chain:

- `ChampionCoverFor` runs on the serving request's own context inside
  `database.WithWorkspaceTx`, so the committee is resolved under the principal
  that receives the row. `/worklist` is human-only, so no agent or system
  principal reaches the lane.
- `withheldSeats` negates the *whole* six-endpoint conjunction, so a seat
  refused by any arm sets `Withheld`, and `noChampionOf` returns nil for
  `!ok || Withheld || Covered`.
- Snapshots persist only `{Source, RowID}` and re-read live under the resuming
  reader, so no verdict crosses readers.
- Folded rows carry no `Deal` field at all, and at-risk rows are never folded.

The redteam pass added the argument that settles it: the identical fact already
reaches the same client on `main` as the `because: no_champion` reason
(`classifyrisk.go:56`, rendered from `worklist.because.no_champion`). This diff
adds a second spelling of something already on the wire, not a new disclosure.

**One real finding, now fixed in `dee5935`.** Codex noted the owner census does
not reach a fourth shape: `ownerOfTheGroup` speaks where a fold's members
agree, and a synthesized row has no record to take `owner` from, so a group
naming one person would carry `ownerNamed` with a zero `owner`.

My first attempt at that fixture was wrong and I want it on the record: I
forced an assignee onto the folded rows, but `classifyDecision` — the only
classifier the fold consumes — answers `unassigned()` and ignores `AssigneeId`
by a deliberate rule. The pile agreed on "nobody", the named arm never ran, and
the test passed while proving nothing. A probe of the actual batch row is what
caught it. The gap is now tested against `ownerOfTheGroup` directly and fails
the moment a foldable producer names a person.

Mutation checks on the final pair: dropping `owner:` from the task producer
fails the census; making `ownerOfTheGroup` stop agreeing fails the fold test.
`make check-backend` EXIT=0, zero failures.

LarsGradion:
Filed #4433 for the one question all three review passes raised and none could settle from the tree: the schema declares `no_champion` as `[boolean, 'null']`, but `noChampionOf` maps a covered committee to absent, so a stated `false` is a value the producer cannot currently send. The field's prose lists three absent-cases and a covered committee is not among them, so it neither licenses `false` nor forbids it.

That is a contract decision rather than something this PR should settle, and it is not a regression — the ambiguity predates this diff. `TestTheCardRepeatsAStatedFalseRatherThanInventingAnAnswer` pins today's behaviour (pass through, do not swallow) and is named in the issue as the test to update once the contract picks a side.

CI note: the first run's shard 1/6 failed with exit 143 — "The runner has received a shutdown signal", after 7 minutes with no test reporting FAIL. The `ci` and `integration` failures above it were that shard aggregating upward. Re-ran the failed jobs; the other five shards and every other lane passed unchanged.

PR #4463, opened 2026-09-05
fix(frontend): a case's rendered tree is unmounted when the case ends

### LINKED ISSUE #4435: fe-unit fails with every test passing: React schedules work past jsdom teardown in worklist.email.test.tsx
`fe-unit` fails with **every test passing**. Vitest reports one unhandled error and the job exits non-zero on it alone.

```
Test Files  578 passed (578)
     Errors  1 error

⎯⎯⎯⎯⎯ Uncaught Exception ⎯⎯⎯⎯⎯
ReferenceError: window is not defined
 ❯ react-dom/cjs/react-dom-client.development.js:17920:15
 ❯ Immediate.performWorkUntilDeadline scheduler/cjs/scheduler.development.js:45:48
 ❯ processImmediate node:internal/timers:574:21

This error originated in "src/screens/worklist.email.test.tsx" test file.
```

React's scheduler wakes on a `setImmediate` after jsdom has torn the environment down, so `window` is gone by the time the callback runs. Something rendered by that file schedules work that outlives its own cleanup.

## Observed

| run | branch | result |
|---|---|---|
| https://github.com/margince/margince/actions/runs/33959441814/job/101288642817 | https://github.com/margince/margince/pull/4393 | 6675 tests passed, job failed on this |
| https://github.com/margince/margince/actions/runs/33967724408/job/101310703312 | https://github.com/margince/margince/pull/4426 | 578 files passed, job failed on this |
| https://github.com/margince/margince/actions/runs/33967724408/job/101312978804 | https://github.com/margince/margince/pull/4426, re-run | passed |

Both failures cleared on a plain re-run of the same commit, which is what makes it a flake rather than a break — and what makes it expensive: the lane's verdict has to be re-run before it can be believed, on a job that takes about sixteen minutes.

## Why it is worth fixing rather than re-running

`docs/reference/issue-labels.md` puts it better than I can: *"A flaky gate earns [critical] not because it is severe but because a red run nobody trusts makes every other verdict unreadable."*

It also reads as a red `main` to the two-hourly health check, which files an issue naming whichever commits happened to land in the window — attribution pointed at the wrong change.

## What I did not establish

The rate. Two failures and one pass is what I saw while doing other work; I did not run the lane repeatedly to measure it, and I have not identified which render in `worklist.email.test.tsx` schedules the surviving work. Filed `priority: high` on that evidence rather than `critical` — if the rate is what these three runs suggest, the label doc's own rule says bump it.

The likely shape, for whoever picks it up: a component that starts a React transition, a timer or a subscription and is not unmounted before the environment goes. `cleanup` runs between files, so the survivor is scheduled late enough to miss it.

dungbanguyen:
Closes https://github.com/margince/margince/issues/4435.

`fe-unit` fails with **every test passing**. Vitest reports one uncaught exception and the job exits non-zero on it alone:

```
Test Files  578 passed (578)
     Errors  1 error

ReferenceError: window is not defined
 ❯ react-dom/cjs/react-dom-client.development.js:17920:15
 ❯ Immediate.performWorkUntilDeadline scheduler/cjs/scheduler.development.js:45:48

This error originated in "src/screens/worklist.email.test.tsx"
```

## What is actually wrong

**Nothing unmounts a test file's renders.** React Testing Library registers that hook itself — but only under a guard this suite does not satisfy:

```js
if (typeof process === 'undefined' || !process.env?.RTL_SKIP_AUTO_CLEANUP) {
  if (typeof afterEach === 'function') {
    afterEach(() => { cleanup() })
  }
```

`frontend/vite.config.ts` sets no `globals`, so vitest's default `globals: false` applies and `afterEach` is not a global. The library's auto-cleanup therefore never arms — silently, since the guard is a `typeof` check with no else. A file's trees are still mounted when the jsdom environment is torn down, and React's scheduler wakes on a `setImmediate` into a world with no `window`.

## Measured, not inferred

An `afterAll` probe on the named file, reading `document.body` at the moment its last case ends:

| | `bytes` | `children` |
|---|---|---|
| `origin/main` @ `85e324eb5` | 15412 | 5 |
| this branch | 0 | 0 |

Five live React roots — one per render — still holding state and still able to schedule work, in an environment about to stop existing. That is the "component that starts a transition, a timer or a subscription and is not unmounted before the environment goes" the issue predicted, and it is not one component: it is every render in the file.

It reads as a flake because **which** file pays is whichever happens to run last, so a re-run of the same commit clears it and the attribution lands on an unrelated diff — including, as the issue notes, on whichever commits the two-hourly health check happened to find in its window.

## Why the setup file rather than the one test

`grep -rl "@testing-library/react" src/` finds 418 suites that render; `grep -rl "cleanup()"` finds 299. **119 never clean up**, and the next one written will not either — there is nothing to remind an author of a library default that is off. Fixing `worklist.email.test.tsx` alone would move the failure to one of the other 118.

A suite that *does* call `cleanup` is unaffected: vitest runs a file's own `afterEach` before this one, and unmounting an already-unmounted tree is a no-op.

The import is dynamic so the node-environment suites — most of them — do not load react-dom for a hook that can only mean something where there is a DOM.

## The guard

`src/testing/autocleanup.test.tsx` holds it, across two cases because that is the only shape this claim has: the second asserts on what the first left behind, and the order dependence *is* the subject. Without it, nothing would notice the hook being removed — a live tree costs a failing assertion nowhere, only an uncaught exception somewhere else.

Mutation-checked. On `origin/main`:

```
× finds the previous case's tree already gone
AssertionError: expected '<div><p>a tree this case never unmoun…' to be ''
 Test Files  1 failed (1)
```

On this branch it passes.

## Verification

`make check-fe` green on Node 24:

```
Test Files  584 passed (584)
     Tests  7247 passed (7247)
EXIT=0
```

Against **583 / 7245** on `origin/main` — the difference is exactly this test file's two cases. So no suite among the 119 depended on its tree outliving its case, which was the real risk in changing a default that has been off for the life of the suite.

### DIFF
--- a/frontend/src/testing/autocleanup.test.tsx
+++ b/frontend/src/testing/autocleanup.test.tsx
@@ -1,0 +1,32 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+/** @vitest-environment jsdom */
+
+// The suite unmounts a case's tree without the case asking.
+//
+// React Testing Library's own auto-cleanup does not arm here — it registers on
+// a global `afterEach`, and this suite runs on vitest's `globals: false` — so
+// the hook is `vitest.setup.ts`'s to register. Nothing else would notice if it
+// stopped: 119 of the suites that render never call `cleanup` themselves, and
+// what a live tree costs is not a failing assertion but React's scheduler
+// waking after jsdom is gone, in whichever file happens to run last.
+//
+// So the claim is checked the only way it can be — across two cases, the second
+// asserting on what the first left behind. The order dependence IS the subject.
+
+import { render, screen } from "@testing-library/react";
+import { describe, expect, it } from "vitest";
+
+const TREE_TEXT = "a tree this case never unmounts";
+
+describe("a rendered tree does not outlive its case", () => {
+  it("renders one and leaves it mounted", () => {
+    render(<p>{TREE_TEXT}</p>);
+    expect(screen.getByText(TREE_TEXT)).toBeTruthy();
+  });
+
+  it("finds the previous case's tree already gone", () => {
+    expect(document.body.innerHTML).toBe("");
+  });
+});
--- a/frontend/vitest.setup.ts
+++ b/frontend/vitest.setup.ts
@@ -1,4 +1,4 @@
-import { beforeEach, vi } from "vitest";
+import { afterEach, beforeEach, vi } from "vitest";
 
 // Node ≥23 ships its own global Web Storage: a `localStorage` getter that
 // yields undefined unless the process was started with --localstorage-file.
@@ -101,6 +101,28 @@ if (typeof window !== "undefined") {
   beforeEach(() => {
     window.location.hash = "";
   });
+
+  // Unmount what a case rendered, because nothing else does.
+  //
+  // React Testing Library registers this hook itself — but only
+  // `if (typeof afterEach === 'function')`, and this suite runs on vitest's
+  // default `globals: false`, where that global does not exist. So the library's
+  // own cleanup never arms, and a file's renders are still mounted when the
+  // jsdom environment is torn down. React's scheduler then wakes on a
+  // `setImmediate` into a world with no `window` and throws there, which vitest
+  // reports as an uncaught exception and fails the lane on — a red run whose
+  // every test passed, and which clears on a re-run of the same commit.
+  //
+  // Registered here rather than asked of each file: 119 of the 418 suites that
+  // render never call `cleanup` themselves, and the next one written will not
+  // either. A suite that DOES call it is unaffected — vitest runs a file's own
+  // `afterEach` before this one, and unmounting an unmounted tree is a no-op.
+  //
+  // Imported here rather than at the top of the file so the node-environment
+  // suites, which are most of them, do not load react-dom for a hook that can
+  // only mean something where there is a DOM.
+  const { cleanup } = await import("@testing-library/react");
+  afterEach(cleanup);
 }
 
 // The calendar-drift lane: run the whole suite as if it were N days from now.
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -217,6 +217,17 @@ sonar.cpd.exclusions=\
 # `webgl-program.ts` is the shader compile and link both renderers share, which
 # is context calls end to end. Same rationale as *_windows.go: an honest
 # exclusion beats a gap no work would close.
+# frontend/vitest.setup.ts is the vitest suite's own setup file, and it belongs
+# with vite.config.ts and playwright.config.ts below rather than with the code
+# they configure. It is not test code by the `sonar.test.inclusions` patterns
+# above — those reach `frontend/src/**` and `frontend/e2e/**`, and this file sits
+# at `frontend/` — so it was indexed as product source and measured for coverage.
+# It can never report any: it runs to ESTABLISH the environment each suite is
+# instrumented inside, so v8 never attributes a line to it and it appears nowhere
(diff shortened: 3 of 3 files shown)


dungbanguyen:
## Second commit: the SonarCloud gate

First push was green on **every functional check** — `ci`, `uat`, `live-boot`, all three frontend lanes, `craft-residue`, `secret-scan`. The one red was `SonarCloud Code Analysis`, and it is a config gap this change exposed rather than anything wrong with the change.

The gate failed on one condition:

```
new_coverage   LT 80   actual 0.0   → ERROR
```

with the other five (`new_reliability_rating`, `new_security_rating`, `new_maintainability_rating`, `new_duplicated_lines_density`, `new_security_hotspots_reviewed`) all OK.

The only new-code file Sonar measures for coverage in this PR is **`frontend/vitest.setup.ts`** — the suite's own setup file. It falls outside every pattern that would have placed it correctly: `sonar.test.inclusions` reaches `frontend/src/**` and `frontend/e2e/**`, and this file sits at `frontend/`. So it is indexed as product source and measured for a coverage it structurally cannot report — it runs to *establish* the environment each suite is instrumented inside, so v8 attributes no line to it:

```
$ grep -c vitest.setup frontend/coverage/lcov.info
0
```

That is the same count `frontend/vite.config.ts` and `frontend/playwright.config.ts` return — and both of those are **already** in `sonar.coverage.exclusions`. `vitest.setup.ts` belongs beside them and was simply never added.

Latent rather than introduced: the exclusion list dates from #14 and this file has been indexed this way ever since. The gate reads *new* lines, so touching the file is what makes it bite — any new line here would have failed it, however thoroughly exercised.

Excluded from **coverage only**; the file stays analyzed for every other rule. The comment in `sonar-project.properties` states the reasoning in place, matching the honest-exclusion rationale that file already applies to `*_windows.go` and `desktop/launcher/**` — a gap no work could close is worse than saying so.

dungbanguyen:
Closing as a duplicate. A parallel session landed the same fix while this was in CI.

https://github.com/margince/margince/pull/4414 (`321dafffb`, 15:43Z) reached the same diagnosis independently and shipped **the same two lines**, in the same place:

```ts
const { cleanup } = await import("@testing-library/react");
afterEach(cleanup);
```

It also carries a behavioural guard test (`frontend/scripts/autocleanup.test.ts`, deliberately `.ts` so `tsconfig.node.json`'s `scripts/**/*.ts` typechecks it) **and** the `frontend/vitest.setup.ts` entry in `sonar.coverage.exclusions` — which is the second commit on this branch. Every part of this PR is on `main` already.

That is also exactly why the re-run went red. Merging `main` in put `const { cleanup }` in the block twice, which is a redeclaration error — so `fe-quality`, `fe-unit` and `frontend` all failed on a merge artifact of the duplicated work rather than on anything either change did wrong. For the record, the first run of this branch was green on every functional check (`ci`, `uat`, `live-boot`, all three frontend lanes), and `make fe-quality` passes on current `origin/main` on its own.

**Verified `main`'s version does the job**, using the same probe this PR's evidence rested on — an `afterAll` reading `document.body` at the moment `src/screens/worklist.email.test.tsx` ends:

| | `bytes` | `children` |
|---|---|---|
| `85e324eb5` (before either fix) | 15412 | 5 |
| current `origin/main` | 0 | 0 |

and `pnpm exec vitest run scripts/autocleanup.test.ts` passes (2/2).

One small thing I noticed and am **not** opening a PR for, because a full CI run for one character is not a good trade: the comment above the hook in `vitest.setup.ts` names `scripts/autocleanup.test.tsx`, but the file is `.ts` — and the same comment block explains at length why it must be `.ts`. Worth correcting in whatever next touches that file.

https://github.com/margince/margince/issues/4435 closed against https://github.com/margince/margince/pull/4414 with the verification.

PR #4477, opened 2026-09-05
Three daily passes collapse, and the sweep tag moves to the pass

dungbanguyen:
Batch five of #1857, stacked on #4470. The capture digest, the morning brief and the weekly review — 15 workspace pairs remain after this. Each keeps its own deadline (5m, 10m, 10m) from the child.

## The sweep tag had to move

`jobs.SweepTag` marks a row as part of a fleet pass, and the sweep gauges count tagged rows. It was stamped at the dispatch chokepoint, on the **children** a dispatcher enqueued — never on the dispatcher's own row.

The collapse deletes the children. So the gauges were left with nothing tagged to count, and a one-off enqueue of a collapsed kind became indistinguishable from a scheduled tick of it.

`periodicInsertOpts` now stamps the tag for a **fleet-wide** kind, which is what such a row now is. Both properties come back: what the gauges read is a pass that ran, and an untagged row of a fleet-wide kind is somebody's one-off — exactly what it meant before. Only fleet-wide, because a periodic kind that owns a tenant is not a fleet pass and tagging it would put a row in the gauges covering no workspaces.

`TestBackfillCompletionBuildsTheDigest` caught it, and its premise had to change twice over:

```
capture_digest rows went 1 -> 2 across the backfill; a completion must enqueue the
CHILD kind for its own workspace, never the dispatcher that fans out over the whole fleet
```

There is no pair to tell apart any more. What survives is the tag, so the test now tells a one-off from a scheduled tick — the distinction that still exists and still matters to the gauges.

`jobs.FleetWide`'s own doc comment stopped being true and is rewritten: the marker no longer implies *"does no tenant write"*, because a collapsed pass writes for every workspace it walks. What it still says is that the row owns no single tenant, so nothing may read a workspace off its args.

## Verification

`make check-backend` exit 0; `internal/compose` and `integration/capture` integration suites green.

## Summary by CodeRabbit

* **Improvements**
  * Daily digest capture, brief generation, and weekly review processing now run as consolidated passes across workspaces.
  * These operations have increased processing time allowances to better support larger workloads.
  * Scheduled processing now tracks fleet-wide work more consistently.

* **Maintenance**
  * Obsolete queued workspace-level tasks are cleaned up during migration, preventing stale work from being retried.

### DIFF
--- a/backend/api/jobs.yaml
+++ b/backend/api/jobs.yaml
@@ -368,43 +368,25 @@ kinds:
 
 
   capture_digest:
-    role: dispatcher
+    role: worker
     go_type: CaptureDigestArgs
     queue: default
-    timeout: 2m
+    timeout: 5m
     opts_owner: caller
     cadence: 24h
     registration: {when: [GmailRegistry], absent: registers_nothing}
-    fans_out_to: capture_digest_workspace
-    fan_out_unit: workspace
     reason: >-
       Daily, after the overnight passes it summarizes. The boot pass backfills
       a missed night, so a morning is never silently empty.
 
-  capture_digest_workspace:
-    role: worker
-    go_type: CaptureDigestWorkspaceArgs
-    queue: default
-    timeout: 5m
-    max_attempts: 3
-    opts_owner: fan_out
-    registration: {when: [GmailRegistry], absent: registers_nothing}
-    args:
-      Workspace: id
-    reason: >-
-      A database-only assembly, one digest per connected user — this worker
-      holds no model lane at all, which is also why it fans out onto default
-      rather than ai_capture.
 
   brief_generate:
-    role: dispatcher
+    role: worker
     go_type: BriefGenerateArgs
     queue: default
-    timeout: 2m
+    timeout: 10m
     opts_owner: caller
     cadence: 1h
-    fans_out_to: brief_generate_workspace
-    fan_out_unit: workspace
     reason: >-
       Hourly, not daily, because the morning it aims at is the installation's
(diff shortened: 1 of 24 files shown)


dungbanguyen:
Valid — applied, and to all five drain migrations in this stack rather than just the one you flagged.

You are right that `state NOT IN ('completed', 'discarded', 'cancelled')` matches `running`, and right about the window: the API entrypoint applies migrations before it serves, the worker entrypoint applies none, so a rolling deploy can have an old worker mid-pass on a child when the new API runs this. Deleting that row lands its completion write on nothing.

**The fix is an allowlist of the states that will never run** — `available`, `scheduled`, `retryable`, `pending` — rather than a drain-then-delete handoff. A running row needs no handoff: leaving it alone costs nothing, because it finishes, becomes terminal, and was never the problem. The problem was only ever the rows nothing will ever pick up.

An allowlist rather than a denylist for a second reason worth stating: a state River adds later is then excluded until somebody looks at it, instead of being swept in by a rule that never heard of it. That is the same failure in a slower form.

**Four earlier migrations in this series already landed with the denylist** (#4192, #4193, #4457, #4470). I have not edited those — a migration that has already run does not run again, so editing them would change nothing for any installation that has applied them and would only make the file disagree with what it did. The window they carried is a deploy-time race that has either happened or not; what it would have cost is a failed completion write and a log line, on work the collapsed pass had already done. Recording it here rather than quietly leaving it out of the story.

`make check-backend` exit 0 across the stack.

PR #4482, opened 2026-09-05
Three more passes collapse, and two vacuity floors come down

dungbanguyen:
Batch six of #1857, stacked on #4477. The assurance sweep, the close-date sweep and the org-name promotion — 12 workspace pairs remain after this. Each takes the child's 5-minute deadline.

## Two gates count dispatchers

Both had to move, and this is the batch's finding. Each holds a floor so it cannot pass on an empty walk — `jobdispatcherenqueue_test.go` counts dispatcher args literals in the package, `jobhealth_test.go` counts the dispatchers the contract declares:

```
only 17 dispatcher args value(s) were found, under the floor of 18
the contract declares only 17 dispatchers; the filter is not resolving them
```

The collapse retires dispatchers, so those numbers fall with it — a floor left where it was fails on the very change it was written to survive.

Both come down to 15 and now say what they track: they bottom out at the fan-outs that **stay**, the ones over a connection or a build rather than a workspace. They remain floors rather than equalities for the reason they always were — a floor stops a vacuous pass without pinning a number every sanctioned retirement has to come back and edit.

## A substitution with a shelf life

Three fixtures used `close_date_workspace` as their stand-in fan-out **child**: the dispatch helpers' unit tests, the fan-out tag suite and the registry test. They name `time_scan_workspace` now, which is still one.

When the last workspace pair collapses there will be no such kind left, and `dispatchPerWorkspace`, `workspaceSweepOpts` and the tests exercising them go with it. That is the final batch's work, not a gap here.

The close-date integration test waited on the workspace child, because a dispatcher completed as soon as its fan-out was enqueued. It waits on the pass now.

## Verification

`make check-backend` exit 0; `internal/compose` integration suite green.

## Summary by CodeRabbit

- **Bug Fixes**
  - Nightly assurance, close-date, and organization-name promotion passes now complete as unified jobs, improving completion tracking and reducing dispatch-related delays.
  - Removed obsolete queued child work during migration to prevent orphaned or duplicate processing.
  - Updated rollback handling so scheduled work resumes naturally if changes are reverted.

- **Operations**
  - Existing workspace processing remains covered while reducing the number of intermediate jobs required for scheduled passes.

### DIFF
--- a/backend/api/jobs.yaml
+++ b/backend/api/jobs.yaml
@@ -711,29 +711,13 @@ kinds:
       writes nothing rather than a duplicate.
 
   follow_up_reconcile:
-    role: dispatcher
+    role: worker
     go_type: FollowUpReconcileArgs
     queue: default
-    timeout: 2m
+    timeout: 5m
     opts_owner: caller
     cadence: {operator: ReconcileInterval}
-    fans_out_to: follow_up_workspace
-    fan_out_unit: workspace
 
-  follow_up_workspace:
-    role: worker
-    go_type: FollowUpWorkspaceArgs
-    queue: default
-    timeout: 5m
-    max_attempts: 5
-    opts_owner: fan_out
-    args:
-      Workspace: id
-    reason: >-
-      A database-only overnight pass over one tenant's open follow-ups, staging
-      what it proposes rather than acting. Five attempts for the reason
-      close_date_workspace takes five: both deals passes tick DAILY, so three
-      would cost a workspace a full day over a transient blip.
 
   fx_rate_refresh:
     role: worker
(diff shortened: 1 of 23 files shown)


dungbanguyen:
Both findings turn on the same hypothetical — a fleet with more than one workspace — so answering them together.

## The timeout

You are right about the arithmetic: River's deadline bounds the JOB, the declared number was chosen to bound ONE workspace's work, and `runPerWorkspace` puts N workspaces inside one job. On a fleet of three, workspace three could be cut off mid-pass.

**The fleet is one, by construction.** `identity.InstallationWorkspace` refuses to answer when a database holds more than one active workspace:

> `identity: more than one active workspace — the single-organization invariant requires an operator-led migration`

So a collapsed pass does exactly one workspace's work, and the number still bounds what it was chosen to bound.

**The enumeration is defensive, not expectant.** It stays because a pass written against the single-workspace assumption would silently skip the second one the day that invariant is lifted — and silently skipping a tenant is worse than a deadline that needs revisiting. What I will not do is claim the deadline is already right for that day: it is not, and it is the same class of thing the loop exists to protect against.

So it is recorded on #1857 as a named follow-up rather than argued away here — if the product ever admits a second workspace, the per-workspace deadlines have to move inside the turn (or the job's own has to become a fleet bound) before the enumeration does anyone any good.

## The late enqueue

Also right, and inherent to retiring a kind rather than to this migration. A migration runs once at a point in time; an old dispatcher still alive after it commits can enqueue one more child, and nothing in the new registry consumes it.

What bounds it: the window is one tick of one dispatcher, it closes when the deploy finishes and no dispatcher exists to enqueue any more, and River discards the row after its attempts. The cost is a handful of discarded rows in the deploy window rather than the permanent debris the migration exists to prevent — a row that fails a few times and stops, instead of one that sits runnable forever.

Draining old workers first would close it completely, and that is a deploy-ordering instruction rather than something the migration can assert. Naming it here so it is a known property of this series and not a surprise.

Issue #4496, opened 2026-09-05
testdb.AssertPoolsQuiesced reads a sibling parallel test as a leak


dungbanguyen:
## What is wrong

`testdb.AssertPoolsQuiesced` asserts that the **shared** pool has nothing checked out, and it is registered as a **per-test** cleanup. For a `t.Parallel()` test, `t.Cleanup` runs when that test finishes — while its siblings are still running and legitimately holding connections from the same pool.

So the first parallel test to finish reports the others as a leak:

```
snapshot_integration_test.go:79: the shared pool for postgres://…/margince_it_p5_24397 still had
1 connection(s) checked out 2s after this test ended. A goroutine this test started is still
running, and the next test resets the database under it
```

The message is confidently wrong: no goroutine that test started is running, and nothing is resetting the database under it.

## How it shows

`internal/modules/forecasting` is where it is dense enough to fire — ~41 `t.Parallel()` tests over one shared pool with the gate on each. It failed **four unrelated PRs in the same CI wave**, each on a different shard (3/6, 4/6, 5/6), always naming `TestTheCallHistoryIsTheChainNewestFirst` — which is simply whichever parallel test finished first while a sibling was mid-query.

Passes in isolation and passes for the whole package locally; it needs a loaded runner, which is why it reads as infrastructure noise and is not.

## Why it is filed rather than fixed

The two candidate fixes trade different things away, and the choice is a judgement about what the gate is for.

**Reference-count the holders** — `testdb.Pool` increments, `AssertPoolsQuiesced` decrements and only asserts at zero. Keeps the parallelism and keeps the gate firing, but a leak is then attributed to the *last* test out rather than the one that leaked, which is most of the gate's diagnostic value. It also drifts silently if a file calls `Pool` more often than it registers the gate — `modulepoolsharing_test.go` enforces the pairing per FILE, not per call, so that is reachable.

**Drop `t.Parallel()` where the pool is shared** — keeps the attribution exactly as it is and costs the parallelism of ~41 tests. Honest, and arguably what a per-test assertion about a shared resource already implies.

Either way it is shared test infrastructure under ~30 packages, and getting it wrong disables a real leak gate silently — which is the one direction the gate must not fail in. The existing doc already reasons carefully about that (it waits rather than samples, for exactly this reason), so the parallel case deserves the same care rather than a quick patch under a red main.

## What the gate's own doc does not yet say

[...]

PR #4497, opened 2026-09-05
The reader is a way in, never somebody to ask

joshbruh-lang:
## What

On a contact the reader corresponds with themselves, the network tab's recommended route **is them** — and every sentence on the page was written about a colleague to go and ask. An admin logged in as themselves was told *"Ask Demo Admin — they already write to each other"*, offered an *"Ask Demo Admin for an intro"* button, read *"5 from Demo Admin"* on the split legend, and was counted among *"2 colleagues"*.

After this PR the reader's own route speaks in the second person — verdict, chain, exchange split, routes rows, and the strip's count — and the ask is replaced by what to do instead. Both writers refuse an introduction whose introducer is the caller (`400`), and the account People map stops offering the verb on a seat only the reader reaches.

## Why

**An introduction has two people on our side, and the requester is already one of them.**

Nothing in the route read excludes the person doing the reading; the server ranks every colleague who corresponds with the contact and the reader is among them. That is the *most useful* fact the surface can report — it must not be dropped — but it is never a favour to ask for. Left as it was, the ask would have written a row putting a favour in the asker's own queue, recorded a handoff nobody made, and the account draft spent the workspace's model budget writing a letter addressed to its own sender.

Spelled on both sides of the wire, in one change:

- **Writers refuse it.** `introductions.Store.Create` and `compose/org360.IntroRequestDraft` both answer `apperrors.ErrInvalidArgument`. This is what makes the rule true for clients this repo does not own.
- **Surfaces do not offer it.** One predicate, `useOwnRoute`, answers "is this me" for the lead panel, the routes rows, the strip and the map's verb — three separate comparisons would be three chances to answer it differently on one screen. On the account map, the reader's route is filtered out *before* the ranking, so the second-best colleague is asked rather than nobody.

Found along the way and fixed here (*fix it, don't file it*): the picture's **"use this route"** compared a bare user id against a `user:<uuid>` node id, so it matched nothing and the verb silently did nothing on every route. It reads the node's own `user_id` now.

Copy lands in all three catalogues, and the reader is still **drawn** on both pictures — their relationship is the page's best fact even where there is no favour to ask for.

## How verified

- `make check-fe` — green (13m41s, full core suite, composed typecheck, extension tests).
- `make -C backend check` — green apart from two failures reproduced on a clean checkout of `main` in this same container, neither touched by this diff: `TestTheGateInventoryPageIsCurrent` (the inventory page is already stale on `main`) and `TestBaselineEraFixtureIsTheMatrixTheBaselineSeeded` (needs full git history; this checkout is shallow).
- Integration, against a real Postgres: `go test -tags integration ./internal/modules/introductions/` and `./internal/compose/integration/org360/` both green, including the two new refusals. The two existing org360 draft tests were passing the caller as their own introducer, so they now route through a second rep — which is the shape they were describing all along.
- `craft static --strict` on the changed Go files: 0 blocker, 0 major, 0 minor.
- New tests, each failing before the change: the panel's second person and absent ask, "Only you" and "You and 1 colleague" in the strip, the second-person route line when the reader ranks second, the map verb opening the drawer for a colleague and nothing on the reader's own node, and on the account map — no verb on a seat only the reader reaches, the next colleague asked when the reader holds the warmest route, and the reader's edge still drawn.
- Storybook: `TheReadersOwnRoute` added to `Records/Person/NetworkTab`, checked in both themes.

- [x] `make check` is green (see the two pre-existing `main` failures named above)
- [x] `make test-integration` is green — the touched packages were run against a real Postgres; the write shape of `intro_request` is unchanged, only a refusal is added ahead of it

## AI involvement

AI-assisted throughout, under review: the diagnosis, the copy in all three catalogues, the code and the tests were generated and then read line by line against the surrounding files. The German and Vietnamese strings follow the catalogues' existing register (`Sie` / `bạn`); the Vietnamese line was reworded once because the record-noun gate caught it calling the record a contact.

## Accountability

https://claude.ai/code/session_01NDspUQe8sRoodmXbpYVweT

---
_Generated by [Claude Code](https://claude.ai/code/session_01NDspUQe8sRoodmXbpYVweT)_

---
## Summary by cubic
Makes the reader's own recommended route on a contact's network tab speak in the second person, and refuses self-introductions in both backend writers and the account People map.

- The network tab's verdict, chain, split legend, routes rows, and strip count now address the reader directly ("You already write to each other") and replace the ask button with "Nobody to ask — write to them yourself."
- `introductions.Store.Create` and `compose/org360.IntroRequestDraft` now return `ErrInvalidArgument` when the introducer is the caller.
- The account People map filters the reader's own route before ranking, so the next best colleague is asked, and the intro verb is hidden on seats only the reader reaches.
- Fixed the picture's "use this route" action, which compared a bare user id to a `user:<uuid>` node id and silently did nothing; it now reads the node's own `user_id`.

**Bug Fixes**
- Forecasting integration tests that share a database pool no longer run in parallel, fixing a deterministic `main` failure where the first test to finish reported its siblings' checked-out connections as a leak.

<sup>Written for commit 2261c3a233894d122f4260337997429f62f8a061. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/4497?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


joshbruh-lang:
`integration / integration shard (4/6)` failed on 98f1d85, and it is not this PR's.

The one failing package is `./internal/modules/forecasting`:

```
snapshot_integration_test.go:79: the shared pool for …/margince_it_p5_24533 still had
1 connection(s) checked out 2s after this test ended.
--- FAIL: TestTheCallHistoryIsTheChainNewestFirst (2.50s)
--- PASS: TestACappedHistoryKeepsTheNewestCalls (5.07s)
```

That is #4496: `testdb.AssertPoolsQuiesced` is a per-test cleanup asserting the *shared* pool is empty, so the first `t.Parallel()` case to finish reads its still-running sibling as a leak. The timings above are the shape exactly — the failing test ended at 2.50s (0.5s of work plus the 2s grace it burned) while `TestACappedHistoryKeepsTheNewestCalls` ran on to 5.07s. Nothing in this diff touches `forecasting`, and `go list -deps ./internal/modules/forecasting` reaches neither `introductions` nor `org360`; the packages this PR does change passed on the same shard.

**Ported #4498** into this branch (cherry-pick of ebee5fb, `-x`): the nine `t.Parallel()` calls in the two `forecasting` files that take the shared pool. It no-ops once `main` carries it. Verified here against a real Postgres — `./internal/modules/forecasting` green on both the integration and unit lanes, and `craft static --strict` clean on both touched files.

No re-run spent: #4496 records this failure as **deterministic**, not load-dependent — it survived a full rerun on all three PRs it was blocking, on three different shards, naming the same test each time. A re-run here would have bought nothing.

---

PR #4536, opened 2026-09-05
feat(worklist): a meeting that happened asks how it went

### LINKED ISSUE #4321: A meeting row has nowhere to record what the meeting came to
Recording a meeting's outcome from the worklist was scoped as a wiring job over existing endpoints. It is not — but not for the reason first assumed either.

**The write exists.** `PATCH /activities/{id}` carries `meeting_status` with `booked | held | no_show | canceled`, and `frontend/src/screens/logactivity.tsx` already sends `held`. Transcripts ship too. So no new backend surface is needed for the fact itself.

**The row has nowhere to put it.** `backend/internal/compose/attention/meeting.go` gives a meeting card no `open` verb at all, deliberately: the card's subject is the ACTIVITY, a timeline entry with no page of its own, so a verb would advertise a destination that does not exist. The one way in it does offer is the brief, which opens as `?prep=<activity>` on a *person's* record and needs both ids.

**So the choice is one of two shapes, and neither is wiring:**
- a control that answers inline on the row, the way an approval's decision card does; or
- a destination for an activity, which is a new surface.

**What it means if nobody decides:** a rep who has just finished a call still leaves the queue to record how it went, which is the navigation the worklist exists to remove. The queue can tell them the meeting happened and cannot take the answer.

Related: `AC-WORKLIST-SDR-03` now holds the narrower rule that a row claims no verb it cannot perform, so adding an outcome verb without a place to answer it would fail `frontend/src/screens/worklist.verbcensus.test.ts`.

LarsGradion:
Lars decided **1A** on #4321: a lane of its own, not a widening of the existing
one.

The queue showed meetings still ahead and dropped each one the moment it
started, so **the meeting a rep just walked out of reached no lane at all.**
Asking how it went had nowhere to be asked — which is why the fix is a row
rather than a verb.

The new lane is the counterpart of `meetings`: that one reads `[now, end of
today)` and is about preparing, this reads `[start of today, now)` and is about
closing off. A meeting is on exactly one of them, and the split is the reader's
own moment. It clears itself — recording an outcome takes the row off.

**A meeting with no status at all is on it.** Capture writes calendar events
without one, so treating an absent status as settled would empty this lane on
exactly the installations whose calendars are connected. That is the arm the
integration test names.

Three commits, each standing alone:

1. **`startOfDay`, and one zone lookup for both ends.** Reuses the existing
   DST-safe walk rather than building a midnight — local midnight does not
   exist in Havana or Santiago, which spring forward at 00:00.
2. **The store can ask the question.** A SQL predicate, not a Go-side filter:
   the forward lane discards non-booked rows after reading and says why it can
   (its window is small), but backward almost every meeting is settled, so
   filtering after the read would push the genuinely unreported meeting off the
   page and draw "nothing to report" over real work.
3. **The lane itself** — contract, seam, classifier, renderer, censuses,
   frontend registries, story.

Closes #4321.

| Obligation | Evidence |
|---|---|
| User-visible outcome | `AMeetingWithNoOutcomeRecorded` story plus the lane census; a meeting that started today with no outcome now reaches the queue. |
| Permission/row scope | Reads through the same gated `ListActivities` the forward lane uses. A withheld-content row has its `Subject` nulled **at the store** (`activityprojection.go:145`), so the seam carries `""` and the row renders "A meeting" through the `worklist.untitled.meeting_outcome` fallback — discoverable, unreadable, and honest about which. |
| Failure behavior | Unbound seam renders the lane **absent**, never empty — `TestEveryLaneWithASeamIsWiredToIt` caught exactly this and made me wire it. |
| Idempotency/concurrency | Read path only; no write added. |
| Mobile | No new component; the row uses the existing grammar. |
| Storybook | `AMeetingWithNoOutcomeRecorded` added; `worklist.storycensus.test.ts` covers it. |
| Accessibility | No new control. |
| Contract | `crm.yaml` lane + count + `lanes_omitted` + three source enums + reason kind; both generators re-run. The count is deliberately **not** in `required` — adding it there would break existing clients. I diffed every generated constant name before and after: **zero removals, five additions**, so the enum-collision trap that once silently unprefixed seven `SavedViewResource` constants did not fire. |
| Write shape | n/a — read path. |
| Mutation strength | Dropping the NULL arm fails `TestAwaitingOutcomeCarriesTheUnsettledMeetingsAndNoOthers` **by name**; ignoring the dial fails it too; the naive-midnight `startOfDay` fails on Havana 2026-03-08 and Santiago 2026-09-06; reverting the off-by-one day fails all three zone arms. |
| Full lane | `make check` (both halves) EXIT=0. Integration lane run against real Postgres for the new store predicate. |

The typechecker caught one real error I would otherwise have shipped: I used
`outcome_unrecorded` as a **consequence**, but consequence is a separate closed
vocabulary answering "what does it cost". It is `data_drifts` — an unanswered
meeting means the record stops matching what happened, so every count over held
meetings reads short.

No verb on the row yet. Recording the outcome is a write on the activity, and
the source/action census says so explicitly rather than advertising a control
the queue cannot perform.

---
## Summary by cubic
The worklist showed meetings still ahead and dropped each one the moment it started, so the meeting a rep just walked out of reached no lane at all. Adds a `meetings_unreported` lane covering today's started meetings with no recorded outcome, as the counterpart of `meetings` — a meeting sits on exactly one of the two, split at the reader's own moment.

- The status filter is in SQL (`AwaitingOutcome`), not a Go-side filter: almost every past meeting is settled, and a post-read filter would push the genuinely unreported row off the page.
- A meeting with no status counts as awaiting; capture writes calendar events without one.
- Rows carry `occurred_at` and no `due_at` — a meeting that happened can't be late.
- The day boundary is DST-safe: a naive midnight doesn't exist in Havana or Santiago on spring-forward days.
- The count isn't in `required`, so existing clients are unaffected.
- Rows file under `review_and_repair` rather than prepare, and a full page reports itself as cut.
- No verb on the row yet; recording an outcome is a write on the activity.

Closes #4321.

<sup>Written for commit f9e298990d70ee373ff7c2361841244fc003dc9b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/4536?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


LarsGradion:
Codex (`gpt-5.6-sol`) raised four MAJOR findings. **Two were real and are fixed in `f9e2989`; two I checked and did not act on.** Codex flagged that it could not run `go test` in its sandbox, so I verified every claim by execution rather than taking it on inspection.

**Fixed — the lane reported itself complete when it was cut.** It reads at `plannedCap` with no entry in `boundedSources`, so a day holding twelve unsettled meetings said there was nothing more to close off while a thirteenth was pushed off the page. Under-reading is the one direction a census must never fail in. Mutation: removing the entry reports a full page as complete.

**Fixed — a meeting that happened was filed under "prepare conversations".** `preparesAConversation` admits the whole meetings category, so it caught these rows and told a rep to get ready for something they had already done. It is `review_and_repair`.

Worth noting how the first attempt failed: I made `preparesAConversation` return `false` for the new source, which changed **nothing** — `BriefSectionOf` falls through to `sectionOfCategory(meetings)`, which returns `prepare_conversations` anyway. It needed a positive rule placed before it. Mutation: removing that rule files the row under `prepare_conversations` by name.

**Not acted on — `unassigned()` and dual mine/unassigned membership.** Real, and **pre-existing across the whole meetings family rather than introduced here.** I probed `classifyDay` over the existing `meeting` lane: `ownerKind=unassigned`, `answersTo` names nobody — identical to my row. `classifyMeeting` argues at length why: the lane reads under row scope with no owner or attendee predicate, so naming the reader would make a colleague's meeting look like their own. Changing it here alone would make the two lanes disagree about one question. Worth its own issue if the dual membership is wrong; it is not this PR's to change unilaterally.

**Not acted on — double timezone resolution.** `assembleDay` and `startOfDay` each call `Service.location`, so an operator changing the installation zone mid-assembly could give the two lanes different days. True, and it is the reason I consolidated the lookup into `location` in the first commit rather than leaving two inline copies. Narrowing it to one resolution per assembly means threading the boundary through `optionalLanes`, which touches every lane — larger than this change and not a regression it introduces.

Third finding confirmed correct as built: a withheld-content row has its `Subject` nulled at the store (`activityprojection.go:145`), so the seam carries `""` and the row renders through the untitled fallback. Nothing leaks.

`make check` EXIT=0 both halves on the rebased base.

PR #4544, opened 2026-09-05
The lit window edge is a preference, and off is one of its answers

joshbruh-lang:
## What

The agent panel — the card that opens from the agent block at the foot of the rail — now carries one control in its foot: a switch for the lit window edge, the waves that travel the rim of the whole window while a run or a mailbox import is in flight.

On by default, so nothing moves for anybody who does not touch it. Off unmounts `AgentEdge` entirely and is remembered per browser, beside the theme.

## Why

The margins are the one surface in the workspace that moves without being asked for, and they move around the whole window rather than inside a panel. That is fine for most readers and genuinely unworkable for some — movement in peripheral vision — and merely unwanted for others. `prefers-reduced-motion` already reaches the loop and calms the waves, but a machine that states no preference leaves those readers with nothing to say.

Turning it off costs a reading, not a fact: every reading the margin carries is already written in words in the rail beside it, which is exactly what lets the edge be `aria-hidden` decoration in the first place.

Two things the shape holds:

- **Off unmounts rather than dims.** An unwanted light must not go on paying for a full-window fragment shader on every screen.
- **The linger is dropped with it.** The edge outlives the reading that lit it by as long as it takes to fade, and switching the light off is not the work finishing — nothing is left mounted to report going dark, so a linger kept across the flip would mount a canvas over an idle window the next time the light came on.

The preference lives in `app/agent-edge-preference.ts`, a store shaped like `theme.ts`: per browser rather than per seat, because it is a choice about the screen somebody is sitting in front of, and it has to hold before any read of `/me` answers. A refused `localStorage` costs the reader their choice on the next page, never on this one — the flip still takes.

The control is the design system's `Switch` (flipping it IS the write; there is no Save on this surface), and it sits last and to the right because everything above it reports.

## How verified

- `make check-fe` green (DS gates, contract drift, biome, unit, build, composed typecheck) — 8m03s, zero failures.
- `node scripts/fe-uat.mjs` renders all 18 agent stories, `PanelOpen` included, with no render error.
- Panel captured in both themes at the story surface to check the foot: the switch reads correctly against light and dark ground.
- New file `frontend/src/app/agent-edge-preference.test.tsx`, 10 cases: the default lights, off draws nothing at all (wrapper gone, no canvas), turning it back on relights work still in flight, no fade held across the switch, the flip is honoured where the browser refuses to persist it, boot honours a stored `off` and an install that never chose, and — through a mounted rail and margins together — the panel's switch takes the margins off the screen, records `off`, and gives them back.
- `make check` could not complete its backend half in this container: the pre-installed `golangci-lint` is built with go1.25 against modules targeting go1.26, so `lint-modules` fails before the Go lanes run. No Go file is touched by this diff.

- [ ] `make check` is green — the frontend half is; the backend half is blocked by the toolchain mismatch above, on a diff with no Go in it
- [x] `make test-integration` is green (or: this change does not touch tenant data / RLS / the write shape) — frontend only, no server call added
- [x] `make craft-static` reports no new blockers — no Go, no extension, no fixture file in the diff

## AI involvement

Written by Claude Code end to end — the preference store, the switch, the stylesheet, the tests, the changelog entry — from Lars's description of the feature. Reviewed against the design-system catalog before building (the control is the existing `Switch`, not a new one).

## Accountability

https://claude.ai/code/session_01JUfTgGRHZdCoZa2Xe75R6m

---
_Generated by [Claude Code](https://claude.ai/code/session_01JUfTgGRHZdCoZa2Xe75R6m)_

---
## Summary by cubic
Makes the lit window edge a preference. The agent panel's foot now carries a switch for the waves that travel the window rim while a run or mailbox import is in flight; on by default, and off is remembered per browser.

**New Features**
- Off unmounts `AgentEdge` entirely rather than dimming it, so an unwanted surface stops paying for a full-window shader.
- The fade-out linger is dropped with the switch, so nothing stays mounted to report going dark.
- The preference is stored per browser in `localStorage`, beside the theme; a browser that refuses storage still honors the flip on the current page.
- `prefers-reduced-motion` still calms the waves on its own for readers who never open the panel.

**Bug Fixes**
- Fixes the unsaved-draft e2e to navigate to `#/settings/admin/general`, the address the settings screen actually mints; the flat `#/settings/company` falls through to the first visible tab, so the dialog landed on a page nobody navigated to.

<sup>Written for commit 37a9b421267d7b6e8934619c4e7fb69814f7dd69. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/4544?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


joshbruh-lang:
The `uat (AC screens + axe WCAG 2.2 AA + 390px, mocked)` lane was red on `ed47b77`, on two cases in `e2e/unsaved.spec.ts` that this diff does not go near. It is not this PR's failure, and it is now fixed here rather than waited on.

**What was failing.** Both cases time out at `typeAndCloseDialog`, waiting for the organization-name edit button after `page.goto("/#/settings/company")`.

**Why it is not this PR's.** The same two cases fail identically on #4541's own run (job `101393777904`, same file, same lines), and that PR is merged — so the failure arrived on main with it and reaches this branch through the merge commit CI builds. This diff touches the agent rail, the lit edge and its preference; nothing under `settings/`.

**The cause.** `#/settings/company` is the catalog's new flat address, resolved by `settingsrouting.ts`. The settings *screen* is still driven by the register in `settingsnav.tsx`, which knows that page as the admin entry `general` — exactly the gap #4541's own description names ("pages the CATALOG names but the SCREEN does not render yet"). So the flat spelling resolves to no tab, `useVisibleSettingsTabs` falls through to the reader's first visible entry, and the dialog the draft is typed into is on a page nobody navigated to. Reproduced locally: the failing run's page snapshot is Konto.

**The fix ported here** (`37a9b42`) is one line plus the reason beside it — the spec opens `#/settings/admin/general`, the address the product actually mints for that page today. It no-ops the moment the screen moves onto the catalog.

Verified on the merge, locally: those two cases pass, the whole e2e suite is green (281 passed, 55 skipped, 0 failed), and `make check-fe` is green.

---

PR #4549, opened 2026-09-05
feat(worklist): a bounced send names the address that refused it

LarsGradion:
Lars decided **3B** on the fix-address item: name the address, no resend from
the queue.

The card said a message failed, named the person and the subject, and never
said **which** address bounced. Open a contact with three addresses and a rep is
guessing — the row reported a failure and left the fix to guesswork.

**The address comes from `bounce_recipient`**, the column `RecordBounce` writes
with the address the delivery report actually named. Not from the send's
recipient list, and that distinction is the whole PR:

> *"A delivery report names ONE failed recipient, but the bounce stamp landed on
> the whole row: a send with a CC would mark every address on it as the one that
> refused."* — the migration that added the column

**My first version had exactly that bug.** I read `recipients->>0` — the first
address on the send. Against a two-recipient send whose second address bounced,
it named `colleague@turbinenbau.de` as dead while `dana@` was the one that
refused. A rep would have "fixed" a working address and left the broken one
alone. The integration test now seeds the bounced address **second**, so
reverting to that spelling fails by name.

The column is per-lane: a park is about the send as a whole and has no address
to name, so the undelivered lane reads a literal `''` rather than borrowing a
recipient the failure was not about.

| Obligation | Evidence |
|---|---|
| User-visible outcome | `TestTheBounceRowCarriesTheAddress` — the card's line reads `dana@turbinenbau.de — 550 no such user`. |
| Permission/row scope | Unchanged, and no new disclosure: the lane is `o.user_id = $N`, the reader's own outbox, so the address was already theirs. When the person is withheld, `PersonID` is zero and the card draws no subject and no link — the address still shows, correctly, because it is the reader's own send. |
| Failure behavior | `TestABouncedSendNamesTheAddressThatRefusedIt` covers all four states: both, address only, reason only, neither. Neither draws no line rather than an empty one. |
| Idempotency/concurrency | Read path only. |
| Mobile | One existing line gains text; no layout change. |
| Storybook | No new component. |
| Accessibility | No new control. |
| Contract | None. `detail` already exists and its description already names "a bounce reason". |
| Write shape | n/a — read path. No migration: `bounce_recipient` shipped in `1788011515` and nothing read it. |
| Mutation strength | Reverting to `recipients->>0` names the bystander, by name. Dropping the read entirely reports `""`, by name. Both against real Postgres. |
| Full lane | `make check-backend` EXIT=0. All **151** comms integration tests pass — the statement is shared with the undelivered lane, so a column change could have broken it silently. |

**Codex did not review this.** Two runs backgrounded past their window and
returned no output. The recipient-index defect above is the finding I had asked
it to check for and found myself; the shared-statement and disclosure questions
I verified by execution. Flagging it rather than implying a review happened.

The test failed twice before it was right, both usefully: the shared fixture
hardcodes `buyer@example.com` in **two** places (the send's recipient and the
bounce report), so a test built on those would have passed against a read that
returned a constant.

---
## Summary by cubic
The worklist bounce card now names the address that refused the send, so a rep no longer has to guess which of a contact's addresses is dead after a delivery fails.

- Reads the address from `bounce_recipient`, the column `RecordBounce` writes with the address the delivery report actually named — not from the send's recipient list, which would name the first address on a multi-recipient send.
- The integration test seeds the bounced address second, so any read that takes the first recipient fails by name.
- The undelivered lane reads a literal empty string: a park is about the send as a whole and has no address to name.
- The card's supporting line pairs the address with the provider's reason, address first; a send carrying neither draws no line rather than an empty one.
- No resend from the queue; this only names the failed address.

No migration needed: `bounce_recipient` shipped in an earlier migration and nothing read it until now.

<sup>Written for commit c15a57c62ccc533b199d53d9edb74ac4057da150. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/4549?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Bounce notifications now identify the specific recipient address that rejected a message.
  * Bounce details display the recipient, rejection reason, or both when available.
  * Bounce entries without recipient or reason details no longer show empty supporting text.

* **Tests**
  * Added coverage confirming recipient-specific bounce reporting and detail rendering.

### DIFF
--- a/backend/internal/compose/attention/bouncedetail_test.go
+++ b/backend/internal/compose/attention/bouncedetail_test.go
@@ -1,0 +1,71 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+package attention
+
+// Which address is dead, on the row that reports the failure.
+//
+// The card names a person and a subject. A rep opening a contact who carries
+// three addresses cannot tell from that which mailbox refused, so the row
+// reports a failure and leaves the fix to guesswork — the defect this lane was
+// built to remove, one step further along.
+
+import "testing"
+
+func TestABouncedSendNamesTheAddressThatRefusedIt(t *testing.T) {
+	t.Parallel()
+	for name, tc := range map[string]struct {
+		send BouncedSend
+		want string
+	}{
+		"both": {
+			BouncedSend{Recipient: "dana@turbinenbau.de", Reason: "550 no such user"},
+			"dana@turbinenbau.de — 550 no such user",
+		},
+		// The provider said nothing usable. The address alone still tells the
+		// reader which mailbox to fix, which is the move.
+		"address only": {
+			BouncedSend{Recipient: "dana@turbinenbau.de"},
+			"dana@turbinenbau.de",
+		},
+		// A send carrying no recipient — an older row, or one stored without
+		// one. The reason stands alone rather than the card claiming to know
+		// where it was aimed.
+		"reason only": {
+			BouncedSend{Reason: "550 no such user"},
+			"550 no such user",
+		},
+		// Neither: no line at all, so the card draws nothing rather than an
+		// empty supporting row under the subject.
+		"neither": {BouncedSend{}, ""},
+	} {
+		t.Run(name, func(t *testing.T) {
+			if got := bounceDetail(tc.send); got != tc.want {
+				t.Errorf("the card's line reads %q, want %q", got, tc.want)
+			}
+		})
+	}
+}
+
+// And it reaches the row, not just the helper.
+//
+// bounceItem sets Detail only when the line is non-empty, so a helper returning
+// the right string proves nothing about what a reader sees.
+func TestTheBounceRowCarriesTheAddress(t *testing.T) {
+	t.Parallel()
+	item := bounceItem(BouncedSend{
+		Subject: "Retrofit quote", Recipient: "dana@turbinenbau.de",
+		Reason: "550 no such user",
+	})
+	if item.Detail == nil {
+		t.Fatal("the row carries no supporting line at all")
+	}
+	if *item.Detail != "dana@turbinenbau.de — 550 no such user" {
+		t.Errorf("the row's line reads %q, and a reader cannot see which address is dead", *item.Detail)
+	}
+	// And a send with nothing to say still draws no line.
+	bare := bounceItem(BouncedSend{Subject: "Retrofit quote"})
+	if bare.Detail != nil {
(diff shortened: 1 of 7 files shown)


LarsGradion:
`craft-reviewer` found one BLOCKER and three advisory notes. All addressed in `c15a57c`.

**BLOCKER — a dead field with an untrue comment.** I added `Recipient` to `ParkedSend` where nothing writes it and nothing reads it, under a comment saying "the address that refused the send" — of mail that **never left**. Nothing refused a parked send. Removed. The parked lane's honesty was already stated where it belongs: `sendLane.recipientColumn` is empty for it, so the shared statement reads a literal `''` rather than borrowing a recipient the failure was not about.

The reviewer's framing is the useful part: the same field is **honest in one layer and dead weight in another**. In `laneSend` it is load-bearing, because the shared statement needs a stable column count across both lanes. In `attention.ParkedSend`, which the bounce lane does not share, it was pure speculation.

**SQL interpolation — confirmed safe, no change.** `recipientColumn` is a compile-time literal in a package-level struct, exactly like the `reasonColumn`, `atColumn` and `only` fields already interpolated into the same statement. The reviewer also noted `pgx.Identifier.Sanitize` would be *wrong* here: `recipientExpr()` returns an expression, not a bare identifier, so quoting it would produce nonsense.

**Backend composition of the detail line — confirmed correct, no change.** `crm.yaml` documents `detail` as prose "in a language a reader reads — a bounce reason", the provider's reason is already an untranslated string, and no client parses the field. Splitting into a typed field would only be right if something had to read the address back out.

**Comment duplication — trimmed.** The CC-bystander story was told in three places (the migration, the struct field, the test). The test keeps its assertion, which demonstrates it, and drops the retelling.

**One thing the reviewer taught me about my own change.** The empty-address case is not only "an older row": an Art. 17 erasure restriction NULLs `bounce_recipient` (`privacy/erasure_restrict.go:204`). I verified that and named it in the test comment, because a reader told only "older row" would be surprised when a restricted send loses its address.

Not acted on, and I agree it is not this PR's: the file header claims every fragment is a literal, and nothing holds that claim if somebody later adds a runtime-derived lane field. It predates this change and deserves its own gate rather than a fix here.

`make check-backend` EXIT=0 and all comms integration tests pass on the rebased base.

Issue #3468, opened 2026-09-01
The Art. 17 redaction leaves text columns standing that may hold subject data


dungbanguyen:
Falls out of #3163. The column-aware census there fixes the column SET so the next `bounce_recipient` cannot arrive unremarked; it does not judge the columns already standing. This is that judgement, and it is a separate piece of work because several of them are legal calls rather than mechanical ones.

The census now names, per redacted PII table, every text/jsonb column the Art. 17 cascade does **not** clear. Most are plainly not subject data — `kind`, `status`, `direction`, `provider`, `source`. These are the ones that are not obviously fine:

**`lead`**
- `linkedin_url` — the subject's own profile address, uncleared. Of everything on this list this is the one that looks most like a straightforward gap.
- `disqualify_note`, `score_override_reason` — free text a rep writes, about the subject.

**`person`**
- `photo_object_key` — a pointer to the subject's photograph. The bytes may well be purged by the attachment arm; the key naming them is not cleared, and a pointer to a deleted object is still a record that the object existed.
- `photo_origin` — where the photo came from.

**`activity`**
- `capture_label`, `audience_reason`, `restricted_reason` — free text, and the last two are written about why a message is restricted, which can name why it is about this subject.

**`comms_outbound`**
- `bounce_reason` — provider text that frequently quotes the recipient address in full, on a table whose whole point here is that the address copy is scrubbed.
- `from_name` — our side, so probably correct to keep, but it should be a decision rather than an omission.

Each needs the same answer as `bounce_recipient` got: clear it, or record why it stays. The census will accept either — what it will not accept, after #3163, is a new column arriving without one.

Not urgent for the ones that are clearly workspace metadata; `lead.linkedin_url` and `comms_outbound.bounce_reason` are the two worth deciding first.

dungbanguyen:
Picking this up.

PR #4551, opened 2026-09-05
fix(privacy): the satellite sweep declares its error instead of testing nil

joshbruh-lang:
## What

`deleteIdentifyingSatellites` opened with a `var err error` and then guarded its
first statement on a nil it had just declared. The chain reads the same without
it, and one fewer line stands between the reader and the comment that says why
every satellite has to be handled in this file.

One file, +2/−5. Nothing else.

### What this PR no longer carries

It opened as the craft-gate fix: `make craft-static` was red on `main` with two
`long-func` findings, `craftsmanship` sits in the `ci` rollup's `needs`, and so
every open pull request was red with it. That is no longer true — `main` at
`0a83bf599` reports `PASS — 0 blocker, 0 major, 0 minor`, and it got there by its
own route rather than this one. Rebased onto that `main`, each of this branch's
remaining pieces resolved to the answer already standing:

- **`AssembleFor` / `measureWeek`** — `main` landed its own `measureWeek`, with a
  wider signature that also folds in the deals read, the prior-review lookup and
  `LearningsState`. Keeping this branch's narrower spelling would define the
  function twice.
- **`newServer`** — `main` moved the forecast and analytics-share wiring out to
  `serverassembly.go`, which is a different answer to the same ceiling and is
  already merged. `newServer` is under 80 lines without this branch's helpers, so
  extracting a lone `newAssuranceHandlers` would add an abstraction with one
  caller and a second spelling of a seam that already exists.
- **`listActivitiesFilter`, `anonymizePersonRecord`, the migration-citation
  register, the send-surface census, `unsaved.spec.ts`** — dropped earlier for
  the same reason, each in a merge rather than as a competing spelling.

What survives is the cleanup above, which stands on its own.

## Why

The recurring defect behind the original findings is worth naming even though
this PR no longer fixes any of them: the pre-push hook runs `craft static`
**diff-scoped**, while CI runs it **whole-tree**. A change that is clean in its
own diff can push a function that was already at the ceiling over it, and the
author cannot see that locally. `AssembleFor` crossed the ceiling twice in one
night that way, at 84 lines and again at 92, from two separate PRs each adding a
few lines. Closing that gap is its own change.

## How verified

On this branch at `3c924ff7e`, rebased onto `main` at `0a83bf599`:

- `go run -C cli/craft . static --strict --root ../../backend`: `PASS — 0 blocker,
  0 major, 0 minor`. The same command on a clean `main` worktree also passes,
  which is what retired this PR's original purpose.
- `go test ./internal/modules/privacy/... ./gates/`: both packages `ok`.
- `go build ./...`, `go vet ./internal/modules/privacy/...`, `gofmt -l`: clean.

The `DELETE` statements themselves are untouched, so the gate that reads this
file's SQL literals to prove every satellite is handled sees the same corpus.

- [x] `make check` is green
- [x] `make test-integration` is green (or: this change does not touch tenant data / RLS / the write shape) — no statement, argument or order changes; only the declaration of `err`

## AI involvement

AI-generated in full: the cleanup, the rebase and the conflict resolutions, and
this description. Each conflict was resolved toward the implementation already
on `main` rather than toward this branch's, so the tree keeps one spelling of
each capability.

## Accountability

https://claude.ai/code/session_01XaNzMSm8r95MiJ5KZfVwcv

### DIFF
--- a/backend/internal/modules/privacy/retentionactions.go
+++ b/backend/internal/modules/privacy/retentionactions.go
@@ -377,13 +377,10 @@ func anonymizePersonRecord(ctx context.Context, tx pgx.Tx, id ids.UUID) error {
 // reads this FILE's SQL literals to prove every satellite is handled, so a
 // helper elsewhere or a loop over identifiers is invisible to it.
 func deleteIdentifyingSatellites(ctx context.Context, tx pgx.Tx, id ids.UUID) error {
-	var err error
 	// The anonymize UPDATES the person row rather than deleting it, so none of
 	// these cascades — one skipped leaves the subject readable beside an
 	// "Erased Subject" record.
-	if err == nil {
-		_, err = tx.Exec(ctx, `DELETE FROM person_social WHERE person_id = $1`, id)
-	}
+	_, err := tx.Exec(ctx, `DELETE FROM person_social WHERE person_id = $1`, id)
 	if err == nil {
 		_, err = tx.Exec(ctx, `DELETE FROM person_email WHERE person_id = $1`, id)
 	}


joshbruh-lang:
**`integration` is red on a failure this PR does not cause, and cannot clear on its own.**

`TestFK_rowScopedTargetsHaveVisibilityDecision` reports four foreign keys on `sdr_handoff` with no entry in `rowScopedFKDecisions`: `deal_id`, `lead_id`, `organization_id`, `person_id`.

Measured rather than assumed: the same test fails identically on a clean `main` worktree at `6772010a5`, with the same four columns. The table, its foreign keys and its writer all arrive with #4568; this PR adds no migration and no writer touching any of them.

This PR exists to clear gates that are red on `main`, so it is a fair question why it does not also clear this one. The map is a classification of how each reference is gated — writing an entry is a statement about the owning module's write path, not a line that makes a gate pass. I could not make that statement truthfully from the outside without guessing, and a wrong one is worse than the red check, because it is the thing the next reader trusts instead of looking. So it belongs with the change that introduced the table, and I have raised the specifics with the maintainer out of band.

The `deterministic-gates` failure on the earlier head `58223d79d` was real and is fixed on the current head: `retentionactions.go` had crossed the 500-line cap.

Green locally on the current head: whole-tree `craft static` (`0 blocker, 0 major, 0 minor`), `go-file-length`, the full `./gates/` package, and the `migrations` integration package apart from the finding above.

---

joshbruh-lang:
A second `integration` failure, distinct from the FK one above and also not this PR's: `./tools/seed-demo`.

The shard log carries its own diagnosis:

```
testdb: rebuilding the schema rather than reusing the clone — 1 table(s) already hold rows
(public.sdr_handoff_reason) — the template this clone came from is not empty
```

`sdr_handoff_reason` is seeded with reference rows by its migration, so the migrated template is no longer empty. `testdb.EnsureSchema` therefore refuses to reuse the clone and rebuilds the schema instead — and the rebuild applies core and custom migrations only, not River's. `river_job` goes with it, and `TestSeedPassesQueueRegisteredJobsAfterTheRecordsExist` fails on `relation "river_job" does not exist`.

Reproduced locally: after `cmd/migrate up` applies River's 7 migrations the table is there, and the next `EnsureSchema` drops it again.

The quieter cost: clone reuse is now off for **every** package in the lane. Each shard's log shows ~508s provisioning slots that used to be `CREATE DATABASE ... TEMPLATE` copies — most of the integration lane's wall clock, on every PR in the repository.

This PR exists to clear gates that are red on `main`, so to be explicit about why it does not clear this one either: the fix belongs with the change that seeds the table — either those rows move out of the migration, or the emptiness probe learns which tables are legitimately seeded. Both are decisions about that table, not about anything here.

No third comment from me on `integration`; these two cover it.

---

dungbanguyen:
Heads-up from stacking #3468 on this branch: `deleteIdentifyingSatellites` opens with

```go
var err error
// …
if err == nil {
	_, err = tx.Exec(ctx, `DELETE FROM person_social WHERE person_id = $1`, id)
}
```

and `govet`'s nilness analyzer calls that first check a tautological condition, so `make check-go` fails on this branch:

```
internal/modules/privacy/retentionactions.go:384:9: nilness: tautological condition: nil == nil (govet)
```

Making the first statement the declaration (`_, err := tx.Exec(...)`) and dropping its wrapper clears it and reads the same. The extraction itself is right — and the comment explaining why the statements stay in this file rather than moving to a helper is the same thing #4590 is about, from the other side.

joshbruh-lang:
Fixed in `26703ab25` — thank you, that was a good catch and I could not have seen it from here.

The reason it is invisible on this branch is worth writing down, because it will bite the next person the same way: `.golangci.strict.yml` is the config that turns `nilness` on (`govet: enable-all`), and it carries `new-from-merge-base: origin/main`. This PR does not touch `retentionactions.go` at all — it takes `main`'s version verbatim after the merge — so those lines are not new, the strict pass has nothing to report, and both `golangci-lint run ./...` and a full `make check-backend` come back clean here. Your stack changes the file, the lines become new, and the finding appears. The code is `main`'s either way.

Applied as you described: the first statement declares `err` and loses its wrapper. The chain below is untouched, and one statement per table stays — `satellite_lifecycle_test.go` reads this file's SQL literals, so a loop over identifiers would be invisible to it, which is what the function's own doc comment says.

Verified: `golangci-lint run --config .golangci.strict.yml ./internal/modules/privacy/...` is `0 issues` with the lines now genuinely new; `go build`, `go vet`, whole-tree `craft static` (`0 blocker, 0 major, 0 minor`), `go-file-length`, and the gates that read this file (satellite lifecycle, person scrub, erasure, table ownership) are all green. The file lands at 498 lines, so it keeps a little room under the 500 cap.

For your stack: this branch is otherwise just the two `long-func` splits that clear `craft-static` on `main` — `AssembleFor` (92 lines) and `newServer` (82). Both are red on a clean `main` right now, so rebasing onto this should clear those for you too.

---

PR #4555, opened 2026-09-06
style(brief,worklist): the Brief opens the day, and a queue row knows its column

joshbruh-lang:
## What

The Brief is drawn at the size of the page it is: the greeting takes the
display rung with the day written out under it, the two dials sit on its own
first line, and the five readings are tiles carrying their figures at that rung
rather than a strip sized for a record column. The page carries a light of its
own and composes itself in reading order on arrival.

A queue row gains a KIND column and a sense of how wide its column is. The kind
of work — `Customer waiting`, `Deal at risk` — moves out of the title line into
a fixed column of its own, and the verbs move out of the title's line entirely:
under the work by default, and beside it in a capped column once the list is
past 60rem. The Worklist's own header puts the day's sentence and the scope on
one line.

Structure is unchanged: the schedule is a rail panel, and the queue is one
continuous feed on both surfaces.

## Why

The Brief is the first page anyone sees after signing in and it read as a
column of counts. The greeting stopped one rung under display so as not to
out-shout five reading panes; the tiles now carry the page's first answer, so
that trade is off. Nothing here invents content — every figure and row is the
one the page already drew.

The row's two changes are measured, not stylistic:

- **The kind was a badge inside the title line**, read only after the title. In
  its own fixed column the kinds can be read down as one list and the titles as
  another, and the titles start at one x whatever the kind says. In the day's
  first band it takes the warn tone, where the kind is also the reason the row
  is there.
- **The verbs were a column beside the work.** A row can carry seven of them,
  and they took the width: the subject, its snippet and its reasons were left a
  strip that wrapped every line. Under the work, the text has the row — but on
  a page-wide queue that left the card's right-hand half empty, measured at
  812px of nothing on a 1358px row. So the row asks the LIST how wide it is, and
  only a list past 60rem spends on both. A media query would read the window,
  and the same queue is drawn in the Brief's work column and on a page of its
  own — two widths on one screen.

`StatStrip` gains a `hero` variant for the first of those, declared in the
catalog: the same readings, sized for a page that opens on them.

## How verified

- `make check-fe` green on the pushed head, run alone: the design-system gates
  (`check-ds-purity`, `check-ds-spacing`, `check-ds-spacing-roles`,
  `check-space-tokens`, `check-font-lock`, `check-icon-glyph`), drift, biome,
  the full unit suite, the build, the composed typecheck and the composed
  workspace screens.
- 60 test files / 551 tests green across `worklist`, `home`, `brief` and
  `record360`.
- Rendered both screens at 1440px in both themes at each step, and measured the
  row's geometry rather than judging it by eye: dead space to the right of a
  page-wide row went from 812px to 32px, and the Brief's column keeps the verbs
  under the work so the subject keeps its full width.
- `make check`'s backend half cannot run in this environment: `golangci-lint`
  there is built against go1.25 while the modules target 1.26, which is the tool
  refusing to load its config rather than a finding. The diff is frontend-only
  and touches no Go.

- [ ] `make check` is green — frontend half green; the backend half did not run, see above

## AI involvement

Written by Claude Code under Jo's direction. Every steer is his: style the
Brief and the Worklist to the design prototype, the readings and the greeting,
"too boring", the task rows needing more room for their content, the dead space
on the right, and finally taking two shapes back off — a day drawn as a
timeline and a stack of cards per task — in favour of the structure `main`
already had. Those two are reverted in this branch from `main`'s own files
rather than rewritten. The CSS, the container query, the kind column, the
`hero` variant and the measurements are AI-authored.

## Accountability

---

---
## Summary by cubic
The Brief now opens at the size of a page — greeting at the display rung, the day written out under it, and the five readings as hero tiles — and worklist rows put the kind in its own fixed column with verbs under the work, or beside it in a capped column once the list passes 60rem. Structure is unchanged: the schedule is a rail panel and the queue one continuous feed.

**Brief**
- The greeting uses the display rung; a new formatter writes the day out in the reader's locale and zone, shown in the morning view only.
- `StatStrip` gains a `hero` variant for the readings, with a hover state that reads as openable; the page composes itself in reading order on arrival, at rest at once under `prefers-reduced-motion`.
- The page's indigo light is now a radial background rather than a positioned box — the old box bled past the edge and pushed the shell's scroller 39px past a 390px viewport, so a phone scrolled sideways.
- The Brief opens on the list, not the deck — the deck hides what is waiting behind a count, so readers land on the list and reach the deck by the switch beside the heading.

**Worklist**
- The kind moves out of a badge into a fixed column between rank and work, taking the warn tone in the day's first band.
- Verbs sit under the work by default; the row measures its own list, not the window, and moves them to a capped fourth column past 60rem.
- The header's day sentence and scope control share one line.

<sup>Written for commit a2ad015d9e52744c904b1def1d4d7214c02e81b6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/4555?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


joshbruh-lang:
`uat` failed on `07c08b2` with five tests. They are two different things.

**Three were mine, and are fixed in `0330183`.**

`§3.8: 390px mobile` — on `#/home`, on the palette over it, and on the agent's panel — all reporting the same thing: `the shell content scroller (.scroll): 39px past the viewport`. The cause was this PR's: the Brief's page light was a positioned box bleeding 10% past the page's right edge, and at 390px that 10% is the 39px. A decoration must never change what the page measures, so it is painted as a background now — the way the shell paints its own two corner glows — with the centre outside the box, so the pool is cut by the edge rather than reaching past it. A background cannot contribute to scroll width.

Verified against the built app rather than reasoned about: the three named tests pass, and so does the whole `390px` block, 37 passed / 1 skipped.

**Two are not this PR's.** `e2e/unsaved.spec.ts:52` and `:76` both time out waiting for `getByRole('button', { name: 'Name der Organisation ändern' })` on `#/settings/company`. The same two, and only those two, fail on `main` — see the `uat (main)` job of the scheduled `main-health` run [33999864438](https://github.com/margince/margince/actions/runs/33999864438), which ran at 00:15 today against `30e088a`. This diff is frontend-only and touches no settings screen. No fix for them exists yet that I could port into this PR; they need somebody who knows what changed on that screen. I have not spent a re-run, since a failure reproduced on the base branch is not a flake.

---

PR #4562, opened 2026-09-06
A clicked file opens over the record instead of leaving for the downloads folder

joshbruh-lang:
## What

Clicking a file card that the browser can draw — a PDF, a photo, a scan — now opens it full-screen over the darkened page, with the filename over it and the verbs a reader of a document has: download, print, close. Everything else keeps the download it always had.

New design-system control `FilePreview` (`filepreview.tsx`), a `placement="full"` lightbox form on the one `Modal`, and `FileChip` wired to open it. The three surfaces that draw file cards — a message's attachment shelf, an account's contract paper, the corpus answer's citations — get it without changing a line, because the chip is what opens it.

## Why

`FileChip` was the end of the road for paper. A reader who clicked a contract on an account got a download: the agreement goes inside whatever application that machine has installed, and the record they were reading comes off the screen. A reader clicking a contract is reading, and the record is still what they are in the middle of.

Three invariants the diff turns on, each stated in the code beside what it binds:

- **The type is decided from the filename, never from the response.** The byte endpoints answer `Content-Disposition: attachment`, so a frame pointed straight at one downloads the file rather than showing it — and that header is not a nuisance to remove, it is what stops an uploaded file from being served as a document under this origin. The preview therefore fetches once and draws from a blob, and a blob URL inherits our origin: a type that can carry script (HTML, and SVG, which is a script host wearing a picture's extension) would run as our own page with our own cookies. `previewMediaType` is the allowlist that decision is made against, and nothing on it can carry one.
- **A PDF goes in a frame and an image in an ``, and Print therefore has two spellings.** A browser fits an image to the window only when the image IS the document, so a scan inside a frame is drawn at its own size and a reader meets an A4 page at 1:1 (measured, not assumed). A PDF's own viewer is better than anything this page could draw, and it prints itself — a page-level print of an embedded viewer puts an empty box on the paper. An image is drawn by this page, so the page prints and the `@media print` rules leave the document alone on the sheet.
- **One provider, above the screens.** Held by `conformance.test.ts` the way `ToastProvider` and `UnsavedGuard` are: a dialog per surface is two `aria-modal` boxes and two Escape handlers the first time two are open together.

What stays a download stays a download: a kind nothing can draw, a click carrying a modifier (a reader asking for a new tab gets one), and a chip with no provider above it — a story, a unit test.

`FileChip`'s two kind lists became one table on the way past. The glyph list and the previewable list are answered from the same key, and the rows where they disagree — HEIC and TIFF are pictures to a reader and to no browser — are exactly the ones a second list gets wrong.

## How verified

- **CI is green on `d684ced47`**, every lane: the three frontend jobs, `uat (AC screens + axe WCAG 2.2 AA + 390px, mocked)`, `live-boot`, `secret-scan`, `craft-residue`, and SonarCloud (0 new issues, 0 security hotspots, 99.0% coverage on new code, 0% duplication). The backend lanes skip on a frontend-only diff.
- `make check-fe` green locally end to end (ds-gates, drift, lint, the vitest suite, build, composed typecheck, unit screens) — re-run after the rebase, on this base.
- New `filepreview.test.tsx`: the file opens over the page rather than navigating; the blob is typed `application/pdf` even when the response calls itself `text/html`; a `.docx` press is not cancelled and opens no dialog; the frame is asked to print itself once it holds a document and Print is absent before that; an image is drawn as an `` and prints through the page; the save rides the card's own href and filename and leaves no link behind; a refused read says so without printing the status; closing revokes the object URL. Two cases added to `filechip.test.tsx` for the allowlist, including that no script-carrying kind is ever offered.
- Stories rendered and screenshotted in **both themes** — ready, refused, and the card that stays a download. `fe-uat` renders all three (`design-system-filepreview--*`) clean.
- One note on the earlier red `uat` run, kept here because the history is on the PR: it was `e2e/unsaved.spec.ts`, broken on `main` by #4541 and already fixed there by #4544. This branch is rebased onto that fix; the two tests pass locally on this head and the lane is green.

- [x] `make check` is green — the frontend half locally and every CI lane on this head; the backend half cannot run in the authoring sandbox (its `golangci-lint` is built with Go 1.25 against a tree targeting 1.26.6), and no Go file is in this diff, so CI's backend lanes skip
- [x] `make test-integration` is green (or: this change does not touch tenant data / RLS / the write shape) — frontend only, no store path
- [x] `make craft-static` reports no new blockers — no Go file changed, so the diff-scoped gate has nothing in scope

## AI involvement

Written by Claude Code end to end under review: the component, the styles, the tests, the stories, the catalog row and the three catalogs. The two design decisions that changed course mid-way were driven by measurement rather than by the first plan — a large image was screenshotted inside a frame to see that browsers do not fit it, and the refused-Print sentence was screenshotted breaking the glyph row before the control became absent instead.

## Accountability

https://claude.ai/code/session_01M2yvVGZC3GEGkGyDkR2VEX


joshbruh-lang:
`uat (AC screens + axe WCAG 2.2 AA + 390px, mocked)` went red on `dd04e46` — two Playwright failures, both in `e2e/unsaved.spec.ts`, timing out on `getByRole('button', { name: 'Name der Organisation ändern' })`.

**Not this PR's.** The spec broke on `main` in #4541, which pointed `typeAndCloseDialog` at the catalog's flat `#/settings/company`. The settings screen is still driven by the register in `settingsnav.tsx`, which knows that page as the admin entry `general`, so the flat spelling falls through to the reader's first visible entry and the dialog is never on the page the test is looking at. This diff touches no settings routing, no e2e spec, and nothing under `src/screens/`.

**Fix ported.** `main` already carries it, in #4544 ("The unsaved-draft e2e opens the page the product actually mints"), which puts the spec back on `#/settings/admin/general`. This branch is rebased onto `6772010a5`, so it now carries that commit — a no-op once the merge ref picks it up anyway, but it means this PR is not waiting on anything to land.

Validated on the rebased head before pushing:

- `make check-fe` green end to end — ds-gates, drift, biome, the vitest suite, build, composed typecheck, and the unit screens.
- `pnpm exec playwright test e2e/unsaved.spec.ts` — the two failing tests, now `2 passed`.

No re-run spent; the push re-triggers the lane on the rebased head.

---

PR #4564, opened 2026-09-06
Main is green again: the new verdict history, twice over

dungbanguyen:
#4550 added `activity_reply_verdict_history` and landed red on two gates. Every PR opened since is red on both. Neither finding is about that change being wrong — both are the gates asking it to say something it had not said yet.

## 1. `TestFK_rowScopedTargetsHaveVisibilityDecision`

```
FK activity_reply_verdict_history.activity_id -> activity has no visibility decision
```

An FK naming a row-scoped record is a read of it, so every such column carries an explicit decision. This one is **gated**, and by the strongest form available:

- both writers reach the history row only *inside the transaction whose `UPDATE` on `activity` already matched*;
- the classifier's write is bounded by the backlog predicate it selected on;
- a human correction takes `auth.EnsureActivityWritable` first;
- and both re-test archived / audience / restricted at write time.

A history row therefore cannot name an activity its writer had not just written, and a correction aimed at a row out of scope answers `apperrors.ErrNotFound` rather than disclosing that the message is there.

## 2. `craft long-func` on `anonymizePersonRecord`

```
239: [MAJOR] long-func — anonymizePersonRecord is 83 code lines (> 80)
```

It crossed the cap because the erasure arm of the same change added a delete to it.

**I tried the extraction craft asks for, and put it back.** Three fitness gates bind these writes to *this file* and *this function name*:

| gate | what it keys on |
|---|---|
| `TestEveryPackageOnlyWritesTablesItOwns` | `package:table:file:function` — every cross-module write ratified under a key naming both |
| `TestEveryPersonSatelliteJoinsEveryLifecyclePathThatApplies` | reads the anonymize path out of this file |
| `TestErasureAndSARReachEveryPIITable` | derives the declared purge set from these statements |

A four-line helper moved writes out from under all three at once and turned one craft finding into **seventeen gate failures**. The waiver records that, so the next person does not repeat the experiment.

The length is not incidental here: it *is* the register — one statement per table, with the judgement that put it there.

## Verification

`make check-backend` exit 0 (gates green, lint 0 issues, craft `PASS — 0 blocker, 0 major, 0 minor`); `backend/migrations` integration suite green, including the FK test that was failing.


dungbanguyen:
Superseded. The FK half landed on main independently (#4550's follow-up wrote the same decision for `activity_reply_verdict_history.activity_id`, in a fuller form than mine), and the craft half has since grown a second finding. Reopened as a PR against current main that fixes both remaining long-funcs.

PR #4569, opened 2026-09-06
fix(main): two functions drop back under the length cap

dungbanguyen:
`make craft-static` runs **whole-tree** in CI, so main being over the bar fails the craftsmanship job on every open PR — six of mine among them, none of which touches either file. I checked the open PRs first: none was in flight for these, so this is that fix.

```
backend/internal/compose/weekly/weeklyassemble.go
  63: [MAJOR] long-func — AssembleFor is 84 code lines (> 80)
backend/internal/modules/privacy/retentionactions.go
  239: [MAJOR] long-func — anonymizePersonRecord is 83 code lines (> 80)
```

## The two splits

Both are by **concern**, not by line count — a split that only moved lines would put the cap back a release later.

**`AssembleFor` → `measureWeek`.** The five reads answer *what the week came to*; the function around them answers *which week, whose, and whether it has been written already*. The reads keep their own comments, including the one explaining why the plan is settled before the counts are written.

**`anonymizePersonRecord` → `deleteSubjectBearerLinks`.** Three deletes, one concern: a URL somebody still holds in a mailbox that reaches the record after it has been anonymized — the opt-in token, the confirm-details link, and what came back through it. Each keeps the paragraph saying why it cannot be left.

The three `crossStoreWrites` waivers move with the statements: that key names the function that writes, and the gate says so explicitly ("A waiver a sibling write in the same package already holds does not cover this one").

## Check

`make craft-static` is `PASS — 0 blocker, 0 major, 0 minor` on this branch, and `make check-go` is green. No behaviour changes: every statement, in order, with its comment.

## Summary by CodeRabbit

- **Refactor**
  - Streamlined weekly reporting data assembly while preserving existing results and error handling.
  - Consolidated privacy retention cleanup operations without changing their behavior.
  - Aligned internal retention waiver identifiers with the corresponding cleanup actions.
- **Tests**
  - Updated coverage to reflect the revised retention cleanup identifiers.


dungbanguyen:
Superseded by #4584, which fixes the same two long functions as part of the combined main-green branch. Closing rather than racing it.

One thing from this branch worth carrying, whichever way #4584 settles `anonymizePersonRecord`: I took the extraction and it did fire those gates — and one of them fires for the wrong reason. `TestEveryPersonSatelliteJoinsEveryLifecyclePathThatApplies` keys a lifecycle path to a single FILENAME, so moving a statement between files in the same package reads as the satellite no longer being handled. An obligation is a property of the code that discharges it, not of the file it sits in; a `files []string` on `satellitePath` makes the gate read the whole path and stops it failing the next time somebody splits a long function.

I have that change working (the gate is green with the deletes in `subjectlinks.go`), but it is not worth holding main for. Filing it as a follow-up.

PR #4573, opened 2026-09-06
craft-static is green on main again: two long functions

dungbanguyen:
```
backend/internal/compose/weekly/weeklyassemble.go
  63: [MAJOR] long-func — AssembleFor is 84 code lines (> 80)
backend/internal/modules/privacy/retentionactions.go
  239: [MAJOR] long-func — anonymizePersonRecord is 86 code lines (> 80)
BLOCK — 0 blocker, 2 major, 0 minor
```

Both are recent arrivals crossing the cap, and every PR opened since is red on them. **They want opposite treatments, which is the point.**

## `AssembleFor` splits

Everything it does before the insert is a *read*: the tallies, the money, the plan's outcome, the deals, the week this one is compared against. `measureWeek` is exactly that set — which is why it can run before the insert has decided whether this rep's week is already taken.

What stays behind is the transaction's judgement: which week closed, which instants bound it, who won the insert, and what may only be frozen onto the row that won.

## `anonymizePersonRecord` does not — and the waiver says why

I tried the extraction first. Three fitness gates bind these writes to *this file* and *this function name*:

| gate | keys on |
|---|---|
| `TestEveryPackageOnlyWritesTablesItOwns` | `package:table:file:function` — each cross-module write ratified under a key naming both |
| `TestEveryPersonSatelliteJoinsEveryLifecyclePathThatApplies` | reads the anonymize path out of this file |
| `TestErasureAndSARReachEveryPIITable` | derives the declared purge set from these statements |

A four-line helper moved writes out from under all three at once and turned one craft finding into **seventeen gate failures**. The length there is not incidental — it *is* the register, one statement per table with the judgement that put it there. The waiver records the experiment rather than leaving the next person to repeat it.

## Verification

`make check-backend` exit 0 — craft `PASS — 0 blocker, 0 major, 0 minor`, lint 0 issues, gates green. `internal/compose/weekly` and `internal/modules/privacy` integration suites pass.

Supersedes #4564, whose FK half landed on main independently.

### DIFF
--- a/backend/internal/compose/weekly/weeklyassemble.go
+++ b/backend/internal/compose/weekly/weeklyassemble.go
@@ -94,49 +94,7 @@ func (e *Engine) AssembleFor(ctx context.Context, now time.Time) (Review, bool, 
 		}
 
 		review = Review{UserID: userID, LocalWeekStart: weekStart, AsOf: now.UTC()}
-		if review.Counts, err = countWeek(ctx, tx, userID, start, end); err != nil {
-			return err
-		}
-		// Leads and meetings read separately from the tallies above: different
-		// tables, different scope clauses, and each dated by a rule that takes
-		// a paragraph to justify. They fold onto the same Counts because a
-		// reader of a week wants one set of figures.
-		c := &review.Counts
-		if c.LeadsRouted, c.LeadsAnsweredInTarget, c.LeadsBreached, err =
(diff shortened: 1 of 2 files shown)


dungbanguyen:
Folded into the combined branch. These fixes block each other — a branch fixing one is red on the others — so they land as one PR.

PR #4578, opened 2026-09-06
A handoff cannot be filed against a record its author cannot open

dungbanguyen:
The schema fitness gate has blocked main since #4568 — every PR opened since is red on it:

```
FK sdr_handoff.lead_id         -> lead         has no visibility decision
FK sdr_handoff.person_id       -> person       has no visibility decision
FK sdr_handoff.organization_id -> organization has no visibility decision
FK sdr_handoff.deal_id         -> deal         has no visibility decision
```

It is right, and **the answer was not a classification.**

## The gap

`SubmitHandoff` takes `lead_id`, `person_id` and `organization_id` from the request body and inserts them behind `auth.Require(sdr_handoff, create)` alone. `DecideHandoff` takes `deal_id` the same way.

`auth.Require` answers *"may this role hand prospects on at all"* — a different question from *"may this seat hand on **this** one"*. That is the same distinction `ensureHandoffSubjectWritable` already draws one function below, for the decision but not for the submission that created it.

So today:

- a seat holding `sdr_handoff:create` can name a lead outside its scope and **learn from the outcome that the id exists**;
- the handoff it wrote points at a prospect its own author cannot open, so every later read of that lead answers about somebody they could not have found;
- `deal_id` is worse in kind — it is the anchor the held-meeting conversion reads, so an ungated one lets that conversion follow a link nobody had scope for.

## The fix

Each caller-supplied reference takes `auth.EnsureLinkTarget` in the transaction that is about to write it — probe, then insert, holding the target `FOR SHARE`, which is what that probe is for. `ErrNotFound` for an id out of scope: the same answer an id that does not exist gets, so the refusal discloses nothing.

## The tests archive their target rather than inventing an id

A missing id proves only that the database refused, and it would refuse with or without the probe. An **archived** row satisfies the FK and is refused by nothing else — so only the probe can produce the failure.

| mutation | result |
|---|---|
| drop the submit probe | `err = <nil>, want ErrNotFound` — the handoff lands on a lead out of scope |
| drop the deal probe | `err = <nil>, want ErrNotFound` — the acceptance anchors to a deal out of scope |

The four map entries record the decision as **gated**, naming the probe.

## Verification

`make check-backend` exit 0. `backend/migrations` green (including the FK gate) and the full `internal/modules/people` suite green.


dungbanguyen:
Folded into the combined branch. These fixes block each other — a branch fixing one is red on the others — so they land as one PR.

PR #4584, opened 2026-09-06
Main is green: eleven blockages, one branch

dungbanguyen:
Eleven separate findings have been blocking every PR opened in the last few hours, and they block each other — a branch fixing any one is red on the other three, so they land together.

Six commits, each reviewable on its own.

---

## 1. `craft-static`: two long functions

```
weeklyassemble.go:63    AssembleFor is 84 code lines (> 80)
retentionactions.go:239 anonymizePersonRecord is 86 code lines (> 80)
```

**They want opposite treatments.**

`AssembleFor` **splits**. Everything it does before the insert is a read — the tallies, the money, the plan's outcome, the deals, the week this one is compared against. `measureWeek` is exactly that set, which is why it can run before the insert has decided whether this rep's week is already taken. What stays behind is the transaction's judgement.

`anonymizePersonRecord` **does not**, and the waiver says why. I tried the extraction first. Three fitness gates bind these writes to *this file* and *this function name* — `TestEveryPackageOnlyWritesTablesItOwns` (key: `package:table:file:function`), `TestEveryPersonSatelliteJoinsEveryLifecyclePathThatApplies`, `TestErasureAndSARReachEveryPIITable`. A four-line helper moved writes out from under all three at once and turned one craft finding into **seventeen gate failures**. The length there *is* the register.

---

## 2. `sdr_handoff`: four ungated references (a real gap, not a classification)

```
FK sdr_handoff.{lead_id,person_id,organization_id,deal_id} has no visibility decision
```

`SubmitHandoff` takes `lead_id`, `person_id` and `organization_id` from the request body behind `auth.Require(sdr_handoff, create)` alone; `DecideHandoff` takes `deal_id` the same way. `auth.Require` answers *"may this role hand prospects on at all"* — a different question from *"may this seat hand on **this** one"*, the distinction `ensureHandoffSubjectWritable` already draws one function below for the decision but not for the submission.

So a seat holding `sdr_handoff:create` could name a lead outside its scope and **learn from the outcome that the id exists**, and the handoff it wrote would point at a prospect its own author cannot open. `deal_id` is worse in kind — it is the anchor the held-meeting conversion reads.

Each now takes `auth.EnsureLinkTarget` in the transaction about to write it. The tests **archive** their target rather than inventing an id: a missing id would be refused by the FK alone, an archived row only by the probe.

| mutation | result |
|---|---|
| drop the submit probe | `err = <nil>, want ErrNotFound` |
| drop the deal probe | `err = <nil>, want ErrNotFound` |

---

## 3. The send census counts one composer, because there is one

Both remaining gates are reporting the same event — #4553 consolidating the composers — from different sides.

`TestEverySurfaceThatSendsAsksTheEngineFirst` floored the walk at **two** composers. #4553 *deleted* `PersonComposer` rather than adopting the component, so every record page now opens the shared drawer and exactly one surface posts to a door. The floor comes down to one: a vacuity floor states what the walk must still see, and two is no longer a fact about this tree.

**Its waiver goes with it, and the distinction matters.** A waiver that matched no subject is normally a census going blind. Here the subject was retired *by the fix the waiver was holding a place for* — the person page's composer is gone and its traffic goes through the drawer, which asks the engine. So the entry is deleted with the reason written above the now-empty map, not carried forward against a file that no longer has a composer in it.

`TestEveryCitedMigrationExists` saw the same move: the comment citing migration 0188 travelled from `persondrawers.tsx` into `compose.head.test.tsx` with the code it describes, so the register's line stopped matching and the citation surfaced as new. The number was the wrong evidence anyway — what the sentence needs is that the door has accepted `bcc` since it landed. Restated; stale register line pruned.

---

## Verification

`make check-backend` exit 0 — gates green, lint 0 issues, craft `PASS — 0 blocker, 0 major, 0 minor`. `backend/migrations` and `internal/modules/people` integration suites green. `compose.head.test.tsx` 14/14, biome clean.

Supersedes #4573 and #4578.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved access controls for handoffs, deal decisions, assurance bundles, and related records.
  - Requests that reference records outside the caller’s permitted scope are now rejected without revealing whether those records exist.
  - Preserved existing weekly review and measurement behavior while improving internal reliability.

- **Analytics**
  - Improved setup and availability of forecasting, analytics queries, context, and sharing capabilities.

- **Consent**
  - Consent records now consistently retain the policy wording presented during approval.

---

## 4. Two harness omissions

`sdr_handoff_reason` is a **boot-seeded reference table** — the migration ships six system reasons so a workspace is usable on day one — and it was not in `PreservedReferenceTables`.

Two consequences, and the loud one is not the important one. The head probe refuses to vouch for a clone whose tables hold rows, so **every integration shard** went red. The quiet one: the reset was **TRUNCATING the seeded catalogue between tests** — the exact thing that exclusion list exists to prevent. An installation always has these six; a lane that empties them tests a state no installation is ever in. It joins the list beside `lead_disqualify_reason`, which is the same shape.

Its fixture follows: with the catalogue preserved, a fixed label collides with the row the previous test left behind, so the seeded reason carries the test's own name — which also keeps the test that *deactivates* a reason from reaching into another test's.

`head_catalog.txt` was missing `weekly_review_scorecard`'s 34 lines (#4579). Regenerated with the flag its own failure message names; **additions only**, and every one of them that table.

---

## 5. Three send fixtures say what the subject was shown

#4583 made a grant **carry real wording or be refused** — the placeholder it replaced produced a row that read like evidence and held none. Three fixtures in the compose suite were granting with nothing, so they fail at the fixture and take four tests with them.

They are about what a *send* does once consent is on record, not about the wording rule — so the wording has to be there, and has to be plainly a fixture's: a sentence that says it is one, rather than a plausible policy line that would read like the real thing in an export.

---

## 6. The four analytics handler sets wire after the literal, like the rest

`newServer` reached 82 lines when #4597 added the assurance surface. The struct literal is where every cross-module edge is declared, so it grows with the product and will cross the cap again — the answer is the seam this file already uses for exactly that: `wireCaptureSettingsSurface`, `wireExportSurface`, `wireOnboardingSurface`, `wireSystemOfRecordReads`, all after the literal.

The forecast and the three analytics surfaces move **together**, because they are one seam and not four adjacent ones: the share routes run in the *forecast* store's transaction, whose `InTx` gates on `forecast:read`, so the whole surface — issuing included — sits behind the grant that reads the thing being shared. Splitting them across the literal and the wiring method would put half of that argument out of sight of the other half.

---

## 7. The catalog records the plan's risks and its room too

#4604 added `weekly_plan.risks` and `.capacity_note` with their length CHECKs and did not regenerate the golden — the same omission this branch already carries for `weekly_review_scorecard`, one migration later. Regenerated with the flag the failure message names: four lines, additions only.

*(The send-census commit was dropped in rebase — main landed the same fix independently while this was open.)*

---

## 8. Two more references say how they were arrived at

Both landed unclassified and block the FK gate. Both are **server-derived**, with nothing a caller could point elsewhere:

- `privacy_notice_case.person_id` — opened by the `person.created` consumer, in the transaction that created the acquisition the duty is owed for. The person is the one the *event* names, and the acquisitions the due date is computed from are read from that same person's own rows.
- `assurance_task_item.task_activity_id` — minted by the pass and then named. The module owns no activity and reaches for none, which is why `BundleInput` carries an id rather than resolving one. There is no production caller yet, so the entry also records the obligation on the pass that lands it: keep minting, do not start accepting an id.

## 9. The anonymize waiver came back out

Main extracted `deleteIdentifyingSatellites` from `anonymizePersonRecord` while this was open — taking it under the cap on its own, and *by moving writes*, so the three gates keyed on (file, function) were repointed as part of that change. The waiver goes with the finding it answered.

It had no headroom left in any case: `retentionactions.go` sits at 499 lines, and a two-line directive put the **merge** over the 500-LOC cap even though neither side was over it alone.


dungbanguyen:
## I force-pushed over this branch by mistake — here is how to get it back

**The pre-clobber head is `6b46a7f8e`** ("The anonymize waiver is not needed after all"). It is intact in the remote's reflog and in my local worktree. The branch currently points at `bf99b367c`, which is mine.

**What happened.** I rebased this branch to unblock it (it was `DIRTY`, so no CI could run at all) and pushed with `--force-with-lease`. On the second push I computed the lease as `--force-with-lease=fix-main-green:$(git rev-parse origin/fix-main-green)` immediately after a `git fetch` — so the "expected" value was whatever had just arrived, including your new commit. That is a lease that can never fail. It is only protective when pinned to the SHA you actually started from.

**What I overwrote was better than what I pushed**, on two counts:

- Your head was rebased onto a **newer main** (it carried #4613, #4620, #4619); mine was several commits behind.
- Your `The anonymize waiver is not needed after all` is the correct fix and mine was not. I had trimmed a comment line to squeeze `retentionactions.go` under the 500-line file ceiling. You spotted that main had since extracted `deleteIdentifyingSatellites` from `anonymizePersonRecord`, so the function is under the cap on its own and the waiver was answering a finding that no longer exists. Confirmed on today's main: `anonymizePersonRecord` no longer appears in `craft-static` at all.

**To restore:**

```
git push --force-with-lease=fix-main-green:bf99b367c origin 6b46a7f8e:fix-main-green
```

The lease is pinned to my clobbering commit, so it fails if anything else has landed since.

## What has moved underneath this branch since

Measured on a clean checkout of `28760e3b9`, so the rebase target has shifted:

- **Already fixed on main** — the two stale registers, and the `sdr_handoff` FK gating is no longer the only FK gap.
- **Still red and still yours**: `craft-static` on `newServer` (82 lines) and `AssembleFor` (**95** lines now, up from the 84 this branch was written against — the split needs redoing against the version main grew); `testdb`'s `sdr_handoff_reason` missing from `PreservedReferenceTables`; the consent-wording fixtures.
- **Taken off your plate**: I opened https://github.com/margince/margince/pull/4634 with the FK decisions and the `head_catalog` regeneration, including your `sdr_handoff` commit cherry-picked (attribution intact). It also adds two references that arrived after this branch: `assurance_task_item.task_activity_id` and `privacy_notice_case.person_id`.

One correction worth having if you pick the assurance one up: `auth.EnsureLinkTarget` **refuses** `activity` — "not a row-scoped table" — because an activity's scope is inherited rather than carried on the row. The probe is `auth.EnsureActivityVisibleLive`.

Sorry for the disruption.

PR #4585, opened 2026-09-06
fix(main): two registers still name the composer that was consolidated away

dungbanguyen:
`deterministic-gates` and `extension-reference` fail on `origin/main`, and therefore on every open PR — mine and everyone else's. Both failures are registers left pointing at `screens/persondrawers.tsx`, the surface #4553 consolidated into the one mail drawer. I checked the open PRs first: none was in flight for either.

```
TestEverySurfaceThatSendsAsksTheEngineFirst
  found 1 frontend file(s) posting to a send door, want at least the two composers
  waiver screens/persondrawers.tsx matched no subject

TestEveryCitedMigrationExists
  frontend/src/screens/compose.head.test.tsx:28 cites migration 0188
  danglingcitations.txt records "frontend/src/screens/persondrawers.tsx migration 0188",
  which is no longer a dangling citation
```

## The send census

The waiver existed because the person drawer's composer led with the person's own purpose-keyed consent guard, so adopting the shared `SendPermission` there meant retiring that guard first. That drawer is gone; the composer every record now opens asks the engine like every other caller. The waiver goes with its subject.

The floor moves from two composers to one, and **stays** — the failure it guards is the walk finding *nothing*, which is a census reporting a pass while seeing no subject. The comment now names why the number moved rather than leaving a lowered number for the next reader to wonder about.

## The citation

The comment moved file with the test. Rather than re-pointing it at a version that is not in the tree — which the gate rightly refuses to launder — it states the invariant instead, which is what the gate's own message asks for: the send door has taken `bcc` and `attachment_ids` for as long as either has existed. The stale register line is pruned with `-update-citations`.

`make check-backend` green; `compose.head.test.tsx` passes (14).

### DIFF
--- a/backend/danglingcitations.txt
+++ b/backend/danglingcitations.txt
@@ -246,7 +246,6 @@ frontend/src/screens/companyraildetails.tsx core 0203
 frontend/src/screens/embedreindex.stories.tsx migration 0115
 frontend/src/screens/embedreindex.test.tsx migration 0115
 frontend/src/screens/embedreindex.tsx migration 0115
-frontend/src/screens/persondrawers.tsx migration 0188
 frontend/src/screens/relationships.test.tsx migration 0007
 frontend/src/screens/relationships.tsx migration 0007
 frontend/src/screens/settings-nav.test.tsx migration 0261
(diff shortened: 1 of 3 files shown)


dungbanguyen:
Superseded by #4584, which carries the same two register fixes — the send census waiver and the citation — inside the combined main-green branch. Closing rather than racing it.

PR #4586, opened 2026-09-06
fix(brief): the page's light is the room's, not the column's

joshbruh-lang:
## What

The Brief's page light no longer stops at a straight edge partway across the
page. It is painted on a viewport-sized layer instead of as a background on the
page's content column.

## Why

A background is painted inside its own box. The light was a background on
`.wrap`, so the pool met that column's top and right edges at full strength and
simply stopped — a hard seam across the page, which is what a light never has.
The comment shipped alongside it even described the mechanism, as a virtue:
"the pool is cut by the edge rather than reaching past it". That reasoning was
about not widening the page, and it was right about the danger and wrong about
the remedy.

The room is the viewport, so the layer is `position: fixed; inset: 0` — the
same size the shell's own two corner glows are painted at. `fixed` is also what
keeps this from repeating the defect the background was reaching for: a fixed
box is laid out against the viewport and contributes nothing to any scroller's
width, where the absolutely positioned box this began as bled 10% past the
column and pushed the shell's content scroller 39px past a 390px viewport.
Behind the content and `pointer-events: none`, because atmosphere that can be
clicked is a control nobody meant to draw.

## How verified

- Rendered `#/home` in dark at 1440px and at 390px: the seam is gone and the
  pool bleeds off the screen's corner.
- Measured rather than eyeballed, which is what the previous fix was for: the
  document's and the shell content scroller's overflow are both empty at 1440px
  and at 390px.
- `make check-fe` green (ds-gates, drift, biome, unit, build, composed
  typecheck, composed workspace screens).
- The `390px` acceptance block, built and run: 31 passed, 1 skipped, and six
  `#/settings/*` sweeps failing on "the address fell back" with
  `ECONNREFUSED 127.0.0.1:18080` from the preview server. Those six fail
  identically on a clean checkout of this same `main` with this diff stashed,
  so they are this environment's missing API rather than this change; the three
  §3.8 criteria this diff is actually about — `#/home`, the palette and the
  agent's panel — pass.

- [ ] `make check` is green — frontend half green; the backend half does not run in this environment (`golangci-lint` built against go1.25, modules target 1.26)

## AI involvement

Jo spotted the seam on the running Brief and reported it. The diagnosis, the
fix and the verification are AI-authored — including the previous change that
caused it, which is the one being corrected here.

## Accountability

---

---
## Summary by cubic
Fixes the Brief page's corner light so it bleeds off the screen's edge instead of stopping in a hard seam where the content column ends; it's now painted on a viewport-sized `fixed` layer behind the content. Also fixes the settings e2e sweep, which was measuring pages' old settings addresses redirected to rather than the pages each test named, and the worklist reply fixture's band name so the typecheck passes.

**Verification**
- Measured at 1440px and 390px: no overflow on the document or the shell content scroller.
- The light layer has `pointer-events: none`, so it can't intercept clicks.
- Extended the e2e grants (`ai_routing`, `audit_log`, `system_reset`) and added an `/ai/routing` fixture so the models and audit pages resolve for the fixture principal.
- The reply fixture used `band: "answer_people"`, which exists nowhere in the contract; it now uses `now`, the classifier's name for a live customer wait.

<sup>Written for commit 5176961393b0a1cd5d505fb43fdad48ef71d24c5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/4586?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


joshbruh-lang:
`uat (AC screens + axe WCAG 2.2 AA + 390px, mocked)` failed here with 28 cases, and none of them is this PR's.

Proof rather than assertion: this branch's whole diff is one CSS file. I checked `frontend/src/screens/home.css` back out at `origin/main`, rebuilt, and got the same 28 failures — 25 settings sweeps plus `AC-settings-16`, `AC-automations-1/2` and `unsaved.spec.ts`. They come from the settings catalog being reshaped: six page ids were renamed, so addresses like `#/settings/general` now redirect, and each sweep measured the page it landed on while reporting the name it asked for. The crumb assertion in `expectSettingsViewLanded` is what caught it.

No fix existed for it, so I opened one — https://github.com/margince/margince/pull/4596 — and have ported it here as `6f2d93a1`. It no-ops as soon as that PR lands on `main`. Full Playwright suite on this head: 289 passed, 55 skipped, 0 failed. `make check-fe` green.

No re-run spent: rebuilding with this PR's diff removed is a stronger answer than a second identical run.

---

joshbruh-lang:
`fe-bundle`, `fe-quality` and `live-boot` all failed on `136a2e0be`. One cause, not three:

```
src/screens/worklist.reply.test.tsx(25,5): error TS2322: Type '"answer_people"' is not
assignable to type '"now" | "review" | "build_pipeline" | "keep_momentum" | undefined'.
```

`live-boot` looks like a different lane, but its last step is the frontend production build — it brought up the stack, seeded it and passed all six verifications before dying on the same line.

Not this PR's: the diff here is one CSS file, and the failing file arrived on `main` with the reply-verb arc. `answer_people` is not a band anywhere — the enum that arc widened is the action list, which gained `reply` — so it is a fixture typo rather than a missed contract addition.

Fix opened as https://github.com/margince/margince/pull/4614 and ported here in `be742f68`. `tsc -b` clean, `make check-fe` green (12m29s).

No re-run spent: the error is deterministic and a re-run on the same commit would reproduce it exactly.

---

PR #4591, opened 2026-09-06
feat(capture): one message is one activity, whichever connector read it

LarsGradion:
## What was wrong

The same email reaching a workspace through two mail connectors landed as **two timeline rows**. `mailmap.ToRecord` keyed the activity's identity on the adapter's name, so `("gmail", <message-id>)` and `("imap", <message-id>)` were two different keys and `uq_activity_source` saw two different messages.

The function's own doc comment claimed the opposite — that a message read over a different transport was "still deduped on (name, Message-ID)" — which is how the defect survived being read. `raw_capture`, the attachment key and the `engagement.reply` idempotency key all inherited the same shape, so one message also banked two originals, two attachment sets and two reply facts.

The send path had the matching hole: the sent row keyed on the *sending* provider, so a Gmail send echoed back over IMAP never collapsed onto it.

## What changed

Mail keys on `connector.EmailSourceSystem` — one identity for every transport carrying an RFC822 message, received or sent. The transport is not lost, it stops being part of the **identity**: it stays on `source`, `captured_by` and `capture_connection.provider`, which is where the readers that actually ask "which mailbox" already look.

Identity is exactly equality of the parsed Message-ID, which the mail parser already requires. Nothing is fuzzy-matched: two different Message-IDs stay two activities however alike they read.

**The rule is enforced, not documented.** `admitMailIdentity` refuses an email record keyed on an adapter's name at the sink door, before any write. A comment could not hold this rule — the tree just proved that.

Three readers took the key for a transport and are corrected:

- **`participantreplay`** matched `capture_connection.provider` against the activity's key, which now matches nothing — every mail row would have been marked "no owner" forever. It resolves the mailbox from the authenticated `captured_by` stamp instead, joining `(provider, user_id)`, which the schema makes unique. That also fixes a case it used to decline: a workspace with two Gmail mailboxes resolved no owner at all. Its parser dispatch reads the **connector**, not the identity, so the demo's JSON payloads stay away from the RFC822 parser.
- **`traceConnector`** answered from the key and would have labelled every mailbox `email` in the capture screen. It reads the authenticated principal, so two connectors on one message still label their own traces.
- **The public create mapper** refuses a client-supplied `source_system: email` on *every* kind — the replay key spans kinds, so a planted `note` under a Message-ID suppresses the real mail just as well.

## Accepted consequences

Stated in code where the old rule was documented:

- The **first** connector to deliver a message supplies the stored original and its attachments. Equal Message-IDs do not promise equal bytes (delivery headers differ per mailbox, a Bcc survives only on the sender's copy), so this is one provider's rendering.
- One seat syncing one mailbox over two connectors records the first connector's trace **per stage and outcome**; a different stage or outcome still writes its own row.

## Not retroactive

Existing rows keyed `gmail`/`imap` do not deduplicate themselves. This is a fresh-database change — resyncing onto old rows would create an `email`-keyed row beside them.

## Tests

`backend/internal/compose/emaildedup_integration_test.go` drives real RFC822 bytes through the **production mapper and sink** (not hand-built records — the defect lived exactly in that seam):

- One message via gmail↔imap, graph↔gmail, imap↔graph, **both arrival orders**: one live activity, one raw original, first reader's provenance kept.
- The original `occurred_at` survives the second connector.
- Two seats whose mailboxes both received the message: one activity, **both** contributions recorded (seat addresses seeded through the real owner-identity writer).
- The sink refuses adapter-keyed mail and writes nothing — no activity, no raw row.

Mutation-checked: reverting the one-line key change fails all four connector orders.

## Gates

`make check-backend`, `make check-fe`, `craft static` (diff-scoped, 22 files), `rls-store-path`, `no-jurisdiction` all pass. Contracts regenerated with `make gen` and `pnpm gen:api`.

⚠️ Two gate failures on `main` are **pre-existing and unrelated** (reproduced with this branch's changes stashed): `TestEveryCitedMigrationExists` and `TestEverySurfaceThatSendsAsksTheEngineFirst` both point at `frontend/src/screens/persondrawers.tsx`, left stale by #4553.

---
## Summary by cubic
Makes one email one timeline activity regardless of which connector (Gmail, IMAP, Graph) read it, with a send echoed back over a different connector folding onto one row. A mailbox-claim corroboration added to close the forgery hole this widening exposed was dropped: the compared sender/time pair is visible to the attacker and genuinely differs between sender and recipient copies, so the hole is filed rather than fixed.

**Bug Fixes**

- The send path and every mail adapter key on `connector.EmailSourceSystem` + Message-ID instead of the adapter's name.
- The sink refuses an email record keyed on an adapter's name before any write.
- `participantreplay` resolves the mailbox from the authenticated `captured_by` stamp, with old stamps lacking a seat falling back to the previous single-connection rule.
- `traceConnector` labels traces from the authenticated connector, not the key.
- The public create mapper refuses client-supplied `source_system: email` on every kind.

**Migration**

- Existing rows keyed `gmail`/`imap` do not deduplicate; this is a fresh-database change.
- The first connector to deliver a message supplies the stored original, its attachments, and the first trace per stage and outcome.

<sup>Written for commit 2794facf89210b1d91d9b3e89134e02dd6e2c301. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/4591?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented duplicate timeline activities when the same email is captured through multiple mail connectors.
  - Preserved the original email’s provenance, timestamp, attachments, and mailbox contributions during deduplication.
  - Rejected email records using connector-specific identities that could create duplicate activities.
  - Ensured sent messages and captured copies share one consistent email identity.
  - Improved replay handling for messages captured from multiple mailboxes.

- **Documentation**
  - Updated API and connector guidance to explain transport-independent email identity and deduplication behavior.


LarsGradion:
## Review round 1 — findings addressed

A Codex review of the first commit raised one BLOCKER and four MAJOR findings. All are handled; two commits follow.

### BLOCKER — a forged Message-ID could buy a content grant

The finding is correct, and it is this change's to close: the weakness pre-exists on `main`, but sharing one identity across connectors widens its reach from one connector to every one.

`capture_import` is read as a **content grant** (`auth.ActivityScopeClause`). The only thing standing between a Message-ID and one of those rows was `mailboxWasARecipientTx`, which proves the *presented* bytes reached the presenter's mailbox — and the presenter wrote those bytes. A seat who learns a Message-ID (the discover scope is deliberately wider than the content scope) could compose a lookalike, mail it to their own connected address, sync it, and be recorded as a contributor to a colleague's correspondence.

The claim is now corroborated against the incumbent: **the sender and the send time must agree with the row already on file**. Both come from the message, so every mailbox that received one message reports the same pair, while a forger is guessing at content they cannot read. Subject and body are deliberately *not* compared — providers rewrite those per mailbox, and requiring them would refuse the real colleagues this path exists to serve. A held row (`restricted_at`) corroborates nothing, so a restriction cannot be probed through it.

Both directions are tested, and the guard is mutation-checked: disabling the corroboration alone records the forger as a contributor.

### MAJOR — legacy `captured_by` broke participant replay

Correct, and a regression I introduced. Rows stamped before provenance carried a seat (`connector:gmail`, no third field) resolved no owner, permanently. Those now fall back to the rule the query used to apply to every row — the provider's single connection — so an old row still replays, while a stamp naming a seat still resolves the two-mailbox case the old rule never could.

### MAJOR — integration fixtures met the new door

Correct. Seven files built email records keyed on an adapter name. They carry the shared key now. Lead and trace fixtures keep the connector name, which is still right for them.

### MAJOR — recipient-first echo collision

Real, and left as-is deliberately. A CC'd colleague's inbound copy can take the stamped identity first; the absorb's `direction = 'outbound'` guard then declines. That guard must stay: `stamped` is parsed out of a remote provider's response, so widening it to inbound would let a hostile answer nominate somebody's *received* mail to be archived. It degrades to one duplicate row plus a `comms_identity_reconcile_failed` breadcrumb — the same outcome this path already had whenever a collision was not provably our echo. Now documented in `messageidentity.go`.

### MAJOR — no migration for pre-existing rows

Out of scope by design, and stated in the PR body: this is a fresh-database change.

### Gates

`make check-backend` is clean apart from the two failures that are **pre-existing on `main`** (`TestEveryCitedMigrationExists`, `TestEverySurfaceThatSendsAsksTheEngineFirst`, both pointing at `persondrawers.tsx` after #4553 — reproduced with this branch stashed). `craft static` diff-scoped: 0 findings. A third gate, `TestEveryReaderOfTheActivityTableExcludesRestrictedRows`, *did* catch the new corroboration query reading the activity table without excluding held rows — fixed rather than waived.

LarsGradion:
## Review round 2 — the guard from round 1 is reverted

The second pass found my round-1 fix wrong twice over, and it was right on both counts. Verified in code before acting:

**It did not close the hole.** The corroboration compared sender and send time against the stored row, on the premise a forger could not know both. They can: `activityprojection.go`'s `record()` nulls `subject`, `body` and `source_id` on a withheld row and leaves **`source`**, which is `connector:<message-id>`. `occurred_at` is projected too. The pair is readable exactly where the attack starts.

**It broke the legitimate case.** `counterparty_email` is the From address for an inbound copy and a *recipient* for the sender's own copy, so the two genuine copies of one message disagree by construction. The guard would have refused the sender-and-recipient pair it was meant to admit, and dropped their postures from audience recomputation.

A second attempt — requiring the seat to be named on the **stored** participants — fails for a related reason: on an inbound message the recipient list is the sender's text, so `StampFurtherParticipants` deliberately does not bind a colleague from it. A genuine second recipient often has no stored row at all.

Both attempts are reverted. I am not guessing a third time.

**The weakness is filed as a private advisory** (per SECURITY.md — it is reproducible, so it does not belong in a public issue). It pre-exists this change on `main`: the replay path gates on the DISCOVER clause while `capture_import` grants CONTENT. Closing it needs evidence capture does not persist today — the provider's own message id, which `messageidentity.go` already names as the missing invariant and which the send receipt already carries as `ProviderMessageID`.

### What this PR now claims

One message is one activity across mail connectors. It does **not** claim to change who can read what: the pre-existing forgery path is untouched, and the advisory tracks it.

### Fixture sweep

The rest of round 2 is the sweep the review asked for. Every reader that looked mail up by an adapter name now asks by the shared identity: capture counts, the scope lookup, the mail-sharing audience probe, attachment keys in `capturedfiles` and `dealfiles`, the erasure seeds, the comms echo-collapse assertion, and the send row's own `source_system`. Lead rows, trace rows and the Telegram contrast keep the connector name.

`sinkinternal_integration_test.go` — the one the review caught that I had missed — is fixed.

### Lanes run

`compose`, `compose/integration`, `compose/integration/capture`, `modules/capture`, `modules/activities`, `modules/comms` all green. `make check-backend` clean apart from the two failures pre-existing on `main` (`persondrawers.tsx`, after #4553). `craft static` diff-scoped over 37 files: 0 findings.

Issue #4594, opened 2026-09-06
The settings UAT sweep is a hand-written list of addresses, and it has gone stale twice


dungbanguyen:
`e2e/ac.spec.ts` sweeps settings by a hand-written list of addresses. The list has now gone stale twice, and the second time it took the `uat` lane on `main` down for 27 failures.

## What the list is

```ts
const CORE_SCREENS = [
  …
  "settings/models",
  "settings/fields",
  "settings/integrations",
  …
];
```

Fourteen settings addresses, written out, feeding four sweeps — 390px overflow, axe in light, axe in dark, and one-h1-per-route. The catalog they are meant to mirror is `SETTINGS_PAGES` in `frontend/src/screens/settingscatalog.ts`, which today holds **27** pages.

## Why a hand-written list is the wrong shape here

The comment above it already says so, and has for a while:

> ALL fifteen. The list below is written out rather than derived, and that is the weakness to know about: it went stale once already, claiming twelve while three pages — capture-activity, knowledge, license — were in neither sweep and nothing failed to say so. **A census that can fall short reports PASS on the smaller tree.**

That is `AGENTS.md`'s rule 8 exactly. The failure mode is under-recognition: the sweep reads a smaller tree, reports PASS, and there is no failing assertion to notice.

The second staleness was worse than silent, which is the only reason it was caught. https://github.com/margince/margince/pull/4575 split the seven groups into real pages, so six of the listed ids became *redirects* to a page under another name — `general`→`company`, `users`→`members`, `data-model`→`fields`, `ai`→`models`, `maintenance`→`system-health`, `license`→`seats`. The sweep went on naming the old ids, `expectSettingsViewLanded` compared the crumb against the *group's* label, and 24 cases failed. Repaired in https://github.com/margince/margince/pull/4590 by correcting the list — which restores green but leaves the shape that will go stale again.

## What deriving it needs

The comment names both blockers. One is gone:

> The blocker named here — that reading the ids would pull the screen's whole module graph through Playwright's transform — is GONE: `settingscatalog.ts` is pure data and imports like `../src/i18n/de` above it already does.

The other is the mock's grants, and https://github.com/margince/margince/pull/4590 moved it but did not finish it. It added `ai_routing`, `audit_log` and `system_reset` because the three pages it had to reach needed them. A derived sweep needs **every** page reachable, and these objects are still ungranted in `E2E_ADMIN_GRANTS`:

| page | `requires` |
|---|---|
[...]

PR #4596, opened 2026-09-06
test(e2e): the settings sweep names the pages that exist

joshbruh-lang:
## What

The `uat` lane goes green again. Twenty-eight e2e cases were failing on every branch: the settings census still named ids the catalog renamed, two pages the fixture principal could not see, and one page that threw on arrival.

## Why

The settings catalog is flat and per-capability now, and six ids were renamed under it — `general` → `company`, `users` → `members`, `data-model` → `fields`, `ai` → `models`, `maintenance` → `system-health`, `license` → `seats`. The old addresses still resolve, so nothing 404'd: they **redirect**, and each sweep measured a page it had not asked for while reporting the name it had. `expectSettingsViewLanded` is the assertion that exists to catch exactly that, and it did.

The invariant is the one the census comment already states and rule 8 of the rulebook restates: a census that can fall short reports PASS on the smaller tree. Three ways it was falling short here, and each is closed:

- **A renamed id is not a page.** The list now names canonical ids only. Automations and the audit trail were each one body of a page already in the list and are pages of their own now, so both are named — a split surface that keeps only its old address is a surface that quietly left the sweep.
- **A page the principal cannot see falls back to Account**, so the sweep measures the shortest page in settings twice while reporting two. `ai_routing:read` opens the models page; `audit_log:read` and `system_reset:delete` together open the audit page, because `AuditLogCard` still stands the reset grant in for the admin role it checks.
- **A page that throws renders almost nothing**, and an axe pass over nothing is clean. `/ai/routing` had no fixture, and the catch-all's `{data,page}` carries neither `tiers` nor `embeddings`, so opening the models page took the whole entry down. It joins the four reads already held here for that same reason, binding two chat tiers and the embedding lane so the widest rows on the page are the ones the 390px and axe passes actually read.

`unsaved.spec.ts` navigated by the legacy `#/settings/admin/general` and then asserted the address had not changed — an address the app rewrites out from under it. It drives the minted spelling now.

## How verified

- `make check-fe` green (8m42s).
- Full Playwright suite on this branch: **289 passed, 55 skipped, 0 failed**. On `main` at `8eeb2037` the same suite fails 28.
- Each of the three failure shapes was reproduced before it was fixed, and the models-page crash was watched turn into rendered lane rows (`ai-routing-tier-cheap_cloud`, `ai-routing-tier-premium`, `ai-routing-embeddings` all present in the axe pass).

- [x] `make check` is green (frontend half; this diff is `frontend/e2e/` only)
- [x] `make test-integration` is green (or: this change does not touch tenant data / RLS / the write shape) — does not touch it
- [x] `make craft-static` reports no new blockers — no Go files in the diff

## AI involvement

Wholly AI-assisted: the diagnosis (reproducing the 28 failures with an unrelated branch's diff removed, then narrowing each shape), the census rewrite, the two grants, the routing fixture and this description were written by Claude Code under human review.

## Accountability

---

---
## Summary by cubic
Fixes the e2e settings sweep so the `uat` lane goes green again: 28 cases were failing because the sweep named settings page IDs the catalog renamed, and old addresses resolve by redirecting, so each case measured a page it hadn't asked for.

- The sweep now lists canonical page IDs only; automations and the audit trail are named as pages of their own.
- Adds `ai_routing:read` for the models page, and `audit_log:read` with `system_reset:delete` for the audit page — without these grants both entries fall back to Account.
- Adds a `/ai/routing` fixture with two chat tiers and the embedding lane so the models page renders instead of crashing.
- `unsaved.spec.ts` now navigates by `#/settings/company` instead of the legacy `#/settings/admin/general` that the app rewrites.

<sup>Written for commit a2c7ad2371f69625cc37a027d7ce81d90fb393c8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/4596?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/e2e/ac.spec.ts
+++ b/frontend/e2e/ac.spec.ts
@@ -101,47 +101,54 @@ const CORE_SCREENS = [
   "partners",
   "analytics",
   "settings",
-  // The automations editor is configuration on the AI settings page now, not a
-  // destination of its own. Sweeping `#/automations` after the route retired
-  // would measure the fallback screen and report it as coverage, so the sweeps
-  // follow the surface to where it actually lives.
-  "settings/ai",
-  // Settings is FIFTEEN pages behind one route (SETTINGS_TABS in
-  // screens/settings.tsx), and a bare `settings` resolves to Account — the
-  // shortest of them. Sweeping that alone and calling settings covered is how
-  // the widest page in the product went unmeasured: `data-model` carries two
+  // Settings is many pages behind one route (SETTINGS_PAGES in
+  // screens/settingscatalog.ts), and a bare `settings` resolves to Account —
+  // the shortest of them. Sweeping that alone and calling settings covered is
+  // how the widest page in the product went unmeasured: `fields` carries two
   // full list surfaces with their own toolbars, `integrations` four
-  // installation-wide cards, `people` a roster of rows that each end in two
+  // installation-wide cards, `members` a roster of rows that each end in two
   // buttons. These are where a narrow viewport actually breaks.
   //
-  // ALL fifteen. The list below is written out rather than derived, and that is
-  // the weakness to know about: it went stale once already, claiming twelve
-  // while three pages — capture-activity, knowledge, license — were in neither
-  // sweep and nothing failed to say so. A census that can fall short reports
-  // PASS on the smaller tree.
+  // Every id below is a CANONICAL page id. A renamed spelling still resolves —
+  // `settingsrouting.ts` keeps the old addresses working — but it resolves by
+  // redirecting, so sweeping one measures the page it lands on while reporting
+  // the name it was asked for. expectSettingsViewLanded below is what makes
+  // that visible rather than silently counting the same page twice.
+  //
+  // The list is written out rather than derived, and that is the weakness to
+  // know about: it went stale twice — once claiming twelve while three pages
+  // were in neither sweep, and again when the catalog was split into its
+  // current shape and six of these ids were renamed under it. A census that can
+  // fall short reports PASS on the smaller tree.
   //
   // The blocker named here — that reading the ids would pull the screen's whole
   // module graph through Playwright's transform — is GONE: `settingscatalog.ts`
   // is pure data and imports like `../src/i18n/de` above it already does. What
   // still blocks deriving it is the mock's grants: E2E_ADMIN_GRANTS in seed.ts
-  // predates the thirteen settings objects, so a derived sweep would name pages
-  // this principal cannot open, each of them landing on the fallback and
-  // reporting a page the run never read — the same short census in a new shape.
-  // Extend the grants first, then derive; doing it in that order is the whole
-  // point.
-  "settings/data-model",
+  // predates the settings objects, so a derived sweep would name pages this
+  // principal cannot open, each of them landing on the fallback and reporting a
+  // page the run never read — the same short census in a new shape. Extend the
+  // grants first, then derive; doing it in that order is the whole point.
+  "settings/models",
+  // Automations and the audit trail each used to be one body of a page already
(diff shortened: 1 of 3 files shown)


dungbanguyen:
Heads-up from a parallel session — I hit the same red lane and landed on the same six renames and the same three grants. **Yours is the more complete fix and I have closed my duplicate of it**: I had not run `e2e/unsaved.spec.ts` and it fails too, which your branch repairs and mine did not.

One thing that is not in your diff, so it does not get lost: `#/settings/models` did not merely land elsewhere — it **threw**.

```
TypeError: Cannot convert undefined or null to object
  at Object.values → boundProviders (src/screens/ai-settings.tsx)
```

Serving `/ai/routing` from the mock fixes the lane, and that is the right call for this PR. It does not stop the component throwing if a real server ever answers a 200 without `tiers` — `tiers` and `embeddings` are required of `AiRouting`, but nothing validates a response, so a server too old or a proxy answering something else produces that body. The boundary sits above the shell, so the throw costs the reader the entire settings page, and it scores a clean axe pass while doing it.

That half is https://github.com/margince/margince/pull/4600 — two files, independent of this one, mutation-checked against the unguarded version. No conflict with yours: it touches only `ai-settings.tsx` and its test.

Also filed https://github.com/margince/margince/issues/4594 for the underlying shape — the settings sweep is a hand-written address list that has gone stale twice now, and the remaining fixture grants a derived sweep would need (`authentication_policy`, `tag`, `product`/`offer_template`, `import_run`, and `reset`'s flag) are enumerated there.

joshbruh-lang:
Thanks — and agreed on the split. I read #4600: it touches `ai-settings.tsx` and its test only, so there is no overlap with this diff, and the two halves are the right shape. The fixture makes the lane measure a page that renders; the `null` guard is what stops a real off-contract 200 costing a reader the whole settings page. Neither substitutes for the other, and a fixture that hides a crash is the worse of the two failures to be left holding.

Nothing to port here: this branch is green without #4600, so it can land on its own timing.

#4594 names the right shape. The census in this diff is still hand-written, and I said so in the comment above it rather than claiming otherwise — the grants you enumerated there are what a derived one needs, and the two I added (`ai_routing:read`, `audit_log:read` + `system_reset:delete`) are the same kind of debt paid one page at a time.

---

dungbanguyen:
This is green on every check — including `uat` (the lane it fixes) — and `mergeStateStatus` is `CLEAN`. It has been mergeable for about twenty minutes.

Flagging because `main` has been red continuously since 01:38 and every PR opened since inherits it: this branch is the whole uat half of that, 28 cases. I have verified the same 28 fail on a clean `origin/main` checkout and that they are the ones this fixes.

Not merging it myself — it is your branch and your call whether more is coming. But nothing is blocking it from where I can see.

The backend half is https://github.com/margince/margince/pull/4584, currently `DIRTY` and being rebased, so the two are independent: landing this does not wait on that.

joshbruh-lang:
Nothing more is coming — this branch is finished as it stands, and I am not planning another push to it.

I am not merging it myself: merging is not mine to do here, on this or any PR. So please treat this as the confirmation you asked for and land it whenever you are ready.

One thing worth saying plainly for whoever picks it up: this closes the sweep's stale addresses, not the shape underneath. The census is still a hand-written list, which is what #4594 is about, and #4600 is the guard that keeps an off-contract 200 from costing a reader the settings page. Green here means the lane measures the pages it names again — not that either of those is done.

---

Issue #4493, opened 2026-09-05
Decide: segment is a published dimension, not a second free-text field


LarsGradion:
Last of six contracts the analytics plan's section 6 says must be decided before PR 6b.

## The question

Managers want to split every rate by segment. There is no segment dimension, and the tempting fix is a `deal.segment` string.

## The recommended default — not approved behaviour

> Publish organization `sizeband` and permitted typed custom fields as analytical dimensions through PR 7a. **Do not add a second generic `deal.segment` string.** A fixed "segment" control appears only when a workspace has a configured permitted dimension.

## Why the tempting fix is the wrong one

A free-text `segment` column is unvalidated, unshared between deals of the same company, and answers a different question per workspace — so a cross-workspace report over it means nothing, and a within-workspace one means whatever the last rep typed. The organization already carries `sizeband`; a deal-level string would be a second answer to a question the account already answers.

## What needs deciding

1. **Which typed custom fields may become dimensions.** All of them, or an allowlist per workspace?
2. **Whether a workspace with no configured dimension sees the control at all.** The default hides it, so nobody groups by a field that does not exist.
3. **The privacy floor on a low-cardinality dimension.** Grouping by a segment with three accounts in it is a question about three named customers.

## Depends on

PR 7a (the multi-domain typed compiler), which publishes custom fields to the analytics schema. Related: #4475, where report filters are invisible to that same schema.

LarsGradion:
## Decision (Lars, 2026-09-06)

**Publish existing typed fields as dimensions. No `deal.segment` free-text column.**

Settled:
1. **Source of the dimension** — organization `sizeband` plus permitted typed custom fields, published through PR 7a. A deal-level free-text `segment` is rejected: it would be unvalidated, unshared between deals at the same company, and would mean something different in every workspace, so a cross-workspace report over it would mean nothing at all.
2. **Which custom fields** — an allowlist per workspace, not all typed fields automatically.
3. **Workspaces with nothing configured** — the segment control stays hidden. Nobody groups by a field that does not exist.
4. **Privacy floor** — the existing floor (`analyticsquery.DefaultFloor`, currently 5) applies to segment grouping. Grouping by a segment holding three accounts is a question about three named customers, and must withhold exactly as any other small cohort does.

Depends on PR 7a.

LarsGradion:
## Half delivered, half blocked — and the halves are independent

### The sizeband half needed no PR 7a and is built

[...]

PR #4613, opened 2026-09-06
feat(reports): a filter carries its grant, and a joined attribute its row scope

### LINKED ISSUE #4576: A report dimension read through a join has no row scope and no filter grant
Found building the sizeband half of #4493, and it is why that work was reverted rather than shipped. Two gaps in the report engine, both pre-existing, both invisible until a dimension came from a joined table instead of the base one.

Every dimension in the catalog today is a column on the report's own table, or an id pointing at another. This was the first ATTRIBUTE read across a join, and the machinery does not cover that case.

## 1. A filter bypasses its grant entirely

`runSpec` checks grants for what the caller GROUPED by and for aggregate fields:

    requireVocabularyGrants(ctx, spec, slices.Concat(asked, aggregateFields(req.Aggregates)))

**Filters are not in that list.** So a caller who may not read organizations can send `{"filters":{"size_band":"1001-5000"}}` and read the count back. A differing count discloses that such a company exists — and with no privacy floor on this surface (#4515), one deal is enough to answer it.

This is not specific to `size_band`. Any current `grants` entry is enforced on grouping and unenforced on filtering. `deals-by-stage` records the same disclosure in its own comment — *"a count filtered to one company answers whether that company exists and has a deal"* — and relies on `referenceScopes` rather than `grants` to close it, which works only because its column is an id.

## 2. Row scope does not follow a join

`referenceScopes` maps a column to a row-scoped table, and the clause fires only when `namedByReport` records that EXACT expression. Grouping by a joined attribute records the joined expression (`co.size_band`), never the base column (`t.organization_id`), so the scope clause never runs.

The consequence: a caller with `organization.read` but excluded from a capture-private organization sees that company's band and its count, and a drill-through can return the band beside the deal id.

My reverted commit claimed the size "travels with organization_id, which IS scoped." **That was wrong**, and it is the kind of wrong that reads as reassuring — I have removed the claim rather than leave it in the tree.

## What needs deciding

1. **Should filters take vocabulary grants?** The straightforward fix is adding filter keys to the `requireVocabularyGrants` call. It is a behaviour change for existing callers: a seat filtering by `project_id` today without the project grant would begin getting 403. That may be correct and it is not mine to decide silently.
2. **Should a dimension be able to declare that it scopes THROUGH another column?** Something like `scopeVia: {co.size_band: t.organization_id}`, so naming the attribute activates the id's existing clause. This is the general fix and it makes joined dimensions safe for every future one.
3. **Until then, is a joined attribute dimension allowed at all?** Today the honest answer is no, and nothing says so — the next author will reach for one exactly as I did.

## What is not blocked

The sizeband dimension is small and ready: a LEFT JOIN on `organization.id` (proven not to multiply or drop rows, with a companyless-deal test) plus the field. It ships as soon as either fix lands.

Related: #4515 (no privacy floor on this surface), which is what makes gap 1 a single-row disclosure rather than an aggregate one.

LarsGradion:
Closes #4576. Unblocks the segment dimension work on #4493.

Two holes in the report engine, both pre-existing, and the company-size dimension that could not be added over them.

## A filter is a read

The vocabulary gate refused a **group-by** on a granted field and let a **filter** on the same field through. So `{"filters":{"project_id":"…"}}` answered what `{"group_by":["project_id"]}` was refused for, and the count is the disclosure: a non-zero one says the record exists and has work filed under it.

`runSpec` now passes the caller's filter keys to `requireVocabularyGrants`. They are collected **before** `withThresholdDefaults`, so a threshold the spec supplied for the caller is never refused to the caller who did not ask for it — the same rule the group-by defaults already followed.

Worth stating plainly, because it changes how much this PR is worth: **every granted filter in the catalog today is `project_id`, and every one of them also carries `filterScopes`**, which refuses the *value* before a row is counted. So the hole is currently unreachable and this is defence in depth — the answer for the granted filter that has no row scope to check, which is the spec somebody writes next. An end-to-end test over a shipped report would be green either way, so the assertion is written against the call site instead and says why.

## A joined attribute had no row scope

`referenceScopes` renders `ref.id = <column>`, so it only reaches a column that IS an id. A dimension reading a company's **size band** is not an id, matched no entry, and carried no row scope at all — a seat excluded from an account could group by that account's size and read its win count off the row.

`spec.scopeVia` names the id column such a dimension inherits from, and `namedByReport`/`namedByDerivation` activate the base column's existing clause. Both surfaces, because a drill-through that forgot it would open rows the count above it never counted.

## The dimension itself

`win-loss` gains `size_band` over `LEFT JOIN organization org ON org.id = t.organization_id` — to-one on the company's primary key, so the grain stays one row per deal, and LEFT so a deal with no company stays in the totals rather than vanishing from them.

`TestWinLossVocabularyMatchesItsPinnedShape` asked for the **bare** deal table. That is a stronger claim than the grain needs and the wrong one to hold: it forbids a to-one lookup, which cannot multiply anything, while saying nothing about the to-many join that would. It now holds what matters — every join is LEFT and matches on the joined table's primary key. (The to-one check compares two capture groups in code rather than using a backreference: Go's RE2 has none, and a `\1` compiles to a pattern that panics at init.)

## Verification

Both fixes are mutation-checked:
- Reverting the filter keys fails `TestTheVocabularyGateReadsTheCallersFilterKeys`.
- Reverting `scopeVia` leaks the hidden company's band in `TestGroupingByAJoinedAttributeIsScopedLikeTheIdItHangsOff` — which asserts **both arms**, so a clause that excluded every company would not pass it.

`TestEveryFilterOverAScopedReferenceIsDefended` walks the catalog rather than a list beside it, and fails if it ever walks zero filters. Writing it turned up that `referenceScopes` is itself the third defence for a filter — the deal reports use it deliberately for `organization_id`, which `reportspecs.go` already explains — so the gate accepts any of the three rather than demanding one.

`make check-backend` green after rebase; 188 report integration tests pass.

---
## Summary by cubic
Closes #4576 and unblocks the segment-dimension work in #4493. Fixes two pre-existing report-engine authorization holes, and adds a `size_band` dimension to `win-loss` on top of them.

**Bug Fixes**
- Filters on granted fields now pass through the vocabulary grant check, which previously let a count leak whether a record exists when the matching group-by was refused.
- Filter keys are collected before threshold defaults, so a threshold the spec supplies is never refused to a caller who didn't ask for it.
- The filter hole is not reachable from today's catalog — every granted filter already carries a value scope — so the fix is asserted against the call site.
- A dimension reading a joined attribute (`size_band`, industry) now inherits the row scope of the id column it hangs off, through `scopeVia`, on the report, its drill-through, and the analytics surface.
- A catalog-wide gate now requires any vocabulary entry read over a join onto a row-scoped table to declare where its scope comes from.

**New Features**
- `win-loss` offers `size_band` as a group-by and filter, read through a LEFT join to `organization` on its primary key so the grain stays one row per deal.
- The win-loss shape test now checks that every join is LEFT and matches the joined table's primary key instead of demanding the bare deal table.

<sup>Written for commit 577a97ee0daab12cf5e578fc63bb31cb07c30794. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/4613?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added `size_band` as a grouping and filtering option for win-loss reports and analytics.
  - Added support for displaying organization size bands in win-loss reporting.

- **Bug Fixes**
  - Improved access control when filtering or grouping by joined attributes.
  - Ensured hidden organization attributes remain excluded across reports, analytics, and drill-through views.
  - Prevented users from filtering by fields they are not permitted to read.

- **Documentation**
  - Updated the published analytics schema and win-loss vocabulary to include `size_band`.


LarsGradion:
**Adversarial review found a leak this PR's first commit introduced. Fixed in a6f7281d84.**

Three surfaces build a `referencedColumns` set. The first commit taught two of them about `scopeVia` and missed `namedByAnalyticsQuery` (`analyticsqueryrun.go:184`).

That matters because the analytics grammar derives its vocabulary from the same `reportSpec`s — `AnalyticsSchemaFor` reads `prebuiltReports`, including `fromClause()` and therefore the new join. So adding `size_band` to win-loss published it on `POST /analytics/query` and the `run_analytics_query` tool in the same commit, on a path where the row scope was not applied. The diff's own `mcp-info.json` hunk shows it: `size_band` was added to `margince://schema/analytics`, not only to `schema/reports`.

Two attacks, both confirmed against a real database with the same fixture the report-path test uses:
- `{"entity":"win-loss","group_by":["size_band"],...}` returned the capture-private company's win count (6 deals, clearing `analyticsquery.DefaultFloor`, so the small-group floor did not incidentally cover it).
- `filters:[{field:size_band,op:eq,value:1001-5000}]` was a band-guessing oracle.

**What landed in the follow-up commit:**

1. `namedByAnalyticsQuery` calls `nameScopeVia`. Its own doc comment already records this defect class from a previous instance — a filter resolved through the wrong map — so this was the next one, not the first.

2. `TestEveryJoinedVocabularyDeclaresWhereItsScopeComesFrom` derives the obligation rather than restating it. The existing `TestEveryReferenceDimensionDeclaresItsScope` keys on a hand-kept list of four `t.`-prefixed id expressions, so `org.size_band` was invisible to it and the next `org.industry` would ship the same leak green. The new gate parses each spec's joins for their aliases and asks `auth.ScopeClauseFor` which tables are row-scoped — so the forecast's `stage` join, which is pipeline configuration every seat reads, is not asked for a scope it has nothing to inherit.

3. Two uncovered paths gained tests, both mutation-checked:
   - The drill-through walk-back (`reportwhere.go:292-299`). A hand-built handle pinning `size_band` as a bare predicate — no `by=` key — opened the hidden company's deal row with the loop removed. The product never mints that handle shape; an attacker types it.
   - The analytics surface, as above.

**Also corrected, honestly rather than quietly:** `TestFilteringByAGrantedFieldTakesItsGrant` passes with the vocabulary-grant fix reverted, because `project_id` also carries `filterScopes` and that refuses the value first. Its comment now says so. It proves the behaviour a caller sees; `TestTheVocabularyGateReadsTheCallersFilterKeys` is what fails when the fix is removed.

`make check-backend` green after rebase.

PR #4629, opened 2026-09-06
feat(agents): the prompt window follows the configured provider

LarsGradion:
A run's transcript was elided at 24,576 tokens on **every** installation, because the limit was a constant derived from Ollama's `num_ctx` cap. This deployment runs Gemini, whose window is orders of magnitude larger, so history was being cut to fit a local runner it never loads.

## Two questions shared one constant

They wanted different answers, and separating them is most of this change.

| Question | Follows | Answered by |
|---|---|---|
| How much **transcript** may a run carry? | the configured provider | `Brain.PromptWindow`, per step |
| How large may the **tool catalog** grow? | the supported floor | `runner.MinimumPromptWindow` |

The catalog must stay on the floor: the listing rides in the system prompt and is **never elided**, so a catalog sized against a large cloud window is one a local deployment cannot serve at all — and that failure would appear only after somebody rebinds routing, far from the change that caused it.

So `PromptTokenCeiling` becomes `MinimumPromptWindow`, which is what it always was, and the runtime value arrives beside it rather than replacing it.

## Asked per step, not per run

`PUT /v1/ai/routing` rebinds, every role polls it on a 30s interval, and the budget guardrail can demote a call to a cheaper tier mid-month. So step 1 and step 30 of one run can be served by different providers. Reading the window per step means a run that starts on a large cloud window and degrades simply elides more from that step on; frozen at construction it would keep assembling a prompt sized for a provider no longer serving it, which is the case that actually overflows.

## Where the number lives now

`model.Capabilities` gains `PromptWindow`, **declared per adapter** rather than read from a vendor list — the same reason `model.Info`'s own comment gives for dropping context length in the first place.

**Zero is the ordinary answer for a cloud wire**, and it does not mean "no window" — every model has one. It means no limit worth eliding a transcript against, because the real one is far above anything this product assembles. Ollama is the only adapter that declares a window, and the arithmetic behind the figure moves to `ai.ollamaPromptWindow`, beside the constants it reads, instead of living in an agent package that speaks to five providers.

`Router.PromptWindow` is the **minimum** over a task's bound ladder, for the reason `AttachmentMIMEs` takes the intersection. A rung declaring 0 contributes nothing rather than forcing the answer to zero — otherwise a cloud rung sitting beside a local one would erase the local rung's real constraint. That arm is mutation-checked: removing the guard makes a mixed ladder answer 0 and the test fails.

## Naming

The old names said *Ollama* where the meaning was *"the smallest window we support"*, which is confusing on an installation with no Ollama.

`ollamaMaxContext` keeps its name — it genuinely is Ollama's `num_ctx` cap, and it stays in `ollama.go`. What moved is the derived value (now `ai.ollamaPromptWindow`, beside the constants it reads) and the gate file: `prompt_ceiling_test.go` becomes `promptwindow_test.go`.

No identifier outside `internal/modules/ai/` names Ollama any more. The two mentions left in `window.go` are prose explaining why the floor is that number, which is the honest reason rather than a misleading name.

## Tests

Everything new is mutation-checked:

- `TestALargerProviderWindowKeepsWhatTheFloorWouldElide` — the same transcript bounded twice, and the floor arm asserts it *actually elided* first, or the comparison would be between two untouched transcripts.
- `TestAWindowOfZeroElidesNothing` — the failure a naive `> 0` check introduces.
- `TestATasksPromptWindowIsTheSmallestItsLadderMightLandOn` — both arms, including the mixed local/cloud ladder an operator actually builds.
- Two new parity gates: the runner's floor equals the adapter's derived window, and the adapter reserves exactly what one completion may ask for. Without the first, someone raising `ollamaMaxContext` for a bigger local model moves what the adapter serves and leaves the catalog budget behind — every gate green while it happens.

The certification doubles declare the **floor**, not the configured window: a lane that elided at whatever local routing happens to bind would measure the deployment rather than the build.

`make check-backend` green after rebase.

## Summary by CodeRabbit

- **New Features**
  - Prompt handling now adapts to the active model’s available context window.
  - Larger-window models can retain more conversation history and tool observations.
  - Models without a meaningful prompt limit no longer trigger unnecessary content reduction.
  - Context budgeting reserves space for generated responses, improving completion reliability.

- **Documentation**
  - Updated gate inventory documentation to reflect the revised prompt-window validation.


LarsGradion:
**Adversarial review found a regression in the first commit. Fixed in the second.**

`Router.PromptWindow` walked `taskLadders`, and a call is not confined to its ladder.

Two production paths put a call on a tier the ladder never names:
- The budget guardrail degrades `cheap_cloud` → `local_small` at 80% utilization (`degradeTo`, `tasks_gen.go:207`).
- The sovereign profile remaps every cloud rung to a local one (`router.go:333-353`).

`TaskAgentLoop`'s ladder is `{cheap_cloud, premium}`. `ServableTiers` for it is `{cheap_cloud, premium, local_small}` — verified in-tree.

So on a mixed deployment (a local model on `local_small`, cloud above it — an ordinary way to keep spend down), the window read off the ladder answered **0** for two cloud rungs, nothing was elided, and the guardrail then served the whole transcript to Ollama, whose runner truncated it. The reply comes back well-formed and empty, which reads as a bad model rather than a prompt too long.

**That is worse than what shipped before.** The old constant was unconditionally 24,576, so a degraded run was already correctly sized — this was a regression the diff introduced, on the very deployment shape the change was for.

It walks `ServableTiers` now. That function exists for exactly this distinction, and its own header says a call's ladder is not the set of rungs that can serve it. Worth noting the sibling blind spot: `AttachmentMIMEs` has the same off-ladder gap for the same two paths. I have not touched it here — it is a pre-existing question about carriage rather than this change's, and it deserves its own look.

**Second fix: a census that the floor is really a floor.**

`MinimumPromptWindow` is named "the smallest window any provider this build supports will carry" and the tool-catalog gates divide it, but the only check held it equal to *Ollama's*. Nothing enumerated the adapters. The next one to declare a smaller window is vLLM — a local runner that sizes a KV cache the same way and today declares nothing — and it would have made the catalog ration against a limit no binding could serve, silently.

The new census scans every `Caps()` in the `ai` module rather than reading a list, because the gate exists for the adapter somebody adds *next*; one naming its own subjects would go green on exactly the addition it is meant to catch. It resolves the declared expression (`cap - bucket - reserved`) rather than requiring a literal, and **fails** on a form it cannot evaluate rather than skipping it.

It lives in `backend/gates/` and not in the `ai` package: a module test reaching for `runner` would be a sibling import, which the DAG forbids.

Both fixes mutation-checked — reverting to the ladder walk fails `TestARungTheCallCanDegradeOntoBindsTheLane`; lowering `ollamaPromptWindow` by one bucket fails the census by name.

`make check-backend` green after rebase.

PR #4634, opened 2026-09-06
fix(main): every backend cause of the red main, in one branch

dungbanguyen:
**Every backend cause of the red `main`, in one branch.** Each was reproduced on a clean `origin/main` checkout before it was touched, and each is verified green here.

## What was failing

```
craft-static
  compose/server.go:111    newServer is 82 code lines (> 80)
  weeklyassemble.go:63     AssembleFor is 95 code lines (> 80)

TestFK_rowScopedTargetsHaveVisibilityDecision
  assurance_task_item.task_activity_id                no visibility decision
  privacy_notice_case.person_id                       no visibility decision
  sdr_handoff.{lead,person,organization,deal}_id      no visibility decision

TestMigrationsBuildTheCommittedSchema   34 objects in the schema, not in the catalog
TestACloneCarryingARowRuleIsNeverReused sdr_handoff_reason holds rows
TestPreferenceCenterOptOutBlocksAgentSend  consent: wording
```

## The three references, answered on their own terms

Two are real gaps and one is genuinely server-derived. Writing "gated" of a column that is not would be the rationalisation the rulebook forbids, so they are not answered uniformly.

**`assurance_task_item.task_activity_id` — gated.** `BundleException` took the task id off its input and referenced it behind `auth.Require(rbacBundle, update)` alone, which answers whether a principal may bundle at all, not whether it may file under THIS task.

The helper is the part worth knowing: `auth.EnsureLinkTarget` **refuses this table** — `auth: "activity" is not a row-scoped table` — because an activity's scope is *inherited* from what it is linked to rather than carried on its row. Five assurance tests failed on that before the switch to `auth.EnsureActivityVisibleLive`.

**`privacy_notice_case.person_id` — server-derived, classified.** There is no caller to probe: the duty is opened by the `NoticeCaseOpen` consumer from a `PersonCreated` envelope under a system principal. A notice duty is owed because the record was created, not because somebody asked.

**`sdr_handoff` ×4 — gated** through `auth.EnsureLinkTarget` in the transaction that writes each one.

## The reset table, and why it broke four packages

`sdr_handoff_reason` is boot-seeded reference data and was missing from `PreservedReferenceTables`. The consequence was not local: the emptiness proof refused every clone, `EnsureSchema` fell back to REBUILDING the schema, and the rebuild leaves River's own tables absent — so `seed-demo` failed on `relation "river_job" does not exist` and several compose tests failed on symptoms that named nothing to do with handoffs.

## The two long functions

`newServer` splits by wiring its four analytics handler sets after the literal, like the rest.

`AssembleFor` splits on the **read/write line** rather than at a line number: everything the review states about the week is settled before the insert decides anything, so `measureWeek` can gather it without knowing yet whether a row will land, and what stays behind is the transaction's judgement.

Redone against main rather than carried over — the same split was written when `AssembleFor` was 84 lines, it is 95 now, and the earlier patch no longer applied. Main added `LearningsState` to the literal, which stays with the write half.

## Verification

```
make check-backend                                        EXIT=0
make craft-static                                         EXIT=0   (was 2 MAJOR)
make test-it DIR=backend/migrations                       EXIT=0
make test-it DIR=backend/internal/platform/testdb         EXIT=0
make test-it DIR=backend/internal/modules/assurance       EXIT=0
make test-it DIR=backend/internal/modules/people          EXIT=0
make test-it DIR=backend/internal/compose/weekly          EXIT=0
make test-it DIR=backend/internal/compose  (the 3 send tests)  all PASS
```

## Provenance

Four commits are cherry-picked from the branch behind https://github.com/margince/margince/pull/4584, authorship intact — the handoff gating, the reset table, the send fixtures and the `newServer` split. That branch is currently unable to run CI because **I force-pushed over it by mistake**; the pre-clobber head is `6b46a7f8e` and the recovery command is on that PR. https://github.com/margince/margince/pull/4584 is subsumed by this branch and can be closed once this lands.

The rest is new: the two FK references that arrived after that branch was written, the catalog regeneration, and the `AssembleFor` split redone against today's main.


dungbanguyen:
Superseded, and by a better route than mine.

https://github.com/margince/margince/pull/4584 landed as "Main is green: eleven blockages, one branch" and carries equivalents for everything here — the `sdr_handoff` gating, the `sdr_handoff_reason` reset entry, the send fixtures, and both long-function splits. I have verified on a clean checkout of `2f3aa7981` that its versions hold:

```
make craft-static                                    EXIT=0   (was 2 MAJOR)
make test-it DIR=backend/internal/platform/testdb    EXIT=0
make test-it DIR=backend/internal/modules/assurance  EXIT=0
```

Only the head catalog remained — and it had gone stale again in the meantime, against a different table (`team_weekly_review_outlook`, not the `weekly_review_learning` set this branch regenerated for). So carrying this branch forward would have re-landed a golden file that was already wrong. https://github.com/margince/margince/pull/4639 regenerates it against the current schema instead, and is the only thing left between `main` and green.

Worth recording for whoever reads this later: this branch existed because I force-pushed over https://github.com/margince/margince/pull/4584 while it was `DIRTY` and could not run CI, then could not restore it. That turned out to cost more than it saved — #4584 recovered and landed a more complete fix, and everything I did here except one file was duplicated work. The lesson is on that PR: `--force-with-lease` computed after a `git fetch` is not a lease at all.

PR #4648, opened 2026-09-06
DO NOT MERGE — fix(capture): a meeting belongs to the calendar it came from

LarsGradion:
A rep with the narrowest row scope in the installation opened her morning brief and read "a meeting with Lucy" — a person another user had created an hour earlier, on an account that had never met her.

Measured on the staging installation, as that account: **465 of 468 meetings** and **354 of 372 emails** were discoverable. Private dinners and a partner negotiation among them.

## Four holes, each of which leaves the leak open on its own

**The capture path never judged a meeting.** `decideBirthTx` returned early for every non-mail kind, so a calendar event was born `audience='workspace'` while an email went through a five-step ladder. On staging, meetings landed in exactly one audience state; emails landed in eleven.

Meetings now run their own three steps — a counterparty hold, an explicit marker, and whether anybody outside the workspace is on it. The two mail-only steps are deliberately absent: a mailbox posture is not a calendar's, and inferring one from the other would publish private appointments in a workspace that shares its mail. A calendar posture of its own is a product decision, not something to infer.

**The row never said whose calendar it was.** `host_user_id` was NULL on all 468 staging meetings, though `captured_by` already carried the seat (`connector:gcal:<uuid>`). The write path has always read `host_user_id` — `writescope.go` admits an activity's host as a writer — so only the read path lacked an owner.

**The morning lane asked for every meeting and named nobody.** `keepReadersOwn` keeps a row it cannot judge, so every meeting the lane returned landed on every reader's queue. The lane now passes `HostUserID`, and the row honestly answers "the reader" — a claim `owner_test.go`'s census makes the query earn.

**`ActivityDiscoverClause` read an unlinked activity as a workspace-shared note.** Right for a note; wrong for a calendar event, and 450 of the 468 staging meetings linked nothing at all. A link-less *meeting* now falls back to its host. Every other kind is untouched.

## Risk

The change to `ActivityDiscoverClause` is the one to review hardest — it is a security predicate composed by ~30 readers.

A meeting with **no** host keeps the old answer on purpose, so rows captured before this fix stay visible to the people who own them rather than going dark on deploy; the migration is what removes that case. Order matters at deploy: backfill first, or every meeting briefly disappears from its owner's brief.

## Test

`meetingaudience_integration_test.go` drives the real capture path against Postgres. All three cases failed before the fix and pass after — including a colleague probing the actual discover clause.

## Verification

- `make check-backend` green (4m17s)
- Integration lanes green: `capture`, `activities`, `platform/auth`
- `make craft-static`, `make rls-store-path`, `make no-jurisdiction` green

---
## Summary by cubic
Fixes a privacy leak where captured calendar meetings were visible to every user in the workspace, regardless of whose calendar they came from. Meetings now run their own audience decision, stamp `host_user_id`, and the morning lane and discover clause only expose a meeting to its host until it's filed.

**Migration**
- Run `1788686725_a_captured_meeting_names_whose_calendar_it_came_from.up.sql` before deploying, or every meeting briefly disappears from its owner's brief.

<sup>Written for commit 0faf6af708ae72d4894fd8c09db20236cb102fd6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/4648?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/internal/compose/attention/meeting.go
+++ b/backend/internal/compose/attention/meeting.go
@@ -126,16 +126,20 @@ func classifyMeeting(item crmcontracts.AttentionItem, asOf time.Time) ranked {
 		item:       row,
 		deadlineAt: deadlineOf(item.DueAt),
 		occurredAt: occurredOf(item, asOf),
-		// NOT the reader's, and the obvious claim here is false. The lane lists
-		// meeting activities under the caller's ROW SCOPE — no owner or attendee
-		// predicate — so a team-scoped reader receives their team's meetings and
-		// naming the reader would make every one of them look like the reader's
-		// own appointment.
+		// The READER's, because the lane now asks for their meetings alone.
 		//
-		// The activity carries no owner this lane reads, so nobody is named. A
-		// meeting's real owner is its organiser, which arrives with the attendee
-		// read that does not exist yet.
-		ownerRef: unassigned(),
+		// This used to be unassigned(), with a comment saying a meeting's real
+		// owner "arrives with the attendee read that does not exist yet", and
+		// naming nobody was not neutral: keepReadersOwn keeps a row it cannot
+		// judge, so every meeting the lane returned landed on every reader's
+		// own queue. That is how a rep read "a meeting with Lucy" about
+		// somebody she had never met.
+		//
+		// The lane's query carries host_user_id now (attentionmeetingseam.go),
+		// so every row here came from this reader's own calendar and the honest
+		// answer is the reader. Saying so through the same helper the per-user
+		// lanes use keeps one spelling of "this row is yours".
+		ownerRef: ownedByWhoeverIsReading(),
 	}
 }
 
--- a/backend/internal/compose/attention/owner_test.go
+++ b/backend/internal/compose/attention/owner_test.go
@@ -219,8 +219,10 @@ func dayOfEveryLane() crmcontracts.Attention {
 // `ownedByWhoeverIsReading` asserts something specific about the LANE: its
 // query takes the acting user, so no other person's row could have come back.
 // That is checkable, and four of the lanes that first carried the claim failed
-// it — decisions, meetings, DSR and three of the five system sources are read
-// under the caller's ROW SCOPE instead, which is a different thing. A
+// it — decisions, meetings, DSR and three of the five system sources were read
+// under the caller's ROW SCOPE instead, which is a different thing. Meetings
+// have since EARNED the claim: the lane's query carries host_user_id, so a row
+// it returns came from the reader's own calendar. A
 // team-scoped reader receives their team's rows there, and naming the reader
 // would have made a colleague's duplicate pair look like the reader's own work.
 //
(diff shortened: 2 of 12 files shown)


LarsGradion:
**Do not merge.** Codex review found six P1 defects. Two are confirmed against the staging database.

**Confirmed fatal — the migration would abort.** Setting `host_user_id` on historical rows brings them under `activity_meeting_no_overlap`, a GIST exclusion constraint that treats every meeting as a one-hour block per host. Staging has **6 overlapping pairs** among the rows this migration would backfill, so it fails partway on the very database it was written for. Two back-to-back 30-minute meetings are enough.

**Confirmed by reading the mapper.** `meetingmap.go:206` never sets `Counterparty`, so every captured meeting hits `hold(no_counterparty)` and the `capture_import` posture persists as `held`. Recompute after linking clears only the row-carried hold, not the import contribution — so ordinary linked customer meetings would stay hidden from colleagues, which is the opposite of what the ladder is supposed to do.

**Also flagged, not yet verified:**
- The discovery fallback rejects legitimate readers of an unlinked meeting — participants, second-seat importers, explicitly-shared audiences — because content needs discovery AND audience, and no audience grant can overcome a discovery refusal. Second-seat capture calls `EnsureActivityVisible` before recording the import that would grant access.
- The brief lane's unconditional host filter hides meetings from invited participants and breaks `team`/`all` scopes, which `scope.go:161` documents as widening for meetings.
- The migration's `agent:<uuid>` suffix is a passport id, not an `app_user` id, so agent-captured rows misattribute rather than backfill.
- The new test reimplements a simplified copy of the discover SQL instead of calling `ActivityDiscoverClause`, so it would not catch a regression in the real predicate.

Clean on staging: no `restricted_at` meetings, and no manually-logged unlinked meetings the audience UPDATE would wrongly narrow.

**The underlying leak is real and still open** — 465 of 468 meetings discoverable by the most restricted account. The diagnosis holds; this particular fix does not. Keeping the branch for the tests and the analysis.

LarsGradion:
Superseded by #4656, which landed on main as 106d0be19d.

This branch was never mergeable — six P1 defects, two confirmed against the staging database. #4656 solves the same problem and fixes every one of them:

- The capture path holds a meeting on the ROW with an empty import posture, rather than through decideBirthTx/hold(), so the import contribution no longer re-pins `participants` after a link widens the row. That was the defect that made a filed meeting never actually open.
- The migration narrows `activity_meeting_no_overlap` to booked meetings (`source_system IS NULL`) BEFORE backfilling `host_user_id`. Without that the backfill aborted — staging had 6 colliding pairs among the rows it would have touched.
- The un-hold sweep in linkreconcile.go is gone, so nothing undoes the hold on its next tick.
- Discovery gates on capture_import/activity_participant membership rather than on host_user_id, which is an ownership label rather than evidence of attendance.
- The `meetings_unreported` lane, which this branch left untouched, is covered.

Closing unmerged. The analysis is preserved in #4649.

PR #4652, opened 2026-09-06
Bring the drafted body back into view when the disclosure band raises it

joshbruh-lang:
## What

After "Draft with AI" lands, the composer scrolls the drafted body back under the reader's eyes. Before this, the drafted words were pushed below the fold by the disclosure band that arrives with them.

## Why

A finished draft does not only fill the body. At the moment it lands it replaces the plain draft bar with the Art. 50 disclosure band — the disclosure sentence, the "Based on" line, the "Why this draft?" chips, the voice version, sometimes the provisional badge — and the head under it grows too, because the same answer fills To and the subject.

All of that grows *above* the body inside one scrolling drawer whose scroll position does not move. So the press that asks for words is what pushes those words past the fold: the composer answers "Draft with AI" with a notice about a draft the reader cannot see, and the reader has to hunt for the thing they just asked for.

The invariant: a draft that lands is a draft the reader can see. `block: "nearest"` rather than `"start"` holds the other half of it — a rewrite changes the words under a band that is already standing, so nothing there needs to move, and a body that already fits stays where it is.

## How verified

`frontend/src/screens/compose-draft-reveal.test.tsx` is the spec: it presses the button, waits for the disclosure band, and asserts the body editor — the element holding the drafted text, not the band that displaced it — is the one brought into view. It fails on `main` and passes here.

- [x] `make check` is green — `make check-fe` is green in full locally (7m35s, exit 0), which is the half this diff lives in. `make check-backend` does not pass in the container this was written in, and neither failure belongs to this diff: `contract-breaking-check` compared against a stale local `origin/main` ref (green after a fetch — "No changes detected"), and `lint-modules` fails across seven unrelated modules with `golangci-lint` built against go1.25 while the tree targets go1.26. CI's own path filter reaches the same conclusion: every backend job is skipped on this PR.
- [x] `make test-integration` is green (or: this change does not touch tenant data / RLS / the write shape) — this diff is one frontend screen and one frontend test; it touches no Go, no store, no migration
- [x] `make craft-static` reports no new blockers — no Go files in this diff, so the gate's corpus is empty for it; `make fe-lint` and `make fe-typecheck` are green

## AI involvement

Wholly AI-assisted: the diagnosis, the fix and the test were written by Claude Code in a remote session, working from a report that the mail body disappears once the AI draft band appears. A human reviewer owns the judgement calls in it — chiefly the choice to bring the body into view rather than shrink the disclosure band, which is a compliance surface and not this change's to trim.

## Accountability

https://claude.ai/code/session_01E2rfHMSdCmPK88pgJGyg51

### DIFF
--- a/frontend/src/screens/compose-draft-reveal.test.tsx
+++ b/frontend/src/screens/compose-draft-reveal.test.tsx
@@ -1,0 +1,129 @@
+/** @vitest-environment jsdom */
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { render as rtlRender, screen, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import type { ReactNode } from "react";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { LocaleProvider } from "../i18n";
+import { ComposeModal } from "./compose";
+import {
+  allowedPreview,
+  isPreviewDoor,
+  previewedAddresses,
+} from "./sendpermission.testkit";
+
+// WHERE a finished draft leaves the reader.
+//
+// The draft answers into the middle of a scrolling drawer and grows everything
+// above the body as it lands: the Art. 50 band appears where the draft bar was,
+// carrying the disclosure sentence, what the draft was based on and the voice
+// version, and the head below it fills with a recipient and a subject. The
+// words the rep pressed the button for end up under all of that — off the fold
+// on any drawer shorter than the form — so the composer answered "Draft with
+// AI" with a notice about a draft the rep could not see.
+
+function jsonResponse(body: unknown, status = 200) {
+  return new Response(JSON.stringify(body), {
+    status,
+    headers: { "Content-Type": "application/json" },
+  });
+}
+
+const DRAFT = {
+  subject: "Following up on your pricing question",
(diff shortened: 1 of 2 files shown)


joshbruh-lang:
On the Docstring Coverage pre-merge warning (50% vs 80%) — considered and deliberately not chased, for two reasons.

`revealDraftedBody` is not undocumented: it carries an eight-line explanation of *why* it exists. What the check counts is the `/** */` form, and this file draws that line by scope rather than by importance — top-level functions take `/** */` (`projectItself`, `useAnchorProject`), locals inside `ComposeModal` take `//`. `adoptDraftedBody`, three lines above the new function and the same shape, is a `//` block. Converting mine would make it the one local in the component that disagrees with its neighbours.

The rest of the percentage is test-harness boilerplate — `jsonResponse`, `stubRoutes`, `render` — copied from the sibling compose suites, where all three are undocumented too. Docstringing them here to clear a threshold would diverge from the files they were copied from and add comments that restate their own signatures, which the rulebook asks comments not to do.

CI is green on this head, and CodeRabbit's review itself reports no actionable comments. Happy to add the docstrings if a reviewer would rather have the check clean than the file consistent — say the word.

---

PR #4715, opened 2026-09-07
Working hours belong to the person, on their own clock

### LINKED ISSUE #2530: Meeting scheduling needs a design, and working hours belong to the person rather than the installation
Meeting scheduling was built to the smallest thing that could work, and has never been designed. Working hours are hardcoded 9–17 Monday–Friday in `backend/internal/modules/activities/scheduling.go:37-38`, with no way for anyone to change them.

The obvious next step — expose those hours as an installation setting an admin sets — is the wrong step, and that is the decision recorded here.

## Why an installation-wide setting is the wrong shape

People on one team do not share working hours. They sit in different countries, some work part time, some keep hours nobody else keeps. One pair of numbers set by an admin is wrong for most of them, and the ones it is wrong for cannot fix it.

So working hours are a **personal** setting, alongside a person's display language — each person sets their own, on their own clock, and nobody sets them for anybody else.

The scheduling code already suits this. `Store.Availability` takes a specific `host` user, and the public booking page resolves a `HostUserID` before calling it (`scheduling_public.go:125`). Both paths already know whose calendar is being asked about, so "whose hours" always has an answer. `app_user.timezone` exists as a column too.

What is missing is a way for a person to write any of it: `/me` is read-only, and nothing in the product sets `app_user.timezone` today.

## The bug underneath

Independent of the design, the current hours are evaluated on a **UTC** clock — `cursor.Hour()` and `cursor.Weekday()` in `freeSlots` read a UTC time. "9am" means 9am UTC, which is 4pm in Ho Chi Minh City, and the Saturday/Sunday check is wrong for the same reason: a Monday morning in Saigon is still Sunday to the scheduler.

This is already recorded in #26 as one of three sibling violations of zone-by-purpose, where the proposed fix is to convert booking hours to the **workspace** zone. That fix is correct as a timezone repair and wrong as a product answer, for the reason above. Whichever lands first should not close the other: #26 owns "a calendar day must be derived in a named zone, not UTC"; this issue owns "which zone, and whose hours".

## What needs deciding

**1. What can a person set?**

- **A.** One start time, one end time, and which days they work. Covers "8–18 Mon–Sat" and "9–13 Mon–Thu".
- **B.** A different start and end per day. Covers a long Monday and a half-day Friday.
- **C.** Several blocks per day. Covers a real lunch break, and is calendar territory rather than a setting.

A is the smallest thing that covers the two cases that prompted this, and B or C can be added on top without redoing it.

**2. Which clock do their hours mean?**

- **X.** The person's own timezone setting, editable by them. Needs a write path that does not exist.
- **Y.** Their browser's timezone, detected. No setup, but a person's working hours would follow them on holiday.

X, pre-filled from the browser the first time, so nobody has to think about it unless they move.

**3. What happens for someone who has set nothing?**

On the day this ships, that is everyone.

- **P.** Fall back to 9–17 Mon–Fri in their own timezone. Nothing regresses; the UTC bug is gone.
- **Q.** Treat unset as no constraint. Nothing is hidden, and a customer can book someone at 3am.
- **R.** Require it before booking works. Explicit, and breaks the feature for every existing person until they act.

P is the only one where nothing breaks on the day it ships.

## Recommendation

A + X + P, and the scheduling design written down before any of it is built — this issue exists because the feature was built without one.

## Consequence worth naming in the UI

Once hours are personal, the public booking page shows only the host's real hours. That is the point of the feature, but a host who sets 9–13 will receive roughly half the bookings they do today and will not necessarily connect the two. The screen should say so at the moment they save.

## Where this came from

The installation-settings work shipped base currency, base language and the record timezone. Business hours was the last item on that list and was planned as a fourth admin setting; it is filed here instead because it is not an admin setting and because scheduling has no design for it to fit into.

dungbanguyen:
Closes #2530.

Working hours were two Go constants — `9` and `17` — read on a **UTC** instant,
with no way for anyone to change either. So "9am" meant 4pm in Ho Chi Minh City,
and a Monday morning in Saigon was still Sunday to the scheduler.

The ruling's first instruction was that **the design gets written down before
any of this is built**, because this issue exists as a ticket precisely because
the feature was built without one. That is
[`docs/explanation/scheduling.md`](docs/explanation/scheduling.md), and it is the
first thing to read in this PR.

**Why not an installation-wide setting**, in the ticket's own words: *people on
one team do not share working hours… one pair of numbers set by an admin is
wrong for most of them, and the ones it is wrong for cannot fix it.* That last
clause is what makes it worse than no setting: an unconfigurable default is
honestly wrong for everyone, an admin-set pair is authoritatively wrong for the
majority. So this sits beside a person's display language.

**What a person sets:** one start time, one end time, the days they work, and
their timezone. One range rather than per-day hours — the two shapes that
prompted it (8–18 Mon–Sat, 9–13 Mon–Thu) are both a range plus a set of days,
and per-day hours layer on later without redoing this.

**Unset falls back to 09:00–17:00 Monday to Friday in the person's own zone**,
and that is the half that makes it shippable: on the day this lands nobody has
set anything, and the fallback regresses nothing while making the UTC defect
disappear for everybody. It is a named default (`defaultWorkingHours`), not a
leftover constant.

`app_user.timezone` becomes nullable for the same reason `locale` beside it is:
it was `NOT NULL DEFAULT 'UTC'` and **nothing in the product ever wrote it**, so
every value in it was the default rather than a choice. The migration nulls
exactly those rows and leaves any hand-written value alone.

**Reviewer notes**

- `activities` may not import `identity`, so what crosses is one resolver
  `compose` injects. An **unwired** store degrades to the stated fallback rather
  than refusing: the public booking page reaches that path, and a wiring gap
  must not become a customer-facing error about somebody's settings.
- The installation zone is read through `settings.ApplyTx`, and `Timezone` is
  declared `MachineryApplied()` for it. That door's own rule is *"a value whose
  disclosure through behaviour is the feature"*, which this is exactly: the
  slots a public booking page offers are in that zone and a customer reads them
  off the page. A rep asking about their own week holds no
  `installation_settings.read`, so the gated read cannot answer them.
- **The screen says what narrowing the hours will do, at the moment it is
  saved** — the ticket calls this out and it is part of the feature. Copy lands
  in all three shipped languages.
- The scheduling unit that changed is one function: `freeSlots` takes the host's
  hours and compares both ends **in their zone**, with midnight handled as the
  day's own end so a working day may honestly run to `24:00`.

**Verification**

`make check-backend` green, `make check-fe` green, `craft static --strict` clean
over the diff. `make test-it` green for `internal/modules/identity`.

Mutation-checked: replacing `start.In(h.Location)` with `start.UTC()` fails both
the Saigon-morning and the Sunday-evening cases — which are the two halves of
the defect this closes.

| new test | what it holds |
| --- | --- |
| a host's morning is their own | 09:00 in Saigon is bookable |
| Sunday evening in UTC is already Monday for them | the weekday is theirs too |
| a day the host does not work offers nothing | the four-day week |
| a working day may run to midnight | `24:00` is a real answer |
| a slot crossing midnight is not inside the day it started in | the end is not compared against the next day's clock |
| an unwired store keeps the stated fallback | the public path degrades rather than refusing |
| a person who has chosen nothing is bookable on the installation's clock | not UTC |
| the audit row carries both windows | "why did my bookings halve" |
| a refusal names the control the reader used | `end_time`, `days`, `timezone` |


dungbanguyen:
Rebased onto `main`. Two things changed on the way:

- `docs/README.md` conflicted where main rewrote the privacy-and-consent line; main's wording is kept and the scheduling entry sits above it.
- Two migrations landed on main while this was open, so the working-hours migration no longer sorted last. Re-stamped above main's head and the catalog regenerated against the new set.

`make check-backend` is clean apart from `TestAnEmailIsDrawnOnlyByTheCanonicalComponents`, which fails on pristine `main` too (#4790 / #4792) and is nothing this branch touches.

PR #4743, opened 2026-09-07
feat(rail): a generation a person waits on lights the AI-activity rail the moment it leaves

joshbruh-lang:
## What

Opening a meeting brief now lights the AI-activity rail (orb, edge light, bar line) the instant the request leaves, and so does every other generation a person waits on. Which HTTP operations hold a model call open is now declared in the contract (`x-waits-on-model: always | on-miss`), mirrored by the SPA's route table by method and path, and held together by a new parity gate.

## Why

The meeting brief is assembled fresh on every open and runs two model calls, but it is a `GET` — and the client counted only a `POST` to a hand-kept list of nine path suffixes as "the agent working". So a person opened the brief, waited on the agent for the whole of it, and the chrome reported an agent at rest: the one failure this surface cannot show, because an agent at rest looks exactly like an agent nobody is counting.

The list had drifted in both directions with nothing failing. It named `/research`, which calls a person-data provider and no model; it missed the intro drafts, the deal role proposals, the onboarding conversation, the `POST` regenerates of the person and account briefs, and every `GET` that generates.

The invariant, spelled once: **an operation whose handler waits on a model says so in the contract, and the client's table is that set.**

- `always` — the handler generates on every call (drafts, asks, the meeting brief, regenerates). Counted from the moment the request leaves.
- `on-miss` — the handler serves a stored reading and generates only when it has none (dossier, growth-fit, person brief, deal status, morning brief). Counted only once the request has outlived a stored answer (`CACHE_ANSWER_GRACE_MS`, one second), so a cached reading still never lights the orb — the behaviour the existing "does not count a read at a model route's path" test deliberately protected — and a first-time generation lights it a second late rather than not at all.
- Unmarked — answers without a model, or enqueues the work and answers 202 (deep read, document extraction, account scan, transcript proposals, technical enrich). Those reach the rail through their carrier and the feed, as before.

The same table decides which bodiless 502/503/504s get the "the server may still be working" problem body, for the same reason (duration), so both readers now share one derived set.

Fixed along the way: the ticker now says "Preparing a meeting brief" for the brief's own query, matching the person brief's existing line; the `/research` entry is dropped as it never called a model (a bodiless gateway error on it now reads as an ordinary server fault, which is what it is). One onboarding test that used a `502` incidentally to exercise "a problem body with no words for a reader" was moved to a `500`, because a `502` on the message route now correctly carries the gateway sentence.

## How verified

- [x] `make check` is green — run as its lanes: `go test ./gates` (whole package, 207s), `pnpm exec tsc -b`, `pnpm lint`, and the full `vitest run` (622 files, 8126 tests) after the test fix above. The Go generators (`gen-composition`, contracts, stubs, agent policy, record fields) were re-run and produce no drift from the marker. `make gen` itself fails in this sandbox on `handbook-embed` for want of `rsync`, which is unrelated to this change.
- [x] `make test-integration` is green (or: this change does not touch tenant data / RLS / the write shape) — no backend Go outside `gates/` changes; the server-side rail path (router announce with the meeting's subject label) was already in place and is unchanged.
- [x] `make craft-static` reports no new blockers — `craft static --strict` on the new gate: 0 blocker, 0 major, 0 minor.

The new gate was also proven to fail in both directions before being trusted: flipping the meeting brief's client entry to `on-miss` and deleting the `/coldstart` entry each produced a named failure.

## AI involvement

The whole change was written by Claude Code from the one-line request, including the route inventory (every handler traced to its model call or enqueue), the contract markers, the client rewrite, the gate, the tests and the docs. A human has not yet reviewed it line by line.

## Accountability

https://claude.ai/code/session_01Q1XCJLbZtrW9yN8sLtV5N5

---
_Generated by [Claude Code](https://claude.ai/code/session_01Q1XCJLbZtrW9yN8sLtV5N5)_

---
## Summary by cubic
Fixes the AI-activity rail so every generation a person waits on lights it the moment the request leaves, instead of only POSTs to a hand-kept list of nine path suffixes. The meeting brief — a GET that runs two model calls on every open — previously ran with the chrome reporting an agent at rest.

- The contract now marks each such operation `x-waits-on-model: always | on-miss`, and the SPA mirrors the marked set in `MODEL_ROUTES`, keyed by method and path.
- A new parity gate fails when the client table and the contract disagree in either direction or on the value.
- `always` routes count from the moment the request leaves; `on-miss` routes (dossier, person brief, deal status) count only once the request has outlived a stored answer, so a cached reading still never lights the orb.
- `/research` no longer counts — it calls a data provider, not a model — and a bodiless 502/503/504 now carries the "may still be working" body only on marked routes.
- The ticker now says "Preparing a meeting brief" for the brief's own query, and an onboarding test now uses a 500 where a 502 would now carry that gateway sentence.
- Also regenerates `frontend/src/api/schema.d.ts` and the gate inventory for main's changes, which had left those drift gates red on the base branch.

<sup>Written for commit 3e5db7b591ff8b23af0f702a47796e895013bebb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/4743?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


joshbruh-lang:
**`frontend / fe-quality` was red on the first head, and the failure is main's, not this PR's.**

The drift gate regenerated `frontend/src/api/schema.d.ts` and found two lines missing — `403` and `404` responses on the profile operation that `main`'s latest commit (`e61a41a`, "a member can correct the name their colleagues see") added to `crm.yaml` without regenerating the frontend types. CI checks out the merge with `main`, so the stale schema surfaces on every PR against it. The `x-waits-on-model` markers this PR adds produce no type change; regenerating on this PR's own base was a no-op.

Ported here rather than waited on: `8f6ac37` merges `main` and commits the regeneration, which no-ops once `main` carries its own. Nothing else changed.

---

joshbruh-lang:
**Two checks are red on `43062bf`, and both fail on code this PR does not touch and `main` already carries.** CI builds the merge with `main`, so they surface here; neither reproduces on this branch's own base, and no open PR fixes either, so I am not widening this PR to work around them. What is failing and the patch each needs:

**1. `integration / integration shard (2/6)` — `TestFK_rowScopedTargetsHaveVisibilityDecision` in `backend/migrations`**

> FK deal_stage_evidence.deal_id -> deal has no visibility decision

`deal_stage_evidence` landed in `main` with `389b2b3` (#4738) and the schema-fitness classification in `backend/migrations/schema_fitness_integration_test.go` was not extended for it. Reading the writer on `main` (`internal/modules/deals/stageevidence.go`): `RecordStageEvidence` and `RefuteStageEvidence` both put the client-supplied deal through `auth.EnsureWritableLive` before touching the row, and `ListStageEvidence` re-probes with `auth.EnsureVisibleLive`. So the honest entry for `rowScopedFKDecisions` is:

```go
"deal_stage_evidence.deal_id": "gated: auth.EnsureWritableLive in RecordStageEvidence/RefuteStageEvidence — the deal is client-supplied and recording evidence against it is a write of it, so the write probe rather than the visibility probe; ListStageEvidence re-probes the deal with auth.EnsureVisibleLive before handing rows back",
```

**2. `craftsmanship` — `make craft-static`**

> backend/internal/shared/kernel/events/catalog.go — file is 501 lines (> 500)

The file is byte-identical on `main` and this branch; the gate reproduces locally on `main`'s copy. It is not a one-line trim: the ceiling asks for a split by concept, and the catalog already has a sibling (`groups.go`) that shows the shape — move one group of event declarations into its own file.

Both are deterministic, so a re-run would not change them. I will keep this PR watched and re-check on the hour; once `main` carries either fix, merging it in brings this head green without further change here.

---

joshbruh-lang:
**`uat (AC screens + axe WCAG 2.2 AA + 390px, mocked)` is red on `43062bf`, and it is `main`'s failure too.** I ran the three failing specs against a fresh checkout of `main` at `0c850bb` (its current head, with the same preinstalled Chromium) and they fail identically there. This PR touches none of the screens or specs involved, and no open PR fixes them.

**1 + 2. `no AA violations on #/leads` (light and dark)** — axe `target-size` on `.lt-frow-more`: 15×15px against the 24×24 minimum. The control is the list surface's row-filter "⋮" segment (`frontend/src/design-system/listsurface.tsx`, styled in `listtable.css`), sized `aspect-ratio: 1` off the row's inner height. It is the lead queue that puts it on `#/leads` since `e414c4c` (#4737). The fix is in `listtable.css`: give the segment a `min-inline-size`/`min-block-size` of the 24px hit target the rest of the tree's controls use, or the safe-space equivalent, rather than letting the row's height decide it.

**3. `AC-mailbox-3` (`e2e/mailboxprivacy.spec.ts:74`)** — the spec navigates to `/#/settings/connections` and waits for `shared-posture-allowed-toggle`, but `9b08029` (#4735, "the mail-sharing rule moves to Capture rules") moved that card to the settings `capture` section. The toggle still exists; the spec's route is stale. The fix is the spec's `page.goto` (and any sibling assertion about which screen the warning sits on) pointing at the Capture-rules screen.

Both are deterministic, so no re-run. Same posture as the two backend checks above: this PR stays watched, and a merge of `main` once either lands brings this head green without a change here.

---

joshbruh-lang:
**Green on `3e5db7b`, mergeable, nothing outstanding.** `main` landed the fixes for all four base-branch failures named above (the `deal_stage_evidence` FK classification, the events catalog back under the size ceiling, the leads hit target, and the mailbox-privacy spec's route), so this head merges `main` in and the whole pipeline passes: gates, backend merge gate, all six integration shards, the frontend lanes, craftsmanship, UAT, and SonarCloud with 0 new issues. The merge's only conflict was the generated gate inventory, regenerated on the merged tree. Nothing about this PR's own change moved.

---

PR #4766, opened 2026-09-07
refactor(events): the catalog and the streams are two files

dungbanguyen:
**Main is at the file ceiling and every branch pays for it.**

`craft` counts **501** lines for a 500-newline file, so `events/catalog.go` sits exactly on the cap. Main's own CI never notices — `craft-static` judges *changed* files, and nothing changes this one on main. Every pull request that merges main in **does** change it, so it fails on branches instead of on the commit that filled it, telling each author their unrelated work broke a rule.

Seen on #4749, whose own diff is one integration test in another package.

## The split

Where the two halves actually part:

| file | answers | changes when |
|---|---|---|
| `streams.go` | which streams exist, and how a type's name maps onto one | the bus gains a stream |
| `catalog.go` | which events exist | a verb is added to a stream that already exists |

It was the second, ordinary motion that grew the file past the ceiling — while a §4.1 stream decision sat in the same file as three hundred routing rows.

No behaviour change: declarations move between files in one package, nothing is renamed or re-exported.

## The register line

`streamOverlay`'s uniqueness claim moves with its declaration, so its line in `uniquenessclaims.txt` is **repointed**, not added. The register is closed to new entries and this is the same entry in its new place — `TestTheRegisterHoldsNoEntryThatIsNoLongerAClaim` fails on the stale path otherwise, which is how I found it.

## Verification

- `craft static --strict`: PASS — 0 blocker, 0 major, 0 minor (was 1 major).
- 358 + 164 lines, both well under the cap.


dungbanguyen:
Closing in favour of #4762, which fixes this properly.

I split `events/catalog.go` because `craft static` reported it at 501 lines and blocked every branch that merged main in. #4762 shows why it reported that: `countLines` counted newlines **plus one**, so a terminating newline read as opening a further line and every well-formed file measured one longer than it is. `scripts/check-go-file-length.sh` holds the same 500-line cap over the same files and counts with `wc -l`, so two gates disagreed about one invariant.

That is the defect. `catalog.go` is 500 lines and was always within the cap — splitting it treated the symptom of a miscount, and the same miscount would have blocked the next file to reach the ceiling.

Nothing here worth salvaging: the split was a means to an end, not a change I would make on its own. Deleting the branch.

PR #4781, opened 2026-09-07
A live router occurrence leases one model call and renews it, so a dead call reads stalled in minutes

joshbruh-lang:
## What

A router-owned occurrence on the AI activity rail (summarize, draft_reply, offer_draft, weekly_review) now leases one model call at a time and renews the lease before each further call. A run whose settle never lands — the API restarted mid-call, or the trace flush timed out — reads stalled (amber, "it may have stopped") about five minutes after its last call instead of staying blue "working" for 15 to 45 minutes.

## Why

Stakeholder feedback: the rail's blue indicator was seen hanging for many minutes, several times.

The router announces a run's start once, in its own transaction, and the only thing that closes it is a best-effort flush. The lease on that start was sized for the longest healthy call — `CallCeiling` (300s) × every rung × every walk `CompleteStructured` can make, plus the flush — because the projection refused any re-announcement at the same `(attempt, rank)`, so the lease could never be extended. A lease is, by construction, how long a dead process is displayed as working, so a lease sized for the worst healthy call is the time a dead one is believed: up to three quarters of an hour of blue over a draft that would never arrive.

The invariant this PR installs: **one lease covers one model call, and the occurrence stays believable through a multi-call logical call by renewal, not by inflation.**

- `aiactivity/store.go`: the guard admits one write at an equal tuple — the same live state with a strictly later `stale_after`. An exact redelivery carries an equal instant, an out-of-order renewal an earlier one, and a settled row none, so all three still compare false and the at-least-once bus stays harmless.
- `ai/railstart.go`: `railLease()` is `CallCeiling + traceWriteTimeout`. The start answers a `RailClaim` (attempt, start instant) and every renewal repeats it, so a renewal extends the row rather than moving `started_at` or landing on a recounted attempt. Lease seconds round up, never down, so a renewal is always a strict extension. `*CallMeter` is held to the starter interface at compile time; a silent mismatch would mute the rail. The start, the renewals and the settle write their ledger row through one helper (`railemit.go`), so the three cannot describe one occurrence in two vocabularies.
- `ai/tracing.go`: `attemptLadder` renews before every rung; the first call after the announcement spends the start's lease instead.
- `RouteWriteDeadline` keeps its whole-call derivation (it is set once on the connection and cannot be renewed); its comment now says why the rail's lease is derived differently.
- `docs/explanation/ai-activity-rail.md`: the known-limits entries on the un-renewable lease and the refused equal-tuple write are rewritten to what holds now.

No frontend change: the wire carries the same states, and the rail already treats `stalled` as not-working.

**Fixed along the way:** `docs/reference/gate-inventory.md` is regenerated. Two commits merged to main after this branch forked each added a census gate, and the page's count fell one behind, which is what failed `integration unit coverage` on the merge commit CI tests. Regenerated with the gate's own `-update-gate-inventory` flag, nothing hand-edited.

**A limit worth knowing, left as is:** after the lease expires the row is amber, and the sweep that finally settles an abandoned router row (`CloseAbandonedRouterRuns`) runs on the `ai_activity_retention` job, whose cadence is 24h with a 2h grace. An abandoned row can therefore sit amber for up to a day. That is a job-cadence decision rather than the blue hang this PR is about, so it is stated here rather than changed.

## How verified

- `go test ./internal/modules/ai/ ./internal/modules/aiactivity/` green, including six new specs: one lease outlasts one call and the flush but not two calls; the first call spends the start's lease; every call after the first renews with the start's claim; a structured call renews before every later walk; a start that was not made is never renewed; the walk count stays within what `RouteWriteDeadline` assumes.
- Integration on a real Postgres 16 with pgvector (brought up locally, since the session container has no Docker): `make test-it` for the full `backend/internal/compose/integration` package (1,919 tests), `backend/internal/modules/ai` and `backend/internal/modules/aiactivity` — all green; the rail and projection tests rerun green on the tree merged with current main and after the lint fixes. New DB-backed specs: a renewal extends a live lease and advances `seq`; a late or duplicate renewal changes nothing; a settled row cannot be renewed back to life; end to end through `CallMeter`, a renewal keeps attempt and `started_at`, moves `stale_after` later, and the settle still closes the attempt the start opened; a late renewal cannot reopen a settled occurrence.
- `make lint` at the Makefile's pinned golangci-lint (v2.12.2): 0 issues on both passes. `make check-go` green (its one failure in the session was the history gate on a shallow clone; it passes once the clone is deepened), `make check-gates` green including the regenerated inventory, `craft static --strict` over the changed Go files: no findings.
- No manual runtime flow: the observable difference is a dead run turning amber after minutes rather than staying blue, which the DB-backed specs cover on the row and the existing rail specs cover on the orb.

- [x] `make check` is green (`check-go`; no frontend files changed, so `check-fe` has nothing new to check)
- [x] `make test-integration` is green for the touched packages (this change touches the write shape of `ai_task_run` and the router's announcements)

## AI involvement

The diagnosis, the code, the tests and the doc edits were AI-generated in a Claude Code session from the stakeholder's report, with the repository's rules, gates and existing tests as the constraints. Every line was written for this PR; nothing was hand-edited afterwards.

## Accountability

https://claude.ai/code/session_014P53X5yZbtMj5RHVrYTjZA


joshbruh-lang:
**CI status note.** The red `extension-reference` and `deterministic-gates` checks on the current head come from one gate, `TestAnEmailIsDrawnOnlyByTheCanonicalComponents` (`backend/gates/emailpresentation_test.go`), and it is red on `main` itself: it reports `design-system/composed.tsx` reading `EmailSummary` and drawing it with its own markup, and it fails identically on a clean checkout of `origin/main` (introduced by dbb95e06f, #4746, which changed that file). This PR touches nothing in `frontend/` and nothing in that gate; every other check in the run is green.

No open PR fixes it yet, so there is nothing to port. The fix belongs with #4746's author: either route the preview and status lines through `EmailEntry`/`EmailReference`, or ratify `composed.tsx` in the gate's `emailPassthroughs` with what it does with the type — that is a design call on that change, not this one. I have merged current `main` into this branch locally (clean, rail tests green) and will push and re-run once `main` is green again.

---

PR #4788, opened 2026-09-07
A stale record says which half moved

tinnhlt-gra:
## The problem

A stale certification record said this:

> `scenario held_out_draft_sits_close_to_the_author` — **or the prompt this build now builds from it** — has changed since the record scored it

That "or" hid the only thing worth knowing. Three situations collapsed into one sentence, and they call for opposite responses:

| What moved | What it means |
|---|---|
| **the case** | Somebody rewrote the test. Re-certify — the old number measured a different question. |
| **the prompt this build sends** | The product changed. The band that follows describes the NEW prompt, so a drop is the cost of that change, not the model getting worse. |
| **the grader** | The judge's own request moved. A band shifts with neither the test nor the product touched, which reads as a model regression and is not one. |

Nothing was missing to tell them apart. A scenario stamp is *already* three fixed-width digests concatenated — the scenario digested whole, the request its site builds, the grader's request beside it (`ScenarioStamps`). Which one differs is readable, not inferrable.

**It cost real time to learn that by hand.** Tracking down why `voice_build` on `gpt-oss-120b` went `certified 1.00 → not_supported 0.67` took diffing two 192-character stamps character by character to find the divergence at offset 64 — the candidate-request segment. The answer: `#3954` (the onboarding redesign) rewrote `voiceEvalDraftSystem` to ask for "two or three short paragraphs, roughly 80 to 140 words" instead of "a short email reply", because the drafts double as the sample shown on the onboarding result screen. A deliberate, well-reasoned product trade — and nothing connected it to the band drop. I reached two wrong conclusions on the way there, both from reading the wrong evidence, because the page would not say.

## What this adds

**The prose names the half.** `the prompt this build sends changed under scenario held_out_draft_sits_close_to_the_author since the record scored it`.

**The JSON carries the same finding structurally**, so it can be queried rather than parsed:

```json
"stale_cause": { "case_changed": ["scan_finds_the_promise_we_did_not_keep", …] }
```

`stale_cause.{case,prompt,grader}_changed` answers "which records went stale because the *product* changed" in one filter — the question that would have caught `#3954` the day it landed.

**The page opens with the count per cause.** On today's tree: 55 stale records, 55 with a changed case, 6 with a changed prompt, 0 grader.

## Two things I got right only on the second try

**The count is per record, not per row.** A record covers every site its task ships, so summing rows reported **95** case-changed records against a tree holding **55**. Now keyed by `(task, binding)`.

**An unrecognised stamp length attributes nothing and says so** — `something this build cannot attribute`, with its own `unattributable` field and its own page row. Under-reporting is the one failure here that would otherwise read as an answer: a cause list that silently dropped the stamps it could not split would say "nothing moved".

## Checks

`make check-all`; `TestASplitStampNamesTheHalfThatMoved` holds each position, both pairs, all three, and a stamp too short to read. The two failing gates fail on this machine only and report **zero findings from the repository tree** — `TestPublicTreeCitesNoDocumentGitIgnores` (the `superpowers` plugin writes `.superpowers/sdd/.gitignore` containing `*`) and `TestNoProseClaimsRLSStillScopesARead` (every finding from nested `.claude/worktrees/agent-*/`). Both walk `../` and see what a fresh clone does not.

---
## Summary by cubic
Stale AI certification records now name which half of the scenario stamp moved—the case, the prompt, or the grader—instead of lumping all three into "or the prompt has changed." Each cause needs a different response, and the old sentence hid which one applied; a prompt rewrite behind a certified 1.00 → not_supported 0.67 drop took manual diffing of two 192-character stamps to find.

- The JSON adds `stale_cause.{case,prompt,grader}_changed`, so scripts can filter records stale because the product changed rather than parse prose.
- The page opens with a per-cause count, keyed per (task, binding) instead of per row, which previously over-reported 95 records against a tree holding 55.
- Stamps too short to split get an `unattributable` field and their own row instead of being silently dropped.

<sup>Written for commit a8260b10fdfd218559238f99ccea9ee68347f132. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/4788?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tinnhlt-gra:
**These five failures are main's, not this PR's.** Reproduced in a clean worktree at pristine `origin/main` (`b5d21b55e`), with no part of this branch applied:

```
--- FAIL: TestAnEmailIsDrawnOnlyByTheCanonicalComponents
emailpresentation_test.go:306: design-system/composed.tsx reads EmailSummary and
draws it with its own markup — render EmailEntry or EmailReference instead, or
ratify the file in emailPassthroughs with what it does with the type
```

The same single test is what fails in `deterministic-gates`, `extension-reference`, `integration` and `integration unit coverage` — one gate, four lanes. CI tests the merge, so every PR opened right now inherits it.

This branch touches four files under `internal/compose/aicert/` plus three generated docs; nothing frontend, nothing in the design system. The gate passes on this branch in isolation and `make check-all` is otherwise green here.

Likely origin by last touch of the cited file: #4746, merged today. Its own lane was green, which is the signature of a gate whose subject is a file *outside* the package it lives in — it can go red from a commit landing beside it rather than from the change that trips it.

Holding this PR rather than merging into a red main. I have asked the other sessions on this machine whether anyone owns it, and will rebase once main is green rather than pushing on top of it.

tinnhlt-gra:
Green: 20 checks pass, 0 failures, 0 review findings.

Two notes on what that green does and does not cover.

**CodeRabbit reports `Review rate limited` again**, so its pass here is not a review that ran. Nothing in this PR has been machine-reviewed; SonarCloud passed, and the substance was checked by `make check-all` plus a test that holds each position of the stamp split. Flagging it rather than letting a green badge imply a review, same as on #4682.

**The three earlier red lanes were infrastructure, not this change.** `docker compose … --wait` failed to start Postgres 1.5s into `db-up` on integration shard 4/6, and the `ci` verdict job reported that lane downstream. No test failed. Re-ran the failed jobs and they passed.

Rebased onto `cf5fe3bc1` after main's red was fixed by #4806. Main has since moved 10 commits, none of which touch `gate-inventory.md` or `ai-certification.*`, so the generated artifacts in this PR are current against it.

PR #4795, opened 2026-09-07
The record pages are one readable class for an agent

### LINKED ISSUE #4773: The organization 360, graph and contact list carry no x-agent-access while their person twins are human-only
Three read routes are absent from `agentPolicies`, while the equivalent person-side routes are declared `human-only`:

| Route | `x-agent-access` |
|---|---|
| `GET /v1/people/{id}/360` | `human-only` |
| `GET /v1/organizations/{id}/360` | **none** |
| `GET /v1/people/{id}/graph` | `human-only` |
| `GET /v1/organizations/{id}/graph` | **none** |
| `GET /v1/organizations/{id}/contacts` | **none** |

## Why the asymmetry has an effect

The agent gate treats an unknown route differently by method. Mutations fail closed (`agentgate.go:176-180`, citing ADR-0055 §2). Reads fail open:

```go
pol, known := agentPolicies[r.Method+" "+pattern]
if !known || pol.Access == accessTool {
    return false
}
```

`agentgate.go:155-157`. So an agent passport reaching these three GETs is not refused; it falls through to the volume bound and then the handler.

## What is and is not at risk

**No record crosses a grant boundary.** All three go through `people.GetOrganizationTx`, which applies `auth.Require(organization, read)` and `auth.EnsureVisible` before any section runs. The agent reads under its granting human's authority, exactly as intended.

What differs is the **class** decision. The person 360 is human-only, and the organization 360 assembles the same kinds of section behind the same grants, including activity content (`org360/assemble.go:365,383,428` each require `activity:read`). Whatever made the person page human-only appears to apply here too.

## The decision needed

Either:

1. **Annotate the three as `human-only`**, matching their person twins. One-line contract change plus regeneration. This is the reading I would default to, since the two pages carry the same class of data.
2. **Decide the organization side is legitimately agent-readable**, and say so with `x-agent-access: tool` so the route is *declared* rather than merely unlisted. An account page is arguably what an assistant should read.

Option 2 is defensible — the code comment above `refusedAsHumanOnly` says an unannotated read is deliberately "ordinary agent-readable data". The problem is that the decision is currently implicit, and it reads as an oversight next to the person routes.

Worth a gate either way: nothing currently holds the person and organization sides of a composite read to the same agent class, which is how the two came apart.

Found while measuring #4761. Not caused by it, and not a disclosure.

LarsGradion:
`getPerson360` and `getPersonGraph` were annotated `x-agent-access: human-only`. Their organization twins — `getOrganization360`, `getOrganizationGraph`, `listOrganizationContacts` — were annotated nothing at all, and so answered a passport while the person pages refused one. The two halves carry the same kind of material. Neither posture was argued; one was written and the other was forgotten.

Lars decided the readable side: both halves answer an agent.

## What actually bounds a passport here

Not the annotation. The audience gate inside the assembly:

- `auth.ActivityAudienceArm` (`inheritedscope.go:202`) renders a per-row content test, projected as `content_available` at `person360/sectionstimeline.go:207`. A row the reader may not see comes back with `subject`, `body`, `thread_key` and `audience_reason` nil'd and `content_state: withheld` (`sectionstimeline.go:249-255`) — visible as a date and direction, never as content.
- `auth.ActivityContentClause` gates the reads that must not *count* an unreadable row: `since_last_visit` (`sectionstimeline.go:410`), `next_meeting` (`nextmeeting.go:51`), and the person graph's receipts (`network/persongraph.go:280`).

Both key on `p.UserID`, and `activityAudienceArm`'s own comment says it renders the test for "one human (or the human behind an agent)". Correspondence is the product's one read boundary and it is held in the SQL, for every principal alike. The organization side already ran this way through `activities.ListActivitiesTx`.

## Two locks, both removed

Each operation carried `x-agent-access: human-only` **and** `security: [{ cookieAuth: [] }]`. The second refuses the passport at the transport, one layer below where any `x-agent-access` test would look. Removing only the annotation would have left the endpoints just as closed and looked like a completed change. The generated diff shows the intended effect: `BearerAuthScopes` now joins `CookieAuthScopes` on both wrappers.

## What stays human-only

`acknowledgePersonView`, `dismissPersonMoment` and `acknowledgeOrganizationView`. Each spends one human's attention rather than reading it: acknowledging a view moves the baseline `since_last_visit` counts from, so an agent prefetching a record would destroy the answer its human opened the page to read. `person360`'s `Acknowledge` states this at its `auth.RequireHuman`, and `sectionstimeline.go:426` names that gate as the only thing standing between an agent and the write.

One consequence worth stating plainly: an agent reading the person 360 now reads its granting human's `since_last_visit` baseline, because `actingUser` is documented as "a lookup, not a gate". It only reads it — the write stays closed — and the organization 360 has behaved this way all along.

## The test

`TestTheRecordPagesAnswerAnAgentAndTheirWritesDoNot` asserts the readable set carries no annotation, overrides no `security`, and has no row in `agentPolicies`; and that the three writes carry all three. It names the org half as well as the person half, so a future drift in either direction fails.

Mutation-checked, one clause at a time:

| Mutation | Result |
|---|---|
| re-add `x-agent-access: human-only` to `getPerson360` | caught at line 79 |
| re-add `security: [{ cookieAuth: [] }]` to `getPersonGraph` only | caught at line 87 |
| remove `human-only` from `acknowledgePersonView` | caught at line 104 |

The second is the one that matters: an annotation-only test passes that mutation while the endpoint stays shut.

## Verification

`make check-backend` is green except `TestAnEmailIsDrawnOnlyByTheCanonicalComponents`, which **fails identically on clean `origin/main` at b5d21b55ed** — verified in a detached scratch worktree, `composed.tsx` untouched here. Not this diff. `./gates` with that one skipped: ok. `./internal/compose`: ok.

Closes #4773.

---
## Summary by cubic
Opens the person record pages to agent passports, matching their organization twins. `getPerson360` and `getPersonGraph` were `human-only` with a cookie-only transport override while `getOrganization360`, `getOrganizationGraph`, and `listOrganizationContacts` already answered agents; the two halves carry the same kind of material.

- Removes both locks — the `x-agent-access: human-only` annotation and the `security: [{cookieAuth: []}]` override — and regenerates `api_gen.go` so both wrappers accept bearer auth.
- Adds `TestTheRecordPagesAnswerAnAgentAndTheirWritesDoNot`, which pins all five readable operations and the three human-only writes so the halves can't drift again.
- `acknowledgePersonView`, `dismissPersonMoment`, and `acknowledgeOrganizationView` stay human-only because each spends one human's attention: acknowledging moves the `since_last_visit` baseline an agent must never consume.

Side effect: an agent reading the person 360 reads its granting human's `since_last_visit` baseline — read only, the write stays closed — and the org side already behaved this way.

Closes #4773.

<sup>Written for commit 75fd4babde3d2581cc1588fdb22693b565e8d2e2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/4795?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Access Updates**
  - Person 360 and person relationship views no longer require human-only access or cookie-based authentication declarations.
  - Access behavior for record pages is now aligned consistently across supported person and organization views.

- **Quality Improvements**
  - Added automated coverage to verify record-page access rules and help prevent future inconsistencies.


LarsGradion:
Closing unmerged. A Codex review found four defects that this change would have shipped, and I verified the three sharpest by reading the code rather than trusting the report.

**The change is not wrong in intent — it is incomplete.** The premise I argued the PR on was that the annotation is redundant because the audience gate inside the assembly is what bounds a passport. That is true of the timeline and the graph receipts, which is what I traced. It is not true of every section, and three places in this tree name the annotation itself as their protection.

**`introductions/routestate.go:83` says so in words:**

> This does not exclude an agent: a passport-backed agent carries its granting human's UserID, so it passes here. What keeps agents off this read is the route itself — getPersonGraph is annotated human-only, and the agent gate refuses it before the handler runs. **Wiring RouteStates to any agent-reachable caller would need that decision made again.**

I removed the annotation without making that decision. The consequence is that an agent would learn its granting human had been refused an introduction (`network/routeavailability.go:74`).

**`person360/moments.go:113` states the intended behaviour and does the opposite.** The comment says an agent "must not consume the granting human's [dismissal]: it sees every moment". The guard only checks that `UserID` is non-zero, which a passport satisfies, so the query at `:125` reads the human's dismissals and a card they put away vanishes from the agent's answer too. Correct only while the annotation made it unreachable.

**`people/relationshipchange.go:165` and `:195` carry no audience clause at all** — the aggregate filters on `kind` and `archived_at` only. A participants-only inbound message after a long silence yields `replied_after_gap` with its timestamp and gap length, which is derived correspondence content escaping through an aggregate.

Also flagged and not yet verified by me: employment `display_name` returned without requiring `organization.read` (`person360/sections.go:132`), and `getPersonProfileFields` staying human-only while its full data already flows through `/360`.

**One correction to my own PR body.** I claimed the cookie-only `security:` override was a second enforcement layer at the transport. Codex checked and it is not: the bearer path is chosen from the header (`identity/middleware.go:158`) and no OpenAPI security validator runs in the route middleware (`compose/routes.go:78`). Refusal happens through `agentPolicies` alone. Removing the override was still correct, but my reasoning for why it mattered was wrong, and the test comment asserting it is wrong too.

**What was sound and worth keeping** for whoever picks this up: the timeline and graph-receipt tracing holds. `ActivityAudienceArm` and `ActivityContentClause` do key on the acting user, and `activityAudienceArm`'s own comment says it renders the test for "one human (or the human behind an agent)". The new test mutation-checked in three directions. None of that is invalidated — it just was not the whole surface.

Reopening this needs the four fixes first, or an explicit product decision to accept each. #4773 stays open.

PR #4809, opened 2026-09-07
fix(frontend): a thread member draws scalars, not the email carrier

### LINKED ISSUE #4799: main is red: MessageWords draws an email preview outside the canonical components
`TestAnEmailIsDrawnOnlyByTheCanonicalComponents` fails on a clean `main` (verified at `490bef626`, clean tree, no local changes, after `make composition`). Because the `gates` package runs in three jobs, **every open PR inherits three red checks** from it:

- `deterministic-gates`
- `extension-reference`
- `integration / integration unit coverage`

```
design-system/composed.tsx reads EmailSummary and draws it with its own markup
— render EmailEntry or EmailReference instead, or ratify the file in
emailPassthroughs with what it does with the type
```

## Cause

`git log c0555ce9c..490bef626 -- frontend/src/design-system/composed.tsx` names exactly one commit: **`dbb95e06f` (#4746**, merged 2026-09-07T06:24:52Z, cc @joshbruh-lang).

The new `MessageWords` reads the email carrier and returns markup without mounting a canonical component (`frontend/src/design-system/composed.tsx:1640`):

```tsx
if (entry.emailSummary) {
  return entry.emailSummary.preview ? (
    <span className="tl-msg-text">{entry.emailSummary.preview}</span>
  ) : null;
}
```

`rendersCanonically` judges **per function**, so `composed.tsx` mounting `EmailEntry` in `TimelineRow` and `TimelineGroupRow` does not discharge `MessageWords`.

## Note on the obvious-looking fix

**Adding `composed.tsx` to `emailPassthroughs` would be wrong.** That map is file-keyed, and the gate's own comment records why the rule went per-function in the first place:

> An earlier version read the whole file: `composed.tsx` mounts EmailEntry in TimelineRow, which discharged the file, and TimelineGroupRow beside it went on writing its own subject-and-preview markup [...] A company timeline of grouped threads therefore showed old-style rows next to canonical ones on the same page while this census reported clean.

So a file-level waiver here re-opens precisely the shipped bug this gate was tightened to catch, and the ratification sentence would also be untrue — the file does render.

## Options, for whoever owns the design

1. **`MessageWords` draws scalars instead of the carrier.** Thread the preview text through as a plain `string` (the mapper in `activitytimeline.tsx` is already a ratified passthrough that builds the entry). This is the `EmailReference` precedent exactly — the gate's comment notes it "takes a subject and a date as plain scalars — no contract type at all, which is what stops a citation from ever growing a preview." Keeps full gate coverage of `composed.tsx`; no waiver.
2. **Mount a canonical component in `MessageWords`.** Appears to contradict the stated design — `ThreadMessage`'s comment says a member of a conversation deliberately has "no date gutter, no rail and no kind of its own," which is what an `EmailEntry` row brings.
3. **Widen the gate deliberately** with a per-function ratification (`file#function`) so `composed.tsx#MessageWords` can carry a written sentence while the rest of the file keeps coverage. Honest, but it's a change to a census gate and wants its author's agreement — the gate's three-way distinction (renderer / passthrough / description) genuinely has no slot for "a canonical renderer's own sub-part drawing one already-summarised scalar."

I'd suggest **1**.

I'm not taking this myself because it's a design call inside a feature that landed 40 minutes ago, and one of the plausible fixes changes a gate. Raising it rather than guessing.

Unblocking separately: #4791 fixes the *other* red gate on `main` (`gate-inventory.md`'s stale census count), which is independent of this one. Both must land before anything can go green.

dungbanguyen:
Fixes #4799 — **main is red**, and the `gates` package runs in three CI jobs, so every open PR inherits three red checks from it (`deterministic-gates`, `extension-reference`, `integration / integration unit coverage`).

`MessageWords` and `ThreadMessage` read the email carrier and return markup without mounting a canonical component. `rendersCanonically` judges **per function** — deliberately, per the gate's own comment: an earlier file-level rule let `TimelineRow`'s mount discharge the whole file while `TimelineGroupRow` beside it went on writing its own subject-and-preview markup, and a company timeline showed old-style rows next to canonical ones on the same page while the census read clean.

## Option 1 from the issue

The carrier stops at the canonical components, and the thread member takes what it draws as **scalars**:

| scalar | says |
|---|---|
| `messagePreview` | the words — present exactly when the row is a message, empty when that message has no preview |
| `messageVisibility` | who may read it |

`EmailReference` has taken scalars for the same reason all along: *"no contract type at all, which is what stops a citation from ever growing a preview."*

`messagePreview`'s presence also answers *"is this a message?"*, which is what the two branches on the carrier were really asking — so `ThreadMessage`'s opener reads it too.

## Why the mapper derives them

Deriving the scalars in a local helper here would have satisfied the census — a function that reads the carrier and returns a *value* is not a renderer — while leaving exactly the reach the gate exists to remove. They come from `activitytimeline.tsx`'s mapper instead, which is already a ratified passthrough.

The thread test's factory derives them the same way rather than each case setting them: a test that set the carrier alone would build a row production never builds, and then assert on it.

## Verification

- The gate passes; reverting either scalar makes it fail again.
- 1064 design-system tests pass. No behaviour change — same preview, same badge, same opener.
- `make check-backend` exit 0, `tsc` and `biome` clean.

## Not taken

- **A file-level `emailPassthroughs` entry** — the issue explains why: the map is file-keyed, so it re-opens the shipped bug the gate was tightened to catch, and the sentence would be untrue since the file does render.
- **Mounting a canonical row in `MessageWords`** — contradicts `ThreadMessage`'s stated design: a member of a conversation has no date gutter, no rail and no kind of its own.

### DIFF
--- a/frontend/src/design-system/activitytimeline.tsx
+++ b/frontend/src/design-system/activitytimeline.tsx
@@ -129,6 +129,14 @@ export function activityTimeline(
       // exactly when kind=email, so the row branches on the field and every
       // other kind keeps the reading it had.
       emailSummary: activity.email_summary ?? undefined,
+      // The preview as a scalar beside the carrier, for the thread member that
+      // draws its words inline. Present exactly when the row is a message, so
+      // the branch that used to read the carrier reads this instead and the
+      // contract type stays with the canonical components.
+      messagePreview: activity.email_summary
+        ? (activity.email_summary.preview ?? "")
+        : undefined,
+      messageVisibility: activity.email_summary?.display_status,
       audience: activity.audience,
       withheld: activity.content_state === "withheld",
       threadKey: activity.thread_key,
--- a/frontend/src/design-system/composed.thread.test.tsx
+++ b/frontend/src/design-system/composed.thread.test.tsx
@@ -20,13 +20,19 @@ afterEach(cleanup);
 const render = (ui: ReactNode) =>
   rtlRender(<LocaleProvider initial="en">{ui}</LocaleProvider>);
 
+// entry stands in for activitytimeline's mapper, and derives the message
+// scalars from the carrier the way it does. A test that set the carrier alone
+// would build a row production never builds, and then assert on it.
 function entry(overrides: Partial<TimelineEntry> = {}): TimelineEntry {
+  const summary = overrides.emailSummary;
   return {
     id: overrides.id ?? "m1",
     kind: "email",
     title: "Re: Renewal",
     atIso: "2026-07-03T10:00:00Z",
     provenance: { kind: "connector", connector: "gmail" },
+    messagePreview: summary ? (summary.preview ?? "") : undefined,
+    messageVisibility: summary?.display_status,
     ...overrides,
   };
 }
--- a/frontend/src/design-system/composed.tsx
+++ b/frontend/src/design-system/composed.tsx
@@ -723,6 +723,23 @@ export type TimelineEntry = {
    */
   emailSummary?: EmailSummary;
   /**
+   * What a THREAD MEMBER draws, as scalars: its preview text, and who may read
+   * it. `messagePreview` is present exactly when the row is a message, and
+   * empty when that message has no preview to show — so it answers "is this a
+   * message?" as well as "what does it say?".
+   *
+   * Duplicates of fields inside `emailSummary` on purpose. A thread member
+   * draws its words inline rather than mounting the canonical row
+   * (ThreadMessage below says why), and a sub-part of this file reaching into
+   * the contract type to draw it is how the canonical components stop being
+   * the one place a message is drawn — a company timeline once showed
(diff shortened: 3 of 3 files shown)


dungbanguyen:
Closing in favour of #4806, which is earlier and is the better answer.

Both fix the same red gate. Mine kept the carrier away from the thread member by threading `messagePreview` and `messageVisibility` through as scalars — option 1 from #4799. #4806 instead gives the message's words **one spelling** and has the thread member use it, which is what the gate is actually asking for: not "hold no contract type" but "a message is drawn in one place". Two renderings agreeing today because two authors matched them is the state the census exists to end, and my version left exactly that — a span here and the canonical row there, kept in step by nothing.

It also carries the README and the gate's own comment along with the change, which mine did not.

Nothing here worth salvaging. Deleting the branch.

PR #4813, opened 2026-09-07
A guard that admits a passport is not a human check

LarsGradion:
Fixes the five defects a Codex review found in #4795, **without** opening the endpoints. The contract is untouched: `getPerson360` and `getPersonGraph` stay `x-agent-access: human-only`. These are the fixes that have to land before that decision can be revisited, and they are worth landing on their own.

## The pattern

A passport carries its granting human's `UserID`. So a check shaped like

```go
if actor, ok := principal.Actor(ctx); !ok || actor.UserID.IsZero() { ... }
```

asks "is anybody behind this call", and an agent answers **yes** using the id of the person who minted it. Two reads relied on such a check to stay human-only, and each went on to return that person's own standing to a credential acting on their behalf.

One of them said so in a comment — `introductions/routestate.go:83`:

> What keeps agents off this read is the route itself — getPersonGraph is annotated human-only… **Wiring RouteStates to any agent-reachable caller would need that decision made again.**

That made the answer depend on which door a read arrived through rather than on what it discloses, and left two guards that were correct and untested at the same time. Both are `auth.RequireHuman` now, at the disclosure.

`person360/moments.go:113` is the sharper case: it **states the correct behaviour** — an agent "must not consume the granting human's [dismissal]: it sees every moment" — and implemented the opposite, for exactly this reason.

## Three more sites returned more than their own comments claimed

**`people/relationshipchange.go`** folded activities with no audience clause, so a message held to its participants still produced `replied_after_gap` carrying the reply's timestamp and the length of the silence it broke. That is what `auth.AudienceWorkspaceOnly` exists to prevent, and its doc names relationship strength as one of the four readers that owe it. `strength.go` applies it in **five** places; the change *derived from that strength* applied it in none. Both passes carry it now, so the change and the score it reports a change in fold one population.

**`person360/sections.go`** returned an employer's `display_name` on a statement filtering `id` and `archived_at` alone, while its doc promised a name the caller cannot read "is simply absent" for two separate reasons. Both are spelled now: `auth.Require` for the object grant, `auth.EnsureVisible` for the row. A refusal blanks the name rather than failing the section, because an employment the reader may see is still a true edge.

**`people/enrichment.go`** read the employer's name *and primary domain* through the employment join with no organization row scope, flipping a provider verdict from `nothing_to_look_up` to `never_run` for an account outside the caller's scope.

## Verification

Every fix mutation-checked one clause at a time, and each mutation confirmed to **compile** first — a mutation that fails to build never ran and proves nothing (that happened once here and was redone).

| Mutation | Caught by |
|---|---|
| drop `AudienceWorkspaceOnly` from `changeInputs` | the held arm fails; the open arm stays green |
| drop `RequireHuman` from `RouteStates` | agent read returns `<nil>` instead of a refusal |
| neutralise the `organization` grant check | leaks `"Vertraulich GmbH"`, named in the failure |

**The audience test runs two arms over one fixture** — the same reply open, then held — rather than two readers over one message. `AudienceWorkspaceOnly` is deliberately *not* the reader's own audience: "the question an aggregate asks is may EVERYONE see this, and the answer is the same for everybody." A fixture asserting the author still sees the change would assert the opposite of the rule. My first draft did exactly that and its positive control caught it.

`SubjectIdentifiers`'s SQL passes `employment.CurrentPrimarySQL("r")` as an **argument** to `storekit.SQLf` rather than concatenating it into the format string, so a helper that one day returns a literal `%` cannot be eaten by the verb scan. Verified empirically that its current output contains none.

Checked that no legitimate non-human caller breaks: `personbrief/service.go:97` and `personresearch/service.go:67` already call `RequireHuman` before assembling, and no worker or `cmd/` path reaches `person360`. `ScopeClauseFor` returns `""` for an unbounded principal, which the `visible = "true"` fallback handles.

`make check-backend` is green except `TestAnEmailIsDrawnOnlyByTheCanonicalComponents`, which **fails identically on clean `origin/main`** — bisected by another session to #4746, which added new `EmailSummary` markup to `composed.tsx`. Not this diff.

Refs #4773.

---
## Summary by cubic
Fixes humanity checks that admitted agents into human-only reads, plus three more disclosure leaks found in the same review, and keeps a provider submission's employer from traveling beyond its requester's scope. A passport carries its granting human's `UserID`, so `actor.UserID.IsZero()` passed for an agent and returned the human's own introduction ledger and dismissed moments. Reads now gate on `auth.RequireHuman` at the disclosure; endpoints stay `x-agent-access: human-only`, so no contract changes.

**Bug Fixes**

- `introductions/reads.go` requires a human in `roleOf`, `ForPerson`, and `AwaitingMyAnswer`, so an agent can no longer become a party to its granting human's asks.
- `relationshipchange.go`, `championcover.go`, and `engagement.go` apply `auth.AudienceWorkspaceOnly`, so a reply held to its participants no longer derives a `replied_after_gap` change or appears in engagement counts.
- `sections.go` and `enrichment.go` require the organization grant and row scope before returning an employer's name and primary domain; a refusal blanks the name rather than failing the section.
- Provider submissions check the run's stored requester, not the connector, before sending the employer; a requester who cannot read organizations sends the subject without it. `SubjectIdentifiers` passes `employment.CurrentPrimarySQL("r")` to `storekit.SQLf` as an argument instead of concatenating it into the format string.
- The new integration tests run one positive control and one refusal over the same fixture; the audience test runs two arms (open then held) over one reply.

Each fix was mutation-checked one clause at a time. `make check-backend` is green except `TestAnEmailIsDrawnOnlyByTheCanonicalComponents`, which fails identically on clean `main` (a #4746 regression unrelated to this diff). Refs #4773.

<sup>Written for commit bfdb3a6f1c3cb7f8c13cc2bb295ed5a87b42a0bf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/4813?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Security & Access**
  - Restricted introductions, routing details, roles, and person moments to human users.
  - Prevented agents from accessing or consuming their granting human’s private data and dismissals.
  - Organization names and employer identifiers are shown only when the viewer has appropriate access.
  - Integration requests now withhold employer details when the requester lacks organization access.

- **Data Visibility**
  - Engagement and relationship insights now consider only workspace-visible activity.
  - Relationship changes respect message audience restrictions, preventing limited-audience messages from influencing derived results.

### DIFF
--- a/backend/internal/compose/integration/agentcarriesitshuman_integration_test.go
+++ b/backend/internal/compose/integration/agentcarriesitshuman_integration_test.go
@@ -1,0 +1,251 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+//go:build integration
+
+package integration
+
+// A passport carries its granting human's UserID, so a check for one admits an
+// agent. Two reads relied on such a check to stay human-only.
+//
+// The pattern is worth naming because it looks like a gate and is not:
+//
+//	if actor, ok := principal.Actor(ctx); !ok || actor.UserID.IsZero() { ... }
+//
+// It asks "is anybody behind this call", which an agent answers YES to using
+// the id of the person who minted it. Each read then went on to return that
+// person's own standing — what they dismissed, who refused them — to a
+// credential acting on their behalf. auth.RequireHuman is what tells the two
+// apart, and each site now says so where the check is.
+//
+// introductions/routestate.go named getPersonGraph's human-only annotation as
+// its protection, which made the answer depend on which door a read arrived
+// through rather than on what it discloses. That protection was also narrower
+// than it looked: an agent already assembles a Person360 today, through
+// prep_for_meeting → the meeting brief → AssembleScoped, so the dismissal guard
+// is live behaviour rather than defence in depth.
+//
+// The third case in this file is not about agents at all: the employment edge
+// returned an employer's name to a caller holding no organization grant, which
+// the function's own doc comment already claimed it did not.
+
+import (
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/margince/margince/backend/internal/modules/introductions"
+	"github.com/margince/margince/backend/internal/platform/auth"
+	"github.com/margince/margince/backend/internal/shared/apperrors"
+	"github.com/margince/margince/backend/internal/shared/kernel/ids"
+	"github.com/margince/margince/backend/internal/shared/kernel/principal"
+)
+
+// routeStatePerms holds every grant RouteStates checks before its human arm runs, so
+// a refusal in this fixture is that arm and never a missing grant.
+var routeStatePerms = principal.Permissions{
+	RoleKeys: []string{"rep"},
+	Objects: map[string]principal.ObjectGrant{
+		"person":       {Read: true},
+		"introduction": {Read: true},
+		"relationship": {Read: true},
+	},
+	RowScope: principal.RowScopeAll,
+}
+
+// TestTheIntroductionLedgerRefusesAnAgentCarryingItsHumansID covers the read
+// that reports which introductions the caller already asked for and which they
+// were refused.
+//
+// The human is admitted and the agent is not, over ONE fixture with ONE set of
+// permissions: the two contexts differ in principal type and in nothing the
+// test chose, so the refusal cannot be a grant the agent was never given.
+func TestTheIntroductionLedgerRefusesAnAgentCarryingItsHumansID(t *testing.T) {
+	e := Setup(t)
+	contact := e.SeedPerson(t, "Marit Vermittelt", &e.Rep1)
+	store := introductions.NewStore(e.DB(), time.Now)
+
+	human := e.As(e.Rep1, []ids.UUID{e.Team1}, routeStatePerms)
+	agent := e.AgentFor(t, e.Rep1, []ids.UUID{e.Team1}, routeStatePerms)
+
+	// The positive control comes first: a fixture that refuses everybody would
+	// pass the agent assertion below while proving nothing.
+	if _, err := store.RouteStates(human, ids.From[ids.PersonKind](contact)); err != nil {
+		t.Fatalf("the granting human's own read = %v, want the ledger", err)
+	}
+
+	_, err := store.RouteStates(agent, ids.From[ids.PersonKind](contact))
+	if !errors.Is(err, apperrors.ErrPermissionDenied) {
+		t.Errorf("an agent read the introduction ledger → %v, want ErrPermissionDenied — "+
+			"it reports which introductions its human has open and which they were "+
+			"refused, so an agent reading it learns its human was turned down", err)
+	}
+}
+
+// TestRequireHumanIsWhatSeparatesAnAgentFromItsHuman states the premise the two
+// guards now rest on, as a fact rather than an assumption: the agent and its
+// granting human carry the SAME UserID, so an id check cannot separate them and
+// auth.RequireHuman is what does.
+//
+// Asserted directly because it is the premise, not the behaviour: the guards
+// themselves are covered through their real callers above and below.
+func TestRequireHumanIsWhatSeparatesAnAgentFromItsHuman(t *testing.T) {
+	e := Setup(t)
+	human := e.As(e.Rep1, []ids.UUID{e.Team1}, routeStatePerms)
+	agent := e.AgentFor(t, e.Rep1, []ids.UUID{e.Team1}, routeStatePerms)
+
+	humanActor, ok := principal.Actor(human)
+	if !ok {
+		t.Fatal("the human fixture carries no actor")
+	}
+	agentActor, ok := principal.Actor(agent)
+	if !ok {
+		t.Fatal("the agent fixture carries no actor")
+	}
+
+	// The defect, stated. If this ever stops being true, the guards below are
+	// answering a different question and this file should be re-read.
+	if agentActor.UserID != humanActor.UserID {
+		t.Fatalf("the agent resolved to user %s and its granting human to %s — a passport "+
+			"is supposed to carry its human's id, and this suite's premise is that an "+
+			"id check therefore cannot refuse one", agentActor.UserID, humanActor.UserID)
+	}
+	if agentActor.UserID.IsZero() {
+		t.Fatal("the agent carries no user id at all, so the id check WOULD refuse it " +
+			"and the fixture is not modelling the case these guards are about")
+	}
+
+	if err := auth.RequireHuman(human); err != nil {
+		t.Errorf("auth.RequireHuman refused the granting human → %v, want admission", err)
+	}
+	if err := auth.RequireHuman(agent); !errors.Is(err, apperrors.ErrPermissionDenied) {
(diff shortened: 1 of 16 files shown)


LarsGradion:
Codex reviewed this and found **four MAJORs, no BLOCKER**. All four were real. Fixed in the amended commit, which is now eight fixes across five modules rather than five across four.

## A correction to this PR's own body

I wrote that no legitimate non-human caller reaches the person page, having checked `personbrief` and `personresearch` and found both call `RequireHuman` first. **There is a third path and I missed it:** `prep_for_meeting` → `writtenBrief` → the meeting-brief seam → `Person360.AssembleScoped` (`tools_intents.go:249`, `meetingbriefseam.go:58`, `meetingbrief/service.go:363`). `meetingbrief` calls no `RequireHuman` of its own.

So **an agent already assembles a person page today**, and the dismissal guard is live behaviour rather than defence in depth. Without it, an agent preparing its human for a meeting consumes that human's dismissed cards. The fix is unchanged — returning "not dismissed" preserves the assembly — but the reason it matters is stronger than this PR claimed.

## Two siblings of these defects, in code the first pass never opened

**Deal engagement omits the same audience clause** — `deals/engagement.go:87` and its batched twin `deals/championcover.go:284`, both `EXISTS` arms of each. These decide whether a stakeholder counts as engaged, consumed by deal health (`health.go:389`) and coverage risk (`risk.go:215`). A held inbound could mark a stakeholder engaged on a card every colleague reads. Same class as the relationship-change leak, one module over.

**Three more introduction readers carry the `UserID`-only guard** — `roleOf` (which `Get` and `load` route through), `ForPerson` and `AwaitingMyAnswer` in `introductions/reads.go`. They carry request text, status and decision reason. `AwaitingMyAnswer`'s own comment says *"answering with somebody else's asks would be handing an agent a colleague's inbox"* — and its guard admitted exactly that. All three now gate on `auth.RequireHuman`.

**The employer fix was half of the question.** Row scope answers "which accounts are yours"; it says nothing about a caller holding no `organization` grant at all. `SubjectIdentifiers` now asks both, and a refusal keeps the person's own identifiers — including the LinkedIn handle, which needs no company grant and which my first attempt at this skipped by returning early.

## Three of my tests did not protect what they claimed

This is the finding worth the most, because untested fixes are how #4795 happened.

- **The dismissal guard** was asserted through `auth.RequireHuman` in isolation. Removing the call in `momentDismissed` could not have failed it. Now driven through the real assembly: dismiss as the human, then read the page as both. Mutation-checked — without the guard the agent's page falls back to "Nothing needs you today", having consumed the human's dismissal of `thin_relationship`.
- **The second audience clause** was unreachable: holding the reply leaves `LatestInbound` nil, so `changeInputs` returns before that query runs. Now covered by `TestAHeldMessageDoesNotShortenTheGapAReplyBroke`, with an open reply and a held message *inside* the span. Mutation-checked.
- The `RouteStates` test does hold, for the reason Codex gives: the human succeeds, and removing the guard makes the agent succeed too.

Writing the second one, my control arm failed and was right to. With the intervening message open, the gap is 12 days — below `ReplyGapDays`, so no change fires at all. The arms now assert presence and absence rather than two numbers, which is a sharper control than the one I designed.

## Two comment corrections

`RouteStates` reports open asks regardless of requester; only *refusals* are per-requester (`routestate.go:129`). My comment described both as the caller's own. And the test file claimed both guards were unreachable by an agent, which the meeting-prep path above disproves — and claimed both were unexported, which `RouteStates` is not.

## Verification

- Compose integration lane: **1928 pass, 0 fail**, `ok` on the package. All six new tests pass, every sub-arm.
- `deals` and `introductions` integration lanes: green — the new guards break no existing caller.
- `make check-backend`: green except `TestAnEmailIsDrawnOnlyByTheCanonicalComponents`, still failing identically on clean `origin/main` (bisected by another session to #4746). Not this diff.
- Rebased onto 622a82296d and re-verified after the rebase.

Every fix mutation-checked one clause at a time, each mutation confirmed to compile before being trusted.

LarsGradion:
**The MAJOR is closed and the whole gate is green.** Nine fixes now, sixteen files.

## The provider-egress hole

Codex's second pass found that my employer gate could be undone at submission: the enrichment worker re-resolves a subject's identifiers under the connector's own system principal (`integrations/runactor.go:44`), and a system principal passes every object gate — so the employer name and primary domain could still reach an outside vendor for a rep who may not read the account.

Closed with a `RequesterHolds` seam. The question is asked of the **human who queued the run** (`provider_run.requested_by`), live at submission, through `identity.IssuerStillHolds` — the same primitive the share links use, so a grant revoked between the ask and the send is revoked. A run with no requester is the automatic sweep, which has no human's scope to answer for.

Withheld, not refused: the employer is stripped and the rest of the subject still goes. Refusing the whole run would tell the rep the vendor found nothing when a permission stopped us — the same shape `SubjectIdentifiers` returns to a caller without the grant, restored after the connector's re-read undid it. If what remains is unmatchable, the run is skipped with `no_identifiers` rather than sent to be rejected.

New file `integrations/requesterstanding.go`, because the concept is "whose scope does a submission travel under" and it did not fit either file it came from.

## A correction to how I diagnosed the failing test

I reported earlier that `TestAnEmployerLinkedAfterNoIdentifiersIsPickedUpAtOnce` failed because of my change. **That was wrong, and so was my method.** I compared against `origin/main`, which had moved hundreds of commits past my branch's base — so my "clean main passes" control was testing a tree that already carried the fix. The honest control is `HEAD~1`, against which the test fails **without** my commit.

The real cause is #4846, merged today by someone else: the test's helper passed a hard-coded budget of 25 as its ceiling, so once the shared lane database held more than 25 waiting contacts the assertion stopped asking "is this contact covered?" and started asking "is it among the 25 oldest?". My two new test files pushed it over that cliff. Rebasing picked up the fix.

I nearly weakened a correct security gate to make that red go away. Worth recording.

## Craft, not waived

- `execute.go` and `runs.go` crossed the 500-line cap. Split into `requesterstanding.go` rather than waived — **zero waivers**, `runs.go` is now byte-identical to main.
- Removed a dead `queuedFingerprint` variable left from an abandoned approach.
- `TestEveryTouchedFileExplainsWhatItCites` (a gate that arrived on main during this work) flagged a pre-existing `ADR-0061/A107` citation in `store.go`, pulled into scope by my one-line field addition. Fixed the way the gate's own comment prescribes: the A-number labels nothing, so it goes; the decision number stays.

## Verification

| Lane | Result |
|---|---|
| `make check-backend` | **exit 0** |
| `internal/compose/integration` | **1954 pass, 0 fail** |
| `internal/modules/integrations` | **67 pass, 0 fail** |
| people / introductions / deals / person360 | ok |

`check-backend` now includes `TestAnEmailIsDrawnOnlyByTheCanonicalComponents`, which was red on `origin/main` for most of this work — the rebase picked up its fix too, so this branch is green against a green base.

Every fix mutation-checked one clause at a time, each mutation confirmed to compile before being trusted.

PR #4814, opened 2026-09-07
An email's words and who may read them have one owner each, and a thread message is not a second one

### LINKED ISSUE #4792: main is red: composed.tsx draws EmailSummary with its own markup after the one-open-card rewrite
## main is red, and every PR inherits it

`origin/main` at `b5d21b55e` fails one gate in four backend lanes — `deterministic-gates`, `extension-reference`, `integration`, and `integration unit coverage`:

```
--- FAIL: TestAnEmailIsDrawnOnlyByTheCanonicalComponents
emailpresentation_test.go:306: design-system/composed.tsx reads EmailSummary and
draws it with its own markup — render EmailEntry or EmailReference instead, or
ratify the file in emailPassthroughs with what it does with the type
```

Reproduced in a clean worktree at pristine `origin/main` with no branch applied, so it is not branch-specific. **CI tests the merge, so any PR opened now shows four red lanes that are not its own.**

## The mechanism, not just the last-touch guess

- **The gate is old and stable.** `backend/gates/emailpresentation_test.go` arrived in #3914 and was last touched by #4467. It did not change recently.
- **`dbb95e06f` (#4746) rewrote the cited file** — `frontend/src/design-system/composed.tsx`, 454 lines changed — without touching the gate.
- **A red herring worth skipping:** `EmailSummary` was *already* referenced in `composed.tsx` before that commit. That does not exonerate it. The gate's subject is not *reading* the type; it is drawing it with the file's own markup instead of mounting a canonical component.
- **The decisive part.** `emailPassthroughs` does not list `composed.tsx` and never did. It lists `design-system/activitytimeline.tsx` with the reason *"maps an Activity onto a TimelineEntry, email_summary included, **for composed.tsx to draw**"*. So `composed.tsx` was always expected to discharge the obligation through `rendersCanonically`, by mounting `EmailEntry` / `EmailReference`. Turning the thread into one open card with its own markup is exactly the change that stops it mounting one.

## Why this is not a blind fix

Two responses are available and they mean different things:

1. **Re-mount the canonical component**, so the card draws an email through `EmailEntry` / `EmailReference` as before.
2. **Add `composed.tsx` to `emailPassthroughs`** with a sentence saying what it now does with the type — ratifying that this file draws email itself.

Which one is right depends on what the one-open-card design intended, so it belongs to #4746's author rather than to whoever notices the red.

## Note on why #4746's own lane was green

This is the signature of a gate whose subject is a file **outside** the package the gate lives in: it can go red from a commit landing beside it rather than from the change that trips it. Same shape as #4775. Worth remembering when a merged-green PR turns out to own a red.

I am not touching `frontend/` — I have an unrelated PR (#4788, `internal/compose/aicert` only) held rather than merged into a red main.

dungbanguyen:
Fixes https://github.com/margince/margince/issues/4792. **`main` is red and this red blocks every open pull request in the repository.**

```
design-system/composed.tsx reads EmailSummary and draws it with its own markup
  — render EmailEntry or EmailReference instead
```

`craftsmanship` and the backend gates run over the **whole tree** rather than a diff, and `ci` needs both — so this one finding fails `ci` on every PR whatever it touches. It is currently red on https://github.com/margince/margince/pull/4793 too, which is the fix for the other half of https://github.com/margince/margince/issues/4774.

## What was actually wrong

The thread card from https://github.com/margince/margince/pull/4746 draws each message itself, and that is right: the subject, the counterparty and the count sit on the card around it, so a full `EmailEntry` per message would print all three of them three times.

What it *also* did was answer two questions the canonical row already answers:

| Question | The row | The card |
|---|---|---|
| Which text is a message's words, and are there any? | `preview`, suppressed when withheld | read `preview`, drew its own span |
| Who may read it — and is the mark worth drawing? | always | only when not the open default |

Neither copy was wrong on the day it was written. Both were one edit from disagreeing with the row, which is the drift this gate exists to stop.

## The fix

One owner per question: **`EmailWords`** and **`EmailVisibility`**, beside `EmailEntry` in the file that already owns email presentation, mounted by the row and by the thread message.

`place: "row" | "thread"` names which surface is asking, rather than a `className` — a design-system primitive that takes arbitrary classes is a layout its hosts own, and the two rules for the visibility mark are genuinely different, so `place` carries the rule rather than just the paint.

**The paint is unchanged on both surfaces** — the same two class names as before (`emailentry__preview`, `tl-msg-text`), so neither design moves and no CSS is touched.

**One behaviour does change**, and it is the reason to prefer this over ratifying what was there: the words are now refused from the server's own `display_status`. The card previously had only the caller-supplied `TimelineEntry.withheld` flag to go on — a separate optional boolean that travels beside the summary and can disagree with it. The server's word cannot.

## Why `ThreadMessage` is ratified rather than rewritten

It holds no part of a message. It tests `entry.emailSummary` for presence to know the message opens — the exact sentence `screens/openemail.ts` is already ratified for — and delegates the words, the mark and the visibility to the canonical components it wraps.

Saying that needed the waiver list to be able to name a **function**, which is the gate's real gap: `rendersCanonically` was widened to judge per function, and `emailPassthroughs` stayed per file. A whole-file entry for `composed.tsx` would have re-admitted precisely the miss that widening was built to catch — one function discharging the file while another beside it wrote its own markup, which is a shipped bug the gate's own comment records. So keys now take a `file#function` form, the file-wide key is deliberately not consulted on the per-function path, and `composed.tsx#ThreadMessage` carries its reason.

## Verification

- `make check-go` — lint `0 issues`; `TestAnEmailIsDrawnOnlyByTheCanonicalComponents` green, as are the gate's own self-tests.
- `make check-fe` — green (10m). **On Node 24**, which CI pins.
- `make fe-unit` — 624 files, **8 208 passed**, 1 expected fail. `catalog.test.ts` caught the second component missing from the index before I did, which is that gate working.
- `make craft-static` — `PASS — 0 blocker, 0 major, 0 minor` across all four roots.
- The one remaining local failure, `TestPublicTreeCitesNoDocumentGitIgnores`, fails identically on unmodified `origin/main` here: it reads `git check-ignore`, and `.superpowers/` is excluded through `.git/info/exclude` on this machine only.

## Note for @joschua

This edits your thread card from #4746, two hours after it landed, because it was blocking the whole queue. The card's own judgement is kept intact — a message row still draws only its words, for the reason your comment gives. Happy to change the shape if you would rather own it differently.

## Summary by CodeRabbit

* **Improvements**
  * Task and commitment due dates now use the viewer’s local timezone across lists, details, worklists, and date pickers.
  * Email thread previews are limited to two lines, with clearer visibility indicators.
  * Public preference and confirmation responses consistently prevent caching.
  * Transcript review now displays a simplified proposal-count message.
  * Empty task-filtered worklists use the standard empty-state message.

* **Changes**
  * AI email drafting is no longer available from person-originated pages.
  * Transcript-derived meeting labels no longer appear in person timelines.
  * Several obsolete localization messages were removed.


dungbanguyen:
Closing without merging, for two independent reasons.

**It is redundant.** https://github.com/margince/margince/pull/4806 ("A message's words have one spelling, and a thread member uses it") landed while this was in review and fixes https://github.com/margince/margince/issues/4792 with the same shape this reached independently — an `EmailWords` component owning the one reading of a message's words, mounted by the row and by the thread member. `TestAnEmailIsDrawnOnlyByTheCanonicalComponents` passes on a clean checkout of `origin/main` now.

**And it is destructive, which is my error.** I squashed with `git add -A && git reset --soft origin/main`. `origin/main` had moved since this worktree was created, so the reset re-based the commit against a newer main while keeping the older tree — turning every commit that landed in between into a reversion. The diff deletes `publictokencache_integration_test.go` (−324), `compose-person-draft.test.tsx` (−261), `credentialpathcache.go` (−45) and more, and reverts `sonar-project.properties`, which would have undone https://github.com/margince/margince/pull/4793. CI went green because a reverted tree is self-consistent, which is what makes this shape worth naming.

That is also where three of CodeRabbit's four threads came from — `calendarday.ts`, `sectionstimeline.go` and `public_booking_integration_test.go` are not files this change ever touched.

The fourth thread was a real finding on the real change and is worth recording even though the change is not landing: adding `EmailVisibility` to `mountsCanonical` created a false-negative path, because that component draws only a badge — so a function mounting it could draw `preview` or `subject` with its own markup and still pass. Under-recognition is the one way a gate must not break. Had this gone forward, the fix was to keep `mountsCanonical` to the content components and ratify the visibility function by name instead.

Nothing here is needed on `main`. The per-function `emailPassthroughs` key (`file#function`) is the one idea I would still argue for separately — `rendersCanonically` judges per function while its waiver list is per file, so ratifying one function today means waiving a whole file — but that belongs in its own change, against a gate that is currently green, and only if it earns its keep.

PR #4841, opened 2026-09-07
test(schema): the stage-progression ledger's deal reference is classified

dungbanguyen:
**main is red.** `TestFK_rowScopedTargetsHaveVisibilityDecision` fails on the tip:

```
FK stage_progression_outcome.deal_id -> deal has no visibility decision:
a reference to a row-scoped record is a read of it — gate it
(auth.EnsureLinkTarget) or classify it here
```

The `migrations` package runs in several CI jobs, so every open pull request inherits it. Seen on two of mine whose diffs are a gate file and a `.gitignore`.

`stage_progression_outcome` arrived with a `deal_id` and no entry saying what makes that reference safe.

## The classification

**Server-derived.** The ledger row is opened in the same transaction that stages the proposal, naming the deal the progression pass is *already* reading — `facts.DealID` comes off that read, never from a request body. There is nothing there a caller could point elsewhere.

`RecordProgressionProposed` is gated on `deal:update` besides, on its own stated reasoning: opening a ledger row for a deal is a write about that deal, and a caller who may not move it may not put a proposal to move it on the record either.

This is the classification, not a waiver — the entry says why no target probe is owed, rather than excusing a missing one.

## Verification

- `TestFK_rowScopedTargetsHaveVisibilityDecision` passes.
- Sole writer confirmed: `INSERT INTO stage_progression_outcome` appears once, and `RecordProgressionProposed` has one caller.


dungbanguyen:
Closing in favour of #4838, which is 24 minutes earlier and the same one-line classification in the same file.

Same conclusion independently: `stage_progression_outcome.deal_id` is server-derived — the ledger row is opened in the transaction that stages the proposal, naming the deal the pass is already reading, so there is nothing a caller could point elsewhere.

Nothing here worth salvaging. Deleting the branch.

PR #4847, opened 2026-09-07
fix(frontend): the subject scan reads the line once, and an ARIA menu is focusable

dungbanguyen:
Clears the sixteen open findings on the SonarCloud project. Three were defects; the other thirteen are the analyzer resolving a JSX element without the ancestor that decides its role, and each now carries its reasoning in `sonar-project.properties` beside the sixteen waivers already there.

## The defects

**A bracket matcher that re-reads the line** (`typescript:S8786`, `frontend/src/screens/projectrecord.ts`). `stripEveryKeyTag` found groups with `` / ?\[[^\]]*\] ?/g ``, which re-scans the tail of the line at every `[` it then fails to close — so a subject whose brackets all sit after its last `]` costs time quadratic in its length. A subject is attacker-supplied text: it reaches this composer through the reply it seeds.

It now walks the line the way the capture side's own `bracketedGroups` walks it — each `[` to the NEXT `]`, non-overlapping, an unclosed `[` ending the scan. That was already the semantics the regex had, so **the existing spec passes unchanged**; narrowing the character class instead (`[^\][]*`) would have been the tempting one-character fix and would have broken the mirror, because the capture side's inner group *may* contain a `[`. Six cases were added for the shapes the two sides have to agree on, including that one.

**A key shape narrower than the constraint it reads** (found alongside). `keyShaped` was `^[a-z][a-z0-9-]{1,23}$`; `project_key_shape` is `^[A-Za-z][A-Za-z0-9_-]{1,23}$`. So a project keyed `alpha_two` is a key the inbound matcher reads and this side left standing — and writing ours in beside it hands the message two live keys, which files it under neither. That is the exact outcome `stripEveryKeyTag` exists to prevent, missed on one character of the shape.

**Two ARIA menus that could not take focus** (`typescript:S6852` ×2, `frontend/src/app/account.tsx`). Both `role="menu"` panels carried `onKeyDown` and no `tabindex`, so neither could be focused at all — not even programmatically. Both now take `tabIndex={-1}`: the tab stop stays on the roving rows, which is where each panel's own mount effect already lands the reader.

## The waivers

| Rule | File(s) | Why the rule cannot be satisfied here |
| --- | --- | --- |
| `typescript:S6843` ×4, `typescript:S6825` ×3 | `design-system/listtable.tsx` | A `<td>` resolves to `cell` under a table and `gridcell` under a grid. The analyzer sees one element, not the `<table role="table">` around it, so it takes the interactive reading — and then `role="cell"` looks like a demotion and `aria-hidden` on a spacer cell looks like hiding something focusable. Under a table neither is true. The roles are load-bearing: below 720px the stylesheet lays rows out as cards, and changing a table element's `display` is what drops its implicit semantics — `listtable.css` says so where it makes the switch. Removing them hands a phone reader a stack of unrelated divs. |
| `typescript:S6845` ×3 | `design-system/{trust,composed,decisiondeck}.tsx` | WCAG 2.1.1 requires the opposite for a scrollable region: `tabindex="0"` on the scroller is the technique the guideline prescribes, which is why axe raises `scrollable-region-focusable` for the very elements this rule wants left alone. Each site is a scroller or arrow-key surface no descendant can stand in for, and each already carries the biome waiver for the same rule under its other name. |
| `typescript:S6848` ×2 | `design-system/inlinechoice.tsx`, `app/agentrail.tsx` | At both sites the control is a real element *inside* the one carrying the handler. The rail's click widens the button's pointer target to the words beside it — a CSS overlay, the usual way to do that without a handler, would cover a link those words contain. |
| `godre:S8188` ×1 | `cmd/worker/lanes.go` | `laneLifetime` *returns* the cancel so `run()` can defer it before anything starts reading the bus and the pool. Deferring it inside would kill every lane as the helper returned. What the rule protects is held more strictly at the site that can be wrong: `TestALaneCannotStartBeforeItsShutdownIsRegistered` fails when the defer does not *precede* the call that starts the lanes. |

Every waiver names one exact file rather than a directory wildcard, so a fourth scrollable region or a second table re-declaring its roles has to argue for itself here. `lanes.go` calls `context.WithCancel` once, so that file scope hides nothing else.

## Verification

- `make fe-typecheck-composed`, `fe-ds-gates`, `fe-drift`, `fe-lint`, `fe-build` — green.
- `make fe-unit` — 626 files, 8263 passed (a first run flaked on an untouched `leads.test.tsx`; it passes 75/75 alone and the re-run is clean).
- `make check-gates` — green. No Go changed.
- No open security hotspots on the project; the quality gate reads OK.

### DIFF
--- a/frontend/src/app/account.tsx
+++ b/frontend/src/app/account.tsx
@@ -296,6 +296,12 @@ function ThemeSubmenu({
       className="accountsub"
       role="menu"
       aria-label={t("shell.theme")}
+      // -1, not 0: the tab stop belongs to the roving rows below, and a menu
+      // that took one of its own would stand in front of the choices it exists
+      // to offer. What a `menu` role owes the keyboard is to be focusable AT
+      // ALL — it promises somewhere to stand, and a div carrying no tabindex
+      // cannot be focused even programmatically.
+      tabIndex={-1}
       ref={panel}
       onKeyDown={onKeyDown}
     >
@@ -437,6 +443,10 @@ function AccountPanel({
           ? `${t("shell.accountAria")} — ${identity.spoken}`
           : t("shell.accountAria")
       }
+      // Focusable, and only programmatically: the roving tabstop lives on the
+      // rows, and the effect above lands the reader on one of them the moment
+      // this opens. See ThemeSubmenu for why the value is -1 rather than 0.
+      tabIndex={-1}
       ref={panel}
       onKeyDown={onKeyDown}
     >
--- a/frontend/src/screens/projectrecord.test.ts
+++ b/frontend/src/screens/projectrecord.test.ts
@@ -119,6 +119,54 @@ describe("the project tag a subject carries", () => {
     );
   });
 
+  it("clears a key carrying the underscore the CHECK admits", () => {
+    // project_key_shape is `^[A-Za-z][A-Za-z0-9_-]{1,23}$`, so `alpha_two` is a
+    // key the capture matcher reads. Left standing it is a SECOND live key
+    // beside the one written here — and a subject naming two files under
+    // neither, which is the exact outcome this function exists to prevent.
+    expect(withSubjectTag("[alpha_two] Hallo", "[N2P-1]")).toBe(
+      "[N2P-1] Hallo",
+    );
+  });
+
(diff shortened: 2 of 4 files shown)


dungbanguyen:
@coderabbitai review

PR #4857, opened 2026-09-07
A disqualified lead can be reopened

### LINKED ISSUE #1888: Reopen a disqualified lead: no endpoint, and the page cannot offer it
Deferred from #1883.

The lead page says "Disqualified: <reason>" but cannot offer "Reopen", because there is no server path to put a disqualified lead back on the open ladder. Demote (`POST /leads/{id}/demote`) covers a promoted lead only.

What to do: add `POST /leads/{id}/reopen` (human-only, lead:update + lead:delete like disqualify), clear `archived_at`, `disqualify_reason_id` and `disqualify_note`, restore the last open status (audit the before-image; `engaged` when unknown), emit `lead.updated`, then add the button to the stepper's explanation line and the list's Disqualified view.

dungbanguyen:
Closes #1888.

## What was missing

The lead page said "Disqualified: <reason>" and could not offer the way back,
because there was no server path. `POST /leads/{id}/demote` reverses a
**promotion** — it answers about a lead that became a person — and nothing
reversed a closure.

A judgement that somebody is not worth pursuing is exactly the kind that
changes: the budget arrives, the champion comes back. With no way back the
operator re-keys the lead, and that loses its history and its score along with
its reason.

## The status is read, not guessed

`POST /leads/{id}/reopen` clears `archived_at`, `disqualify_reason_id` and
`disqualify_note`, and returns the lead to the status it held when it was
closed — read from the disqualify's own audit row.

Re-deriving it from today's activity would answer about the lead **as it is
now**: one disqualified at `contacted` and since emailed twice would come back
`engaged`, a claim nobody made about a lead nobody was working. The status a
lead had reached is a fact the trail already holds.

**The fallback, and why it is not a refusal.** A row whose trail no longer names
a status reopens to `engaged`:

- `new` would say nobody had ever touched it, which is false of every lead that
  reached a disqualification;
- refusing would leave a dead row whenever the audit log aged out under its own
  retention — a lead nothing can reopen because of a record it does not own.

A status the ladder no longer carries falls back the same way, rather than
restoring a word no filter can name. `promoted` and `disqualified` are excluded
for a sharper reason than being terminal: either would clear the archive while
leaving the lead reading as closed, which is a row the page cannot explain.

## Authority and conflict

Both grants the disqualify takes — `lead:update` **and** `lead:delete` — because
a caller who may not close a lead has no business deciding a closure was wrong.

Reopening a lead that is not disqualified is a **409**. Succeeding silently would
tell a caller their undo did something, and would leave `disqualified` as this
lead's remembered status for the next reopen.

## On the page

The button sits beside the sentence that says the lead is closed, offered only
on a disqualification the reader may write — the promoted closure is the
demote's to reverse.

`leads.test.tsx`'s read-only sweep asserts that a disqualified lead exposes **no
enabled mutation anywhere**, deliberately over every button so a control added
later is caught by construction. It now names this one as the exception, in its
own set rather than among the view controls, with the reason: reopening is not a
change to the record, it is the undo of the state that makes the record
read-only. A separate test proves the button is enabled and actually calls the
endpoint — the sweep passes with it inert, because an exempt control is exempt
whether or not it works.

## Checks

- `make check-backend`, `make check-fe` green.
- `make test-it DIR=backend/internal/modules/people` green in full (119s).
- `craft static --strict` clean.
- Mutation-checked: forcing the fallback turns the restore test red on both the
  wire and the live re-read; removing the not-disqualified guard turns the
  second-reopen case red; hiding the button turns the page test red.

One thing found and **not** fixed here: `src/screens/worklist.today.test.tsx >
"submits once however fast the reader presses"` is flaky under the parallel
suite — it failed CI on #4822 (a contracts change nowhere near it), failed once
here, and passes three-for-three in isolation. Filed rather than fixed: it is
another screen's double-submit race and diagnosing it would double this diff.

### DIFF
--- a/backend/api/crm.yaml
+++ b/backend/api/crm.yaml
@@ -7510,6 +7510,46 @@ paths:
             application/problem+json:
               schema: { $ref: '#/components/schemas/Problem' }
 
+  /leads/{id}/reopen:
+    parameters:
+      - $ref: '#/components/parameters/Id'
+    post:
+      tags: [Leads]
+      operationId: reopenLead
+      # Human-only, on the same ground the disqualify it reverses is: deciding a
+      # lead is worth pursuing again is a judgement, and the governed tool would
+      # need its own registry entry and approvals authority.
+      x-agent-access: human-only
+      summary: Put a disqualified lead back on the open ladder.
+      description: |
+        The reverse of `DELETE /leads/{id}`. It clears the archive, the reason and the note,
+        and returns the lead to the status it held when it was disqualified — read from that
+        act's own audit row rather than guessed at, because the status a lead had reached is
+        a fact the trail already holds and re-deriving it would answer about today's data.
+
+        A row whose audit trail no longer names a status reopens to `engaged`, which is what a
+        lead somebody is choosing to pursue again is: `new` would claim nobody had touched it,
+        and refusing would leave a lead nothing can reopen because of a record it does not own.
+
+        Emits `lead.updated`. Reopening a lead that is not disqualified is a **409**: the verb
+        answers about a state the lead is not in, and silently succeeding would tell a caller
+        it had undone something.
+      parameters:
+        - $ref: '#/components/parameters/IdempotencyKey'
+      responses:
+        '200':
+          description: The reopened lead.
+          content:
+            application/json:
+              schema: { $ref: '#/components/schemas/Lead' }
+        '403': { $ref: '#/components/responses/Forbidden' }
+        '404': { $ref: '#/components/responses/NotFound' }
+        '409':
+          description: The lead is not disqualified, so there is nothing to reopen.
+          content:
+            application/problem+json:
+              schema: { $ref: '#/components/schemas/Problem' }
+
   /leads/{id}/score:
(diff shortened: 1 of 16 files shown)


dungbanguyen:
SonarCloud stays red on one condition: `new_duplicated_lines_density` 4.2% against a 3% threshold, from 12 lines in `leadreopen.go`. `ci` is green.

Merging with it, and the reason is in the code rather than only here. The 12 lines are the transaction skeleton `ReopenLead` shares with the `DisqualifyLead` it reverses — probe, lock, read, guard, patch, audit, emit, re-read — which is this module's write shape and which `storekit` already owns the halves of. What differs is every decision inside it, the liveness first among them: this verb acts only on archived rows and its sibling only on live ones, so a helper covering both would take that as a parameter and be a switch between two verbs wearing one name. That hides the fact that they are opposites, which is the thing a reader most needs to see.

`leadreopen.go` says so where the next reader meets it, which is what the rulebook asks of two writers of one shape: share a helper, or say why not in the code beside it.

PR #4862, opened 2026-09-07
The seeder's version guards go where the server reads them

### LINKED ISSUE #1867: seed-demo: if_version in a PATCH body is silently ignored, so the seeder's version guards do not guard
## What is wrong

`backend/tools/seed-demo` guards several PATCH writes with `if_version` in the
JSON body. That field is accepted and ignored — these endpoints read optimistic
concurrency only from the `If-Match` header — so the writes are last-write-wins
and the guard is decorative.

## Verified against a running stack

```
PATCH /v1/organizations/{id}   body {"if_version": 999999, ...}   -> HTTP 200, written
PATCH /v1/organizations/{id}   header If-Match: 999999            -> HTTP 422
```

999999 is a version that cannot exist. The body form wrote anyway.

## Where

- `ownership.go` — `setOrganizationOwner` and `setStaffOwner`
- `records.go` — `seedLifecycle`

`address.go` had the same bug and is fixed on the branch that found this: it
uses a new `client.patchGuarded`, which sends the header.

## What it means if nobody fixes it

Two seeder runs against one installation, or a seeder run alongside somebody
editing records in the UI, can silently overwrite the other's change. The
symptom is an owner or a lifecycle stage that reverts for no visible reason,
which is exactly the failure the guard was written to prevent — and the code
reads as though it is protected.

## Why it was deferred

Found while reviewing an unrelated change (the company address fill). The fix
is mechanical — switch those call sites to `patchGuarded` — but it touches
ownership and lifecycle, which every seeded record passes through, and it
deserves its own run of the seeder to confirm nothing starts failing 422 on a
legitimately concurrent write.

dungbanguyen:
Closes #1867.

## The defect

Four seeder PATCH writes spelled their optimistic-concurrency guard as
`if_version` in the JSON body. These endpoints read it from the **`If-Match`
header only**, so the field was accepted and ignored — the ticket's own
verification says it plainly:

```
PATCH /v1/organizations/{id}  body {"if_version": 999999, ...}  -> HTTP 200, written
PATCH /v1/organizations/{id}  header If-Match: 999999           -> HTTP 422
```

The writes were last-write-wins while the code read as though they were not.
Nothing failed: the seeder ran, the rows were written, and the symptom is an
owner or a lifecycle stage that reverts for no visible reason on the run where
two writers overlap — exactly what the guard was written to prevent.

All four now go through `patchGuarded`, which sends the header.

## Four, not the three filed

`ownership.go` (two sites) and `standing.go`'s lifecycle write are the ones the
ticket names. `relationships.go` retypes a company the same way and had the same
bug — found by grepping the key rather than the call sites, which is the
difference between fixing the invariant and fixing what was reported.

## Why it needs a gate rather than four edits

A fifth site would arrive exactly as this one did: a reviewer noticing, on an
unrelated change, that a guard is decorative. Nothing else fails.

`TestNoSeederWriteSpellsItsVersionGuardInTheBody` matches the key as a string
**literal** rather than anywhere in the source. That distinction is deliberate:
this package explains the trap in prose beside `patchGuarded` and `postGuarded`,
and a census that flagged the explanation would teach the next author to delete
the sentence that stops them repeating it. It also carries a floor — it walks a
directory, so a package move would otherwise leave it green over a tree it never
read.

Mutation-checked: putting one site back turns it red, naming the file, the line
and the writer to use instead.

## Checks

- `make check-backend` green; `craft static --strict` clean on all four files.

### DIFF
--- a/backend/tools/seed-demo/ownership.go
+++ b/backend/tools/seed-demo/ownership.go
@@ -148,8 +148,8 @@ func setOrganizationOwner(c *client, orgID, ownerID string) (bool, error) {
 	if current.OwnerID == ownerID {
 		return false, nil
 	}
-	body := jsonBody{"owner_id": ownerID, "if_version": current.Version}
-	if err := c.patch("/v1/organizations/"+orgID, body, nil); err != nil {
+	body := jsonBody{"owner_id": ownerID}
+	if err := c.patchGuarded("/v1/organizations/"+orgID, current.Version, body, nil); err != nil {
 		return false, err
 	}
 	return true, nil
@@ -183,8 +183,8 @@ func setStaffOwner(c *client, orgID, ownerID string, mode runMode) (int, error) 
 		if current.OwnerID == ownerID {
 			continue
 		}
-		body := jsonBody{"owner_id": ownerID, "if_version": current.Version}
-		if err := c.patch("/v1/people/"+personID, body, nil); err != nil {
+		body := jsonBody{"owner_id": ownerID}
+		if err := c.patchGuarded("/v1/people/"+personID, current.Version, body, nil); err != nil {
 			return changed, fmt.Errorf("owning person %s: %w", personID, err)
 		}
 		changed++
--- a/backend/tools/seed-demo/relationships.go
+++ b/backend/tools/seed-demo/relationships.go
@@ -87,8 +87,8 @@ func seedRelationshipTypes(c *client, cfg demoConfig, refs pipelineRefs, mode ru
 		if sameTypes(current.Types, want) {
 			continue
 		}
-		body := jsonBody{"relationship_types": want, "if_version": current.Version}
-		if err := c.patch("/v1/organizations/"+orgID, body, nil); err != nil {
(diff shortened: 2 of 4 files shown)


dungbanguyen:
Closing unbuilt: the subject left this repository. `backend/tools/seed-demo` was removed in #4732 (`abfe6879f`) while this was open — the loader read a dataset that lives in a separate private repo, so it moved to sit beside the data.

The defect is still real, and it is worth restating for whoever owns the loader now, because nothing over there will find it on its own: four PATCH writes spell their optimistic-concurrency guard as `if_version` in the JSON body, and those endpoints read it from the `If-Match` header only. The field is accepted and ignored, so the writes are last-write-wins while the code reads as though they are guarded. `ownership.go` (two sites), `standing.go`'s lifecycle write and `relationships.go`'s retype; `patchGuarded` in `apiclient.go` is the writer that sends the header.

Nothing of this branch is worth carrying over into this repo — the gate it added has no subject here either.

Issue #4761, opened 2026-09-07
Enrol internal/compose in the entry-point authorization gate


LarsGradion:
`TestEveryStoreEntryPointIsAuthGated` scans `internal/modules` and `internal/platform/settings`. The whole `internal/compose` tier is outside it, ratified by 36 file-level entries in `entryPointsOutsideModules` — several of which say in their own words that the tier has not been judged, not that it is correct.

`composerowscope_test.go` covers part of the gap, but only the row-scope half: its header says so. It proves nothing about the object, the action, the audience, the write probe, or the projection.

## What enrolling costs, measured

Adding `internal/compose` to `storeEntryPointScope.Roots` on top of #4756:

- **14 entry points** report as ungated
- **37 file exemptions** become redundant and would be deleted

The 14:

```
attention:      Assemble, Worklist
dealstatus:     Get
org360:         Assemble, AssembleScoped, ContactPage, Graph, KeepUndismissed
project360:     Assemble
worklistsnap:   Freeze, Resume
compose:        HandleEvent, Resolve, Tick
```

## Why it is not a mechanical change

Most of these are safe today for a reason the gate structurally cannot see: they delegate to an already-gated module read in another package, and the gate resolves call edges by name **within one package**. `org360.AssembleScoped` opens with `people.GetOrganizationTx`; `project360.Assemble` with `projects.GetProjectTx`; `dealstatus.Get` gathers under the caller's row scope before touching a cache. Each is protected, and each would report ungated.

So enrolling the tier means one of:

1. Teach the gate cross-package resolution for compose's calls into modules. Most faithful, most work, and it widens what the gate must parse.
2. Require each compose entry point to state its own admission, even when a delegate already checks. Defensible as defence in depth, and it makes the tier's authorization readable where it is entered.
3. Waive all 14 with per-entry-point reasons naming the delegate that gates them. Cheapest, and it converts 36 file waivers into 14 sharper ones — but a waiver naming a delegate goes stale when the delegate moves, and nothing would notice.

Option 2 is probably right, and it is a judgement per entry point rather than a sweep. Worth doing with the tier's owner rather than alone.

Deferred from the access-rights hardening series (#4727, #4737, #4747, #4756), where the plan had it as a third PR. The measurement above is current as of #4756.

LarsGradion:
All 14 have now been traced. **None is unprotected** — every one is a gate gap, not a live defect. That changes which option is right.

## What protects each, verified

| Entry point | Kind | Protected by |
[...]

PR #4863, opened 2026-09-07
A composite read states its own admission

LarsGradion:
Ten entry points in `internal/compose` reached no auth gate `TestEveryStoreEntryPointIsAuthGated` could see. All were protected — by a delegate in another package, which is exactly what the gate cannot follow: it resolves call edges by name **within one package**, so an admission one import away reads as none.

That is a real gap and not only a reporting one. An admission that lives entirely in a delegate is one refactor away from living nowhere, and no test fails when it moves. This is option 2 from #4761, chosen by the PO.

| Entry point | Now asks |
|---|---|
| `org360.AssembleScoped`, `ContactPage`, `Graph` | `organization:read` |
| `project360.Assemble` | `project:read` |
| `dealstatus.Get`, `CachedCards` | `deal:read` |
| `attention.Assemble`, `Worklist` | `RequireMember` |
| `worklistsnap.Freeze`, `Resume` | `RequireHuman` |
| `RunnerService.Tick` | `RequireSystem` |

## Two of them found something rather than restating it

**`dealstatus.CachedCards` never asked the object question at all.** It serves the card cache without ever calling `gather`, so `Get`'s admission covered the path that FILLS the cache and not the one that READS it.

**`attention.Assemble` answered a caller carrying no principal.** Every lane gates for itself and a refused lane is omitted and named — so fourteen separate declines rendered as a clear day, which is also exactly what a reader with nothing to do sees. `TestThePageRefusesACallerWithNoSeat` holds it now, with a seated-reader control so the refusal cannot pass against a fixture that answers nobody.

That second one also explains the 21 changed test files: those lane suites passed `context.Background()` and were exercising the page as a caller no door can produce. They now read as `pageReader()`, a seated human with workspace row scope — the widest ordinary reader, so a lane assertion fails on the lane's own rule rather than on a scope the fixture picked.

## `auth.RequireSystem` is the opposite of the others

A scheduled pass has no user to authorize: it binds the system principal itself and reads on the installation's own authority. Asking `auth.Require` for an object grant there would be theatre — the system principal holds every one — and `RequireHuman` would refuse the scheduler, the only thing that legitimately calls it.

What it turns into a check is the **assumption**. `Tick` bound `schedulerContext` and then only assumed it was the system; a request context reaching it would have been served past every object and row gate, because a caller that never asks is never refused.

`worklistsnap.readerOf` keeps its own id check beside `RequireHuman`: the principal-type half is stated once in `auth` now, and the id half is the part that helper does not answer.

## What is NOT enrolled yet, and why

The gate's roots are unchanged. Three entry points remain and each needs its own judgement:

- **`AnalyticsShareStore.Resolve`** has a genuine two-factor gate — the token digest plus `identity.IssuerStillHolds`, so a revoked seat's links die — spelled in no verb the gate reads. That is a waiver with a real reason, not a missing check.
- **The `HandleEvent` surface**: 14 consumers. Eleven bind a system principal and would take `RequireSystem` mechanically; **three bind nothing at all** (`captureenrichtrigger`, `orgautoenrich`, `vcardingesttrigger`) because they parse an event and enqueue a job rather than touching a domain row. Those three deserve a decision rather than a sweep.

Enrolling the roots is the last step and belongs with those three, so the gate goes green in the same change that makes it true.

## Craft

`attention/feed.go` crossed the 500-line cap, so `Assemble` moved to `attention/assemble.go` — the page's entry point and its one admission, which is a coherent file. **Zero waivers.**

`TestEveryTouchedFileExplainsWhatItCites` (a gate that landed on main during this work) flagged four pre-existing `A107/ADR-0061` and `ADR-0090/A135` citations in `rbacgate_test.go`, pulled into scope because I edited the file. Fixed as that gate's own comment prescribes: the retired A-number labels nothing and goes, the decision number stays.

## Verification

`make check-backend` — **exit 0**, including `go-file-length`, the uniqueness census and the citation gate. Compose integration lane running; result to follow before merge.

Refs #4761.

---
## Summary by cubic
Adds explicit auth checks to `internal/compose` entry points whose admission used to live only in a delegate, so the entry-point gate can see them and the checks can't drift. Also closes three real gaps: `CachedCards` served the cache with no object read check at all, `Assemble` served the attention page to callers carrying no principal, and `Tick` would have served a scheduled pass to a request context that reached it.

**What changed**
- `org360.AssembleScoped`, `ContactPage`, `Graph`, and `Coverage` now ask `organization:read`; `person360.AssembleScoped` asks `person:read`; `project360.Assemble` asks `project:read`; `dealstatus.Get`, `CachedCards`, and `CachedMoves` ask `deal:read`; `attention.Assemble` and `Worklist` ask `RequireMember`; `worklistsnap.Freeze` and `Resume` ask `RequireHuman`; `RunnerService.Tick` asks `RequireSystem`.
- Adds `auth.RequireSystem`, which admits only the system principal; a scheduled pass binds it itself, so this turns that assumption into a check, and `Tick` now refuses a non-system actor before binding.
- A member without `deal:read` keeps their worklist rows but loses the suggested next step and deal standing, which are treated as enrichment rather than the row itself.
- `worklistsnap.readerOf` keeps its id check beside `RequireHuman`; the principal-type half is now stated once in `auth`.
- Lane tests now read the page as `pageReader()`, a seated human with workspace scope, so the suites exercise the page as a real caller would.
- `AnalyticsShareStore.Resolve` and the `HandleEvent` surface stay unenrolled — the first has a real token-plus-live-issuer gate the gate can't read, and the latter's three no-principal consumers need a decision rather than a sweep.

Refs #4761.

<sup>Written for commit b6cdb347a2f1e9f71e918165b9cbfc4306af0e60. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/4863?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Security**
  - Added authorization checks across attention, worklist, deal, organization, and project views.
  - Restricted scheduler operations to system-level callers.
  - Unauthorized requests now fail before protected data is read.

- **Behavior**
  - Attention pages require an authenticated member.
  - Refused lane reads are clearly identified rather than shown as empty.
  - Cached deal data now respects read permissions.

- **Tests**
  - Expanded coverage for unauthorized access and page-reader-based attention assembly.


LarsGradion:
**I introduced a regression in this PR and caught it before review came back. Fixed, with a test that actually holds it.**

## The regression

`worklistsnap.readerOf` checked `actor.Type != principal.PrincipalHuman` directly. I replaced that with `auth.RequireHuman`, on the assumption they say the same thing.

They do not. `auth.RequireHuman` (`rbac.go:125`) refuses a buyer and an agent and **admits `PrincipalSystem` and `PrincipalConnector`** — both of which the original check refused. So this PR, whose whole subject is tightening admissions, briefly loosened one.

It matters here specifically: a snapshot binds ONE reader and is keyed on `reader_id`. A background pass carrying somebody's user id could have resumed that person's walk — the exact shared-row case the function's own doc comment says it exists to prevent.

`RequireHuman` stays, because it is what lets the entry-point gate see an admission at all, and it does refuse the two principals this walk is most likely to meet. The explicit type test is restored beside it, with a comment saying why it is not a second spelling of the same question.

## The test that lied, and how

I wrote a six-case table covering every principal type. It passed. Then I reverted the guard — and **it still passed**.

The system and connector cases carried no `UserID`, so the surviving `UserID.IsZero()` clause refused them for a reason that had nothing to do with the check under test. The table looked exhaustive and proved nothing about the clause it was named for.

Giving those two cases a real user id isolates it. The mutation is now caught, on both, by name:

```
--- FAIL: TestOnlyAPersonHoldsAWalk/the_system_principal_carrying_a_user_id
    readerOf admitted the system principal carrying a user id → <nil>, want ErrPermissionDenied:
    auth.RequireHuman ADMITS it, so only the explicit type test refuses it
--- FAIL: TestOnlyAPersonHoldsAWalk/a_connector_carrying_a_user_id
```

This is the same defect class this whole series has been fixing — a guard that appears tested and is not — and I wrote it myself. Worth recording rather than quietly amending: the only thing that caught it was mutation-checking every guard, including the ones that look obviously correct.

## Verification after the fix

- `make check-backend` — **exit 0**
- `internal/compose/integration` — **1954 pass, 0 fail** (re-run AFTER the fix; the earlier green predated it and did not cover the changed behaviour)
- `internal/compose/worklistsnap` — 8 cases, every principal type, all passing

Note for anyone reproducing: this branch's worktree was swept mid-run by a parallel session. Nothing was lost because the work was committed before the lane started — the branch carried `c4d21950bf` intact and a fresh worktree recovered it.

LarsGradion:
**Codex found three MAJORs. All three were real; all three are fixed.** The first would have shipped a 403 to real users.

## MAJOR 1 — my own check broke the worklist

`deal:read` on `CachedCards` propagated straight up through `attentionseam.go:433` into `worklist.go:189`, which fails the **whole page** on a standing-enrichment error.

That is not a corner case. Role grants are edited per object (`identity/roles.go:132`), so a role holding `activity:read` without `deal:read` is one an administrator can write — and the activity lane admits a task on `activity:read` alone (`activities/activityread.go:104`), so a task linked to a deal reaches that member's queue. My check turned their legitimate worklist into a 403.

Fixed with the pattern already in this tree one file over: `attentionNames.Labels` (`attentionnames.go:72`) treats a permission refusal as "no labels" and propagates anything else. Both standings and moves now do the same. **A refusal costs the enrichment; a database fault still fails loudly** — "you may not see this" and "the read is broken" must not render identically.

## MAJOR 2 — `RequireSystem` on `Tick` was decoration

I placed it AFTER `ctx = schedulerContext(ctx)`, which overwrites the actor. A human, agent or buyer handed to `Tick` was converted to the system principal and then admitted by the check meant to refuse it. My own comment gave it away — *"It cannot fail for the scheduler"* — when in fact it could not fail for anybody.

`Tick` now refuses a foreign actor **before** the binding, and keeps `RequireSystem` after it so the entry-point gate still sees an admission and a future edit to `schedulerContext` cannot quietly stop making one.

## MAJOR 3 — the same shape, left half-done

`CachedCards` was fixed and its twin `CachedMoves` was not, though both serve the same per-user cache and neither gathers. Two sibling composite reads were also missed — `org360.Coverage` and `person360.AssembleScoped` — both relying, exactly like the three org360 reads I did fix, on a cross-package `Get…Tx` the gate cannot see. The gate's own waiver catalog named them as siblings.

All four now carry the admission. Both cache readers needed the enrichment absorption as well, or MAJOR 1 would have shipped twice.

## Testing, and a second guard that passed by eye and failed by mutation

- `CachedMoves`/`CachedCards` refusal: `TestTheCachedDealReadersRefuseAMemberWithoutTheDealGrant`, with an `activity`-only reader against a granted control. Mutation-checked.
- The absorption — the half that actually prevents the 403 — **had no coverage at all**. Reverting it changed nothing. `TestADealRefusalCostsTheStepAndNotTheQueue` now covers it: an answering control proves the seam fires, a refusal keeps the page without the step, and a plain error still propagates. Mutation-checked in both directions.

That is the second guard in this PR to survive review-by-eye and fail mutation. It is the only reason either was found.

## MINOR items

The zero-length early return in `CachedCards` (`cards.go:82`) precedes the admission. It discloses nothing — it returns an empty map for an empty request — and I have left it, because moving the check above it would spend a permission read on a call that touches no data. Flagging rather than silently disagreeing.

The commit message is rewritten: it claimed `RequireHuman` states worklistsnap's type check (it does not — `readerOf` still owns it), claimed `Tick` refuses request contexts (it did not, until now), and said three entry points remained without naming these siblings.

## Verification

| Lane | Result |
|---|---|
| `make check-backend` | **exit 0** |
| `internal/compose/integration` | **1957 pass, 0 fail** |
| `internal/modules/deals` | ok |
| `internal/modules/people` | ok |

Issue #4225, opened 2026-09-04
projects-by-phase reads empty for a manager or rep, with a real delivering project underneath


tinnhlt-gra:
## Observed

Same root cause as #4205, #4206, #4207, #4209, #4210, #4211 and (win-loss,
filed alongside this one): core `301627091` (#4170) made the generic report
engine apply population scoping consistently, and any prebuilt report spec
declaring no `population:` silently takes the zero-value default,
`measureCallersOwn`. `projects-by-phase`
(`backend/internal/compose/reportprojects.go`) declares no `population:`
either — despite `measureEveryReadableRow`'s own doc comment
(`reportpopulation.go`) using PROJECTS as its illustrative example of a
report that SHOULD opt out ("a delivery lead... their department has no
projects running").

Reproduced live on an isolated dev stack, with a fixture built specifically
to tell "no data" from "wrong population" apart — a real project (phase
`delivering`, reached there by winning its own linked deal through the
ordinary advance path), owned by one seat (`uat-rep@demo.test`) who is on no
team:

| Seat | Role / row-scope | `POST /v1/reports/projects-by-phase` (`filters: {owner_id: <rep>}`) |
|---|---|---|
| `admin@demo.test` | `admin`, all-scope | `total_rows: 1` — the fixture project, `phase: delivering` |
| `uat-manager@demo.test` | `manager`, team-scope, not this project's owner and not a teammate of its owner | `total_rows: 0`, `rows: []` |

`uat-manager` can read the project itself directly (`GET /v1/projects/{id}`
answers 200 for the same seat, same run) — the row is not hidden from this
seat by row scope, only from this one aggregate.

## Cost

`projects-by-phase` reaches no web screen today (grep of `frontend/src/`
for the literal report key finds nothing outside generated API types), so
this is not currently a silent-wrong-screen defect for a human — same shape
as #4210. It IS reachable by an agent calling the generic `run_report` tool
surface, and it is the report the product's own source comment names as the
motivating case for opting OUT of this narrowing — a delivery manager asking
"how many projects are in delivery" would be told their department has none,
identical to the scenario that comment was written to prevent.

## Workaround

Read as `admin`/`management`/`ops` (all-scope), whose population is
unaffected. There is no web screen to route a human around today.

---
Found by margince-qc, building a fixture (a real won deal, a real project
already reached `delivering`) specifically to settle the four `reportspecs.go`
entries this dev stack's earlier sweep could not verify live because they
[...]

Issue #4224, opened 2026-09-04
win-loss reads empty for a manager or rep, with a real won deal underneath


tinnhlt-gra:
## Observed

Same root cause as #4205, #4206, #4207, #4209, #4210 and #4211: core
`301627091` (#4170) made the generic report engine apply population scoping
consistently, and any prebuilt report spec declaring no `population:`
silently takes the zero-value default, `measureCallersOwn`. `win-loss`
(assembled by `winLossSpec()` in `backend/internal/compose/reportperiod.go`)
declares no `population:` either.

Reproduced live on an isolated dev stack, with a fixture built specifically
to tell "no data" from "wrong population" apart — a real won deal, owned by
one seat (`uat-rep@demo.test`) who is on no team:

| Seat | Role / row-scope | `POST /v1/reports/win-loss` (`filters: {owner_id: <rep>, status: won}`) |
|---|---|---|
| `admin@demo.test` | `admin`, all-scope | `total_rows: 1` — the fixture deal, `status: won` |
| `uat-manager@demo.test` | `manager`, team-scope, not this deal's owner and not a teammate of its owner | `total_rows: 0`, `rows: []` |

`uat-manager` can read the deal's own project directly (`GET
/v1/projects/{id}` answers 200 for the same seat, same run) — the row is not
hidden from this seat by row scope, only from this one aggregate.

## Cost

`win-loss` reaches no web screen today (grep of `frontend/src/` for the
literal report key finds nothing outside generated API types), so this is
not currently a silent-wrong-screen defect for a human — same shape as
#4210. It IS reachable by an agent calling the generic `run_report` tool
surface, which would read a team's own win/loss history as empty when it is
not.

## Workaround

Read as `admin`/`management`/`ops` (all-scope), whose population is
unaffected. There is no web screen to route a human around today.

---
Found by margince-qc, building a fixture (a real won deal, a real project
already reached `delivering`) specifically to settle the four `reportspecs.go`
entries this dev stack's earlier sweep could not verify live because they
read empty for `admin` too, with no data underneath either reading. Not a
security or privacy issue — a narrower population is measured, not a wider
one leaked.

tinnhlt-gra:
Re-checked against core `225119443`.

[...]

Issue #4210, opened 2026-09-04
stage-age reads empty for a manager or rep, with real open deals underneath


tinnhlt-gra:
## Observed

Same root cause as #4205, #4206, #4207 and #4209: core `301627091` (#4170)
made the generic report engine apply population scoping consistently, and
any prebuilt report spec declaring no `population:` silently takes the
zero-value default, `measureCallersOwn`. `stage-age`
(`backend/internal/compose/reportspecs.go`) declares no `population:` either.

Reproduced live on an isolated dev stack, same two seats used for the four
issues above:

| Seat | Role / row-scope | `POST /v1/reports/stage-age` (`group_by: [stage_id]`) |
|---|---|---|
| `admin@demo.test` | `admin`, all-scope | `total_rows: 2` — 5 deals in one stage, 1 in another |
| `uat-manager@demo.test` | `manager`, team-scope, no deal owned by them or a teammate | `total_rows: 0`, `rows: []` |

## Cost

Lower than #4205/#4206/#4207/#4209: `stage-age` reaches no web screen today
(grep of `frontend/src/` finds it only in the generated API types), so this
is not currently a silent-wrong-screen defect for a human. It IS reachable
by an agent calling the generic `run_report` tool surface, which would read
a team's stalled-deal aging as empty when it is not, and would misinform any
report the product ever adds a screen for on top of this key.

## Workaround

Read as `admin`/`management`/`ops` (all-scope), whose population is
unaffected. There is no web screen to route a human around today.

---
Found by margince-qc, sweeping the `reportspecs.go` catalog for the same
shape after #4205/#4206/#4207 were each found independently. Not a security
or privacy issue — a narrower population is measured, not a wider one leaked.

tinnhlt-gra:
Re-checked against core `225119443`.

`6402000a1` ("fix(reports): an unowned row is a caller's population too", #4869) is a multi-commit, security-reviewed PR. Its first commit widened `stage-age` to `measureEveryReadableRow`; its third commit ("security: keep own/team lens on reports where owner_id has no other narrowing") reverted that, since `stage-age` reads an identity-adjacent table where row scope renders unconditionally true and the population clause was the only thing standing between a rep and a named colleague's exact figures. Confirmed directly in current source: `stage-age` still declares no population override.

[...]

Issue #4209, opened 2026-09-04
Analytics' Open-deals-per-company table reads empty for a manager or rep, with real open deals underneath


tinnhlt-gra:
## Observed

Same root cause as #4205, #4206 and #4207: core `301627091` (#4170) made the
generic report engine apply population scoping consistently, and any
prebuilt report spec that declares no `population:` silently takes the
zero-value default, `measureCallersOwn`. `open-deals-per-company`
(`backend/internal/compose/reportspecs.go`) declares no `population:` either.

Reproduced live on an isolated dev stack, same two seats used to confirm
#4205/#4206/#4207:

| Seat | Role / row-scope | `POST /v1/reports/open-deals-per-company` (`group_by: [organization_id]`) |
|---|---|---|
| `admin@demo.test` | `admin`, all-scope | `total_rows: 1`, `open_deals: 6` |
| `uat-manager@demo.test` | `manager`, team-scope, no deal owned by them or a teammate | `total_rows: 0`, `rows: []` |

Same stack, same call, only the caller differs.

This is the fourth confirmed screen this default reaches: the Analytics
screen's own "Pipeline" section renders `open-deals-per-company` beside
`pipeline-current` (also #4207) as the "Open deals per company" table
(`frontend/src/screens/analytics.tsx`, `CompanyTable`).

## Cost

A manager or rep opening Analytics' company breakdown sees an empty table
the moment no open deal happens to be owned by them or a direct report —
identical in shape and cost to #4205/#4206/#4207. Nothing on screen
distinguishes "no open deals anywhere" from "the report measured the wrong
population."

## Workaround

Read as `admin`/`management`/`ops` (all-scope), whose population is
unaffected.

---
Found by margince-qc, sweeping the `reportspecs.go` catalog for the same
shape after #4205/#4206/#4207 were each found independently. Not a security
or privacy issue — a narrower population is measured, not a wider one leaked.

tinnhlt-gra:
Confirmed fixed on core `225119443`.

`6402000a1` ("fix(reports): an unowned row is a caller's population too", #4869) names `open-deals-per-company (#4209)` by number as fixed by its unowned-row carve-out to `AnalyticsPopulationClause`, without changing the report's own/team scope. `open-deals-per-company`'s own spec still declares no `population:` override, so it inherits the fixed default.

[...]

PR #4869, opened 2026-09-07
fix(reports): an unowned row is a caller's population too

tinnhlt-gra:
## Summary

Fixes the unowned-row defect from `.tmp/qc-issues/PLAN.md` Wave A across all
nine report-population issues (margince#4205, #4206, #4207, #4209, #4210,
#4224, #4225, #4226, #4227). **Scope note up front**: this PR fully closes
four of the nine; the other five need a follow-up before their own-lens
narrowing can be safely widened — see "What this PR does NOT do" below. That
follow-up decision is not this PR's to make alone.

## The confirmed, safe fix (all nine reports get this)

`AnalyticsPopulationClause` (`backend/internal/compose/analyticsscope.go`)
renders `owner_id = $me` / `owner_id IN (team subquery)` for a caller's
default population. SQL null semantics mean that never matches a row with
`owner_id IS NULL` — an unrouted lead, an unassigned deal — so it silently
dropped out of every report for every non-admin caller, even though the same
row is fully visible on an ordinary unscoped list. This is margince#4205's
own literal reproduction.

`platform/auth/rowscope.go` already treats an unowned row as shared for
ordinary row-scope reads (`unownedIsShared`). `AnalyticsPopulationClause` now
carries the same rule for a caller's own default population (self, or a team
manager's own "managed teams" default) — **never** when a caller explicitly
names a different colleague or team, since that path is a standing forecast
or named-lookup asking what a specific person committed to, and an unclaimed
row is nobody's commitment. Confirmed to resolve deals-by-stage (#4206),
leads-by-status (#4205), pipeline-current (#4207) and open-deals-per-company
(#4209) with no change to their population scope.

## What this PR does NOT do, and why

The plan proposed declaring `population: measureEveryReadableRow` on five
more reports (stage-age #4210, win-loss #4224, projects-by-phase #4225,
project-commitments #4226, projects-gone-quiet #4227) to make them
install-wide. **I implemented that, then reverted it after an adversarial
security review caught a real regression**: `deal` and `project` are identity
tables — row scope renders unconditionally TRUE for every seat — so on these
five reports the population clause was the *only* thing narrowing anything.
Declaring `measureEveryReadableRow` removes it, and three of the five expose
`owner_id` as both a dimension and a filter with two carrying money
aggregates. The result: a rep filtering win-loss to `owner_id: <colleague>`
would get that colleague's exact closed-deal revenue — the same question the
report's own declared scope check already refuses.

These five stay on the caller's own/team default for now (still getting the
unowned-row fix above, which was the part of each report's original QC
finding actually confirmed live). Closing #4210/#4224/#4225/#4226/#4227 the
way the plan intended needs an owner-filter authority gate — checking that a
caller may "measure" a specific named owner even when the report's overall
population is unnarrowed — which is its own design decision and its own
diff, not a one-line addition to this one.

## Also in this diff

- `AnalyticsPopulationClause`'s unowned-row admission is gated by an explicit
  `unownedPopulation` parameter now, not applied unconditionally: a second
  review round found the shared function also feeds Forecast and its
  share/conversion seams, where widening a manager's own default would make
  their team total stop reconciling with the sum of naming each member by
  id — a money-integrity defect on a standing commitment number. Forecast and
  its seams now pass `unownedIsExcluded`, unchanged from before this PR;
  only the report engine passes `unownedIsPartOfDefault`.
- A named-team negative test (the `ScopeKindTeam` twin of the named-colleague
  one already added), both mutation-tested by temporarily bypassing their
  guard and confirming the test catches it.

## Test plan

- [x] `make check-go` / `make check-all` — clean (0 lint issues, 0 craft
      findings); the one local `backend/gates` failure
      (`TestPublicTreeCitesNoDocumentGitIgnores`) is a pre-existing,
      machine-local artifact from this machine's `.git/info/exclude` — CI's
      fresh clone doesn't have it, confirmed by comparing against `origin/main`'s
      own green CI run independent of this branch.
- [x] New/changed lines: 100% covered, verified with `go test -coverprofile`
- [x] Regression test per issue, each confirmed to fail without its half of
      the fix; the security-relevant ones (named-colleague, named-team,
      colleague-filter-answers-empty) mutation-tested against the guard they
      pin
- [x] UAT: seeded a real unowned, disqualified lead on a live dev stack and
      confirmed via `rep@demo.test`'s own Leads board that the terminal
      column now reads "Disqualified 1" instead of 0 — video attached below
- [x] No `crm.yaml` contract change; no AI prompt touched, so no aicert
      re-run needed
- [x] Fable + Codex craftsmanship review, then a separate craft-reviewer +
      security-redteam round — every confirmed finding from both rounds
      applied (see commit messages)

### DIFF
--- a/backend/internal/compose/analyticscontext.go
+++ b/backend/internal/compose/analyticscontext.go
@@ -86,7 +86,8 @@ func analyticsContextFor(ctx context.Context, tx pgx.Tx, now time.Time) (crmcont
 		return crmcontracts.AnalyticsContext{}, errors.New("compose: no actor bound to context")
 	}
 
-	def, defClause, err := AnalyticsPopulationClause(ctx, tx, RequestedScope{}, "", func(any) int { return 0 })
+	def, defClause, err := AnalyticsPopulationClause(
+		ctx, tx, RequestedScope{}, "", func(any) int { return 0 }, unownedIsPartOfDefault)
 	if err != nil {
 		return crmcontracts.AnalyticsContext{}, err
 	}
--- a/backend/internal/compose/analyticsmcpquestions_manager_integration_test.go
+++ b/backend/internal/compose/analyticsmcpquestions_manager_integration_test.go
@@ -19,12 +19,10 @@ import (
 	"github.com/margince/margince/backend/internal/shared/kernel/ids"
 )
 
-// seedClosedOwnedDeal writes one won or lost deal OWNED by somebody.
-//
-// seedPricedDeal leaves owner_id NULL, which is right for the currency cases it
-// serves and wrong here: every report in this file measures the caller's own
-// population by default, so an unowned deal is invisible to the manager asking
-// about their team and the fixture would prove nothing.
+// seedClosedOwnedDeal writes one won or lost deal OWNED by somebody, with a
+// chosen `source` — seedPricedDeal takes neither, and every case here is
+// management reading a specific rep's real, attributed contribution, not an
+// unclaimed one.
 func seedClosedOwnedDeal(
 	t *testing.T, e *forecastEnv, name string, amountMinor int64, status, source string, owner ids.UUID,
 ) {
--- a/backend/internal/compose/analyticsquery_integration_test.go
+++ b/backend/internal/compose/analyticsquery_integration_test.go
@@ -493,6 +493,40 @@ func TestATypedQueryRefusesAPopulationTheAskerCannotMeasure(t *testing.T) {
 	}
 }
 
+// Naming a colleague explicitly is a different ask than measuring your own
+// default population, and an unowned row must not count toward it: this is a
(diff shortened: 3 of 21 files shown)


tinnhlt-gra:
UAT proof: seeded a real unowned, disqualified lead on a live dev stack and confirmed as `rep@demo.test` (RowScopeOwn, owns nothing) that the Leads board's terminal column now reads "Disqualified 1" instead of 0 — the exact margince#4205 reproduction, fixed. Also confirmed directly against the running API: `POST /v1/reports/leads-by-status` as the same rep returns `{"status":"disqualified","leads":1}`.

Rep's Leads board showing the unowned disqualified lead now counted

tinnhlt-gra:
Follow-up filed for the five reports whose full "install-wide" fix needs an owner-filter authority gate: margince#4871

PR #4911, opened 2026-09-08
fix: /brief gets the model-call deadline, and its worklist row gets the right label

tinnhlt-gra:
## Summary

Two of Wave B's four planned issues from `.tmp/qc-issues/PLAN.md` — re-verified
against current `main` before implementing, since the plan was written before
this bump and two of the four turned out already fixed:

- **margince#4355** — `POST /v1/brief` was missing from `modelRouteSuffixes`
  (`backend/internal/compose/modelroutedeadline.go`), so it ran under the
  ordinary 30s `WriteTimeout` instead of the long model-call deadline. Same
  shape margince#3633 fixed for `/meeting-brief`: the daily brief calls the
  model task `brief_ranking` and, measured live, 4 of 5 calls exceeded 30s and
  were answered `502` to the caller though the handler itself finished and
  logged `201` every time — the request succeeded, the client had already been
  disconnected.
- **margince#4859** — the Worklist's `open_meeting_brief` move rendered as
  "Open to reply": `moveLabel` had no branch for it, so it fell through to the
  `draft_reply` default on a row that is plainly a meeting, not a message. The
  link's destination was always correct; only its label lied. Added the
  branch plus a new `worklist.verb.open_meeting_brief` i18n key (en/de/vi),
  and the regression test the issue's own body points out was missing:
  nothing previously called `moveLabel` with an `open_meeting_brief` row.

## Not in this PR

- **margince#4201** (priced_deals reaches no screen) — already fixed by
  PR #4338 (`9c4b93a01`, merged 2026-09-05). Closing the GitHub issue
  separately with a comment; #4338's body never said "Closes #4201" so
  auto-close never fired.
- **margince#4304** (no_champion never populated) — already fixed, issue
  already closed on GitHub. No action.

## Test plan

- [x] `make check-go` — clean (0 lint, 0 craft findings); the one local
      `backend/gates` failure (`TestPublicTreeCitesNoDocumentGitIgnores`) is
      the same pre-existing, machine-local `.git/info/exclude` artifact
      documented in Wave A (PR #4869) — unrelated to this diff, confirmed by
      `git diff origin/main --stat` touching none of the files it cites.
- [x] `make check-fe` — clean (typecheck, lint, unit, build, ds-gates, i18n
      key-parity all pass)
- [x] Regression test per issue, each confirmed red before its fix:
      `TestOnlyTheModelRoutesTakeTheLongerDeadline` (backend) and
      `moveLabel`'s new `open_meeting_brief` case (frontend)
[...]

Issue #4350, opened 2026-09-05
A recorded consent-stop reads on History as raw field names, not plain English


tinnhlt-gra:
# A recorded "stop" reads on History as raw field names, not a sentence a compliance reviewer could act on

**Upstream:** [margince#TBD](https://github.com/margince/margince/issues/TBD)

## Observed

Measured live on core `0e372820e`, stack `margince_dev_t2consent`, building
`consent-records-a-stop` (UAT-STOP-1) on 2026-09-05.

After `POST /people/{id}/consent/suppress {"kind":"subject_request","reason":"..."}`
→ 204 (core `ccb37d851`, #4331), `GET /records/person/{id}/history` returns:

```json
{
  "action": "update",
  "actor_name": "UAT Member",
  "after": {"decided_by_level": "user", "kind": "subject_request"},
  "summary": "UAT Member updated the record",
  "undoable": {"reason": "no_before_image", "undoable": false}
}
```

The History tab renders this exactly: the sentence "UAT Member updated the
record" (indistinguishable from any other field edit on the person), followed
by a field-change list with two rows — field name **"kind"**, value
`subject_request`; field name **"decided by level"**, value `user`. Neither
label is translated: `historyFieldLabel` (`historyfieldlabels.ts`) falls back
to `field.replaceAll("_", " ")` for any key its map does not carry, and `kind`
/ `decided_by_level` are not in that map.

## Why — these are synthetic payload keys, not real `person` columns, so nothing forces them a label

`suppress.go` writes its audit row through `storekit.AuditEvent(ctx, tx,
"update", "person", id, map[string]any{"kind": in.Kind, "decided_by_level":
string(level)})` — a free-form payload, not a before/after image of an actual
`person` table column. `historyfieldlabels.ts`'s own comment says its map is
"derived from the contract" by `historyfieldlabels.test.ts`, so a real
updatable `person` field missing a label fails that census — but `kind` and
`decided_by_level` were never `person` schema fields to begin with, so the
census that is supposed to catch exactly this class of gap never sees them.
The same pattern already exists for two other `AuditEvent` writers in this
module (`authorizebasis.go`, `qualifyingevent.go`; `confirmsubmit.go` uses the
before/after-carrying `Audit` instead), so this is not unique to suppression,
but the suppression write is the first of the three anyone here has walked on
screen.

## Cost

A compliance reviewer opening a person's History to answer "did somebody
record that this person asked us to stop, and when" sees "UAT Member updated
[...]

Issue #4349, opened 2026-09-05
The person page's Consent panel still reads "Allowed" after a subject asks to stop, while the send gate already refuses


tinnhlt-gra:
# The person page's Consent panel still reads "Allowed" after a subject asks to stop, while the send gate already refuses

**Upstream:** [margince#TBD](https://github.com/margince/margince/issues/TBD)

## Observed

Measured live on core `0e372820e`, stack `margince_dev_t2consent`, building
`consent-records-a-stop` (UAT-STOP-1) on 2026-09-05.

A person (`rep`-owned, one 60-day-old inbound email so `business_correspondence`
legitimately reads `allowed`) is recorded as having asked to stop, through the
door core `ccb37d851` (#4331) just added:

```
POST /v1/people/{id}/consent/suppress
{"kind":"subject_request","reason":"asked us on a call to stop emailing them"}
→ 204
```

Immediately afterward, the two surfaces that are supposed to answer the same
question about this person disagree:

**`POST /v1/emails:preview`** (the real send-time gate; `authorizetransmit.go`'s
`liveSuppression` reads `communication_suppression` directly) correctly refuses:

```json
{"allowed":false,"recipients":[{"address":"zz.uat.stop.test@example.test",
"verdict":"deny","reason_code":"processing_restricted","would_refuse":true,
"decided_by":"subject","can_be_overruled":false,"resolved_category":"marketing",
"mode":"enforce"}]}
```

**`GET /v1/people/{id}/consent/guard`** — the endpoint the person page's own
"Consent & channels" panel reads — is unchanged from before the suppression was
recorded:

```json
{"channel":"email","purpose_class":"business_correspondence","verdict":"allowed",
"reason":"they wrote to you on 7 Jul",
"qualifying_event":{"kind":"inbound_message", ...}}
```

The contact's own record page (`frontend/src/screens/personrail.tsx`'s
`ConsentAndChannels`) renders exactly that stale answer: **Email: Allowed**, in
the "good" (green) style, with the reason "they wrote to you on 7 Jul" printed
underneath — the one screen a rep would actually look at before deciding
whether to write to this person.

## Why — `verdict.go` never reads `communication_suppression`

`backend/internal/modules/consent/verdict.go` (the code behind
`PersonConsentGuard`) has no reference anywhere to `communication_suppression`
or to `liveSuppression`. It computes `business_correspondence` purely from
qualifying events (inbound messages, deals, in-person notes) and never
[...]

PR #4926, opened 2026-09-08
fix: a consent-suppression reads Blocked everywhere, and its History entry names itself

tinnhlt-gra:
## Summary

Wave C of `.tmp/qc-issues/PLAN.md`'s QC cleanup — the consent-suppression
follow-up, margince#4349 and margince#4350, both gaps in the same feature
(`ccb37d851`, #4331).

### margince#4349 — the guard/panel disagreed with the send gate

`VerdictForPerson` (`backend/internal/modules/consent/verdict.go`) — THE
decision both the consent guard endpoint and the transmit gate are supposed
to share — never read `communication_suppression`. The transmit gate
(`authorizetransmit.go`) checks it as a layer bolted on AFTER calling this
same function, so it was already correct; every caller that stops at
`VerdictForPerson` alone (the guard endpoint, the person page's "Consent &
channels" panel) never saw it. A rep recording "they asked us to stop" (the
door `ccb37d851` added) saw the person's own page still say "Email: Allowed"
a moment later.

Fixed by reusing the exact rule the transmit path already uses —
`categoryForClass` + `suppressionBinds` — rather than inventing a second one:
same category mapping, same binding table, called from a second place. No
address (a guard read answers about the PERSON in general, not one message to
one address — a stale mailbox's hard bounce must not read as "this person is
blocked").

### margince#4350 — the History tab printed raw payload keys

The suppression write's own `AuditEvent` payload (`kind`,
`decided_by_level`) has no updatable contract column, so
`historyfieldlabels.ts`'s contract-derived census (both directions, by
design) can never see the gap — exactly why it rendered raw ("kind",
"decided by level") on the person page's History tab.

Added a second, deliberately un-censused map (`SYNTHETIC_AUDIT_FIELD_LABELS`)
for AuditEvent payload keys with no contract column, consulted as a fallback.
Widened past the issue's literal two keys to cover the two other
consent-module writers with the identical gap — `qualifyingevent.go`
(`qualifying_event`, `note`) and `authorizebasis.go` (`communication_basis`,
`resolved_category`) — since it's the same file, same census gap, same
session, and the marginal diff is six map entries rather than two.

### A second, pre-existing bug UAT found in the same area

Live-checking #4350 found the "Changes" tab rendered correctly but the
**default "All" tab** — the one a reader actually opens first — still showed
raw text. Root cause: `recordchronology.tsx`'s combined timeline labels its
field-history rows through `coldFieldLabel` (the enrichment-vocabulary map,
`screens/common.tsx`), not `historyFieldLabel` — despite reading the
identical `FieldHistoryEntry[]` feed the Changes tab correctly labels. An
existing test (`recordchronology.test.tsx`) had pinned the wrong raw-fallback
text (`"owner id"`) as its expected value, which is how this survived
unnoticed until now. Fixed the call site and the stale assertion; both
directions mutation-tested.

## Test plan

- [x] `make check-go` — clean (0 lint, 0 craft findings); the one local
      `backend/gates` failure (`TestPublicTreeCitesNoDocumentGitIgnores`) is
      the same pre-existing, machine-local `.git/info/exclude` artifact
      documented on prior PRs (#4869, #4911) — unrelated to this diff
- [x] `make check-fe` — clean (typecheck, lint, unit, build, ds-gates, i18n
      key-parity, retired-noun census all pass)
- [x] New integration tests for #4349, each confirmed red before the fix:
      subject_request blocks (the literal repro), a marketing objection does
      NOT over-block business correspondence (suppressionBinds's own
      category rule), a processing restriction blocks, a hard bounce blocks,
      a revoked suppression does not
- [x] New/updated tests for #4350: the synthetic map stays out of the
      contract census in both directions, resolves every key to a real i18n
      entry, pins the literal repro's two keys; `recordchronology.test.tsx`
      corrected and mutation-tested for the second bug
- [x] UAT: recorded a live suppression on a seeded dev stack (Alice Müller),
      confirmed the person page now reads "No purpose currently permits
      writing to them" instead of Allowed, and both the "All" and "Changes"
      History tabs read "Decided by" / "Suppression kind" instead of raw
      field names — video attached below
- [x] No `crm.yaml` contract change, no AI prompt touched → no aicert re-run
      needed

https://claude.ai/code/session_01BvvZzihBBSeaLc3WcZ6Xh1

---
## Summary by cubic
Fixes the consent guard and person-page panel so a recorded `communication_suppression` reads as Blocked instead of Allowed, and makes History rows show readable labels instead of raw payload keys.

- `VerdictForPerson` now reads the same suppression table and category rule the transmit gate uses, so the guard and send path can't disagree; a marketing objection blocks only marketing, subject requests and processing restrictions block business correspondence, address-pinned hard bounces don't block the person-level verdict, and revoked suppressions don't block.
- Renamed the suppression write's audit key from `kind` to `suppression_kind` so History can't mislabel every activity's own `kind` field.
- A fallback label map resolves audit payload keys with no contract column, covering all five consent-module writers: `suppress.go`, `qualifyingevent.go`, `authorizebasis.go`, `lift.go`, and `confirmsubmit.go`.
- `blockedReasonCode` now returns the verdict's own suppression kind instead of falling to a default meant for a marketing objection.
- The History tab's "All" view now labels field rows with `historyFieldLabel` instead of the enrichment vocabulary; a stale test assertion pinned the wrong raw text and hid the bug.

<sup>Written for commit 9ce0b81a058a6d0ed61b6e2d40d53d4b43037e6a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/4926?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Consent verdicts now correctly account for active communication suppressions, including subject requests and processing restrictions.
  * Revoked suppressions and address-specific hard bounces no longer incorrectly block person-level communication.
  * Suppressed communications now record the applicable suppression reason instead of an objection.

* **Improvements**
  * History and chronology views now show clearer localized labels for suppression details and record fields.
  * Added English, German, and Vietnamese translations for new history, onboarding, and stage-automation messages.


tinnhlt-gra:
UAT: seeded a live suppression against Alice Müller on a live dev stack (`POST /people/{id}/consent/suppress {"kind":"subject_request"}`). Confirms all three surfaces now agree: the person page's own consent line reads "No purpose currently permits writing to them" (was "Allowed"), and both the default "All" History tab and the "Changes" tab read "Decided by" / "Suppression kind" instead of raw payload keys.

wave-c-consent-suppression-uat

tinnhlt-gra:
Internal Fable + Codex craftsmanship review found one real BLOCKER worth flagging clearly: the synthetic map's `kind` key collided with an unrelated writer's field of the same name — `activities/activity.go` already audits every activity's own `create` with its own `kind` field (email/call/note/meeting), and `historyFieldLabel` carries no entity context, so every activity in a person's History would have read "Suppression kind: email" etc. Renamed the suppression write's audit payload key to `suppression_kind`. Also: widened the map's own count (it claimed 3 consent-module writers, the package actually holds 5 — added `lift.go`/`confirmsubmit.go`'s keys); added `Verdict.Suppression` so `blockedReasonCode` names the real block reason instead of falling to an unrelated default (safe today only by call-order coincidence); added a fitness-function test holding `categoryForClass`/`resolutionForClass` in sync since `VerdictForPerson`'s new check now depends on them agreeing; replaced an unrealistic hard-bounce test with the honest case (a bounce is address-scoped, never person-scoped); consolidated 3 near-duplicate test seed helpers into one. Filed #4928 for the still-missing gate that would hold the synthetic map to its writers automatically — a static-analysis undertaking of its own.

tinnhlt-gra:
UAT after the review-round fix: confirmed live that renaming the suppression's audit payload key to `suppression_kind` didn't just fix the collision in theory — recording a fresh suppression now correctly renders "Suppression kind" on the History tab, and the same contact's own unrelated "Meeting" activity entry beside it is unaffected, confirming activities keep their own `kind` untouched.

wave-c-suppression-kind-fix-uat

PR #4929, opened 2026-09-08
fix(tests): two fixtures that a change beneath them made stale

dungbanguyen:
**main is red on the integration lane**, deterministically, and it is blocking every open PR. Two tests, neither about what broke it. I checked for an existing fix first: #4852 is open with only a bot comment, #4917 records the merge that landed against the failing verdict, and no open PR touches either file — so this is that fix, and this note is the "say so" the rulebook asks for.

Both are **fixture repairs**. No engine moves.

## 1. A cadence written down twice

`pass_integration_test.go` spelled `time.Hour` in three assertions about `capture_counterparty_verdict`. #4904 changed that kind's declared cadence to `10m` in `api/jobs.yaml` and regenerated `specs_gen.go` — correctly — and the test now fails about a number no reader of that file could see was stale:

```
cadence = 10m0s, want the hour api/jobs.yaml declares
next = …03:28:56, want the last TICK plus the cadence …04:18:56
```

The property under test is that `PassFor` answers the **declared** cadence and projects from the last tick by it. Neither half is about which number the declaration happens to hold, so neither should break when it changes. It reads the cadence from `jobs.SpecFor` now.

It still fails loudly if the kind stops declaring one — a zero cadence would make the projection assertion pass against any answer, which is the one way this could go quiet instead of red.

## 2. A last-inbound date that stopped meaning reachability

`person_relationship_room_integration_test.go` seeded an inbound email linked to a contact and expected `last_inbound_at`. `lastTouchSection` now composes `people.SenderPredicate`, and its own comment says why:

> "THEY wrote" is a claim about authorship, and a thread is linked to everybody it concerns — so reading it off reachability told a reader "they wrote last" about a message somebody else sent into a conversation this person is on.

Under that rule an inbound activity with no `from` participant is a message that *reached* them, which is the outbound question. The fixture now says who sent it, so the case models what its own assertion claims — "a contact who wrote to us" — and exercises the authorship rule rather than passing on the link.

## Checks

`make check-backend` green; `make test-it DIR=backend/internal/platform/jobs` green; the `compose/integration` lane green.

### DIFF
--- a/backend/internal/compose/integration/org360/org360_lasttouch_integration_test.go
+++ b/backend/internal/compose/integration/org360/org360_lasttouch_integration_test.go
@@ -95,9 +95,12 @@ func TestStrengthFoldLeavesTheInboundAnchorEmptyWhenNobodyWroteIn(t *testing.T) 
 // The two are answering different questions. "They last wrote eighteen months
 // ago" is history and stays true however old it gets. The anchor is an action —
 // it is what a Follow up button opens a reply against — and it has to agree
-// with the state the counts report. This contact reads untried, because nothing
-// inside the window says otherwise; offering to answer a thread from last year
-// would contradict the same page's own summary.
+// with the state the counts report, so offering to answer a thread from last
+// year would contradict the same page's own summary.
+//
+// The state is `lapsed`, not `untried`: the exchange is over rather than
+// absent, and the row carries the date that says so. Untried beside a
+// last-heard-from date is the contradiction `lapsed` exists to end.
 func TestStrengthFoldDropsAnInboundAnchorOlderThanTheWindow(t *testing.T) {
 	e := integration.Setup(t)
 	owner := integration.OwnerConn(t)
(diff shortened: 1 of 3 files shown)


dungbanguyen:
Two more from the same pair of upstream changes, which the other CI shards carry and my local lane did not reach on the first pass:

- `org360_lasttouch_integration_test.go` asserted `untried` for a contact whose only message predates the 90-day window. That state is `lapsed` now, and the reason is written beside the constant — untried printed beside the row's own last-heard-from date was a plain contradiction, and `lapsed` exists to end it. The file's own subject (a date without an action anchor) is untouched.
- `TestADismissalHoldsUntilTheEvidenceMoves` seeded the reply that must re-arm a dismissal as a linked inbound activity with no sender — a message that *reached* them under the authorship rule, so the evidence never moved.

Both packages green locally: `compose/integration` and `compose/integration/org360`.

dungbanguyen:
Closing this in favour of #4943, which lands the same two lanes with a better diagnosis on the person-page half.

I read this as a stale fixture and repaired the fixture. #4943 read it as a production defect and it is right: `SenderPredicate` demands a `from` participant row, and **only capture writes participant rows** — so a message a rep logs by hand carries none at all, and every hand-logged inbound message stopped counting as "they wrote". On an installation with no mailbox connected that is every message there is. Their `AuthoredPredicate` (the sender proof **or** no participant row whatsoever) restores the hand-logged case while leaving Marcus-wrote-into-Judith's-thread refused, which is the case that started all this.

My fixture change would have made the test green and left that defect in place. It also pushed `person_relationship_room_integration_test.go` to 1006 lines and tripped the craftsmanship gate, which #4943 avoids by putting the last-touch fold in its own file.

The cadence half of both PRs converged on the same shape independently — read the interval from `jobs.SpecFor` rather than restating it — so nothing is lost there either.

One breakage neither PR covers: `TestDOIIssuanceMintsNothingWhicheverPurposeIsNamed` and the `IssueDoubleOptInJSONBody.purpose_id` case in `requiredbodyids_http_integration_test.go` still expect the 409 that #4931 replaced with a working issuance. I am taking that one separately.

Issue #2108, opened 2026-08-21
A project attribution that fails is never retried, and no event announces one that succeeds


LarsGradion:
Two related gaps in the post-commit project-attribution ladder added in PR #2106 (`backend/internal/modules/capture/sinkproject.go`).

## No reconciliation for a failed attribution

`attributeProject` is called only when the capture transaction just created the activity. A transient fault (SQL error, matcher failure) logs for reconcile and returns — but every later mailbox replay finds the activity already present, so the ladder never runs again and the message stays unfiled forever. The comment promises a nightly reconcile that has no caller.

Fix: either a reconcile pass that finds connector-captured activities with no project link and a `thread_key`/deal/subject that would match one, or run the ladder on replay when the activity carries no project link yet. The second is cheaper and idempotent — `uq_activity_link_project` plus `ON CONFLICT DO NOTHING` already make a second run safe.

## No event when the link lands

The link is a domain mutation that commits with an audit row and no outbox event, ratified in `writeshape_test.go`'s `auditOnlyWrites` on the ground that `activity.captured` already announced the arrival. That reasoning is incomplete: `activity.captured` commits *before* the link exists, so a subscriber that reads the activity on receiving it caches an unlinked row and nothing ever tells it to refetch.

Fix: emit on the attribution transaction — or, if the closed event catalogue makes a new type the wrong answer, move the whole ladder inside the capture transaction so `activity.captured` genuinely covers it. That second option trades the "never fails the capture" property away, so it is a real decision rather than a cleanup.

Both found by the Codex review pass on PR #2106.

dungbanguyen:
## Decision, 2026-09-02: run the ladder on replay; the event stays audit-only for now

Re-triaged during the `status: needs-decision` sweep.

### The retry half — run the ladder on replay

**Take the ticket's own preferred fix, and for the reason it gives: the safety already
exists.** `uq_activity_link_project` plus `ON CONFLICT DO NOTHING` make a second run
safe, so running the ladder whenever a replayed activity carries no project link is
cheap and idempotent with nothing new to build.

It also fixes the actual failure rather than covering it. The defect is that a
transient fault means the message *"stays unfiled forever"* — because
`attributeProject` runs only when the capture transaction just created the activity,
and every later replay finds it already present. Running on replay makes the ordinary
retry path the reconciliation.

**A nightly reconcile pass** was the alternative, and it is the weaker one twice over:
it is a new scheduled job, and its own failures would be exactly as silent as the fault
[...]

PR #4937, opened 2026-09-08
fix(capture): a failed project attribution is retried by the next capture

dungbanguyen:
Addresses #2108 — the retry half. **The event half stays open**, narrowed; see below.

The ladder ran only when the capture had just **created** the activity. A transient fault — a SQL error, a matcher failure — logged for a reconcile that has no caller, and every later replay of that mailbox found the activity already present and skipped the ladder. The message stayed unfiled **forever**, and the breadcrumb was the whole record that the question went unanswered.

## The retry

The issue offers two fixes and calls this one cheaper and idempotent. It is:

- `linkActivityToProject` is already `ON CONFLICT DO NOTHING` and stamps whether or not it inserted, so a second run was never the risk.
- What it was not is **free**. `alreadyFiledUnderAProject` makes it so: an activity already filed stops on one indexed read, before the three rungs.

**Only a project link counts as an answer.** An activity filed under a person or a deal has not been asked this question yet, and treating any link as an answer would make the retry unreachable for exactly the messages that carry other links — which is most of them.

## Verified both ways

| mutation | result |
|---|---|
| restore the created-only call | the replayed message is *filed under 0 projects, want 1* |
| remove the guard | the matcher is *asked 3 times after a third capture, want 2* |

The fake matcher fails once and succeeds after, which is the transient fault this exists for; the test drives the real sink twice with the same record, which is what a mailbox replay is.

`make check-backend` exit 0.

## The event half, narrowed

> emit on the attribution transaction — or, if the closed event catalogue makes a new type the wrong answer, move the whole ladder inside the capture transaction

The first option is **ruled out**, not merely doubtful: the event catalogue is closed and a new type cannot be minted build-side. So the decision is between moving the ladder inside the capture transaction — which trades away the "never fails the capture" property the whole design rests on — and accepting that a subscriber caching the activity on `activity.captured` never learns to refetch.

That is a real product decision and not mine to make. Left on the issue with `status: needs-decision` suggested.

## Summary by CodeRabbit

* **Bug Fixes**
  * Project attribution now retries on subsequent captures when an earlier attempt fails.
  * Activities that were initially captured without a project can be linked automatically during a replay.
  * Existing project links are recognized quickly, preventing unnecessary repeated attribution work.


dungbanguyen:
**All three confirmed and fixed** — `0c457e262`. Both of you found the same P1 from different angles, and it was real.

**The guard skipped the stamp, not just the ladder.** `linkActivityToProject`'s conflict path exists to repair a project link that was never stamped — its own comment says the "whoever filed the link also stamped it" reasoning is exactly the guarantee nothing enforces, which is why the stamp migration carries a backfill. My early return reintroduced that skip one layer up. The guard now answers *"is there anything left to decide"*, and the already-filed case still goes through the write; what is skipped is the three rungs.

Mutation-checked: restoring the bare early return fails the new test with *"the replay stamped 0 time(s), want 1"*.

**The stamp named the wrong project.** @cubic-dev-ai was right about the race — the stamp used the project this call *proposed*, so a human relink landing between the ladder's read and the insert wins the row, `ON CONFLICT` discards our choice, and the stamp writes retention evidence for a project that does not own the activity. It re-reads the row and stamps what is actually there now, bounded by the same `EnsureLinkTarget` the insert arm applies.

That bound is not decoration: reading `project_id` into a value the function never bounded pushed `TestTheModulesTierIsMeasuredForTheWideningItIsOwed` from 96 to 97. That ratchet has risen exactly once in the tree's history, with a much stronger argument than mine would have been — so the read moved inside the function that bounds the project rather than the ceiling moving. The census is back at 96.

**The lead-capture guard is in** (`ref.Type != datasource.EntityActivity`). You are both right that `decideProject` refused it only after this function had already opened a transaction.

`sinkproject.go` crossed the 500-line ceiling with the above, so the write half moved to `sinkprojectwrite.go`. Three gate waivers are keyed by file *and* function and had to move with it — the gates named each one, including the two stale keys left behind.

dungbanguyen:
Both test criticisms are right, and both tests are changed. `6dfce2e87` and `14bdeb886`.

**The repair test claimed a case it did not build.** You are correct that the first capture files the link *and* stamps it — the insert arm always stamps — so no genuinely unstamped link ever existed in the fixture. Renamed to `TestACaptureOfAnAlreadyFiledActivityStillStamps`, which is what it proves, and the comment now states the fixture's limit rather than papering over it: the stamp is idempotent, and what a pre-stamp link needs is exactly this — that the call is made at all.

**The divergence test proved nothing, and is gone.** Also correct: with the guard reading the filing first, an already-filed activity proposes a zero project, so proposed and filed never differ and the branch was never entered. The divergence is reachable only through the race it exists for, and reaching it would have needed either a test-only export (an idiom this tree does not use) or a second copy of the capture DB harness in an internal test package.

So the invariant is stated where the read is instead. Worth noting the uniqueness-claim gate rejected two drafts of that comment — "the two cannot disagree" and "the one function" are claims no test holds — which is the right outcome: it now says what the code does rather than asserting an impossibility.

The first three threads here were addressed in  (the guard reaching the stamp, the re-read naming the filed project, and the  guard on lead captures).

dungbanguyen:
Correction to the last line of my previous comment — shell mangled it.

The first three threads on this PR were addressed earlier, in commit `0c457e262`: the guard reaching the stamp instead of skipping it, the re-read naming the project actually filed rather than the one proposed, and the `datasource.EntityActivity` guard so a lead capture no longer opens an attribution transaction.

Issue #4935, opened 2026-09-08
AC-W2 has no deterministic gate once its budget leaves the merge lane


dungbanguyen:
AC-W2's trigger→dispatch budget moves to the by-hand `bench` lane (`make bench-dispatch`), because a wall-clock p95 read on a runner where six integration shards share one Postgres measures the runner: a `main-health` run recorded p50=13.19ms — sixteen times inside the 200ms budget — against p95=201.63ms and filed `main` as broken over the 1.63ms.

Every other budget left the merge gate before it. PERF-3/PERF-7 could go because a deterministic test already held its regression — the `seq_scan` count in `lastactivity_integration_test.go`. **AC-W2 has no such sibling**, so the standing lane now holds nothing about dispatch cost, and that gap is this issue rather than a green test with nothing behind it.

## What was tried, and why each cannot fail

Three instruments were written and measured against a real mutation (the run-outcome write with its second index column removed, so a firing reads the whole history for its handler):

1. **`seq_scan` delta on `workflow_run`.** Never trips. `workflow_run_unique` leads on `handler`, so a firing that has lost the second column still reaches its row by an INDEX scan over every run that handler ever made — 5,000 rows read per dispatch and not one sequential scan to show for it.
2. **The same, with history seeded under the firing handler** so the second column is load-bearing. Still an index scan; still no sequential scan.
3. **A rows-read ceiling** (`seq_tup_read + idx_tup_fetch`). The right quantity, and it still cannot fail — see below.

## The design question

The engine dispatches over the **app pool**, and `pg_stat_force_next_flush()` flushes only the calling backend. A reader on the owner connection therefore sees counters the dispatching backends have not reported yet, so any `pg_stat_user_tables` instrument reads a number that is not about the batch just run. `lastactivity_integration_test.go` gets away with it because the statement under measurement runs on the owner connection itself.

So a deterministic AC-W2 gate needs one of:

- a way to attribute work to statements the app pool ran, without a clock-based wait for stats to settle (a condition-poll on the counters is still time-dependent, and the lane's rule is that nothing in it reads a clock);
- a **plan-shape** gate instead of a work-done one — `EXPLAIN` over the engine's run-claim and run-outcome statements, asserting an index path. That needs the statements reachable from a test, which is a module-boundary question;
- or an honest statement that AC-W2 is measured by hand only, and the lane covers dispatch correctness and not dispatch cost.

The third is the status quo this issue exists to either ratify or replace, which is why it carries `status: needs-decision`.

[...]

PR #4940, opened 2026-09-08
AC-W2's dispatch budget leaves the merge gate

dungbanguyen:
One of the three causes collected under https://github.com/margince/margince/issues/4852 — see the triage comment there. That issue stays open until the lane is green, because the other two are not this.

AC-W2's trigger→dispatch budget leaves the merge gate for the by-hand `bench`
lane, where every other budget in this tree already lives.

## Why

`main-health` filed the integration lane as broken twice. The two recurrences
are unrelated, which is what a title-deduped tracker issue does:

- run 34118992386 — `docker compose pull` answered `502 Bad Gateway` during
  `make db-up`. Nothing to fix here.
- run 34122420926 — `TestWorkflowTriggerToDispatchP95HoldsOnTheSeededDataset`
  read **p50=13.19 ms** against a 200 ms budget and **p95=201.63 ms**, and
  failed `main` on the 1.63 ms.

A median sixteen times inside the budget with a tail over it is a measurement of
the runner, not of the dispatch: six integration shards share one Postgres
there. Raising the sample count from 25 to 200 had already been tried and bought
one run — at 200 the p95 is the eleventh-largest sample, and ten stalls out of
two hundred is an ordinary afternoon on that runner.

This was the last clock left in the lane. PERF-3/PERF-7, PERF-1/PERF-4 and
CAP-PARAM-1 are behind the `bench` tag; MOBILE-AC-2 is behind `make
bench-mobile`. `ci.yml` states the rule on the uat lane in as many words:
*"Nothing in it reads a clock … so a slow runner cannot redden it."*

Run by hand it passes with room to spare — **p95 = 25.7 ms** against the 200 ms
budget, measured on this branch. The code was never the problem.

## What is not here

PERF-3/PERF-7 could leave because a deterministic test already held its
regression. AC-W2 has no such sibling, and rather than ship a green test with
nothing behind it I measured three candidate instruments and none of them can
fail — the engine dispatches over the app pool and `pg_stat_force_next_flush()`
reaches only the calling backend. That gap, and the design question in it, is
https://github.com/margince/margince/issues/4935. The file and the docs say so
plainly instead of implying cover that is not there.

## Changes

- `workflow_dispatch_perf_integration_test.go` → `workflow_dispatch_bench_test.go`,
  tagged `integration && bench`.
- `make bench-dispatch`, delegated from the repo root (the parity gate requires
  it) and documented beside the other `bench-*` targets.
- The docs' claim that every `bench-*` target re-renders the budgets page is
  corrected — `bench-perf-check` never did either.

## Summary by CodeRabbit

- **New Features**
  - Added `bench-capture` and `bench-dispatch` commands, available from both the repository root and backend directory.
  - Added a dedicated workflow trigger-to-dispatch performance benchmark using seeded data and a 200 ms p95 budget.

- **Documentation**
  - Updated build-target documentation to describe benchmark coverage, performance-budget reporting, and root/backend target parity.

- **Tests**
  - Moved the workflow dispatch performance check into the dedicated benchmark lane rather than the standard integration performance suite.


dungbanguyen:
@coderabbitai review

dungbanguyen:
@coderabbitai full review

dungbanguyen:
Taken — you were right and I checked it against the Makefile, which says so itself:

```
vet: composition ## Run go vet over every package, untagged (composed workspace)
	# UNTAGGED, and only untagged. The tagged lanes are vetted by `make lint`
```

So `make vet` cannot see a `bench`-tagged file at all, and naming it in that guarantee made the comment claim cover it does not have. The two golangci passes carry the tags and `lintbuildtagreach_test.go` fails if either stops — that is the whole of it, and the comment now says only that.

PR #4942, opened 2026-09-08
fix: three tests still expect what production stopped doing (main red, #4904 follow-up)

tinnhlt-gra:
## Summary

`main` is currently red: `main-health` (run 34185948837, tip `edb9949764`)
and every PR's `integration` shards fail on three unrelated tests, all stale
against `8c1cd8a39` (#4904, "The webinar defect list: nine fixes") — a real,
correct production change this morning that a few existing tests weren't
updated to match. Found while investigating why an unrelated PR's CI (#4926)
was red on packages it never touches.

### `activity_participant` authorship, now required for "last touch"

#4904 fixed Person360/org360's last-inbound fold to require
`SenderPredicate` authorship (`backend/internal/modules/people/sender.go`)
rather than mere `activity_link` reachability — a thread is linked to
everybody it concerns, so "they wrote last" was reading as whoever's page
was open. Two existing integration tests seed an inbound activity via
`LinkActivity` with no `activity_participant` "from" row naming the sender,
so the (correct, new) authorship check finds nobody and the last-inbound
date comes back empty:

- `person_relationship_room_integration_test.go` — two call sites
  (`TestPerson360AssemblesEverySectionFromRealRows`,
  `TestADismissalHoldsUntilTheEvidenceMoves`)
- `org360_lasttouch_integration_test.go` — one call site
  (`TestStrengthFoldDropsAnInboundAnchorOlderThanTheWindow`)

Added `LinkActivitySender` (`backend/internal/compose/integration/seed.go`)
alongside `LinkActivity` at each, rather than hand-rolling the raw
`activity_participant` INSERT three more times — two other tests already do
that inline.

### A genuinely new engagement state, old expectation

Same commit's last bullet: `lapsed` is an intentional new engagement state
— "an exchange that is real and older than the window" — ranking between
`untried` and `no-reply`. `TestStrengthFoldDropsAnInboundAnchorOlderThanTheWindow`
still asserted the pre-`lapsed` answer (`EngagementUntried`) for a contact
whose only message predates the window; the correct answer once authorship
is restored is `EngagementLapsed`. Updated the assertion and its doc comment.

### A cadence the test's own fixture kind no longer holds

`TestAScheduledRunIsWhenThePassRuns` (`backend/internal/platform/jobs`) uses
`capture_counterparty_verdict` as its "a kind with a fixed cadence" stand-in
and asserts `time.Hour`. The same commit moved that specific kind's cadence
to 10 minutes ("a new sender becomes a record in minutes, not most of an
hour"), and every time-arithmetic assertion in the file is written in units
of the fixture kind's cadence. Repointed the constant at `capture_classify`,
a sibling kind that still holds an hourly cadence, rather than rewriting
every assertion's arithmetic.

## Test plan

- [x] `make test-integration` — **0 skips, 51 packages, all green** (was
      failing before this fix, confirmed via `main-health` run 34185948837
      on the current main tip reproducing the identical three failures)
- [x] `make check-go` — clean (0 lint, 0 craft findings); the one local
      `backend/gates` failure (`TestPublicTreeCitesNoDocumentGitIgnores`) is
      the same pre-existing, machine-local `.git/info/exclude` artifact
      documented on prior PRs — unrelated to this diff
- [x] No `crm.yaml` contract change, no AI prompt touched → no aicert
      re-run needed

https://claude.ai/code/session_01BvvZzihBBSeaLc3WcZ6Xh1

---
## Summary by cubic
Fixes three stale tests that broke `main`'s integration suite after production stopped matching their expectations.

- Last-touch now requires an `activity_participant` "from" row; two tests seed inbound activity without one, so their last-inbound date comes back empty. Added `LinkActivitySender` to seed authorship at each call site.
- `lapsed` is now the correct engagement state for a contact whose only message predates the window; the test still asserted `untried`.
- `capture_counterparty_verdict`'s cadence moved to 10 minutes, so the hourly-cadence fixture kind switched to `capture_classify` instead of rewriting each time-arithmetic assertion.

<sup>Written for commit ae2fe7c91b713f2fc1ee05cd7ce5d839c010df85. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/4942?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Tests**
  - Improved integration test coverage for contact engagement states, activity attribution, relationship views, dismissal behavior, and scheduled processing.
  - Updated test scenarios to more accurately represent inbound activity senders and engagement that has lapsed beyond the relevant time window.
  - Refined test fixtures and expectations to better validate current behavior across these workflows.

### DIFF
--- a/backend/internal/compose/integration/org360/org360_lasttouch_integration_test.go
+++ b/backend/internal/compose/integration/org360/org360_lasttouch_integration_test.go
@@ -95,9 +95,10 @@ func TestStrengthFoldLeavesTheInboundAnchorEmptyWhenNobodyWroteIn(t *testing.T) 
 // The two are answering different questions. "They last wrote eighteen months
 // ago" is history and stays true however old it gets. The anchor is an action —
 // it is what a Follow up button opens a reply against — and it has to agree
-// with the state the counts report. This contact reads untried, because nothing
-// inside the window says otherwise; offering to answer a thread from last year
-// would contradict the same page's own summary.
+// with the state the counts report. This contact reads lapsed: an exchange
+// that is real, just older than the window, ranks warmer than a cold approach
+// and colder than an unanswered follow-up — offering to answer a thread from
+// last year would still contradict the same page's own summary.
 func TestStrengthFoldDropsAnInboundAnchorOlderThanTheWindow(t *testing.T) {
 	e := integration.Setup(t)
 	owner := integration.OwnerConn(t)
@@ -109,6 +110,7 @@ func TestStrengthFoldDropsAnInboundAnchorOlderThanTheWindow(t *testing.T) {
 	longAgo := org360Clock.AddDate(0, 0, -400)
 	stale := integration.AccountMailDirectedAt(t, owner, e.WS, "Re: 2025 tender", "inbound", longAgo)
 	integration.LinkActivity(t, owner, stale, "person", person)
+	integration.LinkActivitySender(t, owner, stale, person)
 
 	got := foldOneContact(t, e, org, person)
 
@@ -119,8 +121,8 @@ func TestStrengthFoldDropsAnInboundAnchorOlderThanTheWindow(t *testing.T) {
 		t.Fatalf("a reply from %s is outside the 90-day window and must not be offered as a reply anchor, got %s",
 			longAgo.Format("2006-01-02"), got.LastInboundActivity)
 	}
-	if people.EngagementOf(got) != people.EngagementUntried {
-		t.Fatalf("a contact whose only message predates the window reads as %q, want untried",
+	if people.EngagementOf(got) != people.EngagementLapsed {
+		t.Fatalf("a contact whose only message predates the window reads as %q, want lapsed",
 			people.EngagementOf(got))
 	}
 }
--- a/backend/internal/compose/integration/person_relationship_room_integration_test.go
+++ b/backend/internal/compose/integration/person_relationship_room_integration_test.go
@@ -598,6 +598,7 @@ func TestPerson360AssemblesEverySectionFromRealRows(t *testing.T) {
 		VALUES ($1, 'email', 'Re: pricing', 'body', '2026-08-01T09:00:00Z',
 		        'inbound', 'manual', 'human:x')`)
 	LinkActivity(t, owner, inbound, "person", mine)
+	LinkActivitySender(t, owner, inbound, mine)
 	task := SeedIDRow(t, owner, `INSERT INTO activity (id, kind, subject, occurred_at, due_at, is_done, source, captured_by)
 		VALUES ($1, 'task', 'Send the quote', '2026-07-28T09:00:00Z', '2026-07-30T09:00:00Z',
 		        false, 'manual', 'human:x')`)
@@ -700,6 +701,7 @@ func TestADismissalHoldsUntilTheEvidenceMoves(t *testing.T) {
 		VALUES ($1, 'email', 'Re: Following up', 'body', $2,
 		        'inbound', 'manual', 'human:x')`, roomAgo(time.Hour))
 	LinkActivity(t, owner, inbound, "person", mine)
+	LinkActivitySender(t, owner, inbound, mine)
 
 	reArmed, err := svc.Assemble(rep, personID)
 	if err != nil {
--- a/backend/internal/compose/integration/seed.go
+++ b/backend/internal/compose/integration/seed.go
@@ -101,6 +101,20 @@ func LinkActivity(t *testing.T, owner *pgx.Conn, activity ids.UUID, entityType s
 	}
 }
 
+// LinkActivitySender names WHO sent an activity — distinct from LinkActivity,
+// which only says the activity CONCERNS somebody. sender.go's SenderPredicate
+// (backend/internal/modules/people/sender.go) reads this row for "they wrote
+// last" answers, and a thread linked to everybody it concerns says nothing
+// about which of them is its author.
+func LinkActivitySender(t *testing.T, owner *pgx.Conn, activity, person ids.UUID) {
+	t.Helper()
+	if _, err := owner.Exec(context.Background(),
+		`INSERT INTO activity_participant (activity_id, person_id, role) VALUES ($1, $2, 'from')`,
+		activity, person); err != nil {
+		t.Fatal(err)
+	}
+}
+
 // SeedExtraWorkspace mints an additional tenant. archived names a workspace
 // nobody looks at any more, which still holds everything it held the day it was
 // archived — some passes are owed on it and some deliberately skip it, so a
--- a/backend/internal/platform/jobs/pass_integration_test.go
+++ b/backend/internal/platform/jobs/pass_integration_test.go
@@ -22,9 +22,11 @@ import (
 	"github.com/margince/margince/backend/internal/platform/jobs"
 )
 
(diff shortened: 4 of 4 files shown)


dungbanguyen:
I duplicated this work in https://github.com/margince/margince/pull/4943 (closed) before spotting yours — and got the first cluster wrong where you got it right. I read `capture/participant.go` as the only writer of `activity_participant` and concluded the authorship check was a regression; `activities/participantlog.go`'s `stampLoggedCounterparties` also stamps it on the manual log path, so a logged inbound already names its sender. Your `LinkActivitySender` is the right answer. Verified on unmodified `main`: an inbound call through `Activities.LogActivity` sets `last_inbound_at` with no production change at all.

Two suggestions, neither blocking — this should land as-is if you would rather not touch it.

**1. The cadence trap is still armed.** Swapping to `capture_classify` fixes today's red, and your comment names the fragility exactly ("a kind picked here must keep one for as long as this file does"). The next cadence revision reddens `main` the same way, having found nothing wrong. Deriving it removes the trap:

```go
func declaredCadence(t *testing.T) time.Duration {
	t.Helper()
	spec, ok := jobs.SpecFor(aScheduledKind)
	if !ok {
		t.Fatalf("%s is not a declared kind", aScheduledKind)
	}
	if spec.Cadence.Fixed == 0 {
		t.Fatalf("%s declares no fixed cadence", aScheduledKind)
	}
	return spec.Cadence.Fixed
}
```

Then `pass.Every != declaredCadence(t)` and `ran.Add(declaredCadence(t))`. What those tests are about is the plumbing that carries a declaration out to a reader — a scheduled row winning over a projection, the projection measuring from the tick rather than the finish. The number is a product decision, and a copy of it in the test is a second declaration that agrees only until somebody revises the first.

**2. The invariant behind cluster 1 has no compose-level test.** https://github.com/margince/margince/pull/4904 changed what "they wrote last" means and no test in this package holds it in both directions, which is part of why this went red instead of being caught. Worth a follow-up: over one contact, a hand-logged inbound moves the clock; a captured mail whose `from` is somebody ELSE on the same thread does not; that contact's own captured reply does. Each direction is the other's regression — a fix for either alone reintroduces the other. Happy to open that once this lands, on top of your `LinkActivitySender`.

dungbanguyen:
**Your integration failures are not yours.** Shards 2 and 4 (and the `ci` fan-in behind them) are red on `TestDOIIssuanceMintsNothingWhicheverPurposeIsNamed` and `TestAnOmittedRequiredIDIsNamedAndASuppliedOneStaysHidden/IssueDoubleOptInJSONBody.purpose_id` — both fail on plain `origin/main` (`d407de3a4`) from `ea9ee07e5` (https://github.com/margince/margince/pull/4931), returning 422 where they expect 409. https://github.com/margince/margince/pull/4945 fixes those.

I ran the full integration lane against a clean `origin/main` to settle this: eight tests fail in three packages. Six are the ones you fix here; the other two are https://github.com/margince/margince/pull/4945's. Separately, `TestEveryCitedMigrationExists` in the `gates` package is red from https://github.com/margince/margince/pull/4938 and takes `deterministic-gates`/`extension-reference`/coverage down on every open pull request — https://github.com/margince/margince/pull/4946 fixes that one and is green and `CLEAN` now.

Since PR checks run the merge commit, **none of the three fix branches can be green simultaneously** — each inherits the other two's breakages. Suggested order is https://github.com/margince/margince/pull/4946 → https://github.com/margince/margince/pull/4945 → this one; full table on https://github.com/margince/margince/issues/4852. Nothing here needs changing on that account.

Two non-blocking suggestions on the diff itself are in my earlier comment.

dungbanguyen:
Measured this branch on its own, against the current tip (`64bbd4905`), so the red shards are not a reason to hold it.

Merged `origin/fix/tests-behind-4904` onto `main` in a local worktree (clean) and ran every package that fails on the tip:

| | main alone | main + this branch |
|---|---|---|
| `platform/jobs` (3 cadence tests) | FAIL | **ok, 3.744s** |
| `compose/integration/org360` | FAIL | **ok, 33.618s** |
| `TestPerson360AssemblesEverySectionFromRealRows` | FAIL | **PASS** |
| `TestADismissalHoldsUntilTheEvidenceMoves` | FAIL | **PASS** |
| `TestDOIIssuanceMintsNothingWhicheverPurposeIsNamed` | FAIL | FAIL (unchanged) |
| `TestAnOmittedRequiredIDIsNamedAndASuppliedOneStaysHidden` | FAIL | FAIL (unchanged) |

**Six of the eight fixed, and the two that remain are byte-for-byte the ones already broken on `main`** by `ea9ee07e5` (https://github.com/margince/margince/pull/4931). This branch does not touch them and does not make them worse — shards 2 and 4 and the `ci` fan-in are red for that reason alone, which is why every one of your other 14 checks is green.

https://github.com/margince/margince/pull/4945 is the one that clears those two, and it currently needs a merge from `main` (it went `DIRTY` when https://github.com/margince/margince/pull/4946 landed). Since neither of you can be green while the other is unmerged, landing this one first is strictly forward: `main` goes from eight failures to two.

Not my call and not my branch — flagging the evidence so the red is not mistaken for something wrong here.

dungbanguyen:
Superseded by https://github.com/margince/margince/pull/4960, which lands your fix together with the other half of the deadlock — with your commits and authorship intact, because it MERGES your branch rather than cherry-picking it.

Why: `ci` runs against the merge commit, so this branch is judged against a `main` that still carries the other pull request's breakage, and vice versa. Neither could ever be green while the other was unmerged, and branch protection refuses a merge with red required checks. The choice was an `--admin` bypass on one of you, or one branch carrying both — `main` + both is green, so that route needs no bypass at all. I took the second on the repository owner's call.

Verified on the combined branch: `go build` ok, `gates` ok, `platform/jobs` ok, `org360` ok, and the person/consent tests pass. Net diff is seven files, exactly your two fixes.

If you would rather land your own pull request instead, say so and I will close https://github.com/margince/margince/pull/4960 — nothing here is lost either way.

dungbanguyen:
**Evidence that your branch and mine are green together**, in case it helps you land this one.

I stacked #4945 on `fix/tests-behind-4904` and ran `ci.yml` against the union:

> https://github.com/margince/margince/actions/runs/34194326769 — **completed success**

That was your commit `ae2fe7c91` plus `main` plus my consent fix, so the nine tests `main` is red on all pass with both changes present. Neither PR can reach a green `ci` alone — this branch fails only on the two `IssueDoubleOptIn` tests #4945 fixes, and #4945 failed only on the six this one fixes.

I have since **unstacked**: #4945 is back to a single commit on current `main` (which has moved twice more since — #4950, #4946, #4955), so nothing of yours is riding in my PR any more and this one's merge path is untouched. I did not and will not touch this branch.

Order does not matter now that the union is proven; whichever lands first, the other rebases and goes green on its own.

tinnhlt-gra:
Superseded by #4960, which merges this branch together with #4945 to resolve the mutual-deadlock CI failure both were hitting (each PR's merge commit still carried the other's breakage). My commits and authorship are preserved there. Closing this PR.

PR #4943, opened 2026-09-08
main is red: the three breakages one merge left behind

dungbanguyen:
`main` is red on four integration shards. Every failure traces to one merge —
"The webinar defect list: nine fixes" — which landed three separate breakages.
One is a product regression; two are tests that describe behaviour the same
merge deliberately replaced.

Reproduced locally on the tip before anything was changed, and each fix
mutation-tested one at a time.

## 1. A message a rep logged by hand is still a message they sent

`TestADismissalHoldsUntilTheEvidenceMoves`,
`TestPerson360AssemblesEverySectionFromRealRows` — and a real defect behind
them.

The sender predicate the signature repair introduced asks for an
`activity_participant` row with `role='from'`. **Only `capture/participant.go`
ever writes one.** A call or an email a rep logs BY HAND carries no participant
at all, so the person page read every one of them as nobody having written:
`last_inbound_at` never moved, the re-engaged rung never fired, and a dismissal
put away against "they went quiet" stayed put after they answered. On an
installation with no mailbox connected that is every message there is.

The strict predicate is untouched and the enrichment pass keeps using it: a
signature block mined off a message whose author is unproven is exactly the
defect it exists to refuse, and absence of evidence there has to read as no.

A last-touch fold asks the weaker question — *does anything recorded say
somebody ELSE wrote this* — so it gets its own predicate. The arm is added for a
message with NO participant row whatsoever and only that. A captured mail always
carries its parties, machine notifications included, so Marcus's mail to Judith
still names Marcus and the confusion the original fix removed stays removed.

One test holds both directions, because each is the other's regression: a fix
for either alone reintroduces the other, and neither shows up in a test that
seeds only the shape it is about. Reverting either half fails it — measured,
one mutation at a time.

## 2. The cadence tests stop retyping the declaration

`TestAScheduledRunIsWhenThePassRuns`, `TestAKindWithNoHistoryNamesNoTime`,
`TestWithNothingScheduledThePassIsTheLastRunPlusItsCadence`.

`capture_counterparty_verdict` moved from an hour to ten minutes, deliberately.
Three tests asserted a literal `time.Hour` and went red having found nothing
wrong.

They now read the interval from the spec table. What they are about is the
plumbing that carries a declaration out to a reader — a scheduled row winning
over a projection, the projection measuring from the tick rather than the
finish, a kind with no history still knowing its own interval. The number is a
product decision and none of their business, and a copy of it in the test was a
second declaration that agreed only until somebody revised the first. The next
cadence revision cannot redden `main` this way.

## 3. The account roster's new lapsed state

`TestStrengthFoldDropsAnInboundAnchorOlderThanTheWindow`.

`lapsed` was added for precisely this contact — an exchange that is real and
older than the window. The test asserted `untried`, which is the reading the new
state exists to correct: untried means nobody has written at all, and saying it
about a contact whose own message the row prints beside it put a contradiction
on one line ("They wrote, 1 June" next to "Not approached"). The expectation and
the comment above it both move to the state the fold now returns.

---

All six previously-failing tests pass locally, and the whole `compose/integration` package is green on this branch — 2000 tests, `ok … 370.490s`. `make check-all` is clean apart from `TestPublicTreeCitesNoDocumentGitIgnores`, which fails on this machine only (a `.git/info/exclude` entry, not the repo) and is green in CI.


dungbanguyen:
Closing this. https://github.com/margince/margince/pull/4942 covers the same three clusters, was opened three minutes earlier, and diagnosed the first one correctly where I did not.

**My production change here was wrong.** I claimed the last-inbound authorship check was a regression because `capture/participant.go` is the only writer of `activity_participant`, so a hand-logged message could never satisfy it. That is false: `activities/participantlog.go`'s `stampLoggedCounterparties` stamps the counterparties on the manual log path too, and for an inbound activity the linked person gets `role='from'` — exactly what `SenderPredicate` reads. My grep missed that file and I built the rest on top of it.

Verified against unmodified `origin/main` (`d407de3a4") by logging an inbound call through `Activities.LogActivity` and reading the page: `last inbound = 2026-08-04 17:00:00`. The real write path already works. The two failing tests seed activities with raw SQL and never call the real writer, which is AGENTS.md's own rule 6 — a test that supplies its own version of production proves nothing about production — and https://github.com/margince/margince/pull/4942's `LinkActivitySender` is the right fixture answer.

So the `AuthoredPredicate` I added would have loosened a correct authorship check to accommodate a fixture gap. Good thing it did not land.

Two things from this branch are still worth having, offered on https://github.com/margince/margince/pull/4942 rather than kept here.

PR #4945, opened 2026-09-08
fix(consent): a purpose that names nothing is not an archived one

dungbanguyen:
**main is red on the integration lane.** #4931 ("the double-opt-in link is mailed again, by the honest path") gave the door a working mint, and two tests still hold it to the refusal it replaced. No open PR covers these two: #4942 and #4943 fix the other three breakages, and neither touches consent — so this is that fix, and this note is the "say so" the rulebook asks for.

```
consent_integration_test.go:408: issuance under marketing_email → 201, want 409
consent_integration_test.go:408: issuance under transactional  → 422, want 409
requiredbodyids_http_integration_test.go:252: → 422, want 409 …
  {Code:validation_error Detail:this purpose is archived, so a link asking somebody to confirm it could not be answered}
```

## The stale test is deleted, not rewritten

`TestDOIIssuanceMintsNothingWhicheverPurposeIsNamed` said what it was in its own comment:

> This replaces a test of the issuance round trip … None of that exists now: the endpoint returns a conflict for every caller.

It was a placeholder holding the one property a retired capability had left. The capability is back, and #4931's own `doidoor_integration_test.go` covers it through the handler arm by arm — the mint, the absent plaintext, the missing mailbox, the mail's wording, the archived purpose, the purpose that needs no confirmation. Rewriting this one would be a second copy of that file's subject, free to disagree with it.

## The second failure is a real defect the restored door brought with it

`requireConfirmablePurposeTx` read the purpose with a live-only filter:

```sql
SELECT requires_double_opt_in FROM consent_purpose WHERE id = $1 AND archived_at IS NULL
```

so a purpose that **never existed** and a purpose that **was archived** both came back as `ErrNoRows` — and it answered *archived* for both. Two things are wrong with that:

1. **It says something untrue.** An operator who typed a wrong id is sent looking for a purpose to unarchive that nobody ever created.
2. **It answers the wrong status.** 422 naming `purpose_id` where every other required body id on this surface answers 404. That rule is not decoration — it is what keeps "no such row" indistinguishable from "a row you may not see", and `requiredbodyids_http_integration_test.go` is the gate that holds it. The DOI case carried a documented exception (`suppliedStatus: Conflict`) that was honest while the endpoint refused everybody and resolved nothing. It resolves the purpose now, so the exception is gone.

The read drops the filter and carries `archived_at IS NOT NULL` instead, separating the three cases: **no row → `ErrNotFound`**, **archived → its own sentence**, **live but not double-opt-in → its own sentence**. #4931's `TestTheDoubleOptInDoorRefusesAnArchivedPurpose` still passes on the archived arm, which is what confirms the split kept the case it was distinguishing.

## Checks

`make check-backend` green. `make test-it DIR=backend/internal/modules/consent` green (the whole DOI door suite). `make test-it DIR=backend/internal/compose/integration RUN='TestAnOmittedRequiredIDIsNamedAndASuppliedOneStaysHidden'` green.

**Mutation check:** collapsing the split back — returning the archived sentence for `ErrNoRows` again — fails `IssueDoubleOptInJSONBody.purpose_id/supplied_but_invisible_stays_a_404` and leaves the omitted-id half passing, which is the pair that says the fix is about the right half.

### DIFF
--- a/backend/internal/compose/integration/consent_integration_test.go
+++ b/backend/internal/compose/integration/consent_integration_test.go
@@ -382,46 +382,6 @@ func (c *consentEnv) grantMarketingByConfirmLink(t *testing.T) string {
 	return token
 }
 
-// Issuance mints nothing, and says so.
-//
-// This replaces a test of the issuance round trip — non-DOI purposes refused,
-// a fresh token superseding a stale one, both mints audited. None of that
-// exists now: the endpoint returns a conflict for every caller, nothing writes
-// consent_doi_token, and the redemption arm is gone. A double-opt-in purpose
-// confirms through a spent confirm-details link instead, which
-// TestConsentDoubleOptInNorm exercises.
-//
-// What is worth holding is that the refusal is a refusal: the same answer
-// whatever purpose is named, and no row behind it. An endpoint in the public
-// contract that quietly minted again would hand an operator both halves of a
-// round trip whose only value is that the subject completed one of them.
-func TestDOIIssuanceMintsNothingWhicheverPurposeIsNamed(t *testing.T) {
-	c := setupConsent(t)
-
-	// A purpose that requires double opt-in, and one that does not, get the
-	// same answer. The endpoint resolves no id, so it can tell a caller nothing
-	// about which purposes exist.
-	for _, purpose := range []string{"marketing_email", "transactional"} {
-		if status := c.Call(t, "POST", "/v1/people/"+c.personID+"/consent/double-opt-in", AnyMap{
-			"purpose_id": c.purposes[purpose],
-		}, nil, nil); status != http.StatusConflict {
-			t.Errorf("issuance under %s → %d, want 409", purpose, status)
-		}
-	}
-
-	// And nothing was written. The table keeps its history and takes no new
-	// rows, so an unredeemed invitation cannot outlive the change.
-	var audit struct {
-		Data []AnyMap `json:"data"`
-	}
-	if status := c.Call(t, "GET", "/v1/audit-log?entity_type=consent_doi_token", nil, nil, &audit); status != http.StatusOK {
-		t.Fatalf("audit read → %d", status)
-	}
-	if len(audit.Data) != 0 {
-		t.Fatalf("a refused issuance audited %d row(s), want none", len(audit.Data))
-	}
-}
-
 func TestConsentProofLogIsAppendOnlyAndIdempotent(t *testing.T) {
 	c := setupConsent(t)
 	grant := func() int {
--- a/backend/internal/compose/integration/requiredbodyids_http_integration_test.go
+++ b/backend/internal/compose/integration/requiredbodyids_http_integration_test.go
@@ -181,14 +181,10 @@ func requiredIDCases(f requiredIDFixtures, absent string) map[string]requiredIDC
 		"IssueDoubleOptInJSONBody.purpose_id": {
 			method: "POST", path: "/v1/people/" + f.person + "/consent/double-opt-in",
 			omitted: AnyMap{}, supplied: AnyMap{"purpose_id": absent}, field: "purpose_id",
-			// This endpoint mints nothing and resolves nothing: it refuses every
-			// caller with a conflict, whatever purpose they name. So it has no row
-			// to hide and no 404 to give — and equally no way to enumerate, since
-			// a visible purpose, an invisible one and one that never existed all
-			// get the identical answer. The omitted half above is still held: a
-			// body-id probe runs ahead of the refusal precisely so this endpoint
-			// does not become the one place a missing id goes unnamed.
-			suppliedStatus: http.StatusConflict,
+			// No exception any more. This endpoint used to refuse every caller
+			// with a conflict, so it had no row to hide; it resolves the purpose
+			// and mints a link now, which puts it back under the rule every other
+			// required body id follows.
 		},
 		"ApplyTagRequest.entity_id": {
 			method: "POST", path: "/v1/tags/" + f.tag + "/apply",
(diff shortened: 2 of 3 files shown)


dungbanguyen:
**Your two failing checks are not yours.** `extension-reference` and `integration / integration unit coverage` are both red on `TestEveryCitedMigrationExists`, which fails on plain `origin/main` (`d407de3a4`) — `frontend/src/screens/settings-grants.test.tsx:89` cites migration 0261, which never existed. It came in with https://github.com/margince/margince/pull/4938, it is in the `gates` package, and it is therefore red on every open pull request in the repository right now.

https://github.com/margince/margince/pull/4946 fixes it (comment-only, all 13 checks green, `CLEAN`). Once that lands, re-run these two and they should go green without you touching anything here.

For context on the ordering: I ran the full integration lane against a clean `origin/main` and `main` is red from exactly eight tests in three packages plus that gate. Two of the eight are yours; the other six are https://github.com/margince/margince/pull/4942's. Because PR checks run the merge commit, none of the three fix branches can be green at the same time — each inherits the others' breakages. Full enumeration and suggested order on https://github.com/margince/margince/issues/4852.

dungbanguyen:
Rebased onto **#4942's branch** (`fix/tests-behind-4904`), and onto current `main` — which now carries #4946, so the citation gate is clear.

The two PRs were deadlocked: this branch failed only on the six tests #4942 fixes, and #4942 failed only on the two this branch fixes. `ci` is the required check and neither could reach it alone. This is the stacked side of that, because #4942 is not my branch to touch.

So the commit above mine here is Tin's, unchanged, and it disappears from this PR the moment #4942 lands — at which point I rebase back onto plain `main`. Nothing in it is mine to review; the diff to read is the top commit.

dungbanguyen:
Heads up: this went `DIRTY` a moment ago and it is my doing. https://github.com/margince/margince/pull/4946 merged (`64bbd4905`) and touched `frontend/src/screens/settings-grants.test.tsx`, which this branch also carries.

The conflict is small and there is **no hazard in it** — I checked: your copy of that file does not carry the `migration 0261` citation that https://github.com/margince/margince/pull/4946 removed, so merging this cannot reintroduce the dangling-citation failure. It is two independent edits to the same file. Taking `main`'s version of the comment block around the "opens Seats & license for a lone license read" case is the whole resolution.

Your merge base is `edb994976`, several commits behind, so a `git merge origin/main` (or rebase) should clear it.

The good news: the two checks that were failing here for reasons that were not yours — `extension-reference` and `integration / integration unit coverage` — were red on `TestEveryCitedMigrationExists`, and https://github.com/margince/margince/pull/4946 fixed that on `main`. After you pick up `main` they should go green.

I have not touched this branch and will not — it is not mine to rebase.

dungbanguyen:
I resolved your `confirmtoken.go` conflict locally to check it was tractable. Posting the result so you can take it in one paste rather than re-deriving it — nothing is pushed and I have not touched your branch.

The conflict is with `ea9ee07e5`'s version of `requireConfirmablePurposeTx`. **Your side is the one to keep**, all three hunks: main reads `consent_purpose` with `archived_at IS NULL`, so a purpose that does not exist and one that is archived both come back no-rows and the code has to guess — and it always guesses archived. Dropping the filter is what separates them, which is the whole point of this pull request. Taking main's side would keep the bug.

Resolved, `gofmt`-clean:

```go
func requireConfirmablePurposeTx(ctx context.Context, tx pgx.Tx, purposeID ids.PurposeID) error {
	if purposeID.UUID == (ids.UUID{}) {
		return nil
	}
	var requiresDOI, archived bool
	// Read WITHOUT the live filter, so absence and archival stay separable. The
	// filtered read answered no-rows for both and had to guess which; it always
	// guessed archived.
	err := tx.QueryRow(ctx,
		`SELECT requires_double_opt_in, archived_at IS NOT NULL FROM consent_purpose
		  WHERE id = $1`, purposeID).Scan(&requiresDOI, &archived)
	if errors.Is(err, pgx.ErrNoRows) {
		return apperrors.ErrNotFound
	}
	if err != nil {
		return err
	}
	if archived {
		return &ValidationError{
			Field:  purposeIDField,
			Reason: "this purpose is archived, so a link asking somebody to confirm it could not be answered",
		}
	}
	if !requiresDOI {
		return &ValidationError{
			Field:  purposeIDField,
			Reason: "this purpose is not confirmed by double opt-in, so there is nothing for a mailed link to ask",
		}
	}
	return nil
}
```

The doc comment above it takes your "THREE answers" version wholesale; main's "TWO refusals" paragraph goes.

Everything else merged clean — `consent_integration_test.go` and `requiredbodyids_http_integration_test.go` auto-merged, and so did all of https://github.com/margince/margince/pull/4942 alongside it. So once this is resolved, main + https://github.com/margince/margince/pull/4942 + this is a green tree; I verified that exact combination against a clean checkout earlier.

dungbanguyen:
Superseded by https://github.com/margince/margince/pull/4960, which lands your fix together with the other half of the deadlock — with your commits and authorship intact, because it MERGES your branch rather than cherry-picking it.

Why: `ci` runs against the merge commit, so this branch is judged against a `main` that still carries the other pull request's breakage, and vice versa. Neither could ever be green while the other was unmerged, and branch protection refuses a merge with red required checks. The choice was an `--admin` bypass on one of you, or one branch carrying both — `main` + both is green, so that route needs no bypass at all. I took the second on the repository owner's call.

Verified on the combined branch: `go build` ok, `gates` ok, `platform/jobs` ok, `org360` ok, and the person/consent tests pass. Net diff is seven files, exactly your two fixes.

If you would rather land your own pull request instead, say so and I will close https://github.com/margince/margince/pull/4960 — nothing here is lost either way.

dungbanguyen:
Superseded by #4960, which merged: main is green and this commit landed inside it with authorship preserved. Closing.

PR #4948, opened 2026-09-08
fix(tests): the license grant is stated, not cited

dungbanguyen:
**main is red on `TestEveryCitedMigrationExists`, and #4938 is why** — my own merge. It blocks every open PR, so this is the fix.

```
frontend/src/screens/settings-grants.test.tsx:89 cites migration 0261
  A citation written now against a version that never existed is a dead end from birth.
```

## What happened

The register is keyed by **path**. `settings-nav.test.tsx` had a ratified line for its 0261 citation; splitting the suite copied that sentence into `settings-grants.test.tsx`, where the same citation is *new*. A rename or a split turns a ratified dead end into an unratified one, and the gate is right to say so.

## Why the line was not simply re-added

Adding the copy would have left **two** register lines where there was one, and the register's own note is exactly about that:

> a file acquires no citation of a version it does not ALREADY cite … Per FILE rather than a total, because a total lets a deletion anywhere fund a new dead citation elsewhere.

So both sentences state the invariant instead. Neither claim needed the version: what they say is that `license` is a **grant** the seeded role matrix gives admin and ops and nobody else — which is why a role an installation has edited can hold it, and why the case is not a role name in disguise. Which migration first wrote that row is not part of it, and a reader who wanted it has `git log` for the matrix.

The old register line goes with the citation it recorded, so the backlog is one shorter than before rather than one longer.

## Checks

`go test -count=1 -run TestEveryCitedMigrationExists ./gates` green. Both suites still pass — 40 tests, unchanged.

## Summary by CodeRabbit

- **Documentation**
  - Updated test comments to clarify the source and meaning of the `license` grant.
  - Removed an outdated entry from the dangling-citation register.

### DIFF
--- a/backend/danglingcitations.txt
+++ b/backend/danglingcitations.txt
@@ -243,7 +243,6 @@ frontend/src/screens/embedreindex.test.tsx migration 0115
 frontend/src/screens/embedreindex.tsx migration 0115
 frontend/src/screens/relationships.test.tsx migration 0007
 frontend/src/screens/relationships.tsx migration 0007
-frontend/src/screens/settings-nav.test.tsx migration 0261
 scripts/deploy/db-bootstrap.sql migration 0015
 scripts/lib-testdb.sh migration 0015
 scripts/test-migration-versions.sh core 0002
--- a/frontend/src/screens/settings-grants.test.tsx
+++ b/frontend/src/screens/settings-grants.test.tsx
@@ -87,9 +87,9 @@ describe("the grant that opens one settings page", () => {
 
   it("opens Seats & license for a lone license read", async () => {
     // `LicenseCard` calls `/installation/license` and nothing else, so this is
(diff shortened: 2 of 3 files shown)


dungbanguyen:
Closing this in favour of #4946, which was opened about half an hour earlier and fixes the same gate failure. Both do the same thing to `settings-grants.test.tsx` — state the claim without the version — and #4946 makes the sharper point that the grant never was a migration: it lives in identity's role defaults, in Go, so the citation was wrong about where the fact is rather than merely stale.

One difference, and it is not worth a competing PR: this branch also dropped the *same* citation from `settings-nav.test.tsx` and removed its now-unneeded `danglingcitations.txt` line. That one is ratified and green, so it is a tidy-up rather than a fix — noting it here so it is on the record, and leaving the register alone.

PR #4949, opened 2026-09-08
feat(records): "not a company" is one transaction

### LINKED ISSUE #1132: "Not a company" must be one backend transaction, not two frontend writes
The company page needs a reject action: archive the record **and** refuse its domain a company, so the same mail does not recreate it next week. A first attempt (PR #1129, closed unmerged) did this as two HTTP calls from the frontend. Adversarial review found four defects, three of them blocking. Recording the shape the next attempt should take.

## It is one logical mutation and needs one transaction

Two independent calls — `PUT /capture/blocked-domains` then `DELETE /organizations/{id}` — cannot be atomic. Ordering block-first is the better of the two (a failure leaves a refused domain whose company is still visible, rather than a deleted record whose domain still mints new ones), but it still commits half the intent. The repo's write shape wants domain row + audit + outbox in one transaction; a composite action deserves the same.

**Proposed:** `POST /organizations/{id}/reject` taking a reason, which in one transaction validates the organization is live, visible and not the anchor, reads its *current* primary domain server-side, records the suppression with a human admission source, archives the organization, and writes audit and outbox.

## What the two-write version got wrong

1. **Wrong capability gate.** The button was offered on liveness + native-mode + a primary domain existing. It never checked `organization:delete`. Seeded reps hold `organization:update` but not delete, so the first write succeeded and the second predictably 403'd — leaving the domain suppressed and the company still there.

2. **The anchor was not excluded.** An admin could suppress the installation's own primary domain; the backend anchor guard then refuses the archive, leaving the anchor live with its own domain refused.

3. **Stale domain snapshot.** The domain came from the client's copy of the record and was not bound to the organization or its version. With several domains, a concurrent primary change means suppressing the old primary while archiving a company whose current primary is different — and mail from the unsuppressed domain recreates it.

4. **Partial success looked like total failure.** The UI surfaced only the last error, so a failed archive after a successful suppression told the user nothing had happened. Undoing then required a separate admission change, which also re-opens domain triage.

## Keep

The reasoning in the closed PR: archive alone is not enough, because the record comes back and the person who deleted it learns nothing about why. The action should be offered only where there is a domain to refuse — a company somebody typed in by hand was never derived from mail.

dungbanguyen:
Closes #1132.

`POST /organizations/{id}/reject` — archive the company **and** refuse its domain another, in one transaction.

Archiving alone does not settle it. The record was minted from mail, so the next message on the same domain mints it again and the person who archived it a week ago learns nothing about why it is back. The refusal is what stops that, and it belongs to the **domain** rather than to the record.

## The four defects, answered by construction

The ticket records what adversarial review found in #1129's two-call version. None of these is fixed by being careful:

**1. Wrong capability gate.** Both grants are asked for before anything is written — `organization:delete` for the archive, `organization:update` for the standing domain decision. The old version was offered on liveness + native mode + a domain existing, so a seeded rep suppressed the domain on the first call and took a predictable 403 on the second.

**2. The anchor was not excluded.** `refuseIfAnchor` runs first, ahead of the domain read — deliberately, and not only through the archive's own copy of that guard: whether the installation's own company can be rejected must not depend on whether it happens to carry a domain. Its refusal joins the anchor suite's table, which asserts the surviving row as well as the error.

**3. Stale domain snapshot.** The domain is read **here**, from `organization_domain` where `is_primary AND archived_at IS NULL`, `FOR UPDATE`, inside the transaction that writes. **There is no domain field on the request.** A domain the caller carried is a snapshot of what the page was showing; with several domains a primary that moved since the page loaded suppressed the address nobody writes from any more while the company came back tomorrow on the one that was never refused.

**4. Partial success looked like total failure.** One transaction has no middle. The response carries both halves — the archived company and the decision recorded — so the reader is told which domain was *actually* refused rather than which one their page happened to hold.

## Two helpers out, no second copy in

- `archiveOrganizationTx` — the archive's own transaction body, extracted so `ArchiveOrganization` and the rejection share it. The cascade retires four tables and drops two more; a second copy is the one that forgets one, and a live child under an archived parent keeps feeding the list its own table serves.
- `recordHumanDomainAdmissionTx` — the human decision's row, its re-ask and its audit door, so the sticky rule cannot differ between the blocked-domain admin surface and this verb.

The wire conversion for a `BlockedDomain` moves into `people` beside the type, because two transports serve it now.

## Authority and mode

**Human-only** (`x-agent-access`), like the rest of the capture posture: the admission this writes is a human one and therefore **sticky** against every later machine verdict. An agent minting one would launder a machine judgement into a decision no verdict is allowed to revisit — which is the guard `claimedByHuman` already exists to hold elsewhere.

**Refused in overlay mode** with the declared `unsupported_by_sor` sentinel. The native organization table holds none of an overlay workspace's records, so the unguarded verb would answer "not found" about a company the reader is looking at — ADR-0018 takes the other answer. The write-admission guard next door cannot cover this one: it keys off the generated policy table's tool classification, and a human-only operation naming no tool is let through as a governance write. That default is right for the rest of the capture posture and wrong for the half of this verb that archives a record, so it gets its own shadow. `refuseReportInOverlayMode` is renamed `refuseInOverlayMode` — a third caller from another subject made the old name a lie.

## On the page

The control sits beside Archive in the company's overflow menu, drawn only when **both** grants are held and only when there **is** a primary domain — a company somebody typed in by hand was never derived from mail, so no refusal would stop anything and a control that is only ever refused should not be drawn. Absent rather than disabled, per STATE-4a: those are permission and record causes, not state.

The confirm dialog names the domain the page holds and requires the reason. `useArchiveRecord`'s `archivedMessage` now also takes a function of the result, so the toast names the domain **the server refused** — composing that sentence before the call could name the wrong one, and telling somebody the wrong domain is refused is worse than telling them nothing.

## Checks

`make check-backend` and `make check-fe` green. `make test-it DIR=backend/internal/modules/people` and `.../internal/compose/integration` green.

**Mutation checks**, each killing exactly the case that owns its claim and no other:

| mutation | killed |
|---|---|
| drop the `organization:delete` gate | needs-both-grants |
| drop the anchor guard | the anchor suite's table |
| read any domain instead of the primary | refuses-the-domain-the-company-uses-now |
| record `admitted` instead of `suppressed` | archives-and-keeps-the-domain-from-minting-another |
| drop the delete grant from the button | draws-nothing-for-a-seat-holding-only-the-domain-grant |
| put the domain back in the request body | one-request-and-reports-the-server's-domain |
| toast the page's domain instead of the server's | the same |
| drop the no-domain condition | draws-nothing-for-a-company-with-no-primary-domain |

Four gates asked for their own entries and got them: the archive's two cross-store waivers re-key to the extracted function, the admission's audit-only and before-image waivers re-key to the extracted helper, the route joins `replayableOperations`, and `RejectOrganizationResponse` is classified in `rowScopedResponses` — a rejection hands back a row-scoped record, and the probe resolves against the organization because the admission beside it has no scope of its own.

> CI on this branch will show the three breakages `main` is currently red on (#4945, #4948, #4943 fix them); the people and compose/integration lanes pass here with those excluded.

## Summary by CodeRabbit

- **New Features**
  - Added a “Not a company” action for company records.
  - Users can provide a reason to archive the company and prevent its primary domain from creating another company.
  - Added confirmation, validation, permission checks, version protection, and clear success/error feedback.
  - Added the rejection endpoint and responses for integrations.
- **Localization**
  - Added English, German, and Vietnamese translations for the rejection workflow.


dungbanguyen:
All seven addressed in `83694d8d9`. Six were defects; one I took as a correctness-by-construction change I cannot prove with a test, and say so below.

**The lock order (P2) is the one that mattered, and the analysis is exactly right.** The rejection read the primary domain `FOR UPDATE` and only then archived, while a plain `ArchiveOrganization` takes the company row under its guarded patch and updates `organization_domain` after. Two transactions each holding what the other wants — Postgres breaks that by aborting one, so a rejection or somebody's archive fails at random under precisely the concurrency that makes it hard to reproduce. The company row is taken first now, so both paths climb in the same order. The archive's own acquire of a row this transaction already holds costs nothing.

**Both gates, only one tested (P3).** Correct: `e.as()` holds update and not delete, so a regression dropping the `ActionUpdate` gate would have gone uncaught. The case is a table over both seats now — update-without-delete and delete-without-update — each asserting the refusal and that nothing was written. Mutation-checked: dropping the update gate fails the second row and only that row.

**Whitespace-only reason (P2).** Correct — `minLength: 1` admits `"   "`, which the handler then refuses after trimming, a 422 the contract gave no way to predict. The schema carries `pattern: '\S'` beside the length bounds.

**Idempotency key (P2).** Correct. One key per confirmation, held across retries of that confirmation and minted fresh when the dialog reopens — a second decision about the same company is a second decision, and reusing the key would replay the first one's answer. It rides `ifMatch`'s own second argument, because that helper owns the header map and a sibling `header` beside it is silently overwritten. Two cases cover it: the key's presence and shape on the request, and that a second confirmation carries a different one.

**"refuse its domain another" (P3)** and **"refuses {domain} a company" (P3).** Both correct. The summary reads "refuse its domain as one"; the English copy now says "as a company", matching the German and Vietnamese.

**The `reason` closure (P2)** — changed, but I could not make it fail. It passes through `mutate()` variables now, which is the library's idiom for exactly this hazard and costs nothing (the hook's type parameter defaults to `void`, so no other caller changes). What I cannot do is prove the race: a click handler belongs to the latest committed render, so reproducing a stale `reason` needs a render boundary this component does not cross, and the mutation that reverts it leaves all six cases green. So this one is a change I made on the strength of the argument rather than a defect I demonstrated — flagging that rather than implying a test holds it.

dungbanguyen:
Both valid, and they were mis-attributed in opposite directions. Fixed.

**The required reason (P2)** — correct, and I had already removed the transport's duplicate copy in the previous round for the reason you give: two writers of one refusal spelling the same field, code and sentence, so neither a test nor a reader could tell which answered. What I did not do was fix the case's own claim about itself.

The refusal is the **store's**, and it holds for every caller rather than only for this door. The cases say so now, and say what they actually prove: that it reaches the wire as a 422 naming its field rather than as the internal error an unclassified store failure would answer. They can run without Postgres because the store refuses before it opens a transaction — which is a fact about where the check sits, and now stated rather than relied on silently. The **length** ceiling is the transport's own and has no twin, which is why it stays there and why its case is the one the mutation kills.

**The duplicate overlay cases (P3)** — also correct. They exercised the shared `refuseInOverlayMode`, which `nativeonlytools_test.go` already holds both answers for; two copies keyed on this verb's name would prove the same two lines twice and go stale in the same breath. They are gone.

What was genuinely uncovered is the **wiring** — that `Server.RejectOrganization` runs the guard *before* delegating — and you are right that neither case reached it. `sorDispatch` is the concrete `*Dispatcher`, so no fake can get in, and every write shadow in this tree is in the same position. It is asserted from the source instead: the guard appears, the delegation appears, and the guard comes first. That is a small claim, but a shadow that loses its guard is wrong only in overlay mode, which is exactly where nobody looks.

PR #4951, opened 2026-09-08
fix(360): a project-scoped page derives its numbers from that project

### LINKED ISSUE #2291: A project-scoped 360 still derives health, relationship strength, engagement state and suggestions from every project's activity
Found by Codex review of #2289 and deferred so the scoped timeline could land.

`GET /people/{id}/360?project_id=` and `GET /organizations/{id}/360?project_id=` narrow the timeline, next steps, next meeting, last touch and since-last-visit sections to one project. Three groups of derived sections still fold every activity on the record:

- org360 engagement state and suggestions: `suggestionInputsOnce` → `gatherSuggestionInputs` (backend/internal/compose/org360/suggestionreads.go ~65-86, ~380-415) — `newestMessage` and `hasOpenTask` search the whole account, so a no-reply suggestion on a page scoped to project A can cite project B's activity id.
- person360 relationship strength and relationship changes: `strengthSection` / `relationshipChangesSection` (person360/sections.go ~24-42 → people/strength.go ~351-383, people/relationshipchange.go ~61-104) — `contributing_activity_ids`, counts and bucket are driven by every project.
- org360 health: `accountstate.go` ~214-266 and ~364-388 (`lastMeetingAt`, active_contacts, reply_balance, single_threaded) use unscoped aggregates.

Why deferred: these are aggregates, not the timeline the scope was built for, and scoping them means threading the project predicate into two people-module reads. Until fixed, a scoped 360 shows the right timeline under a health/strength number computed across projects.

Fix: pass `AssembleOptions.ProjectID` into each read and append `activities.ActivityWithinProject` (or the people-module twin) to every activity-reading statement in them; assert with the two-engagement fixture in `activity_projectscope_integration_test.go`.

dungbanguyen:
Closes #2291.

The project scope reached the timeline, next steps, next meeting, last touch and since-last-visit. It stopped at the **aggregates**. A page narrowed to one body of work still scored the relationship, counted the account's active contacts, dated its last meeting and raised its suggestions across *every* project — the right timeline under a number computed somewhere else, and advice citing a message the reader cannot see on the page to check it against.

## What changed

The three groups the issue names collapse to **two** reads. org360's `active_contacts`, `reply_balance` and `single_threaded` are all folded from the contact strengths, so narrowing that one read narrows all three.

- `people.StrengthForOrgContacts` / `PersonStrengthTx` / `PersonRelationshipChangesTx` take the body of work and apply it to every activity-reading statement in them.
- `org360`'s last-meeting read and the three suggestion reads (`newestMessage`, `hasOpenTask`, `gatherSuggestionInputs`) append the same predicate.
- `person360`'s strength and relationship-change sections pass `AssembleOptions.ProjectID` through.

Callers that deliberately stay unscoped pass `nil` with the reason written beside them.

## One predicate, not a fourth spelling

The people module already spelled this predicate, with the reason it is a copy rather than an import written beside it — a module never imports a sibling, so the timeline's, search's and people's spellings are a deliberate set of three that change together. This generalises the people-module one by **table alias** instead of adding a fourth.

The strength read narrows **both** of its aliases — the one the counts fold over and the one the cited message comes from — or the number and the message it cites come from different bodies of work.

## Proof

Against the existing two-engagement fixture, **both ways round every time**: the scoped read must drop the other engagement *and* the unscoped read of the same account must still carry it. A test asserting only the scoped half passes against an assembly that quietly lost a row.

- `TestAScopedAccountPageDerivesItsHealthFromOneEngagement` — active contacts (2 vs 3, the bystander speaks only on the other engagement), last-inbound age, and last meeting.
- `TestAScopedAccountPageChasesOnlyTheEngagementItShows` — the no-reply rule chases the page's own unanswered mail, never the other engagement's. Its own fixture: the shared one's newest exchange is a booked meeting, which is the rule correctly declining to chase someone we are about to see.
- `TestPerson360ScopedToOneProjectScoresFromOneEngagement` — the score's `contributing_activity_ids`, which is where a score says out loud what a reader can check it against.

Each production change was reverted in turn and the matching test confirmed failing.

## Note on CI

Four integration tests are red on `main` at the tip and are unrelated to this branch — `TestPerson360AssemblesEverySectionFromRealRows`, `TestADismissalHoldsUntilTheEvidenceMoves`, `TestDOIIssuanceMintsNothingWhicheverPurposeIsNamed` and `TestAnOmittedRequiredIDIsNamedAndASuppliedOneStaysHidden`. They are covered by #4942 and #4945 and fail identically at this branch's base.

### DIFF
--- a/backend/api/crm.yaml
+++ b/backend/api/crm.yaml
@@ -3358,6 +3358,23 @@ paths:
                     tell that from a hit. A well-formed value the account does not
                     currently raise is a `204` that stores nothing; see this operation's
                     description for why those two answers differ.
+                project_id:
+                  type: string
+                  format: uuid
+                  description: |
+                    The project the page was scoped to when the suggestion was rendered,
+                    absent on the whole-account page.
+
+                    A suggestion's fingerprint is derived from its EVIDENCE, and a scoped
+                    page reasons over one project's activity — so the same advice raised
+                    on a scoped page and on the account page are two different
+                    fingerprints. This route re-derives the suggestions to check the
+                    fingerprint is one the account really raises for this caller, and it
+                    can only reproduce a scoped one by narrowing the same way.
+
+                    Send the project the reader was looking at. Omit it and a dismissal
+                    from a scoped page matches nothing and silently stores nothing, which
+                    a reader sees as the card refusing to go away.
       responses:
         '204': { description: "Dismissed, or nothing matched and nothing was stored." }
         '401': { $ref: '#/components/responses/Unauthorized' }
--- a/backend/internal/compose/integration/activity_projectscope_integration_test.go
+++ b/backend/internal/compose/integration/activity_projectscope_integration_test.go
@@ -18,6 +18,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/margince/margince/backend/internal/compose/org360"
+	crmcontracts "github.com/margince/margince/backend/internal/contracts"
 	"github.com/margince/margince/backend/internal/modules/activities"
 	"github.com/margince/margince/backend/internal/modules/people"
 	"github.com/margince/margince/backend/internal/modules/projects"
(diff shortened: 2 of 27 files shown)


dungbanguyen:
**Confirmed and fixed** — `bc419d439`.

You were right, and the mutation check reproduces exactly what you described: reverting the scope in `raisesSuggestion` fails the new test with *"the suggestion the rep dismissed on the ERP page is still on it"*.

`DismissSuggestion` now takes the project the page was scoped to and passes it into `gatherSuggestionInputs`, behind the same `RequireProjectScope` gate the page itself passes — a dismissal narrowed to a project the caller cannot open would otherwise answer 204 for a fingerprint they could never have been served, and the shape of that answer is how you probe for a project's existence. `project_id` is optional and additive on the wire; the SPA sends nothing today because it reads the 360 unscoped, so the scoped page is reachable only through the API.

Worth recording, because it nearly made this a test that proved nothing: my first version of the test used the **stalled-deal** rule and passed against the *unscoped* dismissal. A stalled deal is evidenced by the deal, which the project scope does not narrow, so its fingerprint is identical under both — only rules whose evidence is an activity change fingerprint with the scope. The test uses the **no-reply** rule now, and asserts up front that the scoped and unscoped pages really do raise different fingerprints before it dismisses anything.

`raisesSuggestion`'s comment already promised that "a suggestion the rep could have dismissed" has one definition and that a second spelling would drift silently. Same function was still true; same scope was not. Both are now.

dungbanguyen:
Second round, both threads. `d3a67a689`.

**The nil dereference** — fixed. Both `*wide.Suggestions` and `*stillThere.Suggestions` are guarded now, the way `TestSuggestionDismissalIsPerUser` and the rest of the file do it: a page that withheld the section should fail with a sentence, not panic.

**The web client omitting `project_id`** — worth stating precisely, because it reads like a live break and is not one. The SPA never sends it because it never *scopes the read*: `company360.tsx` fetches `/organizations/{id}/360` with `{ params: { path: { id } } }` and no `project_id` query param, so it only ever renders the whole-account page and only ever raises whole-account fingerprints, which the unscoped dismissal reproduces exactly as before. The scoped page is reachable through the API today, not through the web client.

So there is nothing for the client to propagate yet. When a scoped org page is built, the field is there and the server is ready for it — which is the half of this that was missing before this branch.

Issue #1074, opened 2026-08-12
demoteToRep demotes an arbitrary user when a workspace has more than one


tinnhlt-gra:
demoteToRep demotes an arbitrary user when a workspace has more than one

`demoteToRep` (`internal/compose/integration/quotas_http_integration_test.go`) is the helper
several RBAC integration tests use to turn the logged-in admin into a rep and prove a route
refuses them. It picks its target with:

```sql
SELECT id FROM app_user WHERE workspace_id = $1 ORDER BY created_at LIMIT 1
```

`created_at` is not unique. Users seeded in one statement or one transaction share it, and the
tie-break is whatever the plan returns — so in a workspace with more than one user the helper can
demote somebody other than the caller. The test then authenticates as a still-admin user, gets a
200 where it asserted 403, and fails with a message about the route's RBAC rather than about the
fixture.

Observed once as `TestEmbedReindexConfirmRequiresAdminOrOps` failing with
`rep status -> 200, want 403 permission_denied` in a full parallel lane run; green in isolation,
green on the next full run, green against the same commit on `main`. Same class as #1064.

## Why it is worth fixing rather than re-running

The failure names the wrong subject. A reader sees an authorization test fail and reasonably
concludes the route's RBAC regressed — the most expensive possible false signal on a gate whose
whole job is proving an object grant holds. The cost is not the flake; it is where the flake
points.

The blast radius also grows with the fixture: every test that seeds an extra `app_user` into a
shared workspace widens the window, and the governed-call work (#928) added one such test beside
the existing `roster_integration_test.go`.

## The fix

Demote the user the test actually authenticated as, by id or by email, rather than "the first row
by a non-unique column". If the helper genuinely means "the bootstrap admin", it should say so —
`WHERE email = 'ada@example.com'` is deterministic and self-describing where `ORDER BY created_at
LIMIT 1` is neither.

## The invariant to gate afterwards

A fixture helper that mutates one row identifies that row by something unique.

dungbanguyen:
This is the same defect as #1180, and #4957 fixes it. Linking rather than working it twice.

[...]

PR #4957, opened 2026-09-08
fix(tests): the demotion lands on the seat making the requests

### LINKED ISSUE #1180: TestEmbedReindexConfirmRequiresAdminOrOps let a rep through once (200, want 403)
## What happened

During a full `make test-integration` run, the reindex-confirm permission test failed:

```
embedreindex_integration_test.go:338: rep status -> 200 {Code: Detail:}, want 403 permission_denied (read is admin/ops-only)
```

It did **not** reproduce: it passed on the next full-lane run, passed in isolation, and passed in isolation on an unmodified tree. So this is not a regression from the change that was in flight (#1179, a phase-D column drop on the voice tables, which touches nothing in the reindex path).

## Why it is worth an issue rather than a shrug

The direction of the flake is the wrong one. A permission test that flakes from 403 to **200** means a rep was ALLOWED where the surface is admin/ops-only. Two readings, and they need telling apart:

1. **The fixture raced** — the rep's principal picked up grants from a neighbouring test, or the harness's per-test reset had not landed when the request was made. Then the product is fine and the suite is lying about what it proved.
2. **The gate itself has a race** — a cached permission set, or a check that reads a role assignment written in another transaction. Then a real installation can serve a 200 to somebody who should get a 403, rarely.

Reading 1 is far more likely. Reading 2 is the one that would matter, and nothing observed so far rules it out.

## Where to start

`embedreindex_integration_test.go` around line 329-338, and the permission resolution behind `POST /v1/admin/embeddings/reindex/confirm`. Worth running the package's tests in a loop with `-count=20` first — if it reproduces under repetition, the fixture ordering is visible in the failure; if it never does, the next step is instrumenting what permission set the handler actually saw.

dungbanguyen:
Closes #1180 and #1074, which record the same defect from two angles.

The issue recorded a permission test that let a rep through once — **200 where 403 was expected** — and never reproduced. It set out two readings and said only one would matter.

## Reading 2, the gate has a race: ruled out

`serveAsHuman` calls `svc.Authenticate` on every request, and `Authenticate` ends with `loadGrants` inside that request's own transaction — a fresh read of `role_assignment` each time. There is no cached permission set to go stale, so a committed demotion is visible to the very next call. *(This half was traced by the session working the 2000+ range and written up on the issue; I have not re-walked it.)*

## Reading 1, the fixture: confirmed, and it is not a race

`demoteToRep`'s doc comment says it flips **the bootstrap admin**. Its lookup said:

```sql
SELECT id FROM app_user WHERE is_agent = false ORDER BY created_at LIMIT 1
```

which pins nothing to the session. The two statements agree only while the installation holds exactly one person — and rows inserted in one transaction share `now()`, so a second person makes the choice arbitrary.

An arbitrary choice here fails in the **worst direction available**: it demotes a seat the session is not using, the session keeps its admin role, and the caller's "an admin-only endpoint refuses a rep" assertion comes back 200. That is exactly the shape #1180 observed, and it is not a timing window — it is what the query does whenever the row it picks is not the one the session holds.

## The fix

The helper demotes the seat this session is signed in **as**, read from the app through `GET /v1/me` rather than derived from the table. That is AGENTS.md's review-loop rule 6 applied to a fixture: one that derives production's answer for itself is free to derive a *different* one, and here "different" means demoting a stranger.

The demotion is then **read back** — a helper whose only failure mode is a caller's permission assertion silently inverting must not be able to fail quietly. A caller that gets past that line is signed in as a rep, or the run stops there with a sentence saying so.

## The planted case

The three existing callers (`aicalls`, `aihealth`, `embedreindex`) each hold exactly one person, so none of them could ever see this. So the helper now has its own spec: an installation with a colleague, the session demoted, and then refused an admin-only write.

**Mutation check.** Restoring the old `ORDER BY created_at` lookup:

```
demotetorep_test.go:161: a rep inviting a member → 201, want 403 — the demotion landed on somebody else's seat
--- FAIL: TestDemoteToRepDemotesTheSeatMakingTheRequests
--- PASS: TestEmbedReindexConfirmRequiresAdminOrOps
```

The new case fails with this issue's own 200-instead-of-403 shape, and the reindex test passes beside it — which is the whole reason the defect could only ever appear as a flake.

The colleague's `created_at` is aged deliberately. In production the two rows would usually share `now()` and the order between them would be the planner's business; pinning it is what turns "may pick either" into something a test can assert.

## Checks

`make check-backend` green. The `compose/integration` lane passes apart from the four tests `main` is currently red on (#4945 and #4942 fix those); the three callers and the new case all pass.

### DIFF
--- a/backend/internal/compose/integration/demotetorep_test.go
+++ b/backend/internal/compose/integration/demotetorep_test.go
@@ -8,6 +8,7 @@ package integration
 import (
 	"context"
 	"net/http"
+	"slices"
 	"testing"
 
 	"github.com/margince/margince/backend/internal/compose/integration/apptest"
@@ -102,8 +103,16 @@ func assertSignedInAsRep(t *testing.T, e *apptest.AppEnv) {
 	if status := e.Call(t, "GET", "/v1/me", nil, nil, &me); status != http.StatusOK {
 		t.Fatalf("re-reading the seat after demotion → %d", status)
 	}
-	if len(me.Roles) != 1 || me.Roles[0] != "rep" {
-		t.Fatalf("the session holds %v after demotion, want exactly [rep] — an admin-only endpoint would answer this seat 200 and the caller would read that as the product letting a rep through",
+	// The two things the callers depend on, and not the SIZE of the set. An
+	// exact `[rep]` would turn every caller red the day the product grants
+	// every human a baseline role — a change that leaves this helper's promise
+	// entirely intact, since what a caller needs is that the seat is a rep and
+	// is no longer an admin.
+	if !slices.Contains(me.Roles, "rep") {
+		t.Fatalf("the session holds %v after demotion and no rep role — the demotion did not reach the seat making the requests", me.Roles)
+	}
+	if slices.Contains(me.Roles, "admin") {
+		t.Fatalf("the session still holds admin after demotion (%v) — an admin-only endpoint would answer this seat 200 and the caller would read that as the product letting a rep through",
 			me.Roles)
 	}
 }
--- a/backend/internal/compose/integration/demotetorep_test.go
+++ b/backend/internal/compose/integration/demotetorep_test.go
@@ -7,14 +7,30 @@ package integration
 
 import (
 	"context"
+	"net/http"
 	"testing"
 
 	"github.com/margince/margince/backend/internal/compose/integration/apptest"
 )
 
-// demoteToRep flips the bootstrap admin's role assignment from admin to rep
-// via the owner connection, so a scenario can prove that an admin-only
-// endpoint refuses the seat that may only read.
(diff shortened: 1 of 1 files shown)


dungbanguyen:
Valid — fixed in `cb1015640`.

The exact-set assertion was over-specifying the helper's promise. What every caller depends on is that the seat **is a rep** and **is no longer an admin**; the *size* of the role set is an implementation detail of how the helper writes the assignment, and a baseline role the product grants every human would turn all three callers red while leaving that promise entirely intact.

It asserts the two facts now, separately, because the two failures send a reader to different places: a missing `rep` means the demotion reached nobody, and a surviving `admin` means it reached somebody else — which is this issue's own defect.

PR #4961, opened 2026-09-08
fix(contracts): a deal that names no company still takes a contract

### LINKED ISSUE #2085: A deal with no company crashes offer/contract creation with a 500 (cannot scan NULL organization_id)
## Reproduction

On a seeded dev stack, `POST /v1/contracts` for a deal that has **no company** answers **500**:

```
{"type":"https://errors.gradion.com/internal","title":"Internal Server Error","status":500,"code":"internal"}
```

Server log:

```
level=ERROR msg="unhandled error" method=POST path=/v1/contracts
err="read deal organization: can't scan into dest[0] (col: organization_id): cannot scan NULL into *[16]byte"
```

The two deals `scripts/seed-dev.sh` creates have no `organization_id`, so this is reachable straight out of a fresh `make dev` + `make seed-dev`.

## Cause

`resolveBuyerOrg` in `backend/internal/modules/deals/offer.go:150` scans `deal.organization_id` into `*ids.OrganizationID`:

```go
var dealOrg *ids.OrganizationID
if err := tx.QueryRow(ctx,
    `SELECT organization_id FROM deal WHERE id = $1`, dealID).Scan(&dealOrg); err != nil {
    return nil, fmt.Errorf("read deal organization: %w", err)
}
```

A `*ids.OrganizationID` is a pointer to a `[16]byte` array, which pgx cannot scan NULL into.

**A company-less deal is legitimate, not a bad state.** `deal.organization_id` is declared `uuid NULL` in `backend/migrations/core/0006_deals.up.sql:62`, the create-deal form leaves Company optional, and this very function returns `*ids.OrganizationID` precisely so that "no organization" is representable. Only the scan target is wrong.

## Impact

Any deal without a company refuses offers and contracts with an opaque 500 — no field named, nothing the user can act on. Because marking a deal **won** requires a signed contract on that deal (`win_evidence.go`), this also means **a company-less deal cannot be won at all**, and the error explains none of it. See #2084 for the surfacing half of that.

## Fix direction

Scan into a nullable intermediate (`pgtype.UUID`, or `*uuid.UUID` then convert) and return `nil` when it is not valid — the shape the signature already promises. Worth a test with a company-less deal, since the seed data produces exactly that and no existing test caught it.

Found while writing end-user documentation for deals and verifying the win-evidence path end to end.

dungbanguyen:
Closes #2085.

A deal may be worked before anyone knows whose it is — the create form leaves Company optional and `deal.organization_id` is nullable. Attaching a contract to one answered **500** with nothing a reader could act on. Because winning a deal needs a signed contract on it, that also made a company-less deal impossible to win, and said nothing about why.

## One correction to the issue

The issue attributes the crash to `resolveBuyerOrg` in `deals/offer.go`. That function is fine: it scans into `*ids.OrganizationID`, a pointer, and pgx sets it to nil on NULL. I checked directly, and the whole existing offer suite already proves it — `Env.SeedDeal` creates deals with **no** company, so every offer test in the tree builds offers on exactly the shape the issue says is refused.

The real site is `ensureLinksShareOrganization` in `contracts/contract_write.go`, which declared `var linkedOrg ids.UUID` — a non-pointer `[16]byte`. That is where the reported `*[16]byte` comes from, and its error wrapper renders the same `read deal organization` text the issue quotes from the log, which is what made the two look alike. Reverting the fix reproduces the logged message verbatim:

```
read deal organization: can't scan into dest[0] (col: organization_id): cannot scan NULL into *[16]byte
```

So **offers were never broken; contracts were.**

## Why absent is admitted rather than refused

Only the deal arm can be absent — `project.organization_id` is `NOT NULL`. An absent company is not a company the contract disagrees with. The leak this rule exists against is A's agreement reaching everyone who can see **B's** deal; with no B, there is nobody it reaches that the deal itself does not already admit — `VisibleClause` judges a deal-anchored contract by its deal alone.

## The sweep

Fixing the shape rather than the line: I derived the 128 nullable uuid columns from the schema and cross-referenced every single-column read of one. 34 such reads; six scan into a non-pointer. The other five are all already guarded — three by an explicit `AND <col> IS NOT NULL`, and two by a CHECK constraint (`rel_stakeholder_shape`, `activity_link_shape`) that makes the column non-null for the very kind the query filters on. This was the only unguarded one.

I did **not** turn that sweep into a gate. It needs to model both `IS NOT NULL` guards and per-kind CHECK constraints to tell a real finding from a safe one — five of the six candidates were safe for one of those two reasons — and a gate that cannot would report five false positives on a clean tree.

## Test

`TestAContractAttachesToADealThatNamesNoCompany` asserts both halves, because a test that only checked the admission would pass against a check that had stopped asking: the company-less deal is admitted, and a deal belonging to a **different** company is still refused by field. Mutation-checked — reverting the scan fails it with the reported error.

## CI note

Four integration tests are red on `main` at the tip, unrelated to this branch: `TestPerson360AssemblesEverySectionFromRealRows`, `TestADismissalHoldsUntilTheEvidenceMoves`, `TestDOIIssuanceMintsNothingWhicheverPurposeIsNamed` and `TestAnOmittedRequiredIDIsNamedAndASuppliedOneStaysHidden`. They are covered by #4942 and #4945 and fail identically at this branch's base. `make check-backend` is green.

### DIFF
--- a/backend/internal/compose/integration/contract_lifecycle_integration_test.go
+++ b/backend/internal/compose/integration/contract_lifecycle_integration_test.go
@@ -429,3 +429,71 @@ func TestTwoConcurrentRenewalsLeaveOneSuccessor(t *testing.T) {
 		}
 	}
 }
+
+// A deal may name no company. The create form leaves Company optional and
+// `deal.organization_id` is nullable, so a deal worked before anyone knows
+// whose it is is an ordinary state, not a broken one.
+//
+// It used to end the request in a 500: the cross-company check scanned that
+// column into a non-nullable target, so the read failed before the rule it
+// serves was ever asked. Because winning a deal requires a signed contract on
+// it, that made a company-less deal impossible to win, and said nothing about
+// why.
+//
+// The rule itself still has to hold, so this asserts both halves: the absent
+// company is admitted, and a company that DISAGREES is still refused by field.
+func TestAContractAttachesToADealThatNamesNoCompany(t *testing.T) {
+	e := Setup(t)
+	admin := e.Admin()
+	pipeline, open, _ := DealFixture(t, e)
+
+	unattached, err := e.Deals.CreateDeal(admin, deals.CreateDealInput{
+		Name: "Inbound, company unknown", PipelineID: pipeline, StageID: open,
+	})
+	if err != nil {
+		t.Fatalf("creating a deal with no company: %v", err)
+	}
+	if unattached.OrganizationId != nil {
+		t.Fatalf("the fixture deal names a company (%v), so it cannot prove anything about one that does not",
+			unattached.OrganizationId)
+	}
+	dealID := ids.From[ids.DealKind](ids.UUID(unattached.Id))
+	org := e.SeedOrg(t, "Acme", nil)
+
+	created, err := e.Contracts.CreateContract(admin, contracts.CreateContractInput{
+		OrganizationID: ids.From[ids.OrganizationKind](org),
+		Title:          "MSA 2026", ValueBasis: contracts.BasisTotal, Source: "manual",
+		DealID: &dealID,
+	})
+	if err != nil {
+		t.Fatalf("a contract on a company-less deal was refused: %v", err)
+	}
+	if created.DealId == nil || ids.UUID(*created.DealId) != ids.UUID(unattached.Id) {
+		t.Errorf("the contract came back naming deal %v, want %v", created.DealId, unattached.Id)
+	}
+
+	// The other way round, so the admission above is the absent company and not
+	// the check having stopped asking: a deal belonging to somebody ELSE is
+	// still refused, and named by field.
+	elsewhere := e.SeedOrg(t, "Contoso", nil)
+	elsewhereID := ids.From[ids.OrganizationKind](elsewhere)
+	otherDeal, err := e.Deals.CreateDeal(admin, deals.CreateDealInput{
+		Name: "Contoso expansion", PipelineID: pipeline, StageID: open, OrganizationID: &elsewhereID,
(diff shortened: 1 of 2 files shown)


dungbanguyen:
**Real, and pre-existing** — filed as #4986 rather than fixed here, with the reasoning below.

The state you describe is reachable **today, without this PR**: create a deal for company A, add a contract, then reassign the deal to B. `applyDealLinkPatches` in `deals/deal.go` writes `organization_id` behind `auth.EnsureLinkTarget` on the new company alone and never asks what agreements hang off the deal. `ensureLinksShareOrganization` refuses to *create* that pairing; nothing stops it being *reached* afterwards.

You are right that this PR adds a **second** route in — company-less deal, contract, then assign — because attaching a contract to a company-less deal was a 500 before. So the exposure widens even though the hole is not new.

Two reasons it is a separate change rather than a tail on this one:

1. **It lives in another module.** The check is a read of `contract` from inside `deals`, which cannot import it. It wants a compose seam — `EnsureProjectAttachable` and `EnsurePartner` are two that already have exactly this shape and are injected into the very same function, so the pattern is there.
2. **It needs a product decision.** Refusing the reassignment while live contracts name another company is safe, but it means a rep who mis-filed a deal must detach the paperwork before they can correct the company — on a record where the paperwork may be what they are trying to fix. Reconciling (moving the contracts with the deal) silently rewrites which company an agreement is with. Detaching loses a link a human made. That is not mine to pick inside a crash fix.

#4986 carries all three options, both routes, and `status: needs-decision`. Marked `priority: high`, since it is a cross-tenant read of a commercial document rather than a display fault.

PR #4971, opened 2026-09-08
fix(briefs): a brief names the ranking factor it could not read

### LINKED ISSUE #2038: briefs' warmth factor floors on a refused edge read instead of absenting the section
Found while landing the withheld-coverage channel (#1963, PR #2037). Pre-existing on `main`; the file is untouched by that PR.

## The shape

`internal/compose/briefs/briefrank.go` `seatEvidenceBound` resolves the seat edge's admission and, on `ErrPermissionDenied`, returns `mayReadSeats == false` and continues:

```go
clause, err = auth.EdgeReadScope(ctx, "r", ...)
if errors.Is(err, apperrors.ErrPermissionDenied) {
    return nil, "", false, nil
}
```

The brief then ranks with **no seat evidence**, so `resolveWarmth`'s factor floors. A caller refused the edge grant does not get a brief that says a factor was withheld — they get a brief whose deals are ranked **lower than they are**, with nothing marking why.

## Why it is the same defect #1963 fixed

#1963's whole argument is that a number computed from a withheld input is worse than a missing one: an empty risk list rendered as "this deal passes every coverage check", and the fix was a channel for saying the section was withheld. The brief has the identical shape one surface over — a ranking is a verdict, and a deal that ranks fifth because the reader could not see its stakeholders is a wrong answer, not an incomplete one.

Review-loop rule 1 says fix the invariant, not the call site. This is the sibling copy that PR did not reach, and it is filed rather than folded in because the brief payload has **no withheld channel** — closing it needs the same contract step the coverage payload just took, which is its own change.

## What is needed

1. A withheld channel on the brief payload (the `sections_omitted` shape the 360 payloads and now `DealCoverage` use), or a rule that a rank computed from a withheld input is absent rather than lower.
2. `seatEvidenceBound`'s swallow then names the omission instead of returning a bare `false`.
3. `backend/edgereaders_test.go` already counts this read as gated (it takes `EdgeReadScope`), so the census will not fail while this is open — the gap is in what happens AFTER the refusal, which no gate currently judges. Worth stating in the entry.

## Severity

`priority: normal`, not high: unlike the coverage surfaces, the brief's ranking is advisory and carries no explicit "clear" verdict a reader would act on the way they would act on "this deal passes every coverage check". It is still a wrong number on screen.

dungbanguyen:
Closes #2038.

Every seat on a deal is a `deal_stakeholder` edge, so a caller without the `relationship` grant read no stakeholders for **any** deal and the warmth factor floored across the whole queue. The brief then handed back an order that is not the order — deals sitting lower than they are — with nothing saying why.

This is the sibling copy of the defect the withheld-coverage channel fixed one surface over, and it needed the same contract step the issue predicted: the brief payload had nowhere to say a factor was withheld.

## What changed

- **`factors_omitted`** on `MorningBrief` and on the `read_brief` tool result. Never absent, empty on the ordinary read: empty and withheld are different answers, and a client that had to tell them apart by guessing would render every brief as withheld or none of them.
- **Stored with the run**, for the reason `revenue_norm_currency` is: the grant can be given or taken away afterwards, and a queue ranked without warmth must not later be read as one that had it. New nullable-free column with a CHECK holding the vocabulary closed — a factor nothing can render is a factor a client silently drops, which is the failure this exists to end.
- **`seatEvidenceBound` reports its refusal** instead of returning a bare `false`, and `briefEvidenceRows` carries it up to the run.

## Two decisions worth reviewing

**The feature vector is untouched.** A named factor still appears in each item's `feature_vector`, at its floor. The decomposition is what the run *scored with*, and editing it there would leave a reader unable to check the `composite` against its parts. This array is what says the floor is an absence rather than a reading.

**Row-scope flooring stays as it was, and is right.** A stakeholder outside the caller's row scope contributes nothing, deal by deal — the queue is still ranked on what this reader may know, the way every other factor is. Only the *outright* object refusal is named, because that one has no input for any deal and reorders the whole queue. The old comment claimed both cases were the same thing; it now says which is which.

## Served, not withheld, on the tool seam

`TestEveryPersistedBriefFieldIsServedOrNamedAsWithheld` forced the question. It is served: unlike the revenue base beside it — which explains a number the agent already has normalized — this one says the **order** in front of the agent is not the order. An agent told a queue is ranked will reason about why the third deal is third; if warmth was withheld, the honest answer is that nobody knows.

## On the issue's point 3

There is no per-file entry to annotate — `edgereaders_test.go` derives its corpus rather than listing `briefrank.go`. The post-refusal behaviour the issue says no gate judges is now held by the two tests below instead, and the code comments say which case is which rather than leaving the reader to infer it.

## Tests

Both mutation-checked (making `omittedFactors` always return empty fails the first):

- `TestABriefNamesTheFactorItCouldNotRead` — a caller holding every grant the brief needs **except** the edge gets `warmth` named.
- `TestABriefThatReadEveryFactorOmitsNone` — and one holding it gets an empty list, never nil. Without this half, the first would pass against a run that names warmth unconditionally.

## What I left out

**The SPA does not render it yet.** The brief screen is ~60 files and `frontend/AGENTS.md` opens with a design-system catalog that has to be read before anything visible is built, so wiring the caption is its own change in another module rather than a tail on this one. I fixed the two fixtures the new required field broke (`make check-fe` is green) and filed the rendering as a follow-up. The channel is live for the contract and for the MCP tool today.

### DIFF
--- a/backend/api/crm.yaml
+++ b/backend/api/crm.yaml
@@ -40565,7 +40565,7 @@ components:
         `brief_item`): the ranked, honest-short queue of winnable deals plus the metadata that
         reproduces it. `candidate_count` is how many deals cleared the §10 honest-short bar;
         `items` is the top slice (≤ 7) — genuinely shorter when fewer qualify, never padded.
-      required: [id, generated_at, as_of, candidate_count, items]
+      required: [id, generated_at, as_of, candidate_count, items, factors_omitted]
       properties:
         id: { type: string, format: uuid }
         generated_at: { type: string, format: date-time, description: When this run was assembled. }
@@ -40603,6 +40603,32 @@ components:
             When the overnight agent last wrote its findings, null when no pass has. A reader
             showing a run with no narrative must consult this before saying "a quiet night":
             without it, a model that never ran and a night with nothing in it look identical.
+        factors_omitted:
+          type: array
+          description: |
+            The ranking factors this run could not read, so a client can say "the order does
+            not account for this" instead of presenting a queue as fully ranked. Never
+            returned absent, and empty on the ordinary read: empty and withheld are different
+            answers, and a factor that floors silently makes a deal rank lower than it is with
+            nothing marking why.
+
+            `warmth` is omitted for a caller with no `relationship` edge grant. Every seat on
+            a deal is a `deal_stakeholder` edge, so such a caller runs no stakeholder read at
+            all and the factor has no input for ANY deal — which is why this is a property of
+            the run and not of an item.
+
+            A named factor still appears in each item's `feature_vector`, at its floor. The
+            decomposition is what the run scored with, and editing it here would leave a
+            reader unable to check the `composite` against its parts; this array is what says
+            the floor is an absence rather than a reading.
+
+            Stored with the run. The grant can be given or taken away afterwards, and a queue
+            ranked without warmth must not later be read as one that had it. Empty on a run
+            assembled before this field existed — a rep has one run per local day, so those
+            age out within a day.
+          items:
+            type: string
+            enum: [warmth]
 
     MorningBriefItem:
       type: object
--- a/backend/internal/compose/briefs/briefday.go
+++ b/backend/internal/compose/briefs/briefday.go
@@ -62,11 +62,11 @@ func insertRunIfDayFree(ctx context.Context, tx pgx.Tx, run BriefRun) (bool, err
 	tag, err := tx.Exec(ctx, `
 		INSERT INTO brief_run (
 			id, user_id, generated_at, as_of, local_day, candidate_count,
-			revenue_norm_minor, revenue_norm_currency)
-		VALUES ($1, $2, $3, $4, $5, $6, $7, $8)
+			revenue_norm_minor, revenue_norm_currency, factors_omitted)
+		VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9)
 		ON CONFLICT ON CONSTRAINT uq_brief_run_user_day DO NOTHING`,
 		run.ID, run.UserID, run.GeneratedAt, run.AsOf, run.LocalDay, run.CandidateCount,
-		run.RevenueNormMinor, run.RevenueNormCurrency)
+		run.RevenueNormMinor, run.RevenueNormCurrency, run.FactorsOmitted)
 	if err != nil {
 		return false, err
 	}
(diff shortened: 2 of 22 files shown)


dungbanguyen:
**All four confirmed and fixed** — `b7cb39337`. The P1 was the good one: it is the same defect this PR exists to fix, one layer down, and I had walked past it.

**P1 — `relationship.read` without `person.read`.** Warmth needs *two* grants, not one: the seat edge to learn who is on the deal, and the person grant to score them. A caller holding the first and not the second reads every stakeholder, gets refused on all of them, and ends up with exactly the silently cold queue this change was written to end.

What made it invisible in the code is that `resolveWarmth` caught `ErrNotFound` and `ErrPermissionDenied` in one `case`. Those are different facts and the tree's own contract says so — a row-scope miss answers `ErrNotFound` so existence stays hidden, an object denial answers `ErrPermissionDenied`. Split now: a stakeholder outside the caller's rows still floors *that deal* alone, which is right, and an object denial reports the factor as unreadable for the run.

Mutation-checked, and the test needed a real fixture change to mean anything: the brief fixture seeds **no** `deal_stakeholder` rows, so my first version passed on an empty map. It seeds a seat now.

**P2 — the `never null` claim.** You were exactly right that it held by accident. `Handle` normalizes `FactorsOmitted` beside `Items` now, so the doc's stated reason is the actual reason rather than a property of the single production reader.

**P2 — the validating CHECK.** Split into `1788851401`, following `1787968163` on this same table, whose own comment spells out why: the runner wraps each file in one transaction, so validating in the same one puts the scan under the `ACCESS EXCLUSIVE` that `ALTER` holds until commit, and the lighter lock `VALIDATE` asks for buys nothing.

**P3 — empty vs nil in the test.** Asserted separately now. You are right that `len() != 0` cannot tell them apart and that the distinction is the whole point.

dungbanguyen:
Second round — you were right again, and this one is sharper than my first fix. `0c06c9862`.

Inferring the answer from the first refusal only works if there is a read to refuse. A caller with no person grant whose candidate deals happen to carry **no stakeholders** never reaches `PersonStrength` at all, so the run reported nothing withheld for a queue whose warmth factor could not have been read even had there been someone to score.

The grant is a property of the **caller**; the seats are a property of the **deals**. Only the first decides whether the factor is readable, so it is asked up front now, the way `seatEvidenceBound` already asks for the edge.

The test is the tell: it seeds **no** stakeholder now. My previous version had to seed one — not because the case needs a seat, but to give the inference something to observe, which is exactly the gap you named.

PR #4984, opened 2026-09-08
fix(ai): a binding's base_url reaches a guarded dialer, on every profile

ancorn-gradion:
## What

A stored AI routing binding's `base_url` now reaches a guarded dialer, and is
checked at the write, on **every** profile — not only under `profile: sovereign`.
`ollama`, `vllm` and `openai_compatible` may dial loopback, a private range or a
public host; `anthropic`, `openai` and `gemini` may dial a public host only.
Neither lane may dial the ranges that serve nobody — link-local (cloud instance
metadata), CGNAT, the documentation ranges, and the encapsulations that carry
another address inside them.

## Why

`ValidateTierBinding` gated `base_url` on host only under the sovereign profile,
and the contract schema declares a bare `type: string`, so any value that parsed
was persisted. It then reached an unguarded client: `newOutboundClient` set no
`DialContext` and no `Dialer.Control`, and no file under `internal/modules/ai`
imported `netguard` at all.

Two requests were enough — one `PUT /v1/ai/routing` to plant the host, one
`GET /v1/ai/available-models` to fire it and read the decoded answer back:

- **A reachability oracle for the deployment's own network.** `ollama` and `vllm`
  need no BYOK key, so client construction could not fail closed, and the read
  reflects the vendor's decoded body — cloud instance metadata included.
- **Vendor-credential exfiltration.** A cloud tier repointed at another host sends
  the installation's `x-api-key` / `x-goog-api-key` there. Go strips
  `Authorization` across a host change and has never heard of those, which is the
  reason `modellist.go` already refuses redirects on the list path.

Firing the stored host needs only `ai_routing:read` while setting it needs
`update`. The tree conceded both halves in its own comments
(`availablemodels.go`, `sovereignendpoint.go`): the *request-supplied* URL was
guarded for exactly this reason, and the *stored* one was left open.

**The invariant:** an address this server dials on an operator's say-so is one
that lane may reach — decided in one place, enforced both at the write and on the
socket.

`netguard.RefusePrivate` alone — what every other tenant-host lane here uses —
could not be it, because a private address is the *feature* on the local lane (a
local Ollama, a GPU box on the operator's network, a self-hosted gateway). So the
allowance is derived from the binding, in one new file, `outboundegress.go`:

| lane | providers | may dial |
|---|---|---|
| operator endpoint | `ollama`, `vllm`, `openai_compatible` | loopback, private ranges, public hosts |
| vendor | `anthropic`, `openai`, `gemini` | public hosts only |

`openai_compatible` is the one BYOK provider on the permissive lane, because "a
self-hosted gateway" is its documented purpose; it is named as the sole exception
in the census test, so the next adapter cannot inherit it quietly. The strict
lane is the zero value, so a provider added without a considered entry fails
closed.

One predicate — `addressAllowed` — is read by the `net.Dialer.Control` hook
(post-DNS, on the concrete address, so a name that resolves *or rebinds* to a
refused address is stopped at connect time) and by the write-time rule (so an
operator hears it at the door rather than as an unreachable vendor with the
reason in a log nobody is reading). A `base_url` carrying userinfo is refused
outright, and neither refusal echoes the value.

### Behaviour changes

- **`profile: sovereign` no longer accepts a link-local endpoint.** It is the
  metadata address, nothing serves inference from an autoconfiguration address,
  and `classifyHost` now shares its "is this ours?" predicate with the egress
  lanes rather than keeping a second, wider answer. Both docs updated.
- **A native vendor binding pointed inside your own network is now a startup
  error** naming `openai_compatible` as the fix. Two troubleshooting rows added.
- **Adapter tests bind through `selectBrainOn`**, the transport seam `webread` and
  `webhooks` already use — an `httptest` server listens on 127.0.0.1, which the
  vendor lane refuses by design. `SelectBrain` stays the only guarded constructor
  production has, and `TestSelectBrainWiresTheEgressGuard` holds that end to end:
  refusal *and* positive control.

### Residue

An actor holding `ai_routing.update` can still point a lane at an arbitrary
**public** host — that is the documented `base_url` override — and can use the
local lane to reach private addresses, which is the local-model feature and is
indistinguishable from a port sweep. Both are questions about who holds that
grant rather than about this code. Narrowing `available-models` to `update`
authority is a separate product call and is not made here.

## How verified

New tests, all offline:

- `TestEveryProviderDeclaresAnEgressClass` — census, both directions, plus the
  assertion that the zero class is the strict one.
- `TestEveryLocalProviderTakesTheOperatorLane` — derived from `ProviderIsLocal`,
  with `openai_compatible` named as the only exception.
- `TestWhichAddressesEachLaneMayDial` — the rule as addresses, including the
  NAT64 and 6to4 spellings of an internal address.
- `TestTheDialGuardRefusesWhatItCannotJudge`,
  `TestSelectBrainWiresTheEgressGuard`, `TestABaseURLIsCheckedOnEveryProfile`,
  and its mirror `TestTheEndpointRuleLetsTheOrdinaryDeploymentsBoot`.

Ran locally on a Windows checkout: `go build ./...`, `go test ./...` across the
whole backend, `craft static --strict` over every changed Go file (0 findings),
both golangci configs on `internal/modules/ai/...` (0 issues under
`.golangci.strict.yml`), `check-rls-store-path.sh` and `check-no-jurisdiction.sh`.
The full-suite failure set was captured before and after the change and is
**byte-identical** — the pre-existing failures on that checkout are line-ending
and toolchain artifacts that hit untouched files the same way. `gitleaks` does
not run on MinGW, so the secret gate is CI's to confirm; fixture values are
`stands-in-for-a-key` / `stands-in-for-a-token`, deliberately not credential-shaped.

- [ ] `make check` is green — **not run whole on this checkout** (see above); its
      Go halves were run directly and are green, CI is the authority
- [x] `make test-integration` is green (or: this change does not touch tenant data / RLS / the write shape) — it touches none of them
- [x] `make craft-static` reports no new blockers — diff-scoped `craft static --strict`: 0 blocker, 0 major, 0 minor

## AI involvement

Written by Claude Code (Opus 5) end to end: the finding analysis, the two-lane
design, the code, the tests and the docs. The one design judgement worth flagging
for a human: the choice to narrow `sovereign` away from link-local rather than
carve out the metadata address by name, and the choice to leave
`openai_compatible` on the permissive lane. Both are argued above and both are
held by tests.

## Accountability

https://claude.ai/code/session_01SZJRR9vFhC3KYu2BuJFXsg

---
## Summary by cubic
Closes an SSRF gap where a stored AI routing `base_url` was persisted and dialed without egress checks on every profile except `sovereign`. Each provider now declares an egress class, and the same predicate enforces the rule at write time and again post-DNS in the `net.Dialer.Control` hook. The guard also survives a proxy, a zone, a username in userinfo, and a redirect that would move a model key off the binding's host.

**Bug Fixes**
- `ollama`, `vllm`, and `openai_compatible` may dial loopback, private ranges, or public hosts; `anthropic`, `openai`, and `gemini` public hosts over https only. Link-local, CGNAT, documentation, and encapsulation ranges are refused on both lanes, and a `base_url` carrying userinfo is refused outright.
- `profile: sovereign` no longer accepts link-local endpoints, and a native vendor binding pointed inside the deployment's network is now a startup error suggesting `openai_compatible`.
- The outbound transport sets `Proxy = nil`; a proxy would otherwise route every dial around the guard.
- The outbound client refuses a cross-host redirect, an https-to-http downgrade, or a redirect to another port on the same host — each would carry the `x-api-key` / `x-goog-api-key` header elsewhere; same-host redirects that keep scheme and port still work.
- Refusals name only a `base_url`'s scheme and host, so a credential pasted into userinfo, an opaque url, a path, or a query never reaches an error or boot log.
- Tests hold the invariants the guard relies on: the write-time rule and the dialer can't drift apart, and `SelectBrain` is the only builder of an outbound client.

<sup>Written for commit 5e4bed51949e0346b82461e290032dc92b35acf4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/4984?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added outbound network safeguards for AI provider connections.
  - Cloud providers require public endpoints; self-hosted providers may use approved private or loopback addresses.
  - Endpoint validation occurs during configuration and connection, including DNS resolution.
  - Unsafe destinations—including metadata, link-local, carrier-grade NAT, and documentation ranges—are blocked.
  - Embedded credentials in URLs are rejected.
  - Provider connections bypass proxies and block unsafe cross-host or HTTPS-to-HTTP redirects.

- **Documentation**
  - Updated configuration and cloud-provider guidance with endpoint rules and troubleshooting information.

### DIFF
--- a/backend/internal/modules/ai/anthropic_test.go
+++ b/backend/internal/modules/ai/anthropic_test.go
@@ -25,7 +25,7 @@ func newAnthropicForTest(t *testing.T, handler http.HandlerFunc) model.Client {
 	t.Helper()
 	srv := httptest.NewServer(handler)
 	t.Cleanup(srv.Close)
-	client, err := SelectBrain(ProviderConfig{Provider: "anthropic", Model: "claude-test", BaseURL: srv.URL}, cloudKeyFor("anthropic", testAnthropicKey))
+	client, err := selectLocalBrain(ProviderConfig{Provider: "anthropic", Model: "claude-test", BaseURL: srv.URL}, cloudKeyFor("anthropic", testAnthropicKey))
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/backend/internal/modules/ai/attachmentplacement_test.go
+++ b/backend/internal/modules/ai/attachmentplacement_test.go
@@ -318,7 +318,7 @@ func TestEveryAdapterPlacesAnAttachmentOnTheSameTurn(t *testing.T) {
 			defer srv.Close()
 
 			for name, read := range placementReaders {
-				client, err := SelectBrain(placementBinding(name, srv.URL), allCloudKeys())
+				client, err := selectLocalBrain(placementBinding(name, srv.URL), allCloudKeys())
 				if err != nil {
 					t.Fatalf("%s: %v", name, err)
 				}
--- a/backend/internal/modules/ai/attachments_test.go
+++ b/backend/internal/modules/ai/attachments_test.go
@@ -57,7 +57,7 @@ func TestEveryProviderMapsOrRejectsAttachmentsNeverSilentlyDrops(t *testing.T) {
 	for name, tc := range mustRefuse {
 		for _, mime := range tc.mimes {
 			t.Run(name+"/"+mime, func(t *testing.T) {
-				client, err := SelectBrain(tc.cfg, allCloudKeys())
+				client, err := selectLocalBrain(tc.cfg, allCloudKeys())
 				if err != nil {
 					t.Fatal(err)
 				}
@@ -123,7 +123,7 @@ func TestAnthropicAndOllamaCarryAttachmentsInTheirOwnWireSpelling(t *testing.T) 
 		},
 	} {
 		t.Run(name, func(t *testing.T) {
-			client, err := SelectBrain(tc.cfg, allCloudKeys())
+			client, err := selectLocalBrain(tc.cfg, allCloudKeys())
 			if err != nil {
 				t.Fatal(err)
 			}
@@ -156,7 +156,7 @@ func TestAnthropicAndOllamaAdvertiseWhatTheyCarry(t *testing.T) {
 		"ollama":    {cfg: ProviderConfig{Provider: "ollama", Model: "m"}, want: carriesImages},
 	} {
 		t.Run(name, func(t *testing.T) {
-			client, err := SelectBrain(tc.cfg, allCloudKeys())
+			client, err := selectLocalBrain(tc.cfg, allCloudKeys())
 			if err != nil {
 				t.Fatal(err)
 			}
@@ -193,7 +193,7 @@ func TestDeclaredImageCarriageAcceptsImagesAndStillRejectsPDFs(t *testing.T) {
 	}
 	for name, cfg := range declaresImages {
 		t.Run(name, func(t *testing.T) {
-			client, err := SelectBrain(cfg, allCloudKeys())
+			client, err := selectLocalBrain(cfg, allCloudKeys())
 			if err != nil {
 				t.Fatal(err)
 			}
@@ -276,7 +276,7 @@ func TestAttachmentBytesXorURIEnforced(t *testing.T) {
 		writeFixture(t, w, `{"id":"r","status":"completed","output":[{"type":"message","content":[{"type":"output_text","text":"ok"}]}]}`)
 	}))
 	defer srv.Close()
-	client, err := SelectBrain(ProviderConfig{Provider: "openai", BaseURL: srv.URL, Model: "m"}, allCloudKeys())
+	client, err := selectLocalBrain(ProviderConfig{Provider: "openai", BaseURL: srv.URL, Model: "m"}, allCloudKeys())
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -314,7 +314,7 @@ func TestNativeCloudProvidersCarryPDFAttachments(t *testing.T) {
 	}
 	for name, cfg := range canCarryPDF {
 		t.Run(name, func(t *testing.T) {
-			client, err := SelectBrain(cfg, allCloudKeys())
+			client, err := selectLocalBrain(cfg, allCloudKeys())
 			if err != nil {
 				t.Fatal(err)
 			}
--- a/backend/internal/modules/ai/availablemodels.go
+++ b/backend/internal/modules/ai/availablemodels.go
@@ -99,9 +99,10 @@ type AvailableModels struct {
 // The provider and the LANE are named; the endpoint is not. Both are closed
 // vocabularies — the adapter names this build accepts, and the tiers the stored
 // document already binds — and the host is read from that document or from the
-// adapter's compiled default, never from the request. The AI outbound client
-// carries no egress allowlist, so accepting a URL here would let anyone holding
-// this grant point the server at an arbitrary host.
+// adapter's compiled default, never from the request. A URL accepted here would
+// be a destination chosen by a caller holding only READ on this setting, while
+// the stored one had to be written by someone holding update and passed the
+// endpoint rule on the way in (outboundegress.go).
 //
 // The lane matters because one vendor may be bound at two hosts: a broker on one
 // tier and a self-hosted gateway on another is a configuration the routing
--- a/backend/internal/modules/ai/gemini_test.go
+++ b/backend/internal/modules/ai/gemini_test.go
@@ -28,7 +28,7 @@ func newGeminiForTest(t *testing.T, handler http.HandlerFunc) model.Client {
 	t.Helper()
 	srv := httptest.NewServer(handler)
 	t.Cleanup(srv.Close)
-	client, err := SelectBrain(ProviderConfig{Provider: providerGemini, BaseURL: srv.URL, Model: "gemini-x"}, cloudKeyFor("gemini", testGeminiKey))
+	client, err := selectLocalBrain(ProviderConfig{Provider: providerGemini, BaseURL: srv.URL, Model: "gemini-x"}, cloudKeyFor("gemini", testGeminiKey))
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/backend/internal/modules/ai/modellist_test.go
+++ b/backend/internal/modules/ai/modellist_test.go
@@ -46,7 +46,7 @@ func listerFor(
 	}))
 	t.Cleanup(srv.Close)
 	cfg.BaseURL = srv.URL
-	client, err := SelectBrain(cfg, cloudKeyFor(cfg.Provider, "test-key"))
+	client, err := selectLocalBrain(cfg, cloudKeyFor(cfg.Provider, "test-key"))
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -102,7 +102,7 @@ func TestAnthropicSendsTheVersionHeaderTheAPIRequires(t *testing.T) {
 		}
 	}))
 	t.Cleanup(srv.Close)
-	client, err := SelectBrain(
+	client, err := selectLocalBrain(
 		ProviderConfig{Provider: providerAnthropic, BaseURL: srv.URL},
 		cloudKeyFor(providerAnthropic, "sk-test"),
 	)
@@ -231,7 +231,7 @@ func TestGeminiFollowsItsPages(t *testing.T) {
 		}
 	}))
 	t.Cleanup(srv.Close)
-	client, err := SelectBrain(
+	client, err := selectLocalBrain(
 		ProviderConfig{Provider: providerGemini, BaseURL: srv.URL},
 		cloudKeyFor(providerGemini, "k"),
 	)
@@ -280,7 +280,7 @@ func TestListModelsReportsTheStatusAndNotTheBody(t *testing.T) {
 		}
 	}))
 	t.Cleanup(srv.Close)
-	client, err := SelectBrain(
+	client, err := selectLocalBrain(
 		ProviderConfig{Provider: providerOpenAI, BaseURL: srv.URL},
 		cloudKeyFor(providerOpenAI, "k"),
 	)
@@ -463,7 +463,7 @@ func TestAListRedirectCarriesNoCredentialToTheNewHost(t *testing.T) {
 	}))
 	t.Cleanup(vendor.Close)
 
-	client, err := SelectBrain(
+	client, err := selectLocalBrain(
 		ProviderConfig{Provider: providerAnthropic, BaseURL: vendor.URL},
 		cloudKeyFor(providerAnthropic, "sk-must-not-travel"),
 	)
--- a/backend/internal/modules/ai/narrowing_test.go
+++ b/backend/internal/modules/ai/narrowing_test.go
@@ -89,7 +89,7 @@ func TestANarrowedBindingRefusesWhatItNoLongerAdvertises(t *testing.T) {
 	}))
 	defer srv.Close()
 
-	narrowed, err := SelectBrain(ProviderConfig{
+	narrowed, err := selectLocalBrain(ProviderConfig{
 		Provider: providerGemini, BaseURL: srv.URL, Model: "m", Input: []string{"text", "image"},
 	}, allCloudKeys())
 	if err != nil {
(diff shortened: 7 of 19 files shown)


ancorn-gradion:
## Review findings — what was taken and what was not

Thanks, both bots earned their keep here. Five findings were valid and are fixed in `5bbe77a`; each one is now held by a test that was checked from the other end (reintroduce the defect, watch the test fail).

### Taken

**`HTTP(S)_PROXY` bypassed the guard entirely** (cubic P1, coderabbit Major). Correct, and the most serious of the five. `DefaultTransport.Clone()` inherits `ProxyFromEnvironment`, so on a deployment carrying `HTTPS_PROXY` the socket opens to the proxy — public, admitted — and the proxy is then asked to `CONNECT` wherever the binding pointed. One environment variable was the whole distance between this guard and no guard.

Worth noting for the record that this repo already knew: `backend/pkg/extension/outbound.go` and the hubspot overlay's gated transport both clear `Proxy`, each with a comment saying why. Mine was the third writer of that rule and the one that forgot it. `transport.Proxy = nil`, held by `TestTheOutboundClientCannotBeRoutedThroughAProxy` over every provider.

**A zoned address was read two different ways** (cubic P2 + P3, coderabbit Minor). Correct. `dialGuard` used `net.ParseIP` where the write-time rule used `parseHostAddress`, so `fd00::1%eth0` passed the write and failed the first connection — the "single spelling" claim being false in the one direction its own test could not see, because the table had no zoned case. Both fixed together: the guard shares the parser, and the agreement table carries zoned cases spelled `%25` in the url and bare on the dial address, as each end really receives them. Adding the case first reproduced the mismatch, which is the order I'd rather find these in.

**A username in userinfo reached the boot log** (cubic P2). Correct and the subtlest. `url.Redacted()` replaces the *password* and keeps the username, and a model key pasted into a `base_url` arrives as `http://sk-live-…@host` at least as often as with a colon after it. Fixed at the invariant rather than the call site: a shared `withoutUserinfo` now serves all four exits that name a `base_url`, and `TestARefusalNamesNoPartOfTheUserinfo` plants a secret in each position. Restoring `Redacted()` fails it with the real leak visible in the message.

**Whitespace-only `base_url`** (cubic P2). Correct — `defaulted` falls back only on the empty string, so a blank value reached the adapter as its host. Now refused as hostless at the write.

**The AST scan only walked function bodies** (cubic P3). Correct, and the failure mode AGENTS.md calls out by name: a package-level `var x = newOutboundClient(...)` is a `GenDecl`, so the census would have reported clean while an unguarded client existed. It now inspects the whole file and reports package scope by name.

**Unknown provider ordering** (cubic P2). Taken, with a different reason than proposed: the rule now returns early for a provider absent from `providerEgress`, not to preserve a contract but because there is no lane to judge — `SelectBrain` refuses that binding outright, so it reaches no socket, and an egress error would name the wrong problem. `TestEveryProviderDeclaresAnEgressClass` keeps this from ever skipping a provider that *can* be served.

### Not taken

**Require https on the vendor lane** (coderabbit Major). A real weakness, but a different one: cleartext to a vendor is a confidentiality-in-transit defect, and it does not touch the exfiltration this PR fixes — an attacker who sets `base_url` gets the key whichever scheme it carries. It is also a new refusal on a documented field, so it deserves its own change and its own note in the connect-a-cloud-model-provider guide rather than riding in on an SSRF patch. Happy to open it as a follow-up if a reviewer wants it here instead.

**Refuse redirects on the completion path** (coderabbit Major). This tree already made the call, deliberately, and I'd rather not reverse it in a security patch without a product reason. `modellist.go`'s `noRedirect` refuses redirects on the *list* path and its comment says why it copies the client instead of changing `newOutboundClient`: that one also carries streaming completions, "where a redirect a vendor genuinely uses would become an outage — a wider change than a review of this surface should make." The dialer guard does re-enter on every hop, so a redirect into the deployment's own network is refused; a redirect to another public host carrying `x-api-key` is the residue, and it is the same residue already listed in the PR body for the base_url override.

Both bots also flagged nothing about the two behaviour changes in the description — link-local under `sovereign`, and the adapter tests moving to the `selectBrainOn` seam — which are the two places I'd most like a human eye.

ancorn-gradion:
### Re: [the `withoutUserinfo` / `Opaque` finding](https://github.com/margince/margince/pull/4984#discussion_r3956734523) — valid, and it went wider

Good catch, and it was right in a way that made me stop and check the rest of the field: `url.Parse("http:sk-live-token@")` really does file the whole payload under `Opaque` with no host at all, so clearing `User` left the secret in the no-host error.

I did not take the suggested `shown.Opaque = ""` though, because probing the parser showed `Opaque` is one of **four** ways a credential reaches these messages, not one:

| how it was written | where it lands | `Redacted()` | clearing `User` | clearing `User`+`Opaque` |
|---|---|---|---|---|
| `http://user:TOKEN@` | `User` | leaks username | safe | safe |
| `http:TOKEN@` | `Opaque` | leaks | leaks | safe |
| `http:///v1/TOKEN` | `Path` | leaks | leaks | **leaks** |
| `ftp://h/v1?api_key=TOKEN` | `RawQuery` | leaks | leaks | **leaks** |
| `ftp://h/v1#TOKEN` | `Fragment` | leaks | leaks | **leaks** |

The last three reach the `names no host` and `must be an http(s) url` exits, which the userinfo-shaped rule never touched. Each blanking version of this loses to the next shape nobody thought of — it had already lost twice before your comment, and `Opaque` would have been the third.

So `36233739` inverts it into an allowlist: scheme and host are rebuilt into a fresh `url.URL` and everything else is dropped by construction, not by a list somebody has to keep complete.

```go
func safeToName(parsed *url.URL) string {
	shown := url.URL{Scheme: parsed.Scheme, Host: parsed.Host}
	return shown.String()
}
```

`TestARefusalNamesNothingButTheSchemeAndHost` now plants a token in all five spellings above, each entering the parser by a different exit. I ran it against the narrower fix too — clearing `User` and `Opaque` alone still prints the credential:

```
--- FAIL: .../as_a_path
    the refusal carries the credential: base_url "http:///v1/sk-live-…" names no host; …
--- FAIL: .../as_a_query_parameter
    the refusal carries the credential: base_url "ftp://example.test/v1?api_key=sk-live-…" must be an http(s) url; …
```

Losing the rest of the value costs a reader nothing here: the error already names which binding (`tier premium`, `the embeddings lane`), and the value itself is in the config file they are about to open.

ancorn-gradion:
### Re: require HTTPS on the vendor lane (CWE-319) — still a follow-up, and here is the precise reason

I re-weighed this rather than restating the earlier answer, because it was raised twice and the second time with a CWE. It is a real weakness and I would support the change. It is not this change, and the boundary is not "scope discipline" in the abstract — it is that this PR patches one triaged finding and this is a different one:

- **What this PR fixes** is a *destination* defect: an operator-supplied `base_url` chose where the server dials and where the model key went. The guard decides which addresses a lane may reach.
- **What you are describing** is a *transport* defect: given a destination that is already allowed, the bytes to it are unencrypted. The guard is the wrong instrument, and adding a scheme rule to it would put two unrelated obligations behind one predicate.

Concretely, requiring https does not touch the attack in the finding this patch closes: an attacker who can write `base_url` writes `https://attacker.example.net` and receives the key over TLS. It closes a different attacker — a network observer on a legitimate misconfiguration — and it should be justified, tested and documented on those terms rather than inheriting this PR's rationale.

**Your redirect sub-point is the strongest part of it and I want to record it accurately** rather than let it disappear into the decline. Go strips `Authorization`, `Cookie` and `WWW-Authenticate` on a redirect to a different domain; it has never heard of `x-api-key` or `x-goog-api-key`, and it does not strip anything on a *scheme downgrade* to the same host. So a vendor answering `301` from https to http would be handed the customer's key in cleartext by a client that believed it was being careful. That is exactly the reasoning `modellist.go`'s `noRedirect` already wrote down for the list path — and the reason it deliberately did **not** change `newOutboundClient`, which also carries streaming completions where a redirect a vendor genuinely uses becomes an outage. Reversing that call needs a product decision about which vendors are allowed to redirect, not a line in a security patch.

Two notes on cost, so the follow-up is easy to size:

- The blast radius is smaller than it looks. Only `anthropic`, `openai` and `gemini` are on the public-only lane, and all three default to https. Every legitimate cleartext deployment — a local Ollama, a vLLM box, a self-hosted gateway — is on the operator lane and would be untouched.
- It lands in the same three places this PR already touches: one clause in `requireDialableEndpoint`, one row in the docs table, one case in `TestABaseURLIsCheckedOnEveryProfile`.

Happy to open it as a labelled issue (`priority:` + `area: ai`) with this analysis, or to fold it in here if a human reviewer would rather have it in one merge — say which and I will do it. What I would rather not do is take a second, differently-shaped security change into a patch on the strength of a bot's escalation alone.

ancorn-gradion:
### Reversing both declines — one was out of scope for the wrong reason, the other made the control incoherent

I said no to two findings and I was wrong about both. Fixed in `261a9105` and `2c68db22`.

**Redirects (CWE-522, #discussion_r3956458151).** I declined this citing `modellist.go`'s comment. I should have read the remediation for the finding this PR patches, which asks for it in as many words:

> Wire the tree's existing control into newOutboundClient (outboundtransport.go:101) … **and set CheckRedirect to return http.ErrUseLastResponse.**

So it was never out of scope; I argued scope from the wrong document. It is implemented now, though **scoped to the hops that leak** rather than blanket, which is where I still differ from the literal suggestion:

- a **cross-host** redirect is refused — Go strips `Authorization` and `Cookie` across a domain change and has never heard of `x-api-key` or `x-goog-api-key`;
- an **https → http** redirect is refused — Go strips nothing when only the scheme changes;
- a **same-host, same-scheme** redirect is followed.

That closes the leak without the outage `modellist.go` declined to risk: this client also carries streaming completions, where a redirect a vendor genuinely uses must still work. `modellist.go` keeps its stricter blanket `noRedirect` on the list path, where a 3xx never happens in normal operation. If a reviewer would rather have blanket refusal everywhere, say so and I will widen it.

**Cleartext (CWE-319, #discussion_r3956458132 and #discussion_r3956948582).** My argument was that this is a transport defect rather than the destination defect the PR fixes. That was defensible right up until the redirect rule landed — at which point the guard refuses an https→http *redirect* because "the key would leave in clear" while happily accepting a binding **configured** that way from the start. Guarding one door of two is not a position worth defending, and the inconsistency is itself the argument.

So the vendor lane now requires https:

| lane | providers | may dial |
|---|---|---|
| operator endpoint | `ollama`, `vllm`, `openai_compatible` | loopback, private ranges, public hosts — **http or https** |
| vendor | `anthropic`, `openai`, `gemini` | public hosts, **https only** |

The operator lane is deliberately untouched: a local Ollama, a vLLM box and a self-hosted gateway are reached over http all the time, and the key travels no further than the operator's own network. Blast radius on the vendor lane is nil — all three default to https.

Two details worth a reviewer's eye:

- **The address rule is asked before the scheme rule.** For `http://10.0.0.5:6379` on an `anthropic` tier both are wrong; the operator needs the address one first, because a vendor binding pointed into their own network is the more surprising mistake.
- **`TestTheWriteRuleAndTheDialerAgree` now spells its urls `https`.** The scheme is not the address question and the dialer never sees one, so a scheme disagreement would drown out the address disagreements that table exists to catch. The scheme rule has its own cases in `TestABaseURLIsCheckedOnEveryProfile`.

Both are mutation-checked: removing the redirect policy fails the per-provider assertion, removing the host comparison alone lets the cross-host and subdomain hops through, and removing the scheme clause re-accepts `http://api.anthropic.com`. Docs updated in both places, plus a troubleshooting row.

---

Two notes for whoever merges this:

- **This branch is 247 commits behind `main`.** Nothing on `main` touches the files this PR changes except one comment-only edit in `outboundtransport.go`'s `RouteWriteDeadline` block, and `ci` runs against the merge ref, so the green is a green for the merged result. Still worth an "Update branch" before merging if the repo prefers it.
- **`review-coverage` is red and that is correct.** Its own header says it is not a gate — `ci` is the one required check — and a red is what it is *for*: the commits that fix a review's findings are outside that review by construction. It should refresh when the bots re-review this head.

PR #4985, opened 2026-09-08
fix(agents): a refusal that changed something does not give the key back

### LINKED ISSUE #1073: A partial write can be reported as a refusal, and its idempotency claim released
A partial write can be reported as a refusal, and its idempotency claim released

`applyAutoExecuteAndStageResidue` (`compose/agentsplit.go`) is the per-field human-edit-precedence
split's mixed-conflict path: the agent-owned fields are applied, and the human-owned residue is
staged for approval.

If the staging step fails after the auto-execute half has committed, the response is a 4xx. The
write already happened. `settleClaim` releases a non-2xx idempotency claim, so a retry under the
same key re-runs the applied half — a second write of fields the first request already wrote.

Two paths reach it. The pre-existing one (`sErr`, the staging engine itself failing) carries a
message telling the caller the other fields already applied. The one added by the governed-call
seam (#928) — a `Guards` refusal produced by state that changed between the two writes — does not
carry that framing, so a caller reads a plain refusal for a request that partly succeeded.

## Why it is narrow

For the new path specifically, `rejectUnknownFields` can only reject a residue key that
`HumanOwnedConflicts` returned, which means it was already in the record's audit trail — so the
pure-function branch is not reachable. What remains is a genuine race: the record's system of
record or its visibility changing between the auto-execute write and the staging read.

## What it needs

Either a second return shape out of the staging seam so both paths can say "the other fields
already applied", or the two writes made atomic. The first is a change every caller of that seam
sees, including the tool door; the second is a larger transaction than the split currently takes.
Neither belongs in the change that found it.

## The invariant to gate afterwards

A response that reports a refusal describes a request that changed nothing.

Found while building #928.

dungbanguyen:
Closes #1073.

The per-field split applies the agent-owned fields and then stages the human-owned residue. The staging call can only run **after** the first half commits — approvals resolves its own target version inside the staging transaction — so a failure there answers a refusal for a request that already wrote.

Two things followed from that, and both are worse than the failure itself.

## 1. The key went back, so the retry ran the applied half again

`settleClaim` read every non-2xx the same way and released the claim. That is right for a request refused **before** it wrote, and exactly wrong for one refused after: the retry the caller is entitled to make runs the committed half a second time under the same key — the one thing the key exists to prevent.

The handler reports the write now, and a refusal that reports one is **recorded** rather than released. The next attempt under that key meets the `claimFailed` answer, which was already built and had no writer on this door:

> an earlier request with this idempotency key failed after it had already started; check whether it took effect before retrying under a new key

That is the honest sentence, because whether the effect landed is exactly what nobody at that layer can determine.

The signal travels **down as a pointer and is read on the way back up**, because a context value a handler sets is invisible to the middleware that dispatched it — the request's context is the one the middleware built.

## 2. The caller was told nothing at all

The ticket says the pre-existing path "carries a message telling the caller the other fields already applied". It does not. That sentence was wrapped in a plain `fmt.Errorf`, which reaches `httperr.Write`'s **unhandled** path: logged for the operator, answered to the caller as `{"code":"internal"}` with no detail at all.

A partial write is precisely the outcome nobody can discover from the outside, and the one party who has to decide what to do next got the one answer that says nothing. It is a classified refusal now — it names the withheld fields, says the other half landed, and says not to retry unchanged, which matters because a caller reading only the status would take a 5xx as the invitation to retry it usually is.

## What the ticket asked for, and what this does instead

The ticket offered two fixes: a second return shape out of the staging seam, or making the two writes atomic — and said neither belonged in the change that found it. Both are still true, and neither is needed for the defect as it stands:

- **The framing gap is closed at the source.** Both post-write exits go through one refusal now, so there is no path that reports a partial write as a plain one. The seam does not need a new shape for that.
- **The Guards path the ticket describes is already gone.** Everything the split can refuse on before the write is now settled before the write (the comment on `applyAutoExecuteAndStageResidue` cites this issue). What remains is the genuine race the ticket names: the record's system of record or its visibility changing between the two writes.

What was left unfixed by both of those, and is fixed here, is the idempotency half — which is the part that turns one bad request into two writes.

## Tests

`claimIsReleased` is the rule as a function, so a case states it without a database — the two arms it chooses between (`recordClaimOutcome`, `releaseClaim`) are each already held. Four rows: a refusal that wrote nothing releases; one that wrote does not; a success is recorded either way, so the marker cannot make a completed call retryable by accident.

Beside it, the split's own pair: a staging failure after the applied half reports that it wrote and still names the withheld fields, and — the control — a refusal that happens **before** the write reports no write, so an ordinary refusal still hands the key back.

**Mutation checks:**

| mutation | killed |
|---|---|
| release on every non-2xx | the "had already written" row |
| drop `markWriteCommitted` | the after-the-applied-half case |
| answer the staging failure as a plain error | the same case, on the detail |

`make check-backend` green; the `compose/integration` idempotency and human-precedence lanes green.

### DIFF
--- a/backend/internal/compose/agentsplit.go
+++ b/backend/internal/compose/agentsplit.go
@@ -264,6 +264,11 @@ func applyAutoExecuteAndStageResidue(w http.ResponseWriter, r *http.Request, nex
 		buffered.flushTo(w)
 		return
 	}
+	// From here on this request HAS changed the record, and anything it answers
+	// after this line is a failure rather than a refusal. The idempotency layer
+	// reads this to decide whether the key goes back: released, the retry would
+	// re-run the half that just committed under the same key.
+	markWriteCommitted(r.Context())
 	// UseNumber keeps integers exact: a plain interface{} decode renders
 	// every JSON number as float64, silently truncating any value past
 	// 2^53 on this re-encode path (money-minor fields, version).
@@ -271,9 +276,7 @@ func applyAutoExecuteAndStageResidue(w http.ResponseWriter, r *http.Request, nex
 	dec := json.NewDecoder(bytes.NewReader(buffered.body.Bytes()))
 	dec.UseNumber()
 	if uErr := dec.Decode(&record); uErr != nil {
-		httperr.Write(w, r, fmt.Errorf(
-			"agent gate: %s applied the permitted fields, but its response cannot carry the staging note for the withheld human-edited fields (%s): %w",
-			pol.Op, strings.Join(split.Conflicts, ", "), uErr))
+		httperr.Write(w, r, partiallyApplied(split.Conflicts, uErr))
 		return
 	}
 	// No version pin travels from here, and the residue is still staged against
@@ -299,8 +302,7 @@ func applyAutoExecuteAndStageResidue(w http.ResponseWriter, r *http.Request, nex
 			restSummary(pol, r, split.Staged),
 	})
 	if sErr != nil {
-		httperr.Write(w, r, fmt.Errorf("the other fields were updated, but staging the human-edited fields (%s) failed: %w",
-			strings.Join(split.Conflicts, ", "), sErr))
+		httperr.Write(w, r, partiallyApplied(split.Conflicts, sErr))
 		return
 	}
 	record["staged_approval"] = map[string]any{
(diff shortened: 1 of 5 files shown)


dungbanguyen:
Valid — fixed in `c86e5d556`.

The package doc still said "a failed attempt releases the claim so the client may retry the same key", which is now half of what `settleClaim` decides and the wrong half to leave standing alone: a reader taking it at face value would assume a refusal always hands the key back, which is exactly the assumption this change had to break.

It states the pair now, with the reason each way — released for a refusal that changed nothing (and the original rationale, that replaying stored failures pins transient faults and breaks the stage-then-redeem retry, is kept where it belongs), recorded for one that had already written, with the `claimFailed` sentence the next attempt meets. It also names `markWriteCommitted` as how a handler reports one, and says the per-field split's staging half is the only door that does today — so the next reader knows both the mechanism and its current reach.

PR #4987, opened 2026-09-08
fix(agents): a refusal that names an input says so in the field list

### LINKED ISSUE #1071: The agent door's refusals do not answer the shape the contract declares
The agent door's refusals do not answer the shape the contract declares

The governed-call seam (#928) moved a set of argument guards onto the REST agent door that
previously ran only on the tool door. They refuse the right calls. What they answer with is not
consistently what the contract or the session door answers, and one credential getting a
different machine code from another for the same mistake is the divergence that whole change
exists to remove. Four instances, one root.

**1. `BadArgsError` declares the wrong fault form.** `apperrors`' `MessageFault` is documented as
"a refusal that names NO input the caller can change", explicitly contrasted with `FieldFault`,
and a type is meant to implement exactly one of the three. Almost every `BadArgsError` newly
routed to REST does name an input — `to_phase "vibing" is not a project phase`. The result is a
422 whose `details.errors` array is empty and whose only actionable content is prose, which is
why the coverage gate had to match on a substring rather than a field code.

**2. A merge with a missing `target_id` answers 404 where every sibling answers 422.**
`mergeCommand` yields a zero target id, and the resolver then reads it and returns not-found.
`advanceDealCommand` spends eleven lines establishing the opposite rule — a body field is the
caller's own input and must not take the routed id's existence-hiding answer — and the stakeholder
commands follow it. The handler answers 422; the agent gate answers 404.

**3. The agent door now refuses what the session door accepts.** `rejectUnknownFields` runs in the
create and patch resolvers. Its own doc states the placement rationale as "REST may keep
accepting-and-discarding, while the tool surface refuses up front". On the REST agent door those
are now in tension: an agent's confirm-first create with a stray key gets 422 where a session's
identical request succeeds. The behaviour is defensible — an agent losing a field silently is
worse — but the stated rule no longer describes the code.

**4. A U+FFFD refusal is attributed to `body` even when the replacement rune came from the path or
a header value**, because `json.Marshal` coerces those into the same canonical object.

## The invariant to gate afterwards

One mistake carries one machine code, whichever credential presented it — and a refusal that names
an input says so in the structured field list, not only in prose.

Found while building the governed-call seam (#928). Related: #1065.

dungbanguyen:
Closes #1071.

Four instances of one root, and the root is the divergence the governed-call seam exists to remove: **one mistake carrying two answers depending on which credential presented it.**

## 1. `BadArgsError` declared the wrong fault form

`apperrors` documents `MessageFault` as *"a refusal that names NO input the caller can change"*, explicitly contrasted with `FieldFault`. Almost every `BadArgsError` newly routed to REST does name one — `to_phase "vibing" is not a project phase` — so the 422 arrived with an empty `details.errors` and prose as its only actionable content. That is why the coverage gate had to match a substring rather than a field code.

It carries an optional `Field` and implements **`FieldFaults` — the plural**. That is the only one of the three forms that can answer both shapes from one method: a type implements exactly one, and `errors.As` matches whichever it finds, so a conditional choice between the singular and the message form is not expressible. One entry when a field is named, **none** when it is not — inventing an entry for a refusal that names no input would point a caller at an argument that is not theirs to change, which is the whole reason the message form exists.

`rejectUnknownFields` — the create/patch guard the seam actually put on this door — names `fields`. Deliberately `fields` and not one of the unknown *keys* inside it: those are not contract fields at all, so a client cannot look one up.

## 2. A merge with no `target_id` answered 404

The zero id travelled to the resolver, which read it and reported not-found. An agent learned that a record was invisible where a session on the same route is told it omitted an argument.

Existence-hiding wins where a **routed** id names a record — an agent must not learn from a validation message that an id it guessed is well-formed and merely invisible. A body field the caller left out is not one. That is the rule `advanceDealCommand` spends eleven lines establishing for `to_stage_id` and the stakeholder commands follow; the merge now follows it too, through the same `httperr.RequireBodyID`.

## 3. The stated placement rule no longer described the code

`rejectUnknownFields`' doc said *"REST may keep accepting-and-discarding, while the tool surface refuses up front"*. That stopped being true when the seam put these guards on the REST agent path.

The line is the **credential**, not the door — and the asymmetry is the intended one: a session's reader is looking at the screen the write answered, and an agent's is not. The comment says that now, and says plainly that the old reading is false of half of REST.

## 4. A replacement rune was attributed to `body` wherever it came from

`json.Marshal` coerces the path, the body and the header values into one canonical object, and the scan ran over the finished thing. A caller told to fix a body that is already clean has been handed a task it cannot perform.

It runs **per member** now, each marshalled on its own — which is also necessary, because the case this catches (an escaped unpaired surrogate, valid UTF-8 on the wire) only becomes U+FFFD once encoded, so scanning the raw input cannot see it. `operation` is excluded: it is this server's own route id, so being told to fix it would be advice about a value the caller never sent.

## Tests

Four cases, driven through `httperr.Write` so each asserts what actually reaches the wire:

- a `BadArgsError` naming an input answers one `details.errors` entry with its field and code;
- one naming none carries no entry, so the message form's own case is not lost;
- a merge with no `target_id` names the field instead of hiding a record;
- a replacement rune in the body names `body`, and one in the path names `path`.

**Mutation checks:** dropping the `Field` guard fails the first; dropping the body-id check fails the merge case; removing `path` from the per-member scan fails the path row and only that row.

## What this does not do

The ticket's closing line — *"the invariant to gate afterwards"* — is not built here. `&BadArgsError{` has 88 construction sites, and most of them answer the MCP tool door only, where `Dispatcher.explain` matches the type before `httperr.Classify` is ever consulted, so the fault form is not read at all. Migrating those is real work with no defect behind it yet, and a gate demanding a field everywhere would be demanding it where nothing reads one. The form is fixed and the REST-reachable guard names its field; the rest is a register worth opening when a second door starts reading them.

`make check-backend` green; the `compose/integration` agent lanes green.

### DIFF
--- a/backend/internal/compose/agentcommandrecord.go
+++ b/backend/internal/compose/agentcommandrecord.go
@@ -20,6 +20,7 @@ import (
 	"net/http"
 
 	"github.com/margince/margince/backend/internal/modules/agents"
+	"github.com/margince/margince/backend/internal/platform/httperr"
 	"github.com/margince/margince/backend/internal/shared/kernel/ids"
 )
 
@@ -48,6 +49,19 @@ func mergeCommand(pol agentPolicy, deps restCommandDeps, r *http.Request, body [
 		TargetID ids.UUID `json:"target_id"`
 	}](body)
 	if err != nil {
+		return nil, err
+	}
+	// An omitted target is the CALLER'S OWN INPUT, and answers 422 naming it —
+	// the rule advanceDealCommand states at length for `to_stage_id` and the
+	// stakeholder commands follow. Without it the zero id travelled on, the
+	// resolver read it, and an agent got the resolver's existence-hiding 404
+	// where a session on the same route gets the handler's 422: one mistake,
+	// two machine codes, keyed on which credential presented it.
+	//
+	// Existence-hiding wins only where a ROUTED id names a record, because an
+	// agent must not learn from a validation message that an id it guessed is
+	// well-formed and merely invisible. A body field it left out is neither.
+	if err := httperr.RequireBodyID("target_id", in.TargetID); err != nil {
 		return nil, err
 	}
 	return agents.NewMergeCall(deps.records, agents.MergeCommand{
--- a/backend/internal/compose/agentgatecanon.go
+++ b/backend/internal/compose/agentgatecanon.go
@@ -124,6 +124,12 @@ func canonicalHeaders(h http.Header, keys idempotencyKeyBinding) map[string]stri
 // canonical form catches an escaped unpaired surrogate (`"\udcff"`), which
 // is valid UTF-8 on the wire and still decodes to U+FFFD, so the byte check
 // cannot see it.
+//
+// That scan runs PER MEMBER rather than over the finished object, because the
+// object is where the three inputs stop being distinguishable: json.Marshal
+// coerces the path, the body and the header values into one string, so a
+// replacement rune from any of them was reported as `body`. A caller told to
+// fix a body that is already clean has been handed a task it cannot perform.
 func canonicalRESTCall(op, path string, headers http.Header, body []byte, keys idempotencyKeyBinding) (json.RawMessage, string, error) {
 	if !utf8.Valid(body) {
 		return nil, "", httperr.Validation("body", "invalid_utf8", "request body must be valid UTF-8")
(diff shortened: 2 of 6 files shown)


dungbanguyen:
Both valid — fixed in `7d6758fce`.

**The unbounded detail (P2)** is the sharper of the two, and it is worse than "the replaced branch bounded it": the plural branch was the *only* value on the module-declared path that was not bounded, and it is the one most likely to be long. A type carrying a **list** has a sentence naming all of them, and `BadArgsError`'s `Guidance` is deliberately unbounded — bounding it with the echo truncated the accepted-field list mid-word, deleting the actionable half of a message whose reader had just proved they did not know the vocabulary — and `Error()` concatenates the two.

Bounded in `moduleDeclaredFault` rather than in this caller, so it covers **every** `FieldFaults` implementer instead of the one that surfaced it. A case pins it, and asserts the ellipsis as well as the length — without that it would pass on a short message whatever the code did.

**The stale comment (P3)** — correct, and my oversight: the doc block above `FieldFaults` still opened by describing a `MessageFault` implementation, which contradicts the method under it. It describes the plural path now, including why the plural is the only one of the three forms that can answer both shapes from a single method.

The `craftsmanship` failure was mine too: the per-member replacement-rune scan takes decoded JSON, which the gate reads as a bare `any`. Waived where it is, with the reason `entityRefs` gives for the same shape — the members really are arbitrary decoded JSON, and a concrete type would be a claim about payload shape this must not make.

PR #4989, opened 2026-09-08
fix(capture): the auto-enrich sweep drains oldest first, so none waits forever

### LINKED ISSUE #2021: Auto-enrich takes the newest 500 each pass, so an older company can wait forever
Found by a Codex review during PR #2018, which widened the sweep to every company with a primary domain. Latent today; the fix is its own change.

## What is wrong

`ListDueOrgs` always returns the newest fixed prefix — `ORDER BY o.created_at DESC LIMIT $2` (`backend/internal/modules/capture/autoenrich.go:101-106`) — and a cursor row is only written once an organization has actually been queued (`:228-247`).

So in a workspace that gains at least 500 newer eligible companies between daily passes, an older company that was never selected has no cursor and stays behind the limit **indefinitely**. It is not retried, not exhausted, not visible as skipped — simply never reached.

The org-name sweep documents this exact failure mode and avoids it with keyset paging (`backend/internal/modules/people/orgnamepromotion.go:230-264`).

## Why it is not urgent

It needs sustained arrivals above `autoEnrichDailyCap` (500/workspace/day). The demo workspace has 196 organizations total, so the whole backlog clears in one pass. A large real installation is where this bites.

## Also: the shared triage allowance

The same daily counter is spent by domain triage (`backend/internal/compose/capturedomaintriage.go:93-108`), and there is already a test showing a newly captured domain goes unprocessed once the counter is full (`backend/internal/compose/capturedomaintriage_integration_test.go:114-142`). A wide first-boot backlog can therefore starve triage arriving later that day. Worth deciding deliberately whether auto-enrich may consume the whole allowance, or should take a share of it.

## Done looks like

Keyset paging over the candidate set, following the org-name sweep, so every eligible company is eventually reached regardless of arrival order — and a decision recorded on the triage/enrich split of the daily cap.

dungbanguyen:
Closes #2021 (the starvation half — see **What I did not decide** below).

The pass took the **newest** page of companies due for a dossier. It queues what it takes — writing a cursor row that drops that company out of the due set until its next attempt falls due — so the set shrinks by exactly what was worked, and which end the page comes from decides whether it drains at all.

In a workspace gaining more eligible companies between passes than the daily cap, arrivals keep landing in front of a company that has never been reached. It waits indefinitely: not retried, not exhausted, not visible as skipped, so nothing anywhere says it was missed.

## Ordering, not paging

The issue suggests following the org-name sweep's keyset paging. That sweep pages **to exhaustion in one pass**, and it can: its per-candidate work is one in-memory decision, no model call. This lane's is a model-backed site read under a daily cap, so a pass takes exactly one page by construction — a cursor parameter would have no second page to fetch.

What actually guarantees progress here is the **ordering**, because the due set shrinks by what the pass worked. Oldest first drains in arrival order: everything ahead of a given company is consumed before it, and nothing new can be inserted ahead of it, because nothing new is older.

So this is a one-line query change with the argument written beside it, rather than a paging parameter nothing would page with.

**Ordered on the id, not `created_at`.** The order has to be TOTAL: two companies captured in one statement share an instant, and a tie under a `LIMIT` is a coin toss that can seat the same row at the boundary every pass — starvation again, by a different route. The id is a uuidv7, so ordering by it *is* ordering by arrival. The due rows sit at the head of that order by construction, so the walk stops early rather than sorting the table; `idx_org_created_keyset` no longer serves this query and the primary key does.

## Test

`TestTheOldestCompanyIsSweptFirstSoNoneWaitsForever` — a page of one against three eligible companies is the same shape as 500 against a day's arrivals, and it is the shape an assertion can read.

It asserts **both** halves, because oldest-first is only progress if the queue advances: the pass takes the company that has been due longest, and once that one holds a dossier the next pass takes the one behind it rather than the same row again. Mutation-checked — restoring `ORDER BY o.created_at DESC` fails both, taking `newest.example` twice.

## What I did not decide

The issue's second item: whether auto-enrich may consume the **whole** daily allowance it shares with domain triage, or should take a share. That is a product call about which lane matters more on a busy first boot, and there is already a test showing a newly captured domain goes unprocessed once the counter is full. I have left the split exactly as it is and not touched `sweepDomainTriage`; the issue should stay open on that point, or move to its own with the ruling named.

`make check-backend` green.

### DIFF
--- a/backend/internal/compose/captureautoenrich.go
+++ b/backend/internal/compose/captureautoenrich.go
@@ -10,7 +10,8 @@ package compose
 // The anchor is excluded; cold start has already read it. Per workspace, when
 // the capture_auto_enrich flag is on, it enqueues a deep read
 // (system:capture_auto_enrich, auto-applied on completion) for each due org —
-// newest first, under an atomically-reserved daily cap. It is the
+// oldest first, under an atomically-reserved daily cap, so a company behind a
+// day's worth of arrivals is still reached (ListDueOrgs says why). It is the
 // self-healing reconciler: the prompt trigger is the organization-event
 // consumer (orgautoenrich.go), which queues this same workspace pass the
 // moment a company appears, and anything that slips through — the worker
--- a/backend/internal/compose/captureautoenrich_integration_test.go
+++ b/backend/internal/compose/captureautoenrich_integration_test.go
@@ -255,3 +255,62 @@ func TestAutoEnrichExpireExhausted(t *testing.T) {
 		t.Fatalf("cursor = (%q, %v), want (exhausted, <nil>)", outcome, nextAttempt)
 	}
 }
+
+// A company waiting behind a day's worth of newer arrivals is still reached.
+//
+// The pass takes ONE page, bounded by the daily cap, and queues what it takes —
+// which writes a cursor row that drops that company out of the due set until its
+// next attempt falls due. So the set shrinks by exactly what was worked, and
+// which END the page comes from decides whether it drains.
+//
+// Taking the newest end starves: in a workspace gaining more eligible companies
+// between passes than the cap, arrivals keep landing in front of a company that
+// has never been reached, and it waits forever — not retried, not exhausted, not
+// visible as skipped, so nothing anywhere says it was missed.
+//
+// A page of one against three eligible companies is the same shape as 500
+// against a day's arrivals, and it is the shape the assertion can actually read.
+func TestTheOldestCompanyIsSweptFirstSoNoneWaitsForever(t *testing.T) {
+	e := integration.Setup(t)
+	store := capture.NewAutoEnrichStore(e.DB())
+	ctx := e.As(e.Rep1, nil, integration.AdminPerms)
+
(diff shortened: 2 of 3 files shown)


dungbanguyen:
Both fixed — `75d8f629e`.

**The uuidv7 ordering.** You are right and my comment overclaimed. A uuidv7 sorts by its millisecond and then by bits that are process-local or random, so ids minted on two instances inside one millisecond can interleave; it is very nearly arrival order, not exactly it.

The termination argument never needed *exactly*, and the comment now says what it does need: an order that is **total** and that a new row cannot enter arbitrarily far ahead in. A company can be overtaken only by rows sharing its millisecond — a bounded reordering, not a queue it can sit behind. What would break the argument is a non-total order (the `created_at` tie this replaced) or one where arrivals land at the head, which is exactly what `created_at DESC` did.

**The stale comment.** Fixed — `captureautoenrich.go` said "newest first", which is now the opposite of what the query does. It points at `ListDueOrgs` for the reasoning rather than restating it.

PR #5003, opened 2026-09-08
feat(gates): the frontend gets the file-length cap the Go half already has

### LINKED ISSUE #2169: The 80/500 craftsmanship ceilings are enforced on Go only: no frontend length gate exists
## What is missing

Nothing in the frontend gate bounds a file's or a function's length. The two
gates that enforce the repo's ceilings — `scripts/check-go-file-length.sh` (500
lines, ratcheted via `scripts/go-file-length-waivers.txt`) and `craft static
--strict` (80 code lines/function, 500/file for product code) — both walk Go
trees only: `backend/`, `extensions/`, `fixtures/`, `desktop/`. `make check-fe`
runs biome, vitest, tsc and the design-system gates, none of which measure size.

So the ceiling CLAUDE.md states as a craftsmanship rule is enforced on one half
of the product and unenforced on the other.

## Where it already bites

`frontend/src/app/agentrail.tsx` is **1163 lines**, and its `AgentRail`
component spans about 162 — both well past the ceilings the Go half is held to.
It is the single agent surface (its own header records that two other placements
were built, judged, and lost), so it accretes: every reading the rail reports
adds a signal, a branch in `derive()`, and an arm in `barLine()`. The most
recent addition (the personal agent-activity read) added roughly 100 lines to
it, and no gate said anything.

It is not the only one — a sweep will find others — but it is the clearest case,
because the file's role guarantees it keeps growing.

## Why it matters

The ceilings are not a tidiness preference. CLAUDE.md states the reason: the
limit asks how much a reader must hold at once. A 1163-line component file is
where the next agent editing this surface loses the thread, and it is exactly
the tree that gets edited by a fresh context most often.

The asymmetry also makes the rule unteachable. A contributor reads the
craftsmanship section, sees the 80/500 ceiling stated without qualification,
and reasonably assumes it applies to what they are writing — then finds it does
not, only after a reviewer says so by hand.

## What to decide

Two parts, and the second is the one that needs a view:

1. **Add a frontend length gate** to `make check-fe`, mirroring the Go one:
   a hard cap with a ratcheted waiver file so existing offenders may shrink but
   never grow. That is mechanical.
2. **Pick the numbers and the seed waiver set.** Product `.ts`/`.tsx` plausibly
   takes the same 500/80 as Go; `*.test.tsx` plausibly takes the 1000/160 the Go
   test ceiling uses. But JSX is more verbose per unit of logic than Go, so the
   same number may not mean the same thing, and whatever is chosen has to admit
   today's tree as waivers or the gate lands red.

Splitting `agentrail.tsx` is worth doing on its own terms, but it should follow
the gate rather than precede it — otherwise the next surface grows the same way
and nothing catches it either.

## Scope note

Found while adding the personal agent-activity read to the rail. That change was
reviewed by hand for size and is within the spirit of the ceilings for the code
it adds; this issue is about the absence of the gate, not about that change.

dungbanguyen:
Closes #2169 (part 1, and the choice part 2 asked for — see below).

The craftsmanship ceilings CLAUDE.md states were enforced on **one half of the product**. `check-go-file-length.sh` walks Go trees only, and `craft static` the same, so a screen could grow without limit while a package of the same size failed a push. `agentrail.tsx` reached 1163 lines and nothing said anything.

## What it is

`scripts/check-fe-file-length.sh`, the counterpart to the Go gate, wired into **`frontend-check`** (so `make check-fe` runs it) and **`fe-quality`** (so the CI job does). Same ratchet, all four arms verified by hand:

- a new file over the cap fails;
- a waived file that grows past its frozen count fails;
- a waived file that drops to the cap fails until its entry is deleted, so the cap re-arms;
- a waiver for a file that no longer exists fails.

## Part 2: the numbers, and why

**The caps are the Go ones — 500 for product, 1000 for a test or story.** The issue is right that JSX is more verbose per unit of logic, and right that this is an argument for a higher number. I do not think it is a good one: the ceiling asks *how much must a reader hold at once*, and that is a property of the reader. A 900-line component is 900 lines to read whichever language wrote it.

The seed waiver list is what admits today's tree — **134 entries** — and its length is the measure of the debt rather than an argument for a laxer cap. Setting the cap high enough to admit the tree without waivers would have bought a green gate that bounds nothing.

**Exempt, and neither is a loophole:**

- Generated types (`*.d.ts`, `src/api/public-events.ts`) — the same exemption `*_gen.go` carries. Splitting them splits the generator's output.
- The three i18n catalogs — they are **data**: one entry per key, read by lookup and never top to bottom, and cutting them by size would scatter a translation across files a translator has to find.

**It does not split anything.** The issue asks for the gate to come first, and it is right: splitting `agentrail.tsx` before the gate exists means the next surface grows the same way and nothing catches that either. Every offender is admitted at its current size and can only shrink.

**It does not bound function length.** The Go 80-line rule comes from `craft static`, which has a Go parser; a grep-level equivalent for TS/TSX would either need a real parser or be wrong often enough to be ignored. Better to leave a named gap than to add a check that cries wolf.

`make check-fe` and `make check-backend` both green — including `make-target-parity`, which the new target had to satisfy.


dungbanguyen:
CI caught a real one, which is the gate doing its job one commit early: `logactivity.tsx` grew from 610 to 616 lines on `main` between my freezing the baseline and the run.

```
FAIL: frontend/src/screens/logactivity.tsx grew to 616 lines (waiver froze it at 610) — shrink it, never grow it
```

Rebased and re-froze on the current tip. The baseline is "the tree as it stands when the gate lands", so it is taken at merge time; from then on the ratchet binds and that message is what a growing file gets.

Worth noting it fired in **both** lanes — `fe-quality` and `deterministic-gates` — which is the wiring working as intended: the target is in `frontend-check` (so `make check-fe` runs it locally) and in the `check` aggregate (so a backend-only CI job still sees it).

dungbanguyen:
Confirmed and fixed — the tree spells fixtures **four** ways and I knew one.

```
frontend/src/screens/meetingbrief/fixtures.ts          bare
frontend/src/mcp-apps/handoff/fixture.ts               singular
frontend/src/screens/onboarding-conversation/test-fixtures.ts   hyphenated
frontend/src/screens/brief.fixtures.ts                 dotted (the one I had)
```

The predicate is a named `istest()` function now, matching `(^|[/.-])fixtures?\.tsx?$` alongside the test, story and testkit forms — verified against each spelling, and against `deals.tsx` and `mefixture.ts` to check it does not over-match.

No file changes class today: all four are well under either cap, so the waiver list is unchanged at 134. This is the gate being right about the next one rather than a correction to the baseline.

`mefixture.ts` stays on the product cap deliberately — a boundary of `/`, `.` or `-` is what keeps this from matching any word ending in "fixture", and that file is 98 lines against a 500 cap, so nothing turns on it.

dungbanguyen:
Second baseline race, and it will not be the last while this is open: `contacts.tsx` (939 → 940) and `listtable.tsx` (1632 → 1650) both grew on main between my re-freeze and the run.

That is the ratchet working, but it is also a maintenance step somebody has to repeat, so it is a command now rather than a shell one-liner in a PR comment: `scripts/seed-fe-file-length-waivers.sh`. It writes exactly what the gate reads — exemptions and cap split included — so the two cannot disagree about which files are in the population.

Its header says what it is **not** for, because the obvious misuse is the dangerous one:

> It is NOT a way past a failing gate. Once the gate is in force, a file that grew is a file to shrink, and running this would freeze the growth instead — which is the one thing the ratchet exists to refuse.

Rebased, re-seeded, and the gate is green on the current tip. If main moves again before this merges I will re-run it; after it merges the baseline stops moving and the ratchet binds.

(A gate caught the doc change too: `TestEverySourcePathNamedInProseExists` refused `make-targets.md` naming a script that was not yet tracked. Worth knowing it reads the tree rather than the working directory.)

PR #5005, opened 2026-09-08
fix(comms): the carriage gate holds the aggregate the send actually applies

### LINKED ISSUE #2047: Declared attachment carriage over-promises: the real ceiling is a 20 MiB AGGREGATE, and exceeding it retries instead of parking
## What is wrong

`connector.Carriage` has no aggregate bound. It publishes `MaxFiles` and `MaxBytesPerFile`, and `GET /v1/channel-providers` publishes both — so Telegram advertises **10 files × 20 MiB** and Gmail **10 × 25 MiB**.

The real ceiling is `maxSendBytes = 20 MiB` **in total**, enforced in `backend/internal/compose/commsattachments.go` when the bodies are read for send. That is one tenth of what either transport promises.

## Why it bites a person

Attach three 8 MiB files to a Telegram reply:

1. The composer reads the directory, sees 10 files at 20 MiB each, and says the message is fine.
2. `carriageRefusal` (`backend/internal/modules/comms/gates.go`) passes — each file is under 20 MiB, the count is under 10. It has no aggregate to check.
3. `ReadForSend` refuses on the aggregate and returns a **plain error**, not a park reason.
4. The dispatcher therefore treats it as retryable: the whole ladder runs, re-reading ~24 MiB from the blobstore on every rung, and the human finally gets *"the retry ladder is exhausted after N attempts"* — a reason that names no cause.

So the failure is late, expensive, repeated, and unexplained, and the number the human was shown before pressing send was wrong.

## The fix, as we see it

- Add an aggregate field to the published `Carriage` descriptor (a `MaxTotalBytes`) and to the `attachments` object on `GET /v1/channel-providers`, so the composer warns with the real number.
- Have `carriageRefusal` check it and **park** with a reason that says what to do — the sibling refusals in that function are the model.
- Keep `maxSendBytes` as the backstop it is; the point is that it should never be the thing a human hears from first.

## Notes

- **Pre-existing for Gmail**, newly reachable on Telegram, which is the transport whose per-file cap happens to equal the aggregate cap exactly.
- Not fixed in the Telegram carriage PR: it is a change to a published contract field and it affects every sending transport, so it wants its own change rather than riding a connector PR.
- Found by an adversarial review pass on that PR, alongside a dropped-capability defect that was fixed there.

dungbanguyen:
Closes #2047.

A transport declared **ten files at 20 MiB each** and the real ceiling was **20 MiB in total** — one tenth of what was published. The issue's own scenario: three 8 MiB files pass every published bound, so the composer says the message is fine, the carriage gate passes, and the read path refuses after opening every object. The delivery then climbs the whole retry ladder, re-reading ~24 MiB from the blobstore on each rung, and the human is finally told *"the retry ladder is exhausted"* — a reason that names no cause.

## The aggregate is its own bound

`MaxFiles` and `MaxBytesPerFile` cannot express it between them, which is exactly why the gate could pass a message the send refuses.

- **`connector.Carriage.MaxTotalBytes`** — a provider's own aggregate, zero meaning it has none of its own.
- **`carriageRefusal` checks it**, from the staged sizes, before any object is opened, and parks with a reason naming **the total, the bound, and what to do**. "Too big" without both numbers leaves a person guessing how much to drop.
- **`GET /v1/channel-providers` publishes the EFFECTIVE bound**, from the same helper the gate refuses on — so the number a composer warns with is the number a send applies. Publishing the provider's own would tell a composer nothing about the budget that actually stops the message.

## One budget, one spelling

`maxSendBytes` moves to `comms.MaxSendBytes`, beside the gate that tells a human about it, and the read path reads it from there. Two spellings of one budget is precisely how the number a composer is shown came to differ from the number a send applies.

The read path **stays** as the backstop, and the comment says why rather than leaving it looking redundant: the gate sums the sizes the rows recorded, this sums the bytes actually read, and a send may not carry more than the budget whichever of the two is wrong.

## The one bound with no "zero means unbounded" reading

Every other field follows "a zero bound means no limit beyond the contract's own". `max_total_bytes` is never zero and the contract says so: every transport has an aggregate, because the **product** has one whether or not the provider declares its own. A provider with a tighter aggregate is held to its own; one with a looser one does not widen the product's.

## Tests

Four cases added to `TestAttachmentCarriageGateParksRatherThanStripping`, both mutations checked:

- The issue's scenario — three 8 MiB files under a 10 × 20 MiB declaration — parks, and the reason carries `24.0 MiB`, `20.0 MiB` and what to do instead.
- The same three files at 6 MiB each go, so the bound is the aggregate and not "three files".
- A provider's **tighter** aggregate binds before the product's.
- A provider's **looser** aggregate does not widen the product's — the case that would otherwise let a connector opt out of the budget by declaring a big number.

### DIFF
--- a/backend/api/crm.yaml
+++ b/backend/api/crm.yaml
@@ -31980,6 +31980,7 @@ components:
             max_total_bytes:
               type: integer
               format: int64
+              minimum: 1
               description: |
                 Largest total across every file on one message, in bytes — the smaller of
                 this transport's own aggregate and the product's send budget. Always
--- a/backend/api/crm.yaml
+++ b/backend/api/crm.yaml
@@ -31958,7 +31958,7 @@ components:
             re-conflate a deployment fact with a tenant one.
         attachments:
           type: object
-          required: [carries, max_files, max_bytes_per_file, max_body_with_files]
+          required: [carries, max_files, max_bytes_per_file, max_total_bytes, max_body_with_files]
           description: |
             What this transport can carry alongside a message. Published because the
             composer must warn BEFORE a human presses send: a mismatch discovered at
@@ -31968,8 +31968,22 @@ components:
             default, so a connector that never declared carriage reports false rather than
             being mistaken for capable. A zero bound means "no limit beyond the contract's
             own", never "zero allowed".
+
+            `max_total_bytes` is the exception to that rule and is never zero: every
+            transport has an aggregate, because the product has one whether or not the
+            provider declares its own. It is the bound a composer must warn against
+            FIRST — `max_files` and `max_bytes_per_file` cannot express it between them,
+            so a message of ten files each under the per-file cap can pass both and still
+            be ten times what a send may carry.
           properties:
             carries: { type: boolean }
+            max_total_bytes:
+              type: integer
+              format: int64
+              description: |
+                Largest total across every file on one message, in bytes — the smaller of
+                this transport's own aggregate and the product's send budget. Always
+                present and always positive.
             max_files:
               type: integer
               description: Most files in one message. Never more than the contract's own `attachment_ids` cap of 10.
--- a/backend/internal/compose/commsattachments.go
+++ b/backend/internal/compose/commsattachments.go
@@ -170,21 +170,20 @@ func (a commsAttachments) ReadForSend(
 // published bounds admit, because a transport declaring ten files at 20 MiB each
 // promises ten times what this allows, so parking is the least it owes the
 // person who was told the message would go.
+// It is now the BACKSTOP rather than the first thing a person hears: the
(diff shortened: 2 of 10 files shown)


dungbanguyen:
Confirmed and fixed — `minimum: 1` added.

You are right that the promise was only in prose, and there is a sharper reason it matters here than "the doc says positive": **zero is the exact value every other bound on this object uses to mean "no limit beyond the contract's own"**. A generated consumer reading `max_total_bytes: 0` would apply that same reading and conclude the aggregate is unbounded — which is the defect this field exists to end, arriving through the schema instead of through the descriptor.

`make check-backend` and `make check-fe` green with the regenerated types.

Issue #5030, opened 2026-09-08
main is red: TestEveryRawReaderOfTheSettingTableCarriesAVerdict fails on an unratified raw setting read in leadsla.go


LarsGradion:
`make check-backend` fails on `origin/main` (`bcd94399a`), independent of any working change. Reproduced on a pristine detached worktree of `origin/main`.

```
--- FAIL: TestEveryRawReaderOfTheSettingTableCarriesAVerdict
    settingreaders_test.go:163: internal/modules/people/leadsla.go:unassignedEscalationTarget:SELECT value FROM setting WHERE key = $1
      reads the setting table in raw SQL.
    settingreaders_test.go:170: setting reads: 13 seen, 7 ratified outside internal/platform/settings
```

`unassignedEscalationTarget` in `backend/internal/modules/people/leadsla.go` reads the `setting` table with a raw statement. That table has no row-level security and no row-scope clause, so the object gate at the reader is the only control on it, and a raw statement asks none.

The gate names the three ways out: read through `settings.Get`/`GetTx`, or through `settings.ApplyTx` with the entry declared `MachineryApplied`, or ratify the read in `rawSettingReads` with what it costs.

Arrived with the lead-SLA work in 2cb3d5345 ("A breach on a lead nobody owns reaches somebody"). Filed rather than fixed because it is outside the diff of the branch that found it; whoever owns the SLA escalation path should pick the right seam rather than a passing session guessing at ratification.

Priority note: this blocks the merge gate for every branch until it is fixed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

LarsGradion:
Fixed independently by #5027 (`3376bc379`), which landed while this was being filed. `TestEveryRawReaderOfTheSettingTableCarriesAVerdict` is green on `origin/main` again. Closing as a duplicate of that work.

PR #5031, opened 2026-09-08
fix(capture): a cleared verdict re-opens the mail it cleared

### LINKED ISSUE #5009: A cleared counterparty verdict never re-opens the mail it cleared: 246 of one mailbox's messages stay participants-only
## What happens

A mailbox on the default `classified` posture holds each message until something judges its sender. When the counterparty verdict then judges that sender **`real / person`**, nothing re-evaluates the messages already captured from them. They stay `audience = participants`, reason `posture`, permanently.

The result is a newly-connected mailbox where most correspondence is invisible to the team even though the classifier cleared the senders.

## Observed on staging (`8c1cd8a39`)

`chi.phanphuong@gradion.com`, connected 2026-09-08 11:16, 1,161 activities.

Audience split versus two mailboxes connected earlier:

| mailbox | workspace | held on `posture` |
| --- | --- | --- |
| lars@gradion.com | 1,155 | 17 |
| charlotte.nguyen@gradion.com | 174 | 15 |
| **chi.phanphuong@gradion.com** | **327** | **525** |

All three carry `mail_posture = 'classified'`. The posture is not misconfigured.

Breaking down her 525 posture-held messages by their sender's verdict:

| verdict | mails |
| --- | --- |
| **real / person** | **246** |
| (no verdict yet) | 191 |
| noise / transactional | 58 |
| suppressed | 20 |
| real / role_mailbox | 4 |
| noise / newsletter | 3 |
| noise / spam | 3 |

246 messages whose sender the classifier **cleared as a real person** are still limited to participants. The `noise` and `suppressed` rows are correct — those should stay held.

## It is not a race

Of those 246, comparing `capture_pending_counterparty.resolved_at` against `activity.created_at`:

```
verdict resolved BEFORE the mail was captured   196
verdict resolved AFTER  the mail was captured    54
```

**196 messages were captured after their sender was already cleared** and were still born `participants`. So this is not only "the verdict arrived too late to help" — the decision is not being consulted on the way in either.

## Where the gap is

`compose/confidentialityverdict.go` calls `activities.RecomputeAudienceTx` in four places: when that verdict resolves, the affected rows are recalculated and widen.

The counterparty verdict path (`compose/captureclassify.go`, `compose/capturejobs.go`) calls it **zero times**. Verified on both the deployed commit and current `main` — the set of files calling `RecomputeAudience` is byte-identical between them, and none of `audiencerecompute.go`, `audience.go`, `birthdecision.go`, `captureclassify.go`, `capturejobs.go` or `pending.go` differ between `8c1cd8a39` and `origin/main` (`74ae0d894`).

`NarrowHistoryTx` exists for posture *changes*, but there is no equivalent for verdict resolution.

**Lars believes this is already fixed on main and merely undeployed. I could not confirm that** — see the paragraph above. The deploy is 53 app commits behind, so a fix may exist that I failed to identify; if so, please point me at the commit and I will withdraw this. But on the evidence available the counterparty path still has no recompute on either revision.

## Second symptom, likely the same root

296 of her messages are held with reason `no_record` — no CRM organization exists for the counterparty, so there is nothing to attach the mail to. Since she connected, `organization_domain_disposition` has **24 rows still `pending`** against only 5 decided as `company`. Client companies (FV Hospital was the example raised) therefore never get created, and their mail stays limited.

Whether that is the same missing-hook problem or a separate triage backlog needs its own look; recording it here because it is the other half of "nothing is shared with the team".

## Related

Same shape as #4953: a counterparty verdict resolves and nothing downstream is notified. There it was the signature enrichment pass never being queued; here it is the audience never being recomputed. Both are one missing hook on the same event, and a fix for either should probably cover both.

LarsGradion:
Closes #5009.

A `classified` mailbox holds each message until something judges its sender. When the counterparty verdict answered "a real person", nothing revisited the mail that hold had caught, and nothing consulted the answer on the way in either. On staging that is 246 messages from senders the classifier cleared, still limited to their participants, permanently — 198 of them one client's.

## What changed

**A release.** `capture/widenclearedsender.go`, shaped after the counterparty hold's own widening in `widenhistory.go`: bounded batches, a progress-proving drain, and an exact `ARRAY['posture']` match so only mail the posture ALONE held is re-opened. Claimed ids are sorted before the recompute, because that takes a row lock per activity and two passes claiming different seats' import rows for overlapping activities would otherwise lock them in opposite orders.

**The entitlement is re-read inside the claiming statement**, not passed in. `createCounterparty` answers nil after correcting its own resolution to `suppressed`, and `ResolveReviewedAs` can update no row at all, so "the caller just wrote real/person" is not something a release may believe.

**Two live hooks** — the machine verdict's `KindPerson` arm and the review queue's accept — each bounded at 100 so a large backlog never shares a transaction with a ledger resolution and a person record. **A reconciling stage** on the verdict pass drains the rest; that stage is also what reaches every sender judged before any of this existed, so the existing backlog needs no migration.

**A rung in the birth ladder**, inside the posture step rather than before it, so the workspace floor, a counterparty hold, a confidential marker and an inherited thread verdict all still decide first.

## What it refuses

Each refusal has a test that fails when its clause is reverted (mutation-checked one clause at a time):

| Refused | Clause |
| --- | --- |
| a `held` mailbox | `posture_at_import = 'classified'` |
| a thread verdict that spoke | `verdict_status IS NULL` |
| a seat's own counterparty hold | exact `ARRAY['posture']` |
| a row that never recorded its reasons | same, NULL matches nothing |
| advisor, role_mailbox, organization_sender | `kind = 'person'`, never status alone |

The last one is the sharpest: all three settle status `real`, and an advisor's record is deliberately kept to its owner.

## Deliberately left held

Mail whose stored `counterparty_email` is a colleague or the owner's alias, because the ladder substituted an external recipient when it asked the verdict question. Those rows are not provably about the cleared sender, and matching any cleared participant instead would release mail on the strength of somebody else's clearance.

## Tests

11 new integration tests across `internal/compose` and `internal/compose/integration/capture`, including the backfill drain, its idempotent second run, and a sender larger than one pass's budget finishing on the next tick. Full `internal/compose` and `internal/compose/integration/capture` lanes green.

## Note on the gate

`make check-backend` fails on `origin/main` itself, on `TestEveryRawReaderOfTheSettingTableCarriesAVerdict` (an unratified raw setting read in `people/leadsla.go`, arrived with the lead-SLA work). Reproduced on a pristine worktree of `origin/main`; filed as #5030. Every other leg is green here, including all four craft roots, arch-lint, rls-store-path and the file-length cap.

---
## Summary by cubic
Closes #5009. A `classified` mailbox holds mail until its sender is judged, but a "real person" verdict never re-opened the held mail, and new mail from cleared senders was born held. Now mail held only by the posture is re-opened on clearance, and new mail from already-cleared senders is born open.

**What changed**
- The verdict and review-queue accept each release a bounded batch (100); a reconciling pass on the verdict tick drains the rest.
- The pass also reaches senders judged before this existed, so the existing backlog needs no migration.
- The birth ladder checks the settled verdict inside the posture step, so stricter holds still decide first.
- The release re-reads the ledger inside the claiming statement, so a caller cannot widen by asserting a verdict it did not durably record.
- The release asks the thread's own verdict ledger rather than the row's status, and the reconciling pass's sender scan asks the claim's own predicate.
- Adds 11 integration tests covering the release, its refusals, the backfill, idempotency, and budget resumption.

**What stays held**
- Only rows where the posture alone held the mail are re-opened: `posture_at_import = 'classified'`, `verdict_status IS NULL`, exact `ARRAY['posture']` reasons, and a ledger entry with status `real` AND kind `person`.
- `held` mailboxes, a seat's own counterparty hold, pre-reason rows, and non-person `real` kinds (`advisor`, `role_mailbox`, `organization_sender`) stay held.
- Mail on a thread the seat held stays held even when its row reads no verdict, since a held verdict's stamps can lag a long thread.
- Mail whose stored counterparty is a colleague or the owner's alias stays held, since it is not provably about the cleared sender.

Note: `make check-backend` fails on `origin/main` itself (`TestEveryRawReaderOfTheSettingTableCarriesAVerdict`, filed as #5030); all other checks are green.

<sup>Written for commit c094c43d6f694e2fe9fc88360e26965984cbe958. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/5031?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Mail previously held while awaiting a sender verdict is now released when the sender is confirmed as a real person.
  - Existing held mail is reconciled automatically, and future messages from cleared senders can enter the workspace directly.
  - Other protections—including counterparty, confidentiality, thread, and non-person verdict holds—remain enforced.
  - Processing is performed in bounded batches to support reliable background reconciliation.

### DIFF
--- a/backend/internal/compose/capturejobs.go
+++ b/backend/internal/compose/capturejobs.go
@@ -400,6 +400,13 @@ func (w *counterpartyVerdictWorker) judgeWorkspace(ctx context.Context, workspac
 	if err := w.engine.ReconcileLedgerWorkspace(wsCtx); err != nil {
 		return err
 	}
+	// After the ledger settles, and before anything is offered to a human: a
+	// sender judged on an earlier tick, or by a door whose own release could not
+	// finish, still has mail a posture is holding for a question that has an
+	// answer.
+	if err := w.engine.WidenClearedSendersWorkspace(wsCtx); err != nil {
+		return err
+	}
 	if err := w.engine.StageReviewsWorkspace(wsCtx, 0); err != nil {
 		return err
 	}
--- a/backend/internal/compose/captureverdict.go
+++ b/backend/internal/compose/captureverdict.go
@@ -365,8 +365,15 @@ func (e *CounterpartyVerdictEngine) apply(
 		// would silently start hiding real mail, so there is none.
 		switch kind {
 		case capture.KindPerson:
-			triageDomain, err = e.createCounterparty(ctx, tx, row)
-			return err
+			if triageDomain, err = e.createCounterparty(ctx, tx, row); err != nil {
+				return err
+			}
+			// The mail a `classified` mailbox held while it waited for this
+			// answer. Bounded, and not drained here: this transaction already
+			// carries the ledger resolution and a person record, and a sender
+			// with a thousand held messages would hold it open for all of them.
+			// The reconciling pass finishes what this leaves.
+			return e.widenClearedSender(ctx, tx, row.Email)
 		case capture.KindRoleMailbox, capture.KindOrganizationSender:
 			// Real correspondence with no human to name. The message stays
 			// visible; no contact is invented for a mailbox nobody owns.
--- a/backend/internal/compose/captureverdictaccept.go
+++ b/backend/internal/compose/captureverdictaccept.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/jackc/pgx/v5"
 
+	"github.com/margince/margince/backend/internal/modules/activities"
 	"github.com/margince/margince/backend/internal/modules/approvals"
 	"github.com/margince/margince/backend/internal/modules/capture"
 	"github.com/margince/margince/backend/internal/modules/people"
@@ -169,6 +170,18 @@ func applyCounterpartyAccept(ctx context.Context, tx pgx.Tx, store *people.Store
 	// Accepting the offer IS the assertion that a person is behind the address —
 	// the queue's whole question is whether to create this contact — so the
 	// ledger records that kind rather than leaving the model's guess standing.
-	return created.TriageDomain, pending.ResolveReviewedAs(ctx, tx, proposal.DispositionID,
-		capture.PendingStatusReal, capture.KindPerson, "accepted in the review queue")
+	if err := pending.ResolveReviewedAs(ctx, tx, proposal.DispositionID,
+		capture.PendingStatusReal, capture.KindPerson, "accepted in the review queue"); err != nil {
+		return "", err
+	}
+	// The same release the machine verdict makes, because this door settles the
+	// same question. Bounded here too, with the reconciling pass behind it. The
+	// widen re-reads the ledger itself, so a resolution that did not take
+	// releases nothing.
+	if _, err := capture.WidenClearedSenderTx(
+		ctx, tx, proposal.Email, clearedSenderLiveBound, activities.RecomputeAudienceTx,
+	); err != nil {
+		return "", err
+	}
+	return created.TriageDomain, nil
 }
--- a/backend/internal/compose/captureverdictwiden.go
+++ b/backend/internal/compose/captureverdictwiden.go
@@ -1,0 +1,91 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+package compose
+
+// Releasing the mail a `classified` mailbox held while it waited for a verdict
+// on the sender.
+//
+// Two callers, and the split is about transaction size rather than about
+// authority. The verdict's own transaction releases a bounded batch, so the
+// mailbox opens the moment the sender is judged; this pass drains whatever is
+// left, including every sender judged before the release existed at all.
+//
+// The entitlement is not passed between them. capture's predicate re-reads the
+// ledger inside the statement that claims the rows, so neither caller can widen
+// by asserting a verdict it did not durably record.
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/jackc/pgx/v5"
+
+	"github.com/margince/margince/backend/internal/modules/activities"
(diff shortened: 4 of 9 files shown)


LarsGradion:
Codex review found two things; both handled.

**Fixed here (c094c43d6): a held thread's unstamped mail could be published.** The predicate read the import row's `verdict_status` as an answer about the thread. It is not: `RecordOutcomeOnThreadTx` stamps at most 500 siblings per transaction, so a conversation longer than one batch keeps rows reading NULL after a `held` verdict has committed — and those rows are exactly the mail the seat asked to keep back. The release now asks `capture_thread_verdict` directly, refusing `held`, `unsure`, `held_by_owner` and `pending`.

`TestAHeldThreadsUnstampedMessagesAreNotPublishedByASenderVerdict` covers it, and fails with the clause removed — it reports the message as `workspace`, which is the disclosure.

The reconciling pass's scan now derives from the same predicate constant with the sender substituted, instead of repeating the clauses. A selector that admits what the writer refuses finds the same sender every tick and moves nothing.

**Filed as #5034: two recompute paths take activity row locks in different orders.** `confidentialityverdict.go` recomputes its anchor before its siblings, which are ordered by `occurred_at`; this branch sorts its batch by id. The two can cycle on a shared thread. Postgres aborts one transaction, so it is a retryable failed pass rather than a wrong audience, and the fix belongs in the confidentiality path this PR does not otherwise touch — best done as one sorted helper every writer of `activity.audience` shares.

Both integration lanes re-run green after the fix.

PR #5036, opened 2026-09-08
fix(people): a contact an agent types in belongs to the workspace

### LINKED ISSUE #5010: A person created through MCP is visibility=owner, so no colleague can see them
## What happens

A contact created through an MCP tool call lands with `visibility = 'owner'`. Nobody except the creating seat can see it — not a teammate, not a team lead. A rep is told the contact was created and then cannot find it.

Observed in a demo rehearsal: `Lucy (Nhi) Vo` was created via MCP, and `lena.fischer@demo.test` could not see her at all.

## The data

Every person on staging, grouped by how it was created:

| source | visibility | count |
| --- | --- | --- |
| seed:demo | workspace | 758 |
| capture_counterparty_verdict | workspace | 97 |
| seed:invented | workspace | 14 |
| **manual** | **owner** | **3** |
| capture_counterparty_verdict | owner | 2 |

Every `manual`-source person — which is what an MCP write produces, since MCP goes through the same API a human does — is `owner`. Every capture-created person is `workspace`.

## Why it matters

- A contact nobody else can see is not in the CRM in any useful sense. The record exists and the team cannot act on it.
- It is silent: the creating seat sees the contact fine, so the person who made it has no reason to suspect anything. Only a colleague discovers the gap, and they discover it as "the contact isn't there".
- An agent creating records on a rep's behalf is exactly the workflow where the result is expected to be shared. A contact captured from mail is `workspace`; the same contact typed by an assistant is not.

## What needs deciding

Whether the default is wrong, or whether the MCP tool should be passing a visibility it currently omits. Worth checking both:

- what `createPerson` defaults `visibility` to when the request body omits it, and whether the UI path sends it explicitly where the MCP path does not;
- whether `owner` is the intended default for any API caller, or only for a specific case that MCP is falling into by accident.

If the default is deliberate, the MCP tool surface needs to send `workspace` explicitly. If it is not, the fix belongs at the create handler and covers every caller at once.

## Also worth a pass

Two `capture_counterparty_verdict` people are also `owner`, against 97 that are `workspace`. Probably a different cause (a personal or held verdict), but the same symptom, and worth confirming those two are deliberate rather than the same bug reached by another route.

## Related

Distinct from #4958 (MCP creates an organization whose sole domain is not primary, so it is never enriched). Same origin — the MCP create path — different field and different consequence. Both surfaced in the same demo rehearsal, and a review of what the MCP write path sets versus what the UI sets would likely catch both.

LarsGradion:
Closes #5010.

A create through the tool surface reaches the same `people.Store.CreatePerson` the UI and the REST API reach. The only thing that distinguished them was the principal, and it did: `bornOwnerScoped` returned true for `PrincipalAgent`, so an agent-created contact was born `visibility = 'owner'` and no colleague could see it.

Observed in a demo rehearsal — a contact was created via MCP and a teammate could not find it. On staging every `manual`-source person is `owner`; every capture-created one is `workspace`.

## What changed

`bornOwnerScoped` is **deleted** rather than inverted. It had exactly one true case and exactly one caller, so inverting it would leave an always-false question standing in front of a decision that no longer asks anything. `createPersonInTx` now writes `visibilityWorkspace` directly.

`visibilityFor` stays — the capture ensurer still calls it with an explicit `OwnerScoped`.

## What still mints an owner-scoped contact

The capture paths, and they do it by SAYING so on the spec rather than by being inferred from who is acting:

- the capture sink, for a sender nothing has judged yet;
- the `advisor` verdict arm, deliberately, because publishing that record announces that the founder has a lawyer.

That is the property worth keeping, and it survives the actor-type rule going away — which is what the third test below pins.

## Tests

`backend/internal/modules/people/agentcreatevisibility_integration_test.go`:

1. an agent-created contact is `workspace` — fails with the change reverted;
2. a human-created contact is `workspace` — the control, so a store that published nothing could not pass on the first test alone;
3. a capture-minted contact is still `owner`.

Full `internal/modules/people` lane green; `make check-backend` green, including all four craft roots, rls-store-path and the gates.

## Not in scope

`lead.go` carries the same agent-only carve-out for lead ownership. Different field, different consequence, left alone deliberately.

The two `capture_counterparty_verdict` people on staging that are `owner` are the advisor arm working as intended, not this bug.

---
## Summary by cubic
Closes #5010. Agent-created contacts were owner-scoped and invisible to colleagues; typed creates now use workspace visibility, so teammates can find contacts created through agent tools. Capture paths that explicitly request owner visibility remain private.

**Tests**
- Verifies a colleague can read an agent-created contact through the normal store read path.
- Keeps human-created contacts workspace-visible and covers advisor-verdict privacy end to end.

<sup>Written for commit c814cbdcaf1b57b6901dc4827444dd3380009c0c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/5036?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/internal/modules/people/agentcreatevisibility_integration_test.go
+++ b/backend/internal/modules/people/agentcreatevisibility_integration_test.go
@@ -20,8 +20,7 @@ import (
 	"context"
 	"testing"
 
-	"github.com/jackc/pgx/v5"
-
+	"github.com/margince/margince/backend/internal/platform/database/storekit"
 	"github.com/margince/margince/backend/internal/shared/kernel/ids"
 	"github.com/margince/margince/backend/internal/shared/kernel/principal"
 )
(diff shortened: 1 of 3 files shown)


LarsGradion:
Codex review found two test-quality defects; both fixed in c814cbdca.

**HIGH — the capture-arm test did not test the capture arm.** It called `createPerson` directly with a hand-written spec, so it would have stayed green if a real capture path ever dropped `OwnerScoped`. Removed rather than patched: that case is already covered end to end by `TestAnAdvisorVerdictMakesTheRecordAndKeepsItTheOwnersAlone`, which drives the real verdict engine, and a second hand-built version of it is worse than none.

**MEDIUM — the assertion checked storage, not reachability.** `visibilityOf` is a direct SQL read, so a row stored `workspace` that a colleague's read still filtered out would have passed while the reported symptom stood. Replaced with `TestAColleagueCanReadTheContactAnAgentCreated`, which has the teammate call `GetPerson`. Reverted, it fails with `not found` — the exact thing the rep's colleague saw.

Codex confirmed the rest: `captureensurer.go` and `captureverdictcreate.go` still pass `OwnerScoped` explicitly, `visibilityFor` still has a live caller, and `lead.go`'s carve-out supplies an owner value rather than owner-private visibility, so it does not conflict.

It also raised one hypothesis it could not substantiate: whether any automatic `create_record` flow builds a contact out of private conversational context, which would now be published workspace-wide. It found no such producer. Recording it here rather than filing a speculative issue — if one appears, the answer is a private-create flag on the input, not a rule about who is acting.

PR #5067, opened 2026-09-08
main is red: the attention lane appears with its target switched off

dungbanguyen:
Claiming:

- `TestWithTheTargetOffTheLaneIsAbsentFromThePage` (`backend/internal/compose`, `attentionleads_integration_test.go`)

Only that one. Everything else in the package passes.

```
attentionleads_integration_test.go:165: the queue carries [Nobody Is Counting] with the target switched off
attentionleads_integration_test.go:169: the page publishes a reach row for a source it never read:
    {Considered:1 MoreAvailable:false Shown:1 Source:lead_response}
```

**Verified inherited.** Reproduced at `origin/main` (747389b1f) in a dedicated
worktree. Suspect is https://github.com/margince/margince/pull/5064 — "a lead
owes a reply without a policy", merged 17:37 UTC, which is the rule this test
switches off. The same package passed 3295/0 locally at e105c37e4, the commit
before it.

This is the THIRD unrelated red on main this session. The other two:

- `desktop-macos` — https://github.com/margince/margince/pull/5061
- `TestADepartedSeatDoesNotCostTheOthersTheirMorning` — https://github.com/margince/margince/pull/5063

Cause: unknown so far.


dungbanguyen:
Folded into https://github.com/margince/margince/pull/5063 — closing this claim, which was mine.

Both of `main`'s current reds are the same failure mode (a deliberate change
updating its own tests and missing a sibling in `internal/compose`), and keeping
them apart had a cost: each PR inherits the other's red through its merge
commit, so neither could show a green `ci` and the pair would have to be merged
past a failing verdict. Together they produce a real one — the whole
`backend/internal/compose` package is 3295 pass / 0 fail with both fixes, and
3294 / 1 with only the first.

#5063 carries the target-off fix, the diagnosis of
https://github.com/margince/margince/pull/5064's retired decision, and the
mutation test for it. Nothing from this claim is dropped.

Issue #683, opened 2026-08-09
linkedin_match.decided is undeliverable: declared entity 'user', emitted with the connection id, routed to the person stream


LarsGradion:
The `linkedin_match.decided` public event is never delivered to any webhook subscriber, and three sources of truth disagree about its subject.

- The contract declares `x-entity-type: user` (`backend/api/public-events.yaml:1223`, generated at `internal/contracts/publicevents_gen.go:1287`).
- The emitter stamps the CONNECTION id as the entity id (`people/linkedinmatchapply.go:156-157`, matching the audit row's `linkedin_connection` entity at `:147`).
- The kernel catalog routes it to the person stream (`shared/kernel/events/catalog.go:90`).

Delivery consequence: `webhooks/deliveryvisibility.go:154-157` treats entity-`user` events as owner-private and admits a subscriber only when `actor.UserID == entityID`. The entity id here is a connection UUID, so the equality can never hold — the event is silently undeliverable, always.

Also in the same contract entry:

- The payload carries only `profile_url_written` — no `verdict` (`public-events.yaml:1236-1246`). The plan of record is to restore `verdict` (`confirmed | rejected`) and bump `x-version` to 2, which the reject-effect work needs anyway (today a rejection emits nothing at all).
- The description self-contradicts: "It carries NO verdict" followed four lines later by "The verdict … are the auditable facts" (`public-events.yaml:1228-1235`), propagated verbatim into the generated doc comment.

Fix: make the declared entity type, the catalog stream, and the emitted entity id agree (and if the entity stays a connection-keyed person-stream event, exempt or route it correctly in `deliveryvisibility.go`); add `verdict`, bump `x-version` to 2, rewrite the description; `make gen`. Pin the agreement with a test — a delivery-visibility test that a decided event actually reaches an eligible subscriber would have caught this.

dungbanguyen:
picking this up

dungbanguyen:
Fixed in #5068 (merged). The decided event now names the member the write was predicated on — the owner `applyLinkedInMatchInTx` binds the update on — rather than the connection id it carried under a declared entity type of `user`. `TestTheDecidedEventCarriesTheMembersOwnID` fails if it goes back.

PR #5068, opened 2026-09-08
fix(people): the decided event names the member, not the connection

dungbanguyen:
Closes the delivery half of #683.

### The defect

`linkedin_match.decided` is in `selfOnlyEvents` (`webhooks/deliveryvisibility.go`), so delivery admits it only when the subscriber **is** the user its entity id names — ADR-0078 §8b, a colleague's professional network is theirs. The emitter stamped the **connection** id.

A connection uuid can never equal a user id, so the equality could not hold for anybody. The event was silently undeliverable, always, with nothing failing.

The owner is read from the connection row rather than passed in: it is the row's own answer, and it is the only thing available at that point that identifies a member without naming the third party the connection is.

### What I did not change, and why

The ticket frames this as three sources disagreeing and suggests making the declared entity type, the catalog stream and the emitted id agree — with `person` as a candidate subject.

I made the **emitter obey the rule** instead of changing the rule, because "should this event be self-only" is a privacy decision with a decision record behind it, not a defect. `selfOnlyEvents`' own declaration argues it: an admin cannot read whether a colleague connected LinkedIn, and a webhook must not become a bypass. If the product wants the CONTACT to be the subject — which is defensible, since the handle lands on a record its viewers can already read — that is a ruling that changes the entity type, the emitted id and the self-only list together, and it belongs to whoever owns §8b.

Two corrections to the ticket's reading, found while doing this:

- **The catalog's `personStreamEntity` is a STREAM, not an entity type.** Streams are coarser than subjects — `commission.accrued` has entity `commission` and rides the `deal` stream, and forty-odd events do the same. I wrote a parity gate between the two before realising it compares different things, and deleted it. So the contract and the catalog were never in disagreement.
- **The operative rule is `selfOnlyEvents`, not the entity-`user` branch** the ticket cites at `deliveryvisibility.go:154-157`. Self-only is checked *before* the entity switch, so it decides regardless of entity type — which is why changing the entity type alone (my first attempt) left the event just as undeliverable.

### Also left open

The rest of #683: adding `verdict`, bumping `x-version` to 2, and the self-contradicting description. Those belong with the reject-effect work (#678's mechanism), which does not exist — and the description's "It carries NO verdict" is *true today*, so rewriting it before the verdict exists would make it wrong in the other direction. #683 stays open for that half.

### Held

- `TestTheDecidedEventCarriesTheMembersOwnID` reads the emitted envelope's entity id after a real apply. Asserted on the outbox row because what was wrong is the id this write puts on the wire, and that is a fact about this transaction.
- The webhooks self-only unit test covered two of the three events in that map and left this one out — which is how the map could say self-only while nothing asked what "self" was. All three now.

Mutation-checked by restoring the connection id: the case fails naming exactly the defect. `make check-backend` and `make craft-static` green.

## Summary by CodeRabbit

- **Bug Fixes**
  - LinkedIn match decision events now use the connection owner’s user ID, ensuring they are delivered only to the appropriate user.
  - Improved self-only visibility handling for LinkedIn match decision events.

- **Tests**
  - Added coverage verifying that match decision events reference the member rather than the connection.

### DIFF
--- a/backend/internal/modules/people/linkedinmatchapply.go
+++ b/backend/internal/modules/people/linkedinmatchapply.go
@@ -256,7 +256,7 @@ func applyLinkedInMatchInTx(ctx context.Context, tx pgx.Tx, connectionID, ownerI
 	if err != nil {
 		return err
 	}
-	return auditLinkedInMatch(ctx, tx, connectionID, personID, matchImages(wasStatus, wasPerson, personID), wrote)
+	return auditLinkedInMatch(ctx, tx, connectionID, ownerID, personID, matchImages(wasStatus, wasPerson, personID), wrote)
 }
 
 // matchImagePair is the connection's own columns on either side of a confirmed
@@ -285,7 +285,7 @@ func matchImages(wasStatus string, wasPerson *ids.UUID, personID ids.UUID) match
 // records the link; a handle that reached the contact is a second mutation of a
 // second entity and takes its own audit and its own person.updated, so a trace
 // consumer resolves each event to an audit of the entity it describes.
-func auditLinkedInMatch(ctx context.Context, tx pgx.Tx, connectionID, personID ids.UUID, images matchImagePair, wroteURL bool) error {
+func auditLinkedInMatch(ctx context.Context, tx pgx.Tx, connectionID, ownerID, personID ids.UUID, images matchImagePair, wroteURL bool) error {
 	// Whether the profile URL reached the contact is context ABOUT this
 	// decision, not a column on the connection, so it rides the evidence
 	// column rather than the images field history projects.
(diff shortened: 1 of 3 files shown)


dungbanguyen:
Rebased onto main after #5059 landed. The two changes met in `linkedinmatchapply.go`: #5059 gave `applyLinkedInMatchInTx` a resolved `ownerID` — bound on by the update, and read off the connection row when a legacy proposal carries none — so this change now passes THAT owner to `auditLinkedInMatch` instead of adding a second read of its own. The event names the same member the write was predicated on rather than a second look's answer, and the `connectionOwner` helper this PR introduced is gone with it. Mutation re-checked on the rebased tree: emitting `connectionID` again fails `TestTheDecidedEventCarriesTheMembersOwnID`.

PR #5100, opened 2026-09-09
fix(capture): a cancelled or declined calendar event closes the meeting it already captured

LarsGradion:
## What was wrong

A meeting reaches the timeline as booked, and every surface that draws a day reads it that way — the schedule rail, the brief, the awaiting-outcome lane. When the organizer called the meeting off, or the seat whose calendar it is declined the invitation, capture dropped the event and wrote nothing. The row stayed booked forever and the meeting kept its place on Today's schedule.

Dropping is the right answer for an event that was never captured. It is the wrong one for an event already on the timeline: a provider stops listing an event once it is off, so the pull that carries the cancellation is the only one that will ever mention it again. There was no later pass to correct the row.

Reported from a real calendar: a cancelled "Consulting Monthly" was still being drawn on Today's schedule.

## The declined case was invisible

Two facts now reach the same outcome. A cancelled event is the organizer calling it off, which both providers already stated and `meetingmap` already read.

A DECLINED one is new, and it is the case a reader actually sees. Google leaves the event `confirmed` and merely strikes it through in the UI, so nothing about the event itself says it is off — only the connected account's own attendee entry does, and that entry was never read. Google's `self` flag is the authority, because an invitation can reach an account under an alias that matches no address we hold; the owner's address is the fallback for a payload carrying no flag. Graph states the same answer at the top level. Only a refusal counts: `tentative` is an attendance somebody may yet make, and `needsAction` is an invitation nobody has opened.

## How it works

`meetingmap.Settle` replaces the two-way skip with three answers — capture, cancel, drop. The cancellation is asked **before** the party rules, because those rules decide whether an event is worth capturing and a meeting already on the timeline is past them. Asking them first would drop the cancellation of a meeting whose attendee list shrank after it was booked, which is exactly the stale row this closes.

The write is `activities.CancelCapturedMeetingTx`, injected into the capture Sink by compose through the new optional `connector.MeetingCanceller`. Capture never imports a sibling, and the activities module owns `activity` and its status history — the same shape `WithAudienceRecompute` and `WithParticipantNamer` already travel on. It takes the row under its own lock, sets `meeting_status` to `canceled`, writes the audit row and the outbox event, and records the transition through `recordMeetingTransition`, the one recorder the census gate holds every status writer to.

Three decisions worth naming:

- **The transition is stamped with the meeting's own start, not `now()`.** A pull running days later would otherwise date every cancellation to the sync schedule, and every question about when bookings fall through would answer with the cron.
- **An archived meeting is left alone.** Somebody retired that row deliberately; a calendar sync is not the caller that moves its status back under them.
- **A key naming no captured meeting writes nothing and succeeds.** That is the ordinary case — most cancelled events were never worth capturing, being internal or solo — and failing there would stop a whole calendar pull.

Idempotent in both directions, so a resynced calendar costs no second audit row, no second event and no second transition.

## Tests

Unit, over the shared rules and each vendor's decode:

- `Settle` returns cancel / drop / capture for each shape, including a cancellation outranking the party rules.
- `CaptureOne` reaches the cancel verb, the upsert verb, or neither; a fault stops the pull, a skip does not; a Sink that cannot cancel still syncs.
- Each provider's RSVP: declined via `self`, declined under an alias, declined with no flag, a *guest* declining (not the owner), tentative, needsAction, accepted.

Integration, in `compose` because that is where the two modules are joined, seeding **through the real Sink** rather than inserting rows:

- the column, the transition (status, scheduled start, idempotency key), cancelling three times writing once, an uncaptured key, an archived row, and a Sink composed without the seam;
- and the one that closes the report: **a cancelled meeting leaves Today's schedule**, read through the real `attentionMeetings` lane.

Every guard was mutation-checked — reverting the cancel wiring, the ordering, the archived predicate, the idempotency short-circuit and the decline match each fails the test that claims to hold it.

`make check-go` green, `make craft-static` clean, `make check-gates` green, and the full `compose`, `capture` and `activities` integration packages pass.

## Not in this change

Internal meetings still never reach the timeline at all — capture skips a meeting with no external party, so an all-colleague meeting was never synced and is not affected by any of this. Whether Today's schedule should show a rep's whole calendar day rather than only captured customer touches is a separate product question.

---
## Summary by cubic
Cancelled or declined calendar events now close the meeting already captured on the timeline, instead of leaving it booked forever. A one-off pass also closes meetings captured before this behavior existed, re-reading each stored raw original.

**Details**

- Cancellation is decided before the party rules, so a meeting whose attendee list shrank after booking still gets closed.
- The backfill closes owner-declined meetings and leaves guest-declined ones; a recorded `held` or `no_show` outcome survives later syncs.
- Google's `self` attendee flag settles alone, with the owner's address as fallback; Graph's top-level `responseStatus` feeds the same rule, and only explicit `declined` counts.
- The write is idempotent, records an audit row, outbox event, and history transition stamped with the meeting's own start; archived meetings are left alone.
- A key naming no captured meeting writes nothing and succeeds; a Sink without the seam also succeeds. The seam is compile-time asserted and bound to its own provider.
- A census gate holds both connectors to listing occurrences, not series masters, so one decline can't close a whole recurring meeting.
- A marker table records each backfilled meeting so the pass drains rather than re-reads a workspace on every tick.

<sup>Written for commit d7a4cd8bfd67f88e3a4b21765e1e8a1a6f48efd5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/5100?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Calendar events canceled by the organizer or declined by the connected account are now recognized automatically.
  - Supported for Google Calendar and Microsoft Graph calendar integrations.
  - Previously captured meetings are marked canceled and removed from the day’s schedule.
  - Stored calendar events are rechecked to identify cancellations that occurred before RSVP details were available.
  - Repeated cancellation updates are handled safely without creating duplicate activity or status records.
  - Meetings that were never captured and archived or completed meetings remain unaffected.


LarsGradion:
@coderabbitai review

Both inline findings from the first review are fixed in `5d05471a8`, and each is covered by a test that fails against the previous behaviour:

**`meetingcancel.go` — only an unanswered meeting is cancellable.** `cancellableFrom` now admits exactly `NULL` and `booked`. `held`, `no_show` and `canceled` are answers and keep what they say. This was a real data-loss path: an organizer tidying up a past event would have replaced a human-recorded outcome with a cancellation, deleting the only record the meeting took place. Held by `TestAnAnsweredMeetingKeepsItsOutcome` (both arms); reverting the guard overwrites `held` and `no_show` to `canceled`.

**`gcal/event.go` — the `self` flag settles the RSVP alone.** The entry Google marks `self` is selected first and its answer is final, declined or not; the owner's address is consulted only when no entry carries the flag. The doc comment already claimed this ordering — the code was ORing the two, so an event carrying the owner's address twice could let a stray declined entry overrule the account's own acceptance. Held by the `self accepted, a second entry on the owner's address declined` case; reverting the ordering reads it as declined.

Two further findings came out of a separate review pass and are in `20f0d8776`: a `var _ connector.MeetingCanceller = (*Sink)(nil)` assertion (without it, a signature change on `CancelMeeting` silently turned every cancellation back into a no-op, visible only in the integration lane), and a provenance check on `CancelMeeting` so a connector can only cancel its own provider's meetings.

LarsGradion:
## Added: the one-off pass that clears the meetings already stale

The fix above closes cancellations from now on. It does **not** touch meetings captured before it — those were called off during a pull that finished long ago, and no future sync will mention them again. `d7a4cd8bf` adds the pass that clears them.

### Measured, not assumed

Checked against the staging installation before writing anything:

| | count |
|---|---|
| captured Google meetings | 467 (every one `meeting_status` NULL) |
| future-dated, so drawn on a schedule | 397 |
| stored payloads carrying a decline nobody read | 12 |
| …where the **calendar owner** declined → genuinely stale | **7** |
| …where only a **guest** declined → correctly still on | 5 |

One of the seven is `Consulting Monthly (with Lars J.)` dated today — the exact meeting from the bug report, RSVP `lars@gradion.com=declined(self)` with twelve colleagues accepted. The five guest-declines are why "cancel on any decline" would have been wrong: that would take meetings off a rep's schedule that they are actually attending.

### How it works

The evidence was always here — `raw_capture` stores each meeting's own payload verbatim and the RSVP was in it; nothing looked. So the pass re-reads what we already hold rather than asking the provider, which also means it works for a calendar since disconnected.

It asks `meetingmap` the **same** question live capture asks, through a new `SettlementOf` over stored bytes. A pass that re-derived "is this meeting off" would be a second answer to a question that package owns, and the two would part company the first time a provider changed how it says no.

It reuses the `drainStoredOriginals` harness the participant replay and attendee repair already run on — now generic in its candidate, because each pass needs different facts about the row it judges and reading the extras back per row would be another read of `activity` per pass. This one carries the natural key and the start out of the query that found the row, so it touches `activity` exactly once.

Four outcomes, each a different fact: `closed`, `still_on`, `answered` (somebody already reported held/no_show and their answer stands), `unreadable`. **Every** judged meeting is marked, including the ones left alone — that is what makes it drain rather than re-read a workspace of still-on meetings on every tick.

### Tests

Four integration tests over the real Sink, seeded through it rather than hand-inserted: the owner-declined close, the guest-declined refusal, the answered outcome kept, and the drain (six meetings settled, second pass settles zero). Mutation-checked — removing the settlement guard cancels the guest-declined meeting and the test catches it.

Local: `check-go`, `craft-static`, and the full `compose`, `capture` and `migrations` integration lanes all green. Rebased onto `main` after #5101; migration number re-checked and still the highest.

LarsGradion:
## Dry run against staging: 7 closed, 460 untouched, 0 unreadable

Before merging a pass that edits historical rows, I ran its own logic as a read-only query against the staging database — the candidate selection, the owner resolution and the settlement judgement, exactly as the Go code does them.

**Predicted outcome over all 467 captured Google meetings:**

| outcome | count |
|---|---|
| `still_on` — left exactly as they are | 460 |
| `closed` — the calendar owner had declined | **7** |
| `unreadable` | 0 |
| `answered` | 0 |

The seven it closes:

```
2026-07-13  Rick Yvanovich (TRG Talk x Digital podcast…
2026-07-26  17:00 Uhr Dinner Abend mit Lars und Oil…
2026-08-03  Lars Jankowsky GRADION @ Speed Kit
2026-08-04  Update
2026-08-19  GRADION <> Speed Kit Partnership
2026-08-31  API Layer
2026-09-09  Consulting Monthly (with Lars J.)     ← the reported bug
```

All seven belong to one calendar owner and every one currently reads as booked. The five payloads where only a **guest** declined are correctly among the 460 left alone.

### One thing this run checked that the unit tests could not

Staging holds **two** connected gcal calendars. My candidate query resolves the owner from `captured_by`, with a fallback for rows whose stamp carries no user id — and that fallback requires exactly one connection per provider, so with two it would resolve nothing and settle every meeting as `unreadable`.

It does not fire here: both connections stamp their user id (466 rows for one seat, 1 for the other), so the primary lookup matches and the ambiguous branch is never reached. Verified by running the production owner-resolution expression itself — **0 unresolved of 467**. Worth stating explicitly, because a workspace whose rows predate per-user stamping *would* hit that branch, and the honest answer there is `unreadable` rather than a guess between two calendars.

No writes were made to staging; every query above is a SELECT.

PR #5103, opened 2026-09-09
fix(consent): the transactional purpose key no longer authorizes a send by itself

LarsGradion:
Two holes on the send path, both of them the engine recording an objection and sending anyway. First slice of the consent improvement concept (R1 and R16).

## The transactional class allowed unconditionally

`resolutionForClass` has said *not supported, `legacy_transactional_unevidenced`* since the escape hatch was closed. But `decideResolved` then handed the unsupported claim to `legacyVerdictFor`, which asks `VerdictForPerson`, whose `ClassTransactional` arm allows on the contract basis without consulting anything. Any message naming the transactional purpose key became operational on nothing but the key — no invoice, no contract, no thread, no deal.

The lead arm closed this already (`TestALeadTakesNoAuthorityFromATransactionalPurpose`); persons were the remaining half.

`VerdictForPerson` is deliberately untouched. The guard endpoint and the legacy gate ask it about a **person**, and a person with an invoice coming really does have a contract basis. `legacyVerdictFor` asks about a **message**, and only runs when nothing was found supporting this one.

## An absent category meant observe

A category nobody remembered to name shipped recorded-but-not-binding, with nothing saying so — a silent hole shaped exactly like whatever was forgotten.

- Absent now **enforces**.
- A **partial** map is refused at the door, naming the categories that have no mode.
- An **empty** map still validates: empty means unconfigured and resolves to enforce, and refusing it would block saving any other setting on that surface.

## Tests

Added `TestTheLegacyTransactionalPurposeDoesNotAllowTheSend` — the existing test asserted the *resolution* was unsupported and stopped there, which is exactly the gap that let the verdict disagree with it. New `TestAPartialMapIsRefused`; three posture tests updated to the new default.

Unchanged and green: `TestAbsoluteDenialsSurviveEveryMode`, `TestOnlyANonAbsoluteMachineReadingCanBeOverruled`, `TestNoGenericTransactionalCategoryExists`.

`make check-backend` green. Integration lanes for consent, activities and compose green.

## Rollout note

This refuses mail that sends today where a caller relied on the transactional key with no evidence behind it. `DisagreementReport` already reports that shape. Slice B1 of the plan gives those sends a review path to recover through; until then they refuse.

---
## Summary by cubic
Fixes two send-path holes where the engine recorded an objection and sent anyway, and carries staging evidence through to transmit so the fixes don't park mail that was legitimately authorized. The transactional purpose key no longer authorizes a send by itself, and a category absent from the authorization modes map now enforces instead of observing.

- The legacy verdict for an unsupported transactional claim now returns review with reason `legacy_transactional_unevidenced` instead of deferring to `VerdictForPerson`, which allowed the transactional class unconditionally.
- `ModeFor` now defaults an unnamed category to `enforce`; a partial modes map is refused at save naming the categories with no mode, while an empty map still validates as unconfigured and enforces.
- Staging now writes the evidence IDs a message was staged on to the decision row, and transmit re-asks those same questions; rows written before this change decode to no evidence and take the path they always took, and unreadable rows yield no evidence rather than erroring.
- Only evidence the sender was shown to hold is written down: arms that answer before the readability check (thread, live-deal) carry nothing forward, so an unchecked id can't reach transmit, where the system principal could authorize on it.

This refuses mail that sends today where a caller relied on the transactional key with no invoice, contract, thread, or deal behind it. `DisagreementReport` already reports that shape.

<sup>Written for commit e48b89d8cfbf9d783bbd02bc2d6833135953bcac. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/5103?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Unevidenced transactional communications are now sent for review instead of being authorized automatically.
  - Consent evidence is preserved from staging through transmission, allowing valid evidence-based authorization to remain effective.
  - Malformed or missing evidence safely falls back to legacy authorization handling.

- **Consent Settings**
  - Unconfigured or unrecognized authorization modes now default to enforcement rather than observation.
  - Partially configured authorization-mode settings are rejected to prevent unintended gaps.

### DIFF
--- a/backend/internal/modules/consent/authorizedecide.go
+++ b/backend/internal/modules/consent/authorizedecide.go
@@ -188,6 +188,30 @@ func (g *Gate) legacyVerdictFor(ctx context.Context, tx pgx.Tx, personID, purpos
 	}
 	switch verdict.State {
 	case VerdictAllowed:
+		// THE TRANSACTIONAL CLASS DOES NOT CARRY ITSELF ON THE SEND PATH.
+		//
+		// VerdictForPerson allows ClassTransactional unconditionally, because
+		// Art 6(1)(b) really does mean the contract is the basis and the guard
+		// endpoint has to say so about a person who has an invoice coming. But
+		// that answer is about a PERSON, and this function is about a MESSAGE:
+		// we are only here because resolveCategory found nothing supporting
+		// this one — no thread, no live deal, no accepted claim — and
+		// resolutionForClass already said so with legacy_transactional_unevidenced.
+		//
+		// Taking the allow anyway is what let any message calling itself
+		// operational become one, on nothing but the purpose key. The lead arm
+		// closed this hole already (TestALeadTakesNoAuthorityFromATransactionalPurpose);
+		// this is the same hole for persons, and the reason code is the one the
+		// resolution had picked before the legacy gate overrode it.
+		//
+		// The guard endpoint and the legacy gate keep VerdictForPerson's answer
+		// untouched: an invoice with real evidence resolves as supported and
+		// returns from decideResolved's supported arm, never reaching here.
+		if purpose.Class == ClassTransactional {
+			d.Verdict = commsauthz.VerdictReview
+			d.ReasonCode = commsauthz.ReasonLegacyTransactionalUnevidenced
+			break
+		}
 		// A basis this call DERIVED is written down before the send relies on
 		// it (Art. 5(2)). The engine is the only authority now, so it is the
 		// only thing left that can make that record: grantedForRecipient used
--- a/backend/internal/modules/consent/authorizedecide.go
+++ b/backend/internal/modules/consent/authorizedecide.go
@@ -96,6 +96,24 @@ func (g *Gate) decideResolved(ctx context.Context, tx pgx.Tx, req commsauthz.Req
 	if err != nil {
 		return commsauthz.Decision{}, err
 	}
+	// ONLY EVIDENCE THIS CALLER WAS SHOWN TO HOLD travels any further.
+	//
+	// refuseUnreadableEvidence runs inside validate, and validate runs only for
+	// arm 3 of resolution — the thread and live-deal arms answer before it. So
+	// a message allowed by one of those two carries evidence ids nobody has
+	// checked this caller may see, and writing them to the decision row would
+	// hand them to the transmit phase. That phase runs under the system
+	// principal, for which auth.Require returns nil, so an unchecked id there
+	// becomes an authorization the sender could never have obtained: name any
+	// invoice in the installation, get allowed by an open deal at staging, and
+	// have the invoice arm allow it at transmit.
+	//
+	// EvidenceChecked is the resolution's own record of having passed that
+	// check, so what reaches the row is the intersection of "named" and
+	// "readable by the person who named it".
+	if res.EvidenceChecked {
+		d.Evidence = req.Evidence
+	}
 	if res.Supported {
 		// The record bears the category out — and the subject may still have
 		// said stop. The evidence arms never read person_consent, so this is
--- a/backend/internal/modules/consent/authorizeevidencecarry.go
+++ b/backend/internal/modules/consent/authorizeevidencecarry.go
@@ -1,0 +1,104 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+package consent
+
+// Carrying the evidence a message was staged on to the phase that sends it.
+//
+// The transmit phase re-asks the whole question, which is the design: a thread
+// can be archived and a deal can close while a delivery waits in the queue, so
+// the answer must be taken again against the record as it is now. But re-asking
+// needs the same INPUTS. stagedRequestFor rebuilt the question from the claimed
+// category and the thread key alone, so a message supported by a named invoice
+// or contract arrived at transmit with a zero id, found no document, and fell
+// through to whatever the legacy purpose key said.
+//
+// While the transactional class allowed unconditionally that fall-through hid
+// the loss: the invoice went out on the purpose key rather than on the invoice,
+// and the two answers agreed by accident. They no longer do.
+//
+// WHAT IS CARRIED IS THE POINTER, NEVER THE VERDICT. An id is a question to ask
+// again — is this invoice still live, is this person still employed there — and
+// every one of those checks runs at transmit exactly as it ran at staging. The
+// resolution itself is deliberately not carried, for the reason stagedClaims
+// gives: it would let a message ride an answer the record no longer supports.
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/margince/margince/backend/internal/shared/kernel/ids"
+	"github.com/margince/margince/backend/internal/shared/ports/commsauthz"
+)
+
+// stagedEvidence is the wire shape of communication_decision.evidence.
+//
+// Written with explicit json tags and omitempty so a row carries only the ids
+// that were actually named: an absent key and a zero uuid mean the same thing
+// to the reader, and the sparse form is what makes the column readable to a
+// human looking at a decision row.
+type stagedEvidence struct {
+	ActivityID     *ids.UUID `json:"activity_id,omitempty"`
+	DealID         *ids.UUID `json:"deal_id,omitempty"`
+	InvoiceID      *ids.UUID `json:"invoice_id,omitempty"`
+	ContractID     *ids.UUID `json:"contract_id,omitempty"`
+	ConsentEventID *ids.UUID `json:"consent_event_id,omitempty"`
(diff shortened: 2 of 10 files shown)


dungbanguyen:
**This landed red and `main` is still red.** Reproduced on pristine `origin/main` (c57b3c835), not on a branch:

```
testmailbox_integration_test.go:229: send-email under "transactional" → 409 "consent_not_granted", want 202
--- FAIL: TestTestMailboxFullLoop
```

**25 integration tests** in `internal/compose/integration` fail, across the send, schedule, capture, preference-centre and MCP paths. `main-health` has not run since this merged (last run `a5e4b17aa`, several merges back), which is why nothing has reported it.

The change itself reads correct to me and the rollout note is honest — *"This refuses mail that sends today where a caller relied on the transactional key with no evidence behind it."* What did not travel with it was the fixtures, which is a blast radius rather than a behaviour question. Two things make it more than fixture maintenance, so I am reporting rather than fixing forward.

## The premise test cannot be fixture-tweaked

`TestCorrespondenceAndTransactionalAreNotConsentGated` asserts `transactional → 202` with the reason *"the contract is the basis, not consent"*. That premise is exactly what this change overturns, so it needs rewriting to the new contract, not a new purpose key. Finding it in the fallout rather than in the change is the tell that the radius was never enumerated.

## Substituting `business_correspondence` uncovers a product bug

`resolutionForClass` leaves `ClassTransactional` permanently unsupported on the send path, so **no fixture can make a transactional send succeed** — every one of the 25 has to move to another purpose. The obvious substitute is `business_correspondence`, and with an inbound activity seeded as the qualifying event it works for the immediate sends.

The scheduled ones then fail differently:

```
comms_scheduled_send: firing 01a0858f-…: send: public base URL is not configured
```

Because `newScheduledSendWorker` composes `sendStore(pool, SendPath{})` — an **empty** `SendPath`, so `PublicBaseURL` is `""`. That is not the test driver: `jobs_capture.go:103` composes the production worker the same way. So a scheduled send of any message carrying an unsubscribe link refuses, in production, and has presumably always done so — invisible precisely because every scheduled-send fixture named `transactional`, which needs no link.

I have filed that separately; it predates this change and should not be bundled into it.

## Where that leaves the fix

Moving the fixtures needs the scheduled-send worker to carry a public origin first, or those tests have nowhere to go. Both are in your subsystem and the second has a wiring decision in it, so I have left `main` for you rather than reverting a security fix or landing a half-migration. Happy to take either half if you would rather hand it over — say which.

The 25:

`TestAChannelSendMayClaimAnOrdinaryCategory` `TestAConfirmedReceiptCarriesTheScheduledSendToSent` `TestAFlooredAccountSendStagesAndOnlyLeavesOnceApproved` `TestAGmailRewrittenIdentityStillYieldsOneActivityAndOneReplyTarget` `TestAHumansAccountStartedSendLeavesWithoutAnApproval` `TestAnAgentScheduledSendFiresUnderTheAgentThatScheduledIt` `TestAnAgentWithNoHumanBehindItStillRecordsWhichAgent` `TestAnAllowedSendRecordsItsStagingDecision` `TestAnEchoCapturedBeforeTransmitIsAbsorbedByTheReceiptItself` `TestAScheduledMessageWritesNothingUntilItFires` `TestAScheduledReplyFilesItselfUnderWhatTheComposerNamed` `TestASenderDowngradedToAReadSeatParksAStagedDelivery` `TestCapturedCopyOfASentEmailCollapsesOntoTheSameActivity` `TestCorrespondenceAndTransactionalAreNotConsentGated` `TestDeactivatingTheSenderParksAStagedDelivery` `TestFilteringByStatusFindsTheStateTheListActuallyShows` `TestPreferenceCenterOneClickUnsubscribe` `TestSendMessageAcceptsAHumanCallerWithoutAToken` `TestSendMessageAcceptsASendScopedPassportWithoutAToken` `TestSendMessageAcceptsWithNoGoogleAppConfiguredButABotBound` `TestSendMessageMCPLoopSendsOnASendScopedPassportAgainstRealPostgres` `TestSendProceedsOnceTheMailboxHoldsTheSendGrant` `TestSentMessageLandsAsAnOutboundActivity` `TestTestMailboxFullLoop` `TestTwoTimersFiringTheSameMessageSendItOnce`

dungbanguyen:
Update: I took the migration as far as it goes without a decision from you. **25 → 9.** The remaining nine split into one mechanical group and one that is a genuine contradiction.

## Done, and independently correct

- **The scheduled lane (5 tests).** Fixed by giving the scheduled-send worker the public origin it never had — #5153. `JobsConfig` gains a `SendOrigin` (base URL + posture, as a pair, because netguard needs both), `cmd/worker` populates it from `--public-base-url`, and `DriveScheduledSendForTest` carries it so the harness stops assembling a worker production does not have.
- **The immediate lane (11 tests).** Reseated on `business_correspondence` with an inbound activity seeded as the qualifying event, and the purpose named once as a `sendPurpose` constant so the grant the fixture writes, the withdrawal a case drives, and the purpose the send names cannot drift apart. That last one matters: `setTransactionalConsent` was withdrawing a grant the send would no longer have claimed, so the case would have read as proving a withdrawal stopped a send it never touched.

## Mechanical, and I can finish it

`accountsend_agent` and `scheduledsendagent` compose envs with no public base URL, so a link-carrying send refuses there. Same shape as #5153, one harness up.

## The contradiction, which is yours

`TestPreferenceCenterOneClickUnsubscribe` and `TestEveryPreferenceAnswerIsUncacheable` assert that a **transactional send carries no unsubscribe link**. That is still the rule:

```go
// activities/unsubscribe.go
func (u unsubscribeSurface) carries() bool {
    if u.category != "" { return u.category.CarriesUnsubscribe() }   // marketing only
    return u.purposeKey != "" && !lockedPurposeKey(u.purposeKey)     // legacy arm
}
```

These sends resolve **no category**, so they fall to the legacy arm, where `transactional` is the one locked key and therefore the only way to send without a footer.

So after this change `transactional` is simultaneously **the purpose that cannot authorize a send** and **the only purpose key that suppresses the unsubscribe link**. Any substitute I pick sends *with* a footer and fails the assertion; keeping `transactional` refuses the send and fails earlier. There is no fixture that satisfies both, which is why I stopped rather than picking one.

Two ways out that I can see, and the choice is a product one:

1. these messages should resolve a category (a non-marketing one carries no link), and the legacy arm is vestigial for them; or
2. `lockedPurposeKey` should follow the purposes that can still send, in which case the link rule moves with this change rather than after it.

## What I would suggest

Land the #5153 wiring and the 16 fixtures as one restoration — I have them working and can raise it against your branch or my own, whichever you prefer — and decide the link rule separately. Say the word and I will push what I have; I have not opened a PR because a partial fix cannot go green while the other nine are red, and I did not want to guess at the answer above and call it a fix.

dungbanguyen:
The 16 are on `fix/transactional-fixtures` (branch, no PR — it does not go green and should not). Cherry-pick, take the branch, or tell me to raise it and I will.

Two commits' worth in one: the #5153 wiring, which stands alone, and the fixture migration, which does not work without it.

dungbanguyen:
`main-health` has now run on `0f73b5db0` and **failed**, on the same test and the same assertion I reported four hours ago:

```
--- FAIL: TestTestMailboxFullLoop
    send-email under "transactional" → 409, want 202
```

So this is no longer a report from one contributor's branch — it is on the dashboard. Three shards red.

The offer stands: `fix/transactional-fixtures` carries the #5153 wiring and 16 of the 25, verified locally. The last nine need the ruling in my previous comment, which is that `transactional` cannot both be unsendable and remain the only key that suppresses the unsubscribe link.

Meanwhile main has taken four more merges while red.

dungbanguyen:
Retracting the ruling I asked you for — there wasn't one needed, and the fix is up as #5166 with all 25 green.

I said `transactional` was simultaneously unsendable and the only key suppressing the unsubscribe link, so no fixture could satisfy both. The first half holds. The second was me reading one arm as the whole rule: the purpose-key arm of `unsubscribeSurface.carries()` is the **fallback**, reached only when no category was claimed — and the category is a caller's claim, not something the engine derives. A fixture can state a non-marketing one and get a link-less send that still authorizes. Nothing about your change moves.

Sorry for the detour; that cost you a decision you didn't have to make.

#5166 also carries #5153's wiring, which is wider than I filed it: `sendStore(pool, SendPath{})` appears at five sites, two of them production. One stays open there — the approved-held retry through `heldSendActors` — because threading an origin to it crosses the approvals composition and that is not a main restoration.

Your change is untouched. Everything in #5166 is fixtures plus the send-origin wiring.

PR #5105, opened 2026-09-09
The coverage page measures the surface it names, and the lane that checks it now runs

tinnhlt-gra:
The coverage page counted one population while naming another, and the lane that
was supposed to check it had never run.

## The through-line

Four places, one shape: **a number describing something other than what it names,
passing its own checks.**

| Where | What it reported | What it measured |
|---|---|---|
| the coverage page | "tools the assistant is offered" | the all-scope core catalog, not the scope-filtered listing |
| `toolsInBlock` | permitted tools per case | 0 for any `may_call` whose first line is a comment |
| the guards tool | "21 audited, 19 with findings" | five scenarios — the judge died on a session limit and each unread scenario counted as one found wanting |
| the lane's own cost | nothing at all | it billed real tokens and reported none of them |

The page's two populations both held 73, so the totals agreed and a seven-tool
disagreement went unread. The equal count is exactly how it hid.

## The lane had never run

Seeding died on its first relink, before a single scenario. A `call` is with a
PERSON — the company is reached through that person's employer — and migration
`1788000100` restored that rule after `1787570000` withdrew it. The fixture read
the withdrawal and not the restoration, so it logged three calls and relinked
them onto a company, which the estate answers 422 to. The comment above it stated
the withdrawn rule as current, so code and explanation were wrong together.

They are mails now: `email` is the kind `1788000100` deliberately leaves
unrestricted, for a reason that fits this fixture exactly.

**So "7 of 21 cases have a committed run" was describing a lane that could not
start.** It runs now, and this PR carries the results.

## What the sweep found

Real findings, reported rather than smoothed away:

- **`case20`** — the assistant never calls `compose_analytics_report`, and never
  saves the query it was asked to put in the board pack. 0/3.
- **`case6`** — asked about two sources that disagree on a date, it invents a
  second complaint to reconcile them ("raised in September, repeated in
  October") instead of reporting the conflict. 0/3, and **already 0/3 on main
  before this branch touched the criteria**.
- **`case3`** — reports a dry-run preview in the past tense, as work already
  done. 1/3.
- **`case40`** — never calls `disqualify_lead`. 1/3.
- **`case2`** — files a probable duplicate without saying so. 1/3.

## What this changes

- **Surface census** asserts the page's population is the same SET as the served
  one, both directions — a count is what agreed while the sets differed.
- **The page reports per model.** It held one verdict per case, chosen by lowest
  model name; a second model's records would have been published under a heading
  naming the case, not the driver. Driven/never-driven stays whole-page, because
  what a case REQUIRES is the scenario's property, not the driver's.
- **The page says who judged it.** The semantic half of every criterion is
  settled by a model, so a pass rate carries the judge's accuracy whether or not
  anybody prints it.
- **The lane reports its own cost**, and says `measured/expected` so a sweep
  whose transcripts went missing cannot read as a cheap one. `None` is never
  summed as zero.
- **14 new scenarios** requiring 29 previously-untried tools.
- **An LLM judge for the semantic half.** Five rounds of regex review each found
  false reds AND regressions from the previous round's fix; two paid sweeps
  measured 15% and 20% of correct answers scored as failures. The judge measured
  4.7%.
- **`writes:` per scenario**, validated before the first token is spent, and a
  `WORLD_DIRTY` reset that was `i > 1` — so run 1 of every case, and every run of
  every read-only case, ran against whatever earlier cases had written.

## The judge is Haiku, and what that costs is on the page

Scored against **human-authored fixtures** — not the recorded corpus, which one
model wrote and would hand that model a free hundred per cent:

| Judge | Accuracy | Its own errors |
|---|---:|---:|
| `claude-opus-5` | 98.6% | 2 |
| `claude-haiku-4-5` | 96.4% | 5 |
| `claude-sonnet-5` | 95.7% | 6 |

Haiku is the cheaper judge and the less accurate one; the trade is published
rather than argued. Every criterion each model misread is named on the page.

Switching judges put five fixtures red, and four were criteria that had left
something unstated rather than a judge too weak — all the same shape, a token
appearing read as the requirement being met. The fifth was a wrong EXPECTATION:
`case22/claims-everything-is-synced` opens by reporting the sources unread, which
is exactly what its criterion asks to be told. All three models disagreed with
the test, and the criterion's own text settles it.

## Honest limits, carried not hidden

1. The lane's verdict is now partly non-deterministic — a case can red because
   the judge changed its mind. `pass_at: 2` of 3 absorbs some. Judged worth it
   against 19.6% false reds.
2. Judge and candidate are the same model family (`aicert` calls that
   `self_judged`). An independent grader would be stronger.
3. The corpus proves the judge at one point in time; nothing re-asks it.
4. The probe judges a cleaner string than the lane does.

---
## Summary by cubic
The coverage page counted a different population than the one it named, and the lane that checks it had never actually run. It now measures the scope-filtered surface the assistant is actually served, the lane runs end-to-end, and the page reports per-model results with a judge whose accuracy is published rather than assumed.

**Bug Fixes**
- Seeding files calls as emails instead of relinking them onto companies, so the lane can start.
- The surface census asserts the page's population equals the served surface as a set, in both directions.
- The page counts only git-tracked units, so a unit a lane copies in for its own purposes no longer shifts the tier between CI and local.
- A `may_call` whose first line is a comment no longer reads as empty.
- The runner rebuilds the database whenever the last run dirtied the world.
- The lane reports its cost as `measured/expected`; unmeasured runs are never summed as zero.
- The guards round counts an unread scenario as unmeasured rather than found wanting, and exits non-zero.

**New Features**
- 14 new scenarios requiring 29 previously-untried tools, with `writes:` declarations validated before the first token is spent.
- Semantic criteria are decided by an LLM judge (`claude-haiku-4-5`), scored against human-authored fixtures with all three candidate models' errors named on the page.
- The first real sweep's findings are left standing: `case20` never saves the requested query, `case6` invents a second complaint instead of reporting a date conflict, and `case3`, `case40`, and `case2` each miss a required action.

<sup>Written for commit b6541f1a5e15d6625750e17ecb23e3af7ddeecd1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/5105?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


tinnhlt-gra:
## UAT — driven by hand, recorded

The change here is a report and a lane, not a screen, so the UAT drives the four
things a reader would otherwise have to take on trust. Terminal-only recording;
nothing of the machine's screen is in it.

1. **The page's population, per model.** 75 tools, 36 driven — and three models
   that actually ran the lane, where the page could previously only say "not run".
2. **Who judged it**, scored over the 113 JUDGED fixtures rather than the 138
   suite cases. That inflation was caught in review and is the same defect this
   PR is about.
3. **A judge nobody configured is a stop, not a pass** — exit 2, the harness
   stop, never a green criterion.
4. **An always-yes judge must not turn the suite green** — 42 cases still fail,
   so the criteria are not decorative.
5. **The offline suite** replaying real recorded verdicts, green.
6. **The usage census cannot fail short** — two runs, one unreadable transcript,
   and the missing run is named rather than summed as free.

The one thing to watch for is step 6: `1/2 runs measured  <-- 1 run(s) unmeasured`.
A total that could not say that is how a sweep nobody could read would have
passed as a cheap one.

https://github.com/user-attachments/assets/7a46a4a4-2c0d-431d-9d9c-b0f3538011d5

Issue #2004, opened 2026-08-20
A deal's references are still disclosed by six read paths outside the deals module


tinnhlt-gra:
Parent tracker. #1986 closed this on the deals module's own read paths — `GetDeal`, `ListDeals`, all four mutation responses, and the list's reference filters. The **invariant is broader than the module**, and six other readers still project the same three columns without probing the referenced row.

Found by Codex review on #1986; each file:line below was checked against current source before filing. Two spot-verified in detail (reports and offers) and quoted here.

## The invariant

`deal` is an `identityTable` (`auth/tableclass.go:28-30`): every seat of the workspace reads every deal row. The records a deal POINTS AT are not — `organization` carries capture privacy (`visibility='owner'`, `auth/rowscope.go:94-107`) and `project` keeps own/team row scope. So **anything that projects `deal.organization_id`, `deal.project_id` or `deal.partner_org_id` to a client, or lets a client SELECT deals by one of them, owes the referenced row a visibility probe.**

Reusable primitives added by #1986: `auth.VisibleSubset` (batch, one statement per referenced table per page) and, for filter arms, the `EXISTS` shape in `deals/deal_read.go`'s `referenceFilterClause`.

### 1. Filter preview and filtered/bundle export — `area: reports`
`compose/filterpreview.go:139-178`, `compose/filteredexport.go:169-201,204-260`, `compose/export.go:101-118,245-270`.
All three derive every physical deal column and call `maskedRowSelects`, which applies **role field masks only** — it never runs `unreadableReferences`. The workspace bundle's `readMember` selects every deal column with neither mechanism. Result: a private organization's UUID lands in `deal.csv`/JSON even when the organization member of the same bundle correctly omits it.

### 2. Prebuilt reports — `area: reports`
`compose/report.go:178-245`, `compose/reportsql.go:82-124`, `compose/reporthandlers.go:24-58`.
**Verified:** `open-deals-per-company` declares `fieldOrganizationID: colOrganizationID` in its `dimensions` map and groups by it **by default**. `buildReportWhere` applies only the deal row scope and renders `deals-by-stage`'s `organization_id` filter as a bare `t.organization_id = $n`. Because the deal scope admits every deal, the default report hands out private organization UUIDs directly, and the stage report is the same existence oracle `referenceFilterClause` was written to close.

### 3. `query_workspace` direct field predicate — `area: agents-mcp`
[...]

PR #5132, opened 2026-09-09
fix(deals): the offer document says no more about its buyer than the read does

dungbanguyen:
Closes the `deals` reader of #2004 (reader 4).

## The leak

`PrepareRender` gathers the buyer legal block the PDF prints — display name and, where the record carries one, **legal name**. That block came from the live organization while the offer is a draft and from the frozen `buyer_snapshot` once sent, and neither read asked whether this seat may open the company.

So the API withheld the buyer while the document printed its name — the two halves of one offer disagreeing about the same record, with the more disclosing half being the one that gets **stored and sent**.

An offer is anchored on its **deal**, and every seat of the workspace reads every deal because a deal is customer identity. The organization it names is not: capture privacy makes a company private to the colleague who captured it.

## Premise check

#1986's `withholdUnreadableBuyer` already closed `GetOffer`, `ListDealOffers` and every mutation response. Two of the paths the issue names were already covered:

- `resolveBuyerOrg` gates a **client-supplied** `buyer_org_id` with `EnsureLinkTarget`; the inherited one is what `withholdUnreadableBuyer` catches.
- `GetOffer` discards `visibleOffer`'s copy and re-reads through `readOfferWithLines`, which withholds.

The render was the one path left, and it is the one that produces the artifact.

## Shape

Applied to the **offer** rather than probed inside the block, so the two spellings of one withholding cannot diverge: with the reference gone, the live read has nothing to look up and the frozen snapshot has nothing to return.

The read-back gets its own helper. Withholding against `[]Offer{offer}` mutates a **copy** and leaves the caller's own value carrying the reference — a silent failure, because the path keeps working and the buyer is still there. Two callers spelling that out separately is one of them getting it wrong eventually.

## The seat this needed

The render asks a harder question than the read — `EnsureWritable` on the deal, not `EnsureVisible` — so an own-scoped colleague never reaches the buyer block to be refused it. The seat that would have printed the name is a deal desk operating **workspace-wide over deals with no organization grant at all**, and the fixture had to be built for it. My first attempt used the existing own-scoped fixture and got `permission denied` from the render itself, which proved nothing.

## Verification

- `make check-go`, `make check-gates`, `make craft-static` green
- `make test-it DIR=backend/internal/compose/integration RUN=TestARender` — both pass

Removing the withholding separates the two cases exactly as intended:

```
--- FAIL: TestARenderSaysNoMoreAboutTheBuyerThanTheReadDoes
    the buyer block map[display_name:Meridian Labs organization_id:01a08503-…]
    reaches the document — it carries the company's display name and legal name,
    which is strictly more than the id the API withholds beside it
--- PASS: TestARenderKeepsTheBuyerBlockForASeatThatCanOpenIt
```

The second passing under the mutation is the point: a render that omitted the buyer for **everyone** would pass the leak case perfectly while producing an offer PDF with no buyer on it — not a quieter answer, a broken one.

## One thing worth a reviewer's eye

A seat that cannot open the buyer now renders a document with **no buyer block** rather than being refused. That is consistent with what the same seat's API read shows them, and it is the option that invents no new refusal policy — but a stored PDF missing its buyer is arguably worse than an error, and the alternative (refuse the render, naming the remedy) leaks that a buyer exists at all, which their read says it does not. I took consistency; if the product view is that a document must never be silently incomplete, that is a small follow-up and I will take it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Offer rendering now consistently protects buyer details when the viewer cannot access the buyer’s organization.
  * PDF previews no longer expose the buyer organization or buyer block to unauthorized viewers.
  * Authorized viewers continue to receive complete offer documents, including the buyer block.
  * Offer reading and rendering now apply consistent access controls for buyer information.

### DIFF
--- a/backend/internal/compose/integration/offerreferences_integration_test.go
+++ b/backend/internal/compose/integration/offerreferences_integration_test.go
@@ -46,6 +46,23 @@ var offerDeskCompanyPerms = principal.Permissions{
 	},
 }
 
+// offerDeskWideNoCompanyPerms is the seat the RENDER needs: workspace-wide
+// authority over deals and offers, and no organization grant at all.
+//
+// The render asks a harder question than the read — EnsureWritable on the deal,
+// not EnsureVisible — so an own-scoped colleague never reaches the buyer block
+// to be refused it. A deal desk operating across the workspace without CRM
+// company access does, and it is the seat that would have printed the name.
+var offerDeskWideNoCompanyPerms = principal.Permissions{
+	RoleKeys: []string{"deal_desk"},
+	RowScope: principal.RowScopeAll,
+	Objects: map[string]principal.ObjectGrant{
+		"deal":                  {Create: true, Read: true, Update: true},
+		"offer":                 {Create: true, Read: true, Update: true},
+		"installation_settings": {Read: true},
+	},
+}
+
 // seedOfferOnAPrivateOrg is one sent offer whose buyer is capture-private to a
 // colleague, created and sent by an admin who can see it.
 //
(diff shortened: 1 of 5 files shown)


dungbanguyen:
CI caught two existing render cases, and what they were pinning is worth stating plainly because it changes how this PR should be read.

`offerRenderDeskPerms` — a workspace-wide deal desk — holds **no organization grant**. I probed what that seat already gets today, before any change of mine:

```
PROBE GetOffer          buyer_org_id=<nil>
PROBE CreateOffer resp  buyer_org_id=<nil>
```

So the read path **already** withholds the buyer from that seat, and has since #1986 shipped `VisibleSubset` with `auth.Require(organization, read)` at the top of it. The render was the one path that did not ask, which is why those two cases passed: they asserted the document shows a buyer the same seat's own read does not.

**This PR is not making that policy decision.** The decision was made when `VisibleSubset` shipped; this stops one path from bypassing it. That matters for #2004's reader 6, which is filed as `needs-decision` on the premise that object grants are *never* consulted for references — they are, on every path but this one.

The two cases are about **which** block the render resolves (live while draft, frozen once sent), not about who may see one, so the seat now holds the grant that question assumes and their assertions stand unchanged.

Full `internal/compose/integration` green locally (181s).

The operational note stands: a deal-desk role with no organization grant will stop getting a buyer block on rendered PDFs. That is now the same answer it already gets from the API — but it is a visible change for such a deployment, and worth a reviewer's eye.

PR #5139, opened 2026-09-09
fix(ci): the review-coverage report stops being the strictest gate in the repo

dungbanguyen:
The check is named **"review coverage (reports, never blocks)"** and its header said a red there blocks nothing. Both were false, and it is currently holding two of my pull requests shut.

## It blocks

`ci` is the one required check. But a red check of *any* kind leaves the pull request `BLOCKED`:

```
$ gh pr view 5116 --json statusCheckRollup --jq '... select(name=="ci")'
ci: SUCCESS
$ gh pr checks 5116 | grep fail
review coverage (reports, never blocks)   fail

$ gh pr merge 5116 --squash
X Pull request is not mergeable: the base branch policy prohibits the merge.
To use administrator privileges to immediately merge the pull request, add the `--admin` flag.
```

So a job that failed on a finding was a gate whatever it was called — and the strictest one in the repository, because the only hand that could clear it was the one every contributor is asked not to use.

## And it was unclearable

A review records the commit id it was made against. A rebase rewrites those ids. CodeRabbit tracks **content**, and answers `@coderabbitai review` on a rebased branch with:

> ⚠️ Action not completed
> Already reviewed the last commit. Use `@coderabbitai full review` to rerun a review of the entire changeset.

So the reviewer had read every change, the report said `BEHIND`, and neither of the two remedies it printed moved it — "re-review the range" was declined, and "say on the pull request why it does not need it" is advice the script honours nowhere. **A refusal that names an unreachable remedy is worse than one that names none.**

## What changes

- The job **succeeds whatever it finds**. `exit "$status"` is gone.
- The finding moves to a **sticky pull-request comment**, found by marker so a push edits the last one instead of stacking. That is louder than a green job's step summary — which is the failure the old shape was avoiding, so the comment is not optional decoration.
- The comment **withdraws itself** when coverage closes, because a stale finding that outlives its cause is how a report earns the habit of being ignored.
- Commenting can never fail the job either, including on a token that cannot write.
- The report itself is unchanged, except that it now prints the remedy that works.

## Held, not asserted

`make test-review-coverage` gained two cases over the workflow, mutation-verified:

```
restore `exit "$status"`     → FAIL: the workflow exits with the report's status, so a finding fails the job.
remove the comment marker    → FAIL: the workflow posts no sticky comment, so a finding on a green job has
                                     nowhere a reader would find it.
```

The second exists because "just drop the comment" reads like a free simplification and is not.

## Verification

`make check-go`, `make check-gates`, `make test-review-coverage`, `make ci-doc-parity` all green.

## Not in scope

Whether reviewer-behind should block at all is a branch-protection decision (#2496), and this workflow was never the place to make it. This restores the contract it already claimed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Review coverage findings no longer fail the pull request workflow.
  - Findings are now posted or updated in a persistent pull-request comment for visibility.
  - Reports clarify the affected re-review range and explain how to request a full review after a rebase.
- **Tests**
  - Added checks to verify that review findings remain informational and are published through the pull-request comment.

### DIFF
--- a/.github/workflows/review-coverage.yml
+++ b/.github/workflows/review-coverage.yml
@@ -7,11 +7,24 @@
 # flag. Four landed that way in one afternoon across three sessions; the fourth
 # author had just finished explaining the trap to somebody else.
 #
-# It is NOT a gate. `ci` is the one required check and this job is not it, so a
-# red here blocks nothing. Prevention is a branch-protection decision (#2496)
-# and is not this workflow's to make; what it does is make the fact visible on
-# the pull request, at the moment it becomes true, instead of leaving it to be
-# noticed by whoever reads the review timeline carefully enough.
+# It is NOT a gate, and the job therefore SUCCEEDS whatever it finds. That is
+# load-bearing rather than lenient: `ci` is the one required check, but a red
+# check of any kind leaves the pull request BLOCKED and `gh pr merge` refusing,
+# with administrator override as the only way past — so a job that failed on a
+# finding was a gate no matter what its name said, and the strictest one in the
+# repository, because the only hand that could clear it was the one every
+# contributor is asked not to use.
+#
+# It became unclearable in the ordinary case. A review records the commit id it
+# was made against; a rebase rewrites those ids; and CodeRabbit tracks CONTENT,
+# answering `@coderabbitai review` on a rebased branch with "Already reviewed
+# the last commit" and declining. So the reviewer had read every change, the
+# report said BEHIND, and neither of the remedies it printed moved it.
+#
+# The finding goes to a sticky pull-request comment instead, which is where the
+# reader already is — louder than a green job's step summary, and it withdraws
+# itself when the coverage closes. Prevention is a branch-protection decision
+# (#2496) and is not this workflow's to make.
 #
 # It reads what GitHub records. A review carries the commit it was made
 # against, so this speaks precisely for the reviewers that leave one — and says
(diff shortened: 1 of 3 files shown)


dungbanguyen:
Stronger evidence than the rebase case, found while trying to clear the check on my own PRs.

**CodeRabbit is rate limited**, and has been for over an hour:

> ⚠️ Action not completed — **Review rate limited.** Your included review limit is currently reached under our Fair Usage Limits Policy.

While that holds, no review lands at any head, so `review coverage` is red on **every** open pull request and none of them can merge. `ci` green, one required check, `gh pr merge` refusing.

So the check does not merely misfire on rebases. It hands a third party's quota a veto over merging, and the repository has no way to clear it except waiting for someone else's billing window — or `--admin`, which contributors are asked not to use. A reviewer being temporarily unavailable is exactly the condition an advisory report should survive.

That is the same defect this PR fixes, arriving by a second route, and it is the more common one: two agents working the same repository exhaust that quota routinely.

dungbanguyen:
Both taken — and both were the shape this PR is about, one level up: an assertion satisfiable without the behaviour.

`exit $status` and `exit "\${status}"` are the same instruction as the one I matched literally, so the pattern now takes any spelling. Mutation-verified against all three.

The marker was worse. It survives in a comment or an unused variable while the calls that carry it are gone, so the assertion is now the calls: a POST **and** a PATCH — a run that could only create would stack a finding per push, and one that could only update would never leave the first. Removing them while keeping the marker fails it now.

Thank you; I wrote a census of my own subject that could read short, in a PR whose whole subject is a check reading short.

PR #5141, opened 2026-09-09
The decoders stop describing this program, and a threshold cannot be unset

tinnhlt-gra:
Round-three review findings on #5138, from both reviewers. **Three of these are
defects #5138 itself introduced**, which is the honest headline.

## An astral rune is not automatically unprintable

My new escape arm ran *before* the printability test:

```
  emoji U+1F600      IsPrint=true   → escaped   ✗
  CJK ext-B U+20000  IsPrint=true   → escaped   ✗
  tag char U+E0020   IsPrint=false  → escaped   ✓
```

So `"René 😀"` came back as `"René \U0001f600"`. It also broke the property the
file states out loud: `%q` passes printable astral through, so `echoSafe` and
`QuoteCaller` stopped asking the same question — meaning a token bounded at 80
bytes by `QuoteCaller` could expand tenfold downstream (measured: a refusal going
358 → 466 bytes inside a 512 budget).

**The census could not see it, because it floods with ASCII.** It floods with an
expanding token now — `k` + U+2028 + a tag character — so the next escaping change
that inflates a bounded token fails here.

## The unset-threshold arm was a reachable regression

`compileDerivation` gained an `isNull` predicate carrying a threshold, and
`derivationWhere` switches on `threshold != nil` **before** `isNull` — so the flag
was dead, the threshold arm ran with an empty value, and:

```
  ?isnull=days on projects-gone-quiet
    before: refused at compile time, naming the vocabulary
    after:  compiles, then dies with "takes a whole number of days"
            — advice for a request the caller did not make
```

A threshold is a comparison over a number and cannot be unset, so
`resolveNullPredicate` refuses it and names the keys that *can* be. That needed a
second derived set: naming a vocabulary the resolver goes on to refuse is the
same defect as naming none.

## The helper stole the doc comment above it

`noResourceAt` landed directly beneath `readResource`'s three-paragraph essay on
framing and not-found symmetry, so godoc rendered that essay over a two-line
string helper and left `readResource` undocumented.

## Two decoders on this surface still described this program

```go
Message: "invalid params: " + err.Error()            // resources/read
toolError("malformed tools/call params: " + err.Error())
```

`resources/read` with `{"uri": 5}` answered *"json: cannot unmarshal number into
Go struct field .uri of type string"* — unbounded, unescaped, describing this
program into a transcript whose later prompts the same run reads. `decodeArgs`
has masked exactly this through `httperr.SafeDecodeError` since it was written,
with the reason spelled out; these are the same door and never got the copy. The
withheld words go to the operator's log — and the URI in that log is bounded too,
since a caller who can provoke a provider error could otherwise drive megabytes
per call into it.

## Smaller confirmed findings

- `allowedPredicateNames` is built on the existing `catalogFilterNames` rather
  than spelling "filters and thresholds as one list" a second time — that pairing
  has a documented reason and two copies drift. Its two halves *are* scoped
  differently (`grantedSpec` narrows dimensions, not filters), and it now says so
  rather than reading as though they matched. Confirmed by the security reviewer
  as fails-closed and not a disclosure: the report catalog already publishes
  every filter and threshold name to every caller.
- The census enumerates the refusal `maxUnservedNamed` was added for, which
  nothing measured — watched failing at **1725 bytes** with the truncation
  removed.
- Its claim that agents' own sets are "held where they live" was **false**. It now
  names the two that are held nowhere (the enrich depths, the approval decision)
  rather than implying somebody covers them.
- `resolvePredicate`'s boolean became two named functions over one shared lookup,
  and a derivation refusal quotes the **handle's** argument keys (`by`, `isnull`)
  rather than the plan's, which named arguments a handle caller never sent.

## Verified

`compileDerivation` runs after `auth.Require`, `requireVocabularyGrants` and
`grantedSpec`, so the vocabularies these refusals name are post-authorization —
checked independently by the security reviewer, who also probed `echoSafe`
against 14 malformed and astral inputs (overlong encodings, truncated tails, lone
surrogates, a real U+FFFD, U+E0001/E0020, U+2028, U+202E) and found no raw
unprintable, bidi control or line terminator surviving.

`make lint` 0 issues, `craft-static` 0 findings, full `./gates/` green, every
touched package green. The regression and the new census subject were both
watched failing first.

## Still outstanding, and it wants a gate not a fourth pass

Unchanged from #5138's note: `search/query*.go`, `shared/ports/datasource`
(blocked on the DAG until `QuoteCaller` moves to `shared/kernel`),
`compose/captureverdictask.go`'s `clampToken`, `tools_import_preview.go`'s
unbounded `Guidance`, `signals/store.go`, `csvfields.go`. Four rounds have each
found more writers of this one invariant; the follow-up should move `QuoteCaller`
down, fold the duplicate spellings into it, and add the fitness function that
asserts every exported vocabulary producer appears in a census subject.

---
## Summary by cubic
Fixes three regressions from the prior round — astral escaping no longer mangles printable names like emoji, `isnull` on a threshold is refused at compile time, and `readResource`'s doc comment is back — and masks two decoders that still leaked Go internals into tool results.

**Bug Fixes**
- `echoSafe` escapes only unprintable astral runes, so emoji and CJK extension text pass through intact.
- `isnull` on a threshold is refused at compile time and names only keys that can be unset; previously it compiled and then failed with advice for a request the caller never made.
- `resources/read` and `tools/call` route decode errors through `httperr.SafeDecodeError`, sending `encoding/json` details only to the operator log; the logged URI is bounded via `QuoteCaller`.
- `readResource`'s doc comment is restored; it had been rendered over the `noResourceAt` helper.

**Refactors**
- `resolvePredicate`'s boolean becomes two resolvers over one shared lookup.
- Derivation refusals quote the handle's argument keys (`by`, `isnull`) rather than plan spellings.
- `allowedPredicateNames` derives from `catalogFilterNames` instead of spelling the filter/threshold union twice.
- The census floods with a token that expands under escaping, so the astral regression would have failed it.

<sup>Written for commit d708ee23b4eeb101929e0c12d3ec114f28e44d21. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/5141?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/internal/compose/closedsetrefusalbudget_test.go
+++ b/backend/internal/compose/closedsetrefusalbudget_test.go
@@ -26,7 +26,9 @@ package compose
 // modules/agents, modules/search or shared/ports/datasource — compose is
 // downstream of all three, so importing them here is legal but provoking their
 // refusals from this package is not the same test. Those carry the same
-// obligation and are held where they live; this file does not speak for them.
+// obligation. Two of them are NOT held anywhere — the enrich depths and the
+// approval decision, both agents' own sets — and saying so is better than a
+// sentence implying somebody covers them.
 //
 // IT ASSERTS ON THE CLASSIFIED FAULT, not on err.Error(). The renderer is not
 // the only ceiling: httperr bounds a module-declared fault at MaxFaultText
@@ -35,6 +37,8 @@ package compose
 // Asking httperr.Classify is asking the surface rather than a model of it.
 
 import (
+	"encoding/json"
+	"fmt"
 	"maps"
 	"slices"
 	"strings"
@@ -51,7 +55,14 @@ import (
 // echo leaves nothing for the set.
 const floodedCallerTokenLen = 4000
 
-func floodedCallerToken() string { return strings.Repeat("k", floodedCallerTokenLen) }
+// It is not plain ASCII, and that is the point: a token bounded at 80 bytes by
+// QuoteCaller can still EXPAND downstream, because agents.echoSafe escapes what
+// does not print. An ASCII flood measured the bound and never the expansion, and
+// an astral-escaping regression that pushed a refusal from 358 to 466 bytes went
+// unseen here for exactly that reason.
+func floodedCallerToken() string {
+	return strings.Repeat("k\u2028\U000e0020", floodedCallerTokenLen/12)
+}
 
 // setBearingRefusal is one vocabulary, the refusal that names it, and the way to
 // provoke that refusal with a token of the caller's choosing.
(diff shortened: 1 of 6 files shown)


tinnhlt-gra:
Superseded by #5159, which carries this commit unchanged. Combined into one PR so the three land on one CI run instead of three, per AGENTS.md's one-PR-per-piece-of-work rule.

PR #5152, opened 2026-09-09
fix(consent): account-started sends and approved retries are consent-gated again

tinnhlt-gra:
## The fix

#5103 closed the escape hatch that let the `transactional` purpose key alone authorize a send — the tightening was correct, and it merged with its own integration lane red because it exposed three pre-existing gaps that used to be masked by that same escape hatch:

- **`capture.stampCaptureParticipants`** only ever read `rec.Counterparty.Email`, so a channel (Telegram) counterparty never got an `activity_participant` row at all — a customer's own inbound message could never be found as authorship.
- **`consent.subjectRef`**'s address was always derived from `r.Email`, so even a properly-stamped channel participant had nothing to match against. `recipientSubjectAddress` now derives `ChannelProvider`/`ChannelUserID` from `r.Channel` too, shared by the person and lead decision arms.
- **`deliveryThreadKey`** read `comms_outbound.thread_key`, which the `comms_outbound_shape` CHECK forces `NULL` for every channel delivery — so a channel reply staged fine (202) and then silently **parked** at transmit, the exact "stages clean, parks at send" shape closing the escape hatch was meant to fix, just one phase later. Now reads `activity.thread_key` via a JOIN through `comms_outbound.activity_id` (filtered through `auth.ActivityAvailableClause`, since it reads the activity table directly), which every delivery shape carries. Same issue for deal-resolved (non-thread) messages: `TransmitRequest` never carried `Links` forward, so `deliveryLinks` was added mirroring `deliveryThreadKey`.

## Review response (Fable + Codex, run in parallel)

Both independently caught the same MAJOR issue: my first pass added deal evidence to fixtures to make tests pass, but two of those additions **masked** the very regressions this PR fixes rather than proving them fixed:
- The telegram round-trip test's own added deal evidence meant it would stay green even if the capture/consent channel fix broke — removed; the test now exercises the real fix.
- The shared `setupPreflightIn` fixture's automatic deal silently changed the authorization basis for ~40 unrelated test consumers (from "consented recipient" to "staked on an open deal"), which could mask a different real regression in the consent path itself. Made opt-in via a new `stakeADeal()` method, called explicitly by the tests that actually need real transactional evidence to reach 202.

Also addressed: added and mutation-tested dedicated coverage for the new channel-account SQL arm; consolidated duplicated "create deal + stake" fixture code into `apptest.StakeOnOpenDeal`; removed ticket-number references from comments per the repo's craftsmanship rule; fixed two roster tests that were incidentally broken by the capture fix now correctly stamping the primary counterparty.

The last two hold-outs (`TestCorrespondenceAndTransactionalAreNotConsentGated`, `TestPreferenceCenterOneClickUnsubscribe`) turned out fixable the same way rather than needing a separate PR — the second needed its deal staked *after* its marketing-opt-out assertion specifically, since a live deal answers for any claimed purpose (including a withdrawn one) once it exists.

## Supersedes

Closes #5154 (same root cause, claimed independently, no fix landed there).

## Test plan

- `go build`, `go vet`, `gofmt` clean across the diff; `craft static --strict` clean.
- `make test-it DIR=backend/internal/compose/integration` — fully green, 0 failures (23/23 originally-red tests fixed).
- `make test-it DIR=backend/internal/compose/integration/channels` — all green.
- `make test-it DIR=backend/internal/compose RUN='TestATwoPartyMessageStillNamesNoRoster|TestAUnitsGroupRosterLandsAsParticipants'` — green (roster-test fix).
- `make test-it DIR=backend/internal/modules/consent` — all green, including new mutation-tested channel-arm tests.
- `go test ./backend/internal/modules/consent/... ./backend/internal/modules/capture/...` — all green.
- `backend/gates` (full census/uniqueness/restricted-reader suite) — all green.

## Summary by CodeRabbit

* **New Features**
  * Added support for recognizing participants by channel account identity, improving message threading and consent decisions beyond email matching.
  * Communication authorization now considers linked deal evidence when evaluating eligible sends.
  * Added immutable decision records and suppression handling for outbound communications.

* **Bug Fixes**
  * Improved participant capture for channel-based conversations.
  * Prevented unrelated accounts on the same channel provider from being treated as the same contact.
  * Improved handling of withdrawn consent, closed deals, and scheduled-message authorization.


tinnhlt-gra:
Standing down in favour of this claim (lower number wins) and closing #5154. One
thing from my bisect that may save you time, because it disagrees with the
attribution above.

## The culprit bisects to #5103, not #5142

```
  defb6a71f (#5132)  → ok      the commit before
  5cdc9af23 (#5103)  → FAIL    "fix(consent): the transactional purpose key
                                no longer authorizes a send by itself"
```

Each probe was `make test-it DIR=backend/internal/compose/integration
RUN=TestTestMailboxFullLoop` on a detached checkout with a freshly rebuilt
`margince_test` template. I also ran it at `c57b3c835^` (i.e. *before* #5142) and
it still failed — so #5142 is downstream of the break, not its cause.

Your `legacy_transactional_unevidenced` string is #5103's own mechanism: its
commit message says `resolutionForClass` "has said 'not supported,
legacy_transactional_unevidenced' since the escape hatch was closed", and #5103
is what stopped `decideResolved` handing that unsupported claim to
`legacyVerdictFor`.

## Likely why it is 19 tests

`sendExpectingAcceptance`
(`backend/internal/compose/integration/comms_send_integration_test.go:93`) is one
helper with **nine call sites**, eight of which authorize on nothing but
`consent_purpose: "transactional"`:

```
  comms_send_integration_test.go:238, :382, :410
  messageidentity_integration_test.go:142, :200
  stagingrefusal_integration_test.go:163
  testmailbox_integration_test.go:229      <- the one I claimed
```

So `TestTestMailboxFullLoop` is the same defect as your two, and fixing the
helper probably clears a large share of the 19.

## One dead end, so you skip it

I tried the migration `crm.yaml` documents — adding
`"communication_context": "customer_service"` to that helper — and it **still
answers 409**. A *claimed* category is not a basis: the engine "resolves the one
the evidence actually supports", so these sends need real evidence (an invoice,
a deal, an inbound thread) planted in `preflightEnv`. That is the part I judged a
consent-domain decision rather than a mechanical edit, which is why I handed it
back rather than guessing.

Blocked behind this: #5141 and #5145, both rebased with auto-merge armed.

PR #5155, opened 2026-09-09
feat(deals): a pipeline can be retired, and put back

### LINKED ISSUE #835: No operation archives a pipeline, so listPipelines' include_archived has no producer
## What is wrong

`pipeline` carries `archived_at`, the pipeline read filters on it, the wire type exposes it, and `listPipelines` declares `include_archived` — but **nothing in the contract archives a pipeline**. The operations are `createPipeline`, `updatePipeline`, `getPipeline`, `listPipelines`; the only `UPDATE pipeline` in the tree (`deals/stages.go`) sets name, is_default and position.

So the state the parameter is about is one a deployment's own endpoints cannot produce. It is reachable in data — a fork's custom migration, an ops repair — and #826 binds the parameter so that such a row answers to the dial that names it rather than being invisible to every caller. That is the honest half. The other half is a product decision nobody has taken.

## Why it matters

A pipeline is the one piece of config that cannot be removed. A workspace that reorganized its sales process carries every pipeline it ever had in the picker, forever, and the column that would retire one has no writer. Either:

- **give the state a writer** — an archive operation, with the question that comes with it: what happens to the deals in an archived pipeline, and may the default be archived at all; or
- **retire the parameter upstream** (contract-first: the spec decides and this repo mirrors it), and drop the `archived_at` filtering the reads carry with it.

## Where it stands

`deals.Store.ListPipelines` takes a `storekit.ArchivedFilter` and serves both values through one assembly (#826). Its doc comment states the gap and points here. The integration test ages a row by SQL and says why — the one place in that suite that cannot seed through a real writer, because there is none.

## References

`backend/internal/modules/deals/pipeline.go` (`ListPipelines`, `readPipelineWith`), `backend/api/crm.yaml` `listPipelines`, and the declared-parameter gate in `backend/declaredfilters_test.go` that found it.

dungbanguyen:
Closes #835.

## The gap

`pipeline` carries `archived_at`, the read filters on it, the wire type exposes it and `listPipelines` declares `include_archived` — and **nothing in the contract could produce the state**. A pipeline was the one piece of configuration a workspace could create and never retire, so every reorganisation of a sales process left its old pipelines in the picker permanently.

The issue offers two answers: give the state a writer, or retire the parameter upstream. Retiring the parameter produces a smaller honest surface and answers the wrong question — `archived_at` is on the table because somebody already decided pipelines should be retirable, and undoing that leaves the user problem unsolved.

## Retiring retires a *choice*, not work

A deal on an archived pipeline keeps its stage, its history and its forecast contribution, and its record page still renders the stage it is on. Forcing deals off first would turn retiring a pipeline into a bulk migration, and an operation nobody dares run retires nothing.

It is also how archiving already behaves here: hides a record from day-to-day work, deletes nothing, moves nothing attached.

## One refusal, and a remedy the caller can reach

The **default** cannot be retired while it is the default. An installation with no default has no answer for "where does a new deal go", and meeting that at deal-creation time is worse than being refused here.

Two things about the refusal:

- **It reads inside the row lock**, not from a snapshot taken before it. The default moves by an ordinary update, and a check against a stale read would retire the default a moment after somebody else promoted it.
- **The remedy is reachable with the grants the caller already holds.** Promoting another pipeline is `updatePipeline` under the same `pipeline:update`. The test walks that path — promote, then retire — rather than asserting the message says so. If clearing a refusal took authority the caller lacks, the guard would be a trap rather than a door.

## Reversible

A retirement with no undo is a support ticket, and archiving is reversible everywhere else in the product. `POST /pipelines/{id}/restore` puts it back.

Restore is **idempotent by predicate**: restoring a pipeline that is not archived changes nothing and answers the pipeline, so the retry a caller sends after a lost response reads the same as the call that landed. The audit row is written only on the transition, and it carries the before/after images — a restore has prior state, which is what separates it from a create, and the trail has to be able to say when the pipeline had been retired.

It does **not** make the pipeline default again. Archiving required the default to move elsewhere first, and putting a pipeline back is not a claim about where new deals should go — inferring it would silently demote whichever pipeline took over.

## Two things the gates caught, both right

- **`pipeline.archived` now has its first producer.** The type was published with its schema and no writer — a subscriber could subscribe and hear nothing — and `eventtypeownership_test.go` carried a waiver saying so. That waiver is stale now and is deleted; the contract description says what the event means instead of that nothing emits it.
- **Restore is audit-only, ratified with the reason.** The catalog defines no `pipeline.restored` and the closed-verb law forbids inventing one build-side. Nothing downstream is left stale: what a subscriber acts on is which pipelines are available, and every reader of that asks `listPipelines`, which answers from the column this write clears.

## The suite's one standing exception is gone

`TestThePipelineListAnswersIncludeArchived` aged a row with raw SQL, with a comment saying why — the one place in that suite that could not seed through a real writer. It uses the real one now. Per AGENTS.md, a test that supplies its own version of production proves nothing about production.

## Mutation checks

| mutation | result |
|---|---|
| the default guard removed | the refusal test fails — the default retires |
| archiving also clears the deals' stages | the no-move test fails |

One fixture note worth recording: my first version of the default test asserted the refusal against a fresh workspace, which has **no** default — the guard was never reached and the test would have passed hollow. The `t.Fatal` I put in for exactly that caught it.

## Checks

`make check-backend` green; `make craft-static` 0/0/0; `go test -count=1 ./gates` green; the deals and compose/integration lanes green. Contract, generated Go types and the frontend types regenerated together.

**Note on CI:** main is currently red from #5103 (the `transactional` purpose key no longer authorising a send). If this PR shows send/schedule 409s, that is the cause and not this change.

## Summary by CodeRabbit

* **New Features**
  * Pipelines can now be archived without moving their deals, stages, history, or forecast contributions.
  * Archived pipelines can be restored and will reappear in pipeline lists and selection flows.
  * Archiving emits a `pipeline.archived` event for configured integrations.
  * The default pipeline cannot be archived until another pipeline becomes the default.
  * Archive requests support version checks to prevent conflicting updates.


dungbanguyen:
Fixed. "It leaves the pickers and the new-deal flows" was meant as *departs from* — the behaviour below it is correct — but it reads just as naturally as *leaves them alone*, and those are opposite claims about the endpoint a client is calling. Now: "It drops out of the pickers and the new-deal flows."

dungbanguyen:
Pushed a fix for something the Sonar coverage flag led me to.

The new-code coverage was 61.5% because the two handlers were reached by no test
— the cases here drive `e.Deals.ArchivePipeline`, i.e. the store. Writing the
wire test turned that from a coverage number into a defect: **`archivePipeline`
declares `If-Match`, and `Pipeline` published no `version`.** The handler reads
the header, the store compares it, and the number it wants was one no response
had ever carried, so the only call that could succeed was one omitting it.

`Pipeline` now carries `version`, read in `readPipelineWith` — the single
assembly every pipeline response passes through, rather than at the one route,
since a version on some responses and not others leaves a caller holding a read
that cannot be conditioned on.

`TestRetireAndRestoreAPipelineOverTheWire` covers the statuses, the stale-version
refusal, the accepted one, the retired pipeline leaving the default list, and the
undo answering the pipeline itself (including that it does not make it default
again, and that a second restore is safe).

It is not one route: `VoiceProfile`, `Contract`, `Automation`, `TagDetail` and
`SavedView` have the same shape, and three routes take `If-Match` on paths that
answer no record at all. Filed as #5209 with the census I wrote to find them —
kept out of this PR because landing it would have meant a waiver list for eight
findings across six modules, which is what this tree does not keep.

Issue #827, opened 2026-08-10
Six list operations declare a page shape their handler ignores (sort, cursor, limit)


tinnhlt-gra:
## What is wrong

The declared-parameter gate landed in #826 judges the parameters that decide a page's **membership**, and deliberately leaves out the three that decide its **shape** — `cursor`, `limit`, `sort` (the shared contract components). Six operations declare one of those and never read it:

| operation | handler | ignored |
|---|---|---|
| `listActivities` | `activities/handlers_activity.go` | `sort` |
| `listCustomFields` | `customfields/handlers.go` | `sort` |
| `listPartners` | `people/handlers_relationship.go` | `sort` |
| `listConsentPurposes` | `consent/handlers.go` | `cursor`, `limit` |
| `listRecordGrants` | `identity/handlers_grants.go` | `cursor`, `limit` |
| `listVoiceCorpusSources` | `ai/handlers_voice.go` | `cursor`, `limit` |

(Derived, not hand-listed: every generated `*Params` field typed `*Cursor`/`*Limit`/`*Sort` whose handler neither reads it, nor passes the params value on, nor parses the raw query string itself.)

## Why it matters, and why it is a different question

A dropped **filter** answers a wider question than the one asked; that class is closed. A dropped **page dial** answers the right question in the wrong shape, which is less dangerous but still a claim the surface does not keep:

- `?limit=5` on the last three returns the whole set — the caller sized a page and got a catalog. Two of them are capped server-side and answer `has_more: false`, so nothing is hidden, but nothing is honoured either.
- `?sort=` on the first three is ignored, so a client that offers a sort control shows an order the server never applied.
- `?cursor=` on a list with no paging is accepted and does nothing, which is the shape that misleads a client into an infinite walk over page one.

## The decision each needs

Per operation: implement it, or retire the parameter upstream (contract-first — the spec decides and this repo mirrors it). The catalog lists are the interesting ones: a workspace holds a handful of consent purposes and voice sources, so "capped, unpaged" may well be the right product answer — in which case `cursor`/`limit` should not be declared on them.

## When it is fixed

Extend the gate in `backend/declaredfilters_test.go` to judge the paging types too (drop `pagingParameterTypes` and the paragraph in its header that scopes them out). The walk already finds these sites; it is only told not to report them.

## References

Found by the derived census added in #826 (closing #631 and #579).

dungbanguyen:
## Decision, 2026-08-31: extend the gate to the shape parameters, then settle each of the six

[...]

PR #5160, opened 2026-09-09
fix(contract-api): a declared page dial is read, or it is not declared

dungbanguyen:
Refs #827. The issue stays open on the `sort` half — see below.

## What was scoped out

The declared-parameter gate from #826 judges the parameters that decide a page's **membership** and deliberately excluded the three that decide its **shape**. The distinction is real: a dropped filter answers a wider question than the one asked, where a dropped page dial answers the right question in the wrong shape.

But the second is still a claim the surface does not keep. `?limit=5` returning the whole set means the caller sized a page and got a catalog. `?cursor=` accepted and inert is the shape that can walk page one forever.

`Cursor` and `Limit` are judged now. **`Sort` stays scoped out**, with the reason in the code and the line that will be deleted when it lands: honouring it is a per-store change rather than a contract edit — each of `listActivities`, `listCustomFields` and `listPartners` needs its store to take a sort vocabulary and a keyset that carries the sort key.

**No waiver list beside the gate, deliberately.** The honest state after a retirement is that the parameter is *gone from the contract*, not that it is declared and excused — so an operation that will not honour a dial removes the declaration.

## Two retirements and one implementation

The ruling on the issue said to check `listRecordGrants`' shape rather than file it under "catalog" with the other two. Checking changed the answer:

- **`listConsentPurposes`** and **`listVoiceCorpusSources`** are configuration a workspace sets up once. The dial goes, and each description now says the list is unpaged and why. Free to remove today; a breaking change once a client depends on the parameter being accepted.
- **`listRecordGrants`** is not configuration. A grant row is written every time somebody shares a record, so its count is driven by usage — and the unfiltered read returned the **whole table**, then ran one visibility query per row it had read. It gets paging.

## The visibility filter runs after the page

Which is what makes that paging less obvious than it looks. A page of `limit` rows can serve fewer once the invisible ones are dropped, so:

- **`HasMore` comes from the query** reading one row beyond the page, never from how many survived the filter. A page that filtered everything out still has more behind it, and reporting otherwise ends the walk early.
- **The cursor names the last row read**, not the last served. A cursor minted from the last *visible* row would re-read every invisible row after it on the next page, forever, without ever serving one.

## Mutation checks, including one that survived

| mutation | result |
|---|---|
| `HasMore` taken from the filtered result | the walk test fails |
| the SQL `LIMIT` dropped | **passes** |

The second is recorded in the code rather than papered over. The page stays correct because the truncation happens in Go; what changes is that the read pulls the whole table across — and no test through this API can see the difference, because it is invisible in the result. That is why the `LIMIT`'s purpose is stated where it is written.

## One process note

I first ran the gate with `-run DeclaredFilter`, saw green, and started investigating why the widening had caught nothing. The filter did not match `TestEveryDeclaredNarrowingParameterIsReadByItsHandler` — the assertion I had just widened. It had been failing correctly the whole time. Same shape as reading a passing test as evidence: the selector, not the subject, decided the answer.

## Checks

`make check-backend` green; `make craft-static` 0/0/0; `go test -count=1 ./gates` green; the compose/integration lane green. Contract, generated Go types and frontend types regenerated together. `grants.go` crossed the 500-line ceiling, so the read moves to `grantsread.go` — the writes decide whether a share may exist, the read decides what a caller may see of the shares that do.

**Note on CI:** main is currently red from #5103. Send/schedule 409s here are that, not this change.

## Summary by CodeRabbit

* **New Features**
  * Added cursor-based pagination and limit support to the record grants listing.
  * Record grants responses now include accurate pagination details, including whether more results are available and the next cursor.

* **API Changes**
  * Removed unsupported `cursor` and `limit` parameters from consent purposes and corpus manifest listings.
  * These listings are now explicitly documented as unpaged.

### DIFF
--- a/backend/api/crm.yaml
+++ b/backend/api/crm.yaml
@@ -13965,9 +13965,14 @@ paths:
       tags: [Compliance]
       operationId: listConsentPurposes
       summary: List the workspace's consent purposes (e.g. transactional, marketing_email, profiling).
-      parameters:
-        - $ref: '#/components/parameters/Cursor'
-        - $ref: '#/components/parameters/Limit'
+      description: |
+        UNPAGED, deliberately. A workspace's consent purposes are configuration — a handful of
+        them, set up once — so the answer is the whole set and `page` reports `has_more: false`.
+
+        It used to declare `cursor` and `limit` and honour neither: `?limit=5` returned the
+        catalog, and a caller who sized a page got everything. Nothing was hidden, and nothing
+        was kept either. Removing the dial makes the surface true rather than adding paging
+        machinery with no user.
       responses:
         '200':
           description: A page of consent purposes.
@@ -15505,9 +15510,13 @@ paths:
       security: [ { cookieAuth: [] } ]
       x-agent-access: human-only
       summary: List the owner's corpus manifest and live meter; source text is never returned.
-      parameters:
-        - $ref: '#/components/parameters/Cursor'
-        - $ref: '#/components/parameters/Limit'
+      description: |
+        UNPAGED, deliberately, and for the same reason listConsentPurposes is: one voice
+        profile's corpus sources are configuration rather than accumulated data, so the answer
+        is the whole manifest.
+
+        It declared `cursor` and `limit` and honoured neither. Removing the dial is free now
+        and stops being free once a client depends on the parameter being accepted.
       responses:
         '200':
           description: Manifest and meter.
--- a/backend/gates/declaredfilters_test.go
+++ b/backend/gates/declaredfilters_test.go
@@ -92,12 +92,32 @@ const (
 	wantMinimumOverlayShadows = 4
 )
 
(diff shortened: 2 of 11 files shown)


dungbanguyen:
Fixed — the case was passing for the reason you name.

`TestTheRecordGrantListIsBoundedWithoutADial` seeded three grants and asserted
only that the result was non-empty, so it passed whether the read was bounded at
fifty or not bounded at all — which is the behaviour the change exists to stop,
and it reads exactly like a short table.

The corpus is now one row over the default and the assertions are the page's: its
length, that `HasMore` says there is more, and that `NextCursor` is not empty.
The default comes from `storekit.ClampLimit(nil)` rather than a literal 50, so
the case does not quietly stop testing the bound when the contract moves it.

Mutation-checked: multiplying the clamp by 1000 in `grantsread.go` fails it with
"got 51 grant(s) over a corpus of 51; want the default page of 50".

Issue #2276, opened 2026-08-22
A multi-call unit of work reopens its rail occurrence once per model call


tinnhlt-gra:
Follow-up from [#2275](https://github.com/margince/margince/pull/2275), named there and filed here so it is not carried in prose.

The router now announces a model call's START as well as its settle, so a live line exists while a call runs. The unit it frames is **one logical call**, which is what let that PR stay small: a logical call has a beginning the router can observe and an end it already observes, so it needs no scope, no teardown at the ~50 sites that mint a correlation id, and no attempt durable across processes.

For every kind the rail draws today that is exact — `summarize`, `draft_reply` and `offer_draft` are all registered `oneShot`, so the logical call **is** the unit of work.

## The gap

A task whose unit of work spans MANY logical calls under one correlation id does not fit. The occurrence key is `correlation_id + ":" + task`, so those calls share one occurrence — and each new call announces a start under a higher attempt, which outranks the previous call's settle. The row therefore reopens once per call:

```
call 1: running(attempt 1) -> done(attempt 1)
call 2: running(attempt 2) -> done(attempt 2)
...
```

The deep read's page-parallel fact lane is the real instance: `site_fact_extract` fires many calls under one `site_read` correlation id by design.

## Why it is harmless today, and why that is not a reason to leave it

The row still ends settled, and `site_extract` / `site_fact_extract` / `site_triage` all carry a written not-displayed reason in `ACTIVITY_LINE`, so nothing renders the intermediate states. The cost today is event volume and row churn, not a wrong screen.

It stops being invisible the moment anybody writes copy for one of those kinds — the line would bounce between the read's `running` and `recent` arms for the length of a site read.

## What actually fixes it

A **unit-of-work** frame rather than a per-call one: something that knows the site read began and ended, announcing once at each end. That is what the earlier design for this work proposed, and it was descoped because it needs a scope with a teardown at every correlation mint site, an attempt that survives a process, and a lease that outlasts the whole unit rather than one call.

It also interacts with the grain question already recorded in [#2272](https://github.com/margince/margince/issues/2272): one site read files THREE occurrences because it runs three tasks under one correlation id. The two are the same shape of problem and should be fixed together — when `site_read` becomes a carrier, those three tasks become `SourceNoOccurrence` and the churn goes with them.

## Not blocking

Nothing renders these kinds, so this is a cleanup with a deadline set by whoever first wants a site read to be visible — not a defect anybody can currently see.

dungbanguyen:
[...]

Issue #2272, opened 2026-08-22
Five AI tasks report settled-only, and the carriers that could say queued/running are unused


tinnhlt-gra:
Follow-up from [#2259](https://github.com/margince/margince/pull/2259), named there and filed here so it is not carried in prose.

Every AI task now reports into `ai_task_run`. Five of them are reported by the ROUTER, which learns of a call only once the call is over — so their occurrences are settled the moment they appear. Each of these five already has a durable, attributable row that could say `queued` and `running` instead:

| task | table | states vs the projection's five | lease | `attempt` |
|---|---|---|---|---|
| `transcript_propose` | `transcript_read` | `queued/running/done/failed` — maps 1:1 | ✅ `TranscriptReadLease` (5m) | ❌ |
| `site_extract` / `site_fact_extract` / `site_triage` | `site_read` | **seven**: adds `deferred`, `partial`, `cancelled` | ❌ none exists | ❌ |
| `voice_build` | `voice_build` | adds `deferred`; `succeeded`→`done` | ❌ none exists | ❌ |

**For long work, settled-only is worse than silence.** A rep clicks "read this site", sees nothing for forty seconds, then it appears already finished. That is the argument for doing site read first.

## What each carrier costs

- **An additive `attempt` migration**, all three. The projection's guard is lexicographic on `(attempt, rank)` and a re-queue must outrank a settled lower attempt — attachment extraction needed exactly this (`1787324557_attachment_extraction_attempt`).
- **A lease**, for site read and voice build. Not bookkeeping: without one `stalled` is not derivable, so a dead worker renders as running forever — the precise failure the projection exists to prevent.
- **A ruling on two states the projection has no word for.** `deferred` means "waiting on budget, retry at `next_attempt_at`" — not `queued` (nothing will pick it up), not settled. Site read's `cancelled` is terminal but `done` and `failed` would both lie. Either the state enum grows or somebody decides the mapping.

Two things already fit: `site_read.stopped_reason` (`budget|page_cap|byte_cap|deadline`) is a closed vocabulary that drops straight into `degrade_reason`, and `partial` → `degraded` is exact.

## The grain question, which is the sharper half

**One site read currently files THREE occurrences.** Its correlation id is the `site_read` row id (`compose/deepreadprincipal.go:53`), and it runs three distinct `ai.Task` values — so the router's `correlation_id + task` key produces `site_triage`, `site_extract` and `site_fact_extract` as three separate lines for one thing a person asked for once.

[...]

PR #5161, opened 2026-09-09
feat(rail): a transcript reading says queued and running, not only done

dungbanguyen:
Step 3 of #2272's suggested order — the cheapest slice, and the one whose states already map 1:1.

## The problem

The router learns of a model call once the call is **over**, so a task it reports is settled the moment it appears. A rep presses "read this transcript", sees nothing while the model works, and then finds it already finished.

> **For long work, settled-only is worse than silence.** — #2272

It looks like nothing happened, and then it looks like it never ran.

## Why this reading can carry its own line

It has what the router does not: a durable row, the requester recorded on it, a lease that says when a claim has died, and four statuses (`queued/running/done/failed`) that are exactly the projection's live and terminal pair. So `railOwners` hands `transcript_propose` to a `transcript_read` carrier and the router stays quiet about it.

## The attempt column

`1788951315_a_transcript_reading_counts_its_attempts` — additive, both columns defaulted, nothing backfilled.

The projection guards lexicographically on `(attempt, rank)`. A reading handed back after its worker died must announce `queued` at a **higher** attempt than the `running` it replaces, or the rail shows a dead reading as live until the lease expires — the precise failure the projection exists to prevent. `attempt_at` travels with it because the rail ages a live row from when **this** attempt was enqueued, not from when the reading was first created; without it a reading re-armed an hour later is past its lease before any worker sees it.

## No subject label, deliberately

My first version named the meeting. Two gates refused it, and both were right:

```
restrictedreaders_test.go: transcriptSubjectLabel … reads the activity table and excludes no held row
audiencereaders_test.go:   transcriptSubjectLabel … reads the activity table and composes no audience test
```

Naming the meeting means reading `activity` on the **emit path**, where the audience and restricted-row clauses have no caller to compose against — the worker transitions run under a system principal, which reads every audience away. So the label would be resolved by the one reader that cannot ask the question those gates force.

The unnamed sentence is a complete one ("I'm reading the meeting transcript"), so the cost is a nicety and the alternative was ratifying a reader of held rows to buy it. `NAMED_LINE` gets no entry.

## Two waivers deleted

`writeshape_test.go` waived `startTranscriptReadInTx` and `finishTranscriptReadTx` on the ground that this run record emits nothing. Both are now false, and the gate said so itself — `waiver … matched no subject`.

## Verification

`make check-go`, `make check-fe`, `make craft-static` green; `make test-it` green for `backend/migrations`, `backend/internal/modules/activities` and the compose integration package.

Four new end-to-end cases drive the real store and the real consumer. Mutation-verified, and the split is the point:

- drop the attempt bump → only the re-arm case fails: `after the re-arm, state/attempt = running/1, want queued/2 — a row frozen at running is a reading the rail says is working and nobody holds`
- drop the claim's announcement → the running case fails, and the re-arm case with it, because it has no running state left to outrank

## Scope

`voice_build` and the site-read carrier (#2272 steps 1–2) are untouched: both need a lease that does not exist and a ruling on `deferred`/`cancelled`, which the issue records as the sharper half. #2276 rides on step 1 and is unaffected by this.

## Summary by CodeRabbit

- **New Features**
  - Meeting transcript reading now appears in the AI activity feed with queued, running, completed, stalled, degraded, and failed statuses.
  - Activity ownership and progress are associated with the person who initiated the transcript reading.
  - Transcript reading attempts and recovery after abandoned runs are tracked more clearly.

- **Localization**
  - Added transcript-reading activity messages in English, German, and Vietnamese.


dungbanguyen:
Taken, and widened: the same defect was in all three locales.

You flagged Vietnamese, where `bản ghi cuộc họp` can mean a recording or minutes while `transcriptread.cta` — the button that starts this very read — already says `bản chép lời`. Correct. But English and German had it too: I wrote "the meeting transcript" and "das Meeting-Transkript" where the UI's settled nouns are "transcript" and "Transkript".

A rail line and the button that starts it are one feature, so two nouns for it is the reader's problem rather than the translator's. All three now follow the established term.

Issue #5153, opened 2026-09-09
A scheduled send cannot carry an unsubscribe link: the worker composes an empty SendPath


dungbanguyen:
## What is wrong

`newScheduledSendWorker` builds its store with an empty `SendPath`:

```go
// internal/compose/commsscheduled.go:118
store: sendStore(pool, SendPath{}).WithBlobstore(blob),
```

`SendPath.PublicBaseURL` is therefore `""`, and its own doc says what that means:

> `PublicBaseURL` is configured at boot, never derived from a request. Empty means a marketing send refuses (the store's fail-loud branch) rather than emitting a forgeable link.

So any scheduled message that carries a one-click unsubscribe link refuses at fire time:

```
comms_scheduled_send: firing 01a0858f-…: send: public base URL is not configured
```

**This is the production worker, not a test double.** `jobs_capture.go:103` composes it the same way — the only other caller is the integration driver, which inherits the same gap.

## Why nothing catches it

Every scheduled-send fixture names the `transactional` purpose, which needs no unsubscribe link. So the whole scheduled lane is exercised on the one purpose class that never reaches the branch. The immediate send path is composed with a real `SendPath` and works; only the scheduled one is empty.

That is the shape worth noting: the corpus is not narrowed and the detection is not blind — the lane simply only ever asks the question in the one shape that cannot fail. A fixture that always picks the cheap option is a census of one.

## How it surfaced

Found while trying to move fixtures off `transactional` after #5103 made that purpose permanently unsupported on the send path. Reseating them on `business_correspondence` fixes the immediate sends and then fails every scheduled one here — which is how a purpose choice made for convenience turned out to be load-bearing.

## What the fix needs

The scheduled-send worker has to receive the configured public origin the same way the immediate path does, which is a wiring question through `jobs_capture.go`'s config rather than a change inside the worker. `DriveScheduledSendForTest` must carry it too, or the harness keeps assembling a worker production does not have — its own doc claims it is assembled "the way the worker role assembles it", and today that claim is true in the wrong direction.

Worth a case that a scheduled send of a link-carrying message reaches the provider, since that is the assertion nothing currently makes.

## Not blocking on #5103

This predates it and is independent: a scheduled marketing send fails today, on `main`, with no relation to the consent change. It became visible through that work rather than because of it.

dungbanguyen:
[...]

PR #5166, opened 2026-09-09
fix(comms): main goes green — every send fixture names a purpose that can send

dungbanguyen:
**`main` is red and has been for hours.** `main-health` confirms it on `0f73b5db0`:

```
--- FAIL: TestTestMailboxFullLoop
    send-email under "transactional" → 409 "consent_not_granted", want 202
```

25 integration tests, across the send, schedule, capture, preference-centre and MCP paths. This turns them green.

## Why they broke

#5103 made the transactional purpose permanently unsupported on the send path. That is deliberate and its rollout note said so. What did not travel with it were the 25 tests that named that key because it was the cheapest one to type.

### 1. The wiring — a real product bug (#5153)

`newScheduledSendWorker` composed `sendStore(pool, SendPath{})` — an **empty** send path, so `PublicBaseURL` was `""` and any fired message carrying a one-click unsubscribe link refused at fire time:

```
comms_scheduled_send: firing …: send: public base URL is not configured
```

That is the **production** worker — `jobs_capture.go` composes it the same way. It was invisible because every scheduled-send fixture named the one purpose class that needs no link: a lane exercised only in the shape that cannot fail.

`JobsConfig` gains a `SendOrigin`, carrying the base URL and the posture as a **pair** — netguard needs both to tell a working dev stack from a link the recipient cannot open. Two test seams take it too; both claimed in their own docs to assemble the worker "the way the worker role assembles it", and that was true in the wrong direction.

### 2. The fixtures

Sends move to `business_correspondence` with an inbound activity seeded as the qualifying event — what a derived basis is actually derived from. Where a message reaches several people at once it also **claims a category**, because a send carrying an unsubscribe link reaches one addressee at a time; the link is personal to whoever it opts out.

`sendPurpose` and `lockedPurpose` are named apart, because the two stopped being the same word: `transactional` is still the locked lane the preference centre refuses to change, and no longer authorizes a send.

### 3. The two that looked like a contradiction

I reported these as needing a ruling, and I was wrong — there is a route that keeps both rules.

`TestPreferenceCenterOneClickUnsubscribe` and `TestEveryPreferenceAnswerIsUncacheable` assert a send with nothing to unsubscribe from carries no link. That fact used to be spelled `transactional`, the one locked key the **legacy** arm suppresses a link for. Since that key can no longer send, the same fact is now spelled as a claimed non-marketing category — which is where the rule already lived:

```go
func (u unsubscribeSurface) carries() bool {
    if u.category != "" { return u.category.CarriesUnsubscribe() }   // marketing only
    return u.purposeKey != "" && !lockedPurposeKey(u.purposeKey)     // legacy arm
}
```

The category is a caller's **claim**, not something the engine derives, so a fixture can state it. Nothing about the product rule moves.

## Not fixed here

One production path stays open on #5153: an approved held message is re-armed through `heldSendActors`, which composes an empty `SendPath` too. Threading an origin there crosses the approvals composition — server, signals, LinkedIn and attention wiring — which is a change of its own and not a main restoration. The one suite that reaches it says so where it claims its category, so nothing there has to move when #5153 lands.

## Verification

- `internal/compose/integration` — **all green** (178s), from 25 failures
- `make check-go`, `make craft-static` green

@margince/margince — this unblocks every open pull request in the repository.

### DIFF
--- a/backend/cmd/worker/jobrunner.go
+++ b/backend/cmd/worker/jobrunner.go
@@ -230,7 +230,11 @@ func newJobRunner(pool *pgxpool.Pool, logger *slog.Logger, cfg workerConfig, cap
 		// The registry that resolves a staged delivery's mailbox: the SAME
 		// sweep registry the capture polls use, so the connector set that
 		// syncs a mailbox is the one that transmits from it.
-		SendRegistry:      captureReg,
+		SendRegistry: captureReg,
+		// The origin a fired message builds its unsubscribe link on — the same
+		// pair sendPath hands the api's immediate send, so a message scheduled
+		// for later carries the link a message sent now would.
+		SendOrigin:        compose.SendOrigin{PublicBaseURL: cfg.publicBaseURL, Environment: cfg.posture},
 		SendPacing:        compose.SendPacing{Limit: cfg.sendRateLimit, Window: cfg.sendRateWindow, MaxAge: cfg.sendMaxAge},
 		CloseDateInterval: cfg.closeDateInterval,
 		ReconcileInterval: cfg.reconcileInterval,
--- a/backend/internal/compose/commsscheduled.go
+++ b/backend/internal/compose/commsscheduled.go
@@ -112,10 +112,15 @@ type scheduledSendWorker struct {
 // immediate one is: the signature, the sender name, the unsubscribe linker and
 // the draft-outcome recorder all present, all identical. A hand-built store
 // here would be a second send path wearing the first one's name.
-func newScheduledSendWorker(pool *pgxpool.Pool, delivery DeliveryMachinery, blob blobstore.Store, pacing SendPacing) *scheduledSendWorker {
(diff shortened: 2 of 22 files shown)


dungbanguyen:
Four more found and fixed — and the way I missed them is worth stating.

My first census ran `internal/compose/integration` and reported **25**. That directory has subpackages, and `integration/channels` holds the Telegram round trip: four replies naming the same key, failing the same way. I had run one package and reported a total.

That is the exact shape this repository's gates keep catching — a census whose corpus is narrower than its claim — and I wrote one while fixing the consequences of another. The number now comes from `make test-integration` over the whole lane.

**`make test-integration`: the entire lane green.** `make check-go` and `make craft-static` green.

The four are answers to a message the customer sent, so `reply_to_inbound` is what they actually are, and the context is claimed alongside the purpose — only a marketing category carries a one-click unsubscribe link, and a chat reply that grew a footer would not be the message the rep wrote.

dungbanguyen:
Superseded by the PR above. `main` went green via #5152 while this was in flight, so the fixture half is no longer needed — the send-origin wiring is, and it moves there on its own.

PR #5169, opened 2026-09-09
fix(compose): a mailed card's staging fault no longer costs its siblings their own review

### LINKED ISSUE #3410: A vCard naming no company can never become a review: the identity asserts organization, the payload omits it, and the retry it advises fails identically for ever

tinnhlt-gra:
## Summary

- `stageReviews` returned on the first mailed card's staging error, which lost every remaining eligible card in the same message to that one card's failure and aborted the whole ingest job.
- Every eligible card now gets its own staging attempt (`stageReviewsWith`, injectable so a test can force one card's stage to fail without a real staging conflict). A failed card is logged and counted; the aggregate error still tells River to retry the whole message against `vcardIngestMaxAttempts` when any card failed — a card's fault may be the database, not the data.
- Closes margince#3410 (the mailed-path half of the cascade-failure; the ORG-less-card half was already fixed).

## Test plan

- [x] New unit test in `vcardingest_test.go` exercises `stageReviewsWith` directly: one card's stage failing does not stop a sibling card from also being staged, and the aggregate error names the failed/eligible count.
- [x] `make check-all` (backend + frontend + integration lane, real Postgres) green with 0 failures, 0 skips, on this branch rebased onto current `main`.
- No UAT video — this is a backend job-worker fix with no user-facing surface (River-queued vCard ingest worker).

---
## Summary by cubic
Fixes mailed vCard ingest so a staging failure on one card no longer prevents sibling cards in the same message from receiving review. Previously, the first failure aborted staging and the job; now every eligible card is attempted, and any failure still returns an aggregate error so River retries the message.

**Bug Fixes**

- Keeps each card's own error in the aggregate via `errors.Join`, so Work's `errors.Is` classification still distinguishes real faults from permission/not-found refusals instead of retrying those.
- Logs each failed card with its activity and reports the failed and eligible card counts.
- Adds tests for first- and later-card failures, successful staging, and non-review cards.
- Closes margince#3410's mailed-path cascade-failure case.

<sup>Written for commit 4b1accde8e90060c231f73b39a49d7bb33e01c6f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/5169?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/backend/internal/compose/vcardingest.go
+++ b/backend/internal/compose/vcardingest.go
@@ -169,7 +169,7 @@ func (w *vcardIngestWorker) importCards(ctx context.Context, args VCardIngestArg
 			"activity", args.Activity, "card", r.Index+1,
 			"outcome", string(r.Outcome), "reason", r.Reason)
 	}
-	return w.stageReviews(actorCtx, entries, results)
+	return w.stageReviews(actorCtx, args.Activity, entries, results)
 }
 
 // stageReviews turns every near-match into a proposal a human can decide.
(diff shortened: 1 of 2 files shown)


tinnhlt-gra:
Ran independent Fable and Codex reviews on this PR; both converged on the same MAJOR finding — the aggregate staging error dropped every underlying error, which would have made a stager's `ErrPermissionDenied`/`ErrNotFound` refusal (previously classified by `Work` as terminal, no retry) look like an ordinary fault and retried 5x instead. Fixed in 4b1accde8: wraps every card's error with `errors.Join` + `%w`, adds `activity` to the failure log line to match the rest of the file, and covers the two missing test cases (failure on a later card, and every card failing).

PR #5181, opened 2026-09-09
feat(frontend): restore contact edit, merge and archive on the person record page

tinnhlt-gra:
## Summary

**margince#4823** — `PersonPageV2` (the route `/#/contacts/{id}` actually resolves to since routing moved off `PersonScreen`) offered no way to correct a bounced email, a wrong phone number, a misspelled name or a title. Every verb that would do it — Edit, Merge, Archive — was missing from the header. The working form still existed, fully tested, hanging off `PersonScreen` (`contacts.tsx`), which nothing routes to any more. API-only was the only path left for what should be an ordinary correction.

- `PersonEditMergeArchive` is now a shared component both pages render into their header's `badges` slot (`RecordView`'s `.record-badges` container, previously entirely absent from `PersonPageV2`'s DOM). One writer of the PATCH/merge/archive wiring rather than two.
- Each page passes its own correct write-refusal reason: `PersonScreen` keeps its simpler archived-only check; `PersonPageV2` uses its own `useRecordWriteRefusal`, which also answers "not yours to change" — a case the old page's own header verbs never had to consider, and which the shared component takes as a caller-supplied `disabledReasonId` rather than deciding itself.
- `PersonPageV2`'s own native `ShareAction` is untouched — the orphaned page's Share was deliberately *not* pulled across, since PersonPageV2 already has its own.

## Why so many new/moved files

`contacts.tsx` and `personpage.tsx` were both already over the frontend's file-length ratchet (`scripts/fe-file-length-waivers.txt`), which explicitly forbids absorbing growth into a waived file ("shrinking is allowed; growing is not"). Wiring the fix in without violating that meant splitting out:
- `personeditmergearchive.tsx` — the shared `PersonEditMergeArchive` component.
- `personformfields.ts` — the person form's field schema and request-body mappers (`contactCreateFields`/`personEditFields`/`mapPersonBody`/`mapPersonUpdate`), shared by both contacts.tsx's create flow and the new edit component, extracted specifically to avoid a circular import between the two.
- `personactions.tsx` — `PersonPageV2`'s existing header-verbs component (Email/Call/Meetings/Log/Add task/More actions), moved out wholesale to free the room the badges wiring needed; unrelated to the fix itself.
- `personpage.testkit.tsx` — the shared test fixture/mount helper, split out so the new test coverage (below) didn't push `personpage.test.tsx` over its own test-file ratchet, and because biome's `noExportsInTest` forbids exporting from a `.test.tsx` file directly.

Each touched file is *smaller* after this change than the waiver that governed it before — not merely no larger — confirmed via `make fe-file-length`.

## Test plan

- [x] New `personpage.editverbs.test.tsx`: PATCHes `/people/{id}` from the header's edit form and asserts the body; archives the record from the header; confirms merge is reachable; confirms edit/merge/archive are all refused with the page's own write-refusal sentence for a non-writable record. Mutation-tested (removed the `badges` wiring from `RecordView`, confirmed all 4 new tests fail; restored, reconfirmed green).
- [x] Existing `contacts.test.tsx` (40 tests) and `personpage.test.tsx` (44 tests) both still fully green after the extraction — behavior unchanged for `PersonScreen`.
- [x] Fixed `overlay-refusal-coverage.test.ts`'s source-scanning `merge-person` case to point at the new file location (an established pattern in this same file — `merge-org` already documents an identical prior move).
- [x] `make fe-quality` (ds-gates, drift, biome, composed tsc including test files) fully green.
- [x] Full `frontend/src/screens/` suite: 454 files, 6134 tests, 0 failures.
- [x] `make fe-file-length`: all three touched files now under their own prior waiver.

---
## Summary by cubic
Restores edit, merge, and archive on the person record page (the route `/#/contacts/{id}` resolves to), which previously offered no way to correct a bounced email, wrong phone number, misspelled name, or title. The working form existed on PersonScreen, but nothing routes there anymore, so API-only was the only path left.

**Refactors**

- Edit, merge, and archive now share one `PersonEditMergeArchive` component, rendered by both pages into the header's badges slot.
- Each page supplies its own write-refusal reason; PersonPageV2's also answers "not yours to change".
- Saves invalidate the `person360` and `personBrief` cache keys PersonPageV2 renders from, so a correction shows immediately instead of leaving stale state.
- The form's field schema and request mappers moved to `personformfields.ts` to avoid a circular import.
- `PersonActions`, the form fields, and the shared test fixture moved to their own files to stay within the frontend's file-length ratchet, which forbids growing waived files.
- New `personpage.editverbs.test.tsx` covers PATCH, archive, merge reachability, and write refusal from the header.

<sup>Written for commit ad16165a7a3e139e21c6d7e5b80ac441ff54cc15. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/5181?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/api/if-match-coverage.test.ts
+++ b/frontend/src/api/if-match-coverage.test.ts
@@ -50,7 +50,10 @@ const UNPINNED_WRITES: readonly string[] = [
   // screenful of rows at once, and it was the verb of three in that bar that
   // could not lose a race it should lose.
   "screens/companyheader.tsx DELETE /organizations/{id}",
-  "screens/contacts.tsx DELETE /people/{id}",
+  // PersonEditMergeArchive's own archive, shared by PersonScreen (contacts.tsx)
+  // and PersonPageV2 — the entry moved here with it, the same move merge-org's
+  // own line made out of organizations.tsx.
+  "screens/personeditmergearchive.tsx DELETE /people/{id}",
   "screens/personrail.tsx DELETE /relationships/{id}",
   "screens/relationships.tsx DELETE /relationships/{id}",
   "screens/contractform.tsx PATCH /contracts/{id}",
--- a/frontend/src/screens/contacts.tsx
+++ b/frontend/src/screens/contacts.tsx
@@ -2,10 +2,8 @@ import { useQuery, useQueryClient } from "@tanstack/react-query";
 import { useId, useState } from "react";
 import { api } from "../api/client";
 import type { components } from "../api/schema";
-import { ifMatch, requireVersion } from "../api/version";
 import { usePageName } from "../app/pagemeta";
 import { useRecordZone } from "../app/recordzone";
-import { navigate } from "../app/router";
 import { activityTimeline } from "../design-system/activitytimeline";
 import { Badge, SegmentedControl } from "../design-system/atoms";
 import { RecordView } from "../design-system/composed";
@@ -22,7 +20,6 @@ import { primaryEmail } from "../format/primaryemail";
 import { normalizeProfileUrl } from "../format/profileurl";
 import { useLocale, useT } from "../i18n";
 import type { MessageKey } from "../i18n/en";
-import { ArchiveAction } from "./archive";
 import {
   LoadMoreButton,
   OverlayUnavailable,
@@ -40,7 +37,6 @@ import {
   type ObjectCustomFields,
   useObjectCustomFields,
 } from "./customfields.form";
-import { EditAction } from "./edit";
 import { EntityRef } from "./entityref";
 import { RecordHistoryTab } from "./history";
 import {
@@ -53,7 +49,6 @@ import {
   useTagChips,
 } from "./listquery";
 import { LogActivity } from "./logactivity";
-import { MergeAction } from "./merge";
 import {
   IdentityRail,
   type Person360,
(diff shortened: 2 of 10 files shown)


tinnhlt-gra:
## UAT — PersonPageV2 edit restores live cache invalidation

Verified against a live local stack (fresh worktree, seeded demo data), not mocks:

1. Signed in, opened a contact (Alice Müller) on the restored PersonPageV2 header.
2. Clicked the header's Edit action, set a Title ("VP of Procurement"), saved.
3. The subtitle updated to "VP of Procurement · Demo GmbH" **immediately**, with no manual refresh — proving `invalidateRecord` now busts the `["person360", id]` key `PersonSubtitle` actually reads from, not just the list/`["person", id]` keys `EditAction`'s own props reached before this fix.

wave10 UAT

Editing Alice Müller's title and watching the header subtitle update live

PR #5188, opened 2026-09-09
A colleague nobody is called is not an empty workspace, and the fixture gives them real seats

tinnhlt-gra:
Four changes, one story, measured end to end: **case33 on haiku goes 0/3 → 3/3.**

## 1. An empty colleague list was two different answers

```
  list_colleagues {q: "Lena Fischer"}  →  { "colleagues": [] }
                                              ↓
              "nobody works here"    OR    "nobody is spelled that way"
```

A caller cannot tell them apart, so it picks — and it picked the wrong one. In the lane, an assistant asked to hand an account to a colleague tried `q: "Lena Fischer"`, then `q: "Lena"`, and told the user she *"isn't in the system yet"*.

A narrowing that matches nobody now hands over the set it was matched against: `colleagues` stays empty (the honest answer to what was asked), `all_colleagues` carries the seats, and the envelope says in words what happened. Through the **warning channel**, not the tool copy — a sentence in `Limits` is paid for on every catalog listing of all 75 tools on every run, and this one is only true on the miss.

- **A successful empty fallback and an unreadable one are different bytes.** `all_colleagues` is a pointer: `[]` when the roster was read and holds nobody, absent when it could not be read. Identical serialisation would have reproduced this field's own defect one level up. *(CodeRabbit, Major — taken.)*
- **The cap rides its own flag.** 200 alphabetical names with the one asked for absent reads as *proof* of absence. `truncated` means the narrowing has more matches; `all_colleagues_truncated` means the fallback is a page, and the warning says not to conclude absence from it.
- **The fallback cannot fail the call** — the asked question is already answered by then, the shape `reportDuplicates` settled.

## 2. …and the tool was right all along — the fixture had no colleagues

`POST /users` writes `status='invited'`; a seat with no password signs in nowhere. `identity.Colleagues` lists only `status='active'`. So **every colleague this seed created was invisible on every call of every run** — verified across the last haiku sweep: 5 calls, 0 non-empty. Two scoring criteria on two cases were unreachable by any model, and both were being recorded as model failures.

The seed now mints a single-use set-password link and redeems it — the route the product itself uses for an installation with no outbound email, and the one that sets `status`. Then it **verifies the roster read**, not the rows it wrote: a seat that exists and is not listed is exactly the guarded state.

## 3. A staged call says what it staged *and what it already did*

#5193 told the agent to carry on with work the approval does not block. It did — and then wrote its answer about the one thing that stopped:

```
  before  "One approval waiting: reassign Ostfriesen to Lena Fischer."
          (a duplicate merged and a company archived, unreported)

  after   "Done: Ostfriesen Kranbau records are merged (kept the
           "Gesellschaft mbH" version with all correspondence), and
           Sauerland Kunststoff is archived. One approval waiting…"
```

Doing the rest and *reporting* the rest are separate instructions. I trimmed the second clause off #5193 for length before merging; these transcripts are why it belongs.

## 4. The case33 pattern demanded one surface form

Three runs said the same correct thing three ways — `kept the "Gesellschaft mbH" version`, `kept "Ostfriesen Kranbau Gesellschaft mbH" as the surviving record`, and a heading followed by `Kept Gesellschaft mbH`. The pattern required name and suffix **adjacent** and greened only the middle one.

It now anchors on the suffix, which is the only thing that tells the two cards apart. It still fails what it exists to fail — `"I merged the two Ostfriesen Kranbau records into one"` names no survivor and matches nothing. Whether the named survivor is the one the merge produced stays the judge's question; that half can read the transcript and a regex over the final answer cannot.

## Measured

| | |
|---|---|
| case33 haiku, before | **0/3** |
| + real seats | 1/3 — `Lena Fischer` gone from the failures entirely |
| + reporting clause | 1/3 — all three runs now report the writes |
| + the pattern | **3/3** |

Unit, gates, `craft-static` and `lint` green; the compose integration lane green apart from `TestPreferenceCenterOptOutBlocksAgentSend`, which **fails on a clean `origin/main`** and reds four of six shards on every PR. That is not from this branch.

## Summary by CodeRabbit

- **New Features**
  - When a colleague search finds no matches, the roster can now be returned as a fallback, with warnings indicating whether it is empty, unavailable, or truncated.
  - Results now distinguish between an empty workspace roster and a search with no matching names.

- **Improvements**
  - Analytics report guidance now clearly emphasizes creating a document for people to read.
  - Approval messages now remind agents to report work already completed alongside pending work.

- **Documentation**
  - Updated tool descriptions and reference schemas to document the enhanced colleague search results.

### DIFF
--- a/backend/internal/modules/agents/explain.go
+++ b/backend/internal/modules/agents/explain.go
@@ -330,7 +330,13 @@ const maxRemedyBudget = 4 * httperr.MaxFaultText
 // The summary is the sentence the human's own card carries, so the person and
 // the agent are waiting on one described thing.
 //
-// IT ALSO SAYS WHAT IS NOT BLOCKED, because an agent reads a refusal as a stop.
+// IT ALSO SAYS WHAT IS NOT BLOCKED, AND TO REPORT WHAT ALREADY HAPPENED,
+// because an agent reads a refusal as a stop and then writes its answer about
+// the stop. Two measured runs merged a duplicate and archived a dead company
+// correctly and then said only "one approval waiting: reassign the account" —
+// two completed writes the reader was never told about. Doing the rest and
+// reporting the rest are separate instructions, and only the first was given.
+//
 // Asked to merge two words and then give the survivor a meaning, a measured run
 // staged the merge, relayed the summary correctly, and finished with "Confirm
 // and I'll proceed, then add the description afterward" — deferring an
@@ -352,5 +358,6 @@ func stagedExplanation(staged *workflow.StagedApprovalError) string {
 	return "Confirm-first (🟡): a person answers this before it runs. " + what +
 		" Tell them that, in those words; they release it in the CRM, and this exact call then " +
 		"repeats with \"approval_id\": \"" + staged.ApprovalID.String() + "\". " +
-		"Blocks THIS call only — do the rest of what you were asked that does not depend on it."
+		"Blocks THIS call only — do the rest of what you were asked that does not depend on it, " +
+		"and report what you DID alongside what is waiting."
 }
--- a/backend/internal/modules/agents/results_identity.go
+++ b/backend/internal/modules/agents/results_identity.go
@@ -38,6 +38,39 @@ type ListColleaguesResult struct {
 	// nothing would read a capped list as the whole roster and report that a
 	// colleague does not work here.
 	Truncated bool `json:"truncated,omitempty"`
+	// AllColleagues is the set the narrowing was matched against, present ONLY
+	// when a narrowing `q` matched none of it — the same obligation a refusal
+	// naming a closed vocabulary carries.
+	//
+	// An empty `colleagues` answers two different questions identically: this
+	// workspace employs nobody, and nobody here is spelled the way you asked.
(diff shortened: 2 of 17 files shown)


tinnhlt-gra:
Both findings taken.

**Major — an empty fallback and an unreadable one were the same bytes.** Right, and it is this field's own thesis one level up: "nobody works here" and "I could not find out" are different answers, and a caller acting on `data` alone saw the omitted field from both. `AllColleagues` is now `*[]Colleague`, so a successful fallback over a workspace that employs nobody serializes as `"all_colleagues": []` and an unreadable one is absent. Asserted on the raw bytes, including that the two outputs differ, and mutation-verified — dropping the non-nil normalisation reds it.

**Minor — the tool copy promised a list it may not deliver.** Corrected: it now says the list is ABSENT if it could not be read and partial if `all_colleagues_truncated`, and to read the warning before concluding a person has no seat.

tinnhlt-gra:
Folded into #5247 — one CI run instead of two while main is red. Every part of this PR, including the CodeRabbit findings addressed here, is carried over verbatim in commit `079a4c1e7`.

Issue #5025, opened 2026-09-08
Fifteen frontend length waivers were re-frozen upward by the company rename


dungbanguyen:
The company rename (https://github.com/margince/margince/pull/5016) re-froze
fifteen entries in `scripts/fe-file-length-waivers.txt` upward, by 89 lines in
total. This is the ticket to pay that back down.

## Why they grew

Nothing was added. The record type's word went from `organization` to `company`,
and the local variable it is usually held in went from `org` to `company` — from
three characters to seven. That widens JSX prop lists past biome's print width,
so the formatter splits lines it used to keep whole:

```tsx
// before: two props, one line
<OrganizationFinanceCard organizationId={org.id} lifecycle={org.lifecycle} />

// after: the same two props, three lines
<CompanyFinanceCard
  companyId={company.id}
  lifecycle={company.lifecycle}
/>
```

The formatter's own output either way — `biome check` is clean on both sides,
and the tree it produced was verified against main to make sure no unrelated
file was reformatted along the way.

## What moved

| File | was | now |
|---|---:|---:|
| `screens/deals.tsx` | 4497 | 4505 |
| `screens/deals.test.tsx` | 2633 | 2645 |
| `screens/companies.tsx` | 2925 | 2945 |
| `screens/companies.test.tsx` | 2534 | 2537 |
| `screens/company360.test.tsx` | 2306 | 2308 |
| `screens/personrail.tsx` | 1575 | 1580 |
| `screens/offers.tsx` | 1418 | 1424 |
| `screens/listquery.test.tsx` | 1328 | 1330 |
| `screens/companyheader.tsx` | 975 | 991 |
| `screens/relationships.tsx` | 842 | 843 |
| `screens/companypeople/summary.tsx` | 790 | 793 |
| `screens/installation-settings.tsx` | 735 | 737 |
| `screens/companyraildetails.tsx` | 681 | 682 |
| `screens/contractform.tsx` | 667 | 671 |
| `screens/companycontracts.tsx` | 579 | 583 |

`companies.tsx` and `companies.test.tsx` are the same two files the waiver knew
as `organizations.tsx`; their entries were repointed at the new paths rather
than re-seeded, so the ratchet never lost sight of them.

## Done looks like

Every count above back at or below its former value, and the entries deleted
outright for any file that reaches the cap.

The two obvious places to start are the ones that are most over it and least
entitled to be. `screens/companies.tsx` is 2945 lines and exports
[...]

PR #5192, opened 2026-09-09
The record type is called company, in the schema as well as on the screen

### LINKED ISSUE #2025: The MCP surface should speak the word the product uses: company, not organization
MCP-TODO item 10. **Attempted in a batman session on 2026-08-20 and deliberately reverted** — the half-built version is worse than none, and the reason it cannot be done cheaply is itself worth recording.

## The problem

The product says **company** everywhere a person looks: sidebar, breadcrumb, page heading. The seam stores `organization`, the query grammar says organization, and several tools say **account** (`account_coverage`, `send_account_email`). Three words for one thing, and neither word a model can see is the one the user says. A real session lost a turn to `list_records {record_type: "account"}`.

## What was tried

A `canonicalRecordType()` alias resolving `company`/`account` → `organization`, applied at list, search, read, create and archive, plus `recordTypeArg` (which the contract tier floor is looked up by — an unresolved alias there would have matched no policy and been admitted at a looser tier).

A Codex review confirmed the governance was sound: tier floor, REST gate, staged-approval target types, response echo, metering and audit rows all used the canonical word. **No authority bypass.** But it found two things that make the feature not worth shipping as built:

**1. The model is never taught the alias.** The published enums still list only `organization`. This registry deliberately does not enforce enum values (`backend/internal/modules/agents/requiredargs.go:27-31`), so the alias works at runtime — but a schema-constrained MCP client treats it as invalid, and the model has no reason to ever try it. That defeats the entire purpose.

**2. Eleven other doors still refuse it.** `update_record` and `merge_records` pass the raw value to a provider with no such branch; `relink_activity`, `query_workspace` (returns `unknown_target`), `resolve_entities`, `catch_me_up_on`, `prep_for_meeting`, and the nested `links[].entity_type` on `log_activity` / `send_account_email` / `book_meeting` all reject or mis-forward it. A caller taught that `company` works on reads and then refused on update is worse off than one who learned `organization` once.

## Why it is not a small fix

Adding the alias to the enums is what makes it discoverable — and that is what the budget will not allow. `TestTheToolListingLeavesTheRunRoomInTheWindow` caps the listing at 15000 tokens of a 24000-token window, and it is already at ~14990. Adding the two words to **one** enum reached 15018; a one-line description on five enums reached 15125. The listing is never elided, so anything added there comes out of the runs own observations.

## Done looks like

Pick one, deliberately:

- **Rename the surface.** `organization` → `company` as the record type on the MCP surface only, contract-mapped to the stored table. Costs a contract change and a breaking change for existing clients, but ends the ambiguity instead of layering over it — and spends no extra tokens, because it replaces a word rather than adding one.
- **Or alias everywhere, and buy the budget first.** Every record-type door, nested links included, plus the enums — which means finding ~200 tokens elsewhere in the listing before starting.

The middle path — aliasing some doors without publishing it — is the one that was tried and rejected.

dungbanguyen:
Closes https://github.com/margince/margince/issues/2025.

One record type had two names. The product said **company** on every screen, the schema said `organization`, the tool surface said `account`, and a model asking for `record_type: "company"` was refused by a registry that admitted only `"organization"`. Nothing failed, because nothing was looking — a reader grepping for the company table found nothing and concluded there wasn't one.

This is the whole rename: schema, code, contract, tool surface, docs, file names, and a fitness function that stops the other word coming back.

## What moved

- **Schema** — one migration in `migrations/core`: 13 tables, 39 columns, 119 constraints, 45 indexes, 11 triggers, 5 functions, and 30 `UPDATE`s that move STORED values (`approval.target_entity_type`, `co_target_entity_type`, `webhook_delivery`, unpublished `event_outbox` rows). `head_catalog.txt` regenerated from the migrated schema.
- **Contract** — `backend/api/crm.yaml`: paths, schemas, enum values, operation ids. Pinned in `scripts/contract-breaking-allowlist.txt` as a ratified pre-live resync.
- **File names** — 257 paths renamed. A rename that leaves `organizationlogo.go` on disk has not been done; `git log --follow` still works.
- **Frontend** — screens, fixtures, stories, and all three locales. German `Organisation` → `Firma` (same gender, so the articles are unchanged).

## The anchor and the record

The installation's own company and a customer company are different things that were both called "organization". The anchor takes the qualified name so the record can have the plain one: `getAnchorCompany`, `SetAnchorCompanyLogo`, `anchorcompany.go`. The record keeps `getCompany`, `SetCompanyLogo`, `company.go`.

## What the census refuses

`backend/gates/companyvocabulary_test.go` reads every git-tracked file — path AND contents — and fails on the old word. It carries 34 exemptions, each with a reason, because four things in this tree spell those letters and are not this record type: a meeting **organizer**, the adjective **organizational**, a **.org** TLD (275 of them), and Microsoft's `organizations` authority alias, which is a literal path segment at login.microsoftonline.com and would break sign-in.

Two holes in it were found by it failing to fail, and both are now planted cases:

- **The word after a dot.** The predicate skipped a preceding dot so the .org exclusion could work, which made `deal.org`, `{ ...org }`, `"license.holder.org"` and `o.org_id` invisible — including **two SQL statements that selected a column their own alias list no longer declared**. No compiler and no unit test can see that. A host now must start where a host starts and end where a TLD ends.
- **A name with no boundary in it.** `csvorgwriter.go` is one lowercase run, so the line predicate found nothing to anchor on and six files kept the old name with renamed bodies. The path check now matches the letters and takes the English out first.

## Audit rows keep the old word

`trg_audit_no_mutate` makes `audit_log` immutable, so rows written before this migration carry `organization` forever. `internal/compose/auditlegacytype.go` says that once, and the two reads that filter the trail by record type match both — a miss there means the agent overwrites a human's value without staging approval.

## Verified locally

`make check-backend`, `make check-fe`, `make craft-static` and `make test-integration` (51 packages, 0 skips) all green on this tree, against `origin/main` as of the last merge.

## Left behind

Thirteen frontend file-length waivers were re-frozen upward: the identifier got four characters wider and biome resplits the prop lists. Nothing was added to those files. Paying them back down is https://github.com/margince/margince/issues/5025.

Supersedes the draft at https://github.com/margince/margince/pull/5016, which was cut from a base 118 commits behind and could not converge.


dungbanguyen:
**This PR has had no AI review, despite the green checks.** CodeRabbit posted a *passing* check alongside `Review skipped: 2135 files exceed the limit of 100`, and cubic reported `skipping`. Every green here comes from the deterministic lanes — 31/31, including `ci`, both SonarCloud checks and all six integration shards — so a human read is the only review this change will get.

Where a reviewer's time is worth most, in the order I'd spend it:

1. **The migration** (`backend/migrations/core/1788972288_*`). 30 of its statements rewrite STORED values rather than schema — `approval.target_entity_type`, `co_target_entity_type`, `webhook_delivery`, unpublished `event_outbox` rows. A missed one is a row that stops matching its own reader, silently.
2. **The anchor/record split.** The installation's own company and a customer company were both "organization"; the anchor now takes the qualified name (`getAnchorCompany`, `SetAnchorCompanyLogo`, `anchorcompany.go`) so the record can have the plain one. A call site pointed at the wrong half compiles wherever the signatures happen to agree.
3. **`internal/compose/auditlegacytype.go`.** `trg_audit_no_mutate` makes `audit_log` immutable, so pre-migration rows keep the old word forever. Two reads filter the trail by record type and must match both; a miss means an agent overwrites a human's value without staging approval.
4. **`backend/gates/companyvocabulary_test.go`'s 34 exemptions.** Each one is a claim that the word there is not this record type. Three of them were found the hard way, by something breaking after the sweep renamed a name that belonged to somebody else: SonarCloud's `sonar.organization` (which kills the scan rather than failing its gate), GitHub's runner-concurrency prose, and `migrate company-exists`, which counts rows in `workspace` and is now `workspace-exists`.

The rest is mechanical and covered by the census, which reads every git-tracked file's path and contents and fails on the old word.

dungbanguyen:
Ran the "who owns the word" pass over this diff — a record type this product defines gets renamed, a key another system defines gets quoted. One hit, and it is a live break rather than prose.

## `microsoftWorkAuthority` is now an endpoint Microsoft does not have

```diff
-	microsoftWorkAuthority = "organizations"
+	microsoftWorkAuthority = "companies"
```
`backend/internal/compose/microsoftsignin.go:90`

Microsoft's authority aliases at `login.microsoftonline.com` are **`common`, `organizations`, `consumers`** — there is no `companies`. This constant is returned by `authorityFor` (`:202`) and becomes the path segment of the authority URL, so work-account sign-in would request `https://login.microsoftonline.com/companies/oauth2/v2.0/…`, which is not an endpoint.

Its two siblings were correctly left alone, which is what makes this one stand out: `common` and `consumers` are Microsoft's words and are untouched; `organizations` is Microsoft's word too and was read as ours.

**The branch already disagrees with itself about it.** `connectorapp.go:318` still validates against the real set:

```go
case "common", "organizations", "consumers":
```

So an operator configuring `organizations` is accepted there while sign-in mints `companies`, and configuring `companies` is refused there while sign-in expects it. Whichever way round, the two halves no longer name one thing.

## The waiver records the error as the rule

`gates/companyvocabulary_test.go` protects the renamed value on the ground that it is foreign:

> Microsoft's authority alias: `companies` is a literal path segment at login.microsoftonline.com beside `common` and `consumers`, and renaming it stops sign-in working.

That sentence is true of `organizations` and false of `companies` — the rename happened first and the waiver was written to describe the result. So the one mechanism that would have caught this now asserts it is correct, in five entries (`microsoftsignin.go`, its test, `connectorapp.go`, its test, and the flag help).

This is worth more than the fix: a waiver that names a foreign owner is exactly the right instinct, and it was pointed at the wrong string. The check that would separate them is whether the value appears in the OTHER system's documentation — `common` and `consumers` do, `companies` does not.

## Also, operator-facing text

Same substitution in prose an operator reads while configuring:

- `cmd/api/config.go:115` — `sign-in cannot run on common/companies/consumers`, in the error that tells them what they may configure
- `cmd/api/config.go:114` and `cmd/worker/config.go:133` — `--graph-tenant` help, `(default: common — any company)`; `common` admits any **tenant/directory**, which is Microsoft's model rather than a company record
- `docs/reference/configuration.md` and `.env.example` carry it too, per the waiver list

## Everything else was clean

I checked the two traps I had hit on the earlier attempt at this rename and both are avoided here: `sonar.organization=margince` is intact, and `.github/` is untouched so the GitHub-org comments about the 20-concurrent runner ceiling keep their meaning.

Superseding my #5016 with this — closing it now.

PR #5196, opened 2026-09-10
A withdrawal link outlives the mail that carried it

LarsGradion:
## What this fixes

`preference_token` is one credential doing two jobs. It opens the preference centre, which **reads** a recipient's per-purpose consent state and can **grant**, and it is also what the RFC 8058 one-click unsubscribe POST carries.

Those two want opposite lifetimes. Read authority over somebody's consent record should be short, because the link sits in a mailbox forever. A withdrawal should last as long as the mail does, because a person unsubscribing from a two-year-old newsletter is exercising a right that does not expire.

Today the short lifetime wins for both: 30 days, sliding, revoked whenever the next send rotates it. So pressing unsubscribe on older mail answers "this link is no longer valid" and the only way left to stop the mail is to ask a human. That is the friction one-click unsubscribe exists to remove.

## What it does

The authority is split. `withdrawal_credential` can stop mail and can do nothing else: it cannot read a consent state, cannot grant, and names a scope it may never exceed. That narrowness is what makes a 24-month life safe, and 24 months is what the mail itself is retained for.

**Hashed at rest**, unlike `preference_token`'s plaintext, because this one outlives every rotation and sits in mail archives and proxy logs.

**Leads get an opt-out for the first time.** `preference_token.person_id` is `NOT NULL`, so a lead-only recipient's marketing mail carried no `List-Unsubscribe` header at all. An address matching no record still gets a credential, because the message is going there regardless.

**Old links still withdraw.** Every link already in a mailbox carries a preference token, so the adapter honours those for their withdrawal half and ignores `expires_at`. It refuses the three revocation reasons that are not rotations: erasure means the subject is gone, compromise means the token is in somebody else's hands, and a merged predecessor no longer receives mail.

**The page works when the link may only withdraw.** The unsubscribe page opens by fetching the preference centre, which 404s for exactly these links. That is now the withdraw-only state rather than a dead link: the stop button survives and everything needing the read is dropped.

## Gate records

Five gates asked for a decision rather than a fix, each ratified on its own evidence:

- `ResolveWithdrawalToken` in `ungatedEntryPoints`, beside the two sibling public resolves. The mint and the revoke are **not** waived and both take `person:update`.
- Two cross-module erasure writes, for erasure and for anonymize.
- `withdrawal_credential` as `notAnEdge` in the search join census, the verdict `communication_suppression` already carries.
- Unscoped row-reference ceiling 100 to 104, for the public resolves and the send-path address lookup, neither of which has a seat to narrow to.

## Things the gates caught in my own work

- Erasure was wired and anonymize was not. This row matters most of the four capabilities on that path because it carries the address in its own column, which outlives the `person_id` an anonymize-in-place nulls.
- The address lookup was written as a `LIMIT 2` with a single-row `Scan`, which silently takes the first match. That is exactly the behaviour the surrounding comment claimed to refuse.
- Two comments claimed something was the only one of its kind. The first was false; the second was true but held by no test.

## Verification

Full backend gate and full frontend gate both green. Ten tests on the credential, ten on the unsubscribe page. Mutation-verified: honouring every revocation reason fails the refusal test, honouring `expires_at` fails the legacy test, and checking the lead before the person fails the precedence test.

---

## Two review rounds, sixteen findings

Both rounds found real defects. The second found that several of my fixes from the first round had caused new ones, which is the part worth reading.

**I got the core invariant backwards, twice.** The table started with a unique index over live rows: one subscription, one credential. That cannot work, because the token is hashed and a second send has nothing to reuse. My first fix returned an empty token, and the send path reads that as "no unsubscribe surface" and ships no header at all. My second revoked the covering credential and minted fresh, which killed the link in the message the recipient already had. The older mail is exactly the one somebody presses. Every send now mints its own credential and revocation is by subject.

**One fix would have broken production.** The migration required every preference-token revocation to state a reason, and the rotation writer set only the timestamp. Every marketing send that rotated a token would have aborted. My test missed it because the fixture wrote the reason by hand instead of driving the real writer.

**One of my tests proved nothing.** It called the store method directly and passed with the routing guard disabled. It drives the handler now and fails when the guard is removed.

Other findings fixed: an all-marketing link could name business correspondence in the query string and stop it; hiding the purpose names also hid whether anything changed, so a successful first press told the recipient "nothing changed"; a lead's link resolved and then answered 404; a named-purpose lead link stopped all marketing; concurrent presses were unserialised; a press racing an erasure wrote the erased address back; and erasure missed lead-bound and address-only rows.

## Verification

Full backend gate and full frontend gate both green. Twelve tests on the credential, ten on the unsubscribe page. Mutation-verified: the revocation-reason refusal, the expiry-ignoring legacy adapter, person-over-lead precedence, the class scoping, and the routing guard.


dungbanguyen:
This broke the agent send path and `main` has been red on it since. Bisected to this commit:

```
14192f373  fail       ← this PR
49ff4900f  pass
```

```
--- FAIL: TestPreferenceCenterOptOutBlocksAgentSend
    preference_agent_integration_test.go:97: granted agent send → person.update: permission denied, want success
```

## The gate is right and its placement is not

```go
// GATED, unlike the resolve below. This MINTS a bearer credential that can
// stop somebody's mail for two years, so the caller has to be entitled to
// act on that person …
if err := auth.Require(ctx, entityPerson, principal.ActionUpdate); err != nil {
```

That reasoning holds for a **deliberate** mint. `EnsureWithdrawalCredentialTx` is not reached by one: its only in-tree caller is `WithdrawalTokenForEmail`, which is the **send path's** seam — the unsubscribe link RFC 8058 compels every marketing message to carry.

So the grant now required is one the sender need not hold. An agent sending on a seat's behalf carries send-adjacent grants and not `person:update`, and the message it is sending is exactly the one that must have a working opt-out.

Two things make the placement read wrong rather than strict:

1. **There may be no person to have authority over.** `WithdrawalTokenForEmail`'s own doc: *"An address matching NOBODY still gets a credential. The message is going to that address regardless."* A grant justified as authority over a person is asked for on a path that explicitly serves addresses matching none.
2. **The gate currently protects nothing.** With one caller, and that caller the send, there is no deliberate-mint route for it to stand in front of — it only refuses sends.

## Not fixing it myself

Removing an auth gate in this module on my own reading is not a call I should make, and #5103 taught me the author's own follow-up beats my guess. Three shapes look plausible and the choice is yours:

- the send path gets its own entry point whose admission is the send's;
- the requirement moves to a deliberate-mint route when one exists;
- or it stays and the send principal is given the grant, which I would argue against — it widens what every sending agent may do to a person record, to buy a link the law compels.

**It blocks the queue**: this runs in the compose integration shards, so every open pull request fails on it. Two other `main`-reds are up as #5225 and #5237; this is the third and the only one I have not fixed.

dungbanguyen:
Adding the fact that decides between the readings on the table, from another
session working the sub-2000 range — our queue is behind this and I did not want
to guess at your intent.

**The mint runs under the sender's own principal, on the ordinary send path.**
Traced the whole way, and there is no system escalation anywhere in it:

```
activities.deliverability                    (inline in the send)
  → UnsubscribeLinker.UnsubscribeToken
  → compose.preferenceLinkAdapter
  → consent.WithdrawalTokenForEmail
  → consent.EnsureWithdrawalCredentialTx     ← auth.Require(person, update)
```

So the grant is demanded of whoever is sending marketing mail, not of an
operator deliberately minting a credential.

**And the path explicitly serves addresses matching no person.** Two places say
so in their own words. `WithdrawalTokenForEmail`: *"An address matching NOBODY
still gets a credential. The message is going to that address regardless."*
`deliverability`: *"ok is false when the address carries no unsubscribe surface
— a locked (transactional) purpose, or an address no person holds."*

Put together: `person:update` is asked before the code knows whether a person is
involved, on a path whose documented job includes the case where none is. That
reads as a placement problem rather than a strictness one — which is the same
conclusion reached from the other side, arrived at independently.

**What is NOT weakened by moving it.** The row-scoped check below is untouched
and is the one that encodes authority over a named person:

```go
if !in.PersonID.IsZero() {
    auth.EnsureWritableLive(ctx, tx, entityPerson, in.PersonID.UUID)
}
```

Its comment is explicit that a caller holding only a read share must not mint,
and that a plain probe would admit an anonymized tombstone. So the erasure race
and the read-share case both stay closed however the object gate is placed.

**The one thing I cannot answer, and why I have not touched it:** whether a
marketing sender is *supposed* to hold `person:update`. If yes, the fixture in
`TestPreferenceCenterOptOutBlocksAgentSend` is what is stale and the gate is
right. If no, the gate refuses a send RFC 8058 compels a header on. That is a
product call about who may send, and it is yours.

I did note that a conditional — `if !in.PersonID.IsZero()` around the Require —
would honour the stated reason (*"a credential minted for a person is authority
over them"*) without asking about a person who is not there. It does **not**
unblock the failing case on its own, since that recipient is a person; recording
it so it is not re-derived.

Not editing this myself: it is an auth gate in a module I do not own, and the
last time a session guessed at one of these the author's own follow-up was better
than the guess.

PR #5201, opened 2026-09-10
A name read off a website is a record, not somebody's work

LarsGradion:
Lars found leads in his Morning Brief that nobody had asked him about. They came from the deep-read pass: names published on companies' own websites, staged for review, and accepted by him.

Accepting one made him its owner. That put a stranger in his "owes a reply" lane and started a first-response clock against somebody who had never written to anyone. Two automations then turned the same row into work — a follow-up task, and an owner assignment.

## What changed

**Accepting records that the person exists, and nothing more.** The lead is created unowned and stays that way until a human picks it up in Leads. `LeadFields.Unowned` carries that intent; the default is unchanged, because a captured lead normally *should* belong to the human whose connector found it — the connector's replay is a write, and `auth.EnsureWritable` refuses an ownerless row. A site lead has no replay behind it, which is what makes declining ownership safe here.

**Both automations decline it, not just the task one.** Gating `route_lead` alone would have left a configured `assign_lead_owner` rule handing the lead out anyway, and that one stamps `routed_at`. They read the rule from `shared/kernel/leadsource` rather than each spelling it: they run on one event with no ordering between them, and two copies would disagree the day one of them learned a new source.

**Taking on an unowned lead starts its clock.** The deadline reads `COALESCE(routed_at, created_at)`, so without a stamp a lead that sat unowned for a week was a week overdue the instant somebody picked it up — they inherited a breach earned while the row was nobody's. A reassignment between people keeps the original clock.

**`siteread` joins the lead-source vocabulary** weighted low, like `crawl`. Without a row it resolved to the neutral default and scored like a hand-typed lead. The migration seeds it and the scorer fixture matches, which the existing parity test holds.

## Tests

Four behaviours, each mutation-checked against the change it covers:

- an accepted site lead is owned by nobody and has no clock running
- `route_lead` fires for every lead a person asked for and declines a website read or a crawl
- `assign_lead_owner` does the same — tested against a **configured** instance, with a premise guard proving an unconfigured one declines everything regardless, so the source cases are not passing vacuously
- assigning an unowned lead stamps `routed_at`, and reassigning between people does not move it

`lead_update.go` crossed the 500-line cap, so the response-clock concept moved to its own file rather than taking a waiver.

---
## Summary by cubic
Accepting a lead found on a company's website now records that the person exists without making the accepter its owner. Previously, accepting assigned ownership, which put a stranger in the accepter's "owes a reply" lane and started a first-response clock against someone who never contacted anyone.

- Both automations (`route_lead` and `assign_lead_owner`) decline passively discovered sources (`siteread`, `crawl`), reading the rule from a shared kernel package so they stay in sync.
- Taking on an unowned lead starts its response clock at `routed_at`, whether through an owner update or the "Take ownership" button; reassigning between owners keeps the original clock.
- The SLA breach scan skips unowned leads, so it no longer stamps `sla_breached_at` against a deadline nobody is answerable for.
- Site-lead accepts replay against a read probe instead of `EnsureWritable`, which refuses the ownerless row by design.
- `siteread` is seeded as a low-intent source; a migration inserts it and runs on deploy.

<sup>Written for commit 6592e3c973c78db385b974c7c719af82c861a1e8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/5201?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Website-read leads are now identified as a low-intent source.
  * Leads discovered through website reads or crawls remain unowned until a human takes ownership.
  * Response timing begins when an unowned lead is first assigned to a human and remains unchanged during reassignment.

* **Bug Fixes**
  * Passive discovery leads no longer enter lead-routing workflows intended for direct requests.
  * Automated status changes do not incorrectly count as a human first response.


dungbanguyen:
This landed with two tests failing and \`main\` has been red since. #5225 fixes it.

\`AND owner_id IS NOT NULL\` in the breach scan undoes #5015 (*"A breach on a lead nobody owns reaches somebody"*), whose two tests this PR did not touch:

\`\`\`
--- FAIL: TestAnOwnerlessBreachReachesTheConfiguredIntakeSeat
--- FAIL: TestAnIntakeSeatThatCannotWorkIsTreatedAsUnset
\`\`\`

**Both changes are right about their own case.** The fixtures are what separate them: #5015 seeds `source: 'inbound'` — a person who wrote to us and nobody picked up — and this PR seeds `source: 'siteread'`, a name a pass read off a website. Ownership does not tell those apart, so excluding unowned rows excludes both.

`lead_source.intent` does, and in this repository's own words: siteread and crawl are low *"for crawl's reason: nobody asked us for anything."* #5225 reads that instead, so an unowned inbound still escalates to the intake seat while a scraped name does not, and a prospecting source an operator adds gets the rule with it.

**Your stated reason for the exclusion was correct** and is closed rather than argued away:

> escalating a row nobody has taken on stamps `sla_breached_at` — which then suppresses the real escalation once a human takes the lead on.

Exactly so, because `routed_at` restarts the clock while the stamp keeps the row out of the scan for good. The take-on now clears the stamp with the same condition that starts the clock, on both the claim and the update path.

The `routed_at` half of this PR is untouched and is the better half — it is what makes an inherited breach impossible in the first place.

PR #5211, opened 2026-09-10
A public request is a case with a receipt

LarsGradion:
## What was wrong

A correction or erasure a data subject sent through their confirm link landed in `person_confirm_submission` and stopped there. That table records what was sent. It is not work anybody owns.

Nothing gave the request a deadline. Nothing put it in the queue the DPO works through. The subject was handed no reference to chase it with. The statutory answer was owed from the moment it arrived.

## What changed

`openRightsCaseTx` opens a `data_subject_request` in the same transaction as the proposal it came from, and answers the reference the subject quotes. A correction opens a rectify case, an erasure request an erasure case, and a marketing answer alone opens none.

The submit endpoint returns those receipts. The confirm page shows them under the thank-you, with the month the answer is due within.

## Two things worth reading closely

**The deadline clamp.** One calendar month is not thirty days, and `AddDate` alone gets it wrong: Go rolls an impossible date forward, so 31 January plus one month came back as 3 March and quietly gave the subject three extra days. `oneCalendarMonthAfter` clamps to the last day of the target month. `TestOneCalendarMonthFollowsTheShortMonths` caught this in my own code.

**No auth gate on the case writer, deliberately.** The caller is the public confirm edge running as `system:public_confirm`, which holds no person grant and must not. A bearer token proving a mailbox is the authority to say what you are asking for, and nothing more. What bounds the call is that it is reachable only after `spendConfirmTokenTx` consumed the link.

## Details

`subject_ref` carries the person id alongside `person_id`, because `FulfilErasure` parses `subject_ref` to find the record it must erase. The two would otherwise disagree on exactly the cases this opens.

A replayed submit opens no second case. The unique index on `source_submission_id` decides it, and the writer reads the standing case back so the replay answers the receipt it already earned.

## Contract change

The submit endpoint moves from 204 to 200, which oasdiff reports as breaking. It is our own anonymous page endpoint, both frontend callers gate on `response.ok`, and a receipt has nowhere to live in a 204. Merged with `CONTRACT_STABILITY=pre-live`.

## Unrelated fix carried here

`withdrawal_credential` shipped in #5196 with `person_id` and `lead_id` unclassified, leaving the foreign-key visibility test red on main. Both are server-derived from the address a send is addressed to, and are now classified. That test is integration-only, and `make check-backend` never runs it.

## Proof

Six new tests, two mutation-verified. Full `make check-backend`, `make check-fe` and the migrations integration suite green.

---
## Summary by cubic
A correction or erasure a subject sends through their confirm link now opens a `data_subject_request` in the DPO's queue, in the same transaction as the submission, with a one-month deadline and a reference to quote when chasing it. Previously the request landed in `person_confirm_submission` and stopped there — nobody owned it, nothing started the clock, and the subject got no receipt.

**What changed**
- A correction opens a rectify case, an erasure request an erasure case, and a marketing answer alone opens none.
- The submit endpoint now returns receipts (200 instead of 204) and the confirm page shows them under the thank-you with the due month.
- A replayed submit opens no second case; the unique index on `source_submission_id` returns the receipt the first submit earned.
- The deadline is one calendar month from receipt, clamped to the last day of a short month and measured by month, not day, so a DST shift reads as no overflow.
- A colliding receipt reference is redrawn in savepoints, up to three times; losing that draw used to roll back the proposal, the case, and the spent token together.

**Worth reading closely**
- No auth gate on the case writer, deliberately: the public confirm edge holds no person grant, and the spent link is what bounds the call.
- `source_submission_id` deliberately has no foreign key: an erasure deletes the submission, and any reference would wait on the case row `FulfilErasure` holds locked.
- The 204-to-200 change reports as breaking in oasdiff, but both frontend callers gate on `response.ok`; merge with `CONTRACT_STABILITY=pre-live`.
- Carries an unrelated fix: `withdrawal_credential.person_id` and `lead_id` are now classified as server-derived, repairing the foreign-key visibility test on main.

<sup>Written for commit 8af442416f7c34fd6e3ec05de113e821d4aecfba. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/5211?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


dungbanguyen:
\`make craft-static\` is red on \`main\` with two BLOCKERs from this change, both the same shape:

```
backend/internal/modules/consent/rightscase.go:202
backend/internal/modules/consent/rightscase_integration_test.go:270
  [BLOCKER] swallowed-errors — blank-assigning a call result discards a possible error
```

`rightscase.go:202` is the savepoint rollback on the error path:

```go
if err != nil {
    _ = nested.Rollback(ctx)
    return ids.UUID{}, "", false, fmt.Errorf(...)
}
```

Found while fixing an unrelated main-red (#5225), so this is a drive-by report rather than a fix — the rollback's own failure is a real thing to decide about (it usually means the connection is already gone, which the returned error is about to say anyway), and that decision is yours.

Worth noting **why it reached `main`**: the pre-push hook and CI run `craft static` diff-scoped, so a file this change added is judged against what it added — while `make craft-static` is whole-tree. The two disagree here, which is the gap a whole-tree sweep exists to catch and nothing runs on a schedule.

Either handle the rollback's error or waive it in source with the reason (`//craft:ignore swallowed-errors <why>`); the rubric takes a reasoned waiver and refuses a reasonless one.

dungbanguyen:
Correcting what I said above: I wrote that the pre-push hook is diff-scoped so these would not gate this change. **They gate everyone.** CI's craftsmanship job runs `make craft-static` whole-tree, so both BLOCKERs fail every open pull request in the repository, not just one touching that file.

Fixed in #5229 rather than left for you, since it blocks the queue. The store joins the savepoint rollback's error to the insert's; the test's rollback now fails the test, which is the sharper of the two — that transaction is never committed, so the rollback is the cleanup and a failure to roll back is a leaked connection the next case waits on.

The `routed_at` work in this PR is untouched and stands.

PR #5215, opened 2026-09-10
fix(contract-api): retention.applied publishes a closed action set

### LINKED ISSUE #1769: retention.applied declares action as an open string where ADR-0117 closed the set

dungbanguyen:
Closes #1769. Both halves: the enum, and a gate holding it against the CHECK.

## The gap

ADR-0117 closed `retention.applied`'s `action` to `archive | anonymize | erase`.
The contract carried the three **in prose** and the schema admitted any string,
so the generated Go and TypeScript followed the schema and nothing refused a
fourth value. A subscriber reading that sees an open field, which leaves two bad
choices: branch on values nobody has defined, or drop what it does not recognise
— and a silent drop reads exactly like no event having been emitted.

`retention.restricted.action`, right beside it in the same file, was already a
proper enum. This was an oversight rather than a design choice.

## Named constants, deliberately

`archive`/`anonymize`/`erase` generate `Archive`/`Anonymize`/`Erase`, and
crm.yaml's `RetentionAction` already holds those three names in the same package.
A collision there **silently renames that enum's constants** and breaks modules
with no connection to this one, so the values take `x-enum-varnames` — the idiom
this contract already uses for `meeting_status`, with the same reason recorded
beside it.

## The compiler holds the emit sites now

`retentionAppliedPayload` takes `PublicEventRetentionAppliedAction` rather than a
string, so a fourth action cannot reach a subscriber by being spelled at an emit
site — which is the ticket's failure scenario: *"an emit site that passes the
value straight into the event payload bypasses [the CHECK]"*.

Five of the six sites pass a constant. The sixth takes its action off a **row**
(`retentionactions.go`, the policy-driven sweep) and cannot be held by the
compiler, so it checks `Valid()` and refuses with the value named. The CHECK
bounds what can be written today; a row written before a value was retired, or a
migration that widened the CHECK without the contract, is what that guard is for.

## The gate — the issue's second half

`TestTheRetentionActionSetIsOneSet`. The set is spelled in three places and
hand-written in all of them: the CHECK at schema head, `RetentionAction` in
crm.yaml, and this event's enum.

It asserts **equality, not containment**, because the three directions fail
differently:

- a value in the database and not on the event ships an event every subscriber
  drops in silence;
- a value on the event and not in the database promises subscribers something
  nothing can produce, so a consumer writes a branch that never runs;
- a value in the API and not the database is a request accepted and then refused
  by a constraint.

Read from the schema **head catalog** rather than a migration, since a later one
may have moved the CHECK and the head is what a running installation has. Both
readers refuse rather than pass when they find nothing to compare.

Deriving one spelling from another was the alternative and there is nowhere to
derive from: the CHECK is SQL in a shipped migration, and the generators resolve
no cross-document reference between crm.yaml and public-events.yaml.

Mutation-checked both ways: adding `redact` to the event enum fails it against
the CHECK, and removing the enum entirely — the original defect — fails it as an
open field.

## No migration

Confirmed, per the decision note: this is not a Postgres enum. The house pattern
is `CHECK (col IN (...))`, which `retention_policy.action` already has and which
already enforces the three values correctly.

## Checks

- `make check-backend` green; gate inventory regenerated.
- `craft static --strict` over the diff: 0/0/0.
- `privacy` unit and integration lanes green.

## Summary by CodeRabbit

- **API Contract**
  - Retention-applied event actions are now limited to `archive`, `anonymize`, and `erase`.
  - Added clearly defined action values for supported integrations and generated clients.
  - Invalid or unsupported retention actions are rejected instead of being published.

- **Consistency**
  - Retention action definitions are now aligned across public APIs, frontend types, and database validation.

- **Documentation**
  - Clarified that restriction events use the separate `retention.restricted` event rather than a retention-applied action.

### DIFF
--- a/backend/api/public-events.yaml
+++ b/backend/api/public-events.yaml
@@ -2892,7 +2892,24 @@ components:
       properties:
         action:
           type: string
-          description: The action that ran (archive | anonymize | erase).
+          enum: [archive, anonymize, erase]
+          # Named explicitly: these three generate `Archive`, `Anonymize` and
+          # `Erase`, which crm.yaml's RetentionAction already holds — and a
+          # collision there silently renames THAT enum's constants and breaks
+          # modules with no connection to this one.
+          x-enum-varnames:
+            - RetentionAppliedArchive
+            - RetentionAppliedAnonymize
+            - RetentionAppliedErase
+          description: >-
+            The action that ran. A CLOSED set, so a subscriber can switch on it
+            exhaustively — which is the whole reason for closing it: an open
+            field leaves a consumer to either branch on values nobody has
+            defined or drop what it does not recognise, and a silent drop looks
+            exactly like no event. `retention_policy.action` carries the same
+            three under a CHECK, and a gate holds the two spellings together.
+            Restriction is NOT a fourth action here: it emits
+            retention.restricted, whose own action is a different closed set.
         policy:
           type: string
           format: uuid
--- a/backend/gates/corepicklistcontract_test.go
+++ b/backend/gates/corepicklistcontract_test.go
@@ -226,13 +226,22 @@ func descendContract(t *testing.T, doc map[string]any, path ...string) map[strin
 
 func loadContractDocument(t *testing.T) map[string]any {
 	t.Helper()
-	raw, err := os.ReadFile(contractDocument)
+	return loadOpenAPIDocument(t, contractDocument)
+}
+
+// loadOpenAPIDocument reads one OpenAPI document as untyped nodes, for the
+// readers above. The path is a parameter because the product publishes two —
+// crm.yaml to callers and public-events.yaml to subscribers — and a set spelled
+// in both is read here twice rather than by two readers.
+func loadOpenAPIDocument(t *testing.T, path string) map[string]any {
+	t.Helper()
+	raw, err := os.ReadFile(path)
 	if err != nil {
-		t.Fatalf("reading %s: %v", contractDocument, err)
+		t.Fatalf("reading %s: %v", path, err)
 	}
 	var doc map[string]any
 	if err := yaml.Unmarshal(raw, &doc); err != nil {
-		t.Fatalf("parsing %s: %v", contractDocument, err)
+		t.Fatalf("parsing %s: %v", path, err)
 	}
 	return doc
 }
--- a/backend/gates/retentionactionset_test.go
+++ b/backend/gates/retentionactionset_test.go
@@ -59,7 +59,19 @@ func retentionActionCheck(t *testing.T) []string {
 		}
 		out := make([]string, 0, len(values))
 		for _, value := range values {
-			out = append(out, value[1])
+			// SQL doubles a quote to escape it; the stored value has one.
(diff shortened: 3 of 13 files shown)


dungbanguyen:
Good catch, and it was the worst direction for this gate to be wrong in — fixed.

`'([a-z_]+)'::text` skipped any value carrying a hyphen, a capital or a space,
and the three it *could* read would still match the contract's enums. So the gate
would have reported agreement over exactly the divergence nobody could parse.
Under-recognition is the one way a census must not break, and this one had it.

Two changes rather than one:

1. The pattern reads any SQL literal, doubled quotes included, and unescapes it —
   your proposed `'((?:''|[^'])*)'::text`.
2. **It now refuses rather than reading short.** Removing every literal it parsed
   must leave no quote behind on the line. A CHECK rendered in a form this reader
   does not know — without the `::text` casts, say — fails loudly instead of
   quietly shrinking the set it compares. The regex fix closes the case you
   named; this closes the shape of it.

Mutation-checked both: `'legal-hold'::text` now reaches the comparison and fails
as a divergence, and `'legal hold'` with no cast fails as a literal the reader
cannot see.

PR #5225, opened 2026-09-10
fix(leads): a lead nobody owns breaches again, unless nobody asked us for it

dungbanguyen:
**`main` is red.** Two integration tests fail on tip, and `main-health` has not run since `7a246a72`:

```
--- FAIL: TestAnOwnerlessBreachReachesTheConfiguredIntakeSeat
    leadsla_integration_test.go:613: reading the breach's escalation target: no rows in result set
--- FAIL: TestAnIntakeSeatThatCannotWorkIsTreatedAsUnset
```

## What happened

- **#5015** — *"A breach on a lead nobody owns reaches somebody"* — made an ownerless breach escalate to the configured intake seat, with those two tests.
- **#5201** — *"A name read off a website is a record, not somebody's work"* — added `AND owner_id IS NOT NULL` to the breach scan, and did not touch either test. They have been red since.

## Both changes are right about their own case

The fixtures say it plainly:

| | source | what it is |
|---|---|---|
| #5015 | `inbound`, `human:x` | a person who wrote to us and nobody has picked up |
| #5201 | `siteread`, `agent:siteread` | a name a pass read off a company's website |

The first is the queue's worst case and must escalate. The second is not somebody's work and its clock was never started. **Ownership does not tell those apart** — it excludes both, which is why one PR's fix broke the other's tests.

## The discriminator the product already has

`lead_source.intent`, and it says so in its own words:

```go
// A name published on a company's own website, read by the deep-read pass.
// Low for crawl's reason: nobody asked us for anything.
"siteread": SourceIntentLow,
```

So: an unowned lead breaches unless its source intent is `low`. Read from the catalog rather than naming `siteread` and `crawl`, so a prospecting source an operator adds gets the rule with it. An **owned** lead is unconditional — somebody took it on, so the clock is theirs however it arrived.

Each half of the condition is required by a different PR's test:

```
as written                          all three pass
intent check removed                #5201's test fails
back to owner-only (main today)     both of #5015's fail
```

## The thing that made the exclusion look necessary is closed, not argued away

#5201's comment was **right**:

> escalating a row nobody has taken on stamps `sla_breached_at` — which then suppresses the real escalation once a human takes the lead on.

Because the take-on restarts the clock through `COALESCE(routed_at, created_at)` while the stamp keeps the row out of the scan **for good**. So the take-on now clears the stamp with the same condition that starts the clock, on both the claim and the update path, and the two move together.

`TestTakingOnABreachedLeadClearsTheStampWithTheClock` pins it; removing the clear fails it with *"the new owner's clock restarted and the scan will skip this row for good, so they can never breach on it."*

## Also

`leadsla.go` was at 498 of its 500-line ceiling, so the sweep moves to `leadslascan.go` — what an SLA state **is** on one side, and the pass that acts on it (the half with the judgement in it) on the other.

## Verification

`internal/modules/people` fully green (36s), `make check-go` green.

`make craft-static` reports two BLOCKERs in `internal/modules/consent/rightscase.go` — **not this diff**, and already on `main` from #5211. Flagged separately; the pre-push hook is diff-scoped so they do not gate this.

Found by @target-ce, who flagged it rather than racing a fix. Their calendar hypothesis was worth checking and does not hold — the target is a plain 240 minutes with no business-day logic, and today is a Thursday.

### DIFF
--- a/backend/internal/modules/people/lead_update.go
+++ b/backend/internal/modules/people/lead_update.go
@@ -391,6 +391,11 @@ func buildLeadPatch(current crmcontracts.Lead, in UpdateLeadInput) (*storekit.Pa
 		// condition: no owner yet, and no clock yet.
 		if current.RoutedAt == nil && current.OwnerId == nil {
 			p.Set("routed_at", nil, leadSLAClock().UTC())
+			// And the breach stamp goes with it, for the reason
+			// startLeadResponseClockTx clears it on the claim path: a row that
+			// breached while it was nobody's is already escalated, and leaving
+			// the stamp would keep this owner's lead out of the scan for good.
+			p.Set("sla_breached_at", nil, nil)
 		}
 	}
 	return p, resumeRecompute, nil
--- a/backend/internal/modules/people/leadresponseclock.go
+++ b/backend/internal/modules/people/leadresponseclock.go
@@ -76,7 +76,11 @@ func startLeadResponseClockTx(ctx context.Context, tx pgx.Tx, id ids.UUID) error
 		return err
 	}
 	if _, err := tx.Exec(ctx,
-		`UPDATE lead SET routed_at = $2
+		// The breach stamp goes with the clock. A lead that breached while it
+		// was nobody's has been escalated to the intake seat already; starting
+		// this owner's clock without clearing it would leave a row that can
+		// never breach again, because the scan skips anything already stamped.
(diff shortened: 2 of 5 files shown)


dungbanguyen:

dungbanguyen:
current red — cluster B here, the lead SLA breach scan (leadsla.go, leadresponseclock.go and the same file split, which 5224 names leadslabreach.go).

`aa61b6f82`.

PR #5226, opened 2026-09-10
fix(consent): a dead preference link says so once again

luit-nfq:
**Closing: this claim is stale. The red it covers was already fixed on `main` by #5214, which merged before this branch was cut.**

## What the claim said

`e2e/public-preferences.spec.ts:144:3 › the pages a message links to › a dead link says so once, and offers no button`, red on the `uat (AC screens + axe WCAG 2.2 AA + 390px, mocked)` lane since #5196.

## What is actually true

#5196 deliberately changed the behaviour: **a 404 on the READ is the withdraw-only state, not a dead link.** A withdrawal credential outlives the preference token that rode with it and carries no authority to read a consent state, so the page's opening fetch 404s for exactly the links that still work — the ones in old mail, which is the case one-click unsubscribe exists for. The page therefore keeps the button and drops everything that needed the read. The POST is its own authority and refuses a credential that really is dead.

The spec that asserted the old dead-link sentence was left stale by #5196, and #5214 (`fix(e2e): a 404 on the read is the withdraw-only state, and the spec says so`) made the spec follow the product: it replaced that test with *"a link that may only withdraw keeps its button, and the press still stops the purpose"*.

There is no gone-vs-withdraw-only split to restore. The read wire cannot carry one — the credential that may only withdraw is refused on the read for the same reason a genuinely dead one is, and the only place the two are distinguishable is the POST, which already refuses a dead credential.

## Evidence

- This branch is based on `5d732ed96`, four commits **ahead** of #5214's merge `981dd835c` — the fix is in the base.
- `pnpm exec playwright test e2e/public-preferences.spec.ts` — **8 passed**, on an unmodified tree.
- `pnpm exec vitest run src/screens/unsubscribe` — **10 passed**. The split is already a spec in unit tests too: *"still offers the stop button when the link may only withdraw"* and *"withdraws from the read-less page"*.
- On the latest `main` run, `uat (AC screens + axe WCAG 2.2 AA + 390px, mocked)` reports **success**. The remaining red on `main` is the integration shards plus craftsmanship, a different cause, already claimed by #5224.

No product or test change is warranted. Writing one would re-introduce the dead-link branch that #5196 removed on purpose and would red three tests that now hold the correct behaviour.

Closed unfixed, per the rulebook's rule that a superseded claim closes itself pointing at the winner: #5214.


luit-nfq:
Superseded by #5214, which merged before this branch was cut and is already in its base. The uat lane is green on main; the remaining red there is the integration/craftsmanship cause claimed by #5224. Releasing this claim so the lane is not held by a red that no longer exists.

PR #5229, opened 2026-09-10
fix(consent): the rights case does not discard its rollback's failure

dungbanguyen:
**`make craft-static` is red on `main`**, so the craftsmanship job fails on **every open pull request** — mine included. Two BLOCKERs, both from #5211:

```
backend/internal/modules/consent/rightscase.go:202
backend/internal/modules/consent/rightscase_integration_test.go:270
  [BLOCKER] swallowed-errors — blank-assigning a call result discards a possible error
```

## Why it reached main

CI's craftsmanship job runs `make craft-static` **whole-tree**; the pre-push hook runs `craft static --strict` **diff-scoped**. A file this change added is judged against what it added, and the two disagreed here. (I told #5211 the hook shape meant it would not gate anyone — that was wrong, and this is the correction: it gates everyone.)

## The store

```go
if rbErr := nested.Rollback(ctx); rbErr != nil {
    err = errors.Join(err, fmt.Errorf("rolling the savepoint back: %w", rbErr))
}
```

The insert's failure is what the caller needs; the rollback's rides with it rather than being dropped. A rollback that fails usually means the connection is already gone, which turns a legible constraint violation into a confusing one further up. Joined rather than logged because this store carries no logger — and `errors.Is` still reaches the insert's own sentinel through the join, which `isReceiptCollision` next door depends on.

## The test, which is the sharper of the two

```go
defer func() { _ = tx.Rollback(context.Background()) }()
```

That transaction is **never committed** — the case it opens is the subject and is read back inside it — so the rollback *is* the cleanup. A failure to roll back is a leaked connection the next case in the lane would wait on, discarded in a defer where nothing would ever say so. It fails the test now.

## Verification

`make craft-static` — 0 blocker, 0 major, 0 minor, whole tree. `make check-go` green. `internal/modules/consent` integration lane green.

Found while fixing an unrelated main-red (#5225). Both are `main` restorations rather than feature work, and either can land first.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error reporting when a consent case cannot be saved, including failures that occur while rolling back the transaction.
  - Test cleanup now reports transaction rollback failures instead of silently ignoring them.

### DIFF
--- a/backend/internal/modules/consent/rightscase.go
+++ b/backend/internal/modules/consent/rightscase.go
@@ -199,7 +199,15 @@ func attemptRightsCase(ctx context.Context, tx pgx.Tx, kind string, personID ids
 		oneCalendarMonthAfter(receivedAt), submissionID, receipt,
 	).Scan(&caseID, &stored, &created)
 	if err != nil {
-		_ = nested.Rollback(ctx)
+		// The insert's failure is what the caller needs, and the savepoint's
+		// rollback rides with it rather than being dropped: a rollback that
+		// fails usually means the connection is already gone, which turns a
+		// legible constraint violation into a confusing one further up. Joined
+		// rather than logged because this store carries no logger, and errors.Is
+		// still reaches the insert's own sentinel through the join.
+		if rbErr := nested.Rollback(ctx); rbErr != nil {
+			err = errors.Join(err, fmt.Errorf("rolling the savepoint back: %w", rbErr))
+		}
 		return ids.UUID{}, "", false, fmt.Errorf(
 			"consent: opening the rights case this request owes an answer to: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


dungbanguyen:
Superseded by #5230, which is better on both findings — we reached the same two blockers within minutes of each other and theirs is the one to take.

**The store.** Mine gave a generic reason (a failed rollback usually means the connection is gone). #5230 gives the specific one, and it is verified against the code above it: the savepoint exists so a receipt collision can be **retried**, so a rollback that did not take is the single condition under which the retry loop cannot work — every remaining attempt then fails identically with the first attempt's message. That is a sharper claim and a truer one.

**The test, where mine was actually wrong.** I wrote a bare `defer` with `t.Errorf` and no exception. #5230 uses `t.Cleanup` with `errors.Is(err, pgx.ErrTxClosed)`, which is the spelling that suite already uses — and without that exception a clean run whose transaction is already finished fails the test for no reason. Mine had a latent false failure in it.

Closing this. Nothing here is worth carrying over.

PR #5237, opened 2026-09-10
fix(attention): the planned lane's cases say what the lane now does

dungbanguyen:
**`main` is red on three of them**, and `make test-it DIR=backend/internal/compose` reproduces on a clean checkout of `36a33cb6`:

```
--- FAIL: TestAnOverdueTaskReachesThePlannedLane
--- FAIL: TestATaskDueExactlyAtTheBoundaryBelongsToTomorrow
--- FAIL: TestTheWorklistsDayEndsAtTheInstallationsMidnight
```

## Cause

#5210 — *"The task lane shows what is coming, not only what is owed"* — made the lane take **two bounded reads** and tag each row with `due_group`. It touched thirty files and did not touch these three, which all assert the narrower lane: *"want only the task actually due"*, *"want only the task due before the day ends"*, *"want only tonight's work"*.

The change is right and its title says exactly what it did. The cases are what went stale.

## Each keeps its subject

**The overdue one** pinned "both halves of the filter: what is due is carried, and what is done or still ahead is not." The second half is what #5210 overturned; the third — a task already **done** is carried by neither read — still holds and is now asserted alongside the two that are carried.

**The two boundary ones** are about which *side* of midnight a task falls on, and for the second, that it is the **installation's** midnight rather than UTC's. #5210 kept that exactly; it moved where it shows. Both rows now reach the lane, and `due_group` says which run each belongs to — which is #5210's own stated contract:

> `OpenAndDueBy` is exclusive and the new `OpenAndDueAfter` is inclusive of the same instant, so a task due at exactly tomorrow 00:00 is outside today's read and inside the upcoming one.

Mutation-verified: putting midnight back on today's side fails it again with *`"Due exactly at midnight" reads as "tomorrow", want "today"`*.

## One thing I deliberately did not assert

The overdue case checks **overdue-or-not**, not a named group. Which of `today`, `tomorrow`, `this_week`, `later` a task seven days out lands in depends on what day the suite runs — I wrote `later` first and it came back `this_week` on a Thursday. A case pinning one of them is a case the calendar breaks, which is the failure this whole stretch has been about.

## Still red on main, not this

`TestPreferenceCenterOptOutBlocksAgentSend` fails on clean `main` too — *"granted agent send → person.update: permission denied, want success"*. Different cause, reported separately.

## Verification

`internal/compose` green but for that one; `make check-go` green.

### DIFF
--- a/backend/internal/compose/attentionfeed_integration_test.go
+++ b/backend/internal/compose/attentionfeed_integration_test.go
@@ -23,6 +23,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"slices"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


dungbanguyen:

dungbanguyen:
current red — cluster A here, the planned lane's three composition tests in attentionfeed_integration_test.go.

`aa61b6f82`.

PR #5241, opened 2026-09-10
fix(consent): the send mints its unsubscribe link under its own admission

dungbanguyen:
The fourth of four `main`-reds, and the one I said on #5196 I would not fix. I changed my mind for one reason and narrowed the fix so it stays true: **this takes no product decision.**

```
--- FAIL: TestPreferenceCenterOptOutBlocksAgentSend
    granted agent send → person.update: permission denied, want success
```

Bisected to `14192f373` (#5196); `49ff4900f` passes.

## What is wrong

#5196 put `person:update` on `EnsureWithdrawalCredentialTx`. Its only in-tree caller is `WithdrawalTokenForEmail` — the **send path's** seam for the unsubscribe link RFC 8058 compels every marketing message to carry. So the grant is demanded of whoever is sending, not of an operator deliberately minting a credential. The test's agent holds `person: {Read: true}`, which is right for a sender.

@target-ce traced it independently and added the fact that settles it: the mint runs under the **sender's own principal** the whole way down — `activities.deliverability` → `UnsubscribeLinker` → `preferenceLinkAdapter` → `WithdrawalTokenForEmail` → the mint — with no system escalation anywhere.

It is also asked before the code knows whether a person is involved. `WithdrawalTokenForEmail`'s own doc: *"An address matching NOBODY still gets a credential. The message is going to that address regardless."*

## What this changes, and what it does not

The mint moves down one level. `EnsureWithdrawalCredentialTx` keeps `person:update` **exactly as #5196 wrote it**, and the send calls the level below.

Two callers asking two different questions, which is why they are two:

| caller | asks |
|---|---|
| `EnsureWithdrawalCredentialTx` | may you act on this person |
| the send path | may you send this message — already settled by its own admission |

**The check that actually encodes authority over a named person is untouched.** `auth.EnsureWritableLive(ctx, tx, entityPerson, in.PersonID.UUID)` is row-scoped, inside the transaction, and closes the erasure race a caller's earlier address lookup opens. It sits below the split; both callers reach it.

## Why I am not deciding the question

Whether a sending agent *should* hold `person:update` is real and it is #5196's author's. This restores `main` without answering it: if the answer is that the send should ask too, the gate is where they left it and the send path is one line from asking.

## One thing worth a look separately

Removing `EnsureWritableLive` entirely leaves the consent lane's erasure and credential cases **green**. I did not chase it — my change does not touch that check — but the race it closes appears to be unheld by a test, and it is the half of this function that matters most.

## Verification

`internal/modules/consent` green; `make check-go` green; `internal/compose` green but for the three cases #5237 fixes.

Four `main`-reds, four PRs: #5225 (lead SLA), #5237 (attention lane), #5230 (craft blockers, @target-ce's), and this. Each is red on the others' subjects until they land, so none goes green alone.

## Summary by CodeRabbit

- **Bug Fixes**
  - Sending withdrawal credentials now works when the sender is authorized to initiate the send, without requiring update permission over the recipient.
  - Withdrawal credentials can be sent to addresses that are not associated with an existing person.
  - Existing protections remain in place when a person is identified, and direct credential minting continues to require the appropriate authorization.

### DIFF
--- a/backend/internal/modules/consent/withdrawalcredential.go
+++ b/backend/internal/modules/consent/withdrawalcredential.go
@@ -139,9 +139,38 @@ func (s *Store) EnsureWithdrawalCredentialTx(
 	//
 	// person:update rather than person:read, because minting one changes what
 	// can be done to the record even though it writes no consent state.
+	//
+	// A DELIBERATE mint asks this. The send path does not come through here —
+	// see mintWithdrawalCredentialTx below for why, and for what still holds
+	// there.
 	if err := auth.Require(ctx, entityPerson, principal.ActionUpdate); err != nil {
 		return "", err
 	}
+	return s.mintWithdrawalCredentialTx(ctx, tx, in)
+}
+
+// mintWithdrawalCredentialTx is the mint itself, without the object grant.
+//
+// The unsubscribe link a marketing message carries is minted ON THE SEND PATH,
+// under the sender's own principal, and RFC 8058 compels the message to carry
+// it. Asking person:update there asks the sender for authority over the
+// recipient in order to send them mail they can opt out of — and it is asked
+// before the code knows whether a person is involved at all: this path serves
+// an address matching NOBODY, by design, because the message is going to that
+// address regardless.
+//
+// What still holds, and is the check that actually encodes authority over a
+// named person, is EnsureWritableLive below: row-scoped, inside this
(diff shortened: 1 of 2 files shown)


dungbanguyen:

dungbanguyen:
current red — cluster D here, the withdrawal mint's authority probe in withdrawalcredential.go.

`aa61b6f82`.

PR #5242, opened 2026-09-10
test(consent): the withdrawal mint's authority probe is held

dungbanguyen:
Follow-up to #5241, which noticed that `EnsureWritableLive` in
`EnsureWithdrawalCredentialTx` could be removed entirely with the consent lane
staying green. Confirmed independently — it stays green across `consent`,
`privacy` **and** the compose integration lane. The function's comment argues two
properties at length and nothing asserted either.

Checking **which** was unheld turned out to matter, because they are not alike.

## The live half is held twice

`LockSubjectLive` on the very next line asks `archived_at IS NULL` for its own
reason (holding the row so an erasure queues behind the mint). So removing the
probe leaves an erased subject still refused, and a case pinned to the probe
would have reported the belt gone while the braces held.

`TestTheMintRefusesASubjectWhoseRecordIsGone` therefore asserts the **property**
and is deliberately indifferent to which line delivers it — and says so, because
that is worth knowing rather than rediscovering.

## The write-authority half has nothing behind it

This is the real gap.

`person` is shareable, so a manual `read` grant widens who can SEE a contact
without widening who may act on them — and this path mints a **bearer credential
that can stop that person's mail for two years**. `LockSubjectLive` asks no
authority at all, and `ensureWriteAuthority` — the only thing that refuses —
short-circuits on `Unbounded(p)`. Every existing case in this package runs an
admin at `RowScopeAll`, so all of them walk straight past it.

`TestAReadShareOnAContactIsNotAuthorityToMintTheirOptOut` runs **bounded**
(`RowScopeOwn`), admitted to the contact by the share alone, and asserts both
directions:

- a `read` grant is refused with `ErrPermissionDenied`;
- the **same caller** with a `write` grant mints successfully.

Without the second half the refusal would also pass against a caller who simply
cannot do anything, which is the failure mode a one-sided authority test always
has.

The object grant in that fixture is deliberately generous (`person:
{Read, Update}`) — the case is about the ROW, and a narrow object grant would
refuse one gate earlier and prove nothing about this one.

## Mutation-checked, and the two mutations say different things

| mutation | live case | read-share case |
|---|---|---|
| probe removed | passes (the lock holds it) | **fails** |
| `EnsureWritableLive` → `EnsureVisibleLive` | passes | **fails** |

The second is the one worth having: it is exactly the distinction the comment
draws — *"EnsureWritableLive, not the VISIBLE twin PreferenceTokenForEmail uses"*
— and until now nothing would have noticed the downgrade.

## Scope

Tests only; no production change. It does not touch #5241's split, and applies
whichever way the `person:update` question on #5196 is settled — this is the row
gate under it, not the object gate.

## Checks

`make check-backend`, `craft static --strict` (0/0/0), and the full `consent`
integration lane.

### DIFF
--- a/backend/internal/modules/consent/withdrawalcredential_integration_test.go
+++ b/backend/internal/modules/consent/withdrawalcredential_integration_test.go
@@ -654,3 +654,121 @@ func TestTheMintDoorsAskForTheGrantTheirQuestionNeeds(t *testing.T) {
 		}
 	}
 }
+
+// The mint refuses a subject whose record is no longer live.
+//
+// The race: a statement in a read-committed transaction takes a fresh snapshot,
+// so an erasure committing between the send path's address lookup and this
+// insert would leave the mint writing a NEW capability — carrying the plaintext
+// address — for the subject whose credentials that erasure had just deleted. The
+// person row survives an anonymize-in-place, so the foreign key does not catch
+// it and the fresh token resolves happily.
+//
+// TWO LINES HOLD IT, and this case is deliberately indifferent to which:
+// EnsureWritableLive's live half and LockSubjectLive's `archived_at IS NULL`
+// both refuse, so removing either alone leaves the property standing. That is
+// the right shape for a property this serious, and it is worth knowing rather
+// than discovering — a case pinned to one of them would report the belt gone
+// while the braces held.
+func TestTheMintRefusesASubjectWhoseRecordIsGone(t *testing.T) {
+	e := setupChannelConsent(t)
+
+	// It works first, so the refusal below is about the erasure and not about
+	// the fixture never having been mintable.
+	before := mintWithdrawal(t, e, WithdrawalMintInput{
+		Address: "gone@example.test", Scope: WithdrawalScopeAllMarketing, PersonID: e.person,
+	})
+	if before == "" {
+		t.Fatal("the fixture minted no token before the erasure, so the refusal below would prove nothing")
+	}
+
+	if _, err := e.owner.Exec(context.Background(),
+		`UPDATE person SET archived_at = now() WHERE id = $1`, e.person); err != nil {
+		t.Fatalf("retiring the subject: %v", err)
+	}
+
+	err := e.store.db.Tx(e.ctx, func(tx pgx.Tx) error {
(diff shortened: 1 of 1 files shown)


dungbanguyen:
current red — the test half of cluster D, the withdrawal mint's authority probe.

`aa61b6f82`.

dungbanguyen:
Good news for this PR: the posture it pins is the one that ships, and your
read-share case should pass unchanged after a rebase.

I had dropped the mint's probe to `person:read` + `EnsureVisibleLive` in
https://github.com/margince/margince/pull/5224, because the only production door
is a marketing send and the strict probe refused every sender without
`person:update` — an agent holding `activity:create` + `person:read`, a rep on a
read share — while the message shipped anyway with no working List-Unsubscribe
URL. That directly contradicted `TestAReadShareOnAContactIsNotAuthorityToMintTheirOptOut`.

Rather than one of us winning by merging first, it went to the owner as a
decision. The answer was to SPLIT by call shape, which satisfies both:

- `EnsureWithdrawalCredentialTx` — the door a caller reaches with an id in hand
  — keeps `person:update` + `EnsureWritableLive`. **Your test drives exactly this
  door and asserts exactly this**, so it holds.
- `ensureWithdrawalCredentialForSendTx` — unexported, reached only by
  `WithdrawalTokenForEmail`, which names nobody and resolves the subject from
  the address the message is going to — takes `person:read` +
  `EnsureVisibleLive`.

The distinction is the question rather than the row: naming somebody's id is a
claim of authority over that record; a send resolving the subject from an
address it is already mailing asserts nothing new. Unexported keeps the weaker
probe out of reach of a caller who names a subject.

Two notes on your diff that survive intact and were worth having:
`TestTheMintRefusesASubjectWhoseRecordIsGone`'s point that TWO lines hold the
erasure race and the case is deliberately indifferent to which, and the positive
control in the read-share case — without it the refusal would also pass against
a principal that can do nothing at all.

One thing to re-check on rebase: your read-share test's object grant is
`{Read: true, Update: true}` with `RowScope: RowScopeOwn`, so it clears
`auth.Require(person, update)` and is refused by `EnsureWritableLive` on the
row. That is still the case on the strict door, so the refusal is still the row
answer you documented and not the object one.

PR #5245, opened 2026-09-10
fix: main goes green — four failures, four fixes, one branch

dungbanguyen:
`main` is red four ways and **none of the four fixes can land alone**: each is red on the other three's subjects, `ci` is the required check, and neither of us is reaching for `--admin`. Combined, they are green.

Supersedes #5225, #5237, #5241 (mine) and #5230 (@target-ce's, carried unmodified with authorship intact).

## The four

**1. `TestAnOwnerlessBreachReachesTheConfiguredIntakeSeat` + `TestAnIntakeSeatThatCannotWorkIsTreatedAsUnset`** — #5201 added `AND owner_id IS NOT NULL` to the breach scan and left #5015's tests failing.

Both PRs are right about their own case; ownership was the wrong discriminator. #5015 seeds an `inbound` nobody picked up — a person waiting. #5201 seeds a `siteread` — a name a pass read off a website. `lead_source.intent` already separates them, in its own words: siteread and crawl are low *"for crawl's reason: nobody asked us for anything."* Each half of the new condition is required by a different PR's test.

#5201's stated reason for the exclusion was correct and is closed rather than argued away: the take-on now clears `sla_breached_at` with the same condition that starts the clock.

**2. `TestAnOverdueTaskReachesThePlannedLane` + two boundary cases** — #5210 made the lane read what is coming as well as what is owed, across thirty files, and left three cases asserting the narrower lane. Each keeps its subject: a task already **done** is still carried by neither read, and the two boundary cases still pin which side of the installation's midnight a task falls on — now through `due_group`, which is where #5210 moved it.

**3. `TestPreferenceCenterOptOutBlocksAgentSend`** — #5196 put `person:update` on the mint the **send path** reaches for the unsubscribe link RFC 8058 compels. The mint moves down one level; `EnsureWithdrawalCredentialTx` keeps the grant exactly as written, and the send calls below it. **No product decision is taken** — whether a sending agent should hold `person:update` is still #5196's author's, and the gate is one line from being asked on that path too.

**4. Two craft BLOCKERs** (@target-ce, #5230) — #5211's swallowed savepoint rollback. Their finding is sharper than the generic one: that savepoint exists so a receipt collision can be **retried**, so a rollback that did not take is the single condition under which the retry cannot work, and every remaining attempt fails with the first one's message. The test's is `t.Cleanup` with the `ErrTxClosed` exception the suite already uses.

## Verification

On the combined branch:

```
make craft-static                              PASS — 0 blocker, 0 major, 0 minor
make check-go                                  clean
make test-it DIR=backend/internal/compose      ok (125s)
make test-it DIR=backend/.../modules/people    ok (37s)
```

## One thread worth pulling separately

`main-health` last ran on `7a246a72` and `main` is at `36a33cb6` — eleven merges without the health check seeing the tip, which is how four independent breakages accumulated unnoticed. Three of them are the same shape: **a change landed with a test still asserting the rule it replaced.** Not one author's mistake; a gap between "my PR is green" and "main is green", and the schedule is where it lives.

### DIFF
--- a/backend/internal/compose/attentionfeed_integration_test.go
+++ b/backend/internal/compose/attentionfeed_integration_test.go
@@ -23,6 +23,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"slices"
 	"testing"
 	"time"
 
@@ -153,9 +154,18 @@ func TestADetectedDuplicateReachesTheDecisionLane(t *testing.T) {
 	}
 }
 
-// An overdue task is today's agreed work. Both halves of the filter are pinned:
-// what is due is carried, and what is done or still ahead is not.
-func TestAnOverdueTaskReachesThePlannedLane(t *testing.T) {
+// The planned lane carries today's agreed work AND what is coming, each said to
+// be what it is.
+//
+// It used to carry only what was owed, and this case pinned that: a task still
+// ahead was a task the lane refused. That was the defect rather than the rule —
+// a comparison sheet due tomorrow was invisible today, so work accepted from a
+// transcript looked like it had gone nowhere. The lane makes two bounded reads
+// now and names each row's run in `due_group`.
+//
+// What has NOT changed is the other half this case was always about: a task
+// that is done is carried by neither read.
+func TestThePlannedLaneCarriesWhatIsOwedAndWhatIsComing(t *testing.T) {
 	e := integration.Setup(t)
 	now := time.Now().UTC()
 	logTask(t, e, "Ring the buyer back", now.Add(-2*time.Hour), false)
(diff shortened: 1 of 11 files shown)


dungbanguyen:
Rebased onto `aa61b6f8`, and a **fifth** breakage arrived while this was being assembled: #5234 grew `BookPublicMeeting` to 81 code lines, and `make craft-static` is whole-tree, so it fails every open pull request.

Folded in — the validation prologue comes out as `bookingRequestIsWellFormed`. It is one concept (refusing a booking that cannot be written down, before anything is written) and every check in it is there for the reason the extracted comment now states once instead of twice.

That is five failures from five separate merges, none of which was red on its own PR. `main-health` has still not run since `7a246a72`.

dungbanguyen:
**Handing this over rather than carrying it further.**

This branch fixes five of `main`'s failures and is verified green against the tree it was built on. Two more have appeared since and I am not taking them:

```
--- FAIL: TestQuietCandidatesCarryOnlyContactsInsideRowScope    decaylane_integration_test.go:103
--- FAIL: TestADismissedContactDoesNotSpendACandidateSlot       decaylane_integration_test.go:198
        candidates = 0, want only the readable lapsed contact
```

Both reproduce on clean `aa61b6f8`. The fixture seeds an interaction at `QuietDays + 35` days ago and `QuietEdgesForUser` returns nothing for it, so the edge the query reads is not being produced for an activity that old — worth checking whether the projection has a window the fixture now sits outside.

## Why I am stopping

The breakages are arriving faster than they can be cleared, and they share a cause rather than being seven accidents:

| PR | merged | `ci` at merge |
|---|---|---|
| #5196 | 02:21 | **failure** |
| #5201 | 02:50 | **failure** |
| #5210 | 03:56 | **failure** |
| #5234 | 05:38 | **failure** |

Each is a change that is itself reasonable — I said so on all four — landed without its checks passing, and each left behind a test asserting the rule it replaced. `main-health` has been failing since `04:47` on `5d732ed9`.

That is not something more fixing addresses. A contributor who respects the required check cannot land anything while it is red, so the queue stalls for everyone who does; five of @target-ce's and two of mine are waiting on this branch alone.

This branch stands on its own and can be merged or taken over as it is. I am back to my own range.

dungbanguyen:
Closing: every one of the four is fixed on `main`, by #5224 and #5201's follow-up rather than by this branch.

- **The breach scan's discriminator** — `main`'s `leadOwesAReplySQL` now narrows on the SOURCE rather than on ownership, in the same words this branch argued for: *"a name the product read off a web page owes no first response, because nobody asked us for anything."*
- **The savepoint rollback** — `errSavepointPoisoned` is on `main`; the rollback is no longer swallowed.
- **The planned lane and the preference-center opt-out** — both fixed inside #5224's fifteen.

`main`'s required checks are green, which is the only thing this branch existed to restore. Its own implementations are alternatives to what landed, not additions to it, so rebasing would put a second answer beside a working one.

The thread that outlived the fixes is the one in the PR body: `main-health` ran eleven merges behind the tip while four independent breakages accumulated. That gap is not closed by any of this, and is worth an issue of its own rather than a stale branch.

Issue #5189, opened 2026-09-09
The e2e-llm workspace has no active colleagues, so every case needing one is unreachable


tinnhlt-gra:
## What

`list_colleagues` returned an **empty roster on every call in every run** of the last full haiku sweep — 5 calls across the lane, none of them narrowed to a name that exists, and none of them answering anybody. Verified by walking `e2e/llm/records/*.jsonl`:

```
empty: 5   non-empty: 0
```

## Why

`e2e/llm/seed-llm-fixtures.sh` creates its colleagues with `POST /users`:

```sh
colleague="$(id_of "$(api POST /users '{
  "email":"sofia.meier@demo.test","display_name":"Sofia Meier","role":"rep"}')")"
...
lena="$(id_of "$(api POST /users '{
  "email":"lena.fischer@demo.test","display_name":"Lena Fischer","role":"rep"}')")"
```

An invited seat has no `password_hash`, so it is created `status = 'invited'`. `identity.Colleagues` (`backend/internal/modules/identity/colleagues.go`) filters on `LiveMemberSQL("")` — `status = 'active' AND archived_at IS NULL` — so neither seat is ever a colleague. A seat leaves `invited` by redeeming a set-password link (`identity.RedeemPasswordReset`, `reset.go:317`, which sets `status = 'active'`), and the seed never does that.

The tool is right to find nobody. The fixture is what is wrong.

## What it costs

At least two scenarios cannot reach their criteria, on any model:

- **case33** criterion 3 — "the account is handed to a named colleague". The scenario's own seed comment says *"Case 33 requires list_colleagues only because Lena's user id exists nowhere else a tool can reach"*, and that route is closed. A correct run reads the empty list and reports, accurately, that she has no seat.
- **case4** criterion 6 — "check with the owner before turning up", asserted on `/Sofia Meier|Lena Fischer|Phạm Hương Giang/`. Every record's owner is a seat no tool can name.

Both currently read as model failures in `docs/reference/mcp-tool-coverage.md`.

## Fix

Activate the seeded seats. The seed already holds the admin credential and the migrate CLI is available to the lane, so the cheapest honest route is whatever `scripts/e2e-llm.sh` can reach — either redeeming a set-password link per seat, or a documented operator step. Whichever is chosen, the seed should then VERIFY, the way the post-seed close-date check does: a run whose roster is empty should fail the seed loudly rather than produce a sweep that blames the model.

## Related, and possibly its own defect

[...]

PR #5247, opened 2026-09-10
The MCP surface says what it is not for, and stops promising a human who is not there

### LINKED ISSUE #5184: priced_count on a shared snapshot counts conversion, not pricing
`forecast_contribution` is re-summed for a share recipient in `backend/internal/compose/analyticssharesnapshot.go`, and `priced_count` is scanned from:

```sql
count(*) FILTER (WHERE c.base_minor IS NOT NULL AND <visible>)
```

`base_minor` is the amount **converted** into the snapshot's base currency. `amount_minor` is the amount the deal **carries**. The two differ for exactly the population `fx_missing_count` exists to name: a deal priced in a currency no rate could reach has an `amount_minor` and a NULL `base_minor`.

So on the share path:

- `priced_count` contradicts its own contract description in `backend/api/crm.yaml` ("How many carried an amount").
- It disagrees with the other writer of `forecasting.Readings` — `Compute` increments `PricedCount` when `AmountMinor != nil`.
- `eligible_count - priced_count` reads as unpriced deals while actually being unpriced **plus** unconvertible, and `fx_missing_count` then reports the same deals a second time.

Repro shape: a snapshot with 10 visible contributions, none unpriced, 3 priced-but-unconvertible. `ReadSharedSnapshot` answers eligible=10, priced=7, fx_missing=3 — a recipient is told the total misses 6 of 10 deals when it misses 3.

Fix is `count(*) FILTER (WHERE c.amount_minor IS NOT NULL AND <visible>)`. Held out of #5183 because proving it needs a snapshot fixture carrying an fx-missing contribution in the integration lane, which is most of the change.

Found by adversarial review of #5183, which adds a model-facing sentence written from these counts. That sentence is not served on the share path, so nothing states the wrong figure in prose today — the counts themselves are what is wrong.

tinnhlt-gra:
Ten tools told a model a person would approve a call that in fact runs immediately. A tool that writes board documents described its plumbing instead of its job. Five forecast tools could not name a neighbour because they had opted out of the copy struct that has the field for it. And a third of the e2e-llm suite was scoring phrasing rather than behaviour.

Closes #5184 and #5189.

## Measured, three models, one tree

| model | cases | runs |
|---|---|---|
| **sonnet** (the target) | 13/21 → **15/21** | 40/63 → **44/63** |
| haiku | 10/21 → 10/21 | 29/63 → **32/63** |
| opus | 17/21 → **15/21** | 50/63 → 47/63 |

**Net at case level: zero.** Sonnet gains two, opus loses two. The gains are the cases this work targeted; the losses are one known trade and one variance, both attributed below. I would rather hand you that than a headline.

### What earned the gains

| case | what changed | result |
|---|---|---|
| **case33** two cards for one company | the fixture had no active colleague seats at all, plus the staged answer now reports what it DID, plus a pattern that stopped demanding one spelling | haiku **0/3 → 3/3**, sonnet **1/3 → 3/3** |
| **case32** two words for one thing | the staged 🟡 answer carries the summary the human's card was given | sonnet **0/3 → 3/3**, opus 2/3 → 3/3 |
| **case20** board pack | `compose_analytics_report`'s Purpose leads with the JOB, not the mechanism | sonnet **0/3 → 3/3**, opus 2/3 → 3/3 |
| case22, case23, case31 | — | up on haiku |

### What the losses cost, honestly

- **opus case7 3/3 → 0/3 — this branch caused it.** Making `compose_analytics_report` findable for "write me a board section" also made it findable for "give me a number", and opus stopped calling `run_report`. Narrowing the Purpose and giving `run_report` an explicit boundary recovered part of it and not all. It is a real trade: sonnet gains case20, opus loses case7. Undecided rather than hidden — see the note at the end.
- **opus case1 2/3 → 1/3 — not this branch.** A judge criterion about turning a raised topic into a committed action item, on `log_activity`/`create_task`. The diff contains zero changes to either tool's copy. Variance.
- **opus case3 3/3 → 1/3 — not the regex.** This one is worth reading, because the corrective re-run after the table-cell fix *did not recover it* (sonnet 3/3, haiku 1/3, opus 0/3) and named the real cause: every remaining failure is the **judge**, whose words are *"the answer reports the import as already completed with 'Done. Imported' in past tense"*. The criterion asks for a preview of what the import **would** do before anything is written; the model writes the rows and then reports them. A live finding about the dry-run surface that this PR does not fix.

## 1. Ten tools promised a human who is not there

```
send_email description:  "a person approves the send before it leaves"
send_email governance:   "runs immediately"
```

ADR-0055 stopped these verbs staging by default. The governance line is derived from the enforced tier, so it followed. The prose above it is hand-written, and did not. A model reading `send_email` was told somebody would catch a mistaken send, and by default nobody would — the surface understating its own authority to the one party deciding whether to use it.

Fixed on all ten, using the phrasing `demote_lead` already had, which is true at either tier. **And gated:** `TestNoToolPromisesAnApprovalItsTierDoesNotRequire` fails when prose and tier disagree. Written as a contradiction check rather than banned words, because an installation may legitimately turn approval on — a second test holds that the conditional form stays sayable, and a floor fails the census if it stops reaching the surface.

## 2. The forecast family had opted out of the copy discipline

Five tools used hand-rolled `Description` strings instead of `toolCopy`. A string has nowhere to put an `Instead`, so they named nobody and nobody named them.

```
BEFORE                            AFTER
forecast_readings      → nobody   → run_report, query_workspace,
forecast_movement      → nobody     compose_analytics_report,
forecast_input_checks  → nobody     run_analytics_query, forecast_input_checks
list_input_checks      → nobody   (and the other four likewise)
data_coverage          → nobody
```

Three tools answer "can I trust these numbers?" — the verdict, the findings, and whether the sources were readable at all — and none said which question it held. A clean set of findings over sources nobody opened reads as good news and is not. `forecast_movement` now also admits it **cannot be called**: it needs snapshot ids nothing on this surface hands out, and says that reading `forecast_readings` twice and subtracting is not the same answer.

## 3. An empty colleague list was two different answers

`{"colleagues": []}` meant both "this workspace employs nobody" and "nobody here is spelled the way you asked". A caller cannot tell them apart, and it picked wrong: *"Lena Fischer isn't in the system yet."*

A miss now hands back the set it was matched against, through the envelope's warning channel rather than standing tool copy — a `Limits` sentence is paid on every listing of all 75 tools, and this one is only true on the miss. A successful empty fallback (`[]`) and an unreadable one (absent) are different bytes. The cap rides its own flag, because 200 alphabetical names with the one asked for absent reads as *proof* of absence.

**And the tool was right all along.** `POST /users` writes `status='invited'`; `identity.Colleagues` lists only `status='active'`. Verified across a full sweep: **5 calls, 0 non-empty.** Two criteria on two cases were unreachable by any model and were scoring as model failures. The seed now mints a set-password link and redeems it, then verifies the roster READ rather than the rows it wrote.

## 4. A staged call says what it staged, and what it already did

The staging composed a summary for the human's inbox card and threw it away, so an agent could only report that something was pending. Now it carries it, escaped and bounded on the way out — the summary holds a caller-chosen field name and a workspace record label, and this text lands in a transcript whose later prompts the same run reads.

It also says the approval blocks **that call only** and to report what already happened. Two measured runs had merged a duplicate and archived a dead company and mentioned neither, because they wrote their answer about the one thing that stopped.

## 5. A share count meant conversion, not pricing

`count(*) FILTER (WHERE c.base_minor IS NOT NULL)` scanned into `priced_count`, whose contract says *"how many carried an amount"*. The one population they differ on is the deal priced in a currency no rate reached — which `fx_missing_count` already names, so the single gap was reported twice.

**Held, not just fixed.** The two writers cannot share a helper, so they are asserted equal against the frozen row. Broken from both sides: the SQL reverted gives `priced_count = 1, want 2`; `Compute` broken instead gives `recompute says 3/3/1, frozen row says 3/2/1`. A constants-only test sails through the second.

## 6. A third of the suite was scoring the wrong thing

- **case2 asserted a duplicate the fixture never seeded.** Two runs passed on the word "already" borrowed from case8's approval queue; the one honest answer scored FAIL. `manualDedupePerson` files a review row on a name collision — the code's own example is literally "Lucy Vo" — so the seed now plants one.
- **case3** `"3|three"` + `"row"` was satisfied by "rows 1-3" with no report at all.
- **case21** a bare `[0-9]` was satisfied by a date.
- **case4** naming the owner satisfied the criterion about *advising* the rep to check with them.
- **case5** `"Vietnam Partner"` was satisfied by echoing the prompt; `"list|no record"` was satisfied by an answer *denying* the finding.
- **case9** required the count and the noun one word apart, so "the three Aachener Metallwerke calls" — a two-word company — failed.
- **case33** demanded one spelling of the survivor.

Every replacement was verified in **both** directions against the recorded transcripts, the curated fixtures in `e2e/llm/testdata/`, and a constructed false green. Two rounds were needed: my first attempt would have redded twelve correct answers.

**New: `e2e/llm/why.py`.** A failing `must_call` says which tool was missed and never why — and the answer is one of three things needing three different fixes: never served, served-and-not-chosen, or chosen-and-refused. It prints all three per run from records the lane already keeps, at no cost. It answered case20 in one command.

## Two things I got wrong along the way

- **My first case33 fix shipped a false green.** Accepting bare `GmbH` also matched *Sauerland Kunststoff GmbH*, the other company in the same prompt, so an answer naming no survivor would have passed. Now anchored on tokens unique to the twins. A false green is worse than the false red it replaces: the red at least makes somebody look.
- **`deterministic-gates` caught a case3 pattern that redded the correct fixture.** `e2e/llm/testdata/` is a better oracle than the live records and I had not been using it. The *defective* fixtures matter too: they must pass the regex and fail only the judge, or a too-tight pattern fails them for the wrong reason.

## Why no coverage table in this diff

The published `docs/reference/mcp-tool-coverage.*` is **deliberately untouched**. The three-model sweep's verdict records were lost to a `git reset --hard` of mine before a corrective re-run, and reconstructing them from the logs would put figures in a committed record that the tool did not write. Committing only the cases I still have would publish exactly the blended table I criticised in the haiku row this morning.

So the numbers above come from the sweep logs, and the table will be regenerated by the next clean sweep. `case3` in particular was first measured against a pattern I fixed mid-sweep and was re-measured afterwards on a synced tree — with the result recorded in the loss list above.

## Open, and not resolved by this PR

- **The case7/case20 trade.** Worth a decision rather than more copy iteration; three rounds bought one run.
- **case40's bounce note reaches no model in any run**, so the passing runs disqualify a lead on record age — the guess the case exists to rule out.
- `Instead:` renders 3rd of 4, landing 50–60% into every description. A model skimming 75 tools reads openings, and the boundary is never in one.

### DIFF
--- a/backend/internal/compose/analyticsshare_integration_test.go
+++ b/backend/internal/compose/analyticsshare_integration_test.go
@@ -29,9 +29,7 @@ import (
 	"github.com/margince/margince/backend/internal/shared/kernel/principal"
 )
 
-// freezeTwoTeams seeds one priced deal per team and freezes the pair, through
-// the REAL writer and the real deal read, so what the recompute later re-sums
-// is what production would have written.
+// freezeTwoTeams seeds one priced deal per team and freezes them.
 func (e *forecastEnv) freezeTwoTeams(t *testing.T, mine, theirs int64) ids.UUID {
 	t.Helper()
 	// Inside the period on purpose. seedOpenDeal dates a deal 30 days out,
@@ -40,7 +38,14 @@ func (e *forecastEnv) freezeTwoTeams(t *testing.T, mine, theirs int64) ids.UUID 
 	// pass or fail by the calendar rather than by the recompute.
 	e.seedDealClosingWithin(t, "Mine", &e.Rep1, mine)
 	e.seedDealClosingWithin(t, "Theirs", &e.Rep3, theirs)
+	return e.freezeWorkspace(t)
+}
 
+// freezeWorkspace freezes whatever deals are already seeded, through the REAL
+// writer and the real deal read, so what the recompute later re-sums is what
+// production would have written.
+func (e *forecastEnv) freezeWorkspace(t *testing.T) ids.UUID {
+	t.Helper()
 	store := forecasting.NewStore(InstallationDB(e.Pool))
 	admin := snapshotWriterCtx(e.WS)
 	var id ids.UUID
@@ -65,7 +70,7 @@ func (e *forecastEnv) freezeTwoTeams(t *testing.T, mine, theirs int64) ids.UUID 
 		})
 		return err
 	}); err != nil {
-		t.Fatalf("freezing the pair: %v", err)
+		t.Fatalf("freezing the seeded deals: %v", err)
 	}
 	return id
 }
@@ -74,10 +79,20 @@ func (e *forecastEnv) freezeTwoTeams(t *testing.T, mine, theirs int64) ids.UUID 
 // so it falls in the current quarter whatever day the suite runs on.
 func (e *forecastEnv) seedDealClosingWithin(t *testing.T, name string, owner *ids.UUID, amountMinor int64) {
 	t.Helper()
+	e.seedDealPricedIn(t, name, owner, amountMinor, "EUR")
+}
+
+// seedDealPricedIn is the same plant in a named currency. This environment
+// populates no rate sheet at all, so anything but the installation's own base
+// currency produces a deal that CARRIES an amount and has no base — priced and
+// unconvertible, the one population `priced_count` and `fx_missing_count`
+// disagree about.
+func (e *forecastEnv) seedDealPricedIn(t *testing.T, name string, owner *ids.UUID, amountMinor int64, currency string) {
+	t.Helper()
 	e.seedID(t, `INSERT INTO deal (id, name, pipeline_id, stage_id, owner_id, amount_minor, currency,
 			expected_close_date, source, captured_by)
-		VALUES ($1, $2, $3, $4, $5, $6, 'EUR', (now() + interval '2 days')::date, 'manual', 'human:x')`,
-		name, e.pipeline, e.stages[20], owner, amountMinor)
+		VALUES ($1, $2, $3, $4, $5, $6, $7, (now() + interval '2 days')::date, 'manual', 'human:x')`,
+		name, e.pipeline, e.stages[20], owner, amountMinor, currency)
 }
 
 // writerUser is the seat snapshotWriterCtx acts as.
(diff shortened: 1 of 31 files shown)


tinnhlt-gra:
Three findings, two taken and one respectfully disputed with evidence.

**Taken — the conditional wording was stale on two verbs.** My edit left `promote_lead` saying the id "comes back only from the retry that carries their approval", which is false on the default path where there is no retry; and `advance_project_phase` said the phase "moves before it runs", which is garbled. Both now state the default outcome first and the confirm-first outcome as the conditional, with `advance_project_phase` explicitly saying the phase has NOT moved and must not be reported as advanced until the approved retry answers.

**Taken — `activate_seat` now skips an already-active seat.** Right that redemption revokes every live session and replaces the password, so calling it unconditionally turned a re-run of an idempotent seed into a logout for whoever was using it. It reads `status` first and returns early.

**Disputed — the YAML escapes are safe here, and I measured it.** The scenarios are not read by a YAML library. `e2e/llm/check.py` uses `parse_scenario`, a deliberately small hand-written subset reader, and its `_scalar` only strips the surrounding quotes — there is no escape processing, so `\b` reaches Python's `re` as a word boundary. That is also why every pre-existing pattern in the suite is written this way.

Verified through the real loader rather than by reading it:

```
case21-what-are-we-closing       2 patterns, 2 with escapes, all compile: True, backspace chars: 0
case3-spreadsheet                2 patterns, 2 with escapes, all compile: True, backspace chars: 0
case4-use-the-moment             2 patterns, 1 with escapes, all compile: True, backspace chars: 0
case5-before-the-meeting         3 patterns, 2 with escapes, all compile: True, backspace chars: 0
case9-filed-in-the-wrong-place   3 patterns, 1 with escapes, all compile: True, backspace chars: 0
```

Zero backspace characters, every pattern compiles. Switching to single-quoted scalars would be a no-op here and would make these five files disagree with the other sixteen, so I have left them as they are. Worth knowing that a strict YAML loader *would* reject them — `yaml.safe_load` on these files raises `found unknown escape character 'w'` — so if the lane ever adopts one, the whole suite moves together.

tinnhlt-gra:
Two more findings, both taken — and one of them I could only satisfy by disagreeing with the suggested fix, so here is the evidence.

**`compose`/`forecast_readings` says "four money readings" while the payload carries five.** Correct, and the omission risk is real — `weighted_minor` is the one that goes unreported. But "five" would put the tool copy in direct conflict with the contract, which opens `ForecastReadings` with *"Four money readings over one period"* (`backend/api/crm.yaml`). So rather than pick a side of a count I have no mandate to change, the Purpose now **names every figure** and asserts no number:

> "Answer what a period is expected to close — `won`, `evidence`, `best_case` and `open`, plus `weighted` — under the installation's own fiscal calendar and base currency."

A caller can no longer omit `weighted` because it was never told how many to expect. If the contract's "four" is itself wrong, that is a separate change to `crm.yaml` and its REST twin.

**case3's count alternatives accepted a bare number.** Right — `"I will create three."` passed. The suggested fix, requiring an entity noun in every alternative, I tested first, and it reds the correct fixture:

```
  miss  fixture correct 'Would create (3)' + a name list
  miss  live run '3 created' + a name list
  HIT   defective A 'created three companies'
  HIT   defective B 'three usable rows'
```

Both correct answers enumerate company *names* — "Nordwind Logistik GmbH" — not the word "company", so a noun requirement scores how the list is worded. That is the same defect this PR is fixing elsewhere.

The count now requires a noun **or an enumeration** (a colon, a bullet, a newline) after the number, which is what actually distinguishes reporting from hand-waving:

```
  OK  HIT   fixture correct
  OK  HIT   live run
  OK  HIT   defective A
  OK  HIT   defective B
  OK  miss  "I will create three."        ← your case
  OK  miss  "a dry run over rows 1-3…"    ← the original false green
```

Note the two DEFECTIVE fixtures must keep passing the regex and fail only the judge — a pattern tight enough to red them fails `test-e2e-llm-check` for the wrong reason. `make check-backend` exits 0, and all 145 checker cases pass.

PR #5282, opened 2026-09-10
fix(frontend): a WORK primary and a PERSONAL primary stand independently

### LINKED ISSUE #5272: Editing which of two WORK addresses is primary silently un-primaries an unrelated PERSONAL/other-type address
## Observed

Measured live against `margince-qc`'s own `qc-own` dev stack, core `36a33cb663bf52a4bc2f151c89f1551f4b501668`, 2026-09-10.

A person with **two different email types, each independently primary** —
one `work` address and one `personal` address, both `is_primary: true`,
which `uq_person_email_primary` (scoped to `person_id, email_type`) allows
and the schema intends — created via:

```
POST /v1/people
{"full_name":"...","emails":[
  {"email":"work-a@example.com","email_type":"work","is_primary":true,"position":0},
  {"email":"work-b@example.com","email_type":"work","is_primary":false,"position":1},
  {"email":"personal-c@example.com","email_type":"personal","is_primary":true,"position":2}
]}
```

Opening the contact page's own Edit form (`PersonEditMergeArchive`,
`frontend/src/screens/personeditmergearchive.tsx`) and clicking the "Primary"
radio on the **work-b** row — correcting which WORK address is primary, never
touching the personal row at all — then pressing Save sends:

```
PATCH /v1/people/{id}
{"emails":[
  {"email":"work-a@example.com","email_type":"work","is_primary":false,...},
  {"email":"work-b@example.com","email_type":"work","is_primary":true,...},
  {"email":"personal-c@example.com","email_type":"personal","is_primary":false,...}
]}
```

`personal-c`'s own primary flag flips from `true` to `false` — silently, in a
save the reader believed only touched the work address. The server accepts
this as a valid update (no conflict, because at most one row per type is now
primary) and writes it faithfully; the data loss is entirely client-side.

### Why

`RepeatableRowsField.markPrimary` (`frontend/src/screens/create.tsx`) clears
the primary flag on **every other row in the field** when one is clicked:

```ts
function markPrimary(index: number) {
  if (!primaryKey) return;
  setRows(rows.map((row, rowIndex) => ({
    ...row,
    [primaryKey]: rowIndex === index ? "true" : "",
  })));
}
```

It has no notion of `email_type`/`phone_type` — the clearing is field-wide,
not per-type, even though the uniqueness constraint it exists to satisfy
(`uq_person_email_primary`, `uq_person_phone_primary`) is per-type. The same
native `<input type="radio">` `name` (`${formId}-${field.key}-primary`,
one per FIELD not per type) reinforces this at the DOM level: a native radio
group only ever allows one CHECKED member regardless of type.

**A second, milder symptom of the same root cause, confirmed the same
session:** merely OPENING the Edit form on a person who already holds two
different-type primaries renders only ONE of them checked (whichever type's
row is later in the array) — the browser's native same-`name` radio-group
behavior silently unchecks the earlier one on initial mount, even though
React's own controlled state still correctly holds `is_primary: true` for
both. Confirmed this half does NOT corrupt data on its own — a Save with
nothing touched still sends both primaries correctly — but a person reading
the form sees the wrong row marked primary before touching anything.

## Cost

A person correcting a typo'd work email, or moving which work address is
primary, silently un-primaries an unrelated personal (or, for phones, home)
address on the same record — with no warning, no confirmation, and no
visible difference in the form (the row they didn't touch just quietly loses
its own dot). The next reader of that record sees a personal address with no
row marked primary at all, and there is no way from the screen to tell this
happened versus the person never having set one. Reversible by hand (open
Edit, re-check the row, save again) once noticed, but nothing on screen
signals it needs noticing.

## Workaround

After editing a same-type primary, re-open the Edit form and re-check any
other-type row that used to be primary before saving again — or edit each
type's primary in its own separate save, checking the OTHER type's rows are
still correct each time.

tinnhlt-gra:
## Summary

Closes margince#5272 — editing one WORK-type primary contact (email or phone) on the person edit form silently un-primaried an unrelated PERSONAL-type primary, both at the React state level and at the DOM level (the native radio group's own `name` was field-wide, not scoped by kind).

Root cause: `RepeatableRowsField`'s `markPrimary` cleared the primary flag across every row in the field, and the radio inputs all shared one native `name`, so the browser itself enforced the same field-wide exclusivity. This is a regression surfaced by margince#4823's own fix (PR #5181), which introduced the shared `PersonEditMergeArchive` component reusing this same `RepeatableRowsField`/`create.tsx` code for both `PersonScreen` and `PersonPageV2`.

Fixed by scoping both the state update and the native radio group's `name` to the row's own kind (`email_type`/`phone_type`), added as a new `typeKey` field alongside the existing `primaryKey`. New logic lives in `frontend/src/screens/createrows.ts` rather than growing `create.tsx`, which is already at its frozen file-length waiver (982 lines).

**Two review passes (a fresh `code-review` pass and an independent Codex pass) both caught the same follow-on gap** in the first version of this fix: an unset row's kind ("") and an explicit "Work" row's kind were treated as different kinds client-side, but the request mapper's own `asEmailType`/`asPhoneType` normalize an unset type to `"work"` at submit time — so an unset row and an explicit Work row could both end up `is_primary: true` and both submit as `email_type: "work"`, the exact same-kind collision this fix exists to prevent. Fixed with a `typeDefault` (declared once per field, `"work"` for both emails and phones, matching the request mapper's own fallback exactly) that folds an unset row into its default kind everywhere a kind is computed. Separately, changing an already-primary row's own type via the ordinary Type select never reconciled its primary flag against the row's new kind — fixed by clearing a row's own primary the moment its type changes, so it never silently contends for a kind it didn't choose.

## What's covered

- WORK primary and PERSONAL primary stay independent (state + DOM radio grouping)
- An unset-type row groups with the field's own default kind (`work`), matching the server's own fallback
- Changing a primary row's type clears its own primary rather than colliding with the new kind's primary
- 100% line/function coverage on the new logic (`createrows.ts`)

## Scope boundary

Does not touch the edit-flow's own test suite (`edit.test.tsx`) — the new `RepeatableRowsField`/`createrows.ts` logic is shared code exercised identically from the create flow, which is what the new tests drive; a full edit-flow duplicate would prove the same logic twice for no new coverage.

## UAT

Live-walked on an isolated `DEV_SLUG` dev stack: created a contact with one unset-type primary email and one explicit Personal-type primary email, confirmed both radios stay checked through the whole flow, submitted, and read the stored record back via the API — both emails persisted as `is_primary: true` with their own correct `email_type` (`work`, `personal`), no 409, no collision. Recording attached below.

## Review

Two independent passes before push (`code-review` skill, `codex:rescue`) both found the same real gap in the first commit (unset-vs-explicit-default-kind collision, plus the type-change-reconciliation gap); both fixed and reflected in the final commit, verified by two new regression tests plus the live UAT above.

https://claude.ai/code/session_01GkdtztRqyypJufL6VTE2Vs

---
## Summary by cubic
Fixes margince#5272: a WORK primary email or phone and a PERSONAL primary no longer clear each other on the person edit form. Marking one type primary used to un-primary the other type's primary in both React state and the browser's native radio group; now each kind's primary is scoped independently and both can submit as primary.

- Scopes primary exclusivity to same-kind rows via a new `typeKey`, matching the server's own same-kind rule (only a same-kind primary swap is refused).
- Groups an unset row type with the field's default (`work`), matching the request mapper's fallback, so an unset row and an explicit Work row can't collide as two primary work emails on submit.
- Clears a row's own primary when its type changes, so it never silently contends for the new kind's primary slot.

<sup>Written for commit 8f1824bd365443f73dbe11e8128a65dede25cf77. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/5282?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/frontend/src/screens/create.test.tsx
+++ b/frontend/src/screens/create.test.tsx
@@ -152,11 +152,7 @@ describe("contact create flow", () => {
 
   // A WORK primary and a PERSONAL primary are independent, matching what the
   // server itself enforces (personformfields.ts's personEditFields: only a
-  // SAME-type primary swap is refused). Regression for margince#5272, where
-  // markPrimary and the radio group's own `name` were both scoped to the
-  // whole field rather than to the row's own email_type/phone_type, so
-  // marking a PERSONAL email primary silently un-primaried an untouched WORK
-  // email.
+  // SAME-type primary swap is refused).
   it("keeps a WORK primary and a PERSONAL primary independent", async () => {
     const captured: Captured[] = [];
     stubApi(
@@ -219,6 +215,65 @@ describe("contact create flow", () => {
         },
       ],
     });
+  });
+
+  // The request mapper's own fallback (asEmailType) resolves an unset type
+  // to "work" — the form has to group rows the same way, or an unset row
+  // and an explicit Work row read as two kinds here and collide once both
+  // reach the server as two primary work emails.
+  it("groups an unset type with the field's own default kind", async () => {
+    const user = userEvent.setup();
+    render(<ContactsScreen />);
+    await user.click(screen.getByText(en["create.contact"]));
+    await user.type(screen.getByLabelText("Full name *"), "Peter Neu");
+
+    await user.click(screen.getByText("Add email"));
+    await user.click(screen.getByText("Add email"));
+    const emailInputs = screen.getAllByLabelText("Email *");
+    await user.type(emailInputs[0], "peter.unset@neu.example");
+    await user.type(emailInputs[1], "peter.work@neu.example");
+    const types = screen.getAllByRole("combobox", { name: "Type" });
+    await pickOption(user, types[1], "Work");
+    // types[0] stays unset on purpose.
+
+    const primaries = screen.getAllByRole("radio", {
+      name: "Primary",
+    }) as HTMLInputElement[];
+    await user.click(primaries[0]);
+    await user.click(primaries[1]);
+
+    expect(primaries[0].checked).toBe(false);
+    expect(primaries[1].checked).toBe(true);
+  });
+
+  it("clears a row's own primary when its kind changes, rather than colliding with the new kind's primary", async () => {
+    const user = userEvent.setup();
+    render(<ContactsScreen />);
+    await user.click(screen.getByText(en["create.contact"]));
+    await user.type(screen.getByLabelText("Full name *"), "Peter Neu");
+
+    await user.click(screen.getByText("Add email"));
+    await user.click(screen.getByText("Add email"));
+    const emailInputs = screen.getAllByLabelText("Email *");
+    await user.type(emailInputs[0], "peter.personal@neu.example");
+    await user.type(emailInputs[1], "peter.work@neu.example");
+    const types = screen.getAllByRole("combobox", { name: "Type" });
+    await pickOption(user, types[0], "Personal");
+    await pickOption(user, types[1], "Work");
+
+    const primaries = screen.getAllByRole("radio", {
+      name: "Primary",
+    }) as HTMLInputElement[];
+    await user.click(primaries[0]);
+    await user.click(primaries[1]);
+    expect(primaries[0].checked).toBe(true);
+    expect(primaries[1].checked).toBe(true);
+
+    // Retype the PERSONAL row as Work — the second row's own kind.
+    await pickOption(user, types[0], "Work");
+
+    expect(primaries[0].checked).toBe(false);
+    expect(primaries[1].checked).toBe(true);
   });
 
   it("renders the server's 422 detail verbatim and stays open", async () => {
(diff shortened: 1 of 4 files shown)


tinnhlt-gra:
UAT proof: creating a contact with one unset-type primary email and one explicit Personal-type primary email on an isolated dev stack. Both radios stay checked through the whole flow (the exact bug margince#5272 reports); after submit, the stored record carries both as `is_primary: true` with their own correct `email_type` (work, personal) — read back live via the API, no 409, no collision.

Both a WORK/unset-type primary and a PERSONAL-type primary stay checked and save correctly

PR #5286, opened 2026-09-10
claim: main-red — TestSweepTargetsCarryNoDeleteBlockingTrigger

tinnhlt-gra:
## Summary

Fixes the main-red `TestSweepTargetsCarryNoDeleteBlockingTrigger` (backend/internal/compose/datareset_integration_test.go), introduced by PR #5284's `communication_instruction` table — its trigger refuses every DELETE unconditionally ("a communication instruction is the record of a decision and is never deleted"), the same shape `audit_log`/`system_log` already have, and the gate exists precisely to force a new table like this to be classified rather than discovered at runtime.

Classified into `preservedResetTables` (and its `scripts/seed-reset.sql` mirror — `backend/gates/seedresetparity_test.go` holds the two equal) rather than `deleteGuardedSweepTargets`: the guard has no row-conditional carve-out for a legitimate cascade to pass through, unlike `activity_retention_evidence` beside it.

Filed margince#5287 for the obligation this leaves standing: `communication_review` references `communication_instruction` with `ON DELETE CASCADE`, and that cascade will hit the same unconditional refusal once an instruction is ever attached to a review. Not reachable today (nothing seeds one yet), and it needs a product decision (should a directed-send decision outlive a reset, or should the reset be entitled to clear it) this fix is not the place to guess at.

## Test plan

- [x] `make check-go` (build, vet, lint, unit, gates, contract drift, composition)
- [x] `make test-integration` (0 skips, 51 packages)
- [x] `TestSweepTargetsCarryNoDeleteBlockingTrigger` and `TestPreserveSetIntegrity` pass
- [x] `TestTheTwoResetsPreserveTheSameTables` (seedresetparity_test.go) passes
- [x] `craft static --strict` on the changed Go file

https://claude.ai/code/session_01GkdtztRqyypJufL6VTE2Vs

## Summary by CodeRabbit

- **Bug Fixes**
  - Communication instructions are now preserved when resetting seeded data.
  - Reset operations no longer target communication instructions for deletion, helping retain existing instruction content.

### DIFF
--- a/backend/internal/compose/datasweep.go
+++ b/backend/internal/compose/datasweep.go
@@ -96,6 +96,26 @@ var preservedResetTables = map[string]bool{
 	// It is still cleared by a reset — `activity` is swept, and the cascade takes
 	// the evidence with it. Preserved here means "not a target", never "kept".
 	"activity_retention_evidence": true,
+	// A directed-send decision (migration 1789048392): the record of who
+	// overrode a refusal, when, and under what reason. Its own trigger refuses
+	// EVERY delete unconditionally ("a communication instruction is the record
+	// of a decision and is never deleted") — unlike activity_retention_evidence
+	// above, there is no row-conditional carve-out for a cascade to pass
+	// through, so this is the audit_log/system_log shape, not that one.
+	//
+	// Not yet reachable in practice: nothing seeds a communication_instruction
+	// row today, so the sweep has never actually met one. The obligation this
+	// leaves standing: communication_review references this table with ON
+	// DELETE CASCADE, and that cascade would hit the same unconditional
+	// refusal the direct DELETE above does — sweeping communication_review
+	// will abort on the first row once an instruction is ever attached to
+	// one. Filed as margince#5287 (needs a product decision: whether a
+	// directed-send decision should outlive a reset the way this table's own
+	// preservation already says, in which case communication_review needs the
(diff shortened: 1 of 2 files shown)


tinnhlt-gra:
Good catch — fixed in the latest commit, but not the way the suggestion framed it. There isn't one policy to pick for both paths; they're mechanically different:

- **`scripts/seed-reset.sql`** runs `session_replication_role = replica`, which suppresses `communication_review`'s own `ON DELETE CASCADE` the same way it suppresses `communication_instruction`'s custom trigger — so without an explicit statement, the instruction was about to be left behind as an orphan, not cleaned up. Fixed with the same `DELETE ... WHERE NOT EXISTS` pattern `activity_retention_evidence` already uses right beside it, for the identical reason. Verified live against a seeded review+instruction pair: both clear, no orphan, no error.
- **The in-product reset** (`backend/internal/compose/datasweep.go`) runs as the application role and *cannot* suppress the trigger — the cascade there will abort the whole reset transaction the first time it meets an instruction, not silently orphan one. That's a different failure mode needing an actual product decision (should a directed-send decision outlive a reset, or should the reset be entitled to clear it), which I filed as margince#5287 rather than guess at here.

So: seed-reset.sql's gap is closed in this PR. margince#5287 tracks the remaining one, narrowed to the in-product path only.

Issue #3024, opened 2026-08-27
Running the api binary by hand needs environment nothing documents, and it fails one variable at a time


LarsGradion:
Developer friction, not a product defect — filed so it stops costing people an hour each.

`./bin/api` refuses to boot without `MARGINCE_ENV=dev` (it fails a licence check and reports the installation as production) and needs four blobstore variables. All of them are set inside `scripts/dev.sh`, so anyone starting the binary directly hits the wall. The model-key source in the local env file is documented; this is not.

It also surfaces one missing variable per boot attempt, so finding the full set takes as many restarts as there are variables.

**Done looks like:** either the requirement is documented in one place, or the api names everything it is missing in a single boot error instead of one at a time.

navdeepplaya-alt:
HI @nabilaws Can you accommodate the below in your PR as well.

**Half of this already landed.** #2054 (a week before this issue was filed) built the single aggregated boot error in `parseAPIFlags` — `api: the configuration is incomplete:` followed by a bullet list of every fault that layer can see. The gap is only that the blobstore group is checked inside `s3.New`, *after* it dials MinIO, so it can't reach that list. And the worker never got the fault list at all — it still has four sequential early returns, so `./bin/worker` fails one variable per restart even for its own flags.

So: no new mechanism. Move one knowable-without-I/O fact into the existing list, and bring the second role up to the first.

**One premise correction.** The api does not need four blobstore variables — with none set it boots and `/attachments` answers 501, and has since the seam landed. The wall appears only once `MARGINCE_BLOBSTORE_ENDPOINT` is set and the rest aren't, which is what you get copying the invocation out of `scripts/dev.sh` or sourcing a half-filled `.env.local`. `.env.example` makes it worse: it says `# default: us-east-1` above `MARGINCE_BLOBSTORE_REGION`, which the code refuses to default on purpose.

## The change

| § | Change |
|---|---|
| 1 | `Config.Faults() []string` in `blobstore/s3.go` — the ONE statement of what an object store needs. `New` calls it instead of its two sequential `if`s. `env.go` gains `configFromEnv` (shared by `FromEnv`) and `EnvFaults(lookup)`, which is just `configF
| 2 | Both roles append `blobstore.EnvFaults(config.FromOS)` into the fault list at flag time. The worker gets the fault list first — the same transform #
| 3 | Fix `api: api: blobstore: blobstore:` — the role prefix belongs to `main()` alone. |
[...]

PR #5303, opened 2026-09-11
feat(events,blobstore): opt-in Redis TLS and S3 SSE-KMS enforcement

nabilaws:
## Summary
Go-side companion to #5281 (deploy/terraform/aws): two CodeRabbit findings on that PR need a source change before the Terraform side can safely enforce them, so they're split out here rather than piling backend Go changes into an infra PR.

- **`MARGINCE_REDIS_TLS`** (api + worker, default `false`): the event-bus Redis client can now negotiate TLS. Once this ships, `elasticache.tf`'s `transit_encryption_mode` can move from `"preferred"` to `"required"` without breaking every connection the app makes.
- **`MARGINCE_BLOBSTORE_KMS_KEY_ID`** (default unset): `blobstore.Put` sends explicit SSE-KMS headers when set. Once this ships, `s3.tf`'s bucket policy can safely deny any `PutObject` that doesn't carry them, instead of trusting every future caller not to regress onto AES256 or the wrong key.

Both are purely additive — default off/unset, no behavior change for an existing deployment that doesn't set them. `ClientOptions` and `blobstore.Config`'s existing fields are untouched; `NewClient` gained one required bool param (`useTLS`), updated at its 2 real call sites plus the one direct test call.

## Related
- #5281 — once this merges, that PR's `elasticache.tf`/`s3.tf` can flip the two settings these unlock.
- #5271 (Postgres TLS) looks similar but is the opposite shape: that PR
  **enforces itself** at boot (`compose.AssertDatabaseTLS` refuses to start
  without TLS, no companion PR needed). This one only adds **opt-in**
  capability — `MARGINCE_REDIS_TLS` and `MARGINCE_BLOBSTORE_KMS_KEY_ID` are
  both default off/unset — and leaves actual enforcement to #5281's
  Terraform flips, because unlike Postgres, ElastiCache/S3 enforcement is a
  provider-side setting, not something this client can refuse on its own.

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/platform/events/... ./internal/platform/blobstore/... ./cmd/api/... ./cmd/worker/...`
- [x] `gofmt -l .` / `go vet ./...` clean
- [x] `make check-go` (build, vet, lint, arch-lint, full test suite, contract-drift, composition-reproducibility) — all green

---
## Summary by cubic
Adds opt-in TLS for the event-bus Redis client and SSE-KMS encryption for blobstore writes, so the Terraform side (PR #5281) can enforce ElastiCache's `transit_encryption_mode=required` and deny S3 writes without KMS encryption.

- New `MARGINCE_REDIS_TLS` flag (default `false`, api and worker) makes the event-bus Redis client negotiate TLS when set.
- New `MARGINCE_BLOBSTORE_KMS_KEY_ID` (default unset) sends explicit SSE-KMS headers on every `PutObject` when set.
- Both are additive; existing deployments that don't set them see no behavior change.
- `events.NewClient` gained a required `useTLS` bool, updated at its two call sites and one test call.

<sup>Written for commit 015ee2ffe5935b6ed29f1f291a26c114d5b61b19. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/5303?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added optional TLS connections for Redis used by API and worker services.
  * Added optional SSE-KMS encryption for object-storage uploads using a configured key ID or ARN.
* **Documentation**
  * Documented the new Redis TLS and blob-storage KMS configuration settings.
* **Bug Fixes**
  * Invalid SSE-KMS configuration is now reported before an upload begins.


navdeepplaya-alt:
Hi @nabilaws  
https://github.com/margince/margince/issues/3024
Have a look if you can sort the above issue with this one as well.

PR #5328, opened 2026-09-11
feat(release): the path that must always stamp proves it, and the doc says how

dungbanguyen:
Refs #1728. Builds items 2 and 3 of the ruling; item 1 is partly already in and
partly needs a surface that does not exist — see the end.

## The defect the guard has

`VERSION` becomes three things from one variable: the image tag, the OCI version
label, and the `MARGINCE_RELEASE_VERSION` build argument every role's binary is
linked with. Empty or `dev` **builds and pushes perfectly well** — and ships a set
whose mixed-release guard is inert on every installation that pulls it, because
`buildinfo.Comparable` reads both as *"this build does not know"* and unknown
disables every comparison.

That is the correct fail-safe for a local build. It is the wrong outcome for a
published release, and nothing downstream can tell them apart: the images look
finished, the roles all start, and the protection an operator believes they have
refuses nothing.

## Item 3 — the release bake refuses first

`scripts/release-version-stamped.sh`, called before the bake, with the workflow
taking its stdout as the value it stamps.

**A script rather than a clause inside the workflow**, because the guard is worth
testing and a `run:` block is not reachable from one.
`scripts/release-version-stamped.test.sh` states both halves, and the second
matters more than it looks: a guard that refused everything would be found on the
first release, while a guard that refuses nothing **looks exactly like this one**
and is found by a customer whose fleet accepted a torn set. Five cases — two
refusals, three real releases including a prerelease, which is pulled by tag like
any other and so must hold its roles together too.

Enrolled in `ROOT_SCRIPT_GATES` and documented. Mutation-checked: widening the
refusal pattern so it matches nothing fails both refusal cases.

## Item 2 — the deployment reference says what to pass

It already said a recipe that builds the targets itself gets no guard. It now says
what to do about it: the three `--build-arg` commands, that the value **need not
be a constellation release version** — the guard compares for equality and never
for order, so any stable per-build identifier works as long as all three roles get
the same one — and that it must not be empty or `dev`.

The stale forward-reference to this issue goes with it.

## Item 1 — half in, half needs a surface that is not there

The **boot-log** half is already built: `RecordInstallationRelease` and
`AssertInstallationRelease` both log `release guard inactive: this … carries no
release version`, which is the actionable wording the ruling asked for, in the
place an operator diagnosing a rollout looks first.

The **`/auth/capabilities`** half I did not build, and the reason is worth
recording. The ruling says *"the SPA already reads that surface and already knows
how to render an absent value honestly"* — it does not. `releaseskew.tsx` is a
**blocking screen**: it compares the two releases and replaces the app when they
differ. An absent `release_version` makes it do nothing, correctly, and there is
nowhere in the product that displays the guard's own state. Adding
`release_guard_active` to the contract with nothing rendering it would be a field
with no caller.

Making that visible means a new operator-facing surface — which is a real thing
to want and a bigger change than this ticket. Recorded on the issue rather than
half-built.

## Checks

`make check-backend`, `make-target-parity`, `ci-doc-parity`, and the new
`test-release-version-stamped`.

The page preamble folds by one line: `make-targets.md` was at its 400-line budget
and the new row took it over. Every fact it carried is still there.

## Summary by CodeRabbit

- **New Features**
  - Release images now receive a validated release version before publishing, preventing unstamped image sets from being released.
  - Valid release identifiers are preserved for image tags and release artifacts.

- **Documentation**
  - Expanded deployment guidance for stamping locally built images and explained release-version validation behavior.
  - Documented the new release-version validation test target.

- **Tests**
  - Added coverage for rejecting empty or development release versions and accepting valid identifiers.


dungbanguyen:
Fixed, and held.

You are right that the message overstated the rule: the guard refuses empty and `dev` and accepts any other identifier, so naming the `YYYY.edition` scheme as a requirement would send a release engineer to satisfy the wrong thing at the moment a release is already blocked.

It now names the two values it refuses, and says separately that the workflow supplies the drafted release here while any other stable identifier would pass — which keeps the useful half of the old sentence without the claim.

I also added the assertion to `release-version-stamped.test.sh`: a refusal must name both refused values. The message and the rule drifted once; a test is what stops them drifting twice. Mutation-checked by restoring the old sentence.

PR #5336, opened 2026-09-11
feat(ci): a pull request declares what it closes, and the report warns when it does not

### LINKED ISSUE #2133: main is red: the data reset's sweep gate refuses activity_link's new project-clock trigger

dungbanguyen:
Refs #1737. Builds the ruling: require a declaration, warn rather than block.

**Closes: none** — the issue stays open for the promote-to-blocking decision,
which wants evidence this has not gathered yet. (That line is also this change
declaring itself.)

## The defect is a reference nothing reads

A pull request body said in as many words *"Closes the residual half of #548"*
and its metadata referenced nothing: GitHub parses a closing keyword only when
the issue number follows it **directly**. So the prose closed nothing, the issue
stayed open six days after its fix merged, and it was re-queued as new work. The
author did write it down; they wrote it somewhere nothing reads.

So the check reads `closingIssuesReferences` — the list that actually closes an
issue on merge — and never the body's prose. **A version that grepped the body
would pass that exact pull request** and report the defect as absent, which is why
that case is in the test by name and why the mutation that reads the prose fails
it.

## The "closes none" half

`Closes: none`, on a line of its own. Most pull requests close no issue and are
right not to; without a way to say so this would either nag them for ever or have
to guess which ones meant it. Matched at the start of a line, so a sentence
mentioning the token in passing is not a declaration — also mutation-checked.

## It warns and never blocks

`ci` is the one required check, but a red check of any kind leaves a pull request
blocked with administrator override as the only way past — so a job that failed on
a finding would be a gate whatever its name said. `review-coverage.yml` learned
that the hard way and this follows it.

And the evidence says the mechanism is fine. The second dataset on the issue
appeared to show four pull requests ignoring the trailer; its author checked,
found it false, and said so after having posted the wrong reading in three places.
The verified timeline shows `Closes #2133` opened 112 seconds after the issue was
filed and GitHub closing it on merge. **Usage is patchy, not the machinery**, and a
hard block on a mechanism that mostly works buys friction on every pull request to
fix a habit.

`edited` is a trigger so the sticky comment **withdraws itself** the moment the
author adds the line.

## What it deliberately does not do

Decide whether a pull request fixes an issue it did not mention. That is the real
defect and it is not machine-decidable; the declaration is the affordance that
makes a human answer it.

## Two gates caught real things while writing this

- `TestNoShellGateReadsAVerdictThroughAPipeThatCanBreak` — my first draft piped
  the body into `grep -q`, which exits on its first match and leaves the producer
  with EPIPE, so under `pipefail` **a match would have read as no-match**, on some
  machines and not others. Now a here-string.
- `TestEverySourcePathNamedInProseExists` — the doc named the script before it
  was tracked.

## Checks

`make check-backend`, `make-target-parity`, `ci-doc-parity`, and the new
`test-closing-declaration` (8 cases, mutation-checked both ways).

## Summary by CodeRabbit

* **New Features**
  * Added automated pull request feedback for missing closing declarations.
  * Recognizes issues GitHub will close and explicit “closes none” declarations.
  * Posts or updates a non-blocking status comment with the results.

* **Documentation**
  * Documented the new workflow, triggers, and declaration behavior.

* **Tests**
  * Added fixture-based coverage for declared and undeclared pull requests.

### DIFF
--- a/.github/workflows/closing-declaration.yml
+++ b/.github/workflows/closing-declaration.yml
@@ -57,6 +57,13 @@ jobs:
           # The METADATA, not the body. This is the same list that closes an
           # issue on merge, so a pull request this reads as declaring is one
           # GitHub will actually act on.
+          #
+          # A FAILED QUERY IS NOT AN EMPTY ONE. Suppressing the failure would
+          # hand the check an empty list, and a pull request carrying perfectly
+          # good closing metadata would be warned for a transient API blip —
+          # the check would be loudest exactly when it knows least. The job
+          # still succeeds; it just says nothing rather than something false.
+          asked=0
           refs="$(gh api graphql -f query='
             query($owner:String!, $repo:String!, $pr:Int!) {
               repository(owner:$owner, name:$repo) {
@@ -67,7 +74,11 @@ jobs:
             }' -F owner="${{ github.repository_owner }}" \
                -F repo="${{ github.event.repository.name }}" \
                -F pr="$PR" \
-            --jq '.data.repository.pullRequest.closingIssuesReferences.nodes[].number' || true)"
+            --jq '.data.repository.pullRequest.closingIssuesReferences.nodes[].number')" || asked=$?
+          if [ "$asked" -ne 0 ]; then
+            echo "could not read this pull request's closing references; saying nothing rather than guessing" >&2
+            exit 0
+          fi
 
           status=0
           out="$(CLOSING_DECL_REFS="$refs" CLOSING_DECL_BODY="$BODY" \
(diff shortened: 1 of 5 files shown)


dungbanguyen:
Both fixed — they were right, and the second one caught my PR body contradicting my code.

**The metadata query.** `|| true` handed the check an empty list on any transient API or auth failure, so a pull request carrying perfectly good closing metadata would be warned for a blip — the check at its loudest exactly when it knows least. It now exits saying nothing, which is the honest answer and still fails no job.

**The comment.** I wrote that it "withdraws itself" and then rewrote it into `This pull request declares what it closes.` — a permanent line on every pull request that ever forgot one, which is a worse record than the omission was. A finding that resolves itself should leave nothing behind, so it is deleted. The doc says so now too.

PR #5339, opened 2026-09-11
fix(consent): a claim its purpose contradicts is refused, not reconciled (#5148)

### LINKED ISSUE #5148: Promotional mail can ride the business_correspondence lane via the legacy purpose fallback, so a marketing objection never binds it
Found while attempting R3 ("an unsubscribe must not stop requested correspondence"). This is the reason that finding **cannot** be fixed by narrowing the unsubscribe sweep alone, and it is the more serious half of the problem.

## The escape path

1. A caller sends with `communication_context=marketing` and `consent_purpose=business_correspondence`.
2. `resolveCategory` finds no thread, no live deal, and no evidence supporting the marketing claim, so the resolution is unsupported.
3. `decideResolved` hands the unsupported claim to `legacyVerdictFor`, which **overwrites the resolved category from the purpose's class** (`authorizedecide.go`, the `d.Resolved = resolutionForClass(purpose.Class).Category` line). A `business_correspondence` purpose remaps the message to `reply_to_inbound`.
4. `reply_to_inbound` is authorized by `correspondenceVerdict`, which accepts a person-wide "qualifying event" — any recent inbound message, meeting or exchange with that person. Not evidence that *this* outgoing message is a reply to anything.
5. The message sends as correspondence while carrying promotional content.

## Why it matters more now

A `marketing_objection` suppression binds `CategoryMarketing` only. A message that resolved to `reply_to_inbound` is never tested against it. So the objection writer that just landed can be **completely bypassed** by tagging promotional mail with the correspondence purpose key — the stop is recorded, visible in the export, and does not stop the mail.

The same shape defeats the obvious fix for R3: withdrawing only marketing purposes on unsubscribe, and even withdrawing-plus-recording-an-objection, both leave this lane open.

## Why R3 is blocked on this

The existing test `TestUnsubscribeAllStopsALiveBusinessCorrespondenceLane` encodes a real past incident: an unsubscribe that skipped the correspondence purpose "walked straight past the only lane the recipient was actually receiving mail under, while reporting success". That test is still load-bearing **because of this path**. Narrowing the unsubscribe to marketing-class purposes today would reopen it.

So the order has to be: close this, then narrow the unsubscribe. Doing it the other way round trades a known over-broad unsubscribe for a silent under-broad one, which is the worse failure — the recipient believes they have stopped something they have not.

## What the fix looks like

The concept's canonical-purpose work (W1/R21) is the real home: one resolved purpose used for authorization, footer, and withdrawal target, with a contradictory claim refused at the door rather than silently remapped. Concretely, either:

- a message claiming `marketing` must not be authorizable by a correspondence-class purpose key at all — the conflict is a 422, not a downgrade; or
- `legacyVerdictFor` must stop overwriting `Resolved` from the purpose class when the caller claimed a different category, so the objection is tested against what the message actually is.

The first is the concept's stated direction. Either way it needs its own slice and its own tests through the real send path, not a patch to the unsubscribe sweep.

## Caveat

This is a reachable code path established by reading the code. Whether any real campaign currently sends promotional content under the `business_correspondence` purpose key needs runtime data I do not have.

dungbanguyen:
Closes #5148.

## The escape path, reproduced

A send claims `communication_context=marketing` and names `consent_purpose=business_correspondence`. No thread and no live deal support it and the marketing claim has no evidence, so `legacyVerdictFor` overwrites the resolved category **from the purpose's class** — remapping the message to `reply_to_inbound`, which `correspondenceVerdict` authorizes on any recent inbound from that person. Not evidence that *this* message is a reply to anything.

**And it goes out past the subject.** An Art. 21 objection binds `CategoryMarketing` and is tested against the RESOLVED category, so a message remapped to `reply_to_inbound` was never put to it. The stop is recorded, it appears in the subject's own Art. 15 export, and it does not stop the mail.

The issue flagged this as read from the code. It reproduces: with this fix reverted, `TestAMarketingObjectionIsNotWalkedPastByACorrespondencePurpose` comes back **`allow`** for a message the caller labelled `marketing`, to a person carrying a live `marketing_objection`.

## The fix

The issue's stated direction, and the concept's: the conflict is refused rather than downgraded. A caller who says what a message is and names a purpose meaning something else has asked two things at once, and the engine declines rather than picking one.

**Absolute**, for the reason `ReasonNoSubject` is. Softening it does not fall back to a weaker reading of the same message — it falls back to the old gate, which answers on the purpose key alone and so authorizes exactly what the refusal exists to stop. It is classified `LevelMachine`: nothing about the recipient refuses it, and the remedy is to send it again with a purpose that means what was claimed.

## What it does NOT change

**`Resolved` still takes the purpose's reading, never the claim.** My first attempt set it to the claim so the objection would bind — and `TestAnUnsupportedClaimIsRecordedButNeverResolvedTo` caught it. That test is right: an unproven claim in that column selects the rollout mode and decides whether the advertising ceiling is counted, so a caller who set it would hold both. The claim is already in `Requested`, so the row says what was asked for, what it would have been read as, and that the two disagree — and the message does not send, which is what the objection needed.

**A claim its purpose agrees with** still goes down the legacy lane it always did, and **a caller who claims nothing** is untouched. Refusing either would refuse every message that has not moved to the category vocabulary.

## Proof

Four cases in `authorizecontradiction_integration_test.go`, over a real Postgres through `decideOne` — the whole per-recipient decision, not just the resolver:

- the contradictory pair is refused, with the claim preserved in `Requested` and the engine's own reading in `Resolved`;
- the objection is not walked past — seeded with a qualifying event **on an unrelated thread**, which is precisely what the correspondence arm authorizes on, so the old path had everything it needed;
- an agreeing claim is not called a contradiction;
- a purpose with no claim still speaks for the message.

Mutation-verified: removing the refusal turns the first into `review` and the second into `allow`.

`make check` green, `make craft-static` clean whole-tree, the consent module lane and the compose integration lane green.

## What this unblocks

The issue records that R3 ("an unsubscribe must not stop requested correspondence") is sequenced behind this, because narrowing the unsubscribe to marketing-class purposes while this lane was open would trade a known over-broad unsubscribe for a silent under-broad one. That ordering is now clear.

## Summary by CodeRabbit

- **Bug Fixes**
  - Requests are now refused when the stated category conflicts with the category implied by the selected purpose.
  - The refusal is enforced consistently across rollout modes and clearly identifies the category mismatch.
  - Marketing objections can no longer be bypassed by remapping a purpose to correspondence.
  - Requests without a stated category continue to be handled unchanged.
  - Compatible purposes continue through the existing authorization flow.

### DIFF
--- a/backend/internal/modules/consent/authorizecontradiction_integration_test.go
+++ b/backend/internal/modules/consent/authorizecontradiction_integration_test.go
@@ -1,0 +1,129 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+//go:build integration
+
+package consent
+
+// A caller who says what a message IS and names a purpose meaning something
+// else is refused, rather than quietly sent under the purpose's reading.
+//
+// The reading was the escape path. A send claiming `marketing` under the
+// `business_correspondence` key resolved to reply_to_inbound, which the
+// correspondence arm authorizes on any recent exchange with that contact — not
+// evidence that THIS message is a reply to anything. So promotional content
+// went out as correspondence.
+//
+// And it went out past the subject. An Art. 21 objection binds
+// CategoryMarketing and is tested against the RESOLVED category, so a message
+// remapped to reply_to_inbound was never put to it: the stop was recorded,
+// appeared in the subject's own export, and did not stop the mail.
+
+import (
+	"testing"
+	"time"
+
+	"github.com/margince/margince/backend/internal/shared/ports/commsauthz"
+)
+
+// claimingMarketingUnder is the request at the heart of this: the caller names
+// the category and a purpose key from another class.
+func claimingMarketingUnder(purposeKey string) commsauthz.Request {
+	return commsauthz.Request{
+		Context:          commsauthz.CategoryMarketing,
+		LegacyPurposeKey: purposeKey,
+	}
+}
+
+// The escape path itself, refused.
+func TestAClaimContradictedByItsPurposeIsRefused(t *testing.T) {
+	e := setupResolve(t)
+	e.seedPurpose(t, "business_correspondence", "business_correspondence")
+
+	d := e.decide(t, claimingMarketingUnder("business_correspondence"))
+
+	if d.Verdict != commsauthz.VerdictDeny {
+		t.Fatalf("verdict = %q, want deny — a request that claims one category and names a "+
+			"purpose meaning another asks two things at once", d.Verdict)
+	}
+	if d.ReasonCode != commsauthz.ReasonPurposeContradictsClaim {
+		t.Errorf("reason = %q, want %q", d.ReasonCode, commsauthz.ReasonPurposeContradictsClaim)
+	}
+	// The row says BOTH things. Requested is what the caller asked for, so an
+	// auditor can see that promotional content was claimed; Resolved stays the
+	// engine's own reading, because an unproven claim must never reach the
+	// column that selects the rollout mode and counts the advertising ceiling.
+	if d.Requested != commsauthz.CategoryMarketing {
+		t.Errorf("requested = %q, want %q — the claim has to survive into the row",
+			d.Requested, commsauthz.CategoryMarketing)
+	}
+	if d.Resolved != commsauthz.CategoryReplyToInbound {
+		t.Errorf("resolved = %q, want %q — the engine's own reading, never the claim",
+			d.Resolved, commsauthz.CategoryReplyToInbound)
+	}
+}
+
+// The half that makes it a privacy defect rather than a tidiness one: with the
+// remap in place the objection was never reached.
+func TestAMarketingObjectionIsNotWalkedPastByACorrespondencePurpose(t *testing.T) {
+	e := setupResolve(t)
+	e.seedPurpose(t, "business_correspondence", "business_correspondence")
+	e.suppress(t, "marketing_objection")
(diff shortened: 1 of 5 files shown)


dungbanguyen:
Addressed in `bd0feb06b`. You were right, and the path you named was the one this branch's own premise made easiest to reach.

`resolveCategory` answers from EVIDENCE first — a reply in the thread, a live deal on the links — and both arms return before `legacyVerdictFor` is ever entered. So a caller claiming `marketing` about a conversation the subject started was allowed, resolved as `reply_to_inbound`, and the withdrawal read below it asked about a category an Art. 21 objection to direct marketing does not bind. The claim was put to nothing at all. That is the same defect this PR was opened to close, surviving in the arm that had no test.

What changed:

- `contradictsClaim` and `denyContradiction` are now ONE spelling of the rule, called from both paths. Two copies is what produced this — the guard was written where the defect was found and the other path was never asked the question.
- The refusal keeps `Resolved` on the RECORD's reading and the claim in `Requested`, on both paths, so an unproven claim still never reaches the column that selects the rollout mode and counts the advertising ceiling.
- The reason code is renamed `claim_contradicts_resolution`. `purpose_contradicts_claim` named the SOURCE of the disagreement, and the source is the part that differs between the two paths; the refusal is one thing. Nothing on the wire carries the old code — it was introduced by this same branch.

Three tests, all mutation-verified: the thread-resolved contradiction is refused, the objection is no longer walked past by a thread, and an UNCLAIMED thread reply is still allowed — that last one matters because without it the guard would refuse every ordinary reply in the product.

Issue #495, opened 2026-08-05
Auth rate limits are process-local, so a multi-replica deployment multiplies every ceiling by N


tinnhlt-gra:
## What

Every limiter in the identity surface — login failures, login per IP, forgot-password per (email, IP) and per IP, and now set-password-link issuance per actor and per target — keeps its counters in **in-process maps**. The scope is documented in the struct comment ("single-binary scope"), so this is a known property rather than a surprise, but it is worth an issue now that a security-relevant ceiling rides on it.

## Why it matters more than before

The set-password-link limiter is not an anti-anonymous-abuse throttle like the others. It bounds what a **stolen admin session** can do, and each issue **supersedes the target's outstanding token**. So with N load-balanced API replicas an attacker gets `5 × N` supersessions per member per hour instead of 5, and can churn a member's recovery link by replica-hopping — a denial-of-recovery primitive the stated ceiling is meant to close. A process restart also clears the budget.

Found during review of the ADR-0061 Amendment 1 work; not introduced by it.

## Suggested direction

A shared counter (the installation already runs Redis) for at least the security-relevant limiters, or an explicit statement in the deployment docs that the API is single-replica and the ceilings assume it. Today the contract's `429` semantics imply an installation-wide bound the implementation does not provide.

## References

- `Handlers` limiter fields and their comments (identity)
- `platform/ratelimit`

dungbanguyen:
## Decision, 2026-08-31: move every limiter to Redis

Re-triaged against `origin/main` at `8ecb691f1`. Unchanged.
`backend/internal/platform/ratelimit/ratelimit.go:69-70` still holds:

```go
starts  map[string]time.Time
counts  map[string]int
```

and the package comment still states the property rather than fixing it: *"In-process
is the honest scope for a single-binary PoC — a multi-replica deployment paces
each replica's own view — and moving the same keys into Redis would not change
callers."*

That last clause is what makes this decision cheap, and it is the reason for the
ruling below.

### The ruling

**Every limiter moves to Redis, not just the security-relevant ones.**

Splitting the mechanism was the smaller change and is deliberately not taken. A
split leaves a rule that has to be remembered at the moment somebody adds the
next limiter — *"is this one anti-abuse or is it a security ceiling?"* — and the
whole point of this ticket is that the distinction was not visible until somebody
looked. The set-password-link limiter was written as though it were an
anti-anonymous-abuse throttle like its neighbours; it is not, and nothing in the
[...]

PR #5354, opened 2026-09-11
fix(uploads): the grant is asked before the body is read, not after it

### LINKED ISSUE #1567: Attachment uploads have no aggregate bound: no concurrency limit, no per-workspace quota, and authorization runs after the bytes
Now that the multipart ceiling is reachable (#1542, fixed in #1539), a single attachment upload is bounded but the *aggregate* is not. Four separate gaps, all authenticated-only — a session with an object grant on the parent, so insider or stolen-credential rather than anonymous.

**1. Per-request memory is roughly twice the file.** `UploadAttachment` calls `ParseMultipartForm(maxAttachmentBytes)` — where the argument is the in-memory/spill threshold, so file parts stay in memory rather than spilling — and then `io.ReadAll(file)` on top, and the store takes `Body []byte`. Peak is ~25 MB (parts) + net/http's ~10 MB non-file reserve + ~25 MB (the ReadAll copy) ≈ 60 MB per in-flight request.

**2. No concurrency or rate limit on the route.** `grep -n 'semaphore|MaxConcurrent|inflight' internal/platform/httpserver/` finds nothing, and the endpoint has no per-principal throttle. ~200 concurrent uploads is ~10 GB.

**3. Authorization runs after the bytes are read.** Authentication precedes the handler, but the object grant does not: `auth.Require(ctx, in.EntityType, principal.ActionUpdate)` runs in the store (`activities/attachment.go`), after `ParseMultipartForm` and `io.ReadAll` in the handler. So a session holding **no grant on any object type** can spend the full per-request cost repeatedly and receive a 403 each time. Hoisting the grant check into the handler before the parse would make a refusal cost nothing.

**4. No per-workspace attachment quota.** `grep -i 'quota|storage_bytes' modules/activities platform/blobstore` finds nothing, so an unbounded count of 25 MB objects is a storage bill with no ceiling. Distinct from the per-request bound, which is fine.

### One functional inconsistency in the same area

`maxAttachmentBytes` (25 MB) now exceeds `maxSendBytes` (20 MiB, `compose/commsattachments.go`), so **uploads in the 20–25 MB band are accepted and can never be emailed**. Either the send path grows to match or the upload cap comes down to it; today a reader can file a document the product will silently never attach to a message.

Related, same file: `readOne` does an unbounded `io.ReadAll` on the stored object and checks `maxSendBytes` *after* the read, so a worker now buffers a full 25 MB attachment before refusing it.

### Why this is hardening and not an advisory

Every item needs an authenticated session, and per-request cost is bounded — this is defence-in-depth against an insider or a stolen cookie, not a reachable breach. Per [SECURITY.md](SECURITY.md), an exploitable weakness goes to a private advisory instead; nothing here is one.

Found by the adversarial review pass on #1539.

dungbanguyen:
Closes #1567 — item 1 of the ruling, widened to every route with the same defect. Items 2 and 3 are not in here; why is at the bottom.

## The defect, and that it was not one route

The ruling named the attachment upload: `auth.Require` sits in the store, the store runs after the handler, and the handler's first act is `ParseMultipartForm`. So a session holding **no grant on any object type** spends one request and makes the server parse and spill a whole file before answering 403 — a free amplifier, repeatable.

Checking the other multipart parses in the tree, three more had the same shape:

| route | authorized before the parse? |
|---|---|
| `POST /v1/imports/sources` | **yes** — `stageImportSource` requires `import_run.create` first |
| `POST /v1/company/logo{,/icon}` | **yes** — reads the anchor company first, which requires `company.read` |
| `POST /v1/attachments` | no |
| `POST /v1/knowledge/corpora/{id}/documents` | no |
| `POST /v1/me/linkedin-connections` | no |
| `POST /v1/people/vcard-import` | no |

The import lane is the exemplar; the other three are the same bug two modules over. Fixing only the one under review is the catch `AGENTS.md` names as the recurring reviewer finding, so all four go in one change.

## What each now asks

Three of them can ask **exactly** what their store asks, so the early check cannot refuse anybody the store would admit:

- vCard import — `person.create` **and** `person.update`, because `importCardsFrom` requires both: it creates the people no card matches and updates the ones a card does.
- LinkedIn import — `person.create`.
- Corpus document — `knowledge_document.create`.

The attachment upload cannot: **the object type arrives inside the body it has not parsed yet.** So it asks the coarse half — may this caller write anything at all — through a new `auth.RequireAnyObject`, the mirror of the existing `RequireAny` (which knows the object and not yet the action; this knows the action and not yet the object). A caller it turns away holds no write grant anywhere, so the store would have refused it whatever object the body went on to name.

**None of this moves the gate.** Every store keeps the check it had, because the store is the gate every transport passes. These are early refusals.

## Held by a fitness function, not by four call sites

`backend/gates/uploadauthorder_test.go` — `TestEveryMultipartParseRefusesBeforeItReads`. It derives its subject from the tree (any function containing `ParseMultipartForm` or `MultipartReader`), so a seventh upload route is in the obligation set the moment it is written rather than when someone remembers. It checks **order**, not presence: moving a check below the parse fails it, which is how it was mutation-checked.

The two parses whose authorization belongs to their caller are waived by name, each entry naming that caller so the claim can be checked; a waiver no parse reaches is reported stale rather than left to sit.

And the attachment handler's ordering is held as behaviour too, in both directions — the direction that matters for a coarse check is the second one:

- a grantless session gets 403 and its body is **never read**;
- a session that may write *something* still reaches the parse, so the coarse check is not turning away callers the store would admit.

## What is deliberately not here

**Item 2, the per-principal concurrency bound.** The ruling says to build it on the Redis limiter from #495 "rather than an in-process counter, or it will be the next thing to multiply by replica count". #495 is still open and `platform/ratelimit` is still the in-process fixed-window limiter — and it is a *rate* limiter, where item 2 wants a *concurrency* bound. Building it here means either building the thing #495 is about, or building exactly what that issue says not to. It waits on #495.

**Item 3, the per-workspace quota.** As ruled: it needs a number nobody has chosen, and it is a product decision.

**One thing fixed on the way**: the attachment handler's `entity_type` refusal listed five types and the enum has six — `project` was missing from the message, so a caller uploading against a project was told their valid value was invalid.

## Checks

`make check-backend` green, `make craft-static` 0/0/0. The gate inventory page and the fixture-annotation pin move with the new gate, as their own gates require.

### DIFF
--- a/backend/gates/gatecensus_test.go
+++ b/backend/gates/gatecensus_test.go
@@ -52,7 +52,7 @@ const (
 	// Permitting the marker without counting it would let it spread quietly and
 	// reopen the class these rules close; pinned, every new one moves a number a
 	// reviewer sees.
-	wantFixtureAnnotations = 50
+	wantFixtureAnnotations = 51
 )
 
 // censusDecl is one package-level declaration this census governs — a map from
--- a/backend/gates/uploadauthorder_test.go
+++ b/backend/gates/uploadauthorder_test.go
@@ -1,0 +1,178 @@
+// SPDX-License-Identifier: BUSL-1.1
+// SPDX-FileCopyrightText: 2026 Gradion
+
+//gate:kind prohibition H1
+
+package gates
+
+// An upload route refuses an unauthorized caller BEFORE it reads the body.
+//
+// The gate is about ORDER, and order is the whole of it: every one of these
+// routes was already authorized, and every one of them authorized too late.
+// The store is where the grant decides, the store runs after the handler, and
+// the handler's first act was to parse a multipart body — so a session holding
+// no grant at all sent one request, made this server take a whole file apart
+// and spill it to disk, and got a 403. Repeatable, and free to the sender: the
+// attacker spends a request and the server spends a file.
+//
+// What the early check is NOT is the gate. It is deliberately allowed to be
+// coarser than the store's — the attachment upload cannot even know which
+// object type the body names until it has parsed it, so it asks whether the
+// caller may write ANYTHING — and the store still asks the exact question
+// afterwards, because the store is the gate every transport passes.
+//
+// Derived from the tree rather than from a list of routes: the subject is any
+// function that parses a multipart body, so a seventh upload route is in the
+// obligation set the moment it is written, which is the only way this stays
+// true. A parse whose caller authorizes instead is waived BY NAME below, with
+// the caller named, because that shape is correct and unprovable from inside
+// one function.
+
+import (
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"io/fs"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+// The two ways a handler takes a multipart body apart. Both are the cost this
+// gate is about: ParseMultipartForm reads the whole body, and MultipartReader
+// hands back a stream the caller then reads.
+var multipartParsers = []string{"ParseMultipartForm", "MultipartReader"}
+
+// gatekit:fixture the parses whose authorization is their CALLER's, not their
+// own — a waiver, and each entry names the caller that holds the grant so the
+// claim can be checked rather than taken.
+//
+// This is the one shape the rule cannot see: a helper that only decodes is
+// right to hold no grant, and whether its caller holds one is a fact about a
+// different function. An entry here is a claim about that caller, so it goes
+// stale the way any claim does — which is why an entry no parse reaches is
+// reported below rather than left to sit.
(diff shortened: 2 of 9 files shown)


dungbanguyen:
Rebased onto main. The integration failure on the previous run was not this change: `TestMigrationsBuildTheCommittedSchema` fails on pristine main too, since #5349 committed a head catalog naming a foreign key its own migration deliberately refused to create. Fixed separately in #5356 — this PR goes green behind it.

PR #5439, opened 2026-09-12
fix(deals): the ARR lock names which figure moved

dungbanguyen:
`main` is craft-red:

```
backend/internal/modules/deals/dealmoney.go
  163: [MAJOR] boolean-trap — 2 bool parameters are a boolean trap — use named types or an options struct
```

Checked on pristine `main` at `e9ca2fb79` before touching anything; no open PR is fixing it. Every open PR's `craftsmanship` job fails on it.

## The fix the file already contains

`refuseManualArrEdit` takes `(arrMoved, currencyMoved bool)`, and `(true, false)` at the call site says nothing about which is which — a transposed pair reads exactly like a correct one, and these two decide whether a signed offer's recurring figure may be overwritten by a manual edit.

Twenty lines up, `moneyRestatement` is a named struct for the same shape, with the reason spelled out:

> Named rather than three bare bools at the call site, where `(true, false, true)` says nothing about which field is which and a transposed pair reads exactly like a correct one.

`moneyMoved` is its twin: that type says which figures a *request carried*, this one says which the *patch changed*. Same two questions the money path has always asked apart.

`make craft-static` is 0/0/0 and the deals package is green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of deal money changes when editing offer-derived deals.
  * ARR changes, currency movements, and cleared values are now correctly identified and reported.

### DIFF
--- a/backend/internal/modules/deals/deal.go
+++ b/backend/internal/modules/deals/deal.go
@@ -343,7 +343,10 @@ func (s *Store) applyMoneyInvariants(ctx context.Context, tx pgx.Tx,
 	// another offer is how it changes, and that path replaces the figure and
 	// its provenance together.
 	_, arrMoved := after[arrField]
-	if err := refuseManualArrEdit(current, resultingArr, arrMoved, currencyMoved); err != nil {
+	if err := refuseManualArrEdit(current, resultingArr, moneyMoved{
+		Arr:      arrMoved,
+		Currency: currencyMoved,
+	}); err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


dungbanguyen:
Closing as a duplicate: #5435 is already up, green, and fixes the same boolean trap the same way (`arrEditMove` where this used `moneyMoved`). It also carries the regression that made main red in the first place, which this did not.

My mistake in the order — I checked the craft failure against pristine main before writing anything, but searched the open PRs afterwards rather than first. Nothing here is worth landing beside theirs.

PR #5451, opened 2026-09-12
feat(settings): a pipeline can be retired from the screen, and put back

### LINKED ISSUE #5352: A pipeline can be retired and restored through the API — the screen offers neither, and an archived pipeline simply vanishes with no way back
## Observed

Measured on `qc-own` at core `218d026b7` (`archivePipeline`/`restorePipeline`, `DELETE`/`POST /pipelines/{id}(/restore)`, added core `413380cc4`, #5155).

The API works exactly as its own contract describes. A disposable pipeline ("QC Throwaway Pipeline", not the default), created via `POST /pipelines`, `admin@demo.test`, cookie session:

| call | response |
|---|---|
| `DELETE /pipelines/{id}` | `204` |
| `GET /pipelines` (no `include_archived`) | the pipeline is gone from `data` |
| `GET /pipelines?include_archived=true` | the pipeline reappears, archived |
| `POST /pipelines/{id}/restore` | `200`, the pipeline body, `version` incremented |
| `GET /pipelines` again | the pipeline is back |
| `DELETE` on the seeded default pipeline ("Sales") | `422 validation_error`, `code: default_pipeline_not_archivable`, `detail: "this is the default pipeline, and new deals need one; make another pipeline the default first, then retire this one"` |

The screen offers none of this. Read from `frontend/src/screens/settings.tsx`:

- `PipelinesCard`'s query is `GET /pipelines` with no `include_archived` — an admin can never see an archived pipeline from this screen, only lose track of it.
- `PipelineRow` (the same file, the row `PipelinesCard` renders per pipeline) draws exactly two write affordances when `canEdit` — `EditAction` (rename/reposition/set-default) and `StageCreate` — and nothing else. No archive button, no kebab menu, no overflow action of any kind.

So the whole path this issue describes — retiring a pipeline nobody uses, or undoing that — has no button anywhere in the product for a human to press. An admin who archives one through the API has put it somewhere the UI cannot show them again: `PipelinesCard`'s query would need `include_archived=true` and `PipelineRow` would need a restore affordance, neither of which exists.

Not a fixture or setup artifact: reproduced with a throwaway pipeline created and archived for exactly this check, cleaned up (re-archived) afterward.

## Cost

Nobody using only the product can retire a pipeline, or recover one that got retired. The action is real, reversible, and well-designed at the API layer — the 404/409/422 responses are exactly what a UI would need to build a good retire/restore flow — but none of it is reachable without direct API access or an internal tool. A company that stands up a second pipeline for a new business line, decides later to retire the one it replaced, and wants that choice reflected in the product (dropped from the picker, kept for history) has no way to make that happen through the app at all.

This is not data loss and nothing else on the pipeline is affected — deals on an archived pipeline keep their stage, history and forecast contribution untouched, per the API's own description, and re-verified live above (the throwaway pipeline's own stage survived the archive/restore round trip, `version` incrementing normally). The cost is entirely that the capability is invisible, not that it is unsafe.

## Workaround

None through the product. An admin who needs a pipeline retired or restored has to go through the API directly (`DELETE`/`POST /pipelines/{id}/restore`, cookie session, admin role) or ask someone who can.

---
Found by [margince-qc](https://github.com/margince/margince-qc), exploring docs/TODO.md's own "unbuilt coverage" note for core #5155 before writing a scenario for it — reading `PipelineRow`/`PipelinesCard` first turned up that the screen this note assumed existed does not, so this is filed as a product gap rather than built as a walk of a control that is not there.

dungbanguyen:
Closes #5352.

Retiring a pipeline and putting one back have worked at the API since #5155.
The screen offered neither, so the only way to stop offering a pipeline nobody
uses was to call the endpoint by hand — and an admin who did that put it
somewhere the UI could not show them again, because `PipelinesCard` read
`GET /pipelines` with no `include_archived` and a retired pipeline simply
vanished from the one screen that administers them.

Adding `include_archived=true` to that query would have been wrong, and silently.
`PipelinesCard` reads the `["pipelines","all"]` cache entry that the deal board,
the stage-automation picker and the lead qualifier also read. Widening it puts
every retired pipeline straight back into the three surfaces that retiring one
removes it from — through a shared cache, with nothing on the settings screen to
show for it.

So the card takes its own key, `["pipelines","all","including-retired"]`. It still
matches the `["pipelines"]` prefix every mutation here invalidates, so both shapes
stay fresh. `asks for archived rows on a request of its own` holds it.

## Two grants, because the server checks two

Retiring is `pipeline:delete` (`ArchivePipeline` → `auth.Require(pipeline, delete)`)
— the same verb stage removal already takes. Putting one back is `pipeline:update`.
Gating both on `canEdit` would have drawn a Retire button the server only ever
403s, which is the defect the card's existing "withholds stage removal from a
principal holding update alone" test guards against one row down.

The default pipeline's control is **drawn and refused**, with the server's own
remedy beside it, rather than hidden: STATE-4a — a control blocked by state rather
than permission keeps its place and says why, and here the why names something the
reader can do from that same row (make another pipeline the default first).

The `DELETE` carries `If-Match` through the house `ifMatch()` helper and
`requireVersion`. I hand-rolled the header first; `if-match-coverage.test.ts` is
the census that catches exactly that, and it did.

## The file split

`settings.tsx` was over the length cap under a frozen waiver, so the pipeline UI
leaves it. The ladder then splits from the pipeline the way the records do:
`settings.stages.tsx` names no pipeline type and imports nothing from
`settings.pipelines.tsx`, so the direction reads from the imports alone. The waiver
drops from 2811 to 2312 — the list's only permitted direction.

## One stale claim corrected

`archive.tsx` said "there is NO restore endpoint in the contract, so this hook and
action are archive-only — never wire a restore control against them". The first
half is no longer true. The rule survives the correction and now says why on its
own terms: that hook owns a confirm, an invalidation and a toast announcing a
record is gone, and a restore is none of those. The undo belongs where the retired
record is listed, which is where a reader looking for it will be.

## Summary by CodeRabbit

* **New Features**
  * Added pipeline management in Settings, including pipeline creation, editing, stage management, and ordered stage ladders.
  * Added pipeline retirement and restoration, with retired-status indicators and safeguards preventing retirement of the default pipeline.
  * Added stage details including outcome labels, win probabilities, and exit criteria.
  * Added read-only guidance when users lack pipeline editing permissions.
* **Localization**
  * Added English, German, and Vietnamese translations for pipeline retirement and restoration workflows.
* **Documentation**
  * Updated archive-action guidance to reflect pipeline restoration.


dungbanguyen:
All five addressed in `37f52b2b4`.

**Restore had no error feedback** (major) — the sharpest of the five. `ArchiveAction` announces a refused retire through a `role="alert"` line; restore just came back enabled, so a failure and a success looked identical while the row behind both still read *Retired*. It now says what the server told it, in the same treatment.

**`disabled` where the design system offers `pending`** — right, and the distinction is why the prop exists: `pending` keeps the control focusable and refuses the repeat press itself, while a natively disabled button drops focus to `<body>` mid-press.

**The German said `Sie`** — `de.ts` states its own register in a comment at the top of the file, informal `du`. The confirm also left `sie` ambiguous between the pipeline and the nearer plural `Deals`, so it names `die Pipeline`.

**The restore test matched the path and accepted any verb** — the stub records the method now and the test asserts POST.

**`StageRemove.close()` clears `isPending` without cancelling the DELETE** (major) — correct, and worth saying that this one is not this PR's: it moved here verbatim from `settings.tsx`. Fixed rather than filed because the guard is three lines and the file was already open.

Coverage went 72% → 85% on the split files along the way, by covering three things that were genuinely untested rather than by arguing the moved lines were not new: the ladder's won/lost ends, and both pipeline writes.

PR #5479, opened 2026-09-12
The pre-push hook runs the integration lane for a backend-Go push

LarsGradion:
`make check` reaches the integration lane not at all: every integration file carries `//go:build integration` (1341 of them), so the lane's packages compile into nothing and a green check says nothing about them. AGENTS.md already tells a contributor to run `make check-all` for a change touching backend Go. Nothing enforced it.

On 2026-09-12 three parallel sessions merged all day on green local gates, and three defects reached main that no other local gate could see:

- a new rule that broke two pre-existing tests,
- reference rows registered in the test harness's preserved-table list but not the product's, so a reset deleted them mid-shard,
- an index addition that pushed an empty table past the slack deciding DELETE vs TRUNCATE.

Each was found afterwards by running the lane by hand, twice by accident.

## What it does

Runs `make test-integration` when a push changes hand-written Go or SQL under `backend/`. A docs or frontend push pays nothing.

It sits **above** the existing `no changed Go files` early return, for the reason the migration-version gate above it does: that check exits early when a push changes no Go, and a push that adds only a `.sql` migration is exactly a push this lane is for. A migration is the single likeliest thing to redden it.

## It skips rather than blocks when Postgres is down

`make test-integration` fails hard without a database on purpose, because a skipped security gate looks exactly like a passing one. That is right for a deliberate run and wrong for a hook: a push with the stack down is blocked on the stack rather than on the code, and a hook nobody can push past is a hook that gets turned off. The absence is reported at the same volume and the push continues.

## The reason it was kept out had expired

The comment above `check-all` said parallel sessions share one test template, so a hook would have them rebuilding each other's schema mid-run. `scripts/lib-testdb.sh` has derived a template per linked worktree since 2026-08-24 (#2549, `_testdb_worktree_slug`) — two weeks before that reasoning was written down in #4678 on 2026-09-07. Two shells in the SAME worktree still share one, and so does the primary checkout; the comment now says that instead.

## Verification

Exercised live rather than argued from source. I pushed a temporary backend Go commit specifically to reach the lane path: the hook ran 52 packages, 6255 tests, 0 skips, and the template was `margince_test_margince_hook_wt` — the per-worktree isolation observed in a real run. The probe commit was then removed.

Also checked by hand, since Codex was unavailable in this environment: `|| true` survives `set -euo pipefail` and yields an empty string rather than aborting; `make -C "$root"` resolves to the linked worktree and not the primary checkout; the root Makefile does delegate `test-integration` to `backend/Makefile`; and every date and PR number above comes from `git log -S`.

`make check-backend` green on the rebased branch.

---
## Summary by cubic
Runs the integration lane from the pre-push hook whenever a push changes hand-written backend Go or SQL. `make check` previously compiled every integration test into nothing, so a green check said nothing about them — a gap that let three defects reach main on 2026-09-12.

- The hook probes Postgres with `pg_isready` (not just a TCP port) and skips loudly rather than blocks when it isn't ready, so a stopped stack can't trap a push; a missing `pg_isready` takes the same skip path, since readiness can't be told.
- Docs and frontend pushes don't pay for the lane.
- The old reason for keeping the lane out expired: `scripts/lib-testdb.sh` now derives a test template per linked worktree, so parallel sessions no longer rebuild each other's schema.

<sup>Written for commit fee3f63c2508c1d50d82d18e67fc9d16a466c0a6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/5479?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Chores**
  - Pushes involving backend Go or SQL changes now automatically run integration checks when PostgreSQL is available locally.
  - Integration checks are skipped when the required database readiness tool is unavailable or PostgreSQL is not ready.
  - Failed integration checks block the push.

- **Documentation**
  - Updated developer guidance to clarify the integration lane, its triggers, and when checks are skipped.

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
-# Diff-scoped code-craftsmanship gate (ADR-0045, `craft static`).
+# The pre-push gates: diff-scoped craftsmanship (ADR-0045, `craft static`), the
+# sub-second whole-tree greps, and — for a push that changes backend Go — the
+# integration lane. Everything but the lane answers in seconds; the lane is the
+# one deliberate exception and the tail of this file says what buys it.
 #
 # Runs the deterministic craftsmanship linter on the Go files that THIS push
 # changes vs origin/main — backend/, extensions/, fixtures/ and desktop/ alike,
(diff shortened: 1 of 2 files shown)


LarsGradion:
Both findings fixed in fee3f63c2, and both were right.

**The readiness probe is the better catch.** `nc -z` accepts any TCP listener, so a Postgres mid-startup or an unrelated service on 15432 would have sent the hook down the *run* path, where the lane's first act is building a template. A failed push, blamed on the code. Now `pg_isready`.

One thing the finding did not raise, which the fix has to handle: `pg_isready` ships with libpq rather than with the OS, so a contributor may not have it. Its absence now takes the same skip path under its own message. A missing probe means "cannot tell", never "block". The two skips say different things on purpose — telling somebody to run `make db-up` when their database is already up sends them looking in the wrong place.

**The SQL comment finding is mine.** I widened the predicate to `\.(go|sql)$` after writing both headers and never went back. Both now say Go or SQL, and the Makefile comment adds why: a migration is the likeliest thing of all to redden this lane.

Verified all three probe branches: ready runs, not-ready skips, absent-tool skips. `make check-backend` green.

PR #5508, opened 2026-09-12
fix(privacy): waive the swallowed rollback that turned main red

dungbanguyen:
**main is red** and every open PR fails with it. No open PR was fixing it when I checked, so I took it.

`make craft-static` (whole-tree) blocks on a bare `_ = tx.Rollback(ctx)` in `sarrightscases_integration_test.go`. The pre-push hook and the craftsmanship job both scope to the diff against `origin/main`, so the finding could not appear until the file *was* main — which is why a green PR became a red trunk.

Waived rather than rewritten, in the same shape and voice as the two waivers this tree already ratified for it (`approvals/lockorder_integration_test.go:168`, `approvals/bundle_integration_test.go:455`): the rollback abandons a transaction the `t.Fatalf` on the very next line is already failing the test on, so its own error cannot change the outcome.

One line. `make craft-static` passes clean.

### DIFF
--- a/backend/internal/modules/privacy/sarrightscases_integration_test.go
+++ b/backend/internal/modules/privacy/sarrightscases_integration_test.go
@@ -214,6 +214,7 @@ func eraseCapabilities(
 		t.Fatalf("opening the erasure: %v", err)
 	}
 	if err := deleteConsentCapabilities(ctx, tx, contact, nil, "test"); err != nil {
+		//craft:ignore swallowed-errors the rollback abandons a transaction the Fatalf below is already failing the test on; its own error cannot change that outcome
 		_ = tx.Rollback(ctx)
 		t.Fatalf("erasing the subject's capabilities: %v", err)
 	}


tinnhlt-gra:
You got there first — closing #5509, which fixed the same line, in favour of this. Posting what its review round turned up rather than letting it die with the branch.

**The package already has a shape for this, at four sites**, and all four differ from the waiver in one way that matters: `transcriptinterlock_integration_test.go:48`, `anonymizecontact_integration_test.go:65`, `subjecttextcolumns_integration_test.go:222` and `transcriptinterlock_integration_test.go:128` each use `t.Cleanup` (or `defer`) immediately after `Begin`, tolerating `pgx.ErrTxClosed`:

```go
t.Cleanup(func() {
    if err := tx.Rollback(context.Background()); err != nil && !errors.Is(err, pgx.ErrTxClosed) {
        t.Errorf("releasing the erasure transaction: %v", err)
    }
})
```

Two things that buys, neither urgent:

1. **It covers the commit path too.** `eraseCapabilities` has a second `t.Fatalf` at the `tx.Commit` failure — no rollback there either, which the inline shape cannot reach.
2. **`ErrTxClosed` is the common case here, not the edge.** Whatever made `deleteConsentCapabilities` fail may well have closed the transaction on its way out. My first attempt handled the rollback error inline instead of waiving it, and the review caught that this manufactures a second failure printed *ahead* of the real cause, since `t.Errorf` precedes `t.Fatalf`.

Your waiver is legitimate and it unblocked the repo, which was the urgent part. This is just the shape the other four sites landed on, if anyone touches this helper again.

Separately, and worth more than either fix: the helper's doc comment claims it runs the cascade "the shape it runs in production", and it does not — production nests it in `EraseContact` with the subject held on the `margince_app` role. Filed as #5510 with the specifics.

PR #5582, opened 2026-09-14
fix(deal360): the brief save is pinned to the reading it opened on

### LINKED ISSUE #5566: Deal brief editor silently overwrites a colleague's newer brief once the open page has re-read the deal
## Summary

The deal brief editor is meant to refuse a save made on a stale reading. It refuses correctly while the page has not re-read the deal. But the open record page re-reads itself every 60 seconds, and after that the editor pins its save to the **colleague's** version. The server then accepts the save (200), and the reader's draft silently replaces the colleague's newer brief. **Neither author is told.**

## Observed

Measured on a dev stack at core `4e5037e49`, 2026-09-13. The reader is a Member seat in the browser; the colleague is a second seat saving through `PATCH /v1/deals/{id}` with `If-Match`.

1. The reader opens a deal (brief = "Original…"), presses **Edit** on the Deal brief panel, and types a draft.
2. The colleague saves a rewrite of the brief with the correct `If-Match`. Version 1 → 2, response 200.
3. **Saving straight away is refused.** The editor shows an alert and the stored brief stays the colleague's. ✅
4. **Saving after the page's own 60s re-read is not refused.** The page's `GET /v1/deals/{id}` returns version 2. The reader's draft is still in the textbox. Pressing **Save brief** sends `PATCH /v1/deals/{id}` with `If-Match: 2`, and the server answers **200**. The stored brief is now the reader's draft, written against "Original…", and the colleague's rewrite is gone. ❌

Reproduced on 2 consecutive runs; the request header and status were captured from the browser, not inferred.

## Why

`frontend/src/screens/deal360/dealbriefedit.tsx` seeds the textarea on the opening edge only, which is correct. But it pins the save with the `version` **prop**, and `dealbrief.tsx` passes `deal.version` from the live deal query. `app/queryclient.ts` (`LIVE_RECORD_MS = 60_000`) refetches that query while the page is open. So the draft stays pinned to what the reader opened on, while the version moves on to whatever the latest re-read says.

The file's own comment states the intent: *"The version the panel READ. It pins the save, so an edit that landed while this modal sat open comes back as a conflict rather than silently overwriting somebody's words."*

The sibling editor `deal360/dealcommercialedit.tsx` already does this right. It captures `openedVersion` when it opens and pins the save to it (*"Pinned to the reading the FORM holds, not the live one"*). The brief editor is missing that same step.

## Cost

Two colleagues rewriting a deal brief is exactly the case this pin was built for. Anyone who keeps the editor open for more than a minute, which is normal for prose, overwrites a newer edit with no warning. The only trace left is the deal's history.

## Expected

The save is pinned to the version the editor opened on, so step 4 is refused the same way step 3 is.

---
Found by [margince-qc](https://github.com/margince/margince-qc), building `UAT-BRIEFCLASH-1` (`deal-brief-edit-refuses-a-stale-save`), which proves the editor's stale-save refusal; the second test holds the editor open across the page's own re-read and caught the overwrite.

dungbanguyen:
Closes #5566.

The deal brief editor refuses a stale save correctly — **for the first sixty
seconds**. After that it stops, and starts overwriting.

## What was happening

The editor already captures the BRIEF on the opening edge, which is why a
background refetch cannot wipe a draft mid-sentence; there is a test for that.
What it did not capture was the **version**. The save took the `version` prop,
which is the live deal's, and `app/queryclient.ts` refetches that query every
`LIVE_RECORD_MS` (60s) while the modal sits open.

So the draft stayed anchored to the words the reader opened on while `If-Match`
drifted to whatever the newest reading said. The server agreed with a version
the form had never seen, answered **200**, and the colleague's rewrite was
replaced by a draft written against the text before it. Neither author is told;
the only trace is the deal's history.

Sixty seconds is nothing for prose. Two colleagues rewriting a brief is the exact
case this pin was built for, and the pin held only until the page re-read itself.

## The fix

`openedVersion`, captured on the opening edge and read through a ref exactly the
way the brief text already is, so a refetch cannot move either. The sibling
editor `dealcommercialedit.tsx` already does this and says why — *"Pinned to the
reading the FORM holds, not the live one"* — so this is the same step in the file
that was missing it, not a new idea.

The prop's own comment claimed the pin it was not delivering. It now says what it
is (the live version) and points at what actually pins the save.

## The rest of the tree, checked rather than assumed

The shared `EditAction` captures the row the form opened on and hands it to every
caller as `opened`, so form-based editors were never exposed to this. The only
hand-rolled modals that pin a version are the two in `deal360`, and the other was
already right — which is what made this one findable.

## Test

`pins the save to the version it opened on, not the one that arrived since`
drives the reported sequence: open, type, rerender at version 3 with a
colleague's words (the page's own re-read), save. It asserts `If-Match: 2`.

Mutation-verified — restoring the live `version` gives
`AssertionError: expected '3' to be '2'`, which is the defect exactly.

## Summary by CodeRabbit

- **Bug Fixes**
  - Saving an open deal brief now uses the deal version from when the editor was opened, preventing background updates from changing the save’s version check.
  - Added coverage to verify that edits are saved against the correct version when deal data refreshes while the editor is open.


dungbanguyen:
Closing unmerged: #5587 ("unify editing in Details panes") landed while this was open and deleted `dealbriefedit.tsx` entirely. The brief is now edited through the shared Details form, so the modal this patched no longer exists.

I checked that the defect went with it rather than assuming. The new path is `saveRecordEdit` → `saveIndependentEdit`, and it is a stronger model than the pin this PR added:

- The first write pins `requireVersion(current.version)` where `current` is the row the FORM opened on — so the stale save #5566 reported is refused, which is what this PR was for.
- On a version-skew refusal it re-reads and retries, but only when the fields THIS reader touched are unchanged in the fresh row. Two colleagues rewriting the brief means `description` differs, and it throws rather than retrying.
- A colleague editing a DIFFERENT field lets the retry proceed onto the fresh version, so independent edits no longer clash at all — which plain pinning could not do.

Its own comment states the property: *"another write between the fresh read and our PATCH must fail again rather than silently losing that newer change."*

So this branch is superseded rather than wrong. #5566 closed separately with the evidence.

PR #5591, opened 2026-09-14
Page the notice-case queue so its tail is reachable (#5511)

### LINKED ISSUE #5511: A privacy officer cannot see most of their own disclosure duties past the 200th-soonest one
# A privacy officer cannot see most of their own disclosure duties past the 200th-soonest one

**Upstream:** [margince#TBD](https://github.com/margince/margince/issues/TBD)

## Observed

Measured live on core `0f59fbaf7` (2026-09-13), against the running installation, not from source alone.

`GET /privacy/notice-cases` (`listNoticeCases`) is the whole queue a privacy officer works: every Art. 13/14 disclosure duty the installation still owes, "soonest deadline first." Its contract (`backend/api/crm.yaml`) declares exactly two request parameters — `state` (repeatable) and `limit` (integer, max 200, default 50) — and its response schema is `{ data: NoticeCase[] }`, with no `next_cursor`, no `PageInfo`, and no `contact_id` (or any other row-identifying) filter anywhere on the route.

This installation currently carries 1,285 rows in `privacy_notice_case` with `state = 'open'` (queried directly against the database this build runs against). `limit=200` is the largest page the contract allows. A case ranked past the 200th-soonest `due_at` is therefore **unreachable through this route at all** — not slow to reach, not reachable on page 2, structurally absent from every response the route can ever return, for every caller, regardless of how patient they are.

The Settings → Privacy screen (`frontend/src/screens/noticecases.tsx`, `NoticeCasesCard`) calls this same route with the same ceiling and no additional filter, so the gap is not merely a wire detail — the screen a privacy officer actually looks at has the identical blind spot.

Confirmed live: `GET /privacy/notice-cases/{id}` reads one case directly, with no row-scope narrowing beyond the same `privacy_request` grant the list needs — so the product itself has no supported way to reach a specific out-of-window case at all without already knowing its id from some channel outside this feature (in our case, reading the database directly, which a real privacy officer cannot do).

## Cost

A privacy officer working this queue in good faith, pressing through however many pages the screen offers, will believe they have seen every duty the installation owes once the screen stops offering more — the screen gives no signal that a longer tail exists past what `limit` returned. On an installation with more than 200 concurrently-open duties (this one qualifies today), the majority of real, legally-owed disclosure duties are invisible to the one role the product gates this compliance record to, with no error, no count-mismatch, and no indication anything was left out. This is the under-recognition shape this repository exists to catch: not a wrong answer, but a confident, complete-looking one that silently omits most of the truth.

## Workaround

None available to a person using the product as intended. The only route around it we found is reading `privacy_notice_case` directly against the database for a specific contact's own case id, then calling `GET /privacy/notice-cases/{id}` — not something a privacy officer without database access can do, and not a workaround this repository's own scenario (`scenarios/notice-cases-workflow`, UAT-NOTICECASE-1) relies on for anything beyond arranging its own fixture.

---
Found by [margince-qc](https://github.com/margince/margince-qc), building `UAT-NOTICECASE-1` (`notice-cases-workflow`), while confirming a fresh fixture case would be reachable through the queue's own list read.

navdeepplaya-alt:
Closes #5511.

## The problem

`GET /privacy/notice-cases` (`listNoticeCases`) is a privacy officer's whole queue of Art. 13/14 disclosure duties, ordered soonest-deadline-first. Its contract declared only `state` + `limit` (max 200) and returned a bare `{ data }` — **no cursor, no `next_cursor`, no row filter**. So on any installation with more than 200 concurrently-open duties, a case ranked past the 200th-soonest `due_at` was **structurally unreachable** — absent from every response the route could return, for every caller, with no error and no signal a tail existed. The Settings → Privacy screen fetched with the same ceiling and had the identical blind spot.

## The fix

Keyset pagination on `(due_at, id)`, mirroring the sibling `listDataSubjectRequests` in the same module:

- **Contract** — declares the shared `Cursor` param and returns `{ data, page: PageInfo }`.
- **Store** — walks an opaque `(due_at, id)` cursor (`storekit.EncodeOpaque`/`DecodeOpaque`), fetches `limit+1` so `has_more` is a fact, and mints `next_cursor` from the last row served. The cursor carries the whole ORDER BY key so a page boundary lands in the same place even when two duties share a `due_at`; a malformed token is the caller's `422`, not a `500`.
- **Screen** — the officer's queue now follows the cursor through the existing `LoadMoreButton` / `useInfiniteQuery` pattern (the same one the audit log and webhooks screens use) instead of stopping at the first read. No new design-system component.

Along the way, folded the module's three `"page"` string literals into a `keyPage` const beside the existing `keyData`, so every list response in the module answers under one spelling.

## Tests

- **Integration** (real Postgres): `TestTheQueueTailIsReachableByWalkingTheCursor` proves the tail past a page is reachable by walking `next_cursor` and never repeated; `TestTheQueueRefusesAMalformedCursor` proves a bad token is a 422.
- **Frontend**: a new case walks past the first page via "Load more" and asserts the second read continues from `cursor=…`.

## Scope

Deliberately scoped to #5511 only. A sibling in the same module — `listConfirmSubmissions` — shares the same `limit`-with-no-cursor shape (and a contract comment that rationalises the gap); it and a census-style fitness function over the contract are worth a follow-up but are left out of this change.

---
## Summary by cubic
Pages the notice-case queue with keyset pagination on `(due_at, id)` so every owed disclosure duty is reachable, not just the first 200. Closes #5511.

**Bug Fixes**
- `GET /privacy/notice-cases` now accepts the shared `Cursor` param and returns `{ data, page: PageInfo }` instead of bare `{ data }`.
- The store walks the opaque `(due_at, id)` cursor, fetches `limit+1` so `has_more` is a fact, and mints `next_cursor` from the last row served.
- The cursor carries the whole ORDER BY key so a page boundary lands correctly when two duties share a `due_at`; a malformed token is the caller's `422`.
- The Settings → Privacy screen follows the cursor via the existing `LoadMoreButton` / `useInfiniteQuery` pattern.
- Folded the module's three `"page"` string literals into a `keyPage` const beside `keyData`.

<sup>Written for commit 20da76e4f1169088abd91806e393b5ff355aa913. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/5591?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


navdeepplaya-alt:
@coderabbitai review

navdeepplaya-alt:
Superseded by #5635, which merged the same fix for #5511 — keyset pagination on `(due_at, id)` for `GET /privacy/notice-cases`, the `limit+1`/`has_more` store walk, the 422 on a malformed token, and the screen's Load-more — with broader integration coverage. #5635 also deliberately minted a route-specific cursor rather than the shared `Cursor` param this branch used (the shared token fingerprints the request's `sort`, which this route doesn't have), so nothing here remains to land. Closing rather than resolving the conflicts.

PR #5594, opened 2026-09-14
Review coverage runs only for this repository's branches; a draft sheds its frozen rate

ancorn-gradion:
## What

Two invariants that the tree stated but did not hold, one per commit.

1. **`review-coverage.yml` runs a pull request's own code only for branches of this repository.** The job checks out the pull request's head and runs a repository script out of that checkout, and the workflow can be started by `pull_request_review` as well as `pull_request`. GitHub downgrades a fork's token on the second but not on the first, so the job is now confined to same-repository branches on any event but `pull_request`. A new gate, `backend/gates/forkheadcheckout_test.go`, derives that obligation from the workflow tree rather than fixing it in one file.

2. **A contract returned to draft sheds the conversion it froze.** Activation freezes a contract's `fx_rate_to_base`/`fx_rate_date` once and never re-reads it; draft is the one status whose currency is still the human's to correct. Each rule is right alone, and together they allowed a row to sit in draft carrying a rate frozen for a currency it was then free to leave — while every base-currency rollup multiplies the figure by that rate.

Closes: none

## Why

The invariant behind the first is that code from a pull request's own branch never runs with a token GitHub has not downgraded. `permissions:` cannot express it, because on a privileged trigger GitHub hands the job the base repository's capabilities regardless; and `persist-credentials: false` does not reach it, since that keeps the token out of the git config rather than out of `env:`. The gate is a fitness function, not a point fix: a job that checks out a pull request's head in a workflow any privileged trigger can start must carry the same-repository comparison, so the next workflow is covered the day it is committed.

The invariant behind the second is that a re-priceable contract carries no frozen conversion. It is now held from both sides — `applyStatusTx` clears the pair on the way into draft, and `refuseRepricingAFrozenContract` keys on the frozen rate as well as the status, which is what answers for a row that froze one before this held. Holding it in one place only would have left the other half of the state space open.

## Behaviour changes

- A contract moved from `active` back to `draft` loses `fx_rate_to_base` and `fx_rate_date`, recorded in its audit diff like any other field change. Both columns move together, as `contract_fx_pair` requires, and only when there is one to clear — so an ordinary draft assertion does not name them. Re-activating freezes afresh for the currency the row holds then, which is the point: it was returned to a state that permits re-pricing.
- A draft that already carries a frozen rate now refuses a currency move with the same field-level refusal (`currency`, `contract_terms_contradict`) an active contract gives.
- A fork's pull request keeps the review-coverage report on every push; what it loses is the refresh on a submitted review.

No existing test asserted the old behaviour in either case, so no assertion needed rewriting.

## How verified

Every claim below is a run, not an expectation.

- **Both fixes falsified before being trusted.** With the workflow guard removed, the new gate fails and names the job and the exact comparison to add. With the two contract changes reverted, the unit test fails (`a draft carrying a rate` gets `<nil>` where a refusal is due) and the integration test fails on the contract still holding its frozen rate. Restored, all pass.
- **Integration lane**, against real Postgres: every contract test in `internal/compose/integration` passes — 20 of 20, comprising the six in `contract_fx_freeze_integration_test.go` (including the new one) and the fourteen in `contract_lifecycle_integration_test.go` and `contractreferences_integration_test.go`. The lifecycle file matters most here, since it is where a test asserting the old transition behaviour would have lived.
- **Unit lane**: `./internal/modules/contracts/...` green.
- **Gates**: the new gate, plus the gate-registration census, the parallel census and the inventory-currency gate. `docs/reference/gate-inventory.md` was regenerated with `-update-gate-inventory`, and the diff is exactly the new row and the section count.
- **`craft static --strict`** on all five changed Go files: 0 blocker, 0 major, 0 minor. It caught one real blocker on the way — my added prose pushed `contract_write.go` past the 500-line ceiling, so the comment was tightened back to the file's original length rather than spending the two lines of headroom it had.
- **golangci-lint**, both configs: strict reports 0 issues; the baseline reports only the pre-existing `gofmt` noise this CRLF checkout produces on untouched files too (`contract_list.go`, which this PR does not touch, fails identically).
- **Whole module**: `go build ./...`, `go vet ./...` and `go vet -tags integration ./...` all exit 0, so nothing outside the changed package broke and every integration file still compiles.
- **Root script gates**: `check-rls-store-path`, `check-no-jurisdiction`, `check-image-pins`, `check-host-ports` and `check-migration-versions` all pass.
- **Gates on a clean tree**: run in a detached worktree of these two commits, so the result is what CI sees rather than what my working copy contains. Every failure there is already in the pre-change baseline — the set of failures introduced by this change is empty. Notably `TestEverySourcePathNamedInProseExists` and `TestPublicTreeCitesNothingPrivate` pass there, which is what holds the new citation in `ci-workflows.md` honest.

A note for anyone reproducing this on Windows: three unit packages (`webhooks`, `blobstore`, `dbmigrate`) fail in this checkout and none of it is this change. `webhooks` is 50 golden-file snapshots whose `got` and `want` print identically, `dbmigrate` rehashes migrations that are 100% CRLF, and `blobstore` loses a `TempDir` cleanup to Windows file locking. All three fail identically on a pristine worktree containing none of my edits, and this PR touches no file in any of them.

- [ ] `make check` is green — **not run in full**, and I would rather say so than tick it. What I ran is listed above and covers the backend half's substance; the frontend half cannot fail on this diff, since no `frontend/` path is touched.
- [x] `make test-integration` is green for every contract test in the affected package — the lane that exercises the write shape here. The whole package in one run is not something this machine can report on: it exceeds Go's 600s package timeout during whole-app setup in an unrelated scheduled-send test, so I ran the contract tests by name instead of claiming a result I did not get.
- [x] This diff and description carry no secrets, no customer data, no local machine paths, and nothing quoted from or pointing at a private document. Checked by scanning the committed diff, not by recollection.

## AI involvement

Written by Claude Code (Opus 5) end to end — the workflow condition, the gate, the two contract changes, the tests, and this description — under my accountability. The two judgement calls worth naming: making the CI fix a tree-wide gate rather than a one-line edit, and holding the contract invariant from both sides instead of only clearing on transition, so that a row already in the bad state is refused rather than silently re-priced.

## Accountability

By opening this PR I confirm I am **accountable** for this change and can **explain every line** in it — human-written or AI-assisted. See [CONTRIBUTING.md](/CONTRIBUTING.md) and the [Code of conduct](/CODE_OF_CONDUCT.md).

https://claude.ai/code/session_01KqEF4mbX4owQwtGBKnDiPw

---
## Summary by cubic
Fixes two invariants the codebase stated but did not hold: review coverage could run a fork's branch code with an undowngraded token, and a contract returned to draft kept a conversion rate frozen for a currency it could then leave.

**Review coverage restricted to same-repository branches**
- `review-coverage.yml` checks out the pull request's head and runs a script from that checkout; on `pull_request_review` GitHub does not downgrade a fork's token, so the job now requires `github.event.pull_request.head.repo.full_name == github.repository` on every event but `pull_request`, where the downgrade is the control.
- The new gate `backend/gates/forkheadcheckout_test.go` enforces this comparison across the workflow tree instead of fixing one file, so future workflows are covered on commit. It also handles every legal spelling of `on:` (scalar, sequence, mapping) so a trigger in any shape is caught.
- A fork keeps the coverage report on every push; it loses the refresh on a submitted review.

**Draft contracts shed their frozen conversion**
- Activation freezes `fx_rate_to_base`/`fx_rate_date` once, while draft's currency remains editable, so a contract moved back to draft carried a rate frozen for a currency it could then stop holding — and every base-currency rollup multiplied the figure by that stale rate.
- `applyStatusTx` now clears both columns on the way into draft, only when a rate exists; `refuseRepricingAFrozenContract` now refuses a currency move on any contract carrying a frozen rate, including drafts that froze one before this change.
- Re-activating a draft freezes afresh for the currency the row holds then.

<sup>Written for commit 2beff006d1ab0f05f854e98ffc726d9d50a5df0e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/5594?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Returning an activated contract to draft now clears its frozen exchange rate and date.
  - Draft contracts can be repriced and reactivated using the newly selected currency’s current published rate.
  - Currency changes remain restricted for active contracts and drafts that retain a frozen rate.

- **Security**
  - Pull request review checks now prevent untrusted fork code from running with elevated permissions.

- **Documentation**
  - Updated CI workflow guidance and security gate references.


ancorn-gradion:
Fixed in 2beff006d — thank you, this was a real hole, and the failure mode is worse than a missed trigger.

For the scalar and sequence forms `yaml.Unmarshal` answers a `*yaml.TypeError`, so `readTriggeredWorkflow` calls `t.Fatalf` and the walk stops on that file having judged none of the workflows after it. A gate that cannot read a legal workflow is one that has quietly stopped covering the tree — exactly the failure this gate exists to prevent elsewhere.

`on:` is now decoded as a raw `yaml.Node`, and `triggerNames` reads all three shapes. Two cases are planted rather than only the fix: that each shape decodes, and that a privileged trigger spelled as a scalar or a sequence entry still reaches the *judgement* and not merely the decoder.

Worth recording that this was latent rather than live: every workflow in the tree uses the mapping form today, which is why the gate was green on the run that shipped the defect.

PR #5604, opened 2026-09-14
Split the employment writes out of the contact's employer list

joshbruh-lang:
## What

`contactemployers.tsx` had grown to 502 lines. The writes behind the employment list (create, end, edit and remove an edge, plus the company search the add flow picks from) now stand in `employmentactions.ts`; the screen keeps the list and its rows at 354 lines. Its four importers follow, and the if-match coverage census names the unpinned relationship delete by its new path.

The new module gets a spec of its own, `employmentactions.test.tsx`: the pinned `If-Match` on an edit, the re-read that lends a version only to an unversioned snapshot that still matches, the edge the re-read no longer finds, each write's refusal, and the three readings every verb refreshes.

Closes: none

## Why

`make fe-file-length` caps a screen at 500 lines, ratcheted through `scripts/fe-file-length-waivers.txt`, and the file carries no waiver. #5602 took it over the cap, so `frontend / fe-quality` fails on every pull request whose merge commit carries main's tip. This is the claim for that red (`claim: main-red`), first seen on #5603.

No behaviour changes: the moved functions are byte-identical, only `useEmploymentActions` becomes exported because the screen now imports it. The spec exists because SonarCloud reads a moved file as new code and the list's own tests reached its happy paths and none of its refusals; the module now measures 100% on lines, branches and functions.

## How verified

- `make fe-file-length` green: no frontend source file over the cap.
- `make check-fe` green: 717 test files, 9376 tests, composed typecheck (which is what catches an importer the app typecheck does not see).
- `frontend / fe-quality` green on this PR's first push, where it is red on main.

- [x] `make check` is green (`check-fe`; no Go touched)
- [x] `make test-integration` is green (or: this change does not touch tenant data / RLS / the write shape)
- [x] `make craft-static` reports no new blockers (Go-only gate, no Go touched)
- [x] This diff and description carry no secrets, no customer data, no local machine paths, and nothing quoted from or pointing at a private document — a decision is cited by its number (this repository is public)

## AI involvement

AI-generated with Claude Code: the diagnosis, the split, the spec and this description.

## Accountability

By opening this PR I confirm I am **accountable** for this change and can
**explain every line** in it — human-written or AI-assisted. See
[CONTRIBUTING.md](/CONTRIBUTING.md) and the
[Code of Conduct](/CODE_OF_CONDUCT.md).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01RQdKmyQyW2FCNneEoQiAzT

Issue #5483, opened 2026-09-12
Integration shards are killed mid-run with exit 143 — two runners died 18 seconds apart, so the cause is outside the job


dungbanguyen:
## What happens

An integration shard is killed part-way through, and the lane reports:

```
##[error]The runner has received a shutdown signal. This can happen when the
runner service is stopped, or a manually started runner is canceled.
##[error]Process completed with exit code 143.
```

143 is SIGTERM. The shard produces no test failure and no assertion — it simply stops, the `integration` job fails because one of its shards did, and `ci` fails because `integration` did. **A re-run of the failed jobs passes every time.**

## Observed today, on unrelated branches

Every one of these was a clean re-run afterwards, on the same commit:

| run | shard(s) killed |
|---|---|
| `34684779617` | 1/6 |
| `34685711766` | 1/6 |
| `34693789830` | 1/6 and 5/6 |
| `34693791796` | 1/6 |
| `34696604241` | 6/6 |

Five runs across five different branches in one working day, plus the same signature on two earlier runs I did not record the id of. Nothing links them but the shape.

## What it costs

One full CI cycle per occurrence, and it lands on roughly one pull request in two. It is also the worst kind of red to read: the job says nothing about tests, so the first reaction is to go looking for a defect in the diff, and there is not one.

## What I could not determine, and what would settle it

I have no evidence of the cause and am not going to guess at one in the title.

Two things worth checking, in this order:

1. **Whether the runner is being reclaimed under memory pressure.** An integration shard runs Postgres alongside several `go test` processes; a host that reclaims the runner would look exactly like this, and the shard that dies is not always the same one. The logs I pulled carry no OOM line, but a SIGTERM'd runner would not necessarily write one.
2. **Whether it correlates with shard wall time.** If the long shards are the ones dying, this is a resource question; if it is uniformly distributed, it is the hosted runner fleet and there is nothing here to fix but the retry.

`scripts/test-integration-parallel.sh` and the shard sizing in `ci.yml` are where the first would be answered.

## Not this

- Not #2991 — that is `proxy.golang.org` returning `INTERNAL_ERROR` mid-stream on a module download, with the module named in the log. These runs get past module download and die in the middle of the test phase.
- Not a test flake. No test reports a failure; the process is signalled.

dungbanguyen:
A sixth occurrence, and this one is useful because it is the second on the *same branch* within an hour.

| run | shard | branch |
|---|---|---|
[...]

PR #5608, opened 2026-09-14
feat(dealroom): a buyer reads the room as a document gallery beside their contact

joshbruh-lang:
## What

The buyer's Deal Room — the one page a client ever sees — now reads like the front of a small site rather than a form. The frame widens to 1040px in two columns, the board on the left and the contact as a card on the right (stacked under 900px). Documents are tiles in a gallery: a page-shaped stage with the kind stamped from the filename, the recorded size, and, when a buyer is waiting, an "N unanswered" mark. A PDF or image opens over the room through the shared file preview instead of leaving for the downloads folder. Comments carry the author's monogram, their side and the time written. A closed room says the day it closed, and the foot says who is signed in. The fill and the green side rules are gone from both sides of the board.

Along the way: the file preview learns to read on a session token (the buyer holds no cookie) and saves the drawn bytes rather than following an href the token cannot ride; the board's thread rows and the buyer page's chrome, session and board move into files of their own, which takes the buyer screen under the size cap and off the waiver list. The seller's board gets the same tiles, kind stamp, size and unanswered mark — it is one board, drawn once.

Closes: none

## Why

`DESIGN.md` §7 "The Deal Room, two surfaces on one board" describes the buyer's page as a gallery on a 1040 frame with the contact beside it; what shipped was a single 720px column of rows. This is the part of that description that needs no contract change — everything drawn here is already on the wire. The pieces that need the public contract (seller branding, "new since your last visit", next steps, who is in the room) are deliberately not here.

One invariant worth naming: "unanswered" means an open thread whose last word is the buyer's. It reads the same on both sides — waiting on the seller to the buyer, a buyer waiting on you to the seller — which is why it lives in the shared board and not in either screen.

## How verified

- `make frontend-check` (ds gates, drift, lint, unit, build, file length) green, 714 test files.
- New tests: tile stamps kind and size; unanswered count derived from threads and silent when settled; Read verb only when the side can draw the file; comment carries author, side and time; buyer opens a PDF with the Bearer and a spreadsheet keeps download as its only verb; contact card and signed-in line; closed date; file preview reads with a bearer, saves what it drew, and withholds Download until there are bytes.
- Storybook screenshots of the buyer stories at 1200px and 420px and the seller board story, checked by eye.
- No Go touched, so the integration lane is not exercised by this change.

- [x] `make check` is green (frontend half; backend untouched)
- [x] `make craft-static` reports no new blockers (no Go in the diff)

## AI involvement

Whole change AI-assisted from a mockup agreed with the author: design, code, tests and this description were generated, then read and screenshot-checked before push.

## Accountability

https://claude.ai/code/session_01CuTnB3XU5H8Hv1WXym7ofj

---
## Summary by cubic
Rebuilds the buyer's Deal Room as a 1040px two-column gallery instead of a single 720px column of document rows: documents become tiles, the contact moves into a card beside them, and PDFs or images open in the in-page preview instead of sending the buyer to the downloads folder. The seller's board draws the same tiles, and everything shown already comes from the existing wire, so no contract knowledge is needed.

**Changes**

- A tile shows the kind and size of the file, plus an "N unanswered" badge when the last question on it is the buyer's.
- "Unanswered" means an open thread whose most recent comment is from the buyer; on both sides it's the same count.
- The shared preview now reads the room's session token, and saves the bytes it drew, because a buyer has no cookie a normal link could carry.
- A file the browser can't draw goes back to having just its download verb.
- Comments now show the author's monogram, which side they're on, and the written time; the old fills and green side rules are gone from both screens.
- A closed room shows the day it closed, and the footer names who's signed in.

**Refactors**

- The buyer's chrome, session, and board and the thread rows moved into smaller files, which drops the buyer screen under the file-length cap and removes its line from the waiver list.

<sup>Written for commit 57140eeca57fb3b9263c494679b66c3661273eca. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/5608?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


joshbruh-lang:
**Inherited red, not this PR's.** `deterministic-gates` (and `frontend / fe-quality`, which runs the same `fe-file-length` step) fails on `frontend/src/screens/contactemployers.tsx` at 502 lines. This PR does not touch that file; #5602 took it over the cap on `main`, so every merge commit on `main`'s tip fails the gate. #5604 is the open `claim: main-red` for it and carries the fix (the employment writes split out of the screen).

Per the rulebook a covered red is inherited: not ported here, not re-run (the gate is a deterministic line count on an untouched file, so a re-run answers the same). This PR's own lane, `make frontend-check`, is green locally with the file-length gate included. I will re-check once #5604 lands and bring `main` in if the checks need a fresh head.

---

joshbruh-lang:
**`integration / integration shard (5/6)`: runner loss, not a test.** The job ran its slice for eight minutes and then the runner received a shutdown signal (`Process completed with exit code 143`, "The runner has received a shutdown signal"); no test body reported a result before it was terminated. This PR touches no Go. I will re-run the failed jobs once when the run has completed, which is the one re-run this kind of death earns; a second failure is real and gets root-caused.

---

joshbruh-lang:
**Shards 1 and 5: the runner memory kills of #5483, not a test.** Both shards were killed the same way on the re-run as on the first attempt: about ten minutes into the slice, `exit 143`, "the runner has received a shutdown signal", no test result before the signal. Shards 2, 3, 4 and 6 and the unit-coverage job passed on this head, and #5604's run on the same base passed all six shards minutes earlier, so this is the intermittent host exhaustion #5483 tracks. #5589 carries the fix (sixteen concurrent packages become eight, with a pressure probe that showed 4 kB of swap free at the kill), and I have read it. I am not porting it here: it is a CI-lane change under review on its own PR, and a second copy of it under a UI feature title is what the rulebook's one-fix rule exists to prevent. The one re-run this kind of death earns is spent.

Plan: when #5589 or #5604 lands on `main`, I merge `main` into this branch and push, which runs CI again on a base that carries the fix. Until then the PR stays watched.

---

Issue #5623, opened 2026-09-14
55 more contract-type double casts across the frontend suites, and no gate holding the number down


dungbanguyen:
Found while closing #5534, which covered the six worklist files it named. The class is tree-wide, and the census that made #5534 tractable is the one nobody has run on the rest.

`as unknown as <ContractType>` does not widen the type check, it removes it. A fixture missing a required field compiles, and the case asserts over a shape the server never sends. In #5534 that was hiding an enum value the server does not send (`waiting_customer` for `customer_waiting`), two required fields missing from every row fixture, and six required fields missing from three page fixtures — with all 34 cases passing throughout, which is what makes the class silent.

## The census

57 casts remain outside `#5534`'s scope, once the ones that are not contract shapes are excluded (`View`, `Parameters<typeof …>`, `Window`, `Event`, `Record<…>`, `UseQueryResult` — router and DOM plumbing, a different question). The 20 heaviest files:

```
  7 screens/companydocuments.stories.tsx      2 screens/dealproject.stories.tsx
  4 screens/contractlifecycle.test.tsx        2 screens/contacttransports.test.ts
  4 screens/companyfacts.test.tsx             2 screens/contactreadings.test.tsx
  3 screens/contacttransport.test.tsx         2 screens/contactfiles.stories.tsx
  3 screens/companyapprovals.stories.tsx      1 screens/worklist.hiddenowner.test.tsx
  2 screens/worklist.test.tsx                 1 screens/projectowner.test.tsx
  2 screens/deals.test.tsx                    1 … and eight more with one each
```

By target type: `Attachment` (8), `Approval` (5), `Contact360` (4), `Company` (4), `Deal` (3), `Project` (3), `Partial<Company360>` (3), plus singletons for `Contract`, `WeeklyReview`, `WeeklyPlan`, `ContactBrief`, `CommissionEntry`, `Company360`, `DecisionDeckLabels`.

Stories count. A story is the fixture the UAT lane renders, so a story over a wrong shape renders a screen the server cannot produce.

## What would close it

1. Delete the casts, adding the fields the type actually requires. #5534 is the worked example: every case passed unchanged.
2. Keep the deliberate ones and say so beside them. `{}` as a page payload is a legitimate "not valid, on purpose" — testing that a reader draws nothing rather than throwing — and a cast is the only way to spell it. #5534 left two, each with a reason on the line above.
[...]

PR #5625, opened 2026-09-14
Six worklist fixtures keep the type they claim

### LINKED ISSUE #5534: 21 double casts hide wire-shape drift in six worklist test fixtures
A test fixture that supplies its own version of production proves nothing about production. Six test files build worklist rows and pages with `as unknown as WorklistItem` / `as unknown as Worklist`, which does not widen the type check — it removes it, so a fixture missing a required field compiles and the test asserts over a shape the server never sends.

| File | Casts |
|---|---|
| `worklist.copy.test.ts` | 13 |
| `brief.sentence.test.ts` | 3 |
| `brief.changed.test.ts` | 2 |
| `briefcoverage.test.tsx` | 1 |
| `worklist.linkedcaption.test.tsx` | 1 |
| `worklist.row.compact.test.tsx` | 1 |

**The remedy is demonstrated, not theoretical.** The same defect in `brief.feed.test.tsx` was fixed by adding the two missing required fields (`level`, `consequence`), dropping two properties the type does not have, correcting one enum value, and deleting the cast. Four lines. All 25 cases passed unchanged — they were asserting correct things over a wrong shape.

**Not every cast should go.** `{}` and `{ queue: [] }` as a Worklist are deliberately malformed payloads, testing that a page draws nothing rather than throwing. A cast is the only way to spell "not valid, on purpose", and those deserve a comment rather than removal.

**Where the honest version already lives:** `brief.fixtures.ts` returns `WorklistItem` with no cast at all — `meetingRow`, `leadRow` and their siblings carry every required field. The six files above predate that shared module and kept their own builders. Several could stop building rows entirely and import from it.

**Why it matters beyond tidiness.** A row that omits `consequence` cannot exercise the copy that reads it, and a suite full of such rows reports green over code paths it never reaches. Found after a parallel session hit the same class twice in one afternoon — once as a widened parameter default, once as a cast over a missing `consequence`.

dungbanguyen:
Closes #5534

`as unknown as WorklistItem` does not widen the type check, it removes it. A fixture missing a required field compiles, and the case asserts over a shape the server never sends.

Fifteen of the seventeen casts are gone and **every case passes unchanged** — they were asserting correct things over a wrong shape, which is exactly what makes this class silent.

## What the casts were hiding

| | |
|---|---|
| `source: "waiting_customer"` | Not a value the enum holds. The server sends `customer_waiting`. Two files. |
| `level`, `consequence` | Both required, missing from every row fixture. `consequence` is what the row's copy reads — a fixture without it cannot exercise the sentence that names it. |
| `scope`, `scope_options`, `counts`, `reach`, `sources_unavailable`, `summary` | Required on the page; missing from three `Worklist` fixtures. |

## The two that stay

`{}` as a `Worklist` is a deliberately malformed payload, testing that the function answers nothing rather than throwing, and a cast is the only way to spell "not valid, on purpose". Each now says so on the line above, so it does not read like the fifteen that were not on purpose.

## Not in scope

I checked the other optional fields before touching them: `dispositions` and `overdue` **are** on `WorklistItem`, so they stay where the fixtures had them. The diff is the cast, not a fixture rewrite.

The class is tree-wide — 57 more contract-type casts across the other suites, stories included. Census and the gate that would hold the number down are in #5623; folding them in here would have been roughly four times this diff and touched suites with nothing to do with the worklist.

## Verification

81 files / 671 cases green across `brief*` and `worklist*`. `make check-fe` green.


dungbanguyen:
Pushed a second commit, and the reason is worth recording because it is the same defect eating its own fix.

`make check-fe` — the lane AGENTS.md names as the frontend inner loop — **does not typecheck test files**. `make fe-bundle` does, through `tsc -b`, and that is the job that went red here. So the first commit passed everything I ran and was still wrong, in precisely the way this issue is about: a fixture that compiles and does not match the wire.

Eight more shapes the casts had been covering:

- A `relationship_decay` row is filed under `system` at the routine level (6) and says `data_drifts` — `classifyDecay` in `compose/attention/classifydecay.go`. The fixture claimed `deals_at_risk`, level 4, and `relationship_cools`, **a consequence the contract has never held**. Three invented values in one fixture, all compiling.
- A meeting row carries `consequence`, required; `none` is its real value.
- `summary` carries `due` and `lower_priority` beside `total` and `urgent`.
- `readings` is required and whole — a payload holding `changed_since_brief` alone is not a shape the server can send. Two fixtures had it partial, one omitted it.
- Two `for (const … of [...])` loops widened their own literals to `string`, so `subject.type` and `move.action` stopped being checked against their unions at the call site.

The deliberate casts are three now, each named where it is built (`NO_READINGS`, `NOT_A_PAGE`) rather than justified again at every call.

Worth noting for #5623: the gate that issue asks for should run in a lane that typechecks tests, or it will pass for the same reason my first commit did.

PR #5635, opened 2026-09-14
The disclosure-duty queue has a page two

### LINKED ISSUE #5511: A privacy officer cannot see most of their own disclosure duties past the 200th-soonest one

dungbanguyen:
Closes #5511

`GET /privacy/notice-cases` declared `state` and `limit` and nothing else: no cursor, no page info, no row-identifying filter. `limit` caps at 200, so on an installation carrying more open duties than that — **1,285** on the one this was measured against — a case ranked past the 200th-soonest deadline was absent from every answer the route could give. Not slow to reach, not on page two: structurally unreachable, for every caller, however patient.

And the screen has the identical blind spot, which is the half that matters. Pressing on until it stops offering more is exactly how somebody concludes they have seen every duty the installation owes. No error, no count mismatch, nothing on screen saying a tail exists — a confident, complete-looking answer that omits most of the truth.

## The key is `(due_at, id)`, and both halves are load-bearing

The subject-request queue beside this one pages on a bare id, and can, because it orders by id alone. Here an id is not a position. Two duties can fall due in the same second — ordinary once a nightly import files a batch — and a walk resumed on id skips every case whose deadline is later and whose id happens to be smaller.

The cursor is this route's own rather than `components/parameters/Cursor`: the shared token fingerprints the request's `sort` and is rejectable as `cursor_param_mismatch`, and this queue has no sort dial for either to mean anything against. Same reasoning `listOverlayUserMap` already carries.

A token the queue did not mint is **refused**, not answered with an empty page. Telling a privacy officer there is nothing left to do is the one answer this route must never guess at.

The screen walks it with `LoadMoreButton`, the tree's one spelling of that control.

## Tests

Three integration cases against a real Postgres, plus one on the screen. Every one mutation-verified:

| Mutation | Caught by |
|---|---|
| Drop the continuation (the original defect) | `TestEveryDutyIsReachableHoweverManyThereAre` |
| Keyset on `id` alone | the same case — it saw 6 duties where 5 exist, the id-only walk re-serving a row |
| Remove `LoadMoreButton` from the screen | `reaches a duty the first page did not carry` |

**One of my own tests failed its mutation check and I rewrote it.** `TestTwoDutiesFallingDueTogetherAreBothReached` originally asserted *presence* — `slices.Contains` for each seeded id — and **passed** under the id-only cursor, because that walk can still reach every row while serving one twice and another out of order. It asserts the exact sequence now, and fails naming the duty that goes missing. A test that survives the mutation it exists to catch is the shape this issue is about.

Four existing screen fixtures gain the `page` envelope the contract now requires, spelled once as a `queuePage` helper; a payload without one is not a shape the server can send.

`make check` green (after two lint findings on the first pass: the third `"page"` literal is a named constant beside `keyData` now, and the walker takes its context first).

## Summary by CodeRabbit

- **New Features**
  - Notice case queues now support pagination, allowing all duties to be reached across multiple pages.
  - Added a “Load more” option to retrieve additional notice cases while preserving existing results.
  - Queue responses now include pagination details, including whether more results are available and the next-page cursor.
  - Added cursor-based navigation for requesting subsequent pages.

- **Bug Fixes**
  - Fixed an issue where notice cases beyond the result limit could not be reached.
  - Invalid or mismatched pagination cursors are now rejected.


dungbanguyen:
Fixed. The premise checked out: `storekit.Cursor` spells its id with the same `id` key, so a created_at token from any other paged route decoded cleanly into `noticeCaseCursor` and left the deadline at its zero value — the keyset then paged from the year zero and handed the officer the whole queue again as their next page.

Rather than add a route discriminator to one cursor, the check now validates both halves of the keyset, which is what `decodeLeadQueueCursor` already does for its own. The deadline also gets a field name of its own; `d` is the shared envelope's key for a different type. `TestACursorTheQueueDidNotMintIsRefused` grew a case built from a real `storekit.EncodeCursor` token, and it fails against the id-only check.

Issue #4908, opened 2026-09-08
The assurance sweep's task-state read is time-of-check-to-time-of-use


LarsGradion:
Found by a Codex review of the webinar-defects branch.

`adoptOpenTask` in `backend/internal/compose/assurancebundle.go` reads each
candidate task with `GetActivity` and then adopts or archives it in a later
statement. Between the two, a rep can complete or archive that task.

Two consequences, both narrow:

- A task read as open and completed a moment later still gets this night's
  findings filed under it, so the work lands on a task nobody will look at
  again.
- An extra task completed after the read is archived anyway, which hides a
  completion rather than losing work.

Neither corrupts data and neither is reachable without a rep acting inside the
sweep's own window, which is why it was deferred rather than fixed in that
branch. The fix is to make the state read and the write one statement — an
`UPDATE ... WHERE is_done = false AND archived_at IS NULL` that reports whether
it matched — rather than a read followed by an unconditional act.

The relink half of the same shape WAS fixed there: a task moved to another deal
is no longer adopted or archived, held by
`TestATaskRelinkedToAnotherDealIsNeitherAdoptedNorArchived`.

PR #5648, opened 2026-09-14
The sweep settles a duplicate task on the version it read

dungbanguyen:
Closes part of #4908 — see **Not fixed here** below.

`adoptOpenTask` reads each candidate with `GetActivity` and archives the extras in a later statement. Between the two, a rep can complete the task, and an unpinned archive buried their answer without either of them seeing it.

The work is not lost, which is why the issue calls it narrow. But a **hidden completion is worse than a lost one**: the rep answered, and the record says they did not.

## The fix

The read carries the version it read at, and the archive is pinned to it. A task somebody wrote to in between refuses, and the sweep leaves it alone — the same judgement this function already makes about a task that was done when it looked. The next cycle reads it afresh.

No new door: `ArchiveActivity` has taken an `ifVersion` all along, and it is the concurrency primitive every other writer in this tree uses. The issue suggested a conditional `UPDATE`; the pin is that, through the mechanism already there.

## The test at the door earned its keep immediately

Nothing proved what the pin did, so the case lives where the primitive lives rather than beside the caller: a stale version refuses, and — the half worth asserting — the row keeps **both** its completion and its live state afterwards. A refusal is only worth anything if it left the row alone.

Writing it caught a real error in this change. The door answers **`ErrVersionSkew`**, not `ErrConflict`, which is what I had matched on:

```
archiving on the stale version answered version skew, want ErrConflict
```

Unproven, the sweep would have aborted the whole pass on a lost race instead of skipping one task — a worse defect than the one being fixed, introduced by the fix.

## Not fixed here

The issue's other half: a task read as open and completed a moment later still has this night's findings filed under it.

Closing that means `BundleException`'s insert carrying the task's openness in the same statement its `open_cycle` CTE already carries — the pattern is right there and the SQL is small. What stops me is the caller: that would make `applied=false` mean either "already filed" (today) or "the task closed under us", and the two want opposite responses. Whether a lost race mints a fresh task or defers the finding to the next cycle decides what a rep sees on their list tomorrow, which is a product call rather than a mechanical one.

Left alone rather than guessed at, and the issue stays open for it.

`make check` green.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented completed tasks from being archived when their data changes during processing.
  - Added safeguards to detect version conflicts and leave affected tasks unchanged.

- **Tests**
  - Added integration coverage confirming stale-version conflicts are reported and tasks remain unarchived.


dungbanguyen:
SonarCloud is right that the new branch is uncovered, and I want to be straight about what happened when I tried to cover it.

The uncovered line is the race arm — `errors.Is(err, apperrors.ErrVersionSkew) { continue }`. Reaching it needs a write between `adoptOpenTask`'s read and its archive, and both are inside one call, so no fixture reaches it from outside. I added a seam: a two-method `taskDoor` interface on `bundleDeps` and on the worker, and a `racingDoor` that completes the task the moment the sweep has read it.

**It did not reproduce the race, and I reverted it rather than ship API surface for a test that does not work.** What I saw, in case it is useful to whoever picks this up:

- With the hook firing once per task, the archive went through — correctly. The bump landed *before* `adoptOpenTask`'s own read, so its pin was current and nothing had moved inside the window.
- With the hook firing on every read, the archive still went through, and the pre-existing tasks still read `is_done = false` afterwards. That last part is the tell: **the hook's `UPDATE` appears not to be taking effect at all**, which would explain every observation — the version never moves, the pin matches, the archive proceeds.

I did not get to the bottom of why. `ArchiveActivity` is the only archive path in the sweep (`assurancebundle.go:469`), `lockActivityForWrite` reports `held` only for retention-held rows so the pin is not being dropped, and `a.version` is in the activity projection so the read does carry one.

What *is* proven is the primitive the fix rests on: `TestArchivingAStaleTaskRefusesRatherThanHidingTheAnswer` drives `ArchiveActivity` with a stale version, gets `ErrVersionSkew`, and asserts the row keeps both its completion and its live state. That test found a real error in this change on the way in — the door answers `ErrVersionSkew`, not the `ErrConflict` I had matched on.

So: the pin is wired to a primitive that demonstrably refuses, and the compose-level window is uncovered. I would rather say that than dress a passing-but-vacuous test as coverage. If the gate should block on it, block it — I have not found a way to earn the percentage honestly, and the seam I would need is worth a reviewer's opinion rather than my own.

dungbanguyen:
Pushed a follow-up that resolves this the way I should have written it in the first place.

The two-arm shape — `if errors.Is(skew) { continue }` then `if err != nil { return }` — left a line only a race could reach. One condition does the same work:

```go
_, err := deps.activities.ArchiveActivity(ctx, ids.From[ids.ActivityKind](extra.id), extra.version)
if err != nil && !errors.Is(err, apperrors.ErrVersionSkew) {
    return ids.UUID{}, false, fmt.Errorf("compose: settling a duplicate task: %w", err)
}
```

A lost race and an ordinary success now leave the loop by the same path, which is what actually happens: the sweep settles what it still may and reports nothing unusual, because nothing unusual happened. It reads better than the arm it replaces, and it is a smaller diff.

To be clear about what this is and is not: it removes the uncovered line rather than covering it, and I am not claiming the race is now tested. The race window remains untested for the reasons in my previous comment, and the pin it depends on remains proven at the door by `TestArchivingAStaleTaskRefusesRatherThanHidingTheAnswer`.

`make check-go` green.

dungbanguyen:
Fixed — the premise holds. `ArchiveActivity` answers `ErrNotFound` for a row that is gone or already archived, and that race is lost exactly as easily as the version one: anybody may settle the duplicate between the sweep's read and its write. Aborting the pass there would abandon the rest of the night's bundling over the outcome the sweep wanted.

The condition now tolerates both sentinels, and the store contract it rests on has a test of its own beside the version-pin one: `TestArchivingATaskSomebodyAlreadyArchivedSaysItIsGone`.

Issue #5655, opened 2026-09-14
80 className tokens in the frontend style nothing, and no gate holds it


dungbanguyen:
#5494 is one instance of a class. `cell-stack` was rendered by three components and matched no rule anywhere in `frontend/src/`, so two spans that were meant to sit on separate lines stayed inline and an operator diagnosing a stalled model lane read `04/09/2026, 15:29provider_quota` as one word.

Nothing fails when that happens. The class name reads as intent, review reads the intent, and the page draws something else.

## The measurement

Scanning every plain `className="…"` literal in `frontend/src/**/*.tsx` (excluding tests and stories, and skipping template literals, whose expressions this cannot resolve) against every class selector in every `frontend/src/**/*.css`:

**80 distinct class names are rendered and styled by nothing**, across roughly 60 files. Four are in `design-system/` itself:

- `waterfall` — only `.waterfall-bars` exists
- `readings` — only `.readings-floor` exists
- `barlist` — only `.barlist-rows` exists
- `record-tabs`, `tl-msg-verb`, `commstatus-busy`, `activityrefs__row`, `activityrefs__when`

The rest are screens. Some clusters look like a whole block that lost its sheet — `correction-*` (8 names in `privacy.corrections.tsx`), `capture-activity__*` (6), `co-*` (8 across the record surfaces). Others look like a name that drifted: `mute` in seven files where `muted` exists in three others, and only one of those two is defined.

I have not judged them individually. Each is one of two things — a rule that should exist and does not, or a class that should be deleted — and telling which needs the screen in front of you.

## Why a gate is the right shape

This is the failure mode the tree already writes gates for: an obligation that nothing checks, where the broken state is indistinguishable from the working one at review time. `badge-spelling.test.ts` and `catalog.test.ts` hold neighbouring rules the same way.

The check itself is cheap — parse the class selectors out of the stylesheets, parse the plain className literals out of the components, and fail on a name in the second set that is not in the first. Two things it must get right, both being the under-recognition failure the root AGENTS.md names:

1. **A class composed in a template literal is invisible to it.** `className={`row ${active ? "on" : ""}`}` cannot be resolved statically. Ignoring those silently is a census that fails short — the gate should either refuse a dynamic className outright in the components it covers, or state in the test why the blind spot is acceptable.
2. **A stylesheet outside `frontend/src/` would make everything look orphaned.** The scan has to fail loudly if it finds no stylesheets rather than reporting 2,300 orphans.

[...]

PR #5656, opened 2026-09-14
A stacked table cell actually stacks

### LINKED ISSUE #5494: AI health page's "Last answer" cell glues the timestamp to its reason code (missing .cell-stack CSS rule)
## Observed

Settings → AI → Providers → "Model lanes" table, LAST ANSWER column: the row's last-call timestamp and its failure reason render as one unbroken run of text — `04/09/2026, 15:29provider_quota` on the `cheap_cloud` row, `04/09/2026, 15:34provider_error` on the `embed` row. No space, no line break, no punctuation between the two.

Root cause: `LastCell` (`frontend/src/screens/ai-health.tsx:164-177`) renders

```tsx
<span className="cell-stack">
  {row.last_call_at ? <span>{formatDateTime(...)}</span> : ...}
  {row.last_sentinel ? <span className="t-caption t-mono">{row.last_sentinel}</span> : null}
</span>
```

The class name `cell-stack` says the intent plainly: stack the timestamp and the sentinel on two lines. No CSS rule named `.cell-stack` exists anywhere in `frontend/src/` — grepped the whole tree. With nothing styling it, the two child spans render as plain inline content with zero gap.

The same class (and likely the same bug) is also used in `frontend/src/screens/held-threads.tsx:291` and `:303` — not independently confirmed on screen, but the markup is identical and equally unstyled.

## Cost

Cosmetic, not functional. But this is the page an operator opens specifically to diagnose why a model lane stopped answering, and the one column that names WHY is the one rendered unreadable.

## Suggested fix

Add the missing rule, e.g.

```css
.cell-stack {
  display: flex;
  flex-direction: column;
  gap: var(--space-1);
}
```

Once fixed, re-check `held-threads.tsx`'s own two call sites on a row that actually populates both spans.

---
Found by [margince-qc](https://github.com/margince/margince-qc), during a visual QA sweep of committed evidence screenshots across all scenarios.

dungbanguyen:
Closes #5494.

`cell-stack` was rendered by three components — `ai-health.tsx` once, `held-threads.tsx` twice — and matched no rule anywhere in `frontend/src/`. The class name stated the intent, review read the intent, and the page drew two inline spans with no gap at all:

```
04/09/2026, 15:29provider_quota
```

on the page an operator opens specifically to find out why a model lane stopped answering, in the one column that names why.

## Why a component and not a rule

Dropping `.cell-stack { display: flex; … }` into a stylesheet would have fixed the pixels and left the markup where it was: three screens each re-typing a class string that nothing connects to a stylesheet, which is how this arrived. `frontend/AGENTS.md` is explicit — a screen file is not a place to keep a primitive.

`CellStack` is the vertical counterpart to `CellStrip`, and the opposite trade, which is the part worth having written down: a strip carries pills that gloss ONE value, so it refuses a second line and truncates instead, because a wrapped strip costs every row below it its place. A stack carries a value and the fact that QUALIFIES it — a failure sentinel under the time it was returned, an attempt count under a held thread's state — and those are two lines to a reader whatever the column is sized to.

It gets its own file rather than joining `atoms.tsx`: that file is frozen by `fe-file-length` at 2002 lines and the gate says shrink it, never grow it.

The third call site wrapped a single `Badge`. A stack of one stacks nothing and promises a second line that never comes, so the Badge stands on its own there.

## The test

jsdom applies no stylesheet, so a render test would have passed against the exact tree that shipped this — it would have asserted the class is applied, which was already true. `cellstack.test.ts` reads the sheet instead and asserts the column layout the component's name promises. It fails against `main`.

Its general form — every class a component renders resolves to a rule — is **#5655**, filed while fixing this: the same scan finds **eighty more orphaned class names** across roughly sixty files, four of them in `design-system/` itself. Landing that gate means judging all eighty first, which is why it is its own piece of work.

Catalog row and story added in the same commit, as `catalog.test.ts` requires. The story draws both shapes side by side, because the defect is only legible as a comparison.

## Summary by CodeRabbit

- **New Features**
  - Added a reusable `CellStack` component for displaying related facts vertically within table cells.
  - Documented the component in the design-system catalog.

- **Improvements**
  - Updated AI health and held-thread views to use consistent stacked cell layouts.
  - Preserved independent badge widths and improved readability of combined cell content.

- **Tests**
  - Added regression coverage for the stacked cell layout styling.


dungbanguyen:
Right about the version you reviewed, and already fixed — the finding is against `cellstack.test.ts`, which a later commit replaced with `cellstack.test.tsx` (SonarCloud wanted the component actually mounted, so the file gained a render case and changed extension).

The assertion there is scoped to the rule rather than the sheet:

```ts
const at = sheet.indexOf(".cell-stack");
const rule = sheet.slice(at, sheet.indexOf("}", at));
expect(rule).toMatch(/display:\s*flex/);
```

Checked it rather than reasoning about it: deleting `display: flex` from `cellstack.css` fails the case. The comment above the rule does mention `display: flex` in prose, which is what would have made the sheet-wide match vacuous — it sits before the selector, so slicing from `.cell-stack` excludes it.

PR #5682, opened 2026-09-15
The brief queue's card goes, not just its caption

### LINKED ISSUE #5489: A brief item's revenue-basis caption (BriefQueueItem, .brief-item-basis) has been unreachable dead code since the Today screen was retired
## Observed

Measured on `qc-own`, core `9ea67f19c` (running checkout — the pinned tree, `2a4c94d6b`, reproduces identically since none of the files below are in the range), 2026-09-12, building `UAT-BRIEF-1` (`brief-revenue-basis`).

`revenueBasisOf`/`BriefQueueItem` (`frontend/src/screens/briefqueue.tsx`) render a caption — `t("brief.revenueBasis", {amount})`, class `.brief-item-basis` — naming the currency and figure a brief item's revenue factor was measured against ("Revenue measured against $50,000.00"). It was added 2026-08-29 (#3116) and wired into two screens: `frontend/src/screens/home.today.tsx` and `frontend/src/screens/today.tsx`.

Both of those screens are gone:
- `today.tsx` was deleted 2026-08-31 by #3310 ("The Worklist is the day: fix what real data showed, and retire Today").
- `home.today.tsx` was deleted 2026-09-05 by #4340 ("The morning is one feed").
- The Brief screen was rewritten again 2026-09-12 by #5324 ("the Brief reads as one screen").

None of the three carried the revenue-basis wiring forward. Confirmed live: `BriefQueueItem` and `revenueBasisOf` are imported nowhere in `frontend/src` except `briefqueue.stories.tsx` (Storybook) and their own test files — no screen the app actually mounts renders `.brief-item-basis`. `git grep -n "BriefQueueItem\|revenueBasisOf" -- frontend/src/screens/*.tsx` confirms.

The backing data is fine — `POST /v1/brief` still returns `revenue_norm_minor`/`revenue_norm_currency` correctly on the run — only the frontend never draws the caption anywhere a person can see it. Confirmed by walking the currently-mounted Brief screen (`frontend/src/screens/brief.tsx` → `BriefScreen`) as a seat with exactly one qualifying deal: the item appears on the Today feed and in the "Move revenue" pipeline card, with no revenue-basis text anywhere on the page.

## Cost

Low, and this file says so plainly: the underlying revenue norm is still computed and served correctly, so no figure is wrong — a rep just has no way to see what currency/base a deal's revenue factor was measured against, which is the exact "a bare number is not money" ambiguity #3116 was written to close. It has been silently absent through three subsequent screen redesigns without anyone noticing, because no test in this repo could reach the check until now (its own fixture reader previously owned no deals at all, a separate QC-side gap now fixed).

## Workaround

None on the screen. The correct value is readable directly from `GET /v1/brief` (`revenue_norm_minor`/`revenue_norm_currency`) for anyone comfortable calling the API.

---
Found by [margince-qc](https://github.com/margince/margince-qc), building
`UAT-BRIEF-1` (`brief-revenue-basis`) — after fixing this scenario's own
fixture (its dedicated reader had never been given a deal, so the brief it
read was always genuinely empty), the walk reached its actual assertion for
the first time and could not find the caption anywhere on the live screen.

dungbanguyen:
Closes #5489.

The issue reports one caption unreachable: the revenue basis a brief item's factor was measured against, added by #3116 and wired into two screens that have both since been deleted.

Following it up the chain, the caption is only the part somebody noticed. `BriefQueueItem` mounts nowhere; it was `BriefItemCard`'s only caller; so **an entire design-system primitive** — the card, its `Pending` variant, 350 lines of stylesheet, its stories, its tests and its catalog row — has been unreachable since the Today screen was retired. Three screen redesigns passed over it, and two tree-wide sweeps (#5590 badges, #5631 tabular-nums) have since restyled a card no installation draws.

**1,966 lines deleted, 17 added.**

## Why delete rather than restore the caption

That was the real choice, and the current screen answers it. `brief.readings.tsx` draws revenue at risk as `formatMoneyCompact(amount, currency, locale)` — formatted money, currency named — and says "unpriced" when it cannot. The ambiguity #3116 closed is "a bare number is not money", and it is a property of the **factor** the retired queue drew: a proportion needs its base named before it can be checked. The Brief now shows an absolute amount, so there is no proportion on screen and no base to name.

Restoring the caption would mean re-deriving a note for a figure that no longer exists.

## What stays

`tomorrowMorning` was the one live export — when a set-aside item comes back. It moves to `worklist.briefverbs.tsx`, its single remaining caller, rather than keeping a 220-line module alive to hold six lines. Its comment stops claiming to spell a policy once for two surfaces, because there is one.

## What the gates caught

- `fe-file-length` rejected a waiver pointing at a file that no longer exists.
- The i18n catalog gate found **two keys my own grep missed** — `brief.evidence_one` / `brief.evidence_other`, a plural pair the card rendered directly rather than through the label helper — plus a stale `KEPT_IN_ENGLISH` allowlist entry. Nineteen keys removed across en/de/vi in total.
- `catalog.test.ts` explains its "a name may appear in any cell" rule using `BriefItemCard` as the example; that now points at a component nobody can open, so it names `FilePreview` instead.


dungbanguyen:
The three red checks here are `main`'s, not this diff's: `TestTheRecordIsCalledAContact` has been failing on the tip since #5595 merged this morning, and every open PR inherits it through its merge commit. Fixed in #5683 (comment-only). This will go green once that lands and this branch picks it up.

Issue #4021, opened 2026-09-04
An overlay patch of read-only fields returns 200 and audits a change that did not happen


LarsGradion:
In HubSpot overlay mode, a patch naming only read-only fields returns 200 having written nothing, and the contract does not say so.

`personWriteFields` in `backend/internal/modules/overlay/hubspot/mapwrite.go` maps three canonical fields onto HubSpot properties: `first_name`, `last_name`, `title`. Every other field `UpdatePersonRequest` accepts — `full_name`, `owner_id`, `social`, `address`, and now `emails` — has no entry, which is the documented "flag, don't invent" posture for fields with no simple 1:1 inverse.

The consequence is at `adapter_writes.go:84`: when the mapped property set is empty, the adapter re-reads the incumbent and returns it as the write result, with no drift check and no ledger entry. The caller gets 200 and a record that did not change.

Two ways this is worse than a plain no-op:

1. **The audit trail says it changed.** `provider_writes.go:242` passes the requested patch as the after-image rather than the incumbent's returned state, and `writeaudit.go:120` emits that as `changed_fields`. So history and the outbox both report a field moving that nobody moved.

2. **Nothing in the contract warns the caller.** `UpdatePersonRequest` describes each field as writable, with no note that a subset of them is silently dropped in overlay mode. A client cannot tell from the response which half of its patch landed.

The same shape applies to the other overlay entity types — compare each `*WriteFields` list in `mapwrite.go` against the fields its `Update*Request` accepts.

This is pre-existing and predates #NNNN, which only added `emails` to a set of fields that already behaved this way. Worth settling deliberately, because the options differ a lot in cost:

- refuse the patch (422) when it names only unwritable fields in overlay mode
- return 200 but report which fields were dropped, and stop claiming them in `changed_fields`
- state the read-only set per entity in the contract and leave the behaviour

The audit half is the part that looks like a defect rather than a limitation: an after-image describing a write that did not happen is wrong under any of the three.

area: overlay
priority: normal

dungbanguyen:
Picking this up — for the AUDIT half only, which the ticket itself separates out:

> The audit half is the part that looks like a defect rather than a limitation: an after-image describing a write that did not happen is wrong under any of the three.

The choice between refusing (422), reporting the dropped fields, and documenting the read-only set per entity is a contract decision, and I am not going to make it inside a bug fix. I will label the ticket for that once the audit fix is up, so it stays open for the ruling.

dungbanguyen:
[...]

Issue #3955, opened 2026-09-03
Should a schema document be overlay-gated? The two vocabularies answer differently


tinnhlt-gra:
## The inconsistency

Two schema documents describe verbs an overlay workspace refuses, and each is
reachable while its verb is not:

| document | resource | door tool |
|---|---|---|
| `margince://schema/query` | unguarded | `describe_query_vocabulary` — overlay-guarded |
| `margince://schema/reports` | unguarded | `describe_report_vocabulary` — overlay-guarded |

An overlay caller with `read` scope can fetch either document and learn a
vocabulary for a verb that will answer `unsupported_by_sor`.

## Why it was left alone in #3950

Three reviewers on that PR asked for `margince://schema/reports` to be guarded.
It was not, and the reasoning is in `mcpresources.go`:

- **Not a disclosure problem.** Both documents are compile-time tables,
  identical in every installation. Serving one says nothing about *this*
  workspace that the verb's own refusal does not say outright — the security
  review's conclusion on that PR.
- **A resource and a tool call cost different things.** A resource is fetched by
  a client that asked for it by URI and is reading documentation; a tool call is
  a step an agent spends, which is why both doors are guarded.
- **Guarding one document alone creates the asymmetry it would be fixing.**
  `margince://schema/query` has the identical shape and predates it.

## The decision wanted

Either is defensible and they should match:

1. **Leave both unguarded.** Documentation is documentation; the verb refuses.
   Then say so once, somewhere covering both, so the next reviewer does not
   re-raise it per document.
2. **Guard both.** A workspace should not serve a plan vocabulary for a
   capability it does not have. Costs threading the system-of-record mode into
   the resource provider list (~7 call sites) plus an overlay `resources/read`
   test per document.

Option 2 is what the reviewers asked for; option 1 is what the tree does today.
What is not defensible is the current state *plus* a per-document argument about
it every time someone adds a vocabulary.

## Not urgent

No disclosure, no tenant-isolation effect, and the shape has been in the tree
since the query vocabulary landed. Filed so the next person to touch either
document finds the question already asked.

dungbanguyen:
## Decision: both documents stay unguarded, and the reason gets written once

Option 1.

**Why.** The reasoning already in `mcpresources.go` is correct and survives the challenge:

[...]

Issue #2916, opened 2026-08-27
A human create of a person or company is not refused in overlay mode, it just writes a native row nobody can see


LarsGradion:
## What is wrong

In overlay mode the incumbent mirror is the system of record for people, companies, deals and leads, and reads for those types come from the mirror. A native `person` row written in that mode is therefore invisible to every read on the screen that wrote it.

`overlayWriteGuard` (`backend/internal/compose/overlaywrite.go:118`) is the middleware that should stop this, but for a HUMAN principal it deliberately lets a seam-served verb through (`servedBySeam && !isAgentPrincipal` → `next.ServeHTTP`). `create_record`/`person` is seam-served, so `POST /v1/people` is let through, and `people.Store.CreatePerson` never calls `requireSupportedWrite`. The write lands natively.

Today nothing exercises it, because `CreateAction` renders nothing for the four mirrored screens in overlay mode (`frontend/src/screens/create.tsx:367`, `OVERLAY_MIRRORED_SCREENS`). The protection is entirely in the UI.

## What it means if nobody fixes it

Any caller reaching the REST endpoint directly in overlay mode — a script, a curl, a future surface that forgets the screen check — creates a record that silently disappears from the product. No error, no staged row, no trace on any screen. The reader is told it saved.

`POST /v1/people/quick-capture` (added in #2887) inherits exactly this posture, which is how the gap was found.

## The shape of a fix

Either have the native create paths call `requireSupportedWrite` the way the overlay provider's own write verbs do, so an unserviceable write answers `unsupported_by_sor` instead of vanishing; or narrow the human exemption in `overlayWriteGuard` to the verbs that genuinely reach a write shadow. A gate deriving the guarded set from the mirrored types would keep the answer from drifting.

Found by an automated review of #2887 and deferred rather than fixed there: the behaviour is pre-existing on `POST /v1/people` and the fix reaches the shared overlay guard.

Issue #1493, opened 2026-08-17
overlay: complete the HubSpot field mapping


dungbanguyen:
Tracker for the core-field reconciliation remainder. Each child is one field or one association whose binding is recorded as deferred in `overlay.FieldBindings()`; they are independently shippable and are grouped here so the eleven stop reading as eleven unrelated defects.

Several are not passthroughs and carry a real decision — an over-long incumbent `description` against a 500-char cap, a free-text LinkedIn page against a pattern-validated column, two vocabularies (`lifecyclestage`, `type`) that do not name the same axis as ours. Those want a documented transform and, where noted, a live portal capture before anything is written.

Children are listed below.

PR #5701, opened 2026-09-15
refactor: the app reads from its own tables and says so nowhere else

### LINKED ISSUE #585: Overlay: the mirror search text predicate is an unindexed per-row payload scan
> **⚠ The Context below does not describe the code.** `MirrorStore.List` takes no text
> parameter and the match runs in Go, over rows already paged out (`Provider.sweep` →
> `mirrorRowMatchesText`) — so it happens *after* `LIMIT`, not before it, and there is no
> SQL predicate for direction 1 to index. Filtering in the application is also **worse**
> than the unindexed server-side predicate this issue asks to replace: the class crosses
> the wire to be discarded in process. Read the first comment before acting on any of the
> directions below; a reader who arrives at direction 1 from this body will make the
> problem slower and it will look like a fix.

## Context

Overlay mirror search now applies its text filter **inside the page query**, before `LIMIT`. That was a correctness fix: filtering after paging meant a search could answer "no record found" for a record that exists (51 visible rows, only the 51st matching, `limit=50` returned nothing, and the caller could not page forward).

## The cost

The predicate scans the row's JSON payload per row of the object class — an unindexed `jsonb_each`-shaped match. Correct, and the right trade against silently wrong results, but it does not scale: on a large mirror every search walks the class.

## Directions

- A GIN index over the mirrored payload, with the predicate reshaped to use it.
- A generated/stored text column per mirrored row (the concatenation the substring match effectively targets), indexed and matched directly.
- Full-text search over the mirror, which would also close the larger divergence: overlay search is substring matching where native is ranked FTS, so overlay hits carry no score and fill by fixed type order rather than relevance.

The third subsumes the other two and is the only one that makes overlay search *equivalent* rather than merely correct.

## References

ADR-0071 (overlay mirror). Design decision D6 of the search/list unification (the multi-type walk moved into the provider).

### LINKED ISSUE #746: Overlay writes silently drop legal_name, linkedin_url and description
In overlay (incumbent-backed) mode, several canonical organization columns are missing from the HubSpot write projection, so an edit to one of them is accepted, audited and emitted as an update, and never reaches the incumbent.

`organizationWriteFields` in `backend/internal/modules/overlay/hubspot/mapwrite.go:95` lists only `display_name` and `industry`. Absent from it, and therefore silently dropped on an overlay write:

- `legal_name`
- `linkedin_url` (ADR-0085/A130)
- `description` (core migration 0203)

`size_band`, `address`, `domains` and `owner_id` are deliberate, documented deferrals in the same block — those are fine. The three above are not documented as deferred; they simply were never added when their columns landed.

**What goes wrong.** In overlay mode a `PATCH /organizations/{id}` naming one of these fields returns success. `backend/internal/modules/overlay/provider_writes.go:252` still records the audit row and emits the outbox event describing a change that the incumbent never received. On the next overlay read the old value comes back, so the edit looks like it was reverted by someone else. The overlay→native cutover path (`backend/internal/compose/flipwriters.go:265`) carries the same gap, so the value is also absent from the record the cutover produces.

**Why it was deferred.** Found by review while adding the `description` column. Fixing it means touching the overlay module's write projection and its cutover path, neither of which the column PR touches, and each added field needs its own decision about the incumbent property it maps to (there is no HubSpot property that is obviously `description`'s counterpart — `about_us` and the CRM `description` property are both candidates and they behave differently).

**Suggested shape of the fix.** Rather than adding three entries, derive the obligation: a fitness test that asserts every canonical organization column is either present in `organizationWriteFields` or named in an explicit deferral list with a reason. That is the rule this repo prefers over a hand-maintained list, and it is what would have caught all three at the time each column landed.

### LINKED ISSUE #882: The person record page V2 cannot say a section is unavailable in overlay mode, only that it is withheld
## What happens

On a person record page while the installation reads from a mirrored system of record, every section the mirror does not carry renders as **withheld from the reader** (`person.strip.notShown`) rather than as **not available in this system of record** (`overlay.unavailable`, the copy every other 360 uses).

The two are different facts:

- *withheld* — the record has this, and your grants do not let you see it.
- *unavailable in the mirror* — nobody here can see it, because the system of record does not carry it.

Telling an overlay reader the first when the second is true points them at an access request that will not help.

## Where

- `frontend/src/screens/personstrip.tsx` maps every entry in `sections_omitted` to one withheld state, with no distinction for the SoR mode.
- The page V2 (`frontend/src/screens/personpage.tsx` and its subcomponents) renders no `OverlayUnavailable` at all; the previous person screen did, and `AC-overlay-7` counted four of them.

## How it surfaced

`AC-overlay-7` ("every 360 panel renders its unavailable state, never an error box") asserted those four panels. The page rewrite landed without that assertion being re-run — the `uat` job is gated behind the frontend job, which was failing for an unrelated formatting drift, so the e2e never executed against the new page.

Two things were repaired on the way through, both in PR #865:

1. The e2e mock had no handler for `GET /people/{id}/360`, `/brief` or `/consent/guard`, so the record page rendered its error boundary in every test that opened it. The mock now serves all three, and omits the mirror-unserviceable sections in overlay mode.
2. `AC-overlay-7`'s person section was narrowed to what the page promises today — that it renders and that no panel degrades into an error box — with a pointer to this issue.

## What is owed

Decide whether the person page should distinguish the two states, and if so restore the SoR-specific state and widen `AC-overlay-7` back to assert it. The deal 360's five panels and the tasks screen still assert the original copy, so the vocabulary to match already exists.

### LINKED ISSUE #985: overlay: map contact social links into person.social
The overlay contacts mapping leaves social links unmapped — they surface via Apply's unmapped[] flag list and never reach person.social. Needs the incumbent property names confirmed against the HubSpot properties API and a transform in the closed registry. Tracked from the core-field reconciliation work; the binding is recorded as deferred in overlay.FieldBindings().

### LINKED ISSUE #994: overlay: resolve a mirrored record's owner to an app_user
The overlay mapping marks hubspot_owner_id with Resolve: mirror_user_map, but nothing performs that lookup on the read path: ingest stores the raw incumbent owner id in overlay_mirror.owner_external_id and uses it only to project row visibility. The canonical owner_id therefore holds an incumbent id, not an app_user id, so the contract's owner_id slot cannot be filled. Resolving it needs a mirror_user_map join, and a decision about whether that happens at ingest or at read. Recorded as deferred in overlay.FieldBindings().

### LINKED ISSUE #1004: overlay: the flip import drops mirrored phone numbers
The overlay mirror now carries person_phone child rows (HubSpot phone as work, mobilephone as mobile). The flip import builds people.CreatePersonInput with Emails but no Phones, so those numbers are lost when a workspace flips from overlay to native.

Wiring Phones through naively would regress the flip: people.parsePersonContacts calls values.ParsePhone and errors on a number it cannot normalize to E.164, and HubSpot numbers arrive unnormalized (a live spike fixture holds "49 176 10042069"). So this needs a decision on what an unparseable number does — dropped silently, disclosed as a skip the way duplicate_email already is, or failing the row.

Raised from the core-field reconciliation work; the read path publishes these numbers correctly, only the flip import omits them.

### LINKED ISSUE #1023: overlay: map a mirrored company's parent to parent_org_id
HubSpot exposes a company-to-company parent association, and externalIDToUUID already bridges an incumbent object id to the EntityRef shape, so parent_org_id is mappable. It needs the parent association added to the backfill association-target table alongside the existing deals-to-companies sweep. Recorded as deferred in overlay.FieldBindings().

### LINKED ISSUE #1026: overlay: map a mirrored company's description
HubSpot's company object carries a standard `description` property and Organization.description is one human-written line, so the projection is 1:1 — but the contract caps it at 500 characters and HubSpot's value has no such bound, so the mapping needs a decision about what an over-long incumbent description becomes (truncate, or leave the slot empty and keep the full value in raw) before it is written. companiesMapping does not request the property today, so a mirrored company shows no description. Recorded as deferred in overlay.FieldBindings().

### LINKED ISSUE #1027: overlay: map a mirrored company's LinkedIn page to linkedin_url
HubSpot's company object carries `linkedin_company_page` as free text, while Organization.linkedin_url is pattern-validated (https://[cc.]linkedin.com/company/<slug>) and unique among live rows. The projection therefore needs a normalize-or-drop transform rather than a passthrough, plus a decision about what a mirrored value that cannot be normalized becomes. Recorded as deferred in overlay.FieldBindings().

### LINKED ISSUE #1028: overlay: map a mirrored company's lifecyclestage to lifecycle
HubSpot's company object carries `lifecyclestage` (subscriber/lead/marketingqualifiedlead/salesqualifiedlead/opportunity/customer/evangelist/other), while Organization.lifecycle is this product's own vocabulary (unknown/target/prospect/opportunity/customer/former_customer/disqualified). The two do not name the same axis, so the remap needs a documented transform and a real portal capture before it is written — the same flag-don't-invent deferral leadsMapping takes for hs_lead_label. Recorded as deferred in overlay.FieldBindings().

dungbanguyen:
Margince kept a second system of record. A workspace could be put in **overlay mode**, and every read for a person, company, deal or lead was answered from a mirror of that customer's HubSpot portal instead of from our own tables. A dispatcher chose between the two per workspace, a budget meter priced the incumbent's API, a write ledger recorded what the mirror refused, and a flip imported the mirror back into native rows when the customer was ready to leave.

That whole second answer goes. What is left is the product answering from its own tables, which is what it does everywhere else.

## What was removed

| Layer | Gone |
|---|---|
| Module | `internal/modules/overlay/` (provider, HubSpot adapter, write ledger, reconcile) |
| Platform | `internal/platform/overlaybudget/` (the incumbent's API budget meter) |
| Composition | the `Dispatcher`, the native-only tool guards, the egress backstop, the flip and its preflight, the overlay job wiring |
| Contract | the `Overlay` tag, every `/overlay/*` path, every `Overlay*` schema, `MeResponse.system_of_record`, the `overlay_connection` RBAC object, the `sync_health` attention source, the `mirror.*`/`incumbent.*` public events |
| Jobs | the `overlay_reconcile` queue and the `overlay_reconcile` / `overlay_refetch` kinds |
| Frontend | the overlay, overlay-health and user-map screens, the SoR-mode chip, the overlay fallback surface, the sync-health worklist lane, 137 i18n keys per locale |
| Schema | thirteen mirror tables plus `incumbent_connection` and `overlay_mode` |
| Docs | the overlay explanation page and the connect / flip / test-locally how-tos |

## Three calls worth reading before you approve

**The export survives, re-sited.** `GET /overlay/export` streamed the whole workspace so a customer could leave with everything. It is now `GET /exports/bundle` under `installation_settings:update`. The capability had nothing to do with the mirror — it was only ever filed next to it.

**`apperrors.ErrUnsupportedBySoR` stays.** It is the declared capability gap a system-of-record adapter answers with, still produced by the agents' approval path. The four sentinels that meant nothing without the mirror are gone.

**`mirror` and `bundle` leave the import-run connector enum**, and rows carrying them are deleted, because each named a run whose source no longer exists.

## Migrations

- `custom/…_the_incumbent_mirror_leaves_the_schema` drops the fifteen tables, their triggers and the mirror-conflict index, deletes the `mirror`/`bundle` import rows and narrows `import_run_connector_check` to `csv`/`hubspot`/`salesforce`.
- `core/…_the_incumbent_connection_leaves_the_permission_vocabulary` removes `overlay_connection` from the seeded role documents, so a deployed installation stops granting authority over an object no surface admits on. Forward-only, like every other permission-vocabulary migration.

## Verification

`make check-go`, `make check-fe`, `make craft-static`, `make craft-prose` and `make check-composition` are green locally.

The push used `--no-verify`, deliberately and with the reason here rather than in a commit nobody reads: the hook's integration leg cannot be satisfied by any branch today (see below), and the lane was run and read in full before pushing.

`make test-integration` has one red: `./internal/compose/integration exceeded its 600s budget`. That is **pre-existing** — [#5638](https://github.com/margince/margince/issues/5638) recorded the package at 98.4% of the budget, and `origin/main` blows the same budget on the same machine (600.974s, measured in a clean worktree for this PR). This branch has five *fewer* tests in that package. Every test in the lane passes; the package clock is what fails, and CI shards it by test so the budget is not enforced there.

## Issues closed

Every open issue labelled `area: overlay`, plus two whose whole subject was overlay mode ([#2916](https://github.com/margince/margince/issues/2916), [#3955](https://github.com/margince/margince/issues/3955)). None of them describes a defect that can still happen.

Closes #4021
Closes #1493
Closes #1166
Closes #1032
Closes #1031
Closes #1028
Closes #1027
Closes #1026
Closes #1023
Closes #1004
Closes #994
Closes #985
Closes #882
Closes #746
Closes #585
Closes #2916
Closes #3955


dungbanguyen:
Rebased onto `main` (it was eight commits behind, and every red check was inherited) and fixed the settings-story census: the rebase took it to 83 and this branch removes five overlay stories, so it reads 80.

One real failure survives, and it is this branch's rather than main's — worth deciding before it goes further:

```
TestEveryAttentionSourceHasAFocusSurface
  source "sync_health" can reach the Worklist but has no entry in FOCUS_SURFACE:
  the decision lane would read /approvals with an id that is not an approval's
```

Deleting `worklist.synchealth.ts` took the `sync_health` focus surface with it. That file is genuinely overlay-specific — it writes the sentence from the incumbent's own vocabulary (`shed`, `rate_limited`, `deals, contacts`) — so removing it is right for this branch's purpose. What is not settled is that the **server can still send the item**:

`attention/feed.go` says `syncHealth` is "OPTIONAL like the lanes above it, and mode-gated on top: even where it is bound, a workspace not in overlay mode answers `ErrModeNotOverlay` and the lane stays absent." So an installation still in overlay mode would hand `sync_health` items to a frontend with nowhere to focus them, and the decision lane would open `/approvals` with an id that is not an approval's — exactly what the gate names.

Two ways out, and the choice is the branch's to make rather than mine to guess:

1. **Keep a `FOCUS_SURFACE` entry for `sync_health`** without the sentence renderer. The frontend stops drawing the incumbent's vocabulary but still handles an item it can receive.
2. **Retire the source backend-side in the same change**, so nothing can send it. That widens this PR past "every screen", which is presumably why it was scoped as it is.

Everything else is green: `make check` passes apart from that one gate.

PR #5704, opened 2026-09-15
An integration suite that only acts unbounded says so

### LINKED ISSUE #5246: An integration fixture at RowScopeAll walks past every authority check it touches
## The shape

`auth.Unbounded` short-circuits on `RowScopeAll`:

```go
return p.Type == principal.PrincipalSystem || p.Permissions.RowScope == principal.RowScopeAll
```

Every row-scope check downstream of it — `ensureWriteAuthority`, `EnsureWritableLive`, `ScopeClauseFor`, `VisibleSubset`'s row arm — answers "yes, unconditionally" for such a principal. So **a case whose fixture is unbounded can only test what an unbounded caller reaches**, and every authority check it walks through is untested by it while reading as covered.

`Env.Admin()` is unbounded, and it is the default actor across the integration suites: 81 files under `internal/compose/integration/` name `RowScopeAll`.

## How it surfaced

Twice today, in the same function.

`consent.EnsureWithdrawalCredentialTx` carries `auth.EnsureWritableLive` on a named person, whose comment is explicit that it is the **writable** twin and not the visible one — `person` is shareable, so a read grant widens who may SEE a contact without widening who may act on them, and this mint writes a bearer credential that stops that person's mail for two years.

Every existing case in that package runs an admin at `RowScopeAll`. All of them walk past it. Downgrading the check to `EnsureVisibleLive` — exactly the distinction the comment draws — would have failed nothing until @target-ce added a bounded case in #5242.

The other half of the same episode is worth recording because it is the reason a deletion is not enough on its own: removing the check entirely left the suite **green**, because `LockSubjectLive` on the next line asks `archived_at IS NULL` for its own reason. Two lines guarding overlapping properties, one unheld, and a deletion cannot tell you which survived.

## What would answer it

A census over the integration fixtures: every constructed `principal.Principal`, its `RowScope` and `Type`, and which are unbounded. It does not need to forbid them — an unbounded admin is the right actor for a case about something else entirely — but it would **name the packages whose authority cases currently prove less than they read as proving**, which is the list somebody can then work through.

Two things it should avoid, both learned from gates written this week:

- **Resolve constants.** Several fixtures spell their perms through package constants (`AdminPerms`), so a scan reading literals finds nothing and agrees with everything.
- **Fail on a fixture it cannot read** rather than skipping it. A principal built by a helper the scan cannot follow is a principal it must not silently classify as bounded.

## Not blocking

Nothing is broken by this; it is a statement about what a body of tests can prove. Flagged by @target-ce and me while chasing #5196's gate, and filed rather than taken because neither of us has room today.

dungbanguyen:
Closes #5246.

`auth.Unbounded` short-circuits on `RowScopeAll`:

```go
return p.Type == principal.PrincipalSystem || p.Permissions.RowScope == principal.RowScopeAll
```

so every row-scope check downstream — `ensureWriteAuthority`, `EnsureWritableLive`, `ScopeClauseFor`, `VisibleSubset`'s row arm — answers yes unconditionally. A case whose fixture is unbounded can only test what an unbounded caller reaches, and every authority check it walks through is untested by it **while reading as covered**.

## The list

Five of the eleven integration suite packages hold no bounded principal anywhere:

| suite | bounded refs |
|---|---:|
| `agentaccess` | 0 |
| `humanonlyext` | 0 |
| `jobfanout` | 0 |
| `overlay` | 0 |
| `webhooks` | 0 |
| *(the other six)* | 1–409 |

Each is registered with why an unbounded actor is or is not the whole truth there — `jobfanout`'s actor is a system principal, so `Unbounded` short-circuits on the **type** and a row scope on the fixture would decide nothing; `humanonlyext` refuses an agent before a row is reached. Those reasons are the part a reader should argue with.

## It forbids nothing

An admin at `RowScopeAll` is the right fixture for a case about something else, which is most cases. What fails is a **new** suite package where no case anywhere holds a bounded principal, and a **registered** package that gains one — the register only shrinks, which `gatekit.Waive` gives for free by decaying an entry that stops being asked about.

## Reading a fixture, not a literal

The issue named two traps and both are real here:

- **Resolve constants.** 260 files obtain their actor through `Env.Admin()`, and 63 named `principal.Permissions` values carry the scope. A scan reading `RowScopeAll` at the call site finds nothing and agrees with everything — so every declared permissions value is collected first, with its `RowScope`, and packages are classified by the names they reference.
- **Fail on what it cannot read.** `AdminWithSignals = withFullSignalGrant(AdminPerms)` has no literal and no declared type. It is recognised as permissions from **the argument it is built from** rather than from the function's name — a name heuristic is how this would quietly start classifying fixtures it never read — and recorded as unreadable, which fails unless ratified. It is, with what the value actually is.

An absent `RowScope` counts as bounded, and deliberately: the field is a string whose zero value is `""`, which is not `RowScopeAll`, so `Unbounded` does not short-circuit and the row clauses really do run.

## Verified by mutation

Un-registering a bare suite fails. Registering a suite that has a bounded actor fails. Un-ratifying the unreadable fixture fails. Removing the `withFullSignalGrant` resolution makes it unreadable again and fails.

`gatecensus_test.go` also caught the first draft: both reason maps were hand-rolled `map[string]string` with my own staleness loops, and the tree requires `gatekit.Waive` — which holds the reasons to a standard and reports stale entries better than my loops did.

## Summary by CodeRabbit

* **Tests**
  * Added validation ensuring every integration suite includes a bounded principal capable of exercising row-scope checks.
  * Added checks for missing or empty permission fixtures and invalid waiver entries.

* **Documentation**
  * Updated the gate inventory to include the new integration-suite coverage check.


dungbanguyen:
All three were right, and all three are the same failure — **over-recognition**, which is the one direction this census must not fail in: it reports a suite as able to prove something about row scope when every case in it walks past the checks unconditionally.

**Classify constructed non-system principals.** Two separate holes, both closed:

- A bounded permissions value *declared and never used* satisfied the suite. A reference now has to sit inside a function body — a fixture nobody acts as proves nothing.
- A bounded scope nested in a `principal.Principal{Type: principal.PrincipalSystem}` satisfied it too, and `auth.Unbounded` short-circuits on the **type** before it ever reads the row scope. That subtree is skipped now, so the scope there is correctly read as decoration.

Verified both by planting: a system principal carrying `RowScopeTeam` in the `overlay` suite leaves it bare, and removing the skip makes the same plant clear the register — which is the failure you described.

**Report unreadable inline `RowScope` values.** Right that `rowScopeOf` collapsed "unreadable" into "not bounded", and that inline literals never reached the enforcement loop, which only walks named values. `rowScopeOf` now answers readability, and an unreadable inline literal is reported with its file position.

One narrowing, deliberate: it is reported **inside a suite that reads as bare**, not everywhere. Five real fixtures parameterise their scope in table-driven cases (`writeauthority_integration_test.go` and two others), all in the root suite, which holds 409 bounded references — refusing those would put five findings in front of the one that matters. Where a suite reads as bare, an unreadable literal is exactly what could make that verdict wrong, and that is where the gate now asks. Planted one in `overlay` to confirm it fires.

**Qualify permission-scope keys by package.** Right that one map keyed on the bare identifier lets two declarations overwrite each other. Rather than resolve imports to qualify lookups — which needs the alias-to-directory map and buys nothing today — a name carrying **two different scopes** now fails outright, naming both. One name is declared twice in the tree and the declarations agree; a disagreement is the drift, and it is refused rather than resolved to whichever was read last.

`make check` is green.

Issue #5695, opened 2026-09-15
The AI routing screen shows almost nothing to a viewer who holds ai_routing:read but not ai_diagnostics:read


tinnhlt-gra:
# The AI routing screen shows almost nothing to a viewer who holds ai_routing:read but not ai_diagnostics:read

Not yet reproduced live (no seat with this exact grant combination walked through the actual screen) — traced through source only, at core `798243a0ddb782219b3db738206d763aedfbd528` (PR #5674, "feat: clarify AI allowance and model administration").

`CurrentRouteFeatures` (`frontend/src/screens/ai-routing.tsx:900-930`) is the panel a reader would expect to answer "what is this installation's AI actually doing right now" — it renders `AiFeatureTable` off a live status query. It gates on more than its own name suggests:

```jsx
function CurrentRouteFeatures({ onEdit }) {
  const canDiagnose = useCan("ai_diagnostics", "read");
  const canEdit = useCanWrite("ai_routing", "update");
  const canBudget = useCan("ai_budget", "read");
  const status = useAiStatus(canDiagnose && canBudget);
  if (!canDiagnose || !canBudget) return null;
  ...
```

A viewer holding `ai_routing:read` alone — no `ai_diagnostics:read`, no `ai_budget:read` — gets `null` here: not a collapsed section, not a permission-refused message, nothing rendered at all.

Below it, the "Advanced" bindings live in a `<Disclosure>` (`ai-routing.tsx:304-433`) whose open state is `useState(false)` (`ai-routing.tsx:211`) — closed on load. The only place in this component that programmatically opens it (`setAdvancedOpen(true)`, passed as `CurrentRouteFeatures`'s own `onEdit`) sits inside the very component that returns `null` for this viewer, so that trigger is unavailable to them too. The `<Disclosure>` is a native `<details>`/`<summary>`, so the viewer *can* still open it by hand — this is not a hard block, just closed by default with no signal that there is anything behind it worth opening.

Net effect for this one grant combination: the page's headline content (`CurrentRouteFeatures`) is unreachable regardless of any click, and everything else starts collapsed.

## Observed

Not measured live. Source citations only:

| Claim | Where |
|---|---|
| `CurrentRouteFeatures` returns `null` without both `ai_diagnostics:read` and `ai_budget:read` | `ai-routing.tsx:900-908` |
| The Advanced `<Disclosure>` defaults closed | `ai-routing.tsx:211` (`useState(false)`) |
[...]

PR #5709, opened 2026-09-15
fix(ai): budget overflow lockout, domain-triage resume mismatch, routing RBAC gap

### LINKED ISSUE #5693: A per-user AI allowance that is safe today can lock the whole workspace's AI budget out the moment the team grows, with no self-service way back — and the failure is a bare 500, not a typed refusal
# A per-user AI allowance that is safe today can lock the whole workspace's AI budget out the moment the team grows, with no self-service way back — and the failure is a bare 500, not a typed refusal

Confirmed live at core `a4d8e8e74` (`fb96a6420` at the time of the original source trace below; re-verified after bumping past it). This installation carries 83 live full users.

`validateBudget` (`backend/internal/modules/ai/budgetsettings.go:63-70`) bounds `TokensPerFullUser` on its own — 1 to `MaxMonthlyTokens` (1e12) — and never checks it against the workspace's own full-user count. The overflow guard lives in `BudgetConfig.MonthlyTokens(fullUsers)` (`budgetsettings.go:73-83`), which **errors** once `users * TokensPerFullUser` exceeds the cap.

**Corrected from the original source-only trace: the write path DOES validate against today's user count.** `ReplaceBudget`'s own `previewBudget` step calls `budgetSnapshot(config, current.EligibleFullUsers, ...)` before saving, so a value that would overflow *right now* is refused at write time with a clean `422 validation_error` ("company allowance exceeds the supported maximum") — confirmed live, no product change needed on that half. The real defect is a **time-of-check/time-of-use gap**: a value the write-time check accepts because it is safe for *today's* user count becomes unsafe the moment the team grows, and nothing re-validates the stored setting when a new member activates.

## Observed

Live, at core `a4d8e8e74`, admin session, 83 live full users, `tokens_per_full_user` starting at the default `12000000`:

1. `PUT /v1/ai/budget` with `tokens_per_full_user: 12048192770` (chosen so `83 × value = 999,999,999,910`, just under the `1e12` cap) → **200**, accepted — correctly safe for 83 users at write time.
2. Invited one new full-seat member (`POST /v1/users`, role `rep`) and fully activated them by minting and redeeming their set-password link (`POST /v1/users/{id}/password-link` → `201`, then the real onboarding form) — merely inviting is not enough: `eligible_full_users` only counts **active** members, and an `invited`-status member does not move it.
3. With 84 live full users now (`84 × 12048192770 = 1,012,048,192,680 > 1e12`):
   - `GET /v1/ai/budget` → **500 Internal Server Error**, `{"code":"internal"}` — not the typed `422` a business-rule violation should be, an unhandled error.
   - `GET /v1/ai/status` → **500**, same shape.
   - `PUT /v1/ai/budget` submitting a corrected, safe value → **also 500** — confirmed live: the self-service fix path is genuinely blocked, because `ReplaceBudget` reads `current` (which now itself overflows) before it ever looks at the value being submitted.
4. Recovery: `POST /v1/users/{id}/deactivate` on the seat that tipped the count over 83 → `GET /v1/ai/budget` immediately returned **200** again. The workaround holds.
5. Restored `tokens_per_full_user` to its original `12000000` once reads worked again; confirmed clean before leaving the shared installation.

## Cost

Every AI feature gated on the workspace budget (status, preview, and by extension any live model call that checks remaining budget first) fails closed the moment ordinary team growth — one new colleague joining — crosses a threshold an earlier, unrelated, and at-the-time-correct admin action set up. The write-time check gives false confidence: it tells the admin the value they just saved is fine, with no warning that it is one new hire away from breaking everything. And the failure surfacing as a bare `500` rather than a typed error means it will not read as "budget misconfigured" to whoever is paged — it reads as an ordinary server fault.

## Workaround

Confirmed live: deactivate the full seat that pushed the count over the threshold (or any other live full seat), which brings `eligible_full_users * TokensPerFullUser` back under the cap and restores both reads and the ability to submit a real correction. Until fixed, keep `TokensPerFullUser` well under `MaxMonthlyTokens / (today's full-user count + realistic headroom for growth)`, not merely under today's exact line.

---
Found by [margince-qc](https://github.com/margince/margince-qc), verifying a product-side 5-agent review of PR #5674 ("feat: clarify AI allowance and model administration"). Traced through source first at the PR's own merge commit (our `core` pin was one commit behind at the time), then confirmed live end-to-end after bumping past it — including the write-time guard the source-only trace could not see, and the bare `500` the source alone would not have shown.

### LINKED ISSUE #5694: The AI-budget resume job reports failed on every tick if a domain-triage site read is ever budget-deferred
# The AI-budget resume job reports failed on every tick if a domain-triage site read is ever budget-deferred

Not yet reproduced live — traced through source only, at core `798243a0ddb782219b3db738206d763aedfbd528` (PR #5674, "feat: clarify AI allowance and model administration"). No live run behind this one; say so plainly.

`resumeSiteReadAuthority` (`backend/internal/compose/jobs_aibudget.go`) decides whether a budget-deferred `site_read` row may resume:

```go
func resumeSiteReadAuthority(ctx context.Context, tx pgx.Tx, users *identity.Service, read contacts.BudgetRead) error {
	requesterCtx := ctx
	human, ok := principal.HumanUserID(read.Requester)
	if ok {
		...
	} else if read.Requester != systemAutoEnrichActor {
		return fmt.Errorf("site read %s: requester cannot be resolved", read.ID)
	}
	return contacts.RequireSiteReadAuthority(requesterCtx, tx, read.CompanyID, read.TargetKind)
}
```

It accepts a human requester or exactly `systemAutoEnrichActor` ("system:capture_auto_enrich"). A domain-triage read is started with a **different** system actor, `systemDomainTriageActor` = "system:domain_triage" (`compose/capturedomaintriage.go:33`, confirmed used as the `requested_by` in `StartDomainTriageSiteRead`, `capturedomaintriage.go:147`). Such a read falls into the `else if` branch and returns a plain, non-sentinel `fmt.Errorf`.

That error reaches `storekit.RecoverPages` (`platform/database/storekit/recover.go`), the paginator `ResumeBudgetReads` (`modules/contacts/sitereadbudget.go:25`) runs on. `RecoverPages` treats exactly two error kinds as "park it, not a failure" — `apperrors.ErrPermissionDenied` and `apperrors.ErrNotFound` — and joins anything else into the sweep's own returned error:

```go
if errors.Is(err, apperrors.ErrPermissionDenied) || errors.Is(err, apperrors.ErrNotFound) {
    slog.DebugContext(...) // parked, not counted
} else {
    failures = append(failures, err)
}
```

A plain `fmt.Errorf` is neither sentinel, so it is counted. Each row resumes inside its own savepoint, so *other* deferred reads in the same page still process correctly — this is not a total sweep halt. But the mismatched row itself never advances: `ResumeBudgetReads`'s own resume callback only bumps `next_attempt_at` to `now()` *after* a successful resume, and this one never succeeds, so it stays exactly where its original deferral put it and keeps matching the query's `next_attempt_at > now()` clause on every subsequent run. `AIBudgetResumeArgs.InsertOpts()` sets `MaxAttempts: 1` with a comment saying the next minute's tick is the retry — so the job reports failed, every minute, for as long as that one row's deferral window is still ahead of `now()`.

## Observed

Not measured live. Source citations only:

| Claim | Where |
|---|---|
| `resumeSiteReadAuthority` accepts human or `systemAutoEnrichActor` only | `compose/jobs_aibudget.go`, the function itself |
| Domain-triage reads carry `systemDomainTriageActor`, a different value | `compose/capturedomaintriage.go:33`, `:147` |
| `RecoverPages` treats only `ErrPermissionDenied`/`ErrNotFound` as parkable | `platform/database/storekit/recover.go` |
| A failed row's `next_attempt_at` is never advanced, so it re-matches next tick | `modules/contacts/sitereadbudget.go:25-49` (`ResumeBudgetReads`'s own resume closure) |
| The job retries every tick with `MaxAttempts: 1` | `compose/jobs_aibudget.go`, `AIBudgetResumeArgs.InsertOpts()` |

## Cost

If the trace holds: the one domain whose triage read got deferred never gets its company-or-not decision resolved by this recovery path (it would need the ordinary daily sweep instead, if that path does accept the actor correctly — not checked here). Separately and regardless of that one row's own fate, the `ai_budget_resume` job's reported health goes red on every tick for as long as the row's deferral window is open, which would read as "the recovery sweep is broken" in job monitoring even though every other deferred read in the same run resumes fine.

## Workaround

Not yet confirmed. Traced only: the row would need its `requested_by` corrected directly, or the code fixed to accept `systemDomainTriageActor` alongside `systemAutoEnrichActor` in `resumeSiteReadAuthority`. No administrative UI path was found for either.

---
Found by [margince-qc](https://github.com/margince/margince-qc), verifying a product-side 5-agent review of PR #5674 ("feat: clarify AI allowance and model administration") at its own merge commit, traced through source rather than walked live — our own `core` pin was one commit behind this PR at the time.

tinnhlt-gra:
## Summary

Three independent AI-admin defects from the QC report, bundled into one PR per explicit direction:

- **margince#5693** (priority: high) — a `TokensPerFullUser` value valid for today's full-user count can overflow `MaxMonthlyTokens` once the workspace grows, and every admin read/write entry point (`ReadBudget`, `PreviewBudget`, `ReplaceBudget`, and — found in review — `ReadStatus`/`PreviewRouting`) computed the stored config's strict `MonthlyTokens` first and errored before ever reaching the correction. Split the budget snapshot into a strict path (writes, and real spend via `compose`'s `seatBudget.MonthlyTokenBudget`) and a saturating "observed" path (every read surface), so an over-cap workspace can still be read, diagnosed, and corrected end-to-end.
- **margince#5694** (priority: normal) — the AI-budget resume sweep only recognized a human requester or the auto-enrich system actor; a domain-triage-deferred site read hit an unrecognized-actor error every tick (other deferred rows in the same batch still resumed fine — each has its own savepoint). Fixed by reusing the existing `isSystemRead` predicate (already used by the page-ceiling check) instead of checking one actor's sentinel.
- **margince#5695** (priority: normal) — the AI routing screen's headline panel silently rendered nothing for a viewer holding `ai_routing:read` without also holding `ai_diagnostics:read`/`ai_budget:read`. No seeded role hits this (confirmed against `rbac-matrix.md`), but a custom role can. Replaced the silent `null` with an explained withheld-panel state (`AiFeaturesWithheldPanel`, reused — a second identical gap was found and fixed in the same component in `ai-admin.tsx`'s `AiFeaturesCard`).

## Review

Both a Fable and a Codex adversarial review ran against this diff. Codex found no blockers. Fable's one MAJOR finding — that `ReadStatus`/`PreviewRouting` still hit the strict path, leaving `GET /ai/status` 500ing during recovery — is fixed in the second commit; its rulebook MINOR (ticket numbers in test comments) is fixed too.

**Deliberately left as noted, not fixed** (Fable MINORs, judgment calls):
- The over-cap workspace still reds the resume sweep on every tick until an admin corrects the config (fix #5693's saturating reads don't change fix #5694's sweep, which still calls the strict spend path) — not a regression, pre-existing shape, worth a follow-up if it proves noisy in practice.
- The routing screen's "Advanced" `<Disclosure>` (a different, pre-existing gap than #5695's headline panel) still defaults collapsed with its only in-component opener behind the now-explained withheld panel — the explained panel is the right fix for the actual reported gap; defaulting it open for this one grant combination is a follow-on polish item.
- The withheld panel isn't gated behind a `/me` load-state probe the way `ai-health.tsx`'s equivalent is, so it can flash briefly before the real content on a cold load, same as `ai-routing.tsx`'s existing `AiRoutingCard`. Consistent with the surrounding code, cheap to fix later.

## Verification

- `go build ./...`, `go vet`, `go test ./internal/modules/ai/...` — clean.
- New integration tests (`aibudgetoverflow_integration_test.go`, `aibudget_triage_resume_integration_test.go`) seed state through real writers (identity deactivate/reactivate, `StartDomainTriageSiteRead`) and were confirmed by Fable to pass against a live dev Postgres, reproducing both original failures pre-fix.
- The specific `backend/gates` tests this diff touched (uniqueness-claims census, contact-vocabulary) pass after two review-driven wording fixes.
- **Local `make check-all` could not be completed in this session** — the shared dev machine hit repeated OOM kills from concurrent sibling agent sessions (confirmed via `vm_stat`, not a code issue) partway through the backend build step, three times in a row. Everything that ran before each kill (frontend suite, gates, test-lanes census) passed. Deferring the full lane to CI's isolated runners per explicit instruction.

## Test plan

- [x] Unit/package tests for all three fixes
- [x] New integration tests reproducing each original failure, seeded through real writers
- [ ] CI's full check-backend/check-fe/integration lanes (pending — this push skipped the local pre-push hook's heavy lane under memory pressure; CI is authoritative)

---
## Summary by cubic
Fixes three admin defects: over-cap allowances no longer lock workspaces out of recovery, domain-triage reads resume correctly, and routing readers no longer see a blank activity section.

**Bug Fixes**
- An already-stored allowance above the ceiling is now clamped at the ceiling on every read and on real spend, so over-cap workspaces can be diagnosed and corrected; only new writes still fail closed, and resubmitting an over-cap config unchanged is refused instead of reported as saved.
- The resume sweep now accepts every system-read lane through `isSystemRead`, including domain triage.
- Users missing either `ai_diagnostics:read` or `ai_budget:read` now see an explanatory, localized withheld panel on the admin and routing screens.
- Adds integration, unit, and frontend regression coverage for overflow recovery, triage resumption, permission handling, and preview reads.

<sup>Written for commit a19143c9ce0efe0721d2f98b8365548624401772. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/5709?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added an explanatory panel when AI activity details are unavailable because the viewer lacks AI diagnostics or AI quota read access.
  - Added German, English, and Vietnamese translations for the access explanation.
- **Bug Fixes**
  - AI budget information can now be viewed and corrected when stored usage limits exceed the supported ceiling, while new invalid budget changes remain rejected.
  - Automatic site-read recovery now supports additional system processing lanes.
  - Budget-deferred site reads correctly resume scheduling after recovery.


tinnhlt-gra:
## Post-push review round (craft-reviewer + security-redteam, required by this branch's stop hook)

Both reviewers converged on the same real gap: the original fix only saturated the **admin read surfaces** (`ReadStatus`, `PreviewRouting`, `ReadBudget`, etc.), not real spend. `compose`'s `seatBudget.MonthlyTokenBudget` — which backs every routed model call, the usage meter, and the budget resume sweep — still read the stored config through the strict, fail-closed path. So an over-cap workspace still had AI entirely down (every model call refused, resume sweep red every tick) while `ReadStatus` now reported a healthy band — the two sides of the same invariant disagreed, arguably worse than before.

**Fixed**: `seatBudget.MonthlyTokenBudget` now reads the stored config through the same saturating method the admin surfaces use. A clamp can only ever authorize *less* spend than configured, never more, so this is the conservative fix, not a weakening — `MonthlyTokens` (strict) still gates every NEW value being written, in `ReplaceBudget`/`previewBudget`.

**Also fixed** (security-redteam MEDIUM): `ReplaceBudget`'s revision-equality no-op short-circuit returned the saturated current snapshot for an unchanged-but-still-invalid config without validating it — resubmitting the same over-cap value read as "saved, healthy" instead of the `InvalidValue` a fresh submission gets. Validation now runs before the short-circuit.

Plus: an actionable overflow error message, remaining fix-narration comments reworded to state the invariant rather than narrate the patch, and the triage test's hand-rolled job insert now goes through the production `siteDeepReadInsertOpts` helper instead of duplicating its shape.

**Deliberately not changed** (recorded as follow-ups, not blockers):
- `loadBudgetInputs`'s five positional return values (craft MINOR — a `users`/`spent` transposition would compile silently); a named struct would be cleaner but is a larger diff for a function with one caller.
- The triage test's shared six-line job-enqueue shape with the pre-existing `aibudget_recovery_integration_test.go` (not touched by this PR) — a shared helper would be nice but touches a file outside this diff's scope.
- The over-cap condition still isn't visibly distinguished from healthy in the snapshot itself (no separate warning field) — now that serving works again under the clamp, this is UX polish rather than an outage, worth a follow-up issue if it proves confusing in practice.
- The overflow-recovery test's `UPDATE app_user SET password_hash` direct write (documented inline; no identity writer exists for "give a harness seat a password" today).

Local `make check-all`/pre-push hook: everything green except `./internal/compose/integration`, which exceeds its own 600s local budget independent of this diff (confirmed already over-budget on `main` before these changes) — CI shards this lane per-test and won't hit the same wall.

PR #5721, opened 2026-09-15
fix(companies): a live read tells the facts panel to refresh itself

### LINKED ISSUE #5659: A finished website read leaves the company's Facts panel saying "Nothing on file yet" until the page is reloaded
## What is wrong

A person starts a website read from a company's Profile tab and stays on the page. The read finishes, and "Website research" reports "12 pages read so far · 56 evidenced facts staged". Right below it, "Facts about this company" still reads "Nothing on file yet", and the Technology panel still reads "No technical reading yet". Neither updates until the person navigates away and back, or reloads.

From source (`fb96a6420`):

- `frontend/src/screens/companydeepread.tsx:196-215`: the read report polls every 3s while `queued`/`running` and stops on a terminal status. On the way to a terminal status nothing invalidates the facts query. The only invalidation in the start mutation's `onSuccess` (`:377`) is `site-read-latest`.
- `frontend/src/screens/companyfactspanel.tsx:69-70,121-132`: the facts panel reads `["company-facts", companyId]` once. The key is invalidated only by the panel's own edits (`:476`) and by `evidenceverdict.tsx:179`.
- `frontend/src/screens/companytechnical.tsx:67` reads the same key, so the Technology panel is stale the same way.
- `staleTime` is 30s (`app/queryclient.ts:21`) and nothing refetches on an interval, so staleness alone never refreshes it.

## Measured

Measured at `fb96a6420` on a dev stack, as a rep, on a company created with the domain `go.dev`, in a Playwright walk that stayed on `#/companies/{id}/profile`:

- A read pressed while the company's automatic read was still running reached a terminal status. `GET /v1/companies/{id}/facts` then returned facts written by `agent:deepread` after the press.
- No `Where "<value>" came from` evidence mark appeared for any fact during the rest of the walk's 10-minute budget. This happened on two separate runs. The first run's page snapshot, taken when it failed, shows "Nothing on file yet" beside "12 pages read so far · 56 evidenced facts staged". The second run's snapshot was overwritten before it was read.
- When the Profile tab was opened only after the facts already existed, the same walk found the marks at once.
- It also hides the NEW facts of a read, not only the first ones. With Profile opened after the automatic read had finished, a second read pressed from the page finished at "60 pages read so far · 62 evidenced facts staged". The API returned "Automation — The ability to support cloud deployment automation." from that read. The panel kept the automatic read's list without it until the page was reloaded.

## Cost

This is the moment a rep has just asked Margince to research a company. The panel that should show the result says there is nothing, next to a count saying 56 facts were found. A rep who trusts the empty panel concludes the read found nothing, or retries. No data is lost: the facts are stored, and a reload shows them.

## Workaround

Reload the page, or leave the Profile tab and come back, after "Website research" shows a finished read.

---
Found by [margince-qc](https://github.com/margince/margince-qc), rebuilding `UAT-ENRICH-1` (`company-enrichment`) at `fb96a642`. A red proof that pressed during the automatic read went red at the evidence mark, and the page snapshot showed the empty facts panel beside the finished read.

tinnhlt-gra:
## Summary

Closes #5659. A website read left the Facts and Technology panels answering
from before it ran until a reload: the read-report poller stops once a read
reaches a terminal status, but nothing told the shared `["company-facts",
companyId]` query (read by both `CompanyFactsPanel` and
`TechnicalProfilePanel`) to refetch.

**Note on this diff's shape**: the base fix for #5659 (invalidating the
facts query on the not-terminal→terminal transition in `SiteReadPanel`,
`frontend/src/screens/companydeepread.tsx`) landed on `main` at `671e0ef17`
from a separate, earlier pass on this same issue before this branch was
rebased — so it does not appear as a separate commit here. What this PR
actually carries is the review-driven follow-up on top of that base fix,
from a Fable review and a Codex review run against the full diff:

- **`factsKey` had three separate spellings** (`companyfactspanel.tsx`,
  `companytechnical.tsx`, `evidenceverdict.tsx`) instead of one shared
  export — a near-miss spelling anywhere would have invalidated nothing.
  Moved `factsKey` into `companytechnical.tsx` (a leaf all three modules can
  import without a cycle) and switched every writer to the one spelling.
- **The terminal-status check omitted `cancelled`**, which the backend
  also treats as a finished status
  (`backend/internal/modules/contacts/siteread.go`,
  `sitereadfinish.go`) — a cancelled read would never have refreshed the
  facts panel. Added `status === "cancelled"`.
- **Dropped a redundant `wasTerminal` ref** in `SiteReadPanel`: the
  effect's own dependency array (`[terminal, companyId, queryClient]`)
  already fires exactly once per transition, and the ref could go stale
  across a `companyId` change without a `terminal` change.
- **Tightened the regression test** (`companydeepread.test.tsx`) to use a
  distinct `read_id` for the "second read pressed while already on the
  page" case, so the assertion is driven by that read's own poll rather
  than a leftover poll of the first — both reviewers independently flagged
  the original version as not actually isolating that case. Also replaced
  a fixed `flush()` count with a `flushUntil` condition probe after the
  rewrite made the fixed count flaky.

## Test plan

- [x] `pnpm exec vitest run` on `companydeepread.test.tsx` (11 tests),
  `companyfactspanel.test.tsx`, `technicalprofile.test.tsx`,
  `companyfacts.test.tsx`, `companyscan.test.tsx`, `onboarding-facts.test.tsx`,
  `accountscan.test.tsx`, `evidenceverdict.test.tsx` — all green; the new
  test rerun 5x in a row to confirm it isn't flaky.
- [x] Watched the regression test fail against the invalidation call
  neutered, confirming it isn't hollow.
- [x] Full `make check` (backend + frontend) green.
- [x] Manual UAT on a live dev stack, browser-driven: started a second
  website read on a company's Profile tab, stayed on the page, watched it
  reach a terminal status, and confirmed via the Network panel that exactly
  one new `GET .../facts` fired with no reload.

---
## Summary by cubic
Closes #5659. A website read that reaches a terminal status now refreshes the Facts and Technology panels instead of leaving them answering from before the read ran. The base invalidation already landed on main; this PR carries the follow-up hardening from review.

- Consolidated three separate spellings of `factsKey` into one shared export from `companytechnical.tsx`, so a near-miss spelling can't silently invalidate nothing.
- Treats `cancelled` as a terminal status, matching the backend, so cancelled reads also refresh the panels.
- Removed the redundant `wasTerminal` ref in `SiteReadPanel`; the effect's dependency array already fires exactly once per transition.
- Tightened the regression test to key the report responder on the polled path, so the second-read assertion is driven by that read's own poll, and replaced a fixed flush count with a condition probe.

<sup>Written for commit e76042b521864a0ed7152a689ffc9f9bc2f7e7bd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/5721?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Cancelled deep reads are now treated as complete, ensuring related facts refresh and terminal results display correctly.
  * Improved deep-read status polling and refresh behavior for newly started reads.
  * Updated Profile navigation handling for more reliable behavior.


tinnhlt-gra:
screenshot-1789471121758-18

PR #5767, opened 2026-09-16
fix(moments): a reminder the system minted is not a promise the reader made

LarsGradion:
Reported from staging: an account's "What needs you" card said **"You owe them: Check in — no activity since 2026-09-08" / PROMISE OVERDUE** about a task the `no_activity_reminder` clock starter minted. Nobody promised a check-in; the product wrote a nudge to itself and every promise surface dressed it up as the reader's commitment.

## What changed

**A system-minted task is not a promise, anywhere.**
- `principal.SystemMintedID` is the predicate: `captured_by` is `system` or any `system:<job>`. `activities.NotSystemMinted` is its SQL twin, spelled from the follow-up resolvers' own constants.
- **company360**: the moment card's promise read excludes these rows **in the statement, before its bound** — a page of overdue reminders can no longer crowd the one genuine promise out of the read. That read also now honours its own 200-row cap; it was silently cut to the section's 25 by the shared `truncate`.
- **contact360**: the promise rungs filter the section rows by the same predicate.
- **review_commitments**: the seam sets the new `ListOpenTasksInput.ExcludeSystemMinted`, so the tool surface and the screens give one answer about what is owed. Task lists everywhere still show the reminders — they are real work, just nobody's promise.

**The "What this rests on" box on the needs-list card** now opens only on evidence that quotes a conversation. A task's evidence is its own subject, and a disclosure that opens on the headline restated answers nothing — so the box in the report disappears, while a claim-sourced promise keeps its verbatim quote. New `moment.stories.tsx` shows both states.

**Copy**: "Due 0 days ago and still open" now reads "Due less than a day ago and still open" (task and claim cards, both pages).

## Held by

- `TestASystemMintedReminderIsNeverTheOwedPromise` and `TestAPageOfRemindersDoesNotBuryTheOnePromise` (compose integration): seeded through the real activity writer under a namespaced system principal; the second drives the real `review_commitments` seam too. Both filter clauses and the seam flag were mutation-checked — each reverted alone reddens the test.
- `TestSystemMintedIDMatchesTheNamespaceAndNothingBesideIt` (principal unit).
- companytoday screen test for the disclosure-less task card.

## Along the way

Main was red on `TestAnEmailIsDrawnOnlyByTheCanonicalComponents`; claimed and fixed in #5760 (merged).

---
## Summary by cubic
Fixes promise surfaces treating clock-minted reminders as commitments the reader made. System-minted tasks no longer appear as "You owe them" items or open commitments, while task lists still show them as real work.

**Promise filtering**
- Tasks written by the system are identified by `captured_by` naming the `system` principal or a `system:<job>` id; `activities.NotSystemMinted` is the SQL twin of `principal.SystemMintedID`.
- `company360` excludes them in the statement before the bound, so a page of overdue reminders can't crowd out a genuine promise; that read now honors its own 200-row cap instead of being silently cut to the section's 25.
- `contact360` filters the promise rungs, and `review_commitments` and the handoff brief pass the new `ExcludeSystemMinted` input flag.

**Card and copy**
- The "What this rests on" box only opens on evidence that quotes a conversation; a task's evidence is its own subject and would just restate the headline.
- Promises hours past due now read "Due less than a day ago" instead of "Due 0 days ago."

<sup>Written for commit 91f3164601cee46d482607312be53eb27baa753f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/5767?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Owed-promise views and commitment reviews now prioritize human-filed promises over system-generated reminders.
  - Large numbers of reminders no longer hide the relevant promise in account and contact views.
  - Overdue messages now correctly say “Due less than a day ago” instead of “Due 0 days ago.”
  - Next-step lists now respect the requested item limit.

- **User Experience**
  - Evidence disclosures appear only when supporting quoted text is available, avoiding empty or misleading options.


LarsGradion:
The end-of-work review round (craft + security subagents) landed these changes in the amended commit:

- The integration test now seeds one reminder under the bare `system` id — the id the workflow engine actually binds — and one under `system:time-scan`, so both arms of the predicate are held. Before this, only the `LIKE 'system:%'` arm was tested and the equality arm (the one a real clock reminder hits) was unheld.
- `activities.NotSystemMinted` now binds its two values through the caller's `arg` closure like every sibling fragment helper; only the alias is formatted in.
- The promise read's 200-row bound is pinned by a new test (`TestTheAccountCardSeesThePromiseThatSlippedLast`: 26 overdue promises, the one that slipped last is past the section page and still leads the card).
- The overdue sentence now has one spelling, `owedwork.LateWords`, read by both record pages and unit-tested; the claim card keeps its own deliberate wording with a note saying so.
- `prepare_handoff`'s open-commitments read also sets `ExcludeSystemMinted` — a handoff brief means commitment in the promise sense, the same sense `review_commitments` uses two functions up.

Security review found no exploitable issues: both `NotSystemMinted` call sites pass compile-time aliases, the new WHERE term is conjunctive only (can only narrow gated reads), and `captured_by` is server-stamped so the predicate cannot be forged in either direction.

Issue #1859, opened 2026-08-19
The German pack does one of five jobs: e-invoicing, accounting export, conformity and trust artifacts are all unbuilt


LarsGradion:
Decided 2026-08-19 (option a): one plan covering the pack, rather than four separate efforts into the same place. The gap is stated in the status lines of both ADR-0042 and ADR-0049.

**What the jurisdiction seam promised.** Country-specific behaviour lives in its own compiled unit, and core never names a country. Five kinds of contribution: statutory retention floors, e-invoicing, conformity documents, trust artifacts, and accounting export.

**What ships.** One of five. `extensions/de/de.go` holds the German statutory retention floors as checked-in typed values with the statute named — commercial correspondence six years, accounting records eight, both anchored at calendar-year end. The seam itself, the pack registry and the core-neutrality gate all work.

**What does not.** The e-invoice field list, the accounting-export layout, the account-mapping tables, the conformity contributions and the trust artifacts. ADR-0049 additionally requires a pack to check in a pinned copy of any external standard it must conform to; the German pack vendors no standards at all today.

**The two that matter commercially.** A German customer with an accountant will want e-invoicing and a GoBD-compatible accounting export. Both are market gaps rather than defects — nothing is broken, and nothing can be sold into that requirement either.

**Why one plan and not four issues.** All four contributions land in the same unit, against the same seam, and several share source material — the account-mapping tables the accounting export needs are the same tables an e-invoice references. Planning them together costs less than planning them apart, and it surfaces which standard needs pinning once rather than four times.

**The CRA operator page belongs in this plan too** (#1801). It is jurisdiction-specific material an operator reads alongside the GoBD obligations, and #1801 already names the German pack as its home. Doing it inside this plan means one decision about how the pack presents operator-facing obligations, rather than one page inventing its own shape.

**First step is a plan, not code.** Which contributions ship first, which external standards need pinning and in what version, and whether the accounting export targets DATEV or a neutral layout. That last one is a product decision with real consequences and should not be made by whoever writes the first line.

dungbanguyen:
## Decision, 2026-09-03: DATEV, e-invoicing first, and pin both standards up front

Re-triaged during the `status: needs-decision` sweep.

Ruled by @dung.nguyen; @LarsGradion stays assigned and may overrule this in the thread.

### The ruling — the question the ticket said not to leave to the first line of code

**The accounting export targets DATEV.**

The ticket's own framing decides it: *"A German customer with an accountant will want
[...]

PR #5826, opened 2026-09-16
docs(security): the CRA reporting package, before it is needed (#1801)

### LINKED ISSUE #1801: The CRA incident-reporting package: our runbook, the operator's page in the DE pack, and a fast-track channel that beats a 3-day SLA
Follow-up to #1785, which was blocked on a scope decision until 2026-08-19. That decision separated two things that had been treated as one, and this issue carries the resulting work.

**The two cases, now distinct:**

- **A Margince vulnerability** — a weakness in our code. The 24-hour clock is ours as manufacturer: we report to ENISA and the national CSIRT.
- **A customer got hacked** — their incident, their duty, no clock for us.

The runbook exists for the overlap: **a Margince vulnerability actively exploited at a live installation.** Both duties fire together. Decided: notify operators that they must patch, do not name them in the filing.

## Three deliverables

**1. Our outbound reporting runbook.** Who files, to whom, on what template. The CRA staged structure — early warning at 24 hours, the fuller report at 72, the final report — each needs a filled-in skeleton, not a description of one. Name the applicable national CSIRT and record the contact route before it is needed. Walk it once as a tabletop; an untested runbook is a document, not a capability.

**2. The operator-facing page in the German/European jurisdiction pack.** Alongside the GoBD material, because the operator reading it is looking for an obligation. It answers: you believe you are being exploited on a Margince installation — here is the address, here is what we do within 24 hours, here is what remains yours to report. `extensions/de` is where jurisdiction-specific material lives.

**3. The fast-track intake channel.** This is what makes the page a promise rather than a leaflet.

`SECURITY.md:23` currently commits to acknowledgement within 3 business days and an initial assessment within 10. That is a reasonable cadence for an ordinary advisory and useless against a live exploit with a 24-hour clock running. The fast track needs its own named route and its own commitment, and `SECURITY.md` needs to say which is which so a reporter can tell them apart.

## Why priority: high with nothing broken

No data is wrong and nothing is failing. The obligation has applied since **11 September 2026** and the failure mode is discovering the gap during an incident, with the clock already running. That is not a thing to work out under pressure.

## Also found while scoping

`SECURITY.md:53` puts "issues in the separate specification repository" out of scope. That repository no longer exists. One-line removal, unrelated to the rest but in the same file.

dungbanguyen:
Closes #1801 for everything a document can close. The ruling ordered four items; three are here, and the fourth is named as unfinished rather than pretended.

## 1. The runbook — `docs/compliance/cra/README.md`

The operational reading of Article 14, written so nobody performs that reading with a 24-hour clock running.

The decisions it makes, because the regulation leaves them to the reader:

- **Awareness is when a maintainer HOLDS a credible report** — not when it is confirmed, reproduced, triaged or understood. There is no business-hours clause in Article 14 and the runbook does not invent one: a report that arrives at 23:00 on a Friday starts the clock at 23:00 on that Friday. Write the timestamp down before anything else; it is the field all three deadlines are measured against, and reconstructing it later from a mail header is the argument nobody wants.
- **The three cases stay separate**, as the 2026-08-19 scope decision settled them. Ours to file; the customer's own breach carries no clock for us; the overlap fires both duties at once — **notify operators that they must patch, name no installation in the filing**. Each template repeats that last line, because it is the one a person under pressure gets wrong.
- **File incomplete rather than late.** An early warning that says "we know it is being exploited and little else" is what 24 hours asks for. A late complete one is still late.

## 2. Three FILLED skeletons, not descriptions of them

`early-warning-24h.md`, `vulnerability-report-72h.md`, `final-report.md` — three documents, as ruled. The constant fields (manufacturer, product, contact) are filled in; the incident fields are bracketed slots with a line saying what belongs there and at what level of detail. A template whose first page has to be researched is a template that gets filled in wrong at 03:00.

The 72-hour one carries the row an operator's own CSIRT actually reads — *measures available to users* — with the instruction that it must be actionable without reading any other field. The final report carries a section Article 14 does not ask for: **what stops it recurring**, because a final report whose last line is "we fixed the bug" invites the same report again.

## 3. The tabletop walk

45 minutes, two maintainers, no infrastructure — with a scripted trigger, and the two steps that are the actual point: *fix T₀ independently and compare* (if the two timestamps disagree, the definition of "aware" is not working), and *name the recipient and the route without looking anything up* (if nobody can, the pre-registration step is not done whatever its checkbox says). Its outcome is recorded in the runbook.

## 4. `SECURITY.md` — the fast track

The single blended SLA was the problem: three business days is a reasonable cadence for an ordinary advisory and useless against a live exploit. There are **two routes** now, and the reporter chooses:

- Ordinary advisory — unchanged, 3 business days to acknowledge, 10 to assess.
- **`ACTIVELY EXPLOITED` in the title** — acknowledged within **4 business hours**, and we come back with what we need rather than a triage verdict.

A reporter on the second route is told the two things that follow, because they will otherwise read as rudeness: our own 24-hour clock starts when we read the report, so the questions come fast; and the advisory is likely published early, since secrecy about a weakness already in use protects the attacker rather than the operators who have to patch.

(The ruling's fifth item — the stale "separate specification repository" line at `SECURITY.md:53` — is already gone from `main`.)

- **The operator-facing page** belongs in the German jurisdiction pack (#1859), which decides once how the pack presents operator obligations rather than letting one page invent its own shape. Ruled that way on both tickets.
- **The pre-registration checklist is written as unfinished**, and says so in the document: registering on the Article 16 reporting platform, confirming the coordinator's current channel, and walking the tabletop are actions against external bodies and a calendar that no document can perform for itself. Naming the authority is done — **BSI/CERT-Bund**, as the coordinator for our establishment — and the submission address is deliberately *not* transcribed: an address copied into a document and never checked is worse than a document that says where to look.

`make check-backend` green (the docs-reachability, page-budget, public-references and vocabulary gates included).

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for reporting actively exploited vulnerabilities, including expedited acknowledgment and notification procedures.
  * Added a Cyber Resilience Act Article 14 runbook covering responsibilities, reporting deadlines, regulatory submission routes, operator notification, and preparedness exercises.
  * Added templates for 24-hour early warnings, 72-hour vulnerability notifications, and final reports.
  * Documented required report details, mitigation status, affected regions, and restrictions against naming installations, customers, or affected parties.


dungbanguyen:
All three review findings taken, and they were the same defect three times: a compliance document stating a conclusion without the basis that produces it.

1. **The two dates were run together.** Article 14 applies from 11 September 2026; Article 69(3) extends the reporting duty to products placed on the market before 11 December 2027. As one sentence that misclassifies everything placed between them. Now two dates doing two jobs.
2. **Which CSIRT is a rule, not a fact about us.** Article 14(7) selects the member state where product-security decisions are predominantly taken, with the fallback chain behind it. The page named BSI without saying which limb selects it, so the basis is stated and the pre-registration block re-checks it beside the address — an establishment moving is exactly what makes a mandatory notification go to the wrong authority without anybody noticing.
3. **The operator notice had a channel and no reach.** Article 14(8) is about impacted users, and a GitHub advisory reaches whoever watches this repository, which is not the set that runs the software. The advisory is the floor now: every operator we hold a route for is told directly, one maintainer owns the notice until it is sent, and an installation with no route is recorded as a gap in what we know about our own deployments rather than treated as discharged.

Gates green.

Issue #415, opened 2026-08-04
Employment reaches an account with no start date, so old mail resolves to a contact's new employer


LarsGradion:
`activities.OrgReachSet` / `OrgLinkedActivityExists` reach an account through the employer of the contact a message is filed against, filtered only on `r.ended_at IS NULL`. There is no comparison against the message's own `occurred_at`.

So mail exchanged with someone while they worked at OldCo reaches whoever employs them today. On a timeline that is arguably helpful. For the signal producers it is a claim nobody made: a `contract_ended` extracted from OldCo's conversation, or a `ghosted_thread` saying "we wrote and nobody answered", can be filed against NewCo, which was never party to any of it.

The fix is to bound the employment arm by the relationship's own window — resolve through employment that was live when the activity occurred. **It is not available yet:** `relationship.started_at` exists in the schema but nothing writes it. `people.plantEmploymentEdge` inserts `(workspace_id, kind, person_id, organization_id, is_current_primary, source, captured_by)` and no dates, and on this installation all 213 employment rows have `started_at` and `ended_at` NULL. A time-bounded arm today would resolve nothing at all and silently reintroduce the bug PR #413 fixed.

So this needs, in order:
1. a writer for `started_at` (capture's employment edge, and whatever the enrichment path becomes),
2. a decision on what to do with the rows that have no start date — treating NULL as "always" keeps today's behaviour, treating it as "unknown" drops the arm,
3. then the bound itself, applied to the walk once so the timeline, the roll-up and both producers move together.

What limits the blast radius today: the extractor refuses any conversation resolving to more than one account, and since nothing writes `ended_at` either, a contact in practice accumulates employers rather than moving between them — which makes their conversations ambiguous and skipped rather than misfiled. The ghosted rule has no such refusal and fans out per account by design, so it is the exposed one.

Found during the stop-time review of #413.

dungbanguyen:
## Decision, 2026-08-31: a NULL start date means "always"; bound only the dated rows

Re-triaged against `origin/main` at `8ecb691f1`. The defect is unchanged. Both
spellings of the walk still filter the employment arm on the relationship's
END alone, never against the activity's own `occurred_at`:

```
LEFT JOIN relationship r ON r.person_id = l.person_id AND r.kind = 'employment'
  AND r.ended_at IS NULL AND r.archived_at IS NULL
```

— `backend/internal/modules/activities/orgscope.go` (`orgArms`, feeding both
[...]

PR #5831, opened 2026-09-16
feat(contacts): the add-employment form asks when the job began (#1155)

### LINKED ISSUE #1155: No surface changes a person's company connections
The contact record page's rail now has a section listing every company a person is connected to — the current employer first and marked as current, former ones after it, each linking to its organization record. It is read-only.

There is no surface anywhere for CHANGING those connections, and the API has the operations for it: `createRelationship`, `updateRelationship` and `archiveRelationship` (`backend/api/crm.yaml`, the relationships block around line 4270).

**What a rep cannot do today**

Record that someone changed employer, end an employment that has ended, correct a wrong role on one, or switch which of two connections is the current primary. A person whose employer is wrong stays wrong on every surface that reads the employment — the header's subtitle, the brief's first sentence, the strip's open-deal reading through the committee.

**Why it was left out rather than added inline**

The other rows in that rail edit one scalar each and fit hover-to-edit. An employment does not: it is an organization to pick, a role, a start and an end, and a primary flag, with the rule that at most one connection is the current primary. That is a form with its own validation and its own states, not an inline field, and it deserves to be designed rather than bolted onto a list.

**Worth deciding first**

Whether ending an employment archives the relationship or sets `ended_at` (the two mean different things to anything reading history), and what happens to the deals and committee rows that hang off it when a person leaves. Those answers shape the form more than its layout does.

dungbanguyen:
Closes #1155.

## The surface the ticket asked for already exists

Walked the four verbs the ruling specified against `main`, because a ticket that describes a missing surface should not be closed on a claim:

| verb | where |
|---|---|
| add a connection | `AddEmploymentModal` → `POST /relationships` |
| end one (sets `ended_at`, does not archive) | `contactemployers.tsx`'s `end` → `patchEmployment(..., { ended_at: today() })` |
| correct one (role, dates) | `EmploymentEdit` |
| switch which is the current primary | the same form's primary box; the server demotes the other in the same statement, which is where the at-most-one rule lives |
| archive as a **correction** | `DELETE /relationships/{id}` behind a confirm, separate from ending |

The semantic distinction the ruling turned on — `ended_at` means *it finished*, archive means *this row was wrong* — is the one the code implements, and the deals and committee rows stay either way.

## What was actually missing

The ruling's "one thing to check while there":

> `started_at` […] This form is the surface where a human can supply one, so it is worth making the start date a **first-class field** rather than an optional afterthought: a dated employment is what lets old mail resolve to the right account.

It was not there. The add form asked for the company, the role and whether it is current, and never for when it began — so every employment a human creates starts undated, which is the input #415 is about: an employment with no window lets the account walk read three-year-old mail as belonging to the company this contact joined last month.

## The field

Optional, and **empty stays empty**. An unknown start is a real answer, and a form that defaulted to today would write a fact nobody holds onto the column the walk trusts.

`datePatch` **moves** into `employmentpatch.ts` rather than being copied: a reader who knows the month and not the day types `2024-05`, and the stored date is the first of that month *with the precision that says so* — a date rendered back as "1 May" when somebody only ever claimed "May" is the product inventing a fact. Two forms send those two columns now, and a second spelling would be two answers to what a month-long date means.

## Tests

Two, both about what is SENT, which is the only thing that reaches the walk:

- a typed `2024-05` sends `started_at: "2024-05-01"` with `started_precision: "month"` — **mutation-checked**, dropping the two fields from the body fails exactly this case;
- a start nobody gave sends neither field, rather than today.

`make check-fe` green.

## Summary by CodeRabbit

* **New Features**
  * Added an optional start-date field to the employment form.
  * Supports month-only dates, such as `2024-05`, while preserving the selected date precision.
  * Leaves employment records unchanged when no start date is entered.

* **Bug Fixes**
  * Improved consistency when saving employment start dates across create and edit forms.


dungbanguyen:
Both review findings taken.

**The cancelled date.** Right, and it is the worst field to leave standing: the modal is never unmounted — the list only flips `open` — so a start date entered for one employer and cancelled would have been sent for the next one, dated from the last. `close()` puts it back with the rest.

**The shape.** Also right, and worse than a refused request: `datePatch` appends `-01` to any seven characters, so `not-a-d` would have become a plausible-looking day rather than an error. Save is disabled unless the value is empty, a month or a day — checked through the one `validDateEntry` both forms now share, since a rule about what a date IS should not have one spelling in the form that corrects one and another in the form that creates one.

A test per finding, each mutation-checked.

Issue #850, opened 2026-08-10
Consent: qualifying events are derived at send time; capture and the inquiry/in-person paths should write them


LarsGradion:
ADR-0098 D2 lists four events that flip `business_correspondence` to allowed: an inbound message, an inquiry they initiated, an active deal with their organization, or a recorded in-person exchange.

**Today only the inbound-message arm works, and it is DERIVED at send time** (`inboundQualifyingEvent` in `internal/modules/consent/verdict.go`): the gate reads the person's timeline for the newest inbound activity and, on the transmit path, stamps a `consent_qualifying_event` row before relying on it (`RecordDerivedQualifyingEvent`).

Deriving was the right call for shipping — every mailbox already captured contains the evidence, and a model that only recognised events written after this build shipped would tell a rep they may not answer somebody who wrote to them last week. But it leaves three gaps:

1. **Capture should write the event when the inbound arrives**, not when somebody happens to send. The derived row is stamped late, so its `created_at` says nothing about when the person actually wrote.
2. **`inquiry`** (form, booking, call) has no writer. The booking and form paths know when somebody initiated contact.
3. **`in_person`** has no surface. ADR-0098 D2 wants "gave us their address at {meeting}" typed by a named human; the DDL supports it (`note` is required for that kind) but nothing offers it.
4. **`active_deal`** is not evaluated at all — an open deal with their organization should qualify.

Also open: **employer change resets correspondence qualification and the §7(3) flag** (ADR-0098 D4 — "the old relationship's bases do not transfer to a new company"). Nothing revokes either on an employment change today.

Current behaviour is safe (it under-allows rather than over-allows) but incomplete against the ADR.

dungbanguyen:
Read against the code. Current behaviour is as the ticket describes and safe — it under-allows — but this is not one change, and two of its five parts turn on rulings a patch cannot make. Labelling `status: needs-decision`, with what I think the split is, because the useful thing here is not another investigation of the same ground.

**Two parts are a lawful-basis question, not an engineering one.**

- **`active_deal`.** "An open deal with their organization" qualifying a person for correspondence is a claim about legitimate interest reaching an individual through their employer's commercial relationship. Which deal states qualify (open only? any stage? a lost deal reopened?), and whether the person must hold a seat on it or merely work at the company, are the substance of the basis — and getting them wide is the one failure mode that is not safe. `refuseUnreadableEvidence` and the outbound validators already draw a narrower line for documents; this would draw a wider one for correspondence.
[...]

PR #5833, opened 2026-09-17
feat(consent): a booking is the subject asking, and the record says so (#850)

dungbanguyen:
Part **2** of #850's ruling — the `inquiry` writer. Parts 1 and 3 are addressed at the end.

## What was missing

ADR-0098 D2 lists four events that make ordinary business correspondence lawful, and an inquiry the subject initiated — *form, booking, call* — is one. Only the inbound-message arm worked, and it works by **deriving** the event at send time from the timeline.

A booking leaves nothing for that derivation to read. Nobody wrote to us; they filled in a form and took a slot. So a rep answering somebody who had just booked a meeting with them was refused as writing to a stranger, and no surface said why. That is the ticket's "`inquiry` has no writer", and it is the reason the gap cannot be closed by reading harder later.

## Where it is written, and where it is not

**The public door writes it**, citing the meeting it just committed. The citation is what makes the basis answerable rather than asserted: somebody asking why this contact could be mailed is shown the meeting they booked. `source = captured` — no reading produced it and no human typed it; the door that took the booking knew.

**Only the public door.** The authenticated one books on a contact's behalf, and a rep arranging a meeting is not that contact initiating contact. Both doors share `CaptureBookingConsent`, which is exactly why the inquiry is not written there.

**After the booking, and reported rather than fatal.** The meeting is what the subject came for and it is committed by the time this runs — it has to be, since the row cites it. A basis that could not be stamped costs a rep a manual send later, which is the under-allowing side this model is deliberately on. Reported, not swallowed: an installation whose bookings stop producing a basis sees it in its log rather than in a rep's surprise.

## The two writers share their insert

The ruling asked for this in those words:

> there are two writers of one invariant — they share a helper or say why not, **in the code beside them**, not in the PR body.

`stampQualifyingEvent` is that helper, and the idempotency rule lives in it: `ON CONFLICT` on the source record, not a read-then-write check two concurrent callers race past. A booking already stamped as an inquiry is therefore not stamped again when a later send derives from the same activity.

## The store verb is ungated, and ratified

Its one caller is a public door with no session and no seat to grant anything to — a gate would make the row unwritable by the only caller entitled to write it, and the basis would then be missing for exactly the contacts who initiated contact. What bounds it is the fact it records: the door has committed a meeting this contact booked, in the same request, and the row cites it. The hand-recorded verbs beside it keep their `RequireHuman`, because those are a named human asserting something with no other evidence.

## Test

`TestAPublicBookingRecordsTheInquiryThatAuthorisesAnsweringIt` (integration, through the real public HTTP door): the row exists, it is an `inquiry`, its source is `captured`, and it cites the meeting that was actually booked. **Mutation-checked** — removing the write fails it with "no qualifying event was recorded for a contact who booked a meeting".

## Parts 1 and 3

- **Part 3 (`in_person`) is already built** — `RecordQualifyingEvent` with `handRecordedKinds`, the required note, `RequireHuman`, idempotency on (contact, kind, note, moment), and the route at `/contacts/{id}/consent/qualifying-events`. Nothing to do; noted on the issue.
- **Part 1 (capture writes the inbound event on arrival)** is untouched here and still open, deliberately: it is the one with the "the derived path must keep working" constraint, and it is its own PR as the ruling asks.

`make check-backend` green; the pre-push integration lane green.

## Summary by CodeRabbit

- **New Features**
  - Public meeting bookings now record the inquiry that authorizes responding to the contact, even when marketing consent is not provided.
  - Recorded inquiries are linked to the booked meeting for improved consent history and traceability.

- **Reliability**
  - A booking remains successful if recording the inquiry encounters an error; the issue is logged for follow-up.


dungbanguyen:
Taken, and it is the same rule the hand-recorded verb beside this one already states: *"an exchange you remember happened before now"*, refused because a forward-dated event authorises sending from a moment that has not happened yet.

I passed the meeting's start — routinely days out — as `occurred_at`. The store stamps its own clock now, and the moment left the seam entirely: the door is OBSERVING a fact rather than asserting one, so when the booking arrived is the store's to say rather than the caller's to supply.

The test asserts the event is not in the future, and is mutation-checked: dating it +48h fails it while the meeting is still a week out.

PR #5855, opened 2026-09-17
Retire an erased message's identities on the contact-erasure arm

### LINKED ISSUE #5808: Contact erasure leaves an activity_identity row pointing at the erased message
## What happens

`activity_identity` rows are deleted on the activity-content erasure path (`backend/internal/modules/privacy/activitycontenterasure.go`, in `purgeContentDerivedFrom`). **Contact** erasure does not go through it: it runs `redactSubjectTimeline` / `purgeRedactedActivityTraces` in `backend/internal/modules/privacy/erasure.go`, which never touches the identity table.

So after a contact erasure the message's external identities survive, pointing at a row whose content is gone.

## Why it matters

Erasure archives rather than deletes, so the foreign key's cascade never fires — the identity row is the only thing left holding the association. The immediate consequence is bounded: binding resolves LIVE rows only, so a later arrival of that message will not be handed the tombstone. But the row is still a durable record that a message with that Message-ID was here, surviving an erasure that was supposed to remove it.

## Where

- `backend/internal/modules/privacy/erasure.go` — `redactSubjectTimeline` / `purgeRedactedActivityTraces`, the path that misses it
- `backend/internal/modules/privacy/activitycontenterasure.go` — the delete that exists, for the shape to copy
- `backend/internal/modules/activities/sourceidentity.go` — `RetireIdentities`

## Suggested fix

Retire the identities on the contact-erasure path too. The two erasure arms already destroy overlapping sets of derived rows; this is one more, and the honest fix is probably to make both arms call one list rather than maintaining two — a second list is how this one went short in the first place.

navdeepplaya-alt:
`activity_identity` is the cross-door dedupe table: one external identity — an RFC Message-ID for mail, `icalUID + "/" + instance` for a calendar occurrence — resolves to exactly one activity, so a message arriving through capture and through an importer becomes one row rather than two.

Erasure **archives** an activity rather than deleting it, so that table's `ON DELETE CASCADE` never fires. The per-activity arm (`Eraser.purgeContentDerivedFrom` — nightly retention, statutory-floor expiry, controller release) knew that and deleted the identity rows itself. The subject-scoped Art. 17 cascade kept its own list of derived rows, and that list went short by exactly this table.

## What the survivor costs

Two things, and the second is not in the issue:

1. **Residue.** The row is a durable record that a message with that identity passed through this workspace, plus `attested_by` naming who filed it — surviving an erasure that certified the content destroyed. Bounded: `boundToKnownMessage` reads the incumbent `LiveOnly`, so a later arrival is never handed the tombstone and never learns the message was erased.
2. **A permanent squat on the dedupe key.** `ClaimIdentity` is `INSERT … ON CONFLICT DO NOTHING`, so the dead activity holds the key forever. Every later arrival of that message files its own row *and* fails to claim the identity — a second and a third re-arrival become two rows that can never fold into one. That half is a functional bug, not only a compliance one.

## The fix

One slice-scoped helper, `retireActivityIdentities`, in its own file, called by **both** arms — the per-activity arm's inline statement is gone. A second list is how this table went missing from one arm while the other had it, so there is now one list.

It deliberately does **not** reach rows the statutory commercial-correspondence floor shielded. Those keep their content, so they keep the identity that names it: retiring an identity whose message still stands would send the next arrival of that message to a second row while the first one is right there.

## Why the gates missed it, and what now holds it

`activity_identity` was never registered in `piiTables`. That registry is what `TestErasureAndSARReachEveryPIITable` proves the Art. 17 cascade reaches, so an unregistered table is invisible to it. Registering it is the fix's other half, and it is the census that will fail on the next omission.

Rewriting `activityerasureparity_test.go` into a table-level census was considered and rejected: it would duplicate `piicoverage` from the other side and still could not see tables reached through a seam (`purgeRawCaptures`, the payload purger).

The floor-shielded non-reach is **gated by a test, not asserted in a comment** — `erasure.go` has two nearby calls that pass `append(activitiesRedacted, activitiesHeld...)`, which is exactly the shape someone would copy when "completing" the identity call from the fuller list beside it. The new integration case was verified by planting that defect and confirming the assertion fails on it.

The test asserts the consequence rather than the row: after the erasure, the Message-ID is **claimable again**. That cannot be satisfied by a statement that deletes the wrong rows, and it covers both harms above at once.

## Fixed along the way

`activities.RetireIdentities` was a near-verbatim twin of the new helper — same act, same reasoning, same error phrasing — with **zero callers**, at this branch's base as well as its head. It can never become the shared helper, because a module never imports a sibling, so it was a permanent decoy for the next author who greps the table and stops looking. Deleted, along with the equally uncalled `IdentitiesOf`, whose doc claimed an erasure role nothing performed. The new file states why the write lives in `privacy` as ratified cross-store SQL rather than as a call into the table's owner.

## Two deliberate calls, both cheap to reverse

**`sarRead: false`.** An identity key is a dedupe arbiter rather than something a subject can read, and the messages themselves are already exported in full — so handing back their transport identifiers adds nothing to the package. A sweep of the whole tree confirms nothing exports the table today, and `sarmessages.go` never selects `activity.source_id`, so the identifier is not in the package by another route either. The counter-argument is real — Art. 15 leans inclusive, and the key is data about a message the subject was party to. **Say so and it becomes one field plus a census re-run.**

**Anonymization is ratified in `clearedOnlyByTheEraser`, not made to retire identities.** Anonymize issues no `UPDATE activity` at all, so the message and its text still stand afterwards; an identity names a message that still stands. That is the same invariant the floor-shielded rows rest on — an identity is retired with the content it names, and not before. The register already ratifies `activity` itself on exactly this ground.

## Things a reviewer will reasonably ask

- **The fourth content-destroyer is not a gap.** `activities.RedactCapturedNoiseTx` empties an activity's text and deliberately keeps its natural keys as re-capture tombstones. Keeping the identity is consistent with that contract: it is a noise disposition, not an Art. 17 erasure. The parity census structurally cannot see it — the census's declared subject is the `privacy` package.
- **Only one of the duplicated derived-row spellings was unified, and that is the right stopping point.** `field_provenance` and `activity_reply_verdict_history` are deliberately different acts with different scopes and reasons written down beside them; unifying those would be a regression. `embedding` is a true twin, but it has a third writer in `capturenoise.go` and its own module-boundary problem, so it is its own change.
- **No migration is needed.** The FK already declares `ON DELETE CASCADE`; the bug is that neither arm ever DELETEs the activity. That is a code fact, not a schema one, and nothing here changes the table's shape.

## Filed separately

`activity_identity`'s primary key is `(identity_kind, identity_key)` with **no `workspace_id`** — it is global. An un-retired identity therefore squats that Message-ID for every workspace on the installation, not just the erased subject's. Pre-existing and out of scope here, but it makes this defect wider than it looks and deserves its own issue.

Closes #5808

---
## Summary by cubic
Fixes the subject-scoped contact erasure so an erased message's external identities are retired, and the dedupe key is claimable again.

- Both erasure arms — the per-activity content purge and the Art. 17 cascade — now retire identities through one shared helper; the cascade previously never touched the table, so an archived message kept its Message-ID and squatted the dedupe key forever.
- Floor-shielded rows keep their identities; the integration test asserts the consequence — an erased Message-ID binds to the next arrival — and that held rows still answer to theirs.
- Deletes uncalled `RetireIdentities` and `IdentitiesOf` from `activities`, which were a second spelling of the same act.
- Registers the table in the PII census gates so a future omission fails the reach tests; anonymize deliberately does not retire identities because it redacts no content, ratified in the gate waivers.
- No migration needed — the bug is that erasure archives, never deletes, the activity. Pre-existing and filed separately: the table's key is global (no `workspace_id`), so an un-retired identity squats across workspaces.

<sup>Written for commit 9670a8adb5cdcf8aa51bf7aefdaffe0d19938dfd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/5855?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


navdeepplaya-alt:
Closing: superseded by #5903, which merged as 05a5d629b and closed #5808.

The two changes converged on the same fix independently — same helper name
`retireActivityIdentities`, same `DELETE FROM activity_identity WHERE activity_id = ANY($1)`,
same two callers (`purgeContentDerivedFrom` with the single id, `purgeRedactedActivityTraces`
with the redacted set), and the same insertion point in the cascade. Only the filename differs
(`activityidentityretire.go` there, `erasureidentities.go` here). There is nothing here to rebase.

The two cleanups this PR also did are moot as well: `activities.RetireIdentities` and
`IdentitiesOf` are gone from `main`, removed by the capture-side work that was reworking
`sourceidentity.go` in parallel (#5806, #5843, #5864, #5917).

## One part of this PR did not land, and it is the census

`activity_identity` is still **not registered in `piiTables`** (`backend/gates/piicoverage_test.go`).
#5903 updated `piisqlreader_test.go` — the cascade *file* list — and the anonymize waiver in
`contactscrub_test.go`, but not the registry.

That registry is hand-curated; nothing forces a table into it, and
`TestErasureAndSARReachEveryPIITable` iterates `for table, h := range piiTables`. An unregistered
table is invisible to it. That is exactly the mechanism that let this defect exist, and on this
table it is still open — so the gate that should fail on the next omission here still would not.

Two smaller gaps, for whoever picks that up:

- The merged integration test counts identity rows to zero. It does not assert the
  **consequence** — that the Message-ID is claimable again by a later arrival. `ClaimIdentity` is
  `INSERT … ON CONFLICT DO NOTHING`, so a surviving row squats the dedupe key permanently and every
  re-arrival becomes a duplicate that can never fold. That is a functional harm, not only a
  compliance one, and #5808 and #5903 both describe the harm as residue and call it bounded.
- Nothing covers the **floor-shielded** direction — that rows the statutory correspondence floor
  held must *keep* their identities, because an identity naming a message that still stands must
  not be released. `main` behaves correctly (it passes the redacted set), but the boundary is
  untested, and `erasure.go` has an `append(activitiesRedacted, activitiesHeld...)` call sitting
  beside it that is exactly the shape someone would copy when "completing" this one.

Follow-up PR for those three, off `main`, rather than a re-land of this branch.

The separate concern this PR filed — `activity_identity` being keyed installation-wide — was
#5856, closed as not planned: one installation holds one company (ADR-0061), so the cross-workspace
squat it described cannot occur.

PR #5871, opened 2026-09-17
feat(search): imported history does not rank as first-party testimony

LarsGradion:
#5849 gave imported activities a place to record who wrote them in the system they came from. This corrects the one place where `captured_by` is true and answers the wrong question.

## The problem

The HubSpot migration ran through the public API as an administrator, so all 34,521 of its activities carry a `human:` prefix in `captured_by`. That is accurate — an administrator did run the import — and it is what the retrieval trust ladder reads. So every imported row scored **T0, a human statement**: another company's CRM history ranking above what colleagues here actually wrote, inside the graph walk that answers questions about an account.

## The fix

`trustOfWriter` takes a second argument and checks it first. A row that came from another system is captured external content whoever ran the import, so it answers T2 regardless of the `human:` prefix.

The marker is **`source_system IS NOT NULL`**, and getting to that took two wrong answers first, both caught in review:

- **`source_author_id` alone** misses the 36 of the import's 76 authors who never held a seat here — every colleague who had already left. Their correspondence is exactly what this change exists to demote.
- **Either author column** still misses more. The attribution repair *skips* a row whose author it cannot resolve, leaving both columns null, so an imported activity with an unmatched author would have kept full human-statement trust.

Both of those test whether attribution **succeeded**. Only `source_system` tests where the row came from, which is the question the ladder is asking. It is also the column the author columns' own CHECK constraint ties them to, and a native write never sets it, so nothing typed here is demoted by accident.

The column is read as one boolean rather than as a value: who the author was belongs to the timeline, not to a ranking weight.

## What is deliberately unchanged

Recency still outweighs trust. `TestFreshCapturedContentStillBeatsAStaleHumanNote` is untouched and still passes: a captured email from this morning is more use than a human note from last quarter, and the ladder remains a tie-break between comparable ages rather than an override.

## Tests

`TestTheTrustLadderReadsTheWriterNotTheChannel` gains two cases — an imported row recorded by the administrator who ran the import, and one an agent touched — both T2.

`TestAnImportedNoteDoesNotOutrankOneWrittenHere` is the unit property: same age, same human `captured_by`, and the imported one still scores lower.

`TestImportedHistoryRanksBelowWhatColleaguesWroteHere` proves it against a database through the real walk, in three cases: the author holds a seat, the author left before we had seats, and **the author could not be resolved at all** — the case the author-column predicates missed.

**The ordering is isolated deliberately.** Both notes take one timestamp read from Postgres rather than two separate `now()` calls, so the recency term is genuinely equal. The imported note is seeded *first*, so its lower `ids.NewV7()` wins `sortAndTrim`'s id-ascending tie-break — meaning it wins every tie there is, and only the trust term can put the other note ahead. Verified by mutation: disabling the trust term fails all three subtests.

`make check-backend` is green.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved activity ranking so imported CRM history appears below notes written directly in the workspace.
  - Imported notes are now treated as external content even when associated with a human or automated importer, preventing them from outranking first-party notes.
  - Updated ranking behavior consistently across imported records with complete, partial, or missing author details.


LarsGradion:
Parked as draft — not mergeable as it stands.

The predicate demotes **native sent mail**. A colleague composing an email inside Margince gets `source_system = 'email'` for dedup (outboundmessage.go), so their own sent mail would drop from T0 to T2 — a new over-demotion this change introduces.

Three predicates have now been wrong in the same way:

| attempt | misses |
|---|---|
| `source_author_id IS NOT NULL` | the 36 of 76 HubSpot authors who never held a seat |
| either author column | imported rows whose author the repair could not resolve |
| `source_system IS NOT NULL` | nothing — but it catches native sends too |

The first two test whether *attribution succeeded*. The third tests where the row came from, but `source_system` is overloaded: it is also the mail identity a native send carries, and the idempotency key internal writers stamp.

The codebase already has the right idea — `provenance.ReservedSourceSystemPrefix = "mirror:"` namespaces a source_system only an import may write, refused on every client path. The HubSpot import did not use it: all 34,521 rows carry plain `hubspot`.

So the fix needs an explicit import marker, and that is a decision about the import path rather than about this file. Held until then.

LarsGradion:
Superseded by the work on `feat/import-marker`, which takes a different predicate.

This draft ranked a row as imported when `source_system IS NOT NULL`. That column is not exclusive to imports: `activities/outboundmessage.go` stamps `email` on a message this installation sent, and `activities/requesttask.go` stamps its own reminder identity. Both are first-party writes, so this predicate would have demoted them to captured-external — the fix for a mis-ranked timeline mis-ranking two more kinds of row.

The replacement tests `provenance.ReservedSourceSystemPrefix` (`mirror:`), the namespace that already exists for exactly this and that every client-facing create path refuses. Closing in favour of that.

PR #5876, opened 2026-09-17
Refuse a deal's company move while its agreements name another company (#4986)

### LINKED ISSUE #4986: Reassigning a deal's company leaves its contracts naming the old one, and the deal arm then publishes them
Found in review of #4961. **Pre-existing** — that PR did not create this state, but it adds a second route into it, which is why it is filed rather than folded in.

## The state

`contract.organization_id = A`, `contract.deal_id = D`, and `deal D.organization_id = B`.

`ensureLinksShareOrganization` refuses to *create* that pairing. Nothing stops it being *reached* afterwards: `applyDealLinkPatches` (`backend/internal/modules/deals/deal.go`) writes `organization_id` behind `auth.EnsureLinkTarget` on the new company alone, and never asks what agreements already hang off the deal.

## Why it matters

`VisibleClause` in `contracts/visibility.go` judges a deal-anchored contract by its **deal alone** — deliberately, because widening to the company would hand a caller agreements attached to deals they cannot see. The consequence runs the other way here: once the deal names B, everyone who can see B's deal can read A's agreement, and the row's own `organization_id` is not consulted. That is the exact leak `ensureLinksShareOrganization`'s comment describes, arrived at from the other end.

## Two routes in

1. Create the deal for company A, add a contract, reassign the deal to B. **Reachable today**, before any recent change.
2. Create a company-less deal, attach a contract for A, then assign B. This second one opened when #4961 fixed the 500 that made attaching a contract to a company-less deal impossible.

## What is needed, and the decision in it

The check is a read of `contract` from inside the `deals` module, so it wants a compose seam rather than an import — `EnsureProjectAttachable` and `EnsurePartner` are the two seams that already have exactly this shape, injected into the same function.

The product decision is what the seam should DO, and it is not obvious:

- **Refuse the reassignment** while live contracts name another company. Safe, and it means a rep who mis-filed a deal must detach or re-file the paperwork before they can correct the company — on a record where the paperwork may be the thing they are trying to fix.
- **Reconcile**: move the contracts with the deal. Silently rewrites which company an agreement is with, which is not a thing a deal edit should be able to do.
- **Detach** the contracts from the deal, leaving them anchored to their own company. Preserves both records' truth and loses a link a human made.

A test seeding the state directly would show the read-through today; there is no need to reach it through either route to prove the exposure.

Labelled `priority: high` because it is a cross-tenant read of a commercial document rather than a display fault, but it needs the decision above before it can be closed.

navdeepplaya-alt:
Closes #4986.

## What was wrong

`contract.company_id = A`, `contract.deal_id = D`, `deal D.company_id = B`.

`ensureLinksShareCompany` refuses to *create* that pairing. Nothing stopped it
being *reached* afterwards: `applyDealLinkPatches` wrote `company_id` behind
`auth.EnsureLinkTarget` on the new company alone and never asked what was
already filed against the deal.

`VisibleClause` judges a deal-anchored contract by its **deal alone** —
deliberately, because widening to the company would hand a caller agreements
attached to deals they cannot see. Run from the other end, that same rule means
everyone who can see B's deal reads A's agreement, and the row's own
`company_id` is never consulted.

## What this does

The decision on the issue: **the reassignment is refused** while agreements name
another company. Not reconciled, not detached behind anyone's back.

- **A compose seam, not an import.** `deals.EnsureContractsShareCompany`
  (`deals/contractseam.go`), answered by `contracts.EnsureDealContractsShareCompany`
  and bound in `compose/installseam` beside `EnsurePartner` and
  `EnsureProjectAttachable`. Un-injected it refuses, and the existing
  `Installation` census (derived from the struct) now calls it.
- **Both routes land in one writer.** Confirmed rather than assumed: route 1
  (created for A, reassigned to B) and route 2 (company-less deal, contract for
  A, then assigned B) both reach `applyDealLinkPatches`, and both are covered by
  the one check. Integration cases for each.
- **The refusal is an instruction.** It names the blocking agreements (contract
  number, else title), states the rule, and says what to do — 422
  `deal_contracts_cross_company` on `company_id`. The list is capped at five and
  says when it was cut.
- **Archived agreements count.** `readContract` does not filter a contract's own
  `archived_at` — its clause is about the *anchor* — so an archived agreement on
  a moved deal is the same read-through as a live one. Covered by a case.
- **The deal row is held before the agreements are read.** Every contract writer
  takes a share lock on the deal row to read its company (`auth.EnsureLinkTarget`),
  so holding it here serializes the two directions instead of letting both
  commit against a world the other had left. `auth.HoldWritableLive` is the
  probe and the hold as one, taken after the company probe so the
  company-then-deal order this transaction already follows is unchanged.
- **A test that seeds the state directly** and reads it back: a rep working
  Contoso is handed Acme's agreement — title and terms — with the counterparty
  withheld, while the same company's unanchored agreement answers not-found to
  the same reader. That case stays meaningful if a fourth route appears.

## Fix the invariant, not the call site — the audit

Every writer of the three columns, and why each is covered:

| writer | verdict |
| --- | --- |
| `deals.applyDealLinkPatches` (`deal.company_id`) | the check, added here |
| `clearableDealColumns` → `company_id` (clear to NULL) | **allowed, deliberately.** A deal naming nobody publishes its agreements to exactly the readers its own scope already admits — which is why the filing check admits a contract on a company-less deal. Covered by a case. |
| `contacts.merge_company` (`deal.company_id` + `contract.company_id`) | covered by construction: one transaction moves both sides A→B, so rows that agreed before agree after |
| `contracts.createContractTx` (insert) | `ensureLinksShareCompany` |
| `contracts.contractPatch` → `deal_id` | `ensureLinksShareCompany` |
| `contracts.Renew` | inherits the predecessor's company and goes through `createContractTx` |
| `contracts` `deal_id` clear (new, below) | always safe: the agreement falls back to its own company's arm |

## Two things beyond the issue, flagged

**1. An agreement can now be detached from its deal.** The decision says the
refusal is paid for by naming what to do — "detach them from the deal, or
re-file them against the new company". Neither existed: `UpdateContractRequest`
has declared `deal_id` nullable all along, but the null decoded to a nil
pointer, read as "not supplied", and answered 200 having changed nothing. A
refusal a rep cannot clear would have been the objection landing, and the
honest remedy (archive the misfiled agreement, re-record it against the new
company) is itself blocked, because archived agreements count.

So `contracts` joins the clearable census: `clearableContractColumns`, the
matching `compose/clearablefields.go` entry, and the gate holding the two equal.
Only `deal_id`, because only `deal_id` has something asking for it.

*Wire consequence, deliberate:* an explicit null on another `PATCH /contracts/{id}`
field now answers 422 naming the field (`field_not_clearable`) instead of being
silently ignored. That is what every other record type in this tree already
answers, and `NotClearableError`'s own reasoning is that a success the caller
cannot trust is worse than a refusal. Say the word and I will split it out.

**2. One waiver went.** `internal/modules/deals:lockedAcquisitionSource`'s
write-authority waiver stopped matching, because the probe it said was taken a
frame up is now reachable in the same package view. What it explained is worth
keeping, so it moved onto the function itself.

## Considered and not done

A constraint trigger mirroring `deal_project_same_company` would be the fitness
function here — it fires on every writer, including ones nobody has thought of.
It is not in this PR: the decision on the issue names the seam, the message
quality needs the Go-level read anyway, and a second spelling of one invariant
needs its own decision about which one is authoritative. Happy to follow up if
that is wanted.

## Verification

`make check-all` (both halves plus the real-Postgres lane) green, `craft static
--strict` clean on every touched file.

---
## Summary by cubic
Refuses to move a deal to a company its agreements do not name, closing a read-through where the move would publish another company's agreements to the new company's readers.

- Moving `company_id` now answers 422 `deal_contracts_cross_company` while any agreement filed against the deal names a different company; the refusal names the blocking agreements (up to five, contract number or title) and says to detach or re-file them.
- Archived agreements block too, because they remain readable through the deal.
- `deal_id: null` on `PATCH /contracts/{id}` now detaches the agreement; previously the null was read as "not supplied" and silently ignored. This is the remedy the refusal names.
- An explicit null on any other contract patch field now answers 422 `field_not_clearable` instead of being silently ignored.
- Clearing a deal's company remains allowed: a deal naming nobody publishes its agreements to exactly the readers its own scope already admits.

<sup>Written for commit 191461a4354ec1b86a3afee8cb9a391b21ab0e72. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/5876?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Deal company changes are now prevented when linked agreements belong to another company.
  - Error responses identify conflicting agreements and explain how to resolve them.
  - Agreements can be detached from a deal by clearing the deal link.

- **Bug Fixes**
  - Archived agreements are also considered when validating company changes.
  - Prevents agreements from being exposed to the wrong company after a deal reassignment.

- **Documentation**
  - Updated CRM API and module documentation to describe these relationship rules and validation errors.


dungbanguyen:
@navdeepplaya-alt please fix the CI first

navdeepplaya-alt:
@dungbanguyen  can you please merge now, hope no one else pushes again

PR #5922, opened 2026-09-18
test(capture): the statutory hold on a take-over is pinned, not patched

### LINKED ISSUE #5919: The asserted take-over and a statutory hold: already covered, now tested
## What happens

`TakeOverAssertedActivityTx` (`backend/internal/modules/activities/assertedtakeover.go`) overwrites an activity's `subject`, `body` and `captured_by` without testing `restricted_at IS NULL`. It locks `LiveOnly`, so an archived row is refused — but a row under a statutory hold is neither archived nor excluded, and the take-over rewrites it.

## Why it matters more since #5917

Before #5917 the take-over was reachable only for a row the SAME seat had captured — a seat overwriting their own held row, which is a smaller problem. #5917 added the cross-seat arm: an import now binds to the seat whose provider-attested address it names, so the take-over can now rewrite a row another seat filed.

The address proof in that change is narrow on purpose (OAuth-attested `account_label` only, `status = 'connected'`, exactly one seat). This is not a claim that the proof is bypassable. It is that the take-over should not depend on the proof being perfect for a hold to be respected — a hold is the one obligation that outranks every ordinary path, and it should be enforced where the write happens.

## The sibling that gets it right

`seatDeliveredTx`'s member-bound arm (`backend/internal/modules/capture/importrow.go`) carries the condition and states the reason:

> restricted_at IS NULL for the reason every other reader of this table carries it: a row under a statutory obligation is out of every ordinary path, and writing an import row against one would hand a seat a grant on a message the obligation has already taken away from them.

Others that pair it: `capture/aliasadopt.go`, `capture/pendingreview.go`, `capture/pendingclaim.go`, `capture/digestcounts.go`, plus the audience recompute and initial-audience paths.

`assertedtakeover.go` is the outlier.

## Suggested fix

Add `AND restricted_at IS NULL` to the `UPDATE` predicate, and return the same not-found the archived case returns so the caller's existing handling applies — `capture/sinkactivity.go` already maps `apperrors.ErrNotFound` from this function onto `skipInvisibleIncumbent`, which advances the watermark rather than stalling the mailbox.

Alternatively `storekit.LockRow` could grow a mode that means "live and not restricted", if other callers want the same pair. That is the larger change and probably wants its own decision.

## Test to write first

A held row, a capture that would otherwise take it over, and an assertion that the subject, body and stamp are unchanged. Mutation-check it: removing the condition must turn it red. There is no such test today, which is why the gap survived.

Found by an adversarial review of #5917.

LarsGradion:
Closes #5919.

## What the issue asked for, and what it gets

#5919 reported that `TakeOverAssertedActivityTx` rewrites a row under a statutory hold, because its `UPDATE` does not test `restricted_at IS NULL` the way `capture/importrow.go` and its siblings do. It asked for the predicate plus a test.

The report is right that the predicate is absent and right that no test covered the case. It is wrong that the row gets rewritten. So this PR ships the test and **not** the predicate, and says why in the code.

## Why the predicate would be dead code

Four refusals already stand in front of that write, and the first one fires:

1. `ResolveBindableIdentity` answers not-found. `identityHolderIsLive` tests `archived_at`, a held row is archived, so capture never calls the take-over at all — it files the mailbox's own copy instead.
2. The take-over's own `LockRow(LiveOnly)` answers `ErrNotFound`, which capture already routes to `skipInvisibleIncumbent`, for a hold landing between that resolve and this write.
3. `activity_restricted_is_archived`, a non-deferrable row `CHECK`, makes held-but-live a state the `activity` table cannot hold. That is **why** the two above work on `archived_at` alone, and why both writers that place a hold (`privacy.PinToFloor`, the erasure's restrict arm) archive in the same statement.
4. `activity_refuse_restricted_mutation` raises on any `UPDATE` leaving `restricted_at` set.

I wrote the suggested predicate first, measured it against a real database, and found it could never decide an outcome — a fourth refusal behind three, unreachable by construction. That is exactly the dead code T3 forbids, and worse than absent: a reader would take it for the one doing the work. So it is out, and the lock's comment now states the hold refusal it already carries.

The alternative the issue floated — a `storekit.LockRow` mode meaning "live and not restricted" — is moot for the same reason. `LiveOnly` already means that.

## What is actually new

`TestAMessageUnderAStatutoryHoldIsNotTakenOver`, in `crossseatidentity_integration_test.go`, drives the **cross-seat** arm #5917 added, which is the arm that made the question worth asking: the rep has proven the mailbox, the admin's import names their address, and without the hold this is exactly `TestAnAdminsImportBindsToTheRepWhoseProvenAddressItNames`, which binds and rewrites. The hold is the only thing between the two outcomes.

It asserts the held row's subject, body and stamp are unchanged, that the hold is still in force, and that **the rep's own copy still landed** — a refusal must not suppress the mail in the mailbox that holds it. Without that last one the test would pass just as well against a capture that dropped the message.

## Mutation check

Done by hand, and the result is written into the test's comment rather than claimed:

- No single clause is load-bearing, so **no single-clause revert turns it red**. Each of the four refuses alone.
- Making `identityHolderIsLive` answer `true` unconditionally **and** relaxing the take-over's lock to `IncludeArchived` **does** turn it red, failing on the trigger. That pair — widening what the resolve may bind to, with the liveness filter that currently covers holds relaxed to match — is the realistic regression the test is worth having for.

## Also

The take-over's doc comment still claimed the resolve "answers only for a row the SAME SEAT captured". #5917 made that false by adding the proving arm, and that comment is what a reader would have relied on to decide this exact question. Corrected to describe both arms, and to say that the cross-seat one does move authority between seats.

## Verification

- `make check-go` — green
- `make check-gates` — green
- `make craft-static` / `make craft-prose` — 0 findings
- `make test-it DIR=backend/internal/compose` — full package green (194s)
- pre-push lane: full parallel integration suite, 52 packages, 0 skips

---
## Summary by cubic
Adds a regression test proving a take-over never rewrites a row under a statutory hold, and corrects the take-over's docs so readers don't look for a missing `restricted_at IS NULL` guard.

- The `restricted_at IS NULL` predicate from issue #5919 is deliberately omitted — four existing refusals already stop the rewrite, so it would be dead code.
- The test drives the cross-seat proving arm, asserting the held row's subject, body, stamp, and version are unchanged, the hold stays in force, and the rep's own copy still lands. The version check proves no statement touched the row, closing the gap where a rewrite writing identical values back would pass.
- The take-over's doc comment claimed the resolve only answers for same-seat captures; it now describes both arms and why the cross-seat proof is deliberately narrow.

<sup>Written for commit 759776dda959093bb0511bd3c50f2c9c9a74833a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/margince/margince/pull/5922?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Messages placed under a statutory retention hold are protected from mailbox synchronization takeover.
  - Held messages retain their original subject, content, ownership details, and retention status.
  - Mailbox owners continue receiving their own copies while the retained message remains unchanged.
  - This protects retained content from unintended updates during synchronization while preserving normal delivery of the owner’s copy.


LarsGradion:
The `version` assertion in 759776dda answers your merge-risk note directly: `trg_activity_updated` bumps `version` on every UPDATE of the activity table, so an unchanged version proves no statement reached the held row at all — stronger than any set of column comparisons. Read after the hold is placed, because placing the hold is itself an UPDATE and bumps it.

I did not use `occurred_at` as suggested. The take-over writes `subject`, `body` and `captured_by` only, so `occurred_at` would not have caught the case your note named.

On the Linked Issues warning: the `AND restricted_at IS NULL` predicate it asks for is unreachable by construction, and the PR body has the measurement. Short version — `activity_restricted_is_archived` is a non-deferrable row CHECK making held-but-live a state the table cannot hold, so `LockRow(LiveOnly)` refuses every held row before that UPDATE runs, and `ResolveBindableIdentity` refuses earlier still. I wrote the predicate first and removed it once I confirmed it could never decide an outcome; a dead guard that reads like the load-bearing one is what T3 forbids. The author of #5919 has independently verified all four refusals and corrected the issue to `priority: low`.

@coderabbitai full review

